PR #3, opened 2026-04-04
feat: add daemon session persistence and recovery

margvez:
## Summary
Add daemon-backed session persistence and recovery as a standalone feature.

## What this adds
- persist daemon session metadata to ~/.wmux/sessions.json
- persist periodic terminal buffer snapshots to ~/.wmux/buffers/*.buf
- recover live and suspended daemon sessions on startup
- restore renderer session/layout and scrollback IPC persistence
- reintroduce tests for StateWriter and ring buffer dump/load behavior

## Why
This is the daemon persistence/recovery functionality that was intentionally split out from the earlier security PR so it can be reviewed on its own.

## Verification
- npx tsc --noEmit
- npm test

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "productName": "wmux",
   "version": "2.3.1",
   "description": "Windows terminal multiplexer with MCP server for AI agents",
-  "license": "MIT",
   "private": true,
   "main": ".vite/build/index.js",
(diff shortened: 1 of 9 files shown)


openwong2kim:
Great work! Thanks for the contribution. Merging this in :tada:

PR #30, opened 2026-05-12
fix(notifications): restore agent task-complete notification pipeline

openwong2kim:
## Summary

Restores the broken agent task-complete notification pipeline. User-visible symptom: when an AI agent in a wmux pane finishes its turn, none of the three signals — sidebar dot, unread badge, OS toast — fire. The detection layer in main was disconnected from the renderer UI in four places, and daemon mode (the production default) bypassed the entire signal chain.

Three commits, each a logical unit:

- `cecdfc4` — **Pipeline restoration** (26 files): wires `AgentDetector.onEvent` + `ActivityMonitor.onActive` into `PTYBridge` + a new daemon-mode `DaemonNotificationRouter`. Unifies the `IPC.METADATA_UPDATE` payload contract (was silently broken between `metadata.handler` and `meta.rpc`). Adds `sendNotification` utility, AgentDetector overhaul (enum unified, unsubscribe support, per-(agent:status) dedup with `resetEmissionState()` for turn N+1, `\r`-split for TUI redraw, extended ANSI strip, removed `esc to interrupt` false-positive). 37 new tests.

- `5aee27f` — **Daemon cycle reset + destroy event** (5 files): post-review fixes after the first Codex pass. Daemon-side `AgentDetector` dedup now resets on each active burst (was stuck after turn 1). `session.destroyed` event is broadcast and handled, so closing a daemon-backed terminal via `pty:dispose` clears `agentStatus` instead of leaving a stale sidebar dot.

- `cddd3bd` — **Cross-model review fixes** (4 files, 5 catches): independent Codex re-review + Claude code-reviewer subagent each surfaced additional cases. Both agreed on the local-dispose path. Resolved:
  - C1: `onActiveToIdle` fallback now clears stale 'running' status when no precise agent event followed
  - C2: `cleanupInstance` broadcasts `agentStatus='idle'` (covers `PTYManager.dispose` path that bypasses `onExit`)
  - C3: in-app `pushToast` now respects `state.toastEnabled` user preference
  - L3: `DaemonNotificationRouter.onActive` omits `agentName` to avoid clobbering the prior name (daemon has no access to `AgentDetector.getLastAgent()`)
  - L4: `AppLayout` session-restore sanitize skips already-idle workspaces (was firing redundant no-op updates)

## Review trail

| Pass | Reviewer | Findings | Status |
|---|---|---|---|
| Plan 1 | `/plan-ceo-review` | 5 proposals, 2 accepted, 3 deferred | clean (SELECTIVE_EXPANSION) |
| Plan 1 | Codex round 1 | 10 | all addressed |
| Plan 1 | `/plan-eng-review` | 11, 1 critical (daemon mode) | all addressed |
| Plan 1 | Codex round 2 | 8 | all addressed |
| Code 2 | Codex round 1 | 2 (1 P1, 1 P2) | all addressed in `5aee27f` |
| Code 3 | Codex round 2 | 3 (2 P2, 1 P3) | all addressed in `cddd3bd` |
| Code 3 | Claude subagent | 7 (2 P2, 5 P3) | 2 addressed in `cddd3bd`, 5 deferred |

Cross-model agreement caught the dispose path from two angles (Codex C2 = agentStatus stale, Claude L5 = elapsed=0 cosmetic, same `cleanupInstance` method).

## Deferred (Claude P2/P3, follow-up issues)

- `DaemonNotificationRouter` regression suite — manual verification for now; mocking the full daemon IPty pipeline is out of scope.
- Session-restore sanitize regression test.
- `onExit` elapsed=0 cosmetic when `cleanupInstance` ran first.
- `AgentDetector` dedup intent comment.
- `DaemonClient` `removeAllListeners` on disconnect (pre-existing, not introduced here).
- `DaemonNotificationRouter` startup race window (events between `client.connect` resolve and `router.start` — recovery sessions are muted that early in practice).

`TODOS.md` also picks up the cherry-picked deferrals from CEO review: E3 (transient dot flash animation), E4 (per-workspace mute), E5 (tray icon unread badge, cross-platform), and the Eureka Phase 2 (Claude Code stop-hook → OSC 9 BEL).

## Test plan

- [x] `tsc --noEmit` — 0 errors
- [x] `npx vitest run` — 980 / 980 pass (was 943; 37 new)
- [x] Regression coverage: R1-R6 unit-tested (R7 pushToast requires jsdom — manual)
- [ ] Manual: Claude Code in local mode — verify sidebar dot transitions `running` → `waiting` on prompt, unread badge increments, in-app toast fires, OS toast fires when window is unfocused
- [ ] Manual: same with daemon mode (production default) — confirms `DaemonNotificationRouter` wiring
- [ ] Manual: close a pane while agent is `running` — sidebar dot clears immediately (covers `cleanupInstance` `idle` broadcast)
- [ ] Manual: disable "Toast notifications" in settings — confirm both OS toast AND in-app overlay suppress
- [ ] Manual: long `npm run build` in a non-agent shell — confirm ActivityMonitor fallback fires once, then sidebar returns to `idle`
- [ ] Manual: restart wmux with a saved session — confirm no stale `running`/`waiting` dots from the previous run

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -39,3 +39,39 @@
 - **Cons:** collectLeafSurfaces로 workspace별 PTY 수집 로직 필요
 - **Context:** `src/company/renderer/provisioner.ts` 및 기존 CompanyPanel.tsx의 destroy 로직
 - **Depends on:** 없음
+
+## (E3) Agent status dot transient flash on completed/waiting
+- **What:** `agentStatusIcon.ts`/`MiniSidebar` dot animation을 `completed`/`waiting` 전환 시점에 한 번만 발화 (예: 2s flash). 현재 `running`만 `animate-pulse`이고 나머지는 정적.
+- **Why:** wmux 창에 포커스가 있어도 측면 시야에서 변화를 잡을 수 있어야 함. inline toast가 transient라 놓치기 쉬움.
+- **Pros:** 다중 agent 워크플로우에서 시각 신호 강화. CSS만으로 가능.
+- **Cons:** 여러 workspace가 동시에 completed로 변하면 시각 노이즈. tuning 필요.
+- **Context:** `src/renderer/components/Sidebar/agentStatusIcon.ts` className 매핑 + `MiniSidebar.tsx:133-140` / `WorkspaceItem.tsx:177-178`. `useEffect`로 status 변경 detect + setTimeout으로 transient class. CC 추정 15분.
+- **Depends on:** validated-pondering-grove plan (알림 파이프라인 복구) ship 후 사용자 피드백
+- **Priority:** P3 (cosmetic)
+
+## (E4) Per-workspace notification mute/snooze
+- **What:** `WorkspaceMetadata`에 `notificationsMuted: boolean` 필드 추가 + 사이드바 우클릭 메뉴 또는 settings에서 toggle. `useNotificationListener`가 muted workspace의 알림은 store에 안 push.
+- **Why:** v2.8.2에서 MAX_SESSIONS 50→200 확장됨. 다중 agent 동시 실행 시 알림 폭주 방지 필수.
+- **Pros:** 사용자 통제권. 시끄러운 빌드 workspace를 silent로.
+- **Cons:** UI 토글 위치/아이콘/i18n key 결정 필요. settings vs context menu 결정.
+- **Context:** `src/shared/types.ts:139` WorkspaceMetadata 확장, `src/renderer/hooks/useNotificationListener.ts:41-66`에서 muted 체크 추가, `notificationSlice.markAllReadForWorkspace`와 같은 패턴으로 mute action 추가. CC 추정 30-40분.
+- **Depends on:** 알림 파이프라인 복구 ship 후 사용자 피드백 (정말 시끄러운지 확인 후 우선순위)
+- **Priority:** P2
+
+## (E5) Tray icon unread badge (cross-platform)
+- **What:** `src/main/tray.ts`에 unread count를 표시. macOS: `app.dock.setBadge(N)`, Windows: tray tooltip prefix `[N] wmux`, Linux: best-effort (NotificationServer 표준은 제한적).
+- **Why:** wmux 창이 minimize/hide 상태에서는 OS 토스트 한 번 외에는 알림 신호 부재. tray가 영구 신호 채널.
+- **Pros:** 창 안 열어도 unread 인지. macOS 사용자에게 자연스러운 UX.
+- **Cons:** Cross-platform 차이 큼 — 3 OS에서 각각 QA 세션 필요. Linux fallback 정책 결정 필요.
+- **Context:** memory `project_cross_platform_branch_policy` 정책에 따라 **feature branch + PR**로 진행. `src/main/notification/ToastManager.ts`와 같은 자리에 `TrayBadgeManager` 추가 고려. Renderer store unread count 변경 → main으로 push → tray 갱신. CC 추정 30-45분 + 3 OS QA.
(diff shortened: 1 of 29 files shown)


openwong2kim:
## Follow-up commit: `42f5bd3` — resize/typing false positives + auto-read

After local manual testing, the user found two false-positive notification patterns the 6-round plan review pipeline missed:

1. **Sidebar dot/badge climbs while typing.** User keystrokes echo back through the PTY → ActivityMonitor crosses the 2000B threshold → pause to think → 5s idle timer fires → "Task may have finished" appears while the user is still composing.

2. **Workspace switch without clicking a terminal adds an unread.** Switching workspaces toggles `display:flex/none`. xterm's `FitAddon.fit()` sends `IPC.PTY_RESIZE`. TUI agents (Claude, Codex) reply to SIGWINCH with a full-screen redraw → ActivityMonitor enters 'active' → user leaves within 5s → idle timer later fires a notification for a workspace they only visited.

Both are the same root cause: ActivityMonitor's byte-count heuristic cannot tell "agent task ending" from "PTY redraw caused by external state change (resize / user keystroke echo)".

### Fix layers in this commit

**F1 — Suppress the activity fallback right after a resize or user keystroke** (`src/main/notification/idleSuppression.ts`)
- New module tracks `lastResizeAt` + `lastUserWriteAt` per `ptyId` with a 30s window (bigger than ActivityMonitor's 5s idle delay so any timer started during the window also observes the suppression when it fires).
- `pty.handler.ts` calls `markResize(id)` in both resize paths (local + daemon) and `markUserWrite(id)` in both write paths.
- `PTYBridge.onActiveToIdle` and `DaemonNotificationRouter.onIdle` check `recentlySuppressed(ptyId)` before emitting. **AgentDetector's precise events are NOT gated** — only the heuristic fallback is suppressed.
- `DaemonNotificationRouter.onIdle` also now broadcasts `agentStatus='idle'` before sending the notification (parity with PTYBridge; previously it left a stale 'running' dot when the fallback fired).
- Suppression state cleared in `cleanupInstance` / `onSessionEnd` to prevent leaks across PTY recycling.

**F3 — In-app surfaces skip notifications the user is already looking at**
- `useNotificationListener.ts`: when a notification arrives whose `ptyId` matches the active surface of the active workspace, `addNotification` + `pushToast` are skipped entirely. The user is already seeing the output. OS-level toast in `ToastManager` still has its own focus gate.
- `workspaceSlice.setActiveWorkspace` marks that workspace's notifications as read on activation. Previously, sidebar tile click did not read anything (only per-terminal click did) so badges climbed monotonically across workspace switches. Guarded with `Array.isArray` so the existing workspaceSlice unit tests (which don't mount the notification slice) still pass.

### Updated test plan

- [ ] Manual: type continuously for 10+ seconds in an agent prompt → pause 6s → confirm NO "Task may have finished" appears
- [ ] Manual: with a Claude-CLI agent running in workspace A, switch A → B → C → A without clicking any terminal → confirm A's unread badge stays at 0
- [ ] Manual: in workspace A's active surface, agent finishes a turn → confirm NO badge increment (user is looking at it); switch to workspace B → trigger same turn → confirm B's badge increments by 1 and OS toast fires (window unfocused) or not (window focused)
- [ ] Manual: long `npm run build` in a non-agent shell while focused on that terminal → confirm fallback notification still suppressed (active surface skip)
- [ ] Manual: long `npm run build` while focused on a different workspace → confirm one fallback notification fires after 5s idle (resize/typing windows expired)

### Why F2 wasn't taken
"Mute hidden workspaces entirely" was on the table but rejected: agents legitimately reach 'waiting' in hidden workspaces and the user should still get the badge / OS toast — F1 kills the false positives without losing the real signal.

TS 0 errors. 980/980 tests pass.

PR #33, opened 2026-05-12
feat(metadata): M0-a — MetadataStore module + 28 unit tests

openwong2kim:
> **Stacked on #32.** Review #32 first — this PR adds code on top of the Phase 0 docs (PROTOCOL.md, M0 design doc) that frame what's being implemented here.

## Summary

Phase 1 **M0-a** — the new main-process `MetadataStore` module. Isolated: no call sites changed yet. Handler rewiring (M0-b), paneSlice mirror-only conversion (M0-d), and SessionManager persistence integration (M0-e) land in follow-up PRs stacked on this one.

Builds on RFC #15 (with @alphabeen) — `expectedVersion` and `mergeMode` wire format are first surfaced here, behind the store, before being exposed on the JSON-RPC surface in M0-b.

cc @alphabeen — per your note in #15, tagging when the first M-series PR opens. This PR is the substrate side of the optimistic-concurrency + `mergeMode` design we discussed; M0-b will wire it to the wire format.

Tracks #31.

## Changes

- **`src/main/metadata/MetadataStore.ts`** (~380 LOC) implementing the full contract from `docs/internal/m0-design.md` §2:
  - `get` / `set` / `clear` / `snapshot` / `hydrate` / `serialize` / `migrate` / `onPaneDeleted`
  - per-pane monotonic `version`
  - optimistic concurrency via `expectedVersion`
  - three `mergeMode` semantics: `merge` (default) / `replace` / `replaceShared`
  - validation guards matching the existing `pane.rpc.ts:30` sanitizer
  - EventBus injection point for testability (module-level singleton in production)
- **`src/main/metadata/__tests__/MetadataStore.test.ts`** (~380 LOC, 28 tests):
  - CRUD + version (5)
  - `mergeMode` (4)
  - `expectedVersion` (5)
  - validation guards (4)
  - snapshot envelope (2)
  - persistence roundtrip + schema migration (4)
  - EventBus emission shape (4)
- **`src/shared/events.ts`** — `PaneMetadataChangedEvent.version?` (optional, additive). MetadataStore emissions carry it; legacy `publisher.ts` paths continue to emit without.

### Added

- Internal `MetadataStore` module (`src/main/metadata/`) — the main-process authority for `PaneMetadata` (M0-a). Not yet wired to RPC handlers; M0-b lands that.
- Optional `version` field on `pane.metadata.changed` events (emitted only when produced by `MetadataStore`).

### Changed

n/a (no public surface change yet)

### Deprecated

n/a

### Removed

n/a

[...]

PR #32, opened 2026-05-12
docs(substrate): phase 0 — inventory, contract, PR template

openwong2kim:
## Summary

Substrate 3.0 Phase 0 — the external-contract paperwork that locks in what already shipped (v2.8.0 PaneMetadata + EventBus + bootId + `mcp.claimWorkspace`) and stakes out the v3.0 stable surface ahead of the M0 metadata-store work.

No code surface changes. Pure docs + a PR template + the README positioning shift to LSP-for-terminals.

Tracks #31. Builds on the RFC discussion in #15.

## Changes

- **README** opens with the LSP-for-terminals framing (substrate identity) while preserving the AI-agent value prop and tmux-alternative keywords.
- **`docs/api/inventory.md`** catalogs every RPC method, MCP tool, and event type with a stability tier (`stable` / `experimental` / `internal`).
- **`docs/api/versioning.md`** formalizes semver + tier semantics. Documents v2.x → v3.0 auto-updater compatibility (winget · Chocolatey · electron-updater all upgrade cleanly) and explains the v3.0 + "Substrate 1.0" tagline decision over a v1.0 reset.
- **`docs/api/stability.md`** freezes the v3.0 stable-surface contract: `pane.list` snapshot envelope, `pane.{set,get,clear}Metadata` with `expectedVersion` + `mergeMode`, `events.poll` cursor + `bootId` + `resync` semantics, `mcp.claimWorkspace`, validation limits, error codes.
- **`docs/PROTOCOL.md`** (draft 1) is the substrate wire contract: layered status, namespacing, optimistic concurrency, mergeMode semantics, cursor opaqueness, snapshot reconciliation, 4-point permission enforcement sketch, Named Pipe token security model.
- **`docs/internal/paneSlice-callsite-inventory.md`** is the M0 pre-work required by the 2nd Codex review. Walks every read/write touching `PaneLeaf.metadata` and confirms the mirror migration has a narrow blast radius — no React component currently subscribes to the display-vocabulary fields.
- **`.github/PULL_REQUEST_TEMPLATE.md`** adds CHANGELOG sections + stability-tier classification so every PR self-classifies its contract impact.

### Added

- Substrate 3.0 contract documentation: `docs/PROTOCOL.md`, `docs/api/{inventory,versioning,stability}.md`.
- PR template with CHANGELOG + stability-tier sections.

### Changed

- README opens with the LSP-for-terminals substrate framing.

### Deprecated

n/a

### Removed

n/a

### Fixed

n/a

### Security

n/a

[...]

Issue #31, opened 2026-05-12
Tracking: Substrate 3.0 — v3.0 milestone roadmap


openwong2kim:
**Vision:** wmux as LSP-for-terminals — a neutral substrate that lets external tools build workflow intelligence on top of any terminal session.

This tracking issue is the v3.0 milestone roadmap. It builds on the RFC discussion in #15 (with @alphabeen) and the external-tooling primitives shipped via #18.

Living document — checkboxes move as PRs land. See the in-repo plan for design rationale: `docs/PROTOCOL.md`, `docs/api/{inventory,versioning,stability}.md`.

> **2026-05-14 status update**: Phase 0 + Phase 1 (M0) shipped in **v2.9.0 GA** via #34 (Substrate 3.0 — Phase 0 + M0 ship unit) and #35 (chopped-dump recovery tool). See comments below for the closeout summary.

> **2026-05-20 status update**: Phase 2.1 MCP plugin convention landed record-only via #46 (path-D call-site inventory pre-work) → #48 (plugin identity + `wmuxPermissions` grammar) → #49 (trust-DB invariants follow-up — widening demote, LRU cap, structured rejection, transport-close clear). Phase 3.2 baseline attempted via #41 (SECURITY.md + filesystem hardening) but the Windows ACL hardening helper caused a user-dogfood lockout regression; #43 reverted the helper and restored a hardening-claims-removed `docs/SECURITY.md` (substrate-not-vault stance, OS delegation table, OS-isolation guidance kept; NTFS ACL + cloud-sync signals dropped). v2.9.x patch stream: #39 (v2.9.1 — Daemon shutdown reliability + .txt active-hazard mitigation), #45 (#35 instrumentation + recovery PTY retry + before-quit timeout), #47 (three-layer Fix 0 — mount gate + race-free reconnect + restore toggle). Adjacent ships outside the substrate roadmap: #50 (i18n 23 locales — cmux parity), #51 (v2.10.0 tmux-style prefix), #52 (`mcp/terminal_send` `submit` flag). See comments below for the PR-by-PR rundown.

---

## Phase 0 — Inventory & Decisions (Week 1–2) ✅

The substrate surface is **already ~50% shipped in v2.8.0** (PaneMetadata, EventBus, bootId, asOfSeq, system.capabilities, MCP host, mcp.claimWorkspace). Phase 0 is **organizational, not new code** — declaring the contract that already exists and the boundaries the substrate is going to formalize.

[...]

Issue #18, opened 2026-05-08
Tracking: External tooling primitives (pane metadata + events poll)


openwong2kim:
Tracking issue for #15 (@alphabeen's RFC).

## Scope decisions

The three primitives proposed in #15 split as follows:

| Primitive | Status | Where |
|-----------|--------|-------|
| Pane metadata | Implementation in review | #16 |
| JSON-RPC events | Implementation in review (stacked) | #17 |
| Optional buffer access | Deferred | already partial via `terminal_read` / `terminal_read_events` |

Stay aligned with #15's framing: wmux remains a neutral terminal/workspace layer. New surfaces are pull-only, additive to RPC, and gated by workspace identity so external tools stay scoped.

### Phase 1 — Pane metadata (#16, ready for review)

- [x] `PaneMetadata { label, role, status, custom: Record<string,string>, updatedAt }` on `PaneLeaf`
- [x] Caps: 8KB serialized total, label ≤64, role ≤64, status ≤128, custom ≤32 entries × 64-char keys
- [x] RPC: `pane.setMetadata` / `pane.getMetadata` / `pane.clearMetadata`
- [x] MCP tools: `pane_set_metadata`, `pane_get_metadata` (auto-pin to caller's `workspaceId`)
- [x] `system.capabilities().features.paneMetadata = true` for client detection
- [x] Persistence: piggybacks on existing `SessionManager` JSON dump (backwards-compat)
- [x] Cross-workspace hijack scoping (same class as v2.7.2 `mcp.claimWorkspace` fix)
- [x] Custom map deep-merges on `merge=true` so cooperating tools don't clobber each other
- [x] Tests: 35 cases covering merge / scope / validation paths
- [ ] Manual smoke test from external Claude Code
- [ ] Merge

### Phase 2 — JSON-RPC events (#17, stacked on #16)

- [x] `WmuxEventType` union: `pane.created` / `pane.closed` / `pane.focused` / `pane.metadata.changed` / `process.started` / `process.exited`
- [x] `EventBus` singleton: 1024-event ring, monotonic `seq` cursor, stale-cursor `resync: true` flag, per-workspace + per-type filtering
- [x] RPC: `events.poll({ cursor, types?, workspaceId?, max? })`
- [x] MCP tool: `wmux_events_poll` (auto-scoped to caller's workspace)
- [x] Producers: `paneSlice` (5 methods) + `PTYBridge` (`onAttach` / `onExit`)
[...]

PR #34, opened 2026-05-12
wmux Substrate 3.0 — Phase 0 + M0 (v2.9.0 ship unit)

openwong2kim:
## Summary

**wmux Substrate 3.0 — Phase 0 + M0 (v2.9.0 ship unit).** Builds on the RFC discussion in #15 with @alphabeen. Closes the gap between the substrate identity we've been talking about and the substrate primitives we actually ship.

This is one PR by design — `feature/m0-b-handler-rewrite` carries the whole v2.9.0 batch so reviewers see the substrate as a coherent unit instead of having to chain through stacked PRs. Commit history is preserved for review-by-commit.

cc @alphabeen — the substrate side of the optimistic-concurrency + `mergeMode` design from #15 is now landed end to end, code + contract docs. Open to feedback on the parts that are still draft.

Tracks #31.

### Phase 0 — contract paperwork (no new code surface)

The substrate is **already ~50% shipped in v2.8.0** (PaneMetadata, EventBus, bootId, asOfSeq, system.capabilities, MCP host, `mcp.claimWorkspace`). Phase 0 declares the contract that already exists and the boundaries the substrate is going to formalize in v3.0.

- **README** opens with the LSP-for-terminals framing while preserving the AI-agent value prop and tmux-alternative keywords.
- **`docs/api/inventory.md`** catalogs every RPC method, MCP tool, and event type with a stability tier.
- **`docs/api/versioning.md`** formalizes semver + tier semantics, documents v2.x → v3.0 auto-updater compatibility, explains the v3.0 + "Substrate 1.0" tagline decision.
- **`docs/api/stability.md`** freezes the v3.0 stable-surface contract.
- **`docs/PROTOCOL.md`** (draft 1) — the substrate wire contract: layered status, namespacing, optimistic concurrency, mergeMode, cursor opaqueness, snapshot reconciliation, permission enforcement sketch, Named Pipe token security model.
- **`docs/internal/paneSlice-callsite-inventory.md`** — M0 pre-work; walks every read/write touching `PaneLeaf.metadata` and confirms the mirror migration has a narrow blast radius.
- **`docs/internal/m0-design.md`** — pre-implementation design doc for the M0 series with 6 race specs and a test budget.
- **`.github/PULL_REQUEST_TEMPLATE.md`** adds CHANGELOG + stability-tier sections.

### M0-a — `MetadataStore` module

The main-process authority for `PaneMetadata` lands as an isolated module.

- `src/main/metadata/MetadataStore.ts` — `get` / `set` / `clear` / `snapshot` / `hydrate` / `serialize` / `migrate` / `onPaneDeleted`, with per-pane monotonic `version`, optimistic concurrency via `expectedVersion`, and three `mergeMode` semantics (`merge` / `replace` / `replaceShared`).
- `src/shared/events.ts` — optional `version` field on `PaneMetadataChangedEvent` (backwards-compatible; legacy publisher.ts paths keep emitting without).
- 31 unit tests cover CRUD + version, mergeMode transactions, optimistic concurrency, validation, snapshot, persistence, EventBus emission, and three codex-found regression cases.

### M0-b — handler authority migration

`pane.{set,get,clear}Metadata` handlers now route through `MetadataStore` for all calls. paneSlice is no longer mutated by the RPC metadata path — M0-d's mirror-only conversion is now nearly a no-op.

- `pane.resolveActiveLeaf` — a new internal IPC channel: when the caller omits `paneId`, main asks renderer for the active leaf id (read-only, no paneSlice write) and then commits via MetadataStore. Closes the codex P1 split-store hole.
- Wire shape preserved from v2.x (reply has no `version` field yet — M0-f exposes that).
- Legacy `merge: boolean` parameter preserved; internally maps to `mergeMode: 'merge' | 'replace'`.

### M0-c through M0-f — folded into this PR as commits

The remaining M0 stages landed onto this branch instead of stacked PRs so v2.9.0 ships as one unit:

- **M0-c** — `MetadataStore.snapshot()` integrated into the `pane.list` envelope, anchored to the pane tree so `asOfSeq` reflects the snapshot lineage.
- **M0-d** — paneSlice mirror-only conversion. Compile-time write protection added; near no-op after M0-b.
- **M0-e** — `SessionManager.saveMetadataSync` wired to `MetadataStore` persist callback. Hydrate path also added; strict field validation per codex P2.
- **M0-f** — wire-format additions (`expectedVersion`, `mergeMode`, `version` on reply / event / list); spec doc lands as `docs/PROTOCOL.md` updates.

### M0 dogfood stability fixes (surfaced during v2.9.0-rc.1 dogfood)

Three follow-ups discovered while daily-driving the substrate build:

- **logSink durable write** (`f6863bf`, `0e316fd`): EPIPE infinite loop on broken stdout; reentrancy guard + `appendFileSync` so log lines reach disk even under stress.
- **IPC handler swap race** (`3e1a375`): session + scrollback handlers moved out of the daemon-connect handler-swap cycle so `scrollback:load` cannot land in the unregister window during cold boot.
- **PTY recovery mute race** (`377d11f`, `3206c6f`): `pty:resize` against a session the daemon hasn't published yet returns `session not found` and used to break recovery PTYs silently. Retry budget = 50 attempts × 20ms with diagnostic logging.
- **Reconcile race** (this PR's `wip(m0)` commit): `daemon.whenReady()` and `daemon.onConnected` both triggered `reconcilePtys` on first connect, racing two reconcile walks against the same session. Symptom: some panes ended up input-mute after launch until a workspace toggle re-ran reconcile. Fix: in-flight guard drops the duplicate trigger; re-read the workspace snapshot per walk so concurrent surface spawns are not masked by a frozen view.

### Scrollback corruption fix (P0 layered defense)

Production v2.8.4 was silently corrupting persisted scrollback dumps. Forensics on real on-disk samples (42 KB and 7.8 KB chopped files) showed the smoking-gun signature: 99% of non-empty lines were 1-2 characters, separated by CRLF, e.g. `om\r\nma\r\nn\r\ndN\r\not\r\nFo\r\nun\r\nd\r\n...` for what was originally `CommandNotFoundException`.

**Root cause.** xterm's `FitAddon` collapses `cols` to ~2 when run against a hidden or zero-width container. The renderer's 5s autosave (and `beforeunload` final dump) iterates physical xterm rows via `getLine(i).translateToString(true)` and joins with CRLF — capturing the cols-collapsed reflow as a column of single characters. On the next launch the renderer restored that garbage into a fresh xterm, then the next 5s autosave re-dumped it. Self-sustaining corruption loop. The visible symptom is scrollback that "looks empty after reboot."

**Fix is layered:**

| Layer | Component | What it does |
|---|---|---|
| Dump-time | `src/renderer/utils/scrollbackDump.ts` (new) | Refuse to serialize when `cols < 12`, `rows <= 0`, or `terminal.element.offsetWidth === 0` / detached. Extracted out of `AppLayout.tsx` so the guard is unit-testable against fixture terminals. |
| Dump-time | `src/renderer/hooks/useTerminal.ts` | Last unguarded `fit()` site (font/theme-change effect) now checks `container.offsetWidth/offsetHeight`. Matches the guards already in place at the other four `fit()` sites. |
| IPC | `src/main/scrollback/corruption.ts` (new) | Cols-collapse detector — `median ≤ 3 char non-empty line length` AND `CRLF byte ratio ≥ 0.3`. Max-line gate was deliberately omitted: real-world chopped fixtures had a small minority of pre-collapse lines up to 60 chars long, so the median is the load-bearing signal. |
| IPC | `src/main/ipc/handlers/session.handler.ts` | `SCROLLBACK_DUMP` rejects writes whose payload matches the corruption signature (defense in depth against renderer guard regressions). `SCROLLBACK_LOAD` validates the read, quarantines a rejected primary into `corrupted/{ts}.bak` (with structured `CORRUPT_FILE` log), and walks `.bak` → `.bak.3` for a healthy fallback. |
| Durability | `src/daemon/util/atomicWrite/text.ts` (new) | Text variant of the existing JSON `atomicWrite` primitives — tmp+rename, 4-generation `.bak` rotation, shared quarantine pipeline. The JSON variant in `core.ts` assumes parseable payloads; this sibling keeps scrollback's raw-bytes contract clean. |

**Verification.** 40 new unit tests cover the text atomic write/read/rotation/quarantine, the detector against synthetic and real production fixtures, and the dump-time eligibility guards. End-to-end verification on a packaged Squirrel build (full uninstall → reinstall → launch) confirmed that pre-existing chopped files get quarantined into `userData/wmux/scrollback/corrupted/` on first launch and that subsequent dump cycles produce clean (sub-1 KB) content with the rotation chain populating four healthy generations. The packaged-build test surfaced an early threshold bug (the initial detector gated on `max ≤ 10` and missed the production samples whose max was 60); the threshold fix is folded into the detector commit.

**v2.8.4 migration cost.** A v2.8.4 user upgrading to v2.9.0 will see their pre-collapse dumps moved into `corrupted/`. Visible scrollback for those panes appears blank/fresh on first launch — but the chopped raw bytes are preserved (cleanup defaults: 30 days / 10 files per dir). A reverse-reflow recovery script can extract usable text from those quarantined files; that lands as a separate follow-up PR.

### Codex review fixes

External voice on this PR caught 7 real findings across the substrate (full-stack pass + M0-b pass) and an additional architecture review on the scrollback fix:

| Codex pass | Finding | Resolution |
|---|---|---|
| Full stack | `replaceShared` overwrote `custom` | `patch.custom` silently ignored in this mode; namespace clobber prevented |
| Full stack | Size cap checked before `updatedAt` appended | check moved to post-merge shape including `updatedAt` |
| Full stack | `custom` entry cap only validated on patch | post-merge entry count now validated |
| Full stack | `stability.md` cap numbers diverged from `types.ts` | doc updated to match (ROLE 32→64, STATUS 64→128, CUSTOM_ENTRIES 64→32) |
| Full stack | `versioning.md` showed `metadata.version` (nested) | corrected to sibling field per PROTOCOL.md §3 |
| M0-b | Split-store read-after-write hole (P1) | unified all 3 handlers via `pane.resolveActiveLeaf` IPC |
| M0-b | `workspaceId ?? ''` clobbered remembered scope (P2) | dropped coercion; MetadataStore's existing fallback now works |
| Scrollback | Diagnosis-correctness review | confirmed cols-collapse root cause; recommended layered defense + max-gate-free detector; recommended daemon-side RingBuffer as long-term durable source (deferred to v3.x) |

### Tracking issue

#31 — Substrate 3.0 v3.0 milestone roadmap (Phase 0 → 4 + parallel v3.1 macOS + v2.9.x patch stream).

### Added

- Substrate 3.0 contract documentation: `docs/PROTOCOL.md`, `docs/api/{inventory,versioning,stability}.md`.
- M0 design + paneSlice callsite inventory: `docs/internal/{m0-design,paneSlice-callsite-inventory}.md`.
- `MetadataStore` module — the main-process authority for `PaneMetadata` (integrated end-to-end through M0-c/d/e/f within this PR).
- Optional `version` field on `pane.metadata.changed` events; `expectedVersion` and `mergeMode` on `pane.setMetadata` wire shape.
- PR template with CHANGELOG + stability-tier sections.
- `atomicWriteText` / `atomicReadText` (sync + async) — text variant of the existing JSON atomic-write primitives, sharing the rotation + quarantine pipeline.
- Cols-collapse corruption detector for scrollback dumps (`src/main/scrollback/corruption.ts`).
- `scrollbackDump` util module (`src/renderer/utils/scrollbackDump.ts`) — extracts the renderer serializer so its eligibility guards are unit-testable.

### Changed

- README opens with the LSP-for-terminals substrate framing.
- `pane.{set,get,clear}Metadata` RPC handlers write through `MetadataStore`; paneSlice retained as renderer-internal mirror only.
- `pane.list` envelope now anchored to `MetadataStore.snapshot()` (M0-c).
- `SessionManager` persists `metadata.json` via `MetadataStore`'s persist callback and hydrates the store on launch (M0-e).
- `SCROLLBACK_DUMP` IPC handler now uses `atomicWriteTextSync` with rotation enabled; pre-write corruption check rejects payloads matching the cols-collapse signature.
- `SCROLLBACK_LOAD` IPC handler validates the read, quarantines corrupt files into `corrupted/`, and walks the `.bak` chain for fallback.

### Deprecated

n/a

### Removed

n/a

### Fixed

- `replaceShared` mergeMode no longer overwrites `custom` when caller sends one (substrate guarantee against tool-namespace clobber).
- MetadataStore size cap (`PANE_METADATA_MAX_BYTES`) now runs on the final stored shape including `updatedAt`.
- MetadataStore `custom` entry cap now validated on the post-merge shape so cumulative writes cannot bypass it.
- Scrollback corruption loop — a `fit()` against a hidden / zero-width container collapsed `cols` to ~2, causing the next dump to persist a column of single characters and the next launch to restore it. Fixed at four layers: dump-time eligibility guards (cols / rows / visibility), font-theme-change `fit()` guard, IPC-level rejection + quarantine, and atomic dump with `.bak` rotation chain.
- Pane input mute after launch — `daemon.whenReady()` and `daemon.onConnected` racing two reconcile walks against the same daemon snapshot left some panes input-mute until the user toggled workspaces. In-flight ref drops the duplicate; per-walk store snapshot fixes concurrent-spawn masking.
- `pty:resize` against a session the daemon hadn't published yet returned `session not found` and broke recovery PTYs. Retry budget = 50 × 20ms with diagnostic logging.
- IPC `session` + `scrollback` handlers moved out of the daemon-connect handler-swap cycle so `scrollback:load` cannot land in the unregister window during cold boot.
- logSink reentrancy + EPIPE infinite loop; appendFileSync makes log lines durable.

### Security

n/a

## Stability tier impact

- [x] Additive change to a `stable` surface (new optional fields on RPC reply / event; `expectedVersion` + `mergeMode` are additive args).
- [x] Internal storage layout change (`userData/wmux/scrollback/corrupted/` subdir created on demand when corrupt files quarantine; cleanup defaults 30 days / 10 files per dir).

The wire contract for `pane.{set,get,clear}Metadata` remains backwards-compatible — `expectedVersion` and `mergeMode` are optional; v2.x clients continue to work. `pane.list` envelope gains an optional `metadata` snapshot anchored to `asOfSeq`.

The new `pane.resolveActiveLeaf` channel is **internal** (main ↔ renderer IPC), not part of the external JSON-RPC surface.

## Test plan

```
npx vitest run
  → all suites pass (delta vs baseline includes +40 scrollback tests
    across atomicText, scrollbackDump, and corruption detector).
  → two known parallelism flakes (ProcessMonitor, crashRestore) pass on
    isolated runs.

npx tsc --noEmit -p tsconfig.json
  → clean

npx eslint --ext .ts <changed files>
  → no new warnings introduced by this PR
```

Manual / dynamic verification — end-to-end packaged-build run:

1. `npm run make` → built Squirrel installer
2. `Update.exe --uninstall` + `wmux-2.8.4.Setup.exe`
3. First launch — quarantined pre-existing chopped scrollback files into `userData/wmux/scrollback/corrupted/`, structured `CORRUPT_FILE` logs emitted to stderr.
4. Fresh xterms spawned; subsequent 5s autosaves produced clean dumps (median line length back to normal values).
5. `.bak` rotation chain populated four healthy generations within 20 seconds.
6. M0 reconcile-race regression repro: pre-fix build showed only 2/7 panes accepting input on first launch; post-fix build had all 7 alive immediately.

## Known limitations carried to follow-up

- The reverse-reflow recovery tool for v2.8.4 chopped data lives in a follow-up PR (the quarantined raw bytes are preserved in `corrupted/`, so recovery is non-destructive and can run any time).
- The daemon-side RingBuffer (already populated for AI agent context) is not yet the durable source of truth for visual scrollback — codex review recommended that direction as a v3.x architectural simplification. Tracked separately.

## Related issues / PRs

- Tracks #31 (v3.0 milestone roadmap)
- Builds on #15 (RFC, with @alphabeen) and #18 (external-tooling primitives)
- Supersedes the originally stacked split (#32, #33 — closed and folded into this PR)


alphabeen:
Thanks for the cc — read through the M0 design and the PR carefully.

The substrate side of this is landing very close to where I’d hoped #15 would point.

**On the architecture:**

The renderer-to-main authority migration is the right call.

Before M0-b, `pane.setMetadata` effectively routed through the renderer / `paneSlice` path after validation, which meant the renderer was not just a UI mirror but part of the metadata authority path. That makes metadata state vulnerable to renderer event drops, hydration timing issues, or split-authority problems.

`489b433` creates `MetadataStore` as the authoritative module in main, and `c055616` wires the paneId-present RPC paths through it — cleanly separating concerns so main becomes the source of truth while the renderer can eventually become a mirror. That feels much more aligned with the substrate contract.

The phased approach also makes sense. Keeping this as one coherent substrate batch while still preserving review-by-commit history feels like the right tradeoff here.

The review-by-commit structure — Phase 0 docs (`a9471d0`, `5b18c13`, `da00806`), M0-a module (`489b433`, `4d633ee`), then M0-b routing (`c055616`, `0cd2a81`) — gives enough context to understand the direction while keeping each step independently reviewable.

The temporary inconsistency window where `pane.getMetadata` reflects `MetadataStore` but `pane.list` may lag until M0-c is acceptable to me as long as it is clearly documented — and it is.

**One thing I’d like to confirm around validation:**

My read from `489b433` is that `MetadataStore` is becoming the authoritative place for validation, caps, merge semantics, and version checks, while the RPC handler (`c055616`) mainly handles request-shape normalization and backwards-compat mapping.

Is that the intended split?

The main thing I’d want to avoid is having two independent validation paths — one in the handler layer and one in `MetadataStore` — drifting over time or silently shadowing each other.

**On `expectedVersion` and retry semantics:**

The `expectedVersion?: number` + `VERSION_CONFLICT` shape looks right to me, and it fits the optimistic-concurrency model we discussed.

One thing I’d like to see documented clearly in `docs/PROTOCOL.md` is the canonical retry pattern for clients. My assumption is:

1. call `pane.getMetadata` or `pane.list` to read the current version,
2. compute / re-apply the intended patch locally,
3. retry `pane.setMetadata` with the new `expectedVersion`.

If there is a smarter reconciliation path planned, it would be useful to make that explicit before external tools start depending on the contract.

**On `mergeMode`:**

The `replaceShared` fix in `4d633ee` that silently ignores `patch.custom` is exactly right — it ensures one tool can update shared display fields without accidentally clobbering another tool’s custom namespace.

That’s the detail that makes the cooperative-tools model actually safe in practice.

**On sequencing:**

M0-a/M0-b look like the right foundation, and the planned M0-c/d/e/f follow-ups are the pieces that make this usable as a full external tooling contract:

* `pane.list` backed by `MetadataStore.snapshot()`
* version fields exposed consistently in list / event / reply surfaces
* `system.capabilities` reflecting the new metadata contract
* persistence / hydration ordering documented and tested

The protocol direction feels much clearer after M0-a/M0-b.

Happy to review M0-c/d/e/f as they land. Honestly, it’s been surprisingly easy to discuss fairly abstract substrate ideas and then immediately see them reflected in coherent implementation details a few commits later. Really appreciate the thoughtfulness and iteration speed here.

PR #72, opened 2026-05-27
feat(daemon): lifecycle hardening — idle shutdown + boot-recovery dialogs + error visibility

openwong2kim:
## Summary

Closes four gaps in the wmux daemon lifecycle so the "1 wmux ≙ 1 daemon" invariant becomes self-healing instead of relying on the next clean shutdown.

- **Idle self-shutdown** — daemon terminates itself after 5 min with 0 RPC clients and 0 PTY sessions. Configurable via `daemon.idleShutdownMinutes`. Closes the "orphan daemon survives in RAM after `taskkill` of wmux main" failure mode.
- **Launcher ping retry** — `ensureDaemon` re-pings once after a 250ms delay before declaring the existing daemon unresponsive. Absorbs cold-boot races (Defender realtime scan / ConPTY init / large recovery loop).
- **Unverified-live PID recovery dialog** — when AV blocks `tasklist.exe` / `Get-CimInstance`, the launcher prompts the user with an Electron dialog offering "Clean up and start fresh" instead of refusing to boot. Cancel re-throws the legacy error annotated with the exact `taskkill /F /PID …` recovery command.
- **Respawn-exhausted error visibility** — `DaemonRespawnController` captures the latest error from the bootstrap / respawn loop and ships it on the `respawn-exhausted` event. `main` surfaces it via `dialog.showErrorBox` with concrete recovery steps, plus the existing renderer IPC channel.

> Also includes commit `9caac08` from @iamwongeeeee — `fix(mcp): anchor workspace identity to ptyId, resolve owner live` — landed on this same branch in parallel and is included in the PR scope.

## Review-driven simplifications

`/plan-eng-review` pass during planning surfaced three structural decisions, all accepted by the author:

- **Single idle state machine.** `idleMs = now − (lastDisconnectAt ?? startTime)` collapses the "never connected" and "connected then idle" branches. No separate first-connection deadline timer.
- **Commit 6 dropped.** `taskkill /F /PID` walks the same Win32 `TerminateProcess` path with the same user token as `process.kill('SIGKILL')` — an EPERM there means EPERM here too. Instead, the SIGKILL-failure throw and the unverified-live dialog both embed the exact elevated-PowerShell command the user needs to run manually. More honest than retrying.
- **`showErrorBox` retained over toast.** Renderer has no toast infrastructure for daemon signals. Building one would expand scope; native OS dialog already meets the "make sure the user sees this" bar and `WMUX_NO_DIALOG=1` keeps automated runs quiet.

## Out of scope

- Dev / prod `~/.wmux` directory separation — operational convention covers this.
- `prctl(PR_SET_PDEATHSIG)` / `kqueue NOTE_EXIT` — native module overhead larger than the residual benefit once idle-shutdown lands.
- `WMUX_DIR` env var — paired with dev / prod decision above; both deferred.

## Wire format

`RespawnEvent` extends additively:

```ts
| { type: 'respawn-exhausted'; lastError?: string }
```

The renderer IPC payload now carries `{ lastError }` on the existing `daemon:respawn-exhausted` channel. No consumer renders it yet (kept out of scope), but the contract is in place.

## Test plan

- [x] `DaemonPipeServer.test.ts` — `getConnectionCount` + `getLastDisconnectAt` 0-edge stamping verified across concurrent / serial socket lifecycle
- [x] `Watchdog.test.ts` — 8 new idle-evaluation cases (disabled / grace / active connections / active sessions / never-connected fallback / fire-once / no-callback / threshold)
- [x] `idleShutdown.test.ts` — source-level invariants pinning env-var names, `onIdleCheck` signal triple, `doShutdown('idle.timeout')` route, default-5 contract, hoist order
- [x] `DaemonRespawnController.test.ts` — new `lastError` propagation test walks two cycles, asserts the SECOND cycle's error message surfaces (proves the `clear-on-install` contract)
- [x] `scripts/daemon-idle-shutdown-dynamic.mjs` — end-to-end against bundled daemon with `WMUX_IDLE_SHUTDOWN_MS=3000 WMUX_IDLE_GRACE_MS=1500 WMUX_WATCHDOG_TICK_MS=500`. Verifies the daemon exits code 0 with the `[shutdown.phase] idle.timeout` breadcrumb in ~5s.
- [x] Full suite: **1959 passed / 3 flaky** in `StateWriter.test.ts` (debounce + flushSync timing) — all 3 pass cleanly in isolation, unrelated to this PR.

## Dogfood checklist

After the PR is reviewed but before landing, the author should walk through:

1. **Idle shutdown fires** — open wmux, close every pane, hide window to tray, wait 5 min, confirm `~/.wmux/logs/daemon-YYYY-MM-DD.log` shows `[shutdown.phase] idle.timeout idleMs=…` and the daemon process is gone.
2. **Idle does NOT fire while a pane is alive** — leave 1 pane open, hide to tray, wait 5 min, daemon still alive.
3. **Unverified-live dialog** — handcraft a fake `daemon.pid` pointing to an unrelated PID (e.g. Notepad) where `tasklist.exe` is blocked, launch wmux, confirm the dialog appears and "Clean up and start fresh" recovers.
4. **Respawn-exhausted dialog** — break the daemon bundle (rename `dist/daemon-bundle/index.js`), launch wmux, walk the 5-retry budget, confirm the `dialog.showErrorBox` appears with the actual `lastError` text and recovery steps.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


openwong2kim:
**Follow-up commit `473f37c` — idle predicate excludes dead/suspended sessions**

External reviewer surfaced that the idle-shutdown callback fed `listSessions().length` straight into the Watchdog predicate. Sessions in `dead` (PTY exited, scrollback retained until the reap TTL fires up to 24h later) and `suspended` (recovery cap-skipped) would keep the count non-zero and block self-termination for up to a full day past the moment the last live PTY closed. Idle shutdown's intent was defeated by tombstone retention.

Fix: added `DaemonSessionManager.listLiveSessions()` that filters to `attached`/`detached`. Idle-snapshot callback switched to use it. Tightened the source-level invariant in `idleShutdown.test.ts` with a negative regex so a future refactor that quietly swaps back to `listSessions()` breaks the build.

**Codex independent review on the follow-up:** P1 none, P2 none. One P3 follow-up surfaced for tracking — `daemon.ping` (`src/daemon/index.ts:758`) and the `[Watchdog] Health:` diagnostic log still use `listSessions().length`. The idle predicate now diverges from the wire-visible diagnostic count. Not a correctness bug but an internal inconsistency worth a separate change with a wire-impact note (external orchestrators parse `daemon.ping.sessions`). Intentionally NOT folded into this PR — daemon.ping is observable contract.

Tests:
- `src/daemon/__tests__/DaemonSessionManager.test.ts` — 2 new cases (mixed-state filter, all-tombstone empty result)
- `src/daemon/__tests__/idleShutdown.test.ts` — invariant tightened, negative regex for the broader accessor
- `scripts/daemon-idle-shutdown-dynamic.mjs` — re-ran end-to-end, PASS in ~5s

PR #73, opened 2026-05-27
fix(mcp): anchor workspace identity to ptyId, resolve owner live

openwong2kim:
## Problem

An in-pane MCP server (e.g. Claude Code) could get **permanently stuck** reporting a workspace id that no longer exists:

- `a2a.whoami` → `"no workspace found for ws-…"`
- `terminal_send` → `"… is not owned by workspace ws-… (actual owner: ws-…)"`

Every identity-gated MCP call (A2A, `terminal_*`, browser routing) failed until the MCP server was restarted. Serious for multi-agent / team-mode use.

## Root cause

Agent self-identity was derived **only** from PTY-create-time state:

- **Path A** — `WMUX_WORKSPACE_ID` env, baked into the shell at create and inherited by Claude Code → MCP child. Frozen for the process lifetime.
- **Path B** — `~/.wmux/pid-map/<pid>` storing the **workspaceId**, written only at create.

The MCP resolver then cached the env value **permanently** (`if (MY_WORKSPACE_ID && workspaceResolved) return`) and never validated it against the live workspace registry. When the workspace id was re-minted (daemon respawn / session restore) while the shell process lived on, the live renderer ownership moved to the new id but the frozen env + pid-map kept the old one. Split-brain, with no self-heal path (`mcp.claimWorkspace` is not exposed as a tool).

## Fix — live-ownership identity

Anchor identity to the **immutable `ptyId`**, not a frozen workspace id:

- **`a2a.resolve.identity`** now treats pid-map entries as `PID → ptyId` and resolves the **current** owning workspace live from the renderer (`input.findOwnerWorkspace`) on every call. Legacy `ws-`-prefixed entries pass through for one cycle; ptyIds with no live owner are omitted (no phantom mappings).
- **`pty.handler` create + reconnect** write `PID → ptyId`. Reconnect re-anchors the map for a surviving shell, so a respawn-adopted session resolves without a restart.
- **MCP resolvers** (`src/mcp`, `src/company/mcp`) prefer the live PID-walk and use `WMUX_WORKSPACE_ID` only as a last-resort hint. An RPC reporting a stale identity (`no workspace found` / `not owned by workspace`) invalidates the in-process cache so the next call self-heals.

`docs/PROTOCOL.md §6.1` updated: path B (live) is now preferred over path A (stale-prone hint), with the ptyId-anchor + self-heal notes.

## Testing

- New unit suite `a2a.rpc.resolveIdentity.test.ts` (6 cases): live resolution, legacy passthrough, dead-pty omission, renderer-error skip, mixed pass, empty dir.
- Reworked `workspace-identity-dynamic.mjs` probe + `mini-pipe-server.mjs` to the new semantics, plus a **stale-env → live regression** scenario (W2): env = `ws-stale-DEAD`, live PID map resolves to `ws-live-current`; asserts the resolver returns the live id, never the stale env. All 6 dynamic scenarios pass.
- Full `vitest run`: 1961 passing (3 pre-existing `StateWriter` fake-timer flakes that pass 34/34 in isolation, unrelated to this change).
- Full `tsc -p tsconfig.json --noEmit`: clean.

Independent `codex review --base main`: **PASS**, no findings against this change.


openwong2kim:
Superseded by #72 — the identity-drift fix (commit 9caac08) was cherry-picked into the daemon-lifecycle-hardening branch and shipped as part of the #72 squash. Closing as already-merged. CHANGELOG and PROTOCOL §6.1 entries are in main.

PR #90, opened 2026-06-01
security: grant token-file ACL by owner SID so non-ASCII usernames can't lock the owner out

junbeom09:
## Summary

`applyRestrictiveWindowsAcl` passed `%USERNAME%` straight to `icacls`. On a Windows profile with a **non-ASCII account name** (e.g. a Korean username), `icacls` parses its argv in the console's legacy OEM codepage and mangles the name into a ghost principal (a corrupted form of the account name with a trailing backslash). It grants Full control to that non-existent account while the **real owner SID gets nothing**, and `/inheritance:r` strips every inherited ACE — so the owner is locked out of their own `~/.wmux-auth-token`. Since `reHardenTokenFileAcl` re-applies this on **every load** (RCA A12 / v2.14.0), the lock-out is permanent and every MCP tool call fails with `wmux auth token not found. Is wmux running?`.

This is an **availability / over-restriction** bug (the owner is wrongly denied), not a token-disclosure bug — no broader-than-stated ACL is written.

## Changes

- Resolve the owner's **SID** via `whoami /user` and grant by `*<SID>` (pure ASCII, codepage-proof) instead of `%USERNAME%`. Fall back to `%USERNAME%` only if the SID can't be resolved.
- Order the icacls argv `/grant:r` **before** `/inheritance:r`, so the owner keeps `WRITE_DAC` through the inheritance strip (granting after the strip can fail silently — the RCA already documented in `docs/SECURITY.md` §1.2).
- Sync `docs/SECURITY.md` §1.2 and `docs/PROTOCOL.md` §5 to the new argv (`icacls /grant:r *<owner-SID>:F /inheritance:r`).
- Tests: add a non-ASCII-username regression and a SID-resolution-failure fallback case; assert the new argv order and that the username is never the grant target.

### Fixed

- Windows: token-file ACL hardening no longer locks the owner out of `~/.wmux-auth-token` on profiles with a non-ASCII account name (e.g. Korean usernames). The ACL now grants Full control to the owner's SID rather than passing `%USERNAME%` to `icacls`, and grants before stripping inheritance. This fixes all MCP tools (`browser_*`, `terminal_*`, `a2a_*`) failing with `wmux auth token not found` on such accounts.

### Security

- Hardening helper now names the owner by SID and grants Full control before stripping inheritance, preventing both the non-ASCII mangle and the silent-grant-failure RCA from `docs/SECURITY.md` §1.2.

## Stability tier impact

- [x] Change to an `internal` surface (no contract impact). — `src/shared/security.ts` is an internal helper; the token-file ACL guarantee in §1.2/§5 is unchanged in intent (owner-only Full control), only the mechanism is corrected. Docs updated in the same PR.

Named Pipe security (`docs/PROTOCOL.md` §5): the token-file ACL **guarantee** is unchanged (owner-only Full control). Only the `icacls` invocation that implements it is corrected. §5 and `docs/SECURITY.md` §1.2 are updated in this PR to match the new argv.

## Test plan

- **Unit tests added/updated:** `src/shared/__tests__/security.test.ts` — non-ASCII-username regression (grant target must be `*<SID>`, never the raw username), SID-resolution-failure fallback to `%USERNAME%`, and the `/grant:r` → `/inheritance:r` argv order for both `secureWriteTokenFile` and `reHardenTokenFileAcl`.
- **Full suite:** `npx vitest run` → 2066 passing / 165 files. `npx tsc --noEmit` clean. `eslint` clean.
- **Manual verification on a real non-ASCII (Korean) Windows profile:**
  1. Reproduced: `icacls ~/.wmux-auth-token` showed a mangled ghost ACE (the account name corrupted with a trailing backslash); owner `Get-Content` → `Access is denied`.
  2. Ran the **compiled** `reHardenTokenFileAcl` (esbuild bundle of the fixed `security.ts`) against the locked file → returned `true`; `icacls` now shows the owner's real SID resolved to the proper `MACHINE\user` principal and the file is owner-readable again (`READ OK, length=36`).

## Reviewer notes

- One residual dead ACE (the old mangled ghost principal) can remain on files already corrupted by the old code, because `/grant:r` only replaces the SID's ACE. It's a non-resolving (dead) principal, is **not** cumulative (new code only ever grants by SID), and is harmless. A single atomic `icacls` call is kept on purpose rather than a `/reset`-then-grant pair, which would briefly expose the inherited ACL if the grant step failed. Deleting the token file forces a clean SID-only rewrite on next launch.


openwong2kim:
Merged to `main` — thank you for this, genuinely well-scoped fix. 🙏

The diagnosis was excellent: pinning the lock-out on the OEM-codepage **principal** mangle (the file *path* stays Unicode-clean, only the account name gets corrupted) is exactly right, and grant-by-SID is the correct codepage-proof solution. Particular thanks for verifying it on a real non-ASCII (Korean) profile rather than only in unit tests, and for the `/grant:r` → `/inheritance:r` reorder — both held up under review.

One thing I added on top before merging (commit `87c3d1d`), so you're aware of the change to your code:

**Guarded the `%USERNAME%` fallback against non-ASCII names.** In `applyRestrictiveWindowsAcl`, when `getCurrentUserSid()` returns `null` the principal fell back to `process.env.USERNAME`. On an ASCII account that's fine — but on the exact account class this PR protects, a transient `whoami` failure (AV/EDR block, locked-down image, a momentary `CreateProcess` failure) would pass the Korean name straight back to icacls and re-create the ghost-principal lock-out. And because `reHardenTokenFileAcl` re-applies on every load, one such failure would re-lock the file permanently — the very thing this PR fixes.

The guard gates the fallback to **pure-ASCII names only** (`Buffer.byteLength(username) === username.length` — avoids `no-control-regex`). For a non-ASCII or empty `USERNAME` it now refuses and throws instead of running a mangling-prone icacls, so callers fail safe:
- `secureWriteTokenFile` deletes the token and rethrows (fail-closed);
- `reHardenTokenFileAcl` returns `false` without touching the existing ACL (best-effort — re-locking the owner out is worse than leaving the current ACL as-is).

Added two regressions (write-path + re-harden-path) that fail against the pre-guard code and pass with the guard, and synced the new clause into `docs/SECURITY.md` §1.2.

Full suite **2068 passing**, `tsc` + `eslint` clean, CI green on windows-latest / macos-14 / ubuntu-22.04. This will go out with the next release. Thanks again! 🚀

junbeom09:
Thanks for the thorough review and merge! 🙏

The fallback guard was something I missed — given that `reHardenTokenFileAcl` re-applies on every load, you're right that a single failed SID lookup would let the Korean username re-create the lock-out.

The split between fail-closed in `secureWriteTokenFile` and preserving the existing ACL in `reHardenTokenFileAcl` makes a lot of sense too. Learned a lot from this — looking forward to the next release 🚀

PR #101, opened 2026-06-04
feat: per-workspace process profiles (env + startup command for new panes)

snowyukitty:
## What

Adds generic **per-workspace process profiles**. Each workspace can define environment variables and an optional startup command that are applied to **new panes only**:

```ts
interface WorkspaceProfile {
  env?: Record<string, string>;
  defaultPaneCommand?: string;
}
```

Right-click a workspace in the sidebar → **"Configure profile…"** to edit env rows + a startup command. A small ⚙ badge marks workspaces that have a profile.

This is **generic by design** — no provider is hardcoded. It lets you point `CLAUDE_CONFIG_DIR`, `CODEX_HOME`, an SSH wrapper, cloud-CLI config dirs, etc. at different accounts per workspace (Workspace A → account A, Workspace B → account B). It is **environment separation, not an OS-level security sandbox**.

## Semantics

- New PTYs get `safeInheritedEnv → profile.env → forced wmux identity vars`.
- **Existing PTYs are never mutated**; **recovered daemon PTYs keep their persisted create-time env** (recovery already replays `DaemonSession.env`, and the merged overlay is persisted there at create time).
- `defaultPaneCommand` is **written into the shell after creation** (deferred, best-effort), not spawned as the executable — this preserves the allowed-shell check and quoting, matching the existing company-provisioning pattern.
- Workspaces with no profile behave exactly as before (helper returns options untouched).

## Security

- **Recursive IPC redaction** (prerequisite, commit 1): the structured IPC error logger previously redacted only top-level keys, so nested secrets in a `pty:create` payload could leak into `args_summary`. It now walks objects/arrays (depth-capped), redacts sensitive + command-like keys at any depth, and summarizes env maps to a key count.
- Profile env is carried as a **separate `profileEnv`** field into the daemon so it survives the daemon's own `buildSafeChildEnv` pass (an intentional `*_KEY`/`*_TOKEN` would otherwise be stripped). The overlay applies *after* the inherited-env denylist and *before* forced identity vars.
- Reserved `WMUX_*` keys are rejected at validation **and** skipped in the spawn layer, so a profile can never spoof `WMUX_WORKSPACE_ID`/`WMUX_SURFACE_ID`/`WMUX_SOCKET_PATH` or reintroduce a stripped auth token.
- Profile values are **not** logged and **not** included in copy-session-info / drag-export markdown (regression-locked by tests). They are stored in plaintext in the local session file; the UI says so and recommends config-dir paths over raw keys.
- Profile lives on `Workspace.profile`, **not** `WorkspaceMetadata` — it never travels the metadata event/RPC bus.

## Callsites

Every new-pane create path is routed through the profile overlay: Terminal self-create, Pane add-tab, CommandPalette new surface, AppLayout empty-leaf auto-create, Ctrl+T (useKeyboard), and the `surface.new` RPC. Floating panes (no owning workspace) and company provisioning (own startup-command path) are intentionally left unchanged.

## Commits

1. `fix(ipc)` recursive `args_summary` redaction (+ tests)
2. `feat(types)` `WorkspaceProfile` + shared validation/normalization (+ tests)
3. `feat(pty)` env-overlay + startup-command plumbing, local + daemon (+ unit + daemon runtime test)
4. `feat(renderer)` helper, store action, loadSession sanitization, all callsites (+ tests)
5. `feat(ui)` sidebar context menu + editor modal (+ markdown no-leak test)
6. `docs(changelog)` Unreleased entry

## Tests

`npm test` green (2134 parallel + runtime). New coverage: recursive redaction, profile validation (incl. reserved-key rejection + caps), `applyProfileEnv`, the renderer helper precedence rules, store set/clear/no-publish, loadSession sanitization, a daemon **runtime** test proving an intentional `*_KEY` survives the daemon filter while a reserved key can't spoof identity, and a security regression test that profile env never leaks into export markdown.


openwong2kim:
Thanks for this, @snowyukitty — genuinely useful feature, and the implementation is careful where it counts (overlay applied *after* `buildSafeChildEnv` so an intentional key isn't stripped, reserved `WMUX_*` skip, the IPC redaction pass, key/value size caps). I read the whole diff. I'd like to land it, and I think we're one focused round away. Splitting what I found into "objective fixes" and "a design direction I'd like your take on."

### Objective fixes (happy to push these to your branch — maintainer edits are on)

**1. `wrapHandler` env-summary can be bypassed.** The summary branch only fires for plain objects:
```ts
else if (ENV_SUMMARY_KEYS.test(k) && v !== null && typeof v === 'object' && !Array.isArray(v))
```
`env` / `profileEnv` don't match `SENSITIVE_KEY_PATTERN`, so a non-object value under those keys (e.g. `{ env: "ANTHROPIC_API_KEY=sk-..." }`, or an array) skips the `keyCount` summary and falls through to being kept/stringified — which can land a secret in the structured log. Suggest: treat any value under `env`/`profileEnv` as summary-or-redact regardless of type.

**2. macOS CI is red.** `DaemonSessionManager.profile.runtime.test.ts` calls `createSession(...)`, which spawns a real shell; on `macos-14` that's `posix_spawnp failed`. What it actually asserts (profileEnv survives the filter and lands in `meta.env`) is pure merge logic — `shared/__tests__/workspaceProfile.test.ts` already covers `applyProfileEnv` without a live PTY. Moving the assertion there (or asserting `meta.env` without depending on a successful spawn) makes it green on all three runners. I'll land this one now, independent of 3/4.

### Design direction (would like your take before changing anything)

**3. The recovery path drops profile keys — and the fix points at a cleaner shape.** `profileEnv` is threaded through the create RPC, but the daemon's three recovery sites in `daemon/index.ts` (the `createSession({ env: session.env, ... })` calls, ~L371/L448/L487) replay only `env`, never `profileEnv`. And `createSession` re-runs `buildSafeChildEnv(params.env)` (`DaemonSessionManager.ts:140`), which strips the intentional `*_KEY`/`*_TOKEN` merged in at create time. Net: after a daemon recovery a pane loses e.g. `GEMINI_API_KEY`, contradicting the docstring's "reproducing the exact create-time environment."

Rather than re-thread `profileEnv` through every recovery site, I'd suggest keeping the daemon generic: resolve the profile in **main** (where `PTYManager` already calls `applyProfileEnv`), hand the daemon a fully-resolved env, and have the daemon **replay the persisted env verbatim on recovery** instead of re-filtering it. That dissolves all three recovery sites at once and keeps per-workspace policy out of the substrate — the daemon shouldn't need to know what a "profile" is. As a bonus the runtime test in #2 naturally becomes a main/shared unit test.

**4. Env storage: reference over secret.** For the common case (per-account `CLAUDE_CONFIG_DIR` / `CODEX_HOME`, SSH wrappers, `PATH` tweaks) the values are *paths*, not secrets — plaintext in the session file is fine. The risk only appears when someone pastes a raw key. We already ship `isSensitiveEnvKey` in `shared/envFilter.ts`; I'd like the profile editor to use it to **warn/decline on secret-looking keys and steer toward pointing at a config dir** instead of pasting the key. That keeps profiles in plaintext-safe territory and closes the "secrets in session.json" footgun by policy rather than by a warning string. (If raw-key support is ever needed, Electron `safeStorage` — DPAPI/Keychain/libsecret, resolved in main — is the path. We'd rather not store raw secrets at all for now.)

If this direction works for you, I'm glad to push the two objective fixes and we can take 3/4 together (or you drive — whatever you prefer). Thanks again for the contribution and the care in the diff.

openwong2kim:
Merged — thank you, @snowyukitty. You turned around all four review points in about 90 minutes and the fixes are clean: env-summary redaction now covers any value shape, the daemon is fully profile-agnostic with env resolved in main (recovery replays verbatim, so create-time env survives), the macOS runtime test is gone, and the secret-key hint steers toward config-dir references. The event-driven startup-command rework — first-output gate plus delivery retry — is a real improvement over the fixed delay too.

One follow-up we'll handle on our side, separate from this PR: now that a profile can persist env into the session file, we want sessions.json to get the same owner-only ACL hardening the auth-token file already has on Windows. That's a repo-wide hygiene item, not anything you missed — just flagging it so it's tracked.

Thanks again for a well-scoped contribution and the fast, thoughtful iteration.

snowyukitty:
Thanks @openwong2kim, and thanks for merging this.

I also wanted to say thanks for building and maintaining wmux. I reached a point where I felt I needed a tool like this, and that's how I found the project. It's rare to find something that stays this focused while still being pleasant to work with.

I really appreciated the review as well. The comments were detailed, practical, and easy to act on. More importantly, they didn't just point out issues — they helped clarify the shape of the solution. The suggestion to resolve env in main and keep the daemon profile-agnostic ended up simplifying the design quite a bit and made the recovery semantics cleaner than what I originally had.

The sessions.json ACL follow-up makes sense too. Given that profiles can now persist env there, matching the auth-token file's owner-only protection feels like the right next step.

For me, this feature solves a very practical workflow: one machine, multiple AI CLI accounts (personal vs. company), with each workspace automatically pinned to the right identity. That's the use case I originally built it for, so it's great to see it land in a much cleaner shape than the first version.

Here's it running locally:
image_2026-06-05_133323078
Snipaste_2026-06-05_11-46-09

I hope this feature will be helpful to people like me who genuinely need it. 💖

PR #103, opened 2026-06-05
fix(profile): allowlist path-pointer credential vars; non-destructive load (follow-up to #101)

snowyukitty:
Follow-up to #101. Two corrections to the secret-name policy, surfaced by a post-merge review pass, plus two doc/comment nits.

## Fixes

**1. `GOOGLE_APPLICATION_CREDENTIALS` was being dropped (major).** It matches the `*_CREDENTIALS` denylist by name, but its value is a *path* to a service-account JSON — the "reference over secret" pattern we encourage, and it's listed as supported in `docs/workspace-profiles.md`. Added a small `SECRET_KEY_ALLOWLIST` (`GOOGLE_APPLICATION_CREDENTIALS`, `AWS_SHARED_CREDENTIALS_FILE`) so path-pointer credential vars aren't treated as raw secrets.

**2. Load-time silent data loss (major).** The policy dropped secret-named keys on both save *and* load, so a `session.json` written before #101 would silently lose a working key (or the whole profile) on next load — without un-storing the already-persisted plaintext value. The drop now applies **only at the editor/save boundary** (`setWorkspaceProfile` passes `dropSecretKeys`); load is non-destructive. A legacy key keeps working until the user re-saves the profile, at which point the editor flags it and drops it. `normalizeEnv` / `normalizeWorkspaceProfile` gained a `dropSecretKeys` flag (default `false` = preserve).

## Nits

- Docs contradicted themselves on `WMUX_SOCKET_PATH` (the mental model says daemon-mode doesn't force it; the identity bullet said it always did). Corrected: `WORKSPACE_ID`/`SURFACE_ID` in both modes, `SOCKET_PATH` local-only.
- Clarified in comments/tests that the spawn layer (`resolveSpawnEnv` / `applyProfileEnv`) applies whatever profile keys it's given — the editor policy is the layer that decides which keys may be configured. They now differ, so the old "intentional `*_KEY` survives" wording was misleading.

## Tests

`npm test` green (2157 parallel + runtime). New coverage: allowlist exemption, save-drops-but-load-preserves, and a load-time legacy-key-preserved regression.

### DIFF
--- a/docs/workspace-profiles.md
+++ b/docs/workspace-profiles.md
@@ -40,9 +40,10 @@ Consequences worth internalizing:
   restart).
 - **No cross-workspace leakage.** A pane only ever sees its own workspace's
   profile.
-- **Identity is protected.** `WMUX_WORKSPACE_ID`, `WMUX_SURFACE_ID`, and
-  `WMUX_SOCKET_PATH` are forced last; a profile cannot override them. Any
-  `WMUX_*` key you type is rejected.
+- **Identity is protected.** The wmux identity vars are forced last, so a
+  profile cannot override them, and any `WMUX_*` key you type is rejected.
+  (`WMUX_WORKSPACE_ID` / `WMUX_SURFACE_ID` in both modes; `WMUX_SOCKET_PATH`
+  in local mode only — see the mental-model note above.)
 - **Profile env overwrites, it does not append.** If you set `PATH` in a
   profile it *replaces* the inherited `PATH` — almost never what you want. Set
   tool-specific vars (config-dir pointers), not `PATH`.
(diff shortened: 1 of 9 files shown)


openwong2kim:
Merged — thanks again, @snowyukitty. Good catches on the post-merge pass: the `GOOGLE_APPLICATION_CREDENTIALS` false-drop and the destructive load were both real, and the save-boundary/load split is the right call — dropping on load would only break working config without un-storing the plaintext value. The `dropSecretKeys` default-false and the narrow allowlist are nicely scoped, and the mechanism-vs-policy doc/comment cleanups are appreciated.

One thing I'd genuinely like your take on, separate from this merge. The editor's amber hint reads "stored in plaintext," but with the save-time drop a secret-named key isn't stored at all — it's dropped. The message and the behavior point in different directions, and I'm honestly torn on which way to reconcile them:

- **(a)** reword the hint to match the drop ("secret-named keys aren't saved — point at a config directory instead"), keeping the strict reference-only policy; or
- **(b)** switch to warn-and-store — let a deliberately-pasted key persist behind the plaintext warning the text already implies, trusting the user who chose to paste it. (This one reopens the plaintext-at-rest question for the session file, so it's not purely cosmetic.)

You've been living in this code — which direction feels right to you? No rush, and not blocking anything; I just don't want to pick the policy unilaterally when you've got the context.

Thanks for the fast, careful iteration — two clean follow-ups in an afternoon.

PR #104, opened 2026-06-05
fix(mcp): browser DOM tools read app shell instead of webview in packaged builds

junbeom09:
## Summary

In packaged (production) builds, the MCP browser DOM tools
(`browser_evaluate` / `browser_snapshot` / `browser_extract_text` /
`browser_smart_snapshot`) read the wmux **app shell** (main window) DOM instead
of the embedded `<webview>`'s real page DOM, so page content extraction
silently failed. Dev builds were unaffected. Root cause: `PlaywrightEngine`
recognized the shell only by dev-server/devtools/chrome URLs, so the packaged
shell — a `file:///.../.vite/renderer/main_window/index.html` URL — slipped
past `getPage()`'s "first non-shell page" filter and was returned as if it were
the webview. (`navigate`/`screenshot` were fine because they go through the
`webviewCdpManager` RPC path, which targets the guest `webContentsId`
directly.)

This PR hardens shell detection using the main process's **actual runtime shell
URL** and reorders `getPage()` to prefer **positive** webview identification
(registered `targetId`) over URL-based negative filtering.

## Changes

- **`browser.cdp.info` (main, `browser.rpc.ts`)**: expose an optional
  `shellUrl` = `getWindow().webContents.getURL()` — the real main-window URL
  (dev: `http://localhost…`, packaged:
  `file:///.../.vite/renderer/main_window/index.html`). Empty (mid-load) or
  throwing (window destroyed) URLs are omitted so the engine keeps any prior
  value. The guest `<webview>` is a separate `webContents` and never appears
  here.
- **`PlaywrightEngine.ts`**:
  - Cache `shellUrl` from every `browser.cdp.info` response; clear on
    `disconnect()`.
  - New `isShellPage()` — exact-match against the runtime `shellUrl`, with the
    static `isElectronShellUrl()` heuristic retained as a **secondary
    defense-in-depth fallback** when the runtime URL isn't known yet. All four
    negative-filter sites now route through `isShellPage()`.
  - `getPage()` reorder: run `findViaTargetDomain()` (positive `targetId`
    match) **before** the "first non-shell page" filter, so the shell is never
    returned when a registered target exists.
  - The static `file://` heuristic itself was narrowed earlier in this branch
    to `.vite/renderer/main_window/index.html` so a user-opened
    `…/main_window/index.html` page stays drivable.
- **Tests**: +9 cases across `PlaywrightEngine.test.ts` (shellUrl
  caching/exact-match, empty-ignore, heuristic fallback, dev-shell,
  disconnect-reset, and a `getPage()` integration test that returns the webview
  not the shell) and `browser.rpc.test.ts` (shellUrl exposed / omitted on
  empty / omitted on throw). Also fixes a latent test-mock path bug:
  `vi.mock('../wmux-client')` resolved to a non-existent module from the
  `__tests__/` dir → corrected to `'../../wmux-client'` (previously unnoticed
  because no prior test drove a `getPage()` path that calls `sendRpc`).

### Fixed

- MCP browser DOM tools (`browser_evaluate`, `browser_snapshot`,
  `browser_extract_text`, `browser_smart_snapshot`) now read the embedded
  page's DOM instead of the wmux app-shell DOM in packaged builds.

## Stability tier impact

- [x] Change to an `experimental` surface (note in release notes).

`browser.cdp.info` is classified `experimental` in `docs/api/inventory.md`
(line 111). The change is additive — a new optional `shellUrl` field on the
response — and backward compatible (older engines ignore it; the field is
omitted when unavailable).

New optional field on an existing experimental RPC response (`browser.cdp.info`
→ `shellUrl`). No change to PaneMetadata, EventBus, permission enforcement
(`browser.read` gate unchanged), or Named Pipe security. No new RPC method or
event type. `docs/api/inventory.md` already lists `browser.cdp.info` with a
"various" shape; an explicit field note there is an optional follow-up since
the surface is experimental.

## Test plan

- **Unit/integration tests added (+9)**:
  - `src/mcp/playwright/__tests__/PlaywrightEngine.test.ts` — shellUrl
    cache/exact-match, empty/missing ignore, heuristic fallback, dev-server
    shell, disconnect reset, and `getPage()` returns the webview over the
    exact-match shell.
  - `src/main/pipe/handlers/__tests__/browser.rpc.test.ts` — `shellUrl` exposed
    from `getURL()`; omitted when empty (mid-load) or when `getURL()` throws
    (destroyed).
- **Full suite**: `npx vitest run` → **2081 passed / 1 failed (2082 total)**.
  The single failure is environment-specific (see Reviewer notes), not a
  regression.
- **Typecheck/build**: `tsc -p tsconfig.mcp.json --noEmit` ✅, `npm run
  build:mcp` ✅.
- **Lint (changed files)**: no new issues. Two pre-existing warnings/errors in
  `PlaywrightEngine.ts` (`BrowserContext` unused import, empty constructor)
  predate this branch and are unrelated.
- **Manual verification**: not run in a packaged build this cycle (see
  Limitations). The fix is covered by the unit/integration tests above.

## Related issues / PRs

n/a

## Reviewer checklist

- [x] CHANGELOG entry above is filled in.
- [x] Stability tier impact is correctly classified (experimental, additive).
- [x] Substrate contract docs: experimental surface; inventory note optional
      follow-up.
- [x] Tests cover the new behavior and the regression case.
- [x] No secrets/tokens/.env/large binaries committed.

---

### Reviewer notes (limitations, risks, trade-offs)

**Limitations**
- No packaged-build manual verification this cycle; correctness rests on the
  unit/integration tests. A quick prod smoke test (`browser_extract_text` on a
  real page) before/after is the highest-value manual check if a reviewer can
  run it.
- The extraction tools (`smart_snapshot`/`extract_text`/`extract_data`) still
  have **no RPC fallback** — they depend on `getPage()` returning the real
  webview. This PR makes that reliable but does not add a fallback for them
  (that was the larger "route DOM tools through `browser.rpc`" option, scoped
  out to avoid reimplementing extraction as injected JS and losing Playwright
  features).

**Risks / trade-offs**
- **Extra CDP round-trip on the common path**: `getPage()` now runs
  `findViaTargetDomain()` (which issues `Target.getTargets` + a
  `browser.cdp.info` RPC) *before* the cheap in-memory page filter. This adds a
  small, bounded latency to the hot path in exchange for never returning the
  shell. The negative filter remains as the immediate next step, so behavior
  degrades gracefully if positive matching yields nothing.
- **`shellUrl` exact-match depends on the runtime URL**: if the main process
  doesn't supply it (older main, or `getURL()` empty mid-load), detection falls
  back to the static `file://` heuristic (A). This is intentional
  defense-in-depth — two independent mechanisms must both fail to regress — but
  it does mean the exact-match guarantee is "best-effort, with a heuristic
  floor."
- **Heuristic floor is still build-path-coupled**: the retained `A` fallback
  matches `.vite/renderer/main_window/index.html`. If forge/vite output layout
  changes *and* the runtime `shellUrl` is simultaneously unavailable, the
  fallback could go stale — but the primary runtime-URL path is immune to that,
  so this only matters in the double-failure case.

**Environment-specific test failure (not a regression)**
- The one failing test is `src/shared/__tests__/security.test.ts >
  secureWriteTokenFile`. It fails because this dev machine's `USERNAME` is
  **non-ASCII (Korean)**, which the token-file ACL hardening (commit #90)
  deliberately refuses. Verified via `git stash` that it fails identically on
  `main` without this branch's changes — so it is pre-existing and unrelated to
  this PR. CI (ASCII runner usernames) should be green.

### DIFF
--- a/src/main/pipe/handlers/__tests__/browser.rpc.test.ts
+++ b/src/main/pipe/handlers/__tests__/browser.rpc.test.ts
@@ -42,7 +42,7 @@ describe('registerBrowserRpc', () => {
     sendToRendererMock.mockResolvedValue({ ok: true });
   });
 
-  function register(): RpcRouter {
+  function register(getWindow: () => BrowserWindow | null = () => null): RpcRouter {
     const router = new RpcRouter();
     const webviewCdpManager = {
       getTarget: vi.fn(() => ({ surfaceId: 'surface-1', webContentsId: 42, targetId: 'target-1', wsUrl: 'ws://127.0.0.1/devtools/page/target-1' })),
@@ -51,8 +51,14 @@ describe('registerBrowserRpc', () => {
       waitForTarget: vi.fn(),
     };
 
-    registerBrowserRpc(router, (() => null) as () => BrowserWindow | null, webviewCdpManager as never);
+    registerBrowserRpc(router, getWindow, webviewCdpManager as never);
     return router;
+  }
+
+  /** Build a fake main window whose webContents.getURL() returns `url`. */
+  function windowWithUrl(url: string | (() => string)): () => BrowserWindow {
+    const getURL = typeof url === 'function' ? url : () => url;
+    return () => ({ webContents: { getURL } }) as unknown as BrowserWindow;
   }
 
   it('does not expose browser.cdp.send through the RPC router', async () => {
@@ -95,10 +101,61 @@ describe('registerBrowserRpc', () => {
 
     expect(response.ok).toBe(true);
     if (response.ok) {
+      // No window (getWindow → null): shellUrl is omitted, not null.
       expect(response.result).toEqual({
         cdpPort: 18800,
         targets: [{ surfaceId: 'surface-1', targetId: 'target-1' }],
       });
+    }
+  });
+
+  it('browser.cdp.info exposes the main-window URL as shellUrl', async () => {
+    const router = register(
+      windowWithUrl('file:///x/.vite/renderer/main_window/index.html'),
+    );
+
+    const response = await router.dispatch({
+      id: '3b',
+      method: 'browser.cdp.info',
+      params: {},
+    });
+
+    expect(response.ok).toBe(true);
+    if (response.ok) {
+      expect(response.result).toMatchObject({
+        cdpPort: 18800,
+        shellUrl: 'file:///x/.vite/renderer/main_window/index.html',
+      });
+    }
+  });
+
+  it('browser.cdp.info omits shellUrl when the URL is empty (mid-load)', async () => {
+    const router = register(windowWithUrl(''));
+
+    const response = await router.dispatch({
+      id: '3c',
+      method: 'browser.cdp.info',
+      params: {},
+    });
+
+    expect(response.ok).toBe(true);
+    if (response.ok) {
+      expect(response.result).not.toHaveProperty('shellUrl');
+    }
+  });
+
+  it('browser.cdp.info omits shellUrl when getURL throws (window destroyed)', async () => {
+    const router = register(windowWithUrl(() => { throw new Error('destroyed'); }));
+
+    const response = await router.dispatch({
+      id: '3d',
+      method: 'browser.cdp.info',
+      params: {},
+    });
+
+    expect(response.ok).toBe(true);
+    if (response.ok) {
+      expect(response.result).not.toHaveProperty('shellUrl');
     }
   });
 
--- a/src/main/pipe/handlers/browser.rpc.ts
+++ b/src/main/pipe/handlers/browser.rpc.ts
@@ -246,8 +246,22 @@ export function registerBrowserRpc(router: RpcRouter, getWindow: GetWindow, webv
     }
 
     const cdpPort: number = webviewCdpManager.getCdpPort();
+
+    // Expose the actual runtime URL of the main-window webContents (the app
+    // shell) so the Playwright engine can recognize the shell by exact-match
+    // instead of guessing from build-path shape. dev → http://localhost:..,
+    // packaged → file:///.../.vite/renderer/main_window/index.html. The guest
+    // <webview> is a separate webContents and never appears here. Suppress an
+    // empty URL (window still mid-load) so the engine keeps any prior value.
+    let shellUrl: string | undefined;
+    try {
+      const url = getWindow()?.webContents.getURL();
+      if (url && url.length > 0) shellUrl = url;
+    } catch { /* window destroyed — omit shellUrl */ }
+
     return {
       cdpPort,
+      ...(shellUrl && { shellUrl }),
       targets: targets.map((t) => ({
         surfaceId: t.surfaceId,
         targetId: t.targetId,
--- a/src/mcp/playwright/PlaywrightEngine.ts
+++ b/src/mcp/playwright/PlaywrightEngine.ts
@@ -10,6 +10,14 @@ interface CdpTargetInfo {
 
 interface CdpInfoResponse {
   cdpPort: number;
+  /**
+   * The actual runtime URL of the main-window webContents (the app shell),
+   * as reported by the main process. Optional: absent on older mains or when
+   * the window is mid-load (empty URL is suppressed). When present it is the
+   * authoritative shell identifier; when absent we fall back to the static
+   * isElectronShellUrl() heuristic. See browser.cdp.info handler.
+   */
+  shellUrl?: string;
   targets: CdpTargetInfo[];
 }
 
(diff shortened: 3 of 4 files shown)


openwong2kim:
Merged — thank you, @junbeom09. Really well-put-together PR. The root-cause writeup (the packaged `file://` shell slipping past the dev-URL filter, while navigate/screenshot stayed fine through the RPC path) made the diagnosis easy to follow, and the fix is well-chosen: preferring the runtime `shellUrl` exact-match is the robust call — self-correcting regardless of build-path layout — and reordering `getPage()` to pin the webview by `targetId` before the negative filter tightens multi-surface selection too, not just the shell case. The defense-in-depth heuristic floor, the consistent caching across every `cdp.info` site, and the +9 tests (including the getPage-returns-the-webview regression) are all solid. Nice catch on the latent test-mock path, and the correct read that the non-ASCII `USERNAME` token-ACL test failure is pre-existing (#90), not from this branch.

One honest follow-through on our side, matching the limitation you flagged: since this only manifests in packaged builds, we'll run a packaged smoke test (`browser_extract_text` on a real page) at the release gate before it ships, on top of the unit coverage.

Thanks for the thorough writeup and the care — this was an easy review.

junbeom09:
**Follow-up: end-to-end verification, separating what I observed from what I'm inferring.**

I couldn't capture the MCP server's `[PlaywrightEngine]` stderr in this run, so I've split **observed facts** from **inferences read off the code**.

### How I verified (reproducible)

Built this branch (`npm run make`) and loaded its bundle (`out/wmux-win32-x64/resources/mcp-bundle/index.js`) as a second MCP server in a Claude session inside a wmux pane (inherits `WMUX_*`, connects to the daemon), next to the unmodified installed `app-2.16.2` bundle, then ran identical calls on the same workspace.

*(Fact: `npm run make` packaged a runnable app under `out/`; the only failure was the final `electron-winstaller` `.exe` step, `Fatal error: Unable to load file`. Inference: that installer step looks unrelated to this change — it's the `.exe` wrapper, not the app build — but I did not root-cause it.)*

### Observed (facts), same target `https://github.com`

| call | installed (unfixed) | this branch |
|---|---|---|
| `browser_evaluate` → `location.href` | `file:///…/.vite/renderer/main_window/index.html` (the shell) | `https://github.com/` |
| `browser_extract_text` | returns app-shell UI text | errors: `No browser page available …` |

Also a fact, via the code path: that exact `browser_extract_text` error is thrown only when `getPage()` is falsy (`extraction.ts:103-106`), so on this branch **`getPage()` returned `null`** for that call. `browser_extract_data` / `browser_smart_snapshot` share the same `getPage()`-or-throw shape and no RPC fallback.

### Confirmed in the code (facts)

- Extraction tools need a Playwright `Page` and have no fallback: `extract_text` → `extractMarkdown(page, …)` (`extraction.ts:103-106`), `extract_data` → `extractStructuredData(page, …)` (`extraction.ts:144-147`).
- `browser_evaluate` *does* have a fallback: `getPage(...).catch(() => null)`, and if null it calls `sendRpc('browser.evaluate', …)` (`inspection.ts:321-331`), which the main process serves by driving the guest webContents directly: `webContents.fromId(target.webContentsId).debugger.sendCommand('Runtime.evaluate', …)` (`browser.rpc.ts:300-313`) — the `webviewCdpManager` path `navigate`/`screenshot` use.
- `_getPageImpl` returns `null` after all strategies fail (`PlaywrightEngine.ts:367`); `findViaTargetDomain` has an explicit "Attach did not create a page" → null branch (`PlaywrightEngine.ts:456`); `getAllPages()` is just `browser.contexts().pages()` (`PlaywrightEngine.ts:258`).

### Inference (read off the code; not directly observed — no runtime logs captured)

- **Likely cause of the separate bug:** `playwright-core` (`connectOverCDP`) doesn't surface the Electron `<webview>` guest as a `Page` even after `Target.attachToTarget`, so every `getPage()` strategy comes up empty → `null` → the Playwright-`Page`-only extraction tools throw. I'm inferring this specific path from the code; I did not confirm which strategy failed via logs.
- **Likely why `browser_evaluate` returned the real page on this branch:** `getPage()` was `null`, so it took the RPC fallback above (main-process webContents), which doesn't need a Playwright `Page`. (Observed: it returned `https://github.com/`. Inferred: that it went through the fallback rather than a Playwright `Page`.)
- **Likely relationship to this PR:** independent of the shell-detection change. Before the fix, `getPage()` returned the shell `Page`, so the extraction tools returned shell content instead of erroring — i.e. this PR didn't cause that bug, it unmasked it. (Observed: installed bundle returns shell content. Inferred: that this is the masking mechanism.)

Caveat (inference): this ran the branch *bundle* against the *unmodified* daemon, so `browser.cdp.info` wouldn't supply `shellUrl` and the engine would fall back to the static `.vite/renderer/main_window/index.html` heuristic. The runtime-`shellUrl` exact-match path is covered by the unit tests, not this run.

---

Apologies that part of this is inferred rather than directly observed — I was working under time constraints and couldn't capture the runtime `[PlaywrightEngine]` logs to nail down every step, so please take the inference section as "best read of the code," not verified fact. And thank you for building wmux — it's genuinely a pleasure to work in, and I've learned a ton digging through this codebase. 🙏

openwong2kim:
Great write-up, and thanks for actually packaging it and running the two side by side. That's exactly the verification I was hoping someone could do.

Your read matches mine. The shell-detection change does its job (evaluate landing on `https://github.com/` instead of the shell URL is the tell), and you've pinned the other half correctly: the extraction and snapshot tools have no RPC fallback, so once getPage() stops handing back the shell they just go null and throw. evaluate only survives because of its `sendRpc('browser.evaluate')` path. So the PR didn't break those tools, it stopped them from quietly returning the wrong DOM and made the failure honest.

The "connectOverCDP doesn't surface the webview as a Page" theory lines up with what I see in the code too. The real fix for those tools is to route them through `browser.rpc` the way evaluate and navigate already do, which is the larger piece you scoped out here. Opened #105 to track it.

And no need to apologize for the inference section. Splitting observed from inferred is the right way to write up a result you couldn't fully instrument, and it made the whole thing easy to trust. Thanks again for digging this deep.

Issue #105, opened 2026-06-05
Browser extraction/snapshot tools have no RPC fallback and error in packaged builds


openwong2kim:
Splitting this out of #104. That PR fixed the shell-detection bug (getPage() no longer hands back the wmux app shell), but @junbeom09's packaged verification surfaced a second problem sitting underneath it.

In a packaged build, `connectOverCDP` (playwright-core) doesn't surface the Electron `<webview>` guest as a Playwright `Page`. So once getPage() correctly refuses the shell, every strategy comes up empty and it returns null. The tools that need a real `Page` and have no fallback then throw `No browser page available`:

- `browser_extract_text`, `browser_extract_data` (`extraction.ts`)
- `browser_smart_snapshot`, `browser_snapshot`, and the other `getPage()`-only paths in `inspection.ts`

`browser_evaluate` keeps working because it already falls back to `sendRpc('browser.evaluate')`, which drives the guest webContents directly in the main process (`browser.rpc.ts`). `navigate` and `screenshot` use the same RPC path. So the RPC route reaches the webview fine; the Playwright-`Page` route doesn't.

Before #104 these tools quietly returned the shell's DOM instead of erroring, so the breakage was masked. #104 made the failure honest. This issue is about making them actually work.

### Direction

Give the extraction/snapshot tools the same fallback evaluate has: when getPage() is null, route through `browser.rpc` against the guest webContents. The catch (already called out in #104) is that these tools lean on Playwright features, so doing it as injected JS over CDP means reimplementing some of that. Probably worth deciding per-tool how much to reroute vs reimplement rather than one blanket approach.

### Verification

This only reproduces in a packaged build, so the fix needs a real packaged smoke test (whole app, new daemon included), not just dev or unit tests. @junbeom09's run used the new bundle against the 2.16.2 daemon, so the runtime `shellUrl` path wasn't exercised either. A full packaged build covers both.

Repro and analysis are in #104.

PR #108, opened 2026-06-05
feat(mcp): RPC fallback for browser_console/network/response_body in packaged builds (#106)

### LINKED ISSUE #106: Browser console/network/response_body tools have no RPC fallback in packaged builds
Split out from #105.

#105 added an RPC fallback for the one-shot DOM tools (`browser_extract_text`,
`browser_extract_data`, `browser_smart_snapshot`, `browser_highlight`) so they
work in packaged builds where `connectOverCDP` doesn't surface the `<webview>`
guest as a Playwright `Page`. Three inspection tools were left out on purpose:

- `browser_console` (`page.on('console')`)
- `browser_network` (`page.on('request' | 'response')`)
- `browser_response_body` (depends on the network capture buffer)

These don't fit the same fix. The others read the DOM once, so they reroute
cleanly through `browser.evaluate` (a one-shot `Runtime.evaluate`). console and
network instead rely on Playwright's **event subscription** model — they need a
listener attached for the lifetime of the page. A one-shot evaluate can't stream
those events.

Making them work in packaged builds means a main-process CDP event-capture
subsystem on the guest `webContents`:

- enable + listen to `Runtime.consoleAPICalled` (console) and the `Network`
  domain (`requestWillBeSent` / `responseReceived` / `loadingFinished`),
- buffer entries in the main process (mirroring the bounded WeakMap capture the
  MCP side already does in `inspection.ts`, with the same caps from #102),
- expose new RPC methods to drain/clear the buffers,
- wire the three tools to fall back to those when `getPage()` is null.

That's a separate feature, not a fallback, so it was scoped out of #105. Until
then these three keep their honest "No browser page available" error in packaged
builds (they were silently returning the app shell's data before #104).

Context and analysis: #104, #105.

openwong2kim:
## Summary

Completes #105. After #104 made `getPage()` refuse the app shell, the browser
inspection tools with no fallback throw `No browser page available` in packaged
builds (where `connectOverCDP` doesn't surface the `<webview>` guest as a
Playwright Page). #105/#107 fixed the one-shot DOM tools by rerouting through
`browser.evaluate`. These three couldn't use that route:

- `browser_console`, `browser_network`, `browser_response_body` rely on
  Playwright's **event subscription** model (`page.on('console'|'request'|'response')`).
  A one-shot `Runtime.evaluate` can't replicate it — a page can't report its own
  console history, and `performance` entries lack method/status/body.

So this captures the data in the **main process** off the guest webContents' CDP
debugger (the same `wc.debugger` `WebviewCdpManager` attaches and
`browser.screenshot`/`browser.evaluate` already drive), and the tools drain it
over three new RPC methods when `getPage()` is null. The Playwright path is
unchanged.

## Changes

- **`BrowserCaptureManager`** (new, `src/main/browser-session/`): lazily enables
  `Runtime`/`Network` CDP domains on the guest webContents and buffers console
  messages + network requests/responses. Drain API: `getConsole` / `getNetwork`
  / `getResponseBody` (+ clear).
- **Three RPC methods** (`browser.console.get`, `browser.network.get`,
  `browser.responseBody.get`) in `browser.rpc.ts`, gated on the existing
  `browser.read` capability (`methodCapabilityMap` + `ALL_RPC_METHODS`).
- **`inspection.ts`**: the three tools fall back to those methods when no
  Playwright page is available; shared console/network formatters so both
  transports render identically.
- **`WebviewCdpManager`**: `setCaptureCleanup` hook so capture listeners are torn
  down on unregister (manual `detach()` does not remove EventEmitter listeners).

### Design notes (codex-reviewed, 3 rounds)

- **Lazy**: domains enabled + listener attached on the first capture call,
  matching the dev tools' "listener attaches on first call" semantics. Zero cost
  when unused; nothing captured before the first call.
- **Ordering**: listener attached BEFORE `Runtime.enable`/`Network.enable` so no
  event is missed. `ensure()` is singleflight (concurrent first calls don't
  double-attach).
- **Console**: CDP `RemoteObject[]` formatted to text; `warning` mapped to `warn`.
- **Network**: correlated by `requestId` (not URL); redirect hops recorded; ring
  eviction only drops the index entry when it still points at the evicted object
  (redirect hops share an id).
- **Body**: `Network.getResponseBody` base64-decoded, textual-only, truncated by
  UTF-8 **bytes** (256KB/body) with a 4MB total-body budget per webContents
  (evict oldest) so a chatty page can't pin ~256MB; `Network.enable` buffer sizes
  set accordingly; fetch failures leave a valid body-less entry.
- **Cleanup**: drop on webContents `destroyed`, on debugger `detach` (opening
  webview DevTools terminates the session), and on `unregister`.

### Fixed

- MCP browser tools `browser_console`, `browser_network`, and
  `browser_response_body` now work in packaged builds (they fall back to
  main-process CDP capture over RPC when no Playwright page is available) instead
  of erroring with "No browser page available".

## Stability tier impact

- [x] Three new `experimental` RPC methods (`browser.console.get`,
  `browser.network.get`, `browser.responseBody.get`), gated on the existing
  `browser.read` capability. Additive; classify experimental in
  `docs/api/inventory.md`.

Three new experimental RPC methods on the browser surface. No change to
PaneMetadata, EventBus, permission grammar (reuses `browser.read`), or Named Pipe
security.

## Test plan

- **Unit (new):** `BrowserCaptureManager.test.ts` (16) drives a fake debugger
  EventEmitter — console RemoteObject formatting + `warning`→`warn`, requestId
  correlation, redirect-hop recording, base64 decode, non-textual skip,
  `getResponseBody` rejection survival, byte-based truncation, ring cap, clear,
  drop/detach cleanup, singleflight. `browser.rpc.test.ts` (+5) covers the three
  handlers + the no-target error path.
- **Full suite:** `npx vitest run` -> 2231 passed (0 failed).
- **Typecheck/build:** `tsc -p tsconfig.json` + `tsc -p tsconfig.mcp.json` +
  `npm run build:mcp` all green.
- **Lint:** no new findings.
- **Packaged smoke test: PENDING.** Like #105, this only reproduces in a packaged
  build. The real check is opening a page that logs to console + makes XHRs, then
  `browser_console` / `browser_network` / `browser_response_body` in an
  `npm run make` bundle with a fresh daemon. The capture rides the same
  `wc.debugger` channel `browser.screenshot`/`browser.evaluate` already use in
  packaged builds, and unit tests cover the buffering/formatting logic — but
  please do not merge before a packaged run. @junbeom09 has the setup.

## Reviewer notes (limitations / trade-offs)

- **Lazy = no pre-first-call history.** Matches the dev tools; documented.
- **Body byte budget.** 256KB/body, 4MB total/webContents — a multibyte body may
  re-encode a couple bytes over the per-body cap (split char → U+FFFD), a
  negligible fuzz, not a correctness issue.
- **DevTools opens detach the debugger** (Electron single-client). Capture drops
  itself on `detach` and re-enables on the next call; this also surfaces the
  pre-existing fragility that screenshot/evaluate share that debugger.

Closes #106. Builds on #104 / #105 (#107). Browser automation is a de-prioritized
area; this is "restore a deployed helper tool", not new headline capability.


openwong2kim:
Merged (unreleased). @junbeom09, this closes the loop on the packaged-mode browser tooling you surfaced back in #104. #105/#107 fixed the one-shot DOM tools, and this one covers the three that needed real event capture: browser_console, browser_network, browser_response_body.

Since these only misbehave in a packaged build, the same caveat from #105 applies. The unit tests cover the buffering and formatting, but the actual CDP capture path only runs in a real bundle. If you have a few minutes on your npm run make setup, a smoke test would be the highest-value check before this goes into a release:

- open a page that logs to the console and fires a few fetch/XHR calls (your github.com run works, or any SPA)
- browser_console should return the page's console messages, not empty and not the app shell
- browser_network should list the requests with method + status
- browser_response_body with a pattern like *api* should return a real body

Capture rides the same wc.debugger channel that screenshot/evaluate already use in packaged builds, so I expect it to hold, but your side-by-side method (branch bundle next to the installed app) is exactly the verification this needs. No rush. Thanks again for the original writeup that mapped this whole thing out.

junbeom09:
**Follow-up: packaged-build smoke test, separating what I observed from what I'm inferring.**

I couldn't capture the MCP server's `[PlaywrightEngine]` stderr in this run, so I've split **observed facts** from **inferences read off the code**.

### How I verified

_(Done on a different PC than the #104 verification.)_ Built this branch (`npm run make`) and ran the packaged app from `out/wmux-win32-x64/` as the sole instance, then exercised the three tools against `https://jsonplaceholder.typicode.com`. In the page I emitted `console.log/info/warn/error` plus `fetch` (GET + POST) and an `XHR` (`browser_evaluate` with `allowDangerous:true` once — fetch/XHR are blocked by default), then called each tool.

### Observed (facts), target `https://jsonplaceholder.typicode.com`

| call | result |
| --- | --- |
| `browser_console` | all four levels + the fetch/XHR result lines my page logged — page console, **not** the app shell |
| `browser_network` | `GET /todos/7 200`, `GET /users/5 200` (XHR), `POST /posts 201` — method + status |
| `browser_response_body` | real server JSON per pattern (`*todos/7*`, `*users/5*`) |
| `browser_evaluate` | `location.href = https://jsonplaceholder.typicode.com/`, `isFileProtocol:false` — webview, not `file://` shell |
| `browser_cookies` | `[]`, no error |

Also a fact, via the code path: `browser_cookies` throws `No browser page available` only when `getPage()` is falsy (`state.ts:62-64`), so its clean `[]` return means **`getPage()` returned a live `Page` on this run**. The three inspection tools branch `if (page) {…Playwright…} else { sendRpc('browser.console.get' | 'browser.network.get' | 'browser.responseBody.get') }` (`inspection.ts:445-459`, `:495-509

### Confirmed in the code (facts)

* `browser_console` / `network` / `response_body` each do `getPage(surfaceId).catch(() => null)` and only fall back to `sendRpc('browser.*.get', …)` when that's null (`inspection.ts:445`, `:495`, `:550`).                                                                                      * `browser_cookies` (and the `wait`have no fallback —`getPage()`-or-throw shape (`state.ts:62-64`).                                                * `getPage()` returns null only aftlaywrightFailed` is set(`PlaywrightEngine.ts:277`, `:367`).                                                          
### Inference (read off the code; not directly observed — no runtime logs captured)           
* **The three calls took the Playwright path, not the RPC fallback.** `browser_cookies` proves `getPage()` was non-null this run, so the `if (page)` branch ran. _(Observed: `cookies → []`. Inferred: that console/network/respPlaywright path, not `sendRpc`.)_
* **Why `getPage()` succeeded in a packaged build:** this looks like the inverse of the #104 run, where `getPage()` was null and `browser_evaluate` took the RPC fallback. I'm inferring #104's
shell-detection / `targetId` changerface the `<webview>` as a `Page` in packaged builds, so `getPage()` succeeds and the new #108 fallback is a safety net rather than the live path. _(Inferred from the code; not confirmed via `[PlaywrightEngine]` logs.)_

### Note on the fallback

The fallback branch (`getPage()` → ole.get' / 'browser.network.get' /'browser.responseBody.get')`) isn't exercised by this run, but it's covered by unit tests for `inspection.ts` (both page-present and page-null), alongside the `browser.rpc` handlers and
`BrowserCaptureManager` core. Full

_Minor:_ network capture is lazy — requests are only collected after the first `browser_network`
call enables the domain (a warm-up ntended behaviour, and thePlaywright path is lazy the same way via `ensureNetworkListener`).

_Caveat:_ the functional ask — consurning real page data, not shell,not empty — is satisfied live. But since Playwright attached, this run verifies the Playwright path; the RPC fallback itself is verified by the unit tests, not this run. Exercising it live would need `getPage()` forced to null.

Apologies that the path attribution is inferred rather than confirmed via logs — I couldn't capture the runtime `[PlaywrightEngine]` st the inference section as "best read of the code." Thanks again for wmux 🙏

openwong2kim:
Following up on the gap you flagged. You confirmed the tools work in the packaged build, and you were right that on your run getPage() returned a live Page, so the Playwright path ran and this PR's RPC fallback (browser.*.get into BrowserCaptureManager) acted as a safety net rather than the live path, covered by unit tests but not exercised end to end. I drove that missing piece.

I ran the fallback's server side against a real Electron <webview> guest: registered it through the real WebviewCdpManager (debugger attach), enabled capture via BrowserCaptureManager, then made the exact drain calls browser.console.get / browser.network.get / browser.responseBody.get make. No Playwright or getPage in the path, so by construction it is the getPage()==null branch, and on the real packaged target type (the webview guest, not a plain window).

Everything came back off the guest: console at all levels with CDP "warning" mapped to "warn", network correlated by requestId with method and status, XHR/fetch response bodies, and listener teardown when the surface unregisters. One thing worth recording: the main navigation document body comes back as "No resource with given identifier found" from Network.getResponseBody. That is a known Chromium behavior for the top-level resource, not something on our side, and the code handles it gracefully (you still get the network entry, just no body). Subresource and XHR bodies are fine, and browser_extract_text is the right tool for page text anyway.

I also did a visible-window run against example.com to watch it live, and the capture pulled the page's console and network exactly as expected.

So between your live packaged run on the Playwright path and this on the fallback path, both branches are covered. Thanks again for the precise observed-vs-inferred writeup, it made the gap easy to target.

Issue #111, opened 2026-06-06
browser_cookies (and getPage-only tools) lack RPC fallback, error on packaged builds


openwong2kim:
## What happens

On a packaged build, `browser_cookies` errors out instead of returning cookies:

```
browser_cookies action="get"
```

```
No browser page available. Call browser_open with a URL first to establish a
CDP connection (required even if a browser panel is already visible).
```

A browser panel was open and active at the time (the same surface that every other
browser tool was reading from successfully in the same run). So the message is
misleading: a page exists, it's just that `getPage()` (the Playwright path) can't
hand back the embedded webview as a Page on packaged builds.

## Why this is the same class of bug as #105/#106

This is the exact failure mode the RPC fallback work fixed for the extraction,
snapshot, and capture tools. In this dogfood run, the extraction/console/network
tools all returned real page data while `browser_cookies` failed `getPage()` in the
same surface. That mismatch is the tell: the other tools fall back to RPC, cookies
doesn't, so cookies still hard-depends on the Playwright page and breaks in packaged
builds.

## What I expected

`browser_cookies action="get"` returns `[]` or the cookie list without an error,
the same way it works in dev. Cookies are available over CDP (`Network.getCookies` /
`Storage.getCookies`), so this should be reachable through the RPC path even when
`getPage()` can't produce a Playwright Page.

## Scope to check

`browser_cookies` is the one I hit, but anything that resolves through `getPage()`
without an RPC fallback likely has the same gap on packaged builds. Worth auditing
the sibling tools in one pass, e.g.:

- `browser_cookies`
- `browser_storage`
- any other tool that calls `getPage()` directly with no `browser.*` RPC fallback

## Environment

- Packaged wmux build
- Tool path: Playwright `getPage()` (fails here); extraction/capture tools in the
  same run used the RPC fallback and worked

openwong2kim:
Fixed in #112 (merged to main as 8c306c3). browser_cookies/storage/emulate/resize now fall back to the main-process CDP RPC channel when no Playwright Page is available in packaged builds. browser_wait is the one tool still missing a fallback — tracked separately in #114.

Issue #110, opened 2026-06-06
browser_extract_data fills every field with the same row metadata instead of mapping title/url


openwong2kim:
## What happens

`browser_extract_data` reads the live page fine, but every field in the result gets filled with the same value instead of being mapped to the requested field. Both `title` and `url` come back holding the row's metadata line.

Found this while dogfooding the browser tools on a packaged build (HN front page).

Call:

```
browser_extract_data
  goal="front page story list"
  fields={ "title": "string", "url": "string" }
```

Returned (trimmed):

```json
[
  {
    "title": "117 points by transistor-man 4 hours ago  | hide | 39 comments",
    "url": "117 points by transistor-man 4 hours ago  | hide | 39 comments"
  },
  {
    "title": "94 points by 0xkato 3 hours ago  | hide | 13 comments",
    "url": "94 points by 0xkato 3 hours ago  | hide | 13 comments"
  }
]
```

## What I expected

Each record mapped to the right field, e.g.

```json
{
  "title": "How LLMs work",
  "url": "https://www.0xkato.xyz/how-llms-actually-work/"
}
```

The story titles and hrefs are clearly present on the page. The same dogfood run pulled them correctly through `browser_snapshot`:

```
[ref=18] a "How LLMs work" -> https://www.0xkato.xyz/how-llms-actually-work/
```

So the data is reachable. The problem is in how `extract_data` splits rows into fields: it's grabbing one text blob (the subtext line) and stuffing it into every requested field, rather than resolving `title` to the link text and `url` to the href.

## Notes

- The source data is real (not the app shell, not an error), so this is unrelated to the #105/#106 packaged-build fallback work. Those tools pass.
- Two things look off: (1) field names aren't honored at all (title and url are identical), and (2) the value picked is the metadata subtext rather than the story link/title.
- Repro is page-agnostic, but a list page with separate anchor text + href per row (like HN) makes it obvious.

## Environment

- Packaged wmux build
- Tool path: RPC fallback (confirmed in the same run by `browser_cookies` failing `getPage()` while the extraction tools still returned real data)

openwong2kim:
Fixed in #112 (merged to main as 8c306c3). extract_data now maps each field independently — title/name to headings or class hints, url to the primary link href — and skips empty vote/icon anchors when picking the primary link.

PR #112, opened 2026-06-06
fix(mcp): browser_extract_data field mapping (#110) + packaged RPC fallback for state tools (#111)

openwong2kim:
Two browser-tool bugs found dogfooding on a packaged build. Both came out of the same session, so they ride together here.

Stacked on #109 (the first-party MCP enforce-mode fix). #111 edits `firstParty.ts`, which only exists on that branch, so the base is set to `fix/first-party-mcp-trust` and the diff stays scoped to these two fixes. GitHub will retarget this to `main` automatically once #109 merges. #110 is independent of #109.

## #110 — browser_extract_data filled every field with the same value

On a real page the tool read the DOM fine but put the row's subtext line into both `title` and `url` instead of the story title and href. Three separate gaps:

- A blank or one-char table header matched every field name (`''.includes(x)` is always true), collapsing all fields onto the first column. Now requires header length >= 2 before partial-matching, so layout tables with empty header cells fall through to the repeated-element strategy.
- `tr` wasn't in the repeated-element scan, so table-layout lists (Hacker News `<tr class="athing">` rows) were never recognized as repeated items. Added `tr`.
- The repeated-element path grabbed the first `a[href]` in a row, which on link lists is an href-less vote/icon anchor. Now resolves one primary anchor per row (first link with visible text, skipping `javascript:`/`#`/empty hrefs) and uses it for `url` (href) and as a `title` fallback (text). url-like fields read `href` over text in the table path too.

Regression coverage runs the in-page callbacks against a jsdom document (jsdom added as a devDependency).

## #111 — cookies/storage/emulate/resize errored on packaged builds

`browser_cookies`, `browser_storage`, `browser_emulate`, and `browser_resize` all resolved through `engine.getPage()`, which returns null for the embedded `<webview>` on packaged builds. Each threw `No browser page available` even with a browser panel open and active in the same surface every other browser tool read from. Same failure class the RPC fallback already fixed for the extraction/snapshot/capture tools (#105/#106).

Each tool now tries the Playwright Page first and falls back to the main-process CDP channel when there isn't one:

- `browser_storage` is pure `page.evaluate`, so it reuses the existing `evalFunctionOrRpc` transport with no new RPC method.
- `browser_cookies` gets a new `browser.cookies` RPC over CDP `Network.getCookies` / `getAllCookies` / `setCookies` / `clearBrowserCookies`. Sensitive-domain redaction stays in the tool, so it applies identically on both paths.
- `browser_resize` gets a new `browser.resize` RPC over `Emulation.setDeviceMetricsOverride`.
- `browser_emulate` gets a new `browser.emulate` RPC applying offline / headers / geolocation / media / timezone / locale / device over CDP. Device presets are resolved to viewport + UA in the tool so playwright-core's device table stays out of the main process. HTTP credentials stay unsupported over CDP and report the same guidance as the Playwright path.

The three new methods are gated under the existing `browser.evaluate` capability (a caller that can run arbitrary JS can already read/write cookies, storage, and emulate the page, so no new capability is invented) and added to the first-party allowlist so the bundled server can call them under enforce mode.

Scope note: `file.ts` (upload/download) and the PDF tool also resolve through `getPage()`, but they need a file-transfer fallback rather than a CDP command, so they're left for a separate follow-up.

## Testing

- New `state.fallback.test.ts`: each tool's RPC path, sensitive-domain redaction over RPC, device-preset resolution, and the unchanged Playwright path.
- `markdown-extractor.dom.test.ts`: HN + card fixtures that previously duplicated metadata.
- Full suite 2267 passing, `tsc --noEmit` clean.

Still pending: packaged smoke on a real Electron webview (the CDP command paths can't be exercised in unit tests). Tagging for a packaged check the same way #104/#107/#108 were verified.


openwong2kim:
@codex review

openwong2kim:
@codex review

PR #115, opened 2026-06-06
fix(mcp): CDP fallback for browser_wait in packaged builds (#114)

### LINKED ISSUE #114: browser_wait fails in packaged builds (no CDP fallback)
@
`browser_wait` is the one browser tool that still hits "No browser page available" in packaged builds. The geo verification on #112 surfaced it — `browser_open`/`navigate`/`emulate`/`evaluate`/`cookies` all worked, but `browser_wait` errored out and I had to poll with `browser_evaluate` instead.

## Cause

`browser_wait` calls into the Playwright `Page` API directly (`page.waitForURL` / `waitForSelector` / `waitForFunction` / network idle). In packaged builds there is no Playwright `Page` — only the CDP fallback (main-process `webContents` debugger). The other tools route through that fallback; `browser_wait` never got one, so it just throws.

## Repro

In a packaged build:
1. `browser_open` + `browser_navigate` to any https page
2. `browser_wait { fn: "..." }` (or selector/text/url)
3. -> `No browser page available. Call browser_open with a URL first...`

Dev build works because the real Playwright Page is present.

## Fix direction

Add a CDP fallback path for `browser_wait`, same as the other tools:
- `url` / `selector` / `text` / `fn` -> `Runtime.evaluate` polling loop (can reuse the existing fallback evaluate, low effort)
- `networkidle` -> track `Network` domain events (a bit more work)

Out of scope for #112 — filing as a follow-up.
@

openwong2kim:
Closes #114.

`browser_wait` was the last browser tool with no packaged fallback. It only used the Playwright Page API (`waitForURL` / `waitForSelector` / `waitForFunction` / `waitForLoadState`), so on packaged builds — where `engine.getPage()` returns null because playwright-core can't surface the guest `<webview>` as a Page — it threw `No browser page available` instead of waiting. Surfaced during the #112 geo verification, where every other browser tool worked but `browser_wait` errored out.

## Change

Same shape as the #105/#106/#111 fallbacks: try the Playwright Page first, and when none is available poll the condition over the main-process CDP channel (`browser.evaluate`) until it holds or the timeout elapses.

| condition | fallback |
|-----------|----------|
| `url` | match `location.href` against the glob (anchored regex; `*` and `**` both match any run of characters) |
| `selector` | `!!document.querySelector(sel)` |
| `text` | `document.body.innerText.includes(text)`, guarded for a null body |
| `fn` | evaluate the predicate expression (Playwright `waitForFunction(string)` semantics); the dangerous-pattern warning is preserved |
| `networkidle` | approximated by `document.readyState === "complete"` — the page-target debugger has no network-idle signal over this transport, and the completion message says so |

Transient evaluate errors mid-navigation are swallowed so polling continues; a deadline miss throws a `Timeout` error that flows into the existing timeout messaging. The Playwright (dev) path is unchanged.

## Tests

`wait.fallback.test.ts` — 8 cases (selector/text/fn/url/networkidle success, timeout messaging, transient-error retry, and the Playwright-Page-present no-RPC path). Full suite 2263 green, mcp tsc clean.

## Not covered

Packaged smoke (real CDP round-trip) is not in this PR — the unit tests mock the RPC. Worth a quick dogfood in a packaged build before release, the same way the geo fix was verified.

### DIFF
--- a/src/mcp/playwright/__tests__/wait.fallback.test.ts
+++ b/src/mcp/playwright/__tests__/wait.fallback.test.ts
@@ -105,6 +105,38 @@ describe('browser_wait RPC fallback', () => {
     expect(res.content[0].text).toContain('Timed out');
   });
 
+  it('matches zero path segments for a double-star URL glob (**/ is optional subpath)', async () => {
+    // `**/settings` must also match the zero-segment case `/settings`, like waitForURL.
+    mockSendRpc.mockImplementation(evalRouter({
+      'location.href': 'https://site.test/settings',
+      'document.readyState': 'complete',
+    }));
+    const res = await wait({ url: 'https://site.test/**/settings', timeout: 1000 });
+    expect(res.isError).toBeUndefined();
+    expect(res.content[0].text).toContain('URL matched');
+  });
+
+  it('calls a function-expression predicate instead of testing the function object', async () => {
+    mockSendRpc.mockResolvedValue({ value: true });
+    const res = await wait({ fn: '() => window.ready' });
+    expect(res.content[0].text).toContain('custom predicate satisfied');
+    const expr = (mockSendRpc.mock.calls[0][1] as { expression: string }).expression;
+    // The arrow function is invoked, not coerced to a (truthy) function object.
+    expect(expr).toBe('!!((() => window.ready)())');
+  });
+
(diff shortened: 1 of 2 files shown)


openwong2kim:
@codex review

openwong2kim:
@codex review

Issue #113, opened 2026-06-06
Verify first-party MCP identity beyond the self-asserted clientName


openwong2kim:
## Background

The Phase 2.2 permission enforcer recognizes the bundled wmux MCP server by the host `clientName` it reports (`claude-code`) and grants it a scoped allowlist of methods (`firstParty.ts`). This unblocked the packaged enforce-mode lockout (#109).

Codex flagged on #109 that the recognition is spoofable: any other MCP server Claude Code launches also presents `clientName=claude-code`, so a third-party server that reaches the daemon pipe with the auth token would be treated as first-party and reach the whole `FIRST_PARTY_METHODS` set without declaration or approval.

## Current stance

This is a documented, deliberate trade-off (see the `firstParty.ts` header and the plugin spec's "declared, not verified" model). On a single-user OS, a same-user process that wants to impersonate the host already holds the daemon auth token and could call the pipe directly. The scoped allowlist bounds the blast radius: even an impersonator only reaches the curated method set, never `daemon.*`, `workspace.new`, or company mutation (pinned by source-invariant tests).

## What real verification would take

Telling the bundled server apart from another Claude-Code-launched MCP server needs a verified identity, not a self-asserted name. Options to evaluate:

- Named-pipe peer PID (`GetNamedPipeClientProcessId` on Windows; `SO_PEERCRED` / `LOCAL_PEERPID` on Unix) then attest the peer's executable path against the known bundled MCP server binary.
- A per-launch nonce the daemon injects into the bundled server's environment and requires on the envelope. This only works if the daemon owns the spawn, which it currently does not (Claude Code launches the server).

Both carry cross-platform and wire-format implications, so this is its own design and PR rather than a fix on #109/#112.

## Acceptance

- The first-party bypass applies only to a server whose identity is verified, not merely self-asserted.
- Works cross-platform (win32 + darwin + linux), or is explicitly gated per-platform with a documented fallback.
- Existing scoped-allowlist tests still pass; new tests cover a spoofed `clientName` being rejected.

openwong2kim:
Took a proper design pass at this one. Writing up where it landed since the conclusion is a little counterintuitive.

Short version: deferring the verified-identity work and documenting the limitation plus hardening the allowlist tests instead. Not because it's hard (it is), but because verifying first-party identity on its own doesn't move real risk on a single-user box.

[...]

PR #116, opened 2026-06-06
fix(mcp): authenticate first-party allowlist

openwong2kim:
### Motivation
- The first-party allowlist previously relied only on the caller-supplied `clientName` (e.g. `claude-code`), which a malicious authenticated local client could spoof to bypass normal permission checks. 
- The goal is to ensure the bundled MCP bridge can still function without reworking the declaration/approval UX while preventing a spoofed `clientName` from gaining sensitive capabilities.

### Description
- Introduce a separate, on-disk first-party bearer credential and path (`getFirstPartyTokenPath`) and persist a per-install token alongside the normal RPC auth token via `McpRegistrar`.
- Wire the registrar token into the router with `RpcRouter.setFirstPartyToken(...)` and stamp `RpcContext.firstPartyAuthenticated` when an incoming request carries the verified `firstPartyToken` envelope field.
- Require `ctx.firstPartyAuthenticated === true` in `PermissionEnforcer.check` before applying the scoped `FIRST_PARTY_METHODS` bypass, while preserving fail-closed behavior for denied records and lookup failures.
- Make the bundled MCP client (`src/mcp/wmux-client.ts`) read and attach the first-party token to outbound RPC envelopes and update tests to assert that spoofed `clientName` is rejected unless the private token is present.

### Testing
- Ran unit tests: `npx vitest run src/main/mcp/__tests__/PermissionEnforcer.firstParty.test.ts src/main/pipe/__tests__/RpcRouter.firstParty.enforce.test.ts src/main/mcp/__tests__/firstParty.test.ts` and all modified tests passed (24 tests across the changed suites passed).
- Exercised the high-level router dispatch tests with `npx vitest run src/main/pipe/__tests__/RpcRouter.firstParty.enforce.test.ts` and they passed.
- Type-checked the repo with `npx tsc --noEmit -p tsconfig.json` and built the MCP bundle via `npm run build:mcp` with no errors.
- Ran linting (`npm run lint`) which reports pre-existing repository-wide lint issues unrelated to this change (lint failures are not new regressions introduced by this PR).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a24182fa6888320a1bbcc00f901ff74)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -193,6 +193,7 @@ const autoUpdater = new AutoUpdater(() => mainWindow);
 const rpcRouter = new RpcRouter();
 const pipeServer = new PipeServer(rpcRouter);
 const mcpRegistrar = new McpRegistrar();
+rpcRouter.setFirstPartyToken(mcpRegistrar.getFirstPartyToken());
 const webviewCdpManager = new WebviewCdpManager(cdpPort);
 
 const claudeWorker = new ClaudeWorker(() => mainWindow);
--- a/src/main/mcp/McpRegistrar.ts
+++ b/src/main/mcp/McpRegistrar.ts
@@ -1,7 +1,8 @@
 import * as fs from 'fs';
 import * as path from 'path';
+import * as crypto from 'crypto';
 import { app } from 'electron';
-import { getAuthTokenPath, getPipeName } from '../../shared/constants';
+import { getAuthTokenPath, getFirstPartyTokenPath, getPipeName } from '../../shared/constants';
 import { secureWriteTokenFile } from '../../shared/security';
 import { isMac } from '../../shared/platform';
 import { formatMacosError, MACOS_ERRORS } from '../../shared/errors/macos';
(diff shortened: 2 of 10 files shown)


openwong2kim:
Closing this one.

The extra `~/.wmux-first-party-token` sits in the same `$HOME` dir under the same owner-only ACL as `~/.wmux-auth-token`, and a client already has to present the auth token to reach the router at all. So any process that could spoof `clientName` can read both files just as easily and attach both, which means it doesn't add a trust boundary against the same-user attacker described in #113.

This is also the approach we already weighed and rejected in `plans/first-party-mcp-trust.md` section 0 when #109 landed ("token-in-args buys ~nothing... a same-user process that can read it can already read the daemon auth token", and a second bearer secret in a worse location). The proper fix for #113 is named-pipe peer-PID attestation (GetNamedPipeClientProcessId) against the bundled-server binary path, not a co-located token. The new envelope compare is also not constant-time, unlike the auth-token path.

Leaving #113 open to track the real fix.

PR #117, opened 2026-06-06
fix(daemon): prevent reconnect ping token disclosure

openwong2kim:
### Motivation
- A Windows reconnect path treated any successful connect to the canonical named pipe as a trusted daemon and had the launcher send the plaintext daemon auth token to that pipe, allowing a local pipe squatter to steal the token. 
- The goal is to prevent disclosure of the daemon token during liveness/reconnect probes while preserving the split‑brain avoidance and fallback behavior.

### Description
- Add a `daemon.authChallenge` RPC that returns a per-connection nonce so peers can prove possession of the token via an HMAC instead of receiving the token in cleartext (`src/daemon/DaemonPipeServer.ts`, `src/shared/rpc.ts`).
- Implement HMAC signing/verification on the daemon (`signAuthProof`, `verifyRequestAuth`) and emit an `authProof` on RPC responses when the request was authenticated via a challenge (preserves legacy direct-token callers). 
- Update the launcher `daemonPing` to perform the challenge/HMAC flow (request challenge → send HMACed `daemon.ping` → verify server `authProof`) instead of sending the token directly (`src/main/daemon/launcher.ts`).
- Register `daemon.authChallenge` in the RPC metadata and capability map so it is recognized by the permission subsystem (`src/main/mcp/methodCapabilityMap.ts`).
- Add a regression test that exercises the challenge+proof flow and asserts the plaintext token is never written to the socket (`src/daemon/__tests__/DaemonPipeServer.test.ts`).

### Testing
- Ran the focused unit tests covering the pipe server and launcher ping logic with `npx vitest run src/daemon/__tests__/DaemonPipeServer.test.ts src/daemon/__tests__/DaemonPipeServer.reclaim.test.ts src/main/daemon/__tests__/launcher.liveness.test.ts src/main/daemon/__tests__/launcher.reping.test.ts --reporter=verbose` and the new test passed.
- Type-checked the project with `npx tsc --noEmit` and the full test suite with `npm test`, both of which succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a225912bb5c832085a714194bc3e34e)

### DIFF
--- a/src/daemon/DaemonPipeServer.ts
+++ b/src/daemon/DaemonPipeServer.ts
@@ -7,6 +7,11 @@ import type { RpcRequest, RpcResponse } from '../shared/rpc';
 import { secureWriteTokenFile, reHardenTokenFileAcl } from '../shared/security';
 
 const MAX_LINE_BUFFER = 1024 * 1024; // 1 MB — prevent OOM from malicious clients
+const AUTH_CHALLENGE_METHOD = 'daemon.authChallenge';
+
+function signAuthProof(token: string, payload: string): string {
+  return crypto.createHmac('sha256', token).update(payload).digest('hex');
+}
 
 type RpcHandler = (params: Record<string, unknown>) => Promise<unknown>;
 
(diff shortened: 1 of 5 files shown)


openwong2kim:
Closing this and consolidating the daemon-token-disclosure work.

The HMAC challenge is keyed by the same shared auth token, so it doesn't actually authenticate the server: an attacker without the token can't forge a proof but also can't usefully squat, and one with the token forges any proof trivially. The token also still flows in cleartext on every session-pipe connect and RPC immediately after the ping, so hardening just the probe ping doesn't close the scenario. Per PROTOCOL.md section 5.3 the model is "OS isolates users, token isolates processes", and a same-user squatter can already read the token straight off disk.

#138 targets the same vulnerability with overlapping changes to the same files, so I'm closing both. If we pursue pipe-squat hardening later it needs a real server-auth primitive (verify the pipe peer's SID/PID), which neither PR does.

PR #124, opened 2026-06-06
fix(security): rebuild Windows token ACL

openwong2kim:
### Motivation

- The Windows startup hardening previously used `icacls /grant:r` which does not remove explicit ACEs for other principals nor rebuild child ACLs, leaving existing files (token, scrollback) potentially readable by other local users and breaking the documented guarantee.
- The change preserves the existing SID-first identity logic and ASCII-only username fallback to avoid non-ASCII principal mangling while fixing the enforcement gap.

### Description

- Replace the `icacls`-based flow with a PowerShell DACL rebuild: a `RESTRICTIVE_WINDOWS_ACL_SCRIPT` disables inheritance, removes every explicit ACE, and adds a single owner `FullControl` ACE, invoked via `powershell.exe -EncodedCommand` with a JSON payload. (`src/shared/security.ts`)
- Keep `secureWriteTokenFile` and `reHardenTokenFileAcl` routed through the repaired helper so both fresh-writes and re-hardening use the DACL-rebuild path. (`src/shared/security.ts`)
- Update unit tests to assert the PowerShell call, its JSON payload, and that the rebuilt script disables inheritance and removes explicit ACEs. (`src/shared/__tests__/security.test.ts`)
- Update documentation to reflect that on Windows the token-file DACL is rebuilt to remove inherited and pre-existing explicit grants before granting the owner `FullControl`. (`docs/SECURITY.md`, `docs/PROTOCOL.md`)

### Testing

- Ran `npx vitest run src/shared/__tests__/security.test.ts` and the updated security tests passed. 
- Ran `npm run build:daemon` which completed successfully and produced the daemon bundle. 
- Ran `npm run test:parallel` (project test suite) which completed successfully (existing test run shown as passing). 
- Ran `npx eslint --ext .ts src/shared/security.ts src/shared/__tests__/security.test.ts` and linting of the touched files completed without errors; running the repository-wide `npm run lint` still fails due to pre-existing unrelated lint issues outside the modified files.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a2264802518832080bd7c31ab860a71)


openwong2kim:
Closing this and reworking it.

The diagnosis is right: `icacls /grant:r *<SID>:F /inheritance:r` only replaces the named principal's ACE and only strips inherited ACEs, so a pre-existing explicit Everyone/Users ACE survives and the token can stay world-readable. But the fix as written doesn't work. The `Set-Acl` cmdlet path fails with SeSecurityPrivilege on the upgrade case (reproduced 10/10 on a real box): because `Set-Acl` also persists the Owner+Group sections (group is ...-513), a non-elevated process can't write it, so on every existing install reHardenTokenFileAcl would fail at startup and the re-harden guarantee silently stops working. That's why the Windows CI checks here are red.

The correct primitive is a DACL-only write that never touches owner/group: either icacls with an explicit remove-then-grant, or .NET FileInfo.SetAccessControl with a freshly-built FileSecurity (not Set-Acl -AclObject, which was the 0/10-success path in testing). I'll redo it on a fresh branch with a dynamic test that exercises the fresh / already-icacls'd / explicit-Everyone states, and the SECURITY.md section 1.1 real-profile dogfood before merge.

Closing in favor of that rework.

PR #127, opened 2026-06-06
fix(terminal): require explicit right-click paste

openwong2kim:
### Motivation
- Prevent an implicit right-click paste path that read the system clipboard and wrote it to the PTY when the right-click target was not a real DOM anchor, which created a pastejacking / clipboard-leak risk for xterm/WebGL-rendered terminal output.
- Preserve the user experience for selection-copy and explicit context-menu actions while eliminating an unsafe automatic paste behavior.

### Description
- Removed the automatic right-click paste flow from the terminal `contextmenu` handler in `src/renderer/hooks/useTerminal.ts` so a no-selection/no-DOM-link right-click now opens the explicit context menu instead of reading `window.clipboardAPI` and writing to the PTY.
- Kept right-click-on-selection behavior: selection still copies (keeps selection) via `copySelectionWithFeedback` and does not trigger an immediate paste.
- Left DOM-link handling intact but documented that link detection is only a menu enhancement and missed detections must not fall through to paste; updated the comment in `src/renderer/components/Terminal/ContextMenu.tsx` to reflect that `Paste` is an explicit menu action.
- Removed the copy→paste suppression timer / immediate-paste code and unified behavior to require explicit paste via the context menu or keyboard shortcut.

### Testing
- Ran `npx tsc --noEmit` which completed successfully.
- Ran the unit/test suite with `npm run test:parallel` which completed successfully (reported `171` test files and `2157` tests passed).
- Ran `npm run lint` which failed due to pre-existing repository lint errors unrelated to this change.
- Ran `npx eslint src/renderer/hooks/useTerminal.ts src/renderer/components/Terminal/ContextMenu.tsx` which surfaced pre-existing lint issues (errors/warnings) in `useTerminal.ts` outside the modified control-flow; these are unrelated to the security fix itself.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a226e5841b8832082bf265541263251)

### DIFF
--- a/src/renderer/components/Terminal/ContextMenu.tsx
+++ b/src/renderer/components/Terminal/ContextMenu.tsx
@@ -73,8 +73,9 @@ export default function ContextMenu({ x, y, hasSelection, selectedText, linkUrl,
         />
       )}
 
-      {/* Paste is handled inline by right-click when no link/selection;
-          only surface it here if no link is present (legacy callers) */}
+      {/* Paste is always an explicit menu action. Do not paste directly from a
+          right-click event: xterm-rendered links are not reliably DOM anchors,
+          so missed link detection must not fall through to clipboard input. */}
       {!linkUrl && (
         <MenuItem
           label={t('contextMenu.paste')}
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -136,14 +136,6 @@ export function copySelectionWithFeedback(
   });
 }
 
-// How long after a right-click copy we suppress a right-click paste. A second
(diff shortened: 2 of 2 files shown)


openwong2kim:
Closing this. There's no actual vulnerability here: the right-click path only fires on a deliberate right-click (no middle-click/auxclick/auto-paste trigger exists), and pasted content is already neutralized by bracketed-paste + ESC sanitization from v2.16.0, so a poisoned clipboard can't auto-execute. The change is a reasonable UX tweak (explicit menu paste) but not a security fix, and as written it also drops the one-click image right-paste affordance. Not carrying it as a security PR; if we want the UX change we can do it deliberately later.

PR #128, opened 2026-06-06
fix(security): rotate session pipe tokens

openwong2kim:
### Motivation
- Token rotation previously only updated the daemon control-pipe token and dropped control sockets, leaving per-session `SessionPipe` instances authenticated with a stale token and allowing attackers holding an old token to continue accessing active sessions.
- The change ensures daemon token rotation immediately invalidates session pipes and closes any authenticated session clients.

### Description
- Add a token-rotation subscription mechanism on `DaemonPipeServer` with `onTokenRotated()` and invoke subscribers from `rotateToken()` after the new token is persisted to disk.
- Change `SessionPipe` to accept an auth-token provider (`() => string`) instead of a creation-time snapshot and validate incoming clients against `authTokenProvider()` so it observes current daemon token state.
- Add `SessionPipe.handleAuthTokenRotated()` to disconnect any authenticated client and reset local state when the daemon token rotates.
- Wire the daemon (`src/daemon/index.ts`) to construct session pipes with a token provider and register a `pipeServer.onTokenRotated()` handler that calls `handleAuthTokenRotated()` for every active session pipe.
- Add a regression test in `src/daemon/__tests__/DaemonPipeServer.test.ts` proving that a pre-existing session pipe accepts the old token, is dropped on rotation, then rejects the old token and accepts the newly issued token.

### Testing
- Ran `npx vitest run src/daemon/__tests__/DaemonPipeServer.test.ts --reporter verbose` and all tests in that file passed (`18 passed`).
- Ran `npx tsc -p tsconfig.daemon.json --noEmit` to validate types and it succeeded.
- Ran `git diff --check` which reported no problems for the modified files.
- Ran targeted `eslint` checks; `eslint` reported existing repository baseline issues in other files unrelated to this patch (these are pre-existing and not introduced by this change).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a226e6ae7d083208fb568b722af2aec)

### DIFF
--- a/src/daemon/DaemonPipeServer.ts
+++ b/src/daemon/DaemonPipeServer.ts
@@ -9,6 +9,7 @@ import { secureWriteTokenFile, reHardenTokenFileAcl } from '../shared/security';
 const MAX_LINE_BUFFER = 1024 * 1024; // 1 MB — prevent OOM from malicious clients
 
 type RpcHandler = (params: Record<string, unknown>) => Promise<unknown>;
+type TokenRotationHandler = (newToken: string) => void;
 
 /** Action a reclaim probe implies, distinguishing a live owner from a zombie. */
 export type ReclaimOutcome = 'live-owner' | 'reclaimed' | 'unreclaimable';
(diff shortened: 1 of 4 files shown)


openwong2kim:
Closing this. The fix is technically correct (session pipes did authenticate against a stale token snapshot), but `rotateToken()` has zero production callers, no RPC/IPC/tray path triggers it, and tokens are reused across boots per SECURITY.md. So it hardens a dormant capability against a leak vector the owner-SID ACL on the token file already covers. If we ever wire up rotation we'll also need a client-side token-refresh path (which this doesn't add), so it would be revisited then anyway.

PR #134, opened 2026-06-06
Limit MCP registration lifetime and use absolute executable for MCP entries

openwong2kim:
### Motivation
- A prior change made `McpRegistrar.unregister()` a no-op and wrote user-level MCP entries using the unqualified `node` command, which causes durable autostart and opens an executable-search-order attack surface. 
- The change restores session-scoped cleanup and prevents PATH-based hijacking by writing an absolute executable path for MCP entries.

### Description
- Restore quit-time cleanup: `McpRegistrar.unregister()` now removes the keys owned by the current wmux process from `~/.claude.json` and clears owned state instead of being a no-op. (`src/main/mcp/McpRegistrar.ts`) 
- Use absolute executable: MCP registration now writes `command: process.execPath` (and an `env` payload when running the packaged Electron executable) instead of `command: 'node'`, preventing later PATH/current-directory resolution attacks. (`src/main/mcp/McpRegistrar.ts`, `src/cli/commands/mcp.ts`) 
- Add a helper `createMcpEntry()` to build the safe MCP entry shape and reuse it for both `wmux` and `wmux-a2a` registrations. (`src/main/mcp/McpRegistrar.ts`) 
- Update the CLI `wmux mcp register` to use the same absolute executable behavior. (`src/cli/commands/mcp.ts`) 
- Extend and adjust unit tests to verify absolute-command registration, that `unregister()` removes keys written by the current process on quit, and that foreign entries remain untouched. (`src/main/mcp/__tests__/McpRegistrar.test.ts`)

### Testing
- Type-check: ran `npx tsc --noEmit -p tsconfig.json` and it succeeded. 
- Unit tests: ran `npx vitest run src/main/mcp/__tests__/McpRegistrar.test.ts --reporter=verbose` and all tests passed (`11 tests, 11 passed`). 
- Lint (targeted): ran `npx eslint src/main/mcp/McpRegistrar.ts src/main/mcp/__tests__/McpRegistrar.test.ts src/cli/commands/mcp.ts` and the targeted files were clean after the change. 
- Build: ran `npm run build:cli` and the CLI bundle built successfully. 
- Note: a full repository lint `npm run lint -- --max-warnings=0` was attempted but fails on many pre-existing, unrelated lint issues; this PR only changes MCP-related files and includes focused linting for them.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a2280ed539083209ef21daa20ebe8ef)

### DIFF
--- a/src/cli/commands/mcp.ts
+++ b/src/cli/commands/mcp.ts
@@ -231,11 +231,11 @@ function registerEntries(): RegisterOutcome {
 
   const wrote: string[] = [];
   if (wmuxScript) {
-    servers['wmux'] = { command: 'node', args: [wmuxScript] };
+    servers['wmux'] = { command: process.execPath, args: [wmuxScript] };
     wrote.push('wmux');
   }
   if (a2aScript) {
-    servers['wmux-a2a'] = { command: 'node', args: [a2aScript] };
+    servers['wmux-a2a'] = { command: process.execPath, args: [a2aScript] };
     wrote.push('wmux-a2a');
   }
   config.mcpServers = servers;
--- a/src/main/mcp/McpRegistrar.ts
+++ b/src/main/mcp/McpRegistrar.ts
@@ -1,7 +1,7 @@
 import * as fs from 'fs';
 import * as path from 'path';
 import { app } from 'electron';
-import { getAuthTokenPath, getPipeName } from '../../shared/constants';
+import { getAuthTokenPath } from '../../shared/constants';
 import { secureWriteTokenFile } from '../../shared/security';
 import { isMac } from '../../shared/platform';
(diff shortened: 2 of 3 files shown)


openwong2kim:
Closing this. It bundles two unrelated changes: the absolute-executable part is a small, valid PATH-hijack hardening, but the unregister-on-quit part fires on the default detach quit and re-introduces the MCP-discovery chicken-and-egg bug (Claude Code in a live pane can no longer find the wmux server). It can't merge as-is, and the lifetime-limiting half isn't security. If we want the absolute-exec hardening on its own it's a clean ~3-line change we can do directly without the coupling.

PR #136, opened 2026-06-06
fix: gate CDP port disclosure

openwong2kim:
### Motivation

- Prevent unauthorised disclosure of the randomized Chromium DevTools (CDP) port via the `browser.cdp.info` RPC while preserving Playwright's ability to connect for MCP automation.

### Description

- Add a separate per-process CDP access token and on-disk path (`getCdpAccessTokenPath`) and persist it securely from the main process so ordinary RPC clients cannot use the normal auth token to discover the raw CDP port.  
- Change the `browser.cdp.info` handler to omit `cdpPort` by default and only include it when the caller supplies the internal CDP access token that matches the one passed to `registerBrowserRpc`.  
- Teach the bundled MCP-side client to read the CDP access token (`readCdpAccessToken`) and update `PlaywrightEngine` to request `browser.cdp.info` with the token only when it needs the raw port, while continuing to use non-port metadata for normal discovery.  
- Update unit tests (`browser.rpc.test.ts`) to assert public responses omit `cdpPort` and that token-bearing internal requests still receive it.

### Testing

- Ran the test suite with `npm test -- --runInBand` and all tests passed (`2172` tests total).  
- Built MCP and daemon with `npm run build:mcp && npm run build:daemon` and both builds succeeded.  
- Type-check verified with `npx tsc --noEmit` (no errors).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a2289e63024832095eb096458f98303)

### DIFF
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -23,7 +23,7 @@ import { registerMetaRpc } from './pipe/handlers/meta.rpc';
 import { registerSystemRpc } from './pipe/handlers/system.rpc';
 import { registerHooksRpc } from './pipe/handlers/hooks.rpc';
 import { UsagePoller } from './claude/UsagePoller';
-import { IPC } from '../shared/constants';
+import { getCdpAccessTokenPath, IPC } from '../shared/constants';
 import { HookSignalRouter } from './hooks/HookSignalRouter';
 import { SignalLatencyMeter } from './hooks/SignalLatencyMeter';
 import { registerBrowserRpc } from './pipe/handlers/browser.rpc';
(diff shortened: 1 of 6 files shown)


openwong2kim:
Closing in favor of #139, which covers this more fundamentally.

Gating `browser.cdp.info`'s port disclosure helps only at the margin: the CDP port is loopback-only already, the caller has to hold the auth token to call the RPC at all, the new access token shares the same owner-only ACL (so it adds no isolation a same-user attacker can't defeat), and the `/json` endpoint is discoverable by scanning the small 18800-18899 range regardless.

#139 removes the listener in packaged builds and binds debug to loopback explicitly, which subsumes this. The real surface reduction is tracked there.

PR #138, opened 2026-06-06
fix: avoid daemon token disclosure during discovery

openwong2kim:
### Motivation
- Prevent the launcher from disclosing the daemon bearer token to an arbitrary/local endpoint during startup discovery and reuse probes. 
- Provide a way for the launcher to verify the daemon identity without sending the token to a potentially attacker-controlled pipe.

### Description
- Add an unauthenticated `daemon.identify` RPC handler in `DaemonPipeServer` that returns an HMAC proof (over a client nonce) instead of echoing the auth token. 
- Change launcher probing in `src/main/daemon/launcher.ts` to send `daemon.identify` with a fresh `clientNonce`, verify the response `id`, `status`, `algorithm`, and HMAC proof with `crypto.timingSafeEqual`, and only accept endpoints that present a valid proof. 
- Stop using the predictable default pipe as a fallback when there is no trusted `daemon-pipe` file, and refuse the `EDAEMON_ALREADY_RUNNING` reconnect path when a trusted pipe name is not available to avoid token-bearing reconnects. 
- Wire `daemon.identify` into RPC metadata (`src/shared/rpc.ts` and `methodCapabilityMap`) and add a unit test proving the identity endpoint returns a proof without exposing the token (`src/daemon/__tests__/DaemonPipeServer.test.ts`).

### Testing
- Ran `npm run build:daemon` and the daemon bundle built successfully. (passed)
- Ran targeted unit tests `npx vitest run src/daemon/__tests__/DaemonPipeServer.test.ts src/main/daemon/__tests__/launcher.reping.test.ts src/main/mcp/__tests__/methodCapabilityMap.test.ts` and all tests passed. (passed)
- Ran `npx vitest run src/main/mcp/__tests__/phase-2-2-dynamic.test.ts` and it passed. (passed)
- Ran the full parallel suite `npm run test:parallel` and the full test suite passed (`171` test files, `2172` tests). (passed)

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a228e08d7608320a10e844d303e7fda)

### DIFF
--- a/src/daemon/DaemonPipeServer.ts
+++ b/src/daemon/DaemonPipeServer.ts
@@ -422,6 +422,11 @@ export class DaemonPipeServer {
       return;
     }
 
+    if (request?.method === 'daemon.identify') {
+      this.handleIdentify(socket, request);
+      return;
+    }
+
     // Authenticate before rate limit check (prevents DoS via rate exhaustion)
     // Use timing-safe comparison to prevent timing attacks
     const tokenBuf = Buffer.from(request.token || '');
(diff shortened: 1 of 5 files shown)


openwong2kim:
Closing as a duplicate of #117 (same daemon-token-disclosure concern, same files, overlapping HMAC approach).

The underlying issue only bites a same-user pipe squatter, who can already read the token off disk given the owner-SID ACL on the token file (PROTOCOL.md section 5.3). So neither this nor #117 crosses a boundary the model doesn't already concede. The `daemon.identify` mechanism here is reasonable, but turning the EDAEMON_ALREADY_RUNNING reconnect into a hard throw also regresses the split-brain reconnect from #93.

If we decide real pipe-squat defense is worth it, it should verify the pipe peer's SID/PID rather than HMAC keyed by the shared token. Happy to file a focused issue if so.

PR #139, opened 2026-06-06
Disable packaged CDP by default and bind debug to loopback

openwong2kim:
### Motivation
- Prevent exposure of an unauthenticated, predictable Chrome DevTools Protocol (CDP) endpoint in production builds that could be reached by untrusted webview content. 
- Require an explicit operator opt-in for enabling CDP in packaged apps to reduce the attack surface for remote `Runtime.evaluate` targeting privileged preload APIs. 
- Ensure any enabled CDP sessions are bound to loopback and use a randomized port (or validated explicit port) to make accidental exposure harder.

### Description
- Add a centralized resolver `src/main/browser-session/cdpConfig.ts` that decides whether CDP should be enabled, validates `WMUX_CDP_PORT`, supports `WMUX_DISABLE_CDP`, and requires `WMUX_ENABLE_CDP=true` to opt-in for packaged builds. 
- Update main process startup in `src/main/index.ts` to use the resolver and only append `remote-debugging-address`/`remote-debugging-port` when CDP is explicitly enabled, otherwise CDP remains disabled. 
- Bind enabled CDP sessions to `127.0.0.1` and select a randomized port within the configured range when no explicit port is provided. 
- Add unit tests in `src/main/browser-session/__tests__/cdpConfig.test.ts` covering default packaged behavior, explicit opt-in, explicit port handling, disable override, development randomization, and out-of-range port rejection.

### Testing
- Ran the new unit tests with `npm run test:parallel -- src/main/browser-session/__tests__/cdpConfig.test.ts` and all tests passed. 
- Ran type-checking with `npx tsc -p tsconfig.json --noEmit` and it completed successfully. 
- Linted the new files with `npx eslint src/main/browser-session/cdpConfig.ts src/main/browser-session/__tests__/cdpConfig.test.ts` and no new lint errors were introduced. 
- Running the repository-wide `npm run lint` surfaced pre-existing lint errors unrelated to this change and did not block this targeted fix.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_b_6a228d077dac8320b522aa86c9f401d2)

### DIFF
--- a/src/main/browser-session/__tests__/cdpConfig.test.ts
+++ b/src/main/browser-session/__tests__/cdpConfig.test.ts
@@ -1,0 +1,58 @@
+import { describe, expect, it } from 'vitest';
+import { resolveCdpRemoteDebuggingConfig } from '../cdpConfig';
+
+describe('resolveCdpRemoteDebuggingConfig', () => {
+  it('disables CDP by default in packaged builds', () => {
+    expect(resolveCdpRemoteDebuggingConfig({ env: {}, isPackaged: true })).toMatchObject({
+      enabled: false,
+      port: 0,
+    });
+  });
+
+  it('allows an explicit CDP opt-in for packaged builds on a randomized loopback port', () => {
+    expect(
+      resolveCdpRemoteDebuggingConfig({
+        env: { WMUX_ENABLE_CDP: 'true' },
+        isPackaged: true,
+        randomInt: () => 42,
+      }),
+    ).toEqual({ enabled: true, port: 18842 });
+  });
+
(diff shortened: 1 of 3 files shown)


openwong2kim:
Closing this. The premise is overstated: the CDP port is already loopback-only (no remote-debugging-address means Chromium binds 127.0.0.1), so it isn't the remote-network RCE the description implies. The packaged-build opt-in idea has some defense-in-depth merit, but it needs the framing corrected and a packaged smoke test confirming the first-party browser tools still work over the RPC fallback with CDP disabled. Not carrying it in this state; can revisit as a focused change if we decide packaged CDP should be off by default.

PR #140, opened 2026-06-06
fix(security): rebuild Windows token DACL with a DACL-only primitive

openwong2kim:
Supersedes the closed #124. The diagnosis there was right; the `Set-Acl` implementation wasn't.

### The bug
The Windows token-file hardening used `icacls /grant:r *<sid>:F /inheritance:r`, which only *replaces* the named principal's ACE and only strips *inherited* ACEs. A pre-existing **explicit** broad ACE (e.g. `Everyone:(R)` stamped by a redirected/roamed/MDM profile) survives, leaving the auth token world-readable. That's a SECURITY.md section 1.2 violation, since a leaked token drives the daemon RPC surface.

The closed PR #124 diagnosed this correctly but rebuilt the DACL with the PowerShell `Set-Acl` cmdlet, which re-stamps owner+group+DACL. On the already-protected state the shipped icacls flow leaves, that needs `SeSecurityPrivilege` a normal user process lacks, so it throws `PrivilegeNotHeldException` deterministically (10/10) on the v2.14.0+ installed base, silently turning the re-harden into a no-op (and failing Windows CI).

### The fix
Rebuild the DACL with the .NET `FileInfo.SetAccessControl(FileSecurity)` overload, which writes the **DACL section only**, never owner/group/SACL. It needs no privilege, succeeds on the upgrade-from-icacls state, and leaves exactly one owner FullControl ACE (every inherited/explicit/custom ACE dropped). Invoked via `powershell.exe -EncodedCommand` with the identity over stdin and the path over an env var, so non-ASCII names and paths never hit argv codepage mangling.

icacls stays as a fallback for SKUs where PowerShell is absent (owner FC, `/inheritance:r`, `/remove:g` of the well-known broad SIDs). I tried a blind icacls remove-all loop first and rejected it, since it locks the owner out. One documented bound: the fallback strips only well-known broad principals, not an arbitrary custom explicit SID. The primary path, which is virtually always taken, covers the general case.

Preserves #90 SID resolution, the pure-ASCII USERNAME guard, write-path fail-closed, and re-harden fail-soft.

### Verification
- `scripts/issue-124-acl-dynamic.mjs` drives the real compiled function against three seeded on-disk states (fresh-inherited, shipped-icacls, explicit-Everyone): owner-only DACL, owner read-back, no throw on the upgrade case. 16/16, and it confirms the original leak too.
- `security.test.ts` updated for the new invocation and icacls fallback (13 pass). The unit tests mock `execFileSync`, so the privilege/runtime behavior is covered by the dynamic harness instead.
- SECURITY.md section 1.2 and PROTOCOL.md section 5 updated. tsc/eslint/build:daemon clean.

### Merge gate
Per SECURITY.md section 1.1, this still wants real-`%USERPROFILE%` dogfood before merge. The `%TEMP%` harness covers the three ACL states deterministically but can't reproduce provider-specific (OneDrive/redirected/roaming) owner/group/inheritance descriptors. Please don't merge until that passes.


openwong2kim:
@Codex

openwong2kim:
@codex review

openwong2kim:
@codex review

PR #141, opened 2026-06-06
Workspace UX: duplicate, per-terminal cwd & working-dirs menu — plus cwd/identity fixes

snowyukitty:
## Summary

Follow-up UX and bug fixes on top of the per-workspace profiles feature
(#101, #103). It adds the workspace-management actions that feature implied
(duplicate, per-terminal working-directory visibility) and fixes a set of
cwd-reporting and identity-isolation bugs found while dogfooding.

Everything here applies to **new panes only** — existing and recovered daemon
PTYs are untouched, consistent with the profile feature's contract.

### Features (right-click a workspace)
- **Duplicate workspace** — clones the layout (fresh pane/surface ids, cleared
  ptyIds so new PTYs spawn) and the profile (re-normalized through the
  secret-name policy). Named `<name> (copy N)`.
- **Working directories menu** — lists each terminal's live cwd, with copy.
- **Per-surface cwd tracking** — every terminal tracks (and persists) its own
  cwd, shown in the tab tooltip.
- **Tab rename** — double-click a terminal tab to name it.
- **Close confirmation** — guards an accidental workspace close.

### Fixes
- **Per-terminal cwd always showed the startup home dir** (local *and* daemon
  mode). Two compounding parser bugs: OSC 7 left Windows paths as
  `/C:/Users/me`, and prompt detection matched the stale echoed prompt and
  froze cwd at startup. Parsing is extracted into a shared, unit-tested
  `cwdDetect` module (native-path + UNC normalization, reads the live prompt).
  Daemon mode additionally never forwarded its cwd to the renderer — a new
  `session:cwd` event closes that gap.
- **Context menu width** — the right-click menu rendered a blank gutter beside
  short items: its items are `<button>` (default `display: inline-block`), so
  under `w-max` the action buttons summed their widths on one line instead of
  taking the widest. It is now a flex column, so it hugs its content.

### Security / isolation
- **Child shells never inherit a stale wmux identity.** A wmux launched from
  inside a wmux pane (e.g. `npm start` while dogfooding) inherited the parent
  pane's `WMUX_WORKSPACE_ID` / `WMUX_SURFACE_ID` / `WMUX_SOCKET_PATH`, which
  could survive into new child shells (daemon mode never forces a socket path;
  some create paths run before a surfaceId exists). The reserved `WMUX_*`
  namespace is now cleared from the spawn baseline before identity is forced,
  making the "identity can't be spoofed" guarantee unconditional (matching the
  daemon's existing `WMUX_AUTH*` strip).

## Testing
- `tsc --noEmit` clean; `npm test` green, with new unit tests for `cwdDetect`,
  the `resolveSpawnEnv` identity strip, the `DaemonClient` cwd event, and the
  daemon env fallback.
- Heads-up: `StateWriter` / `crashRestore` tests are a pre-existing flake (they
  wait a fixed 50ms for an async fs write under fake timers) — unchanged by this
  PR and passing in isolation.

## Known limitations
- Recovery replays a pre-fix `sessions.json` env verbatim, so a session created
  before the identity fix keeps its old identity on restore. New sessions are
  clean; documented in code.
- New `workspace.*` / `workspaceProfile.*` i18n keys are English-only and fall
  back to English in other locales.

## Notes
- Rebased onto current `main`; already-merged work (#101–#137) is excluded from
  this diff.
- Self-reviewed across multiple passes before opening.


openwong2kim:
@codex

openwong2kim:
Merged, thanks for the thorough work on this one.

Really appreciate how much ground this covers without overreaching. Scoping everything to new panes and leaving recovered daemon PTYs untouched was the right call, and it keeps the contract from the profile feature intact.

The cwd parsing bugs were subtle. Pulling the OSC 7 and prompt-detection logic out into a shared cwdDetect module with its own tests makes that whole path much easier to reason about, and the daemon session:cwd event fills a gap I would have missed.

The identity strip is the part I like most. Clearing the WMUX_* namespace before forcing identity turns "can't be spoofed" into something unconditional instead of best-effort, and lining it up with the existing AUTH strip is clean.

Good catch on the context menu width too. The flex column fix is exactly what that needed.

You will get a credit in the next release notes. Thanks again for the careful self-review before opening this.

snowyukitty:
Thanks so much! I really appreciate the kind words and the credit.

I’m glad the scope and implementation choices landed well, especially keeping recovered daemon PTYs untouched and making the identity handling unconditional.

Happy to contribute, and thanks again for the thorough review! ^^

PR #142, opened 2026-06-07
fix(mcp): stop ghost workspace identity from stale pid-map entries

openwong2kim:
The MCP server resolves its workspace identity by reading ~/.wmux/pid-map/ (PID -> ptyId). Two compounding defects let it adopt a dead workspace id, and a third gap let the map keep leaking after the first fix.

## What was happening

The pid-map accreted one file per shell forever (over 1100 entries seen on my machine across two months). The OS recycles those PID numbers onto unrelated live processes (saw Notepad, Discord, RuntimeBroker in the wild), so a stale entry ends up sitting on a live, unrelated PID. On top of that, a2a.resolve.identity passed legacy "ws-" entries through verbatim, so a legacy entry on a recycled PID resolved to a re-minted dead workspace id. That ghost is absent from workspace.list, so browser_open failed with "no active workspace" and terminal ops threw "not owned by workspace ws-...", and it looped until the MCP server restarted.

## Fixes

**Read path (a2a.resolve.identity):** drop legacy "ws-" entries outright and resolve only current-format (ptyId) entries to their live owner. A dead or recycled current-format entry resolves to null and is excluded, so it can never produce a ghost. No liveness probe needed on the hot read path.

**Write boundary (pty.handler.ts):** prune the pid-map anchor when a daemon session ends, instead of letting it accrete.

This started as the session:died handler only. During dogfood we found the common close path does not go through session:died at all: a pane or workspace close goes PTY_DISPOSE -> daemon.destroySession, which removes the bridge exit listener before killing and emits session:destroyed, not session:died. So the died handler never fired for an explicit close and every UI-driven close still leaked its anchor (caught live: PID 37008 -> daemon-33d7bc91 lingering after a workspace close). The dispose handler now prunes too, so both teardown paths clean the map.

The fs primitives (writePidMap / removePidMapByPtyId) moved into src/main/pty/pidMap.ts so their content-keying and unlink behavior can be tested at runtime. The handler module pulls in electron and can't be imported under vitest, which is why its other tests are source-structural.

**Identity hint (mcp + company/mcp resolveWorkspaceId):** verify the frozen WMUX_WORKSPACE_ID env hint against workspace.list before trusting it. Drop it only on positive proof it is a ghost ('absent'); keep it on 'unknown' (list transiently unavailable, e.g. the renderer's retryable boot-reconcile envelope) so a startup race never turns a live workspace into a hard "identity unknown" failure. Shared classifier in workspaceIdentity.ts so both MCP surfaces behave identically.

## Tests

- 19 unit tests for the resolve path (legacy drop + file purge, live resolve, non-destructive read of dead current-format entries, env-hint 3-state classify).
- Behavioral pidMap.test.ts (5): multi-anchor sweep, content-vs-filename keying, whitespace trim, empty-ptyId and missing-dir no-op.
- Structural pty.handler.pidmap-prune.test.ts (3): both prune call sites plus shared-helper routing.
- Dynamic harness (workspace-identity-dynamic.mjs) updated, 7/7 including a ghost env-hint drop scenario.

Full suite green, tsc clean.

## Note on the dispose prune

The ghost itself is already blocked by the read-path defense regardless of the prune, so the dispose-path leak was a hygiene issue (unbounded growth), not a live ghost regression. The end-to-end "close a workspace, watch the anchor disappear" check needs a GUI restart to load the main-process change, so it's deferred to the next natural close rather than gated on a forced restart.

### DIFF
--- a/scripts/helpers/mini-pipe-server.mjs
+++ b/scripts/helpers/mini-pipe-server.mjs
@@ -80,10 +80,21 @@ function getSimulatedWorkspaces() {
   }
 }
 
+// First-call-only mode: when WMUX_MINISERVER_RESOLVE_ONCE is set, the PID map
+// resolves on the FIRST a2a.resolve.identity call and returns empty afterward.
+// Models a workspace that resolved successfully once, then had its pid-map entry
+// vanish (pane closed / re-minted) — used to exercise the stale-cache fallback
+// gate (resolveWorkspaceId must not return a confirmed-dead cached id).
+let resolveCallCount = 0;
+
 function handleA2aResolveIdentity() {
   const dir = getPidMapDir();
   const owners = getSimulatedOwners();
   const mappings = {};
+  resolveCallCount += 1;
+  if (process.env.WMUX_MINISERVER_RESOLVE_ONCE && resolveCallCount > 1) {
+    return { mappings }; // empty on subsequent calls
+  }
   try {
     if (fs.existsSync(dir)) {
       for (const file of fs.readdirSync(dir)) {
--- a/scripts/helpers/mini-pipe-server.mjs
+++ b/scripts/helpers/mini-pipe-server.mjs
@@ -15,7 +15,15 @@
  *     workspace lookup (real handler: input.findOwnerWorkspace on the
  *     renderer) is simulated here via WMUX_MINISERVER_OWNERS, a JSON
  *     { <ptyId>: <workspaceId> } map. Legacy "ws-"-prefixed entries are
- *     passed through; ptyIds with no live owner are omitted.
+ *     DROPPED (file deleted) — not passed through — mirroring the fix that
+ *     stopped legacy debris from resurfacing as ghost workspaces. ptyIds with
+ *     no live owner are omitted but their files are left on disk (the read path
+ *     is non-destructive for current-format entries).
+ *
+ *   workspace.list
+ *     Returns the simulated live workspace array from WMUX_MINISERVER_WORKSPACES
+ *     (JSON array of { id }). Used by the probe to gate the env hint: a hint
+ *     absent from this list is a confirmed ghost and must be dropped.
  *
  *   mcp.claimWorkspace
  *     Returns an explicit "no window" error to mirror what the real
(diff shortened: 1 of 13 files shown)


openwong2kim:
@codex

openwong2kim:
@codex review

Addressed the P1 in 61848d3: legacy "ws-" pid-map entries are no longer purged unconditionally. a2a.resolve.identity now verifies each against the live workspace.list — resolves and keeps it if the workspace is live (an upgraded pane's anchor), purges only on confirmed-absent, and keeps it on 'unknown' (list unavailable). Lazy single fetch, cached per call.

openwong2kim:
@codex review

On the R2 finding (keeping a live legacy ws- entry can mis-route a recycled PID to a live-but-wrong workspace): you're right, and it exposed that the R1 keep-approach was unsafe. Reverted in 8731ff5 back to unconditional purge of legacy "ws-" entries.

Reasoning: legacy PID->workspaceId entries are both unverifiable (no ptyId to live-resolve, and workspace.list can't prove the PID still maps to that pane) and unprunable (removePidMapByPtyId keys by ptyId content, so it can never match ws- files). Anything kept therefore lives forever and becomes a recycled-PID mis-route. The earlier R1 case (a live pane from an older version losing its anchor) is bounded and self-healing instead: the pid-map writers have only ever stored ptyIds, and a genuinely-live pane re-anchors with a current-format entry on its next reconnect. The R1-vs-R2 tradeoff is now recorded in a code comment so it doesn't get re-litigated.

Back to 17 unit + 7 dynamic green, tsc clean.

openwong2kim:
@codex review

Fixed the R3 P2 in 2812571. resolveWorkspaceId now gates the cached-identity fallback the same way as the env hint: it returns MY_WORKSPACE_ID only if that workspace isn't a confirmed ghost, and clears the cache (id + resolved flag) on 'absent' so the next call re-resolves clean. Kept on 'unknown' to ride out a transient workspace.list outage. Applied to both src/mcp and src/company/mcp.

Dynamic coverage added (W2d): first resolve caches the owner via a PID-map hit, then a forced re-resolve with an empty pid-map + no env hint + the cached workspace absent from workspace.list returns "" instead of the dead id. 8/8 dynamic, tsc clean.

PR #144, opened 2026-06-07
fix(mcp): remove the empty pane left by browser_close

### LINKED ISSUE #143: MCP browser_close leaves an empty pane that backfills with a new terminal
Closing a browser via the MCP `browser_close` tool removes the browser surface but leaves its pane behind, and the pane then gets backfilled with a fresh PowerShell terminal. Repeating `browser_open` then `browser_close` accumulates empty terminals in the workspace.

## Repro

In a workspace with one terminal:
1. `browser_open({ url: "https://example.com" })`
2. `browser_close()`
3. repeat 1-2 once more

`surface_list()` then shows the original terminal plus two extra PowerShell panes (one per close). The UI path (clicking the surface close button) does not do this.

## Root cause

The teardown cascade that the UI path runs is missing from the MCP path:

- MCP: `browser_close` -> `browser.close` RPC (src/main/pipe/handlers/browser.rpc.ts) -> renderer `useRpcBridge.ts` calls `store.closeSurface(leaf.id, surfaceId)` only. `closeSurface` (surfaceSlice.ts) splices the surface out but leaves the pane in place.
- UI: `Pane.tsx` `handleCloseSurface` calls `closeSurface` and then, if that was the last surface, `closePane(pane.id)` (the cascade `SurfaceTabs.tsx` also documents). The MCP handler has no equivalent `closePane` line, so it leaves a leaf pane with `surfaces.length === 0`.
- `AppLayout.tsx`'s "auto-create initial surface for empty leaf panes" effect detects the empty leaf and spawns a new terminal to fill it. That effect is intended (it stops empty placeholder panes from lingering); the actual defect is the missing cascade upstream, so the auto-recovery plants a terminal where the pane should have been removed.

So this is the same shape as the pid-map leak just fixed in #142: a teardown cascade that exists in the primary/UI path is missing from its RPC mirror. Worth a quick sweep for other RPC handlers that mirror a UI action but skip its cleanup cascade.

## Direction (not yet implemented)

In the `useRpcBridge.ts` `browser.close` handler, mirror `handleCloseSurface`'s conditional cascade after `store.closeSurface`: if the target leaf has `surfaces.length <= 1`, also `closePane(targetLeaf.id)`. A browser sharing a split pane with a terminal (`surfaces.length > 1`) then correctly keeps the pane and only drops the browser surface. Mind the mutation-timing snapshot: `Pane.tsx` judges the length before calling `closeSurface`, so the mirror should read it the same way.

## Follow-up (from #142 review, fold in here)

`resolveWorkspaceId` (src/mcp/index.ts and src/company/mcp/index.ts) can issue two `workspace.list` calls in one resolve on the rare fallback path where the env hint exists, is absent, and a cached identity also exists (one to gate the env hint, one to gate the cache). `a2a.resolve.identity` already caches the live set per call via `fetchLiveWorkspaceIds`; resolveWorkspaceId could fetch once on entry and classify both the hint and the cache against that set. Non-blocking efficiency cleanup, fallback path only.

openwong2kim:
Closes #143

Closing a browser through the MCP `browser_close` tool removed the browser surface but left its pane behind, and the empty pane was then backfilled with a fresh PowerShell terminal. Repeating `browser_open` then `browser_close` accreted blank terminals.

## Cause

The MCP path was missing the cascade the UI path runs. `Pane.tsx` `handleCloseSurface` closes the surface and then, when it was the pane's last one, closes the pane too. The `browser.close` handler in `useRpcBridge.ts` only called `store.closeSurface`, leaving a `surfaces.length === 0` leaf that AppLayout's "auto-create initial surface for empty leaf panes" effect filled with a terminal. Same shape as the #142 pid-map leak: a teardown cascade present in the UI path but absent from its RPC mirror.

## Fix

Mirror the cascade in the handler. Snapshot whether this is the pane's last surface BEFORE closing it (so the decision matches `Pane.tsx`, which reads the count pre-close), then `closePane` when it was. A browser sharing a split pane with a terminal has `surfaces.length > 1`, so only the surface is removed and the pane stays.

`closeSurface` and `closePane` are two synchronous `set()` calls; zustand batches them so React never renders the intermediate empty-leaf state and the auto-terminal effect stays dormant. There's a comment marking that the two calls must stay synchronous and adjacent.

## Scope note (root pane)

`closePane` is a no-op on the root pane by design (a workspace can't have zero panes), so a browser that is a workspace's only pane still gets an auto-terminal. The UI path behaves identically there, so it's consistent, not an asymmetry. Only the non-root case was leaking, and that's what this fixes.

## Tests

Source-structural test (handleRpcMethod can't be imported under vitest, same constraint as the pty.handler tests): asserts the cascade is present and that the last-surface decision is taken before the surface is removed. 303 renderer tests green, tsc clean.

## Follow-up

The #143 issue also tracks a non-blocking efficiency cleanup carried over from the #142 review (resolveWorkspaceId can issue two `workspace.list` calls on a rare fallback path).

### DIFF
--- a/src/renderer/hooks/__tests__/useRpcBridge.browserClose.test.ts
+++ b/src/renderer/hooks/__tests__/useRpcBridge.browserClose.test.ts
@@ -1,0 +1,66 @@
+import { describe, it, expect } from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+/**
+ * Regression test for issue #143: MCP browser_close left an empty pane that
+ * AppLayout backfilled with a fresh terminal.
+ *
+ * Root cause: the UI close path (Pane.tsx handleCloseSurface) removes the pane
+ * when its last surface is closed (closeSurface then closePane), but the MCP
+ * mirror (browser.close in useRpcBridge.ts) only called closeSurface, leaving a
+ * surfaces.length === 0 leaf that the "auto-create initial surface for empty
+ * leaf panes" effect then filled with a terminal. browser_open/close in a loop
+ * accreted blank terminals.
+ *
+ * Fix: the handler snapshots whether this was the last surface BEFORE closing,
+ * then cascades into closePane — mirroring the UI path. (Root panes are a no-op
+ * in closePane by design, so a browser that is a workspace's only pane still
+ * gets an auto-terminal on both paths; only the non-root asymmetry was a bug.)
+ *
+ * This is a source-structural test: handleRpcMethod is not exported and pulls in
+ * the whole store, so it can't be imported under vitest — the same reason the
+ * pty.handler tests scan source. It fails if a refactor drops the cascade or
+ * reorders it so the last-surface decision is made AFTER the surface is gone.
+ */
+describe('useRpcBridge browser.close cascade (issue #143)', () => {
+  const src = fs.readFileSync(
+    path.join(__dirname, '..', 'useRpcBridge.ts'),
+    'utf-8',
+  );
+
+  /** Isolate the browser.close handler region so assertions can't match elsewhere. */
+  function closeBlock(): string {
+    const match = src.match(
+      /method === 'browser\.close'[\s\S]*?method === 'browser\.navigate'/,
+    );
+    if (!match) {
+      throw new Error(
+        "browser.close -> browser.navigate handler region not found in " +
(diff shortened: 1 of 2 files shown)


openwong2kim:
@codex review

PR #97, opened 2026-06-04
test: serialize runtime tests

snowyukitty:
## Summary

Split Vitest into a regular parallel lane and a serialized runtime lane so tests that spawn real OS resources do not contend with the full suite.

## Changes

- Add a `*.runtime.test.ts` / `*.runtime.test.tsx` / `*.runtime.test.mjs` convention for tests that spawn real OS resources.
- Run the regular suite in parallel while excluding runtime tests.
- Add `vitest.runtime.config.ts` to run runtime tests with `fileParallelism: false`.
- Rename `ProcessMonitor.test.ts` to `ProcessMonitor.runtime.test.ts` because it exercises real process liveness probes.
- Use `npm test` in the cross-platform baseline workflow so CI uses the same test entry point documented for contributors.

## CHANGELOG entry

no CHANGELOG - internal test/CI only

## Stability tier impact

- [x] No external surface touched (internal refactor / docs / tests / CI).
- [ ] Additive change to a `stable` surface (new optional param, new field, new event type).
- [ ] Breaking change to a `stable` surface (requires major bump - block until release planning is in scope).
- [ ] Change to an `experimental` surface (note in release notes).
- [ ] Change to an `internal` surface (no contract impact).

## Substrate contract impact (Substrate 3.0)

n/a

## Test plan

- `npx tsc --noEmit`
- `npm test`
- `npm run test:runtime`
- `npx vitest run --config vitest.runtime.config.ts --silent` repeated 3 times locally

## Related issues / PRs

n/a

## Reviewer checklist

- [ ] CHANGELOG entry above is filled in or explicitly marked n/a.
- [ ] Stability tier impact is correctly classified.
- [ ] Substrate contract docs (PROTOCOL.md, inventory.md, stability.md) updated if the surface changed.
- [ ] Tests cover the new behavior and the regression case (if a bug fix).
- [ ] No accidental commit of secrets, tokens, `.env`, or unrelated large binaries.

PR #153, opened 2026-06-08
fix(terminal): Ctrl+J inserts a newline under a CJK IME

snowyukitty:
## Problem

Inside in-pane TUIs (codex, Claude Code), **Ctrl+J intermittently failed to insert a newline** — it worked with the IME off and broke with a Chinese / Japanese / Korean IME on.

## Root cause

The byte pipeline *below* xterm is fully transparent to `\n` (`onData` → `chunkOnDataIfNeeded` → `pty.write` → `sanitizePtyText` → daemon `SessionPipe` → node-pty; `sanitizePtyText` only strips NUL + C1). The keystroke was lost at **xterm's keyboard layer**:

- `attachCustomKeyEventHandler` did not intercept Ctrl+J, so it fell through to xterm.
- xterm 6 derives `Ctrl+<letter>` from the **deprecated `KeyboardEvent.keyCode`** (matches `65–90` → `String.fromCharCode(keyCode-64)`).
- With an IME enabled the keydown reports `keyCode === 229` ("Process") and `key !== 'j'`, so that branch never matched and **emitted nothing**. This is IME-state-dependent, which is the "sometimes" users saw. `keyCode === 229` is the IME-universal "Process" value, so all CJK IMEs (and others) are affected.

## Fix

A small pure module `resolveNewlineKeyByte()` resolves the deterministic newline keys (Shift+Enter, Ctrl+J) from the **physical `event.code`** and writes the byte directly, bypassing xterm — the same IME-safe `e.code` approach already used for the split shortcuts (`KeyD`, …). `e.code` is unaffected by the IME (which only mangles `key`/`keyCode`/`which`).

Guards (added after an independent review):
- **Defers during an active IME composition** (`isComposing`) so a preedit is never split. The reported bug is Ctrl+J while the IME is *idle* (no preedit, `isComposing === false`), so the fix still applies there.
- **Defers to an explicit user Ctrl+J custom keybinding** so the implicit newline never shadows a configured binding.

Notes:
- xterm 6 has **no kitty/modifyOtherKeys path** (only `applicationKeypad`, unrelated), so the emitted byte is identical to xterm's legacy output when no IME is active — no behavior change off-IME.
- Keyed on the physical `KeyJ`, consistent with every other wmux shortcut; the Dvorak/Colemak trade-off is documented in-code.

## Testing

- New unit tests for `resolveNewlineKeyByte` (11 cases): physical-code match, IME `key='Process'`, modifier exclusions, `isComposing` deferral, custom-binding deferral, and the preserved Shift+Enter behavior.
- `tsc --noEmit`: 0 errors.
- Full `test:parallel`: **191 files / 2345 tests pass**.


openwong2kim:
Merged, thanks! Clean fix. The keyCode 229 root cause is the kind of thing you can stare at for an hour and miss, and routing the newline keys through the physical event.code with the isComposing + custom-binding guards was exactly the right call. The 11 cases covering the IME-mangled path made it easy to verify on my end.

That's your fourth one now (after #97, #101, #103). Really appreciate you taking on the less glamorous terminal-input bugs. CHANGELOG credit goes out with the next release.

PR #165, opened 2026-06-09
Close #15: substrate reference plugin, Diátaxis docs, perf characterization

### LINKED ISSUE #15: Feature Request: Expose Workflow-Friendly APIs for External Tooling
## Summary

wmux already provides a strong Windows-native terminal multiplexer experience for AI-assisted development workflows.

It would be valuable if wmux exposed a small set of workflow-friendly APIs and metadata hooks that external tools/plugins could optionally use for orchestration, observability, and workspace coordination.

Importantly, this proposal does **not** request AI automation, agent control, prompt injection, or AI output collection.

Instead, the goal is to expose generic terminal/session primitives that external tooling can build upon.

---
### 1. Pane Metadata

Allow optional metadata to be associated with panes/workspaces.

Example:

```json
{
  "paneId": "pane-1",
  "label": "Backend",
  "role": "service",
  "status": "running-tests"
}
```

Possible uses:

* workspace organization
* pane labeling
* workflow visualization
* developer tooling integrations

---
### 2. Event Hooks / JSON-RPC Events

Expose lightweight events such as:

* pane created
* pane closed
* pane focused
* process started
* process exited

This would allow external tools to react to workspace activity without directly interacting with AI systems.

---
### 3. Optional Pane Buffer Access

Optional read-only access to terminal buffers/logs could help external tools monitor:

* build logs
* test runners
* task completion
* shell workflow state

This would be useful for workflow tooling and observability dashboards.

---
## Why This Fits wmux

This proposal stays aligned with the current wmux philosophy:

* wmux does not automate AI systems
* wmux does not capture AI outputs for external services
* wmux remains a neutral terminal/workspace layer

The proposed APIs would simply allow external tooling to build richer developer workflows around wmux.

---
## Possible Future Ecosystem

These APIs could enable:

* workflow dashboards
* multi-pane coordination tools
* workspace observers
* task-aware layouts
* external orchestrator plugins

while keeping wmux itself clean and policy-safe.

openwong2kim:
Closes #15.

Issue #15 asked what it actually takes to build a plugin/dashboard/orchestrator on top of the wmux substrate. This lands the missing pieces: a runnable reference plugin, a real docs structure, honest perf numbers, and a few doc-vs-code drift fixes found while writing them.

## What's here

**Reference plugin** — `examples/event-recorder/`
A small, dependency-free plugin that connects over the named pipe, identifies, declares permissions, and records the event stream to NDJSON. Runs in both legacy (no `clientName`) and identity modes. This is the thing a new integrator copies from.

**Docs (Diátaxis)** — `docs/`
- `tutorials/build-a-substrate-plugin.md` — learning-oriented walkthrough
- `how-to/{connect-to-wmux,write-pane-metadata,react-to-events,handle-daemon-restart}.md` — task recipes
- `api/reference.md` — generated from the live RPC surface by `scripts/gen-api-reference.mjs`
- `README.md` — index tying it together

**Performance characterization** — `docs/internal/substrate-perf.md` + `scripts/substrate-bench.mjs`
End-to-end bench over the named pipe against the packaged app (the path a real plugin uses, IPC and rate limiter included), answering the open m0 §8 Q5 profiling note:
- B1 metadata write: paced under the 50/s per-socket wire limit; the achieved 31.35/s is dispatch-cadence-bound (sequential await + timer granularity), not the cap itself. Per-write latency p50/p95/p99 = 5.7/9.3/10.6 ms, version monotonic.
- B2 `events.poll`: p50 ~1–2 ms across page sizes; the scan is the ring walk, larger `max` only adds serialization.
- B3 ring overflow: first drop at backlog ~720, `resync: true` + `droppedCount`, recovered via `pane.list`.
- B4 snapshot idle vs mid-burst: p50 flat (1.75 vs 1.78 ms) — snapshot does **not** block the write path, no copy-on-write needed. p95 tail rises (1.8x) from shared event-loop queueing, not store contention.

Numbers are from a real run (errors=0); they're environment-dependent and the doc says so.

**Drift fixes** — `docs/PROTOCOL.md` §5, `docs/api/inventory.md`, `docs/api/mcp-plugin-spec.md`
The documented transport was stale: the Windows pipe is `\\.\pipe\wmux-<username>` (not `wmux-rpc-<token>`), the token is a plain UUID at `~/.wmux-auth-token` (not `%APPDATA%\wmux\pipe-token`), plus the POSIX socket, the loopback TCP fallback, and PTY-inherited `WMUX_AUTH_TOKEN`. Corrected to match `src/shared/constants.ts` and `PipeServer`. No wire-format change.

**Also** — `src/main/mcp/firstParty.ts` (+ belt test): #113 residual-risk guard. Small and self-contained; happy to split it into its own PR if you'd rather keep this docs-only.

## Verification
- `scripts/issue-15-verify.mjs` — recorder runs end to end + identity-mode `pendingApproval` contract holds (10/10).
- `firstParty.test.ts` green locally.
- bench + verify were run against the packaged build with wmux fully shut down (isolated temp HOME, daemon spawned/killed per run).

## Post-review pass (e4c972a)

A full review pass over this branch turned up doc-vs-code drift and a few reference-plugin bugs, all fixed in the follow-up commit: regenerated `reference.md` (it was stale at v2.17.1) and wired `gen-api-reference --check` into vitest so the no-drift claim is enforced; corrected the `mcp.claimWorkspace` row, the tutorial's winget id / output samples / first-gated-RPC / denied-recovery note, and the rate-limit teardown claim; the recorder now keeps its cursor on transient poll errors and records the page a resync reply delivers (only `droppedCount` events are lost); the bench counts only confirmed writes and labels B1's bound from the run. 2562 tests green.


openwong2kim:
@codex

PR #178, opened 2026-06-10
fix(pty): prefer PowerShell 7 over Windows PowerShell 5.1 as default shell

### LINKED ISSUE #176: 기본 셸이 PowerShell 7 설치 시에도 Windows PowerShell 5.1을 우선 선택합니다.
**유형**: bug  
**영역**: 기본 셸(default shell) 결정 로직

> 소스 링크는 커밋 [`9e1f69d`](https://github.com/openwong2kim/wmux/tree/9e1f69d4c177e893d7533552e19e33b3a7353d7a) 기준 permalink입니다.

## 요약

`C:\Program Files\PowerShell\7\pwsh.exe`(PowerShell 7)가 설치돼 있어도 wmux가 자동으로 **Windows PowerShell 5.1**을 기본 셸로 선택한다. PowerShell 7을 쓰려면 Settings에서 매번 수동으로 선택해야 하고, 자동으로는 선택되지 않는다.

## 재현

1. PowerShell 7을 설치한 Windows 환경에서 wmux 실행.
2. 별도 셸 설정 없이 새 터미널 surface 생성.
3. 시작된 셸이 PowerShell 7이 아니라 Windows PowerShell 5.1(`%SystemRoot%\System32\WindowsPowerShell\v1.0\powershell.exe`)이다.

## 기대 동작

PowerShell 7이 설치돼 있으면 그것을 기본 셸로 우선 선택한다. (설치된 최신 PowerShell 우선 — 신규 escape-sequence 지원 등 호환성 이점이 있고, Settings 셸 목록도 이미 PowerShell 7을 먼저 나열한다.)

## 원인 — 셸 기본값 결정 경로 간 우선순위 모순

| 경로 | PowerShell 7 우선순위 | 실제 사용처 |
|---|---|---|
| [`ShellDetector.detectWindows()`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/ShellDetector.ts#L21-L67) | ✅ 7을 먼저 나열 | Settings 셸 **목록** 표시 ([`shell.handler.ts`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/ipc/handlers/shell.handler.ts#L29-L33)) |
| [`ShellDetector.getDefault()`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/ShellDetector.ts#L165-L171) | ✅ 7 반환 | ❌ 프로덕션에서 호출되지 않음 (테스트만) |
| [`PTYManager.getDefaultShell()`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/PTYManager.ts#L260-L276) | ❌ **5.1을 먼저** | ✅ 실제 자동 기본값 |

[`PTYManager.getDefaultShell()`의 후보 배열](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/PTYManager.ts#L264-L269)이 5.1을 7보다 먼저 둔다:

```js
const candidates = [
  `${process.env.SystemRoot}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe`, // 5.1 — 항상 존재 (L265)
  `${process.env.ProgramFiles}\\PowerShell\\7\\pwsh.exe`,                         // 7 — 도달 못 함 (L266)
  'powershell.exe',
  'cmd.exe',
];
// 존재하는 첫 번째를 반환
```

5.1은 모든 Windows에 항상 존재하므로 항상 먼저 매칭되어, PowerShell 7 후보에 도달하지 못한다. 또한 기본 셸 설정값이 레거시 문자열 [`'powershell'`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/renderer/stores/slices/uiSlice.ts#L599)(실행 경로 아님)이라 이 폴백으로 떨어진다.

결과적으로, 올바른 우선순위를 가진 [`ShellDetector.getDefault()`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/ShellDetector.ts#L165-L171)는 dead code 상태이고, 실제 동작은 5.1을 강제한다.

## 제안 수정

다음 중 하나(또는 조합):

1. [`PTYManager.getDefaultShell()`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/PTYManager.ts#L264-L269)의 후보 순서에서 **PowerShell 7을 5.1보다 먼저** 둔다 ([`ShellDetector`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/ShellDetector.ts#L21-L40)와 정렬).
2. 자동 기본값 산정을 [`ShellDetector.getDefault()`](https://github.com/openwong2kim/wmux/blob/9e1f69d4c177e893d7533552e19e33b3a7353d7a/src/main/pty/ShellDetector.ts#L165-L171)로 일원화하고 `PTYManager`가 이를 사용하도록 배선 — 중복된 셸 우선순위 로직 제거.

두 곳에 흩어진 셸 우선순위 로직을 한쪽으로 정렬하면 모순과 dead code가 함께 해소된다.

zer0ken:
## Summary

On Windows the auto-selected default shell always resolved to Windows PowerShell 5.1 even when PowerShell 7 was installed. Both default-shell paths now delegate to `ShellDetector.getDefault()`, which prefers pwsh 7 (5.1 remains the fallback).

## Changes

- `src/main/pty/PTYManager.ts` — `getDefaultShell()` Windows branch now delegates to `ShellDetector.getDefault()` instead of its own candidate array that listed 5.1 first. Removes the duplicated, conflicting ordering.
- `src/main/ipc/handlers/pty.handler.ts` — daemon-mode default shell also uses `ShellDetector.getDefault()` on Windows instead of a bare `"powershell.exe"`, so local and daemon modes pick the same default.
- `src/main/pty/__tests__/ShellDetector.test.ts` — locks the Windows contract: pwsh 7 is listed before 5.1 and chosen by `getDefault()`; falls back to 5.1 when pwsh 7 is absent.

### Changed

- On Windows, PowerShell 7 is now preferred as the default shell when installed (Windows PowerShell 5.1 remains the fallback).

### Fixed

- New terminals no longer always launched Windows PowerShell 5.1 when PowerShell 7 was available.

## Stability tier impact

n/a — no PaneMetadata, EventBus, permission, Named Pipe, or RPC/event surface is touched.

## Test plan

- Unit tests added: `src/main/pty/__tests__/ShellDetector.test.ts` — Windows pwsh-7-over-5.1 ordering + `getDefault()` selection, and the 5.1 fallback when pwsh 7 is absent.
- Suite run locally: `vitest run src/main/pty src/main/ipc/handlers` → 183 passing; `tsc --noEmit` clean.
- Manual verification: launched a fresh-profile dev instance with no explicit shell setting; the new pane's default shell was created as PowerShell 7 (`session.json` recorded `"shell": "PowerShell 7"`). Before the fix the same conditions produced 5.1.

## Related issues / PRs

Closes #176

## Reviewer checklist

### DIFF
--- a/src/main/ipc/handlers/pty.handler.ts
+++ b/src/main/ipc/handlers/pty.handler.ts
@@ -4,6 +4,7 @@ import path from 'node:path';
 import { StringDecoder } from 'node:string_decoder';
 import { PTYManager } from '../../pty/PTYManager';
 import { PTYBridge } from '../../pty/PTYBridge';
+import { ShellDetector } from '../../pty/ShellDetector';
 import { DaemonClient } from '../../DaemonClient';
 import { IPC, ENV_KEYS } from '../../../shared/constants';
 import { writePidMap, removePidMapByPtyId } from '../../pty/pidMap';
@@ -209,7 +210,10 @@ export function registerPTYHandlers(
 
       const safeCwd = validateCwd(options?.cwd);
       const effectiveCwd = safeCwd ?? require('os').homedir();
-      const shell = options?.shell || (process.platform === 'win32' ? 'powershell.exe' : (process.env.SHELL || '/bin/bash'));
+      // Daemon-mode default shell. On Windows prefer PowerShell 7 over 5.1 via
+      // ShellDetector (issue #176) — mirrors PTYManager.getDefaultShell() so
+      // both modes pick the same default.
+      const shell = options?.shell || (process.platform === 'win32' ? new ShellDetector().getDefault() : (process.env.SHELL || '/bin/bash'));
 
       // Generate a unique session ID
       const crypto = require('crypto');
--- a/src/main/pty/PTYManager.ts
+++ b/src/main/pty/PTYManager.ts
@@ -5,6 +5,7 @@ import path from 'node:path';
 import { getPipeName, ENV_KEYS, getPidMapDir } from '../../shared/constants';
 import { resolveSpawnEnv } from './resolveSpawnEnv';
 import { isWindows } from '../../shared/platform';
+import { ShellDetector } from './ShellDetector';
 
 export type ShellType = 'powershell' | 'bash' | 'cmd' | 'unknown';
 
@@ -259,18 +260,11 @@ export class PTYManager {
 
   private getDefaultShell(): string {
     if (process.platform === 'win32') {
-      // Try PowerShell paths in order — basename alone may fail
-      // if Electron's PATH is limited (e.g. installed via install.ps1)
-      const candidates = [
-        `${process.env.SystemRoot}\\System32\\WindowsPowerShell\\v1.0\\powershell.exe`,
-        `${process.env.ProgramFiles}\\PowerShell\\7\\pwsh.exe`,
-        'powershell.exe',
-        'cmd.exe',
-      ];
-      for (const s of candidates) {
-        try { if (fs.existsSync(s)) return s; } catch { /* skip */ }
-      }
-      return 'cmd.exe';
+      // Single source of truth for shell preference: ShellDetector lists
+      // PowerShell 7 before Windows PowerShell 5.1, so pwsh 7 is the default
+      // when installed (issue #176). 5.1 is the fallback — present on every
+      // Windows box, and absolute paths sidestep a limited Electron PATH.
+      return new ShellDetector().getDefault();
     }
     return process.env.SHELL || '/bin/bash';
   }
--- a/src/main/pty/__tests__/ShellDetector.test.ts
+++ b/src/main/pty/__tests__/ShellDetector.test.ts
@@ -1,5 +1,6 @@
 import { describe, it, expect, beforeEach, afterEach, vi } from 'vitest';
 import fs from 'node:fs';
+import path from 'node:path';
 
 // Hoisted mutable state shared between the vi.mock factory and the tests.
 const platformState = vi.hoisted(() => {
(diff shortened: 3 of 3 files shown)


zer0ken:
중복 구현이 없는지, 구현의 활용 여부가 일관되는지 검증 중입니다.

openwong2kim:
Merged — thank you for the excellent contribution, @zer0ken! 🙏

This is exactly the right fix: collapsing the duplicated shell-priority logic onto a single `ShellDetector.getDefault()` source of truth (instead of just reordering the array) kills both the 5.1-first bug **and** the dead-code path in one move. The diagnosis in #176 was unusually thorough, and the tests lock in both the pwsh7-first ordering and the 5.1 fallback. The `path.join` detail so the 5.1 mock matches on POSIX CI runners was a nice touch.

**Verification performed before merge:**
- Static review: no regression in `detect()` ordering (PowerShell always precedes Git Bash/WSL), daemon mode moves from a bare `powershell.exe` basename to an absolute path (more robust under a limited Electron PATH), and the returned pwsh7 path still classifies as `powershell` via `detectShellType()`.
- `vitest run src/main/pty src/main/ipc/handlers` → 183/183 passing, including your two new cases.
- `tsc --noEmit` clean; CI green across all three OS baselines.

**One thing I found while dogfooding (out of scope for this PR — not a defect here):** on my box pwsh7 is installed *only* as the Microsoft Store build, which lives at `%LOCALAPPDATA%\Microsoft\WindowsApps\pwsh.exe` as an 85-byte App Execution Alias (a reparse-point symlink). `fs.existsSync()` returns `false` for that alias, so `ShellDetector.detectWindows()` doesn't pick it up and still falls back to 5.1 — independent of this PR, since the detector's path list is untouched here. I'll file a follow-up issue for Store-build / alias detection. If that's something you'd want to take on, it'd be very welcome — but no obligation at all.

You'll be credited in the CHANGELOG Contributors section on the next release. Thanks again!

PR #184, opened 2026-06-10
feat(workspace): add Rename to right-click context menu

matdac6:
## What

Adds a **Rename workspace** item to the top of the workspace right-click context menu in the sidebar.

## Why

wmux already supports renaming a workspace via:
- double-click on the workspace name
- `Ctrl+Shift+R`
- tmux prefix + `,`

…but the right-click context menu had no rename entry, which is the first place many users look. As a daily user I kept right-clicking expecting to find it. This adds that missing, most-discoverable entry point.

## How

The rename machinery (inline editor, `renameWorkspace` store action, session persistence) already exists. This change only wires a new menu button to the existing flow: clicking **Rename workspace** closes the menu and enters the inline editor (`setEditing(true)`), exactly like double-click does.

## Changes
- `src/renderer/components/Sidebar/WorkspaceItem.tsx` — new context-menu button as the first item, before *Configure profile…*
- `src/renderer/i18n/locales/en.ts` — new `workspace.rename` label (other locales fall back to `en`, consistent with the sibling menu items `configureProfile`/`duplicate` which are also en-only)

## Testing
- `npx tsc --noEmit -p tsconfig.json` passes clean
- Verified the menu item opens the existing inline rename editor and commits via Enter

### DIFF
--- a/src/renderer/components/Sidebar/WorkspaceItem.tsx
+++ b/src/renderer/components/Sidebar/WorkspaceItem.tsx
@@ -353,6 +353,13 @@ export default function WorkspaceItem({ workspace, isActive, isMultiview, index,
           <button
             className="w-full text-left px-3 py-1.5 text-xs transition-colors hover:bg-[var(--bg-overlay)]"
             style={{ color: 'var(--text-main)' }}
+            onClick={() => { setMenuPos(null); setEditName(workspace.name); setEditing(true); }}
+          >
+            {t('workspace.rename')}
+          </button>
+          <button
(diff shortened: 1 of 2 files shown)


openwong2kim:
Merged — thank you, @matdac6! 🎉

This is a genuinely nice UX touch. You're right that the right-click menu is the first place most users look for Rename, and wiring it to the existing inline-rename flow was exactly the right call — no new rename path, just a more discoverable entry point next to double-click, `Ctrl+Shift+R`, and the tmux prefix `,` binding.

Verified before merge: `tsc` clean, CI green on both Windows and the cross-platform baseline, and the menu item opens the inline editor and commits on Enter as expected. Looking forward to more contributions!

PR #227, opened 2026-06-14
fix(pty): emit OSC 7/7727 with [char]27 so the prompt hook works on PowerShell 5.1

snowyukitty:
## Problem

The PowerShell prompt-integration hook builds its **OSC 7** (cwd) and **OSC 7727** (git branch) sequences with the backtick-e (`` `e ``) / backtick-a escape-character escapes. Those only exist in PowerShell 6+.

Under **Windows PowerShell 5.1** — the default `powershell.exe` on Windows — `` `e `` is *not* an escape; it collapses to the literal string `e`. So instead of a real ESC-introduced OSC sequence, every prompt render emits visible text:

```
e]7;file://HOST/C:/Users/...
```

The stray glyphs corrupt the prompt and, worse, shift the cursor baseline that in-pane TUIs (codex, Claude Code) lay their UI against — so keystrokes like Ctrl+J / Backspace appear to make the cursor jump to the footer and back — and cwd reporting is broken (the URI never parses).

## Fix

Switch to `[char]27` / `[char]7`, which evaluate to ESC / BEL on **both** 5.1 and 7+, matching the OSC 133 hook in `daemon/shell-integration.ts`. Confirmed: under `powershell.exe` the emitted sequence now starts with byte 27, not 101 (`'e'`).

Scope: a single file — `src/main/pty/shell-hooks/pwsh.ps1` (no behavior change on PowerShell 7+, which already produced ESC from `` `e ``).

## Release Notes

* **Refactor**
  * Updated PowerShell shell hook's internal signal transmission mechanism to use explicit control character encoding, improving terminal compatibility and reliability for directory and git branch information display.

### DIFF
--- a/src/main/pty/shell-hooks/pwsh.ps1
+++ b/src/main/pty/shell-hooks/pwsh.ps1
@@ -27,19 +27,31 @@ function prompt {
     $oscPrefix = ''
     if (-not $script:__wmux_skip_osc) {
         try {
+            # ESC / BEL as [char] codes, NOT the PowerShell `e / `a escapes.
+            # The `e escape (escape char) only exists in PowerShell 6+. Under
+            # Windows PowerShell 5.1 "`e" is the literal two-char string and
+            # collapses to "e", so the prompt would emit visible `e]7;...` text
+            # instead of a real OSC 7 sequence — a garbled prompt plus broken
+            # cwd reporting, and the stray glyphs throw off the cursor baseline
+            # that in-pane TUIs (codex, Claude Code) render against. [char]27 /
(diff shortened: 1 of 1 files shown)


openwong2kim:
@snowyukitty Thanks for the great contribution! You pinpointed exactly why the prompt was breaking on PowerShell 5.1 — `e doesn't evaluate to ESC there. Switching to [char]27/[char]7 keeps PS 7+ behavior unchanged and stays consistent with the OSC 133 hook, so it's a clean fix. The thorough root-cause analysis and verification notes made the review really easy too. Much appreciated!

PR #230, opened 2026-06-14
feat(x6): resume the exact Claude session on every pane after a reboot

openwong2kim:
## What

X6 ③ — resume-by-id. After a daemon restart or OS reboot, wmux revives the
*exact* Claude conversation that was running in each pane (`claude --resume <id>`,
with the captured permission mode), not a fresh shell or a fuzzy `--continue`. And
it now works for **every** pane that ran Claude, not only the one whose live
startup banner happened to be caught.

## Why

The reboot-resume "▶ Resume Claude" pill reliably appeared on only one pane. An
eng-review found the data model (`ptyId` as the join key) and the display path are
sound — this was purely a *capture* problem with two independent fragilities:

1. The pill's sole gate, `lastDetectedAgent`, is written only when the live
   `AgentDetector` catches Claude's startup banner. That is once-per-session and
   restored scrollback is never re-fed, so a pane whose banner streamed before the
   daemon attached never re-arms it after a reboot.
2. The exact-uuid `resumeBinding` had no durable fallback, and its hook routing
   collapsed every pane in a workspace to the active surface (`WMUX_SURFACE_ID` is
   never injected — the renderer mints a surface only *after* `pty.create`). Split
   / shared-cwd panes clobbered each other's binding.

## How

The `ptyId` join key is unchanged. Three capture-reliability rungs:

- **Per-pane routing.** The daemon stamps `WMUX_PTY_ID` (its own session id) into
  each pane's env at spawn — the one per-pane key that reliably reaches the child
  shell and the hook bridge. The bridge echoes it; `resolvePtyIdForSignal` trusts
  `signal.ptyId` when it maps to a live pane *and* belongs to the claimed
  workspace, falling back to workspaceId/cwd otherwise.
- **Second pill writer.** `daemon.setResumeBinding` also arms `lastDetectedAgent`
  (a captured hook proves the pane ran Claude), so a banner-missed-but-hook-landed
  pane shows the pill with its exact binding instead of nothing.
- **Durable spool.** On a failed capture RPC (main pipe down at boot,
  no-workspace-match, timeout) the bridge appends a self-describing record to
  `~/.wmux/resume-spool/<ptyId>.json` (pipe-free, last-write-wins). The daemon
  ingests it on recovery and a 60s timer, attributing by exact `ptyId` with the F7
  cwd-match, D5 transcript-existence, and a stale-clobber guard. Disk
  cwd-inference is deliberately *not* used — it mis-attributes under a shared cwd.

This branch also contains the earlier X6 ③ steps: the id-aware resume builder +
permission-mode mapping, the hook → main → daemon binding capture across the
SIGKILL-survival path, the progressive resume pill, and the D5 existence-probe.

## Verification

- `tsc` daemon/root/cli/mcp clean; full unit suite **3256** (+12 new).
- New `scripts/x6-multipane-resume-dogfood.mjs` **11/11** on the real daemon
  bundle — two shared-cwd panes each resume their **own** uuid via the spool (pane
  B never inherits pane A's conversation); a banner-less pane still gets the pill;
  the D5 / F7 / stale-clobber guards hold.
- Existing `scripts/x6-resume-binding-dogfood.mjs` **6/6** (no regression).
- Bridge resume-capture 4/4 + env-capture 4/4 + spool producer.
- Codex review: **PASS** (no P1); 4 P2 hardenings applied and re-reviewed clean.

## Remaining

Deterministic verification is complete. The live packaged-build reboot visual
check (the "▶ Resume Claude" pixels rendering across all workspaces) is the one
manual step left, since a packaged renderer cannot be store-injected.

## Release Notes

* **New Features**
  * Resume now persists exact session bindings for more reliable recovery, including optional permission-mode restoration.
  * Multi-pane resume routing is improved with per-pane identifiers to reduce mix-ups in shared/cwd scenarios.
  * The resume pill now follows a progressive, permission-aware flow and can offer exact `--resume <id>` when context matches.

* **Documentation**
  * API reference updated to include the new `daemon.setResumeBinding` RPC method.

* **Tests**
  * Added/expanded automated coverage and verification scripts for resume binding durability, routing guards, and replay behavior.

### DIFF
--- a/docs/api/reference.md
+++ b/docs/api/reference.md
@@ -26,7 +26,7 @@ returns `EPERM`. Wire framing: newline-delimited JSON, one object per line.
 
 ## RPC methods
 
-Total: **99** methods (`ALL_RPC_METHODS` in
+Total: **100** methods (`ALL_RPC_METHODS` in
 `src/shared/rpc.ts`). Capability and risk class are read from
 `src/main/mcp/methodCapabilityMap.ts`:
 
@@ -208,6 +208,7 @@ Total: **99** methods (`ALL_RPC_METHODS` in
 | `daemon.compact` | `wmux.internal` |  |
 | `daemon.superviseRearm` | `wmux.internal` |  |
 | `daemon.superviseStop` | `wmux.internal` |  |
+| `daemon.setResumeBinding` | `wmux.internal` |  |
 
 ### `hooks`
 
--- a/integrations/claude/bin/wmux-bridge.mjs
+++ b/integrations/claude/bin/wmux-bridge.mjs
@@ -119,6 +119,13 @@ function spoolResumeBinding(record) {
     if (!safe) return;
     const dir = getResumeSpoolDir();
     const file = join(dir, `${safe}.json`);
+    // Deliberately a SHARED per-pane temp (not a unique name): a crashed bridge
+    // leaves at most ONE orphan `<pane>.json.tmp` that the next write overwrites,
+    // whereas unique temps would accumulate unboundedly. Two concurrent same-pane
+    // hooks (e.g. Stop + SubagentStop) can race here — last rename wins and the
+    // loser may log ENOENT — but the daemon ingest is the real ordering authority
+    // (it skips a spool whose session already matches or whose ts is older), so a
+    // racy spool is reconciled correctly, never mis-applied (codex P2).
     const tmp = join(dir, `${safe}.json.tmp`);
     writeFileSync(tmp, JSON.stringify(record), { encoding: 'utf8', mode: 0o600 });
     renameSync(tmp, file);
--- a/integrations/claude/bin/wmux-bridge.mjs
+++ b/integrations/claude/bin/wmux-bridge.mjs
@@ -19,7 +19,7 @@
 // Codex review 2026-05-22 P0 #2: bridges must be JS-only.
 // Codex review 2026-05-22 P0 #4: token is read from disk, not env.
 
-import { readFileSync, existsSync, mkdirSync, appendFileSync, statSync, openSync, readSync, closeSync } from 'node:fs';
+import { readFileSync, existsSync, mkdirSync, appendFileSync, statSync, openSync, readSync, closeSync, writeFileSync, renameSync } from 'node:fs';
 import { homedir, userInfo } from 'node:os';
 import { join } from 'node:path';
 import { createConnection } from 'node:net';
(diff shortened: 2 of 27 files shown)


openwong2kim:
Addressed all CodeRabbit findings in e0265ca (verified: tsc clean, full unit 3257, multi-pane dogfood 11/11, single-session 6/6):

- **mergeResumeBinding sticky leak** — sticky `permissionMode`/`transcriptPath` now only carry forward for the same conversation (agent+sessionId+cwd); a fresh SessionStart in a reused pane no longer inherits the prior conversation's bypass/transcript. +test.
- **durableChange ignores transcriptPath** — now persists on `transcriptPath`/`agent` change too, so the SessionStart→Stop transcript fill is saved before a reboot (one-time transition, no per-turn amplification).
- **spool drained after exec replay** — added `readResumeSpoolMap()` (read, no consume) + a shared `spoolRecordToBinding()`; `resumeLaunchCommand` now consults the spool so an exec/supervised pane with a spool-only binding replays as `--resume <id>`, not `--continue`.
- **recoveredResumeBindings not cleared on death** — `session:died`/`session:destroyed` now clear it too, preventing a stale binding on session-id reuse.
- **shared spool temp** — unique temp name (pid+uuid) + skip-rename-if-existing-newer; daemon prunes abandoned `*.json.tmp` on ingest.
- **Pane.tsx blanket toLowerCase** — lowercases only a leading Windows drive letter now, matching the daemon's normalizeCwd (POSIX case-sensitivity preserved).
- **dogfood timeout-and-proceed** — `waitForExitOrKill` force-SIGKILLs on timeout instead of leaving a possible orphan daemon (both harnesses).

openwong2kim:
@codex

openwong2kim:
Addressed all 6 Codex findings in 7ef41c4 (verified: tsc clean, full unit 3259, multi-pane dogfood 11/11, single-session 6/6):

- **relay failure loses binding** — `setResumeBinding` now spools from **main** when the daemon relay can't land (daemon down/restarting); symmetric with the bridge's spool for the main-down case.
- **stale capture overwrite** — `setResumeBinding` ignores a capture whose `ts` is older than the existing binding (mirrors the spool ingest guard the live path lacked).
- **SessionStart fallback id clobber** — a provisional SessionStart capture (no transcript yet → #12235-unsafe `payload.session_id`, no `transcriptPath`) no longer overwrites an existing transcript-derived binding for a different session (live + spool paths).
- **supervised restart drops binding** — `restartSupervisedSession` carries `resumeBinding`/`lastDetectedAgent` onto the recreated meta before saving.
- **raw cwd comparison** — new shared `normalizeResumeCwd` (drive-case + trailing-slash insensitive, POSIX case-sensitive) used by the resume builder + daemon recovery/spool guards, so `D:\repo` vs `d:/repo` no longer drops to `--continue`.
- **pill agent mismatch** — the pill now requires `resumeBinding.agent === launcher`, preventing `codex --resume <claude-session-id>` from a stale hint.

Issue #163, opened 2026-06-09
Cross-workspace terminal read/write via spoofable workspace identity (input.readScreen missing ownership assert)


openwong2kim:
While checking the "buffer access is policy-gated" claim for the #15 update, I found it doesn't hold for the read path. A token-holding MCP caller can read, and through the same root cause write, another workspace's terminal. Cross-workspace isolation is otherwise something wmux enforces, so this is a real gap.

## What

`terminal_read` (RPC `input.readScreen`) can return a different workspace's terminal viewport. Two problems stack.

**1. `input.readScreen` skips the ownership assert its siblings have.** `input.send` (`input.rpc.ts:107`), `input.sendKey` (`:164`), and `terminal.readEvents` (`:208`) all call `assertWorkspaceOwnsPty` before delegating to the renderer. `input.readScreen` (`:187-189`) is a bare passthrough with no such check. The helper's own docstring (`input.rpc.ts:50-54`) says it exists to stop "any caller that learned a foreign PTY id from reading or writing that workspace's terminal", and the read handler is the one that skips it.

**2. The `workspaceId` the gate trusts is spoofable.** `terminal_read` resolves it via the weak `requireWorkspaceId()` (`index.ts:317`), which falls back to the `WMUX_WORKSPACE_ID` env hint (`index.ts:140-142`) for any caller whose PID isn't in the live map, i.e. an external Claude Code launched outside a wmux PTY. The hint is accepted as long as the named workspace merely exists (`isLiveWorkspace !== 'absent'`), with no binding to the caller. Both the renderer check (`useRpcBridge.ts:715-728`) and `assertWorkspaceOwnsPty` only verify that `ptyId` belongs to `workspaceId`, which is internal consistency of two attacker-supplied values, not that the caller is entitled to that workspace.

## Repro (token-holding external MCP client)

1. Read the shared daemon token (same-user, on disk).
2. `workspace_list` returns every workspace with its ptyIds. Harvest a victim `(workspaceId, ptyId)`.
3. Set `WMUX_WORKSPACE_ID` to the victim ws. It is live, so `requireWorkspaceId()` returns it.
4. `terminal_read({ workspaceId: <victim ws>, ptyId: <victim ptyId> })`. The renderer dumps the victim's viewport (`useRpcBridge.ts:731-754`).

[...]

PR #239, opened 2026-06-16
feat(a2a): same-workspace multi-agent messaging + instance-suffix propagation

openwong2kim:
## Why

A wmux workspace can **display** N collaborating agents (split panes, Fleet, Agent Teams) but they could not **address one another**. Every same-workspace path was closed:

- `a2a_task_send` to the same workspace → hard-rejected `cannot send to yourself` at *workspace* granularity, **before** the pane-level addressing PR #235 added. The exact "one workspace, many agents" case `pane_id`/`surface_id` were built for was reachable only **across** workspaces — the feature's own contract, dead on arrival.
- `terminal_send` with a sibling `ptyId` → fail-closed by the #163 hardening (correct as a security boundary, but it left zero sanctioned intra-workspace channels).

This is a CORE dependency for S-C2 (Agent Teams / multi-model compare-execute), which is *by definition* same-workspace-multi-pane. Found during live v3.5.0 dogfood (4 agents in one workspace, no working path to greet each other).

## What

Two tracks, two commits.

### Track 2 — `WMUX_DATA_SUFFIX` propagation (`fix(pty)`)
A child PTY (pane shell + any agent/MCP/CLI it spawns) did not inherit `WMUX_DATA_SUFFIX`, so it computed `getPipeName()` with an **empty** suffix and connected to the **production** control pipe — even inside an isolated dev/dogfood instance. A dogfood agent could drive the user's **real** workspaces. The blanket `WMUX_*` strip (built to stop identity spoofing) took the isolation suffix as collateral.

- `resolveSpawnEnv`: capture the suffix from the **spawning process's own env** before the strip, re-apply last (like identity). Covers local PTYManager **and** the daemon-via-main path.
- `DaemonSessionManager`: force the child's suffix to the **daemon's own** suffix, with a **delete-on-unset** branch — closes the recovery-replay vector where a persisted/hand-edited `session.env` suffix survives `stripReservedAuth` and points a recovered pane at a different instance's pipe.

Net **security-positive**: closes a sandbox→prod leak rather than opening one.

### Track 1 — same-workspace pane-to-pane A2A (`feat(a2a)`)
The self-guard was workspace-granular; it should mean "same **pane**", not "same **workspace**", once addressing exists.

- `a2aAddressing.decideSameWsSend` (pure, unit-tested): cross-ws unchanged · same-ws **no address** → reject (ambiguous, would loop to own active pane) · same-ws resolving to the sender's **own pty** → reject (true self-send = bracket-paste + forced submit into your own prompt = loop) · same-ws sibling with a **verified** sender pty → loud paste · same-ws sibling with an **absent** sender pty → deliver but **suppress** the paste.
- MCP self-pty capture (KS-1): `lookupPidMapWorkspace` reads the ptyId anchor from `a2a.resolve.identity` `entries[]`, cached on a verified hit and threaded as `senderPtyId`. It is **absent on the env-hint fallback — the common case** — which is precisely why the renderer fails closed on the *paste* (silent delivery) instead of trusting an absent guard. The task is still persisted + on the EventBus, so a sibling gets it via `a2a_task_query`; a self-addressed send is at worst a no-op pointer, never a loop.
- KS-2: same-ws task **replies** also suppress the paste (reply-to-`from` has no pane address → active-pane delivery could loop). Same-ws reply/status authz stays workspace-granular pending the per-pane `from.ptyId` model — documented, not a cross-ws regression.
- KS-3: the execute-approval dialog is **honest** when sender ws == receiver ws (an agent spawning an autonomous `bypassPermissions` agent in its own workspace).

The cross-workspace **fail-closed boundary (#163) is untouched**: `resolvePaneAddress` only ever searches the target's own tree, and the cross-ws path skips the same-ws gate entirely.

## Design provenance
Two independent adversarial reviews (Opus 4.8) hardened the design before implementation:
1. **senderPtyId is absent in the dominant pid-map-miss/env-hint path** → the guard can't be the only defense; the same-ws silent-delivery default is what actually closes the loop.
2. **Recovery replay survives a forged suffix** (`stripReservedAuth` keeps non-auth `WMUX_*`) → the daemon force **with the delete-on-unset branch** is required, not defensive garnish.

## Out of scope (S-C2 follow-up)
The full per-pane `from.ptyId` model: symmetric reply pinning, role-per-pane, pane-granular reply/status authz.

## Verification
- `tsc --noEmit` (main/daemon/mcp) = 0 · full unit suite **3432 + 16 runtime** green · API reference drift none · lint clean (pre-existing baseline only).
- **Live dogfood 10/10** in a single isolated packaged instance (`scripts/same-ws-a2a-dogfood.mjs`): same-ws A→B sibling send succeeds + pins to pane B (same-ws task) · true self-send + no-address rejected · senderPtyId-absent sibling falls back to silent · cross-ws back-compat + foreign-id fail-closed.

> Merge gate: pending the author's **GUI dogfood** with real Claude/Codex agent panes (exercises the MCP process-tree `senderPtyId` capture, which the RPC harness injects directly).

## Release Notes

* **New Features**
  * Enhanced same-workspace multi-agent communication with safer pane/surface addressing and improved “silent” handling.

* **Bug Fixes**
  * Strengthened self-send detection to prevent agent-to-agent loops.
  * Improved instance isolation for development and dogfood runs.
  * Updated the execution approval dialog to clearly describe same-workspace spawn scenarios.

* **Tests**
  * Added/expanded coverage for same-workspace addressing decisions, pane identity, and spawn environment handling.

### DIFF
--- a/scripts/same-ws-a2a-dogfood.mjs
+++ b/scripts/same-ws-a2a-dogfood.mjs
@@ -1,0 +1,222 @@
+/*
+ * Live dogfood — same-workspace multi-agent A2A (Track 1).
+ *
+ * Proves the headline fix: two agent panes in the SAME workspace can now address
+ * each other via a2a_task_send (previously hard-rejected "cannot send to
+ * yourself"), while a true self-send (own pane) and an ambiguous no-address send
+ * are still rejected, and the cross-workspace fail-closed boundary is unchanged.
+ *
+ * Single ISOLATED packaged instance (out/wmux-win32-x64/wmux.exe + a unique
+ * WMUX_DATA_SUFFIX so it never touches the user's real wmux). Drives the pure
+ * main-pipe RPC (clientName omitted → grandfather), passing senderPtyId verbatim
+ * the way the MCP server would after a verified PID-map hit.
+ *
+ * Verifies over the main-pipe RPC:
+ *   - ws1 pane A → ws1 pane B (sibling surface_id + sibling senderPtyId) SUCCEEDS
+ *     and pins to.surfaceId; the task is same-ws (from.ws === to.ws).
+ *   - addressing your OWN pane (surface_id == sender, senderPtyId == that pty)
+ *     is REJECTED ("cannot send to your own pane").
+ *   - a same-ws send with NO pane address is REJECTED ("without addressing a
+ *     specific pane").
+ *   - same-ws sibling send with senderPtyId ABSENT still succeeds (silent
+ *     fallback — delivery suppressed, task persisted).
+ *   - cross-ws ws2 → ws1 paneB still SUCCEEDS (back-compat) and a cross-ws /
+ *     foreign surface_id is REJECTED (fail-closed #163).
+ *
+ * Run (PowerShell): npm run package; node scripts/same-ws-a2a-dogfood.mjs
+ */
+import { spawn } from 'node:child_process';
+import fs from 'node:fs';
+import net from 'node:net';
+import os from 'node:os';
+import path from 'node:path';
+import { randomUUID } from 'node:crypto';
+import { fileURLToPath } from 'node:url';
+
+const SCRIPT_DIR = path.dirname(fileURLToPath(import.meta.url));
+const REPO_ROOT = path.resolve(SCRIPT_DIR, '..');
+const APP_EXE = path.join(REPO_ROOT, 'out', 'wmux-win32-x64', 'wmux.exe');
+const USERNAME = os.userInfo().username || 'default';
+
+const results = [];
+let app = null;
+function check(name, ok, detail = '') {
+  results.push({ name, ok: !!ok });
+  console.log(`  [${ok ? 'PASS' : 'FAIL'}] ${name}${detail ? ` — ${detail}` : ''}`);
+  return ok;
+}
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+if (process.platform !== 'win32') { console.log('same-ws-a2a-dogfood: SKIP (win32-only)'); process.exit(0); }
+if (!fs.existsSync(APP_EXE)) { console.error(`packaged exe not found: ${APP_EXE} — run \`npm run package\` first`); process.exit(2); }
+
+const suffix = `-samewsdog${process.pid}`;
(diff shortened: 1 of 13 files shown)


openwong2kim:
@codex review

openwong2kim:
## Round-2: review feedback addressed (commit 3cfbf89)

Thanks @coderabbitai and Codex. Resolutions:

**Codex**
- **`a2a.task.update` same-ws paste (P2)** — fixed. The reply branch already suppressed same-ws delivery; the update branch did not. Now both gate delivery on `sameWsTask` (the update is still persisted + teed onto the bus → pollable via `a2a_task_query`).
- **`MY_PTY_ID` lost when a terminal tool warms the cache (P2)** — fixed. `MY_PTY_ID` is now captured inside `lookupPidMapWorkspace` on every verified hit, so the terminal-route path populates it too (previously only `resolveWorkspaceId` set it, which the cache fast-path skips).
- **Case-insensitive suffix scrub (P3)** — fixed. The daemon now deletes any case-variant of `WMUX_DATA_SUFFIX` from a replayed blob before applying its own (Windows env is case-insensitive).
- **Same-ws reply/update role (P2)** — documented as an interim workspace-granular limitation pending the per-pane `from.ptyId` model (S-C2, explicitly out of scope here). Delivery is suppressed for same-ws tasks, so only the stored history role is affected — not a cross-ws regression. Added comments at both role-derivation sites.

**CodeRabbit**
- **`senderPtyId` provenance (Major)** — hardened. `senderPtyId` is now only trusted if it resolves to a real terminal pty in the **sender's own workspace** (`isTerminalPtyInLeaves`); a bogus/foreign value is treated as absent → the safe silent fallback. Note: `senderPtyId` is set by the trusted MCP server from a verified PID-map hit and is **not** an agent-settable tool param (`sendMessageParams` has no such field), so the spoofing surface is the token-gated main-pipe path only; the silent-default remains the floor.
- **Harness flaky sleep / empty callbacks** — fixed (bounded polling for the second pane's PTY; `() => undefined`).

Tests: `isTerminalPtyInLeaves`, case-insensitive scrub, `task.update` same-ws suppression + `senderPtyId`-provenance wiring guards. Full suite **3438** green; **live dogfood 10/10** re-verified on the rebuilt package.

PR #241, opened 2026-06-16
fix(terminal): yield right-click paste to the app when it owns the mouse

snowyukitty:
## Summary

In a pane whose foreground app has enabled mouse tracking (`vim :set mouse=a`, `htop`, `tmux`, `lazygit`), a plain right-click pasted the clipboard **twice** — once via the app's own mouse handling, once via wmux's `contextmenu` paste. This makes wmux yield the right-click to the app when it owns the mouse; **Shift+right-click** forces wmux's own paste (matching Windows Terminal).

## Changes

- `useTerminal.ts`: in the `contextmenu` no-selection paste branch, bail when `terminal.modes.mouseTrackingMode !== 'none'` and Shift is not held — the app already received the right-click as a mouse event, so wmux must not also paste.
- Make Shift a true override: the pre-existing post-copy suppression window now also honours Shift (`!e.shiftKey && …`), so a deliberate Shift+right-click is never swallowed.
- Add a source-level regression lock scoped to the `contextmenu` handler, matching the `atlasClear` / `webglTeardown` locks in the same dir (the handler runs against a real xterm `Terminal` + DOM events jsdom can't faithfully drive).

### Changed

- In a pane where the foreground app has mouse tracking enabled, a plain right-click no longer pastes — use **Shift+right-click** to paste (matches Windows Terminal).

### Fixed

- Right-click no longer pastes the clipboard twice in panes whose foreground app owns the mouse (`vim :set mouse=a`, `htop`, `tmux`, `lazygit`).

## Stability tier impact

Renderer-only terminal UI behavior; no RPC / protocol / API surface changes.

n/a — no PaneMetadata / EventBus / permission / Named Pipe / RPC changes.

## Test plan

- **Unit test added:** `src/renderer/hooks/__tests__/useTerminal.rightClickPasteMouseMode.test.ts` — a source-level regression lock asserting (a) the contextmenu handler reads `mouseTrackingMode`, (b) it bails when mouse mode is on and Shift is not held, (c) Shift overrides the post-copy suppression window, and (d) the guard sits after the selection→copy branch and before the paste write (so selection-copy is unaffected).
- **Full suite:** `npm run test:parallel` → 270 files / 3412 tests passed locally. `tsc --noEmit` clean; ESLint introduces no new issues.
- **Manual:** in a pane running `vim` with `:set mouse=a`, confirmed a plain right-click no longer double-pastes; **Shift+right-click** pastes; a plain PowerShell pane (no mouse mode) still right-click-pastes exactly as before. Selection + right-click still copies.

## Related issues / PRs

n/a

## Reviewer checklist

## Release Notes

* **Bug Fixes**
  * Fixed right-click paste behavior in the terminal when mouse tracking is enabled. Shift+right-click now provides a consistent paste override option.

### DIFF
--- a/src/renderer/hooks/__tests__/useTerminal.rightClickPasteMouseMode.test.ts
+++ b/src/renderer/hooks/__tests__/useTerminal.rightClickPasteMouseMode.test.ts
@@ -1,0 +1,68 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'node:fs';
+import path from 'node:path';
+
+/**
+ * Source-level regression lock: right-click paste must yield to the foreground
+ * app's mouse mode.
+ *
+ * When a TUI enables mouse tracking (xterm mouseTrackingMode x10/vt200/drag/any,
+ * i.e. DECSET 9/1000/1002/1003 — vim `:set mouse=a`, htop, tmux, lazygit), a
+ * plain right-click is already delivered to it as a mouse event. wmux's
+ * `contextmenu` handler must therefore NOT also paste, or the single click is
+ * handled twice — the reported right-click double-paste. Shift+right-click
+ * forces wmux's own paste, matching Windows Terminal's Shift-override.
+ *
+ * The contextmenu handler runs against a real xterm Terminal plus DOM events
+ * that jsdom can't faithfully drive, so the invariant is pinned at the source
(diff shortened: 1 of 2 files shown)


openwong2kim:
Thanks for this. Deferring to the app when it owns the mouse is the right fix, and matching Windows Terminal with the Shift override keeps it predictable. Merging now.

PR #242, opened 2026-06-16
feat(identity): harden multi-agent identity & addressing

openwong2kim:
## Summary

Hardens the adjacent identity & addressing bug cluster surfaced by #239's same-workspace A2A dogfood (4 agents in one workspace had no working cross-pane paths once you looked past the headline fix). Six guards, each reusing #239's verified `MY_PTY_ID`/`senderPtyId` pattern.

> **Stacked on #239** (base `feat/same-ws-multiagent-a2a`). This code depends on #239's `decideSameWsSend` / `MY_PTY_ID` / `isTerminalPtyInLeaves`, so it cannot rebase onto `main` until #239 lands. **Re-target the base to `main` after #239 merges.**

## Guards

**P0 — terminal_send/_key self-loop.** A verified first-party caller (`senderPtyId` present) that omits `ptyId` is refused. "The active terminal" would resolve to the caller's own pane (bracket-paste loop) or, in a multi-pane workspace, a non-deterministic UI-focus sibling that `assertWorkspaceOwnsPty` cannot catch (it only blocks cross-workspace, never an intra-workspace sibling). MCP forwards `MY_PTY_ID`; main rejects via the pure `decideTerminalOmittedTarget`. External callers (no `senderPtyId`) and an explicit `ptyId` are structurally unaffected. `resolveActivePtyId` is now scoped to the caller's workspace instead of UI focus.

**P1b — `a2a_whoami` pane-level identity.** Returns the caller's own pane address (`ptyId`/`surfaceId`/`paneId`) plus that pane's detected agent via the pure `resolveSelfPaneIdentity`, scoped to the caller's ws tree so a forged/foreign `senderPtyId` degrades to the ws-level answer. Read-only, no capability — whoami output never flows into terminal routing (#163 boundary preserved).

**P1a — duplicate-name refusal.** `resolveWorkspaceTarget` refuses a duplicate EXACT name (ambiguous → error listing both IDs) instead of silently picking the first in list order. ID short-circuits; number/substring keep first-match (the documented number/partial addressing contract). Removes the dead, unimported `companyRpcHandlers.ts` duplicate.

**P2a — `surface_list`/`pane_list` caller-scoped (fail-soft).** An omitted `workspaceId` resolves to the caller's own workspace (converging with `a2a_whoami`) via the weak `resolveWorkspaceId`; a read never throws on identity miss (active fallback), while write tools keep `requireWorkspaceId`.

**P3 — `a2a.task.update` message.** A rejected transition now surfaces the allowed next states from `VALID_TRANSITIONS` (e.g. `submitted` must first move to `working`), not a bare "Invalid transition". Message-only; the transition graph and the receiver-permission gate are unchanged.

**P2b — `WMUX_DATA_SUFFIX` isolation.** Diagnostics + fail-loud (`app.exit`) when a non-empty suffix is not reflected in `userData` (setPath failed → an isolated instance would boot onto prod's `session.json` and restore the wrong workspaces). An empty suffix (normal production) never trips the guard.

## Process

Authored via plan → eng-review gate. An adversarial Opus panel (4 reviewers in REFUTE mode) materially strengthened the plan before implementation:
- **HOLE1**: mode (b) is not an "orthogonal reject" — in a multi-pane workspace it is a silent wrong-sibling write → ws-scoping had to be folded into P0, not deferred.
- **P1a**: making the number tier strict would break the documented number addressing contract → only the EXACT-name tier refuses duplicates.
- **P1b**: the "senderPtyId is unspoofable" rationale was wrong for the direct-pipe path → corrected to "whoami is capability-free and structurally separate from the #163 router".
- **P2b**: a naive fail-loud would crash every normal production boot → the guard trips only on `suffix set AND userData not suffixed`.

## Verification

- `tsc --noEmit` 0 · `eslint` 0 · `gen-api-reference --check` clean
- Unit: 3467 (parallel) + 16 (runtime)
- **Live dogfood 13/13** (`scripts/multiagent-identity-hardening-dogfood.mjs`): isolated packaged exe + main-pipe RPC reproducing P0 (omit-ptyId reject + sendKey parity + explicit bypass + external pass-through), P1b (two siblings differ; absent/foreign degrades), P1a (duplicate name → both IDs; ID-direct succeeds). P2a (MCP-server identity path) and P2b (boot launch) are covered by unit + source-invariant tests.

Plan: `plans/multiagent-identity-hardening-plan.md`.

## Summary by CodeRabbit

* **New Features**
  * Pane-level “whoami” identity addressing using the sender’s verified terminal mapping.
  * Safer A2A routing with workspace target disambiguation and workspace-scoped listing when appropriate.
  * Terminal self-loop prevention when sending to the active terminal.
* **Bug Fixes**
  * Startup now verifies session isolation suffixes are applied correctly and exits if not.
  * Clearer guidance for invalid A2A task status transitions.
* **Tests**
  * Added/expanded coverage for pane identity, workspace targeting edge cases, and data isolation validation.
* **Documentation**
  * Added a phase-based multi-agent identity hardening implementation plan.


openwong2kim:
@codex review

openwong2kim:
Addressed codex review:
- **P3 (dataIsolation `endsWith`)** — fixed in 9d45fc2: exact-path comparison (`original + suffix`) instead of a tail substring, with a regression test for the `'x'`-tails-`'wmux'` case.
- **P2-1 (stale cache) + P2-2 (external pin)** — tracked in #243. Both are fail-soft edges of the omitted-workspace read path, not regressions (pre-P2a these reads were always UI-active-scoped). The root cause is the shared `resolveWorkspaceId` fast-path / pin behavior used by every a2a tool, so a scoped read-resolver fix lands separately to avoid regressing the a2a identity path.

PR #245, opened 2026-06-16
fix(mcp): scoped read resolver for surface_list/pane_list (codex P2, #243)

### LINKED ISSUE #243: surface_list/pane_list omitted-workspace resolution: stale cache + external pin (codex P2 follow-up from #242)
codex P2 follow-ups from PR #242 (multi-agent identity hardening). **Neither is a regression** — both are fail-soft edges of P2a's caller-scoped `surface_list`/`pane_list`; before P2a these reads were always renderer-active-UI scoped. The root cause is the shared `resolveWorkspaceId`.

## P2-1: stale cached workspace on fail-soft reads
`src/mcp/index.ts` `surface_list`/`pane_list` (omitted workspaceId) call the weak `resolveWorkspaceId()`, whose fast path (`workspaceResolved && MY_WORKSPACE_ID`) returns the cached id without a liveness check. After a workspace re-mint (daemon respawn / session restore) the cached id is stale; the renderer returns `[]` for the unknown workspace rather than a stale-identity error, so `callRpc()` never invalidates the cache and these reads keep returning an empty list until some other op happens to fail and recover.

Fix direction: revalidate the cached id for fail-soft reads, or make a forwarded stale id produce an invalidating error.

## P2-2: external caller ignores the pinned route
For a confirmed-external MCP caller (no PID/env identity), `resolveWorkspaceId()` returns `''` and never consults `getPinnedRoute()`, so an omitted `surface_list` after `terminal_read()` claimed a workspace falls through to `{}` and the renderer lists the UI-active workspace — handing the external caller the wrong PTY IDs and breaking the "omit → your own workspace" contract for that path.

Fix direction: prefer the existing pin before the active-UI fallback in the fail-soft read path.

Both touch the shared `resolveWorkspaceId` fast-path / pin behavior (used by all a2a tools), so a `surface_list`-only patch is partial; this needs a scoped read-resolver design that does not regress the a2a identity path.

openwong2kim:
## codex P2 hardening (#243) — for v3.5.1

Follow-up to the codex review on #242. v3.5.0 shipped with these as a tracked follow-up (#243); pulling them forward as a hotfix.

`surface_list`/`pane_list` now resolve the caller's workspace through `resolveScopedReadWorkspaceId`:

- **P2-1 (stale cache).** After a workspace re-mint (daemon respawn / session restore), the `resolveWorkspaceId` fast path could return a stale cached id, and the renderer's `[]` for an unknown workspace never invalidated it — so a fail-soft read kept reporting an empty list until some other op happened to recover. The helper now revalidates via `isLiveWorkspace` and re-resolves clean once the id is proven gone.
- **P2-2 (external pin).** A confirmed-external caller (no PID/env identity) that claimed a workspace via `terminal_read` now has that pin preferred over the renderer's UI-active fallback, so `surface_list` reports the caller's OWN workspace, not whatever the user has focused.

Still fail-soft — degrades to `''` (renderer active-ws) on a true miss, never throws. **a2a/write tools are unchanged** (they keep `requireWorkspaceId`), so the a2a identity path cannot regress.

Closes #243.

### Verification
- `tsc --noEmit` 0
- full unit suite **3477** passed (no regression)
- workspaceRouting invariant + P2 source assertions (`isLiveWorkspace` / `invalidateWorkspaceId` / `getPinnedRoute`) — 11/11

## Summary by CodeRabbit

* **Tests**
  * Enhanced workspace routing test coverage with more precise validation checks for read tool behavior.

* **Bug Fixes**
  * Improved workspace reference resolution reliability with better stale detection and external pinning support for read operations.

### DIFF
--- a/src/mcp/__tests__/workspaceRouting.test.ts
+++ b/src/mcp/__tests__/workspaceRouting.test.ts
@@ -45,25 +45,40 @@ describe('MCP workspace routing (source-level invariants)', () => {
     return src.slice(start, next > start ? next : start + 800);
   }
 
-  it('only requireWorkspaceId() and the two fail-soft READ tools call the weak resolveWorkspaceId() — exactly three call sites', () => {
+  it('the weak resolveWorkspaceId() is called only by requireWorkspaceId and the fail-soft read resolver — exactly three call sites', () => {
     // requireWorkspaceId is the sanctioned caller for WRITE/identity tools: it
     // throws when the resolver returns falsy, so a write (browser_open, a2a_*,
     // terminal routing) never silently lands on the UI-active workspace — the
-    // exact bug this invariant guards.
+    // exact bug this invariant guards. It calls the weak resolver ONCE.
     //
-    // surface_list and pane_list are READ tools that deliberately call the weak
-    // resolver directly: caller-scoped when identity resolves (converging with
-    // a2a_whoami so a multi-agent workspace's surfaces are the CALLER's, not the
-    // GUI-focused ones), and a fail-soft active fallback on a genuine miss — a
-    // read must not throw during the boot reconcile window. They forward an
-    // explicit workspaceId once resolved, so the fallback only fires on a true
-    // identity miss. Any NEW direct call site must be reviewed against this
-    // read-vs-write split — hence the exact count.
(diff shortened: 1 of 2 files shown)


openwong2kim:
@codex review

PR #247, opened 2026-06-17
fix(pane): keep both sides visible in the terminal+browser split

openwong2kim:
## Problem
In a pane that holds **both** a terminal and a browser surface, `SplitSurfaceView`'s `hasBoth` branch lays them out in a side-by-side resizable split — but each surface gated its `display` on `surface.id === activeSurfaceId`. A pane has a single `activeSurfaceId`, so only one side was ever visible: focusing the terminal `display:none`'d the browser and vice versa. The unfocused side went blank, and the blank toggled as the user switched surfaces.

## Root cause
`isActive` conflated two concepts — **visibility** (which surface renders, correct for the stacked/tab case) and **focus** (keyboard / devtools / toolbar). In the side-by-side split both sides must stay visible, but the single-`activeSurfaceId` gate hid one. `Terminal.tsx` also tied `isVisible` (the xterm fit/render) to `isActive`, so even a shown-but-inactive terminal stayed blank.

## Fix
Decouple visibility from focus:
- `pickSplitShownSurfaces()` (pure, unit-tested) picks the shown surface per side — the active one if it is on that side, else the first — so neither side blanks.
- A new `visible` prop on `TerminalComponent` / `BrowserPanel` drives `display` (and the xterm fit) independently of `isActive`, which now only drives focus. It defaults to `isActive`, so the stacked/tab path is unchanged.

## Testing
- New `Pane.splitVisibility.test.tsx` — 7 cases pinning the invariant (focusing either side never blanks the other; multi-tab per side; stale/empty fallbacks).
- `tsc` clean · full suite **3484/3484** · no new lint findings.

## Notes
- Out of scope (separate latent bug, noted for follow-up): an `editor` surface coexisting with a terminal **and** a browser in one pane falls outside the `hasBoth` terminal/browser filters and is not rendered.

## Bug Fixes
* Fixed visibility behavior when a terminal and browser are displayed side-by-side in a split pane. Previously, switching focus between them would hide the unfocused surface. Now both surfaces remain fully visible and rendered, with only keyboard focus changing as you switch between the active surface.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to wmux are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- **Terminal + browser split no longer blanks the unfocused side ([#247](https://github.com/openwong2kim/wmux/pull/247)).** A pane holding both a terminal and a browser surface renders them in a side-by-side split, but each surface gated its visibility on the pane's single active-surface id — so focusing one side hid the other (`display:none`) and the unfocused pane went blank, toggling as you switched. Visibility is now decoupled from focus: both sides stay rendered, and the active-surface id only drives keyboard focus.
+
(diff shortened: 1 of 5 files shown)


openwong2kim:
@codex review

PR #248, opened 2026-06-17
feat(a2a): symmetric from.ptyId reply + role-per-pane + pane authz (S-C2)

openwong2kim:
## Summary

Follow-up to #239 (same-workspace agent messaging). Closes the "send AND receive" gap: a reply now returns to the exact pane that originated a task, and same-workspace history is told apart per pane.

The task address model was asymmetric — `to` carried a pane anchor (#235) but `from` did not, so a reply had no pane to return to (it fell back to the workspace's active pane, or was suppressed same-workspace) and same-workspace history role collapsed to `user` for both parties. #242 already captured and validated the sender's `senderPtyId` on the send path, then discarded it; persisting it into `metadata.from` opens symmetric reply with **no new trust surface** (the value is server-derived and re-validated against the caller's own workspace tree).

## What changed

- **P0 — capture the sender pane.** A new task stores the validated sender pane anchor in `metadata.from`, symmetric with `to`. New pure helper `resolveSenderPaneAddress` (the `a2a_whoami` pane path is refactored onto it). Absent/forged `senderPtyId` → `from` stays ws-only (no regression).
- **P1 — symmetric reply pinning + role-per-pane.** A reply destined for the original sender pins to the `from` anchor (mirror of the `to` pin), fail-closed on a lost pin. History role is computed from the caller's verified pane (`resolvePaneRole`, paneId granularity) with an exact ws-level fallback — same-workspace no longer collapses both parties to `user`.
- **P2 — pane-granular status authz.** `a2a_task_update` status transitions are restricted to the addressed receiver *pane* when a caller pane is known. MCP injects `senderPtyId` into `a2a_task_update` (agent path only).

## Adversarial corrections (locked at eng-review)

- **Worker lockout guard (P0/A5).** The headless `execute:true` worker (`ClaudeWorker`) reports `working→completed` with no `senderPtyId`, yet `to.paneId` is stored for pane-addressed tasks. Authz gates on **caller-pane absence**, never on `to.paneId` presence — an absent caller pane always falls back to ws-authz, so the worker is never locked out. Dedicated regression test + dogfood scenario.
- **Same-ws delivery is a nudge, not a paste.** A same-workspace reply/update is delivered as a one-line pointer (`deliverPtyNudge`) to the addressed sibling, suppressed entirely on no-anchor or self-loop — preserving the #239 "paste into your own prompt" loop guard.
- **Role = paneId, pin = surfaceId.** Role compares at paneId granularity (one pane = one agent identity); surfaceId is used only for delivery pinning + the self-loop check.

## Verification

- `tsc` (renderer/shared/mcp) = 0
- `gen-api-reference --check` drift = 0
- vitest: full suite + new `a2aAddressing` / `a2aSlice` matrices (incl. worker-path authz)
- **Live dogfood 24/24** (`scripts/a2a-symmetric-reply-dogfood.mjs` — isolated packaged exe + main-pipe RPC, `senderPtyId` verbatim): same-ws round-trip with from-pane capture + per-pane role, P2 sibling rejection vs addressed-pane allow, worker-path authz pass-through, and cross-ws no-regression.

## Out of scope

`a2a_task_query` role pane-filtering (changes query semantics) and pane-granular `cancel` (ws-authz already correct) — both decided out at CEO/eng review.

## Summary by CodeRabbit

* **New Features**
  * Implemented symmetric A2A replies with pane/surface pinning back to the exact sender pane and per-pane history role handling.
  * Added pane-granular authorization for A2A task status updates so only the addressed receiver pane can apply updates (sibling panes are rejected).
* **Improvements**
  * Same-workspace replies are reduced to nudge-only when the target can’t be safely verified, while self-loop protections remain.
  * Task metadata now preserves sender pane/surface anchors to improve delivery accuracy.
* **Tests**
  * Expanded automated coverage and added a Windows dogfood harness for symmetric reply/update assertions.


openwong2kim:
Addressed all three findings in `cbfca8c`:

1. **🟠 `useRpcBridge.ts` same-ws self-route (Major) — fixed.** Added a `sameWsUnverified = sameWsTask && !callerPtyId` guard to both the reply and update branches: a same-workspace delivery with no verified caller pty is now suppressed (the message is still persisted + teed onto the bus, pollable via `a2a_task_query`), mirroring `decideSameWsSend`'s absent-senderPtyId rule. This closes the self-route where an unverified receiver reply defaulted to the `'user'` role and routed the nudge to `to` (= itself), which the `selfLoop` guard couldn't catch because `callerPtyId` was empty. Cross-workspace is unaffected (the guard is gated on `sameWsTask`). A structural test asserts the guard exists in both branches, and the live dogfood still passes 24/24 — the verified-senderPtyId paths are unchanged.
2. **🟡 dogfood `sendResp` transport fallback — fixed.** `sendResp` now treats a missing handler payload as failure (`{}`) rather than reading the transport-level envelope `ok`.
3. **🟡 dogfood `wss[0]` deref — fixed.** The boot path now throws on an empty workspace list instead of dereferencing it.

openwong2kim:
@codex review

openwong2kim:
@codex Addressed the non-participant finding in `cbb8d55`.

Both the reply and update branches now **reject** a verified caller pane whose `resolvePaneRole` is null (matches neither `from` nor `to`) in a fully pane-anchored same-ws task, instead of falling back to the ws-level `'user'` role:

```ts
if (task.metadata.from.workspaceId === task.metadata.to.workspaceId
    && task.metadata.from.paneId && callerAddr && paneRole === null) {
  return { error: 'a2a.task.send: caller pane is not a participant of this task' };
}
```

Gated tightly so nothing else regresses:
- **Cross-ws** keeps the ws-level model (the whole `from` ws is the sender side) — the guard is `from.ws === to.ws`-only.
- An **unverified** caller (no `callerAddr`) still takes the existing `sameWsUnverified` suppress path, not this reject.
- A task with **no `from.paneId`** anchor is untouched (the gate requires it).
- A **status-only** update from a non-participant was already rejected by the pane-granular `updateTaskStatus` authz; this covers the message-only path.

Verified: tsc 0, vitest `test:parallel` 3500, and the live dogfood extended to **3 panes** — pane C (verified, neither `from` nor `to`) gets both its reply and its update rejected (**26/26**).

openwong2kim:
Addressed both in `0e32631` (harness-only — no production code change):
- **line 103 (`sendResp`)** — adopted the proposed diagnostic version: a missing/invalid `r.result` now returns `{ ok: false, error: 'unexpected response: …' }` instead of `{}`, and never reads the transport-level envelope `ok`.
- **line 120 (`waitRendererReady`)** — now polls until `r.result.length > 0`, so it never returns `[]` for the caller to dereference (it times out with "renderer not ready" instead).

Dogfood still **26/26**. All production code (S-C2 + the CodeRabbit/codex review fixes) is unchanged.

openwong2kim:
This line is already the diagnostic version proposed in the previous round (the line shifted 103→108 after the harness edits): a missing/invalid `r.result` returns `{ ok: false, error }` and the transport-level envelope `ok` is never read. Incremental re-anchor false positive — no change needed.

PR #269, opened 2026-06-21
feat(a2a-channels): U1 channel domain types + persistence layer

AnandSundar:
## Summary

First cut of persistent multi-party A2A channels at the company level — Slack-style rooms with scoped membership, durable history, and public-or-private visibility (immutable post-creation). This PR lands the foundation only: domain types + on-disk persistence. Channel service, IPC, and renderer UI follow in U2+.

Channels are a companion to `MessageQueue` (which still owns 1-to-1 + broadcast A2A primitives). Channel posts fan out through the same idle-targeted delivery path, but the channel owns its own state, history, and membership.

## What's in this PR

- **`src/shared/channels.ts` (NEW, 169 lines)** — Domain types: `Channel`, `ChannelMember`, `ChannelMessage`, `ChannelRecipientStatus`, `ChannelState`. Plus `EMPTY_CHANNEL_STATE` default, `canonicalizeChannelName` + `isValidChannelName`, topic + idempotency bounds, and the 7-day empty-TTL default.
- **`src/daemon/channels/ChannelStateWriter.ts` (NEW, 323 lines)** — Persistence layer mirroring `StateWriter` exactly: `saveImmediate` / `saveDebounced` / `load` / `flush` / `flushSync` / `dispose`. Same concurrency model — `AsyncQueue`-keyed async write with race recovery against a concurrent `saveImmediate`; process-exit-friendly `flushSync` drains the queue before the staged inline fallback. `load()` prunes channels empty longer than the configured TTL but **keeps channels with no `emptySince` (real but never-joined) intact** — a bug the test suite caught and that this PR fixes (see below).
- **`src/daemon/util/atomicWrite/migrate.ts`** — Registers `CHANNEL_STATE_REGISTRY` (v1, identity) so `channels.json` gets the same migration path as `sessions.json`. One-file change, additive.
- **`src/daemon/channels/__tests__/ChannelStateWriter.test.ts` (NEW, 21 tests)** — saveImmediate roundtrip, debounce coalesce + flush, dispose flushes pending, TTL prune (stale / fresh / live / custom-TTL / default-TTL), `.bak` recovery on primary corruption, prototype-pollution guard, rotation chain, `flushSync` order, `load()` recovers from malformed shape, validator rejects bad visibility/status, `EMPTY_CHANNEL_STATE` invariant.

## Bug found during dogfood (fixed in this PR)

The initial `load()` prune logic only kept empty channels if `emptySince` was set AND within the TTL. That meant a real channel that exists on disk but has never had a member join (no `emptySince`) was being **silently pruned on every load** — a P0 for the never-joined-channel case. The `.bak recovery` test caught it (the recovered channel had no `emptySince`, so it was getting dropped). Fixed by adding an explicit "no `emptySince` → keep" branch. Added a dedicated regression test (`load keeps a channel with 0 members and no emptySince`) so this can't drift back.

## Plan reference

Code references plan IDs in comments: `KTD1` (channels.json separation), `KTD2` (per-channel mutex), `KTD3` (recipient-set frozen at critical-section entry), `KTD7` (immutable visibility), `KTD8` (7-day empty purge), `KTD9` (`historyFromSeq`), `KTD10` (creator auto-joined), `R4` (state machine), `R10` (structured data), `R13` (idempotency), `R14` (per-recipient status). The full plan doc is **not** in this repo — U1 implements what the comments describe without claiming to land the rest of the plan. U2 derivation (service layer, per-channel mutex, fanout, idempotency LRU, IPC) is deferred to a follow-up.

## Test results

- `npx vitest run src/daemon/channels/` → **21/21** passed (incl. the new regression test)
- `npx vitest run src/daemon/` → **484/484** passed (no regressions in existing daemon code)
- `npx vitest run src/shared/` → **362/362** passed
- `npx tsc -p tsconfig.daemon.json --noEmit` → **0 errors**
- `npx eslint src/daemon/channels/ src/shared/channels.ts src/daemon/util/atomicWrite/migrate.ts` → **0 errors**

## Test plan for review

1. **Roundtrip + recovery**: `saveImmediate` then `load()` returns the same shape; corrupting the primary file lets `load()` recover from `.bak`.
2. **Empty-TTL prune**: a channel with `emptySince = now - 8d` is pruned under the default 7d TTL; `emptySince = now - 3d` survives; a channel with a member is never pruned by empty-TTL; a channel with 0 members and no `emptySince` survives (regression case).
3. **Debounce + flush**: `saveDebounced` doesn't write immediately; multiple calls within the 30s window coalesce to the latest; `flush()` writes pending state synchronously; `dispose()` flushes then cleans up; `flushSync()` drains the queue before falling back to inline write.
4. **Atomic write**: `rotationEnabled` produces `.bak` (and `.bak.1`) correctly; corrupt primary recovers via `.bak`; no `.tmp` residue after a successful save.
5. **Validator + proto guard**: channels with bad `visibility` / `status` are rejected (load returns empty); `__proto__` / `constructor` keys don't pollute the loaded object.

## Summary by CodeRabbit

* **New Features**
  * Added channel state persistence with automatic saving, debounced writes, and manual flush options.
  * Implemented empty-channel cleanup based on configurable retention (default: 7 days).
  * Added shared channel domain model and channel-name normalization/validation utilities.
  * Added support for channel visibility (public/private) and status (active/archived).
* **Security & Reliability**
  * Hardened persisted state loading with strict schema validation and safer JSON parsing (prototype-pollution resistant).
* **Tests**
  * Expanded test coverage for persistence, recovery, pruning/TTL edge cases, debouncing/flush behavior, and channel-name rules.

### DIFF
--- a/src/daemon/channels/ChannelStateWriter.ts
+++ b/src/daemon/channels/ChannelStateWriter.ts
@@ -289,6 +289,7 @@ export class ChannelStateWriter {
       if (typeof ch !== 'object' || ch === null) return false;
       const c = ch as Record<string, unknown>;
       if (typeof c['id'] !== 'string') return false;
+      if (!isSafeObjectKey(c['id'])) return false;
       if (typeof c['companyId'] !== 'string') return false;
       if (typeof c['name'] !== 'string') return false;
       if (c['visibility'] !== 'public' && c['visibility'] !== 'private') {
@@ -318,6 +319,21 @@ export class ChannelStateWriter {
       if (list.length === 0) continue;
       if (!isValidChannelMessageRow(list[0])) return false;
       break;
+    }
+    // Reject members/messages/idempotency keys that name the prototype
+    // chain. The JSON.parse guard upstream normally strips these, but we
+    // double-check so a custom-parsed file (e.g. from a future migration
+    // step) cannot smuggle `__proto__` past validation.
+    for (const key of Object.keys(obj['members'] as Record<string, unknown>)) {
+      if (!isSafeObjectKey(key)) return false;
+    }
+    for (const key of Object.keys(obj['messages'] as Record<string, unknown>)) {
+      if (!isSafeObjectKey(key)) return false;
+    }
+    for (const key of Object.keys(
+      obj['idempotency'] as Record<string, unknown>,
+    )) {
+      if (!isSafeObjectKey(key)) return false;
     }
 
     return true;
@@ -397,13 +413,39 @@ function isValidChannelMessageRow(row: unknown): boolean {
   );
 }
 
+/**
+ * Returns false for object keys that name the well-known JS prototype
+ * chain — `__proto__`, `constructor`, `prototype`. Used to guard
+ * `pruneKeys` and validator map lookups against a corrupt file that
+ * could otherwise leak prototype references into the running process.
+ */
+function isSafeObjectKey(s: unknown): s is string {
+  return (
+    typeof s === 'string' &&
+    s !== '__proto__' &&
+    s !== 'constructor' &&
+    s !== 'prototype'
+  );
+}
+
+/**
+ * Build a new record containing only the keys in `survivors`. Returns a
+ * null-prototype object and reads with own-key checks, so a corrupt
+ * `rec` with `__proto__` as a literal own property cannot pollute
+ * `Object.prototype`.
+ */
 function pruneKeys<T>(
   rec: Record<string, T>,
   survivors: Set<string>,
 ): Record<string, T> {
-  const out: Record<string, T> = {};
+  const out = Object.create(null) as Record<string, T>;
   for (const id of survivors) {
-    if (id in rec) out[id] = rec[id];
+    if (
+      Object.prototype.hasOwnProperty.call(rec, id) &&
+      isSafeObjectKey(id)
+    ) {
+      out[id] = rec[id];
+    }
   }
   return out;
 }
--- a/src/daemon/channels/ChannelStateWriter.ts
+++ b/src/daemon/channels/ChannelStateWriter.ts
@@ -1,0 +1,323 @@
+// ─── ChannelStateWriter ────────────────────────────────────────────────────
+// Persists ChannelState (channels.json) to disk using the shared atomic-write
+// helpers in `../util/atomicWrite`. The public API mirrors `StateWriter`
+// (saveImmediate / saveDebounced / load / flush / flushSync / dispose) so
+// future waves can layer behaviour without changing call sites.
+//
+// Concurrency model is identical to StateWriter's:
+//   - `saveImmediate` is synchronous and remains so — emergency-exit
+//     paths (SIGINT/SIGTERM/etc.) rely on it running inline. Before
+//     writing it clears any queued async write so a stale debounced
+//     snapshot cannot overwrite the newer immediate one.
+//   - `saveDebounced` funnels through an AsyncQueue keyed
+//     `'channel-state'` so only one async write is ever in flight.
+//     Repeated debounced calls coalesce to the latest snapshot.
+//   - `flushSync` drains the queue by invoking the registered sync
+//     fallback (used by process-exit handlers where the event loop
+//     has stopped).
+//
+// Plan reference: U1 (channel domain types and persistence layer).
+
+import path from 'node:path';
+import {
+  atomicReadJSONSync,
+  atomicWriteJSON,
+  atomicWriteJSONSync,
+  createMigrator,
+  CHANNEL_STATE_REGISTRY,
+} from '../util/atomicWrite';
+import { AsyncQueue } from '../util/AsyncQueue';
+import {
+  CHANNEL_EMPTY_TTL_HOURS_DEFAULT,
+  EMPTY_CHANNEL_STATE,
+  type ChannelState,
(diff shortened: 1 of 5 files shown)


openwong2kim:
Thanks for this — genuinely strong first contribution, and it's well aligned with where the project is heading.

The design choices are senior-level: ordering by per-channel seq instead of timestamps (with the right rationale about same-ms posts in one mutex window), idempotency with an LRU memory bound, the memberName snapshot to avoid stale-name drift, the frozen recipient set, and especially the epoch-fenced race recovery between saveImmediate and the debounced write in ChannelStateWriter. Splitting channels.json from sessions.json for blast-radius isolation, plus finding and regression-testing the never-joined-channel prune bug during dogfood, are exactly the instincts we want. Really nice work.

On direction: we're building a cross-PC A2A transport (agents on different machines talking over the LAN, local-first). Channels and that transport are two axes of the same thing — who's in the conversation vs where it can reach. Your channels layer is the room/persistence half; the transport is the cross-machine half. They compose into cross-PC rooms, which is exactly the product we want.

For U2, the one thing that protects that composition: please build channel fanout/delivery through a delivery (transport) abstraction rather than hardcoding the local idle-targeted path. If "deliver this post to member X" goes through an interface, the LAN transport slots in later as just another transport and rooms span machines for free — channels never needs to know about the LAN, and the transport never needs to know about rooms. I'll share the transport-side interface as it firms up.

I also ran an independent second-model review (Codex) over the diff and verified its findings. Two are worth acting on:

Should fix before merge:

isChannelState is too permissive (the isRecordOfArrays / isRecordOfRecords helpers). typeof [] === 'object', so an array passes as a valid map — isRecordOfArrays([]) === true — and the inner member/message/idempotency rows are never shape-checked. A subtly-corrupt channels.json then loads as "healthy" and gets silently reshaped by pruneKeys() instead of falling back to .bak, defeating the recovery path that's the whole point of the atomic-write design. Cheap fix: reject arrays (Array.isArray(v) → false) in both helpers and spot-check the nested rows the way you already do for channels[].
Design question worth resolving:

The empty-channel GC has a blind spot that's the flip side of the prune bug you fixed: a channel with 0 members and no emptySince is kept forever. Correct for a never-joined channel, but one that had members, went empty, and lost its emptySince (a crash between leave and persist, or a legacy record) is now immortal and leaks. Worth deciding: fall back to createdAt when emptySince is missing, or track a "was ever joined" bit to tell the two cases apart.
Small U1 polish (non-blocking):

canonicalizeChannelName can produce a name isValidChannelName rejects (leading hyphen, or >64 chars), so they don't quite compose. Either make canonicalize guarantee a valid result (strip leading hyphen, clamp to CHANNEL_NAME_MAX) or document the normalize-then-validate contract. Also CHANNEL_NAME_MAX and the regex bound {0,63} duplicate the limit.
ChannelMessage.deliveryStatus's comment references a recipientSnapshot field that isn't on the struct yet. Add it or soften the comment to "U2 attaches per-recipient status."
For U2 (not now): saveImmediate swallowing write errors matches StateWriter, so it's fine here — but on the U2 post path, a failed persist of a durable-history post should surface rather than silently diverge from memory.

CI needs a maintainer approval to run (first-time-contributor gating) — I'll kick that so we get the build/test signal. Thanks again, great foundation and exactly the direction we want.

AnandSundar:
@openwong2kim Thank you for reviewing my PR. I will push an update soon with the changes you recommended.

openwong2kim:
Thanks for this, @AnandSundar — really clean foundation. 🙏

Reviewed the full diff and verified locally (separate worktree):
- `tsc -p tsconfig.daemon.json` → 0 errors
- `vitest run src/daemon/channels/ src/shared/__tests__/channels.test.ts` → 54/54 passing
- `eslint` on all changed files → 0 errors
- CI green across the board (validate, bench, Baseline on macOS/Ubuntu/Windows, CodeRabbit)

A few things I appreciated:
- `ChannelStateWriter` mirrors `StateWriter`'s concurrency model faithfully (AsyncQueue keying, epoch-based race recovery against `saveImmediate`, `flushSync` drain-then-inline order).
- Strong prototype-pollution defense in depth — `isSafeObjectKey` guards on both validation and `pruneKeys`, plus the `Object.create(null)` build.
- The `emptySince ?? createdAt` fallback for the "lost emptySince" recovery case is a nice catch, and the dedicated regression test keeps it from drifting back.
- `channels.json` separated from `sessions.json` so channel loss can't cascade into session failure — exactly the right boundary.

This lands as foundation-only (domain types + persistence, not yet wired into the daemon), so it's low-risk. It dovetails with the LanLink direction — channels (topology) over LanLink (transport) is where these converge.

One housekeeping note: the branch is a bit behind `main` (predates #270), but there are no conflicts. Squash-merging. Looking forward to U2 (service layer, per-channel mutex, fanout, IPC). 🚀

AnandSundar:
@openwong2kim Thank you for being so quick to respond to my PR and merging it with main. I appreciate it.

openwong2kim:
Hey @AnandSundar, awesome foundation! 🎉 Any rough timeline on when U2 (service layer, IPC, fanout, renderer UI) might land? No pressure — just curious what to expect. Thanks!

AnandSundar:
@openwong2kim I started working on U2. It might land today or tomorrow. :)

PR #280, opened 2026-06-21
feat: A2A Channels U2 — Service, IPC, and Renderer UI

AnandSundar:
# feat: A2A Channels U2 — Service, IPC, and Renderer UI

Implements U2 of the A2A Channels feature on top of the U1 persistence layer shipped in PR #269. Closes the maintainer's two U2 directives from [PR #269#issuecomment-4761325754](https://github.com/openwong2kim/wmux/pull/269#issuecomment-4761325754):

1. **Build channel fanout/delivery through a `ChannelDelivery` interface abstraction**, not a hardcoded local idle-targeted path. The local transport (`LocalPtyDelivery`) ships now; a LAN transport slots in later as a sibling.
2. **Surface `saveImmediate` errors on the U2 post path**. The post path now returns `{ ok: false, error: { code: 'PERSIST_FAILED', ... } }`; the renderer turns this into a toast and an inline composer error.

## Review fixes (this update)

The first pass of this PR drew ~30 findings from CodeRabbit and Codex, which the PR author folded into four tiers (🔴 must-fix, 🟠 should-fix, 🟡 nice-to-fix, plus nits). **This update lands all four 🔴, all four 🟠, all six 🟡, and most of the in-scope nits** in nine focused commits pushed on top of the original implementation. The diff is +1,500 / -100 lines on the renderer, daemon, MCP, pipe, and docs layers.

### Critical fixes (the four 🔴 — user-facing send path was dead in production)

| Unit | Commit | Req | Finding | Fix |
|---|---|---|---|---|
| U1 | `6259b13` | R1 | `dc.rpc('channel.<method>', …)` calls at the pipe handler hit a `Unknown method` error at the daemon — the daemon registers `a2a.channel.*` | Re-prefixed all 9 RPC method names to the `a2a.channel.*` namespace |
| U2 | `b3a1556` | R2 | `channel.message` was broadcast as a bare object; the main-side `DaemonClient.handleControlMessage` only switches on the `DaemonEvent` envelope shape, so every event was silently dropped | Wrap the broadcast in `{ type, sessionId, data }`; new `wrapChannelMessageEnvelope` helper owns the wire shape; unit test pins it against drift |
| U3 | `6ca7ed9` | R3 | The renderer's `useChannelsEventSubscription` reads a global that no one installed; even when installed, the poll sent no `workspaceId`, so the per-recipient post-filter dropped every event | Install `__wmuxEventsPoll` in `useRpcBridge`, thread the `workspaceId` from `company.ceoWorkspaceId`, add the new `RPC_INVOKE` (`rpc:invoke`) IPC channel + main-side handler + preload surface |
| U4 | `8cc61ed` | R4 | `ChannelsPanel.handleCreate` and `Composer.handlePost` only mutated the slice — the slice's `*Optimistic` thunks are documented as state-mirror-only, the daemon round-trip was missing entirely | Install `__wmuxChannelsRpc` in `useRpcBridge`; the slice's `*Daemon` thunks round-trip to the daemon; `createChannelOptimistic` / `postMessageOptimistic` now return `{ ok: false, error: ChannelError }` on failure and the composer/panel render it |

### Authorization and resource-cap fixes (the four 🟠)

| Unit | Commit | Req | Finding | Fix |
|---|---|---|---|---|
| U5 | `bf3f2ab` | R5, R6 | Sender was spoofable — `ChannelService.post` checked membership against the client-claimed `params.sender.workspaceId`; `archive` performed zero authorization despite the MCP docstring claiming creator-or-CEO gating | `params.verifiedWorkspaceId` (server-pinned from the MCP layer's `requireWorkspaceId()`) is threaded through every mutating call; `post` rejects `NOT_AUTHORIZED` on `sender.workspaceId !== verifiedWorkspaceId`; `archive` rejects when the verified caller is neither `channel.createdBy` nor a company CEO |
| U6 | `837fb9c` | R7, R8 | Read methods applied no membership/visibility filter; `post` had no body clamp, no control-char strip, no cap on channel count or member count | All 4 read methods gate on the verified caller's membership; `getMessages` floors at `max(sinceSeq, viewer.historyFromSeq)` for non-public channels; `post` rejects bodies over `CHANNEL_BODY_MAX = 8192` and strips C0 control chars before the length check; `create` rejects topics over `CHANNEL_TOPIC_MAX = 256` and bounds `CHANNEL_MAX_COUNT = 1000` / `CHANNEL_MAX_MEMBERS = 256`; new `CHANNEL_DATA_MAX = 4096` cap on the `data` payload |

### Correctness fixes (the six 🟡)

| Unit | Commit | Req | Finding | Fix |
|---|---|---|---|---|
| U7 | `ea67509` | R9, R10 | Idempotency was not preserved across restart — `ChannelService` constructor never rebuilt `this.idempotency` from `state.idempotency`; `join()` did not restore `emptySince` on persist failure | Constructor hydrates `this.idempotency` with a deterministic `hydrationSeq` counter for stable LRU order; `join()` snapshots `emptySince` before delete and restores it on `saveOrFail()` failure (modeled on the symmetric `leave()` rollback) |
| U8 | `e4f2157` | R12, R13, R19, R20 | Rules-of-Hooks violation in `ChannelView` (early `return null` before `useMemo`); `channels.*` i18n keys used but not registered; `LocalPtyDelivery` missed lone `\x1b` and the `try` block didn't cover `resolveRecipient`/formatters; `Composer` `useCallback` dep was defeated by a fresh `t` each render | Hoist the `useMemo` above the early return; register the `channels.*` namespace in `en.ts` and `ko.ts`; strip lone `\x1b` bytes; widen the per-recipient `try` to cover resolution + formatters, preserving the resolved `ptyId` on the catch path so the snapshot keeps a stable surface handle; `ComposerContent` accepts `t` as a prop with an identity-function default |
| U9 | `d21f746` | R14, R15, R16, R17, R18 | Count corrections (7→9, "six" → "nine"), duplicate `vi.mock('electron', …)` block in the channel RPC test, "fan-out fan-out" typo, A2A-surface inventory table missing a description column, retention wording in `stability.md` ambiguous | All five nits fixed in a single commit |

### Per-file summary

- `src/main/pipe/handlers/a2a.channel.rpc.ts` — method-name prefix (U1)
- `src/daemon/index.ts` — envelope wrap (U2), verified-workspace threading (U5)
- `src/daemon/channels/channelEventEnvelope.ts` *(new)* — envelope helper (U2)
- `src/main/ipc/registerHandlers.ts`, `src/preload/preload.ts`, `src/renderer/hooks/useRpcBridge.ts` — bridge install (U3, U4)
- `src/renderer/stores/slices/channelsSlice.ts` — `*Daemon` thunks (U4)
- `src/renderer/components/Channels/ChannelsPanel.tsx`, `Composer.tsx` — round-trip + error surfacing (U4, U8)
- `src/daemon/channels/ChannelService.ts` — verified-workspace gate, membership/visibility gate, body/data/topic/count clamps, idempotency hydration, emptySince rollback (U5, U6, U7)
- `src/shared/channels.ts` — new size caps (U6)
- `src/renderer/components/Channels/ChannelView.tsx` — Rules-of-Hooks fix, viewer selection, composer t-prop wiring (U8)
- `src/renderer/i18n/locales/en.ts`, `ko.ts` — `channels.*` namespace (U8)
- `src/renderer/channels/LocalPtyDelivery.ts` — lone-ESC strip, wider try/catch (U8)
- `src/main/mcp/firstParty.ts`, `src/main/mcp/methodCapabilityMap.ts` — comment + typo fixes (U9)
- `docs/api/inventory.md` — A2A-surface table 3-column layout (U9)
- `docs/api/stability.md` — retention wording (U9)
- `src/main/pipe/handlers/__tests__/a2a.channel.rpc.test.ts` — duplicate mock removed (U9)
- `src/mcp/__tests__/index.channel.test.ts` — count correction (U9)

## What's in this PR

Twenty-nine commits on `feat/a2a-channels-rebased` — the original ten implementation units (U1 → U9 below) plus the nine review-fix units (R1 → R21 above). The branch is rebased on `main` at `db54e5f`.

### Original implementation (10 commits)

| Commit | Unit | Summary |
|---|---|---|
| `c860d53` | U1 | `ChannelStateWriter.saveImmediate` returns `boolean`; `StateWriter.saveImmediate` aligned for parity. |
| `c02ff7a` | U2 | `ChannelDelivery` transport seam in `src/shared/channels.ts`; `LocalPtyDelivery` wraps the existing PTY write path with the per-recipient live-TUI nudge split. |
| `0d4d634` | U3 | `ChannelService` in the daemon — per-channel mutex around `nextSeq` / recipient-snapshot freeze / idempotency LRU. |
| `b63f45f` | U3.1 | Tighten channel formatters to strip ESC + NUL (defense-in-depth on the bracketed-paste path). |
| `452b0a4` | U4 | `a2a.channel.*` pipe RPCs (9 methods); `channel.message` EventBus projection. |
| `1f7faf6` | U5 | Six `channel_*` MCP tools; `a2a.channel.read` / `a2a.channel.send` capabilities added to `FIRST_PARTY_METHODS`. |
| `aca074c` | U6 | `channelsSlice` in the renderer; `useChannelsEventSubscription` hook (events.poll loop scoped to `channel.message`). |
| `09bd54e` | U7 | `ChannelsPanel` + `ChannelItem` in the sidebar (always-visible section, active/archived grouping, unread badges). |
| `319cccf` | U8 | `ChannelView` (active-channel message list, historyFromSeq filter, per-recipient delivery indicator) and `Composer` (Enter-to-submit, Shift+Enter newline, inline + toast error surfacing). |
| `3689bc6` | U9 | JSDoc on new service methods; `PROTOCOL.md` §2.8, `api/inventory.md`, `api/stability.md` updated. |

### Review-fix commits (this update — 9 commits on top)

| Commit | Unit | Summary |
|---|---|---|
| `6259b13` | U1 (review) | Re-prefix all 9 `a2a.channel.*` RPC method names with the `a2a.` namespace. |
| `b3a1556` | U2 (review) | Wrap `channel.message` broadcast in `DaemonEvent` envelope. |
| `6ca7ed9` | U3 (review) | Install `__wmuxEventsPoll` bridge. |
| `8cc61ed` | U4 (review) | Wire renderer create/post actions to the daemon. |
| `bf3f2ab` | U5 (review) | Server-pin sender + archive authorization gates. |
| `837fb9c` | U6 (review) | Membership/visibility gate + body/data/topic/count clamps. |
| `ea67509` | U7 (review) | Hydrate idempotency on startup + restore emptySince on join rollback. |
| `e4f2157` | U8 (review) | Renderer-side polish (Rules-of-Hooks, viewer selection, i18n, ESC stripping, try/catch widening, mount-gate test). |
| `d21f746` | U9 (review) | Comment and doc nits. |

### Original R1–R29 (preserved)

- **Service layer** (R1–R8): ✅ `ChannelService` is the only writer; per-channel mutex serialises `nextSeq` / recipient freeze / idempotency LRU; `saveImmediate` returns boolean and `PERSIST_FAILED` is surfaced verbatim.
- **Idempotency & concurrency** (R9–R11): ✅ LRU cap = 1000 per channel; eviction under the mutex; persisted as `ChannelState.idempotency`.
- **IPC & protocol** (R12–R15): ✅ Nine `a2a.channel.*` RPCs, granular `a2a.channel.read` / `a2a.channel.send` capabilities, `channel.message` event with multi-party scoping documented in `PROTOCOL.md` §2.8.
- **Transport abstraction** (R16–R18): ✅ `ChannelDelivery` interface in `src/shared/channels.ts`; local transport implemented; LAN deferred.
- **Renderer UI** (R19–R23): ✅ `channelsSlice`, always-visible sidebar section, `ChannelView` + `Composer`, unread badge in `MiniSidebar`. (R22 multiline via `formatBracketedPastePayload` is plumbed but the post path is single-line in v1; deferred per scope.)
- **MCP tools** (R24–R26): ✅ Six `channel_*` tools; `isError: true` MCP responses on `PERSIST_FAILED`.
- **Observability & policy** (R27–R29): ✅ JSDoc on new files; no new `vi.useFakeTimers()` usage; no literal hex colors in renderer (theme tokens only).

### Review R1–R21 (this update — all addressed)

- **R1 (🔴)** RPC method-name prefix at the call site — **fixed (U1 review)**
- **R2 (🔴)** `DaemonEvent` envelope wrap on `channel.message` broadcast — **fixed (U2 review)**
- **R3 (🔴)** `__wmuxEventsPoll` bridge install + per-recipient `workspaceId` — **fixed (U3 review)**
- **R4 (🔴)** Renderer create/post round-trip to the daemon — **fixed (U4 review)**
- **R5 (🟠)** Sender-pin authorization on `post` — **fixed (U5 review)**
- **R6 (🟠)** Archive authorization (creator or company CEO) — **fixed (U5 review)**
- **R7 (🟠)** Membership/visibility filter on reads + `historyFromSeq` floor — **fixed (U6 review)**
- **R8 (🟠)** Body/data/topic/count clamps + C0 control-char strip — **fixed (U6 review)**
- **R9 (🟡)** Idempotency preserved across restart — **fixed (U7 review)**
- **R10 (🟡)** `emptySince` restored on `join()` persist failure — **fixed (U7 review)**
- **R11 (🟡)** Composer surfaces `PERSIST_FAILED` via toast + inline error — **fixed (U4 review)**
- **R12 (🟡)** Rules-of-Hooks fix in `ChannelView` — **fixed (U8 review)**
- **R13 (🟡)** `channels.*` i18n keys registered in `en.ts` + `ko.ts` — **fixed (U8 review)**
- **R14 (nit)** Count corrections (7→9, "six" → "nine") — **fixed (U9 review)**
- **R15 (nit)** Duplicate `vi.mock('electron', …)` block removed — **fixed (U9 review)**
- **R16 (nit)** "fan-out fan-out" typo → "fan-out" — **fixed (U9 review)**
- **R17 (nit)** A2A-surface inventory table 3-column layout — **fixed (U9 review)**
- **R18 (nit)** `CHANNEL_EMPTY_TTL_HOURS_DEFAULT` retention wording — **fixed (U9 review)**
- **R19 (nit)** Lone `\x1b` strip + wider `try` in `LocalPtyDelivery` — **fixed (U8 review)**
- **R20 (nit)** `Composer.tsx` accepts `t` as a prop — **fixed (U8 review)**
- **R21 (nit)** `ChannelView.test.tsx` mount-gate test is now assertive — **fixed (U8 review)**

## Test results

- `npx vitest run src/renderer/` — **1376 / 1376 pass** (no regressions from review-fix commits)
- `npx vitest run src/daemon/` — **526 / 526 pass**
- `npx vitest run src/main/pipe/handlers/ src/mcp/` — **393 / 393 pass**
- New U1–U9 review-fix test files: 11 new `ChannelService` cases (visibility gate, body/data/topic/count clamps, idempotency hydration, emptySince rollback), 4 new `a2a.channel.rpc` cases (authz/sender-pin/round-trip), 3 new `LocalPtyDelivery` cases (lone-ESC, wider try/catch), and assertive `ChannelView` mount-gate tests.
- ESLint clean on every file touched by U1–U9 (review).
- TypeScript: no new errors from U1–U9 (review); pre-existing errors in `src/mcp/playwright/tools/*` are untouched per scope discipline.

## Design notes

- **Verified caller threads through `params.verifiedWorkspaceId`, not `RpcContext` (KTD1).** The MCP layer's `requireWorkspaceId()` is the verified-identity source; it derives `workspaceId` from the PID-map and throws on miss. The simplest thread is to forward `verifiedWorkspaceId` as a non-overridable input alongside the (overridable) `params.sender` from the MCP server to the pipe layer to the daemon. We do NOT extend `RpcContext` because the daemon's hot path is the service, not the router.
- **Authorization gate is server-pinned, not client-pinned (KTD2).** `ChannelService` mutating methods read `params.verifiedWorkspaceId` and compare it against the relevant trust anchor (`channel.createdBy` for archive; `params.sender.workspaceId` for post; channel membership for create/join/leave). Read methods read the same field and gate on membership + `visibility`. The MCP layer's `requireWorkspaceId()` throws on miss, so `params.verifiedWorkspaceId` is always a non-empty string by the time the daemon sees it.
- **Renderer subscription scope uses `company.ceoWorkspaceId` as the baseline (KTD3).** The renderer's `useChannelsEventSubscription` hook passes `company.ceoWorkspaceId` as the `workspaceId` on the `events.poll` call. This admits events for the renderer's own workspace but drops events for channels the renderer is a member of whose base scope is a different workspace. A full multi-workspace solution is a meaningful follow-up; the hook's header comment carries a `FIX-MULTI-WS` note.
- **Recipient-snapshot freeze (KTD3 in original).** `ChannelService.post` captures the member list at critical-section entry. A concurrent `join` that lands after the post starts does NOT retroactively change who sees the message. The frozen set lands on `ChannelMessage.recipientSnapshot`.
- **Multi-party event scoping.** `channel.message` is delivered to the sender's `workspaceId` AND every `recipientWorkspaceId` in the frozen snapshot — never a third workspace, and an **unscoped** poll receives zero `channel.message` events. Documented in `PROTOCOL.md` §2.8 and `api/inventory.md`.
- **Idempotency (R13).** Repeat post with same `(channelId, clientMsgId)` returns the original `seq` with `idempotent: true`; no second emit, no second `nextSeq` bump. LRU cap = 1000 per channel, persisted as `ChannelState.idempotency` so daemon restart preserves the cache. The constructor rebuilds `this.idempotency` from the persisted state with a deterministic `hydrationSeq` counter for stable LRU order across restarts.
- **Transport seam (R16).** `ChannelDelivery` accepts the full snapshot so transports that batch or batch-and-defer don't need a renderer round-trip. `LocalPtyDelivery` is the only shipping implementation; a LAN transport is a sibling that satisfies the same interface.
- **Composer round-trip.** Per the U7 precedent, the composer calls `postMessageOptimistic` with a synthesized `ChannelMessage` so the UI updates immediately; the authoritative row arrives via the `channel.message` event fan-out. The `a2a.channel.post` RPC round-trip is wired via the `__wmuxChannelsRpc` global installed in `useRpcBridge`; failures surface as toast + inline composer error.
- **Body clamp + C0 strip.** The post path strips C0 control chars (keeping `\n` 0x0A, `\r` 0x0D, tab 0x09) and trims whitespace BEFORE the length check — this prevents whitespace padding from bypassing the `CHANNEL_BODY_MAX` cap. The `data` payload is JSON-serialized and the result must fit in `CHANNEL_DATA_MAX = 4096` bytes; non-serializable shapes (e.g. circular references) fail with `CHANNEL_DATA_TOO_LARGE` rather than crashing the post path.

## Scope boundaries (deferred to follow-up)

- `channel.history` MCP tool — pagination / streaming shape is unsettled.
- LAN transport for `ChannelDelivery` — interface is the seam; implementation is future PR work.
- Channel invite flow — U2 ships the closed public-channel flow; private channels are creator-managed.
- `archived → active` reverse transition — archive is one-way (data-migration concern).
- Reactions / threads / file attachments / rich-text posts — all deferred; `data?: unknown` is plumbed for forward-compat.
- Virtualized message list — U2 ships a flat list; virtualization lands when channel usage data justifies it.
- Channel rename / topic edit — not exposed via RPC; the underlying `Channel` row is mutable on the daemon side.
- Member removal by the creator — `ChannelService.leave` is the only path.
- `historyFromSeq` mutation mid-channel — only `include_history: false` at join time is supported.
- Multi-workspace renderer subscription (KTD3 in the review-fix plan) — the renderer subscribes for `company.ceoWorkspaceId` only; subscribing for all joined channels' member-workspaces is a meaningful follow-up.
- Two `deliver()` seams to reconcile — `ChannelDelivery.deliver(message, snapshot)` in `src/shared/channels.ts:262-272` vs the LanLink PR-4 `Deliverer.deliver(msg, recipient)`. Different interfaces, different `DeliveryResult` types. **OUT OF SCOPE** — author defers to U3.5.
- The six missing KTDs (4, 5, 6, 11, 12, 13) — deferred to the maintainer.

## Open Questions

1. **KTD4, KTD5, KTD6, KTD11, KTD12, KTD13 are not on disk.** U2 ships against the ten known constraints (KTD1, KTD2, KTD3, KTD7, KTD8, KTD9, KTD10, R4, R13, R14). The missing six are silent holes — the maintainer should provide them before this PR merges, or U2 ships with an explicit "ship on 10 known" note (this PR description serves that role).
2. **Multi-workspace renderer subscription.** The review-fix U3 lands `company.ceoWorkspaceId` as the `workspaceId` on the renderer's `events.poll`. The full multi-workspace story (channels the renderer is a member of whose base scope is a different workspace) is a follow-up; the gap is documented in the `useChannelsEventSubscription` header with a `FIX-MULTI-WS` note. The author accepts this gap as out-of-scope for this PR.
3. **Two `deliver()` seams to reconcile.** `ChannelDelivery.deliver(message, snapshot)` in `src/shared/channels.ts:262-272` vs the LanLink PR-4 `Deliverer.deliver(msg, recipient)`. Different interfaces, different `DeliveryResult` types. The author defers reconciliation to U3.5.
4. `Channel.list` metadata vs. messages — U2 ships metadata only. Inline-message summaries in the list would change the perf shape of the MCP tool and the slice's `refreshChannels` cost.
5. `recipientSnapshot` deserialization shape — U2 uses `ChannelRecipientStatus` (the existing shape). The `deliveryStatus` enum is `pending | delivered | target_gone`. If the maintainer wants `failed` or `retrying`, this is a one-line type addition.

## Verification

- All 10 original U1–U9 units + all 9 review-fix units land in focused commits on `feat/a2a-channels-rebased`, pushed to the `AnandSundar/wmux` fork.
- `npx vitest run` green across `src/daemon/`, `src/renderer/`, `src/main/pipe/handlers/`, `src/mcp/`.
- `npx tsc -p tsconfig.{daemon,main,mcp,renderer,cli}.json --noEmit` — pre-existing errors in `src/mcp/playwright/tools/*` are unrelated to U2; everything U2 touched is clean.
- `npx eslint` clean on new and modified files.
- A review of the PR diff against the Requirements table confirms every original R1–R29 has a corresponding change and every review R1–R21 is addressed.
- The six missing KTDs are explicitly acknowledged above (Open Question 1).
- The multi-workspace subscription gap is explicitly acknowledged above (Open Question 2).
- The `deliver()` seam reconciliation is explicitly acknowledged above (Open Question 3).

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added A2A channels end-to-end: channel create/list, membership management, archiving, and message posting with idempotent `clientMsgId` handling and multi-recipient fanout.
  * Introduced `channel.message` events with correct multi-party visibility; unscoped polls return zero results.
  * Added Channels UI: sidebar, active-channel drawer, composer, archived badge, and unread indicators with safe PTY delivery.
* **Documentation**
  * Updated protocol and public API/stability docs for the channel surface and event scoping rules.
* **Tests**
  * Expanded end-to-end RPC/event, sync behavior, and UI coverage.
* **Chores**
  * Improved synchronous state write success reporting and aligned related tests/wiring.


openwong2kim:
@codex review plz

openwong2kim (commented):
# PR #280 — A2A Channels U2 review (draft for posting)

Thanks for this — the bones are right. `ChannelService`'s per-channel mutex + persist-then-rollback, the read/send capability split, and the no-PTY-paste rule for displayed channel text all hold up. The issues below are wiring bugs and missing server-side enforcement, not a design problem: fix them and the live fan-out works and is safe same-machine. Happy to pair on any of these.

I ran a multi-dimension review (capability/trust, deliver-seam, service logic, renderer). The unit suite is green, but the **live path has never run end-to-end** — four separate breaks each stop it (the renderer never calls the daemon on send, the main→daemon method names are wrong, the daemon→renderer envelope is wrong, and the receive bridge is never installed), and the tests mock past all of them. That's also why the authz gaps below weren't caught.

codex and CodeRabbit independently flagged many of the same issues — the envelope, idempotency hydration, archive authz, and the conditional hook converge across all three reviews. I've folded their catches in below; the framing here is the synthesis.

## 🔴 Must-fix before merge — the live channel path is dead in production

**1. RPC method names are missing the `a2a.` prefix** — `src/main/pipe/handlers/a2a.channel.rpc.ts:47-93`
Each handler calls `dc.rpc('channel.list', …)` etc., but `DaemonClient.rpc()` sends the method verbatim and the daemon only registers `a2a.channel.*` (`src/daemon/index.ts:1479-1551`). Every call returns `Unknown method: channel.list`. The mocked `DaemonClient` in the unit tests hides it.
**Fix:** call `dc.rpc('a2a.channel.list', …)` for all 9. Add one integration test that drives a real daemon round-trip so the method-name contract is pinned.

**2. `channel.message` is broadcast as a bare object, not the DaemonEvent envelope** — `src/daemon/index.ts:2168-2174`
The emit sink calls `pipeServer.broadcast(event)` with top-level `channelId/seq/sender/...`, but the wire contract is `DaemonEvent = {type, sessionId, data}` and `DaemonClient` reads `event.data` (`src/main/DaemonClient.ts:638`) — which is `undefined` here, so `DaemonNotificationRouter` drops every message. `tsc` can't catch it (broadcast/`event.data` are untyped).
**Fix:** wrap at the emit site, mirroring the LanLink nudge at `index.ts:2196`:
`emit: (event) => pipeServer.broadcast({ type: 'channel.message', sessionId: '', data: event })`. Add one test that feeds the actual JSON `ChannelService.emit → broadcast` produces through `DaemonClient.handleControlMessage` and asserts an EventBus entry appears.

**3. The renderer subscription bridge is never installed + the poll is unscoped** — `src/renderer/hooks/useChannelsEventSubscription.ts:63-117, 127`
`window.__wmuxEventsPoll` is referenced only inside this hook — nothing installs it (the comment says `useRpcBridge` mounts it, but it doesn't). At runtime the hook `console.warn`s and bails, so no `channel.message` is ever polled. Even if installed, the poll passes no `workspaceId`, and `events.rpc.ts:121` does `if (!caller) return false` → every channel event is dropped.
**Fix:** install `__wmuxEventsPoll` in `useRpcBridge` (or call the `events.poll` IPC path directly), and pass the renderer's workspace identity into the poll params so `caller` is defined. Note the renderer is multi-workspace, so a single `workspaceId` isn't enough — inject the member-workspace set and match per-recipient. Add a renderer test asserting `appendMessageFromEvent` fires on a delivered message.

**The renderer also never calls the daemon on send (create/post/list are optimistic-local only)** — `ChannelsPanel.tsx:437,462`, `Composer.tsx:217`
Creating a channel, posting from the composer, and the sidebar catalog only mutate the local Zustand slice (a synthesized `ch-local-*` row / message); none call `a2a.channel.create/post` or `a2a.channel.list`. So even with the three breaks above fixed, nothing the user does reaches the daemon and nothing loads from it on mount — the panel is a local-only mock. (codex flagged each of these separately.)
**Fix:** wire the create/post actions to the `a2a.channel.*` RPCs (optimistic update + reconcile on the real result), and load the catalog via `a2a.channel.list` + members on mount.

## 🟠 Should-fix before merge — capability gate authorizes *that* you may call, but the service then trusts caller-asserted identity

These are same-machine-scoped today (gated by `a2a.channel.send/read`, both same-user capabilities), so the blast radius is intra-machine. But they become real confidentiality/integrity holes the moment the LanLink cross-PC `deliver()` seam lands (U3.5+), and two of them silently violate a documented contract. The root cause is the same: the verified caller identity is dropped at the RPC layer and never reaches the service, so the service trusts `params`.

**4. Sender is spoofable** — `src/daemon/channels/ChannelService.ts:404-418`
`post()` checks membership against `params.sender.workspaceId`, supplied by the caller; the pipe + daemon handlers forward it verbatim (`index.ts:1545`, comment at `1475-1478` admits "no per-call identity context"). Any `a2a.channel.send` holder can post as any workspace, and the fan-out carries the forged sender to every member.
**Fix:** server-pin the sender from the verified RpcContext (same `requireWorkspaceId` identity the `a2a.task.send` path uses), and reject any `post/join/leave` whose claimed `workspaceId` ≠ verified caller.

**5. `archive` performs zero authorization** — `src/daemon/channels/ChannelService.ts:257-276`
The MCP docstring (`src/mcp/channels.ts:218`) states "caller is the creator (createdBy) or the company CEO," but `archive()` only records `archivedBy` and never checks it. Any `a2a.channel.send` holder can archive any channel (one-way, makes it read-only for everyone).
**Fix:** reject with a typed `NOT_AUTHORIZED` unless the server-pinned caller == `channel.createdBy` (or CEO).

**6. Reads apply no membership/visibility filter** — `src/daemon/channels/ChannelService.ts:173-192`
`list/get/getMessages/getMembers` never consult membership or the persisted `visibility`/`historyFromSeq`. Any `a2a.channel.read` holder can read full history + roster of any channel, including private ones they never joined; the `includeHistory:false` join option is cosmetic (never applied on read).
**Fix:** thread the verified caller into the read methods; reject `NOT_A_MEMBER` for private-channel non-members; floor returned messages at `max(sinceSeq, historyFromSeq)`. (Public read-without-join can stay if intended.)

**7. No body clamp / sanitize / resource caps** — `src/daemon/channels/ChannelService.ts:419-434`
`post.text`, `create.topic`, `post.data` are persisted + broadcast with no length bound or control-char stripping (MCP schema is bare `z.string()`), and there's no cap on messages/members/channels — `messages[]` grows unbounded and `saveImmediate` rewrites the whole file (fsync) every post, so cost is O(history)/post. A single agent can write a multi-MB body to `channels.json` and fan it out.
**Fix:** add `CHANNEL_BODY_MAX`, clamp + strip C0 (except `\n\t`) at the top of `post()`, mirror `.max()` in the MCP schema, and add a per-channel message cap + `CHANNEL_MAX_COUNT`/`CHANNEL_MAX_MEMBERS`.

## 🟡 Nice-to-fix

- **Idempotency isn't preserved across restart** (`ChannelService.ts:159-168, 444-447`): the constructor never rehydrates `this.idempotency` from the persisted map, so a replayed `clientMsgId` after a daemon restart duplicates the post. Rebuild the LRU from `state.idempotency` in the constructor.
- **Two `deliver()` seams** (`src/daemon/lanlink/deliver.ts` vs `src/shared/channels.ts`): the LanLink track defined/owns `Deliverer.deliver(msg, recipient)` (PR-4, roadmap §11 — header literally says "channels U2 fanout calls deliver(msg, recipient)"), but U2 added a parallel, incompatible `ChannelDelivery.deliver(message, snapshot)` (renderer-side `LocalPtyDelivery`, currently dead code), and there are now two exported `DeliveryResult` types with different shapes. Direct `broadcast` for same-machine fan-out is fine for U2, but before U3.5 we should reconcile these so U3.5 wires one seam — let's align on which one (I'd keep the daemon-side `Deliverer` so a `transport:'lanlink'` recipient can route through the peer's durable inbox).
- **Renderer i18n keys missing**: `channels.*` / `sidebar.channelUnreadCount` aren't in `en.ts`/`ko.ts`; because `t()` returns the key on a miss, the `?? 'fallback'` never fires and the UI shows raw keys like `channels.title`. Add the keys.
- **Rules-of-Hooks**: a conditional `useMemo` after an early return in the channel view can crash on toggle — hoist it above the return. (codex + CodeRabbit also caught this.)
- **`join` doesn't restore `emptySince` on persist failure** (`ChannelService.ts:300-301`): an empty channel clears `emptySince` before the synchronous save, but the failure rollback only pops the member — leaving the channel wrongly non-empty so the reaper won't reclaim it. Restore `emptySince` in the rollback. (codex + CodeRabbit both caught this.)
- **`post` can never surface `PERSIST_FAILED`** (`channelsSlice.ts:254`): the optimistic action always returns `{ok:true}`, so once the send RPC above is wired, a daemon `PERSIST_FAILED` still won't reach the caller — thread the real daemon result through. (CodeRabbit.)

## ✅ Solid
Capability gating (9 RPCs mapped, read/send split correct, `firstParty` bounded to the bundled MCP server, correctly NOT `wmux.internal`); the mutation core (mutex, persist-rollback, frozen recipient snapshot); and the no-PTY-paste invariant for displayed text. Good foundation — the above is finishing it, not rebuilding it.

openwong2kim:
Incredible turnaround — all of these in an hour, and the wiring (R1–R4), clamps (R8), idempotency/emptySince (R9/R10), and the renderer polish all hold up end-to-end on re-review. Really nice work.

One thing I'll take off your plate: the R5/R6/R7 sender/archive/read gates are correct *server-side*, but `verifiedWorkspaceId` is still client-supplied on the non-MCP paths — the daemon pipe handler and the main RpcRouter handler read it straight from `params`, and `RpcContext` carries no per-call verified workspace identity. So a caller can set both `sender.workspaceId` and `verifiedWorkspaceId` to the same target and the pin passes (the bundled MCP path is safe because it overwrites with `requireWorkspaceId()`). The real fix is threading a verified caller identity through `RpcContext`, the way `a2a.task.send` server-pins `senderPtyId` — and that's RPC-infrastructure on our side, not something this PR should have to carry.

So I'll handle the verified-identity plumbing as maintainer commits on top of your branch — your work stays as-is and you keep the credit. Nothing here is a skill gap; it's core trust-model wiring that's ours to own. I'll post the approach before I push so you can see the shape and weigh in.

Thanks again for how fast and thoroughly you turned the review around.

AnandSundar:
@openwong2kim Thank you for your comments and for reviewing my PR. I'll leave the rest in your capable hands. It was fun building this feature. :)

PR #284, opened 2026-06-22
chore(release): v3.9.0 — agent channels with a verified sender (#280)

### LINKED ISSUE #281: perf bench: 1-pane inputLatency flaky in CI (0 samples / pin priming fails) — harness, not a regression
## Summary

The `bench` CI job fails deterministically on the **1-pane `inputLatency`** scenario:
`perf-compare` reports `inputLatency.echoMs.p95` / `inputLatency.frameMs.p95` as
**"baseline present but current missing (scenario skipped?)"** → `FAIL`. The
**8-pane** scenario (`inputLatency8`) measures fine in the same run.

This is **not a perf regression** and **not a required check** — it's a
measurement-harness reliability bug in `scripts/perf-bench.mjs`. Filing so it's
tracked and fixed separately (it predates the channels work and was observed
flaky across prior releases).

## Evidence

From `scripts/perf-bench.mjs --mode ci` (e.g. runs 27953271609, 27954309585 — both fail identically):

```
[input 1-pane] WARNING: pin priming failed — falling back to first-echo pinning
[input 1-pane] 3 consecutive drops — resetting ptyId pin     (repeats)
[input 1-pane] zero samples — clickedBox={...} debug={"activeElement":"TEXTAREA.xterm-helper-textarea","dataEvents":2,"lastData":{"id":"pty-1","snippet":"]7;file://.../wmux-perf-3-FdA"},"pinnedId":null}
[input 1-pane] echo p50=null p95=null | frame p50=null p95=null (n=0, dropped=80)
[input 8-pane] echo p50=10.5ms p95=12.1ms | frame p50=13.7ms p95=25.5ms (n=40, dropped=0)   ← fine
```

Healthy metrics in the same run: `coldStart.firstPtyDataMs` +1.9% (PASS),
`inputLatency8` p95 within bounds, `ram.idle1Pane` **−28.7%** / `ram.panes8`
**−10.6%** (improved). So there is no input-latency or memory regression — only
the 1-pane *measurement* fails.

## Root-cause analysis

`dataEvents: 2` is the tell: the bench's `pty.onData` hook saw only **2 data
chunks during the entire ~2-minute 1-pane measurement** (the last being an OSC 7
cwd report — shell startup). The measured pane's shell emits **no echo** for the
~86 typed characters, so pin-priming and every sample drop.

Key asymmetry: the **1-pane** scenario measures the **default boot pane**
(`Terminal will self-create` its pty), while the **8-pane** scenario measures a
pane created by `pane.split` — and 8-pane echoes reliably. So the default
boot-pane's pty/echo path behaves differently from a split-created pane in
headless CI.

`pty.onData` itself is an add-listener (`ipcRenderer.on(IPC.PTY_DATA, …)`), so
the hook isn't being clobbered by the Terminal mount — the pty genuinely isn't
producing echo data on this pane.

### Hypotheses to confirm (CI-only; not locally reproducible)
1. The default boot pane's self-created pty isn't echo-ready / its data isn't
   routed to the renderer the same way a `pane.split` pty is, specifically in
   headless CI.
2. A readiness gap — measurement begins before the boot pane's shell reaches an
   interactive, echoing prompt (though 0 samples over ~2 min argues against a
   pure timing issue).

## Suggested fix direction

- Make the 1-pane `inputLatency` scenario measure a **freshly `pane.split`
  pane** (the kind 8-pane already measures reliably) instead of the default boot
  pane — or add an **echo-readiness gate** (probe + wait for a real echo) that
  re-creates the measured pane via split if the boot pane never echoes.
- Do **NOT** soften `perf-compare`'s fail-closed-on-missing behavior — it's a
  deliberate safety gate (`perf-compare.mjs:184`, "a skipped-but-expected
  scenario is a gate FAILURE, not a free pass"). The harness should reliably
  measure, not the gate accept gaps.

## Impact / priority

- Non-blocking: `main` has no required status checks, so this does not gate
  merges (e.g. #280 merged with it red, on substance — all real checks green).
- Pre-existing, unrelated to the channels feature.
- Annoying signal noise; worth fixing so `bench` is a trustworthy green/red.

openwong2kim:
Release PR for **v3.9.0**. Bundles the unreleased work since v3.8.0:

- **A2A channels — multi-party rooms with a server-verified sender (U2 + D5, #280).** Slack-style rooms for the agents in a workspace, with every channel message stamped to a workspace identity the main process resolves from the real sending pane — a forged `verifiedWorkspaceId` is rejected, private channels stay member-only. Channels by @AnandSundar; verified caller-identity hardening (D5) by the wmux team.
- **Bundled CLI stable client identity + scoped enforcer allowlist (grandfather deprecation Stage 2, #282).** Additive groundwork toward closing the legacy permission grandfather.
- **perf-bench CI harness recovery for the 1-pane inputLatency flake (#283, closes #281).** The `bench` job is now reliably green.

### Release mechanics
- Version bumped to `3.9.0` (package.json + package-lock.json).
- CHANGELOG `[3.9.0]` section added.
- API reference (`docs/api/reference.md`) already current — no drift.
- tsc green across all four tsconfigs.

After this merges, tag `v3.9.0` is pointed at the squash-merge commit to trigger the Build & Release workflow.

## Summary by CodeRabbit

* **New Features**
  * Added A2A multi-party “agent channels”/rooms with server-verified sender capability
  * Implemented membership-scoped reads for more precise channel access control
  * Improved workspace identity attribution to better protect channel mutations
* **Chores**
  * Updated the released version to **3.9.0** and refreshed API reference/version metadata

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [3.9.0] — 2026-06-23 — Agent channels, with a verified sender on every message
+
+Headline: **A2A channels** grow up. The multi-party half (U2) lands, so several agents in one workspace can talk in a shared, named room instead of only the one-to-one task messages A2A started with — and every channel message now carries a **server-verified sender** an agent cannot forge. Building on the channel domain types and persistence from U1 ([#269](https://github.com/openwong2kim/wmux/pull/269)), agents create, join, leave, post, and archive channels; the daemon pins each message's sender, each membership, and each channel's authorship to a workspace identity that the main process resolves from the *actual sending pane* rather than trusting a tag the caller put on the wire. A forged `verifiedWorkspaceId` is rejected outright — never attributed to the workspace it tried to impersonate — and private channels stay readable only to their members. Alongside it, the bundled CLI takes a stable identity so the legacy permission grandfather can start closing.
+
+### Added
+
+- **A2A channels — multi-party rooms with a server-verified sender (U2 + D5, [#280](https://github.com/openwong2kim/wmux/pull/280)).** Channels are Slack-style rooms for the agents in a workspace: a shared, named thread several agents post into, rather than the one-to-one task messages A2A began with. This release lands the multi-party operations on top of U1's domain types and persistence ([#269](https://github.com/openwong2kim/wmux/pull/269)) — create, join, leave, post, and archive — and makes **caller identity server-verified** end to end. Every mutating channel call is stamped with the workspace identity the main process resolves from the sender's real pane (`senderPtyId`), not a `verifiedWorkspaceId` the caller supplied:
+  - the daemon's `ChannelService` pins the sender on each post, the member on each join/leave, and `createdBy` on each channel to that resolved identity, so a forged sender, member, or author is impossible;
+  - the main process strips any client-supplied workspace tag and re-derives it from the owning pane, failing closed on a mutating call it can't attribute;
+  - a forged `verifiedWorkspaceId` aimed at another workspace is **rejected**, never silently attributed to the victim;
+  - channel reads are membership-scoped, so a non-member can't read a private channel's messages, and message bodies are length-clamped so an oversized post can't stall the pipe.
+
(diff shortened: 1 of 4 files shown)


AnandSundar:
@openwong2kim awesome. Love your speed

PR #299, opened 2026-06-26
fix(mcp): recover pane ptyId from WMUX_PTY_ID when the identity walk misses (WI-002)

openwong2kim:
## What

The bundled MCP server resolves its own pane ptyId by walking its OS process tree with per-hop PowerShell calls. When that walk misses, senderPtyId goes empty and same-ws pane-level A2A fails closed (the renderer suppresses the same-ws paste when it has no sender anchor). The daemon already stamps the immutable WMUX_PTY_ID on the shell env, but the server never read it.

This reads WMUX_PTY_ID as a weak senderPtyId fallback for the A2A task and terminal tools, preferring the verified PID-map walk result when present. Local-mode spawn now stamps WMUX_PTY_ID too, matching daemon mode. It also adds stderr-only resolver diagnostics.

## Provenance split (security)

The verified MY_PTY_ID still feeds `a2a.channel.*`, whose mutation authz gates on a resolvable senderPtyId and must stay unforgeable. The weak env hint is confined to the task/terminal tools, where a forged value can at worst mislabel the caller's own pane or arm a reject-only guard, all within the same-user trust ceiling the WMUX_WORKSPACE_ID hint already exposes. `a2a.channel.*` stays verified-only and fails closed on a walk miss. `senderProvenance.test.ts` locks this wiring, and is mutation-tested so an inlined weak-hint downgrade fails the test.

## Honest scope

A live investigation while validating this turned up the real shape of the multi-agent identity problem. Measured directly with a marker env plus an env-dumping stub MCP server:

- Claude Code propagates the parent env (including `WMUX_*`) to its MCP child, so this fallback and the existing `WMUX_WORKSPACE_ID` hint both reach the server. (The `pidMap.ts` comment claiming Claude does not propagate env is outdated.)
- Codex strips the env entirely, so its MCP server gets no identity env and relies solely on the flaky process-tree walk. Env-based identity does not help Codex.

So this change is a Claude-side hardening plus diagnostics, not a complete fix for the multi-agent demo. The robust fix is server-side identity correlation (resolve the pane from the pipe connection's client PID, independent of env and the child-side walk), landing in a follow-up. The diagnostics added here are what will characterize the walk-miss in production.

## Verification

- Full unit suite (4169) green; tsc clean (root + mcp configs).
- Source-invariant lock on the provenance split, mutation-tested against three downgrade variants.
- Live dogfood on the packaged bundle (`scripts/wi-002-mcp-identity-dogfood.mjs`, 10/10): the env fallback recovers the ptyId on a walk miss (before/after), and a channel mutation still fails closed with the weak hint present.
- Real-pane proof: a real wmux pane shell carries WMUX_PTY_ID, and an MCP server spawned from it recovers pane identity.

## Summary by CodeRabbit

* **New Features**
  * Improved MCP identity recovery by using a weak `WMUX_PTY_ID` hint when verified pane identity lookup misses, enabling more reliable same-workspace actions.
  * Expanded identity and routing diagnostics to report whether resolution came from verified lookup or env hint fallback.
* **Bug Fixes**
  * Fixed fail-closed behavior on Windows/daemon-mode setups by allowing safe recovery of missing pane identity while preserving verified-only security provenance.
  * Ensured local PTY sessions expose the required identity env hint to support downstream recovery.
* **Tests / Quality**
  * Added a Windows dogfood harness and new provenance/regression tests to lock in the intended behavior.

### DIFF
--- a/plans/wi-002-mcp-identity-quick-fix.md
+++ b/plans/wi-002-mcp-identity-quick-fix.md
@@ -160,7 +160,8 @@
 5. **defer:** 사이드카 파일/TTL/ACL/daemonBootId/disk-read — dogfood가 env 미전파 증명 시에만 복귀(§4 설계 보존).
 
 ### 11.3 테스트 커버리지
-```
+
+```text
 [+] src/main/pty/PTYManager.ts (create)
     └── identity에 WMUX_PTY_ID 포함 + forced-last 불변 ── [GAP→ADD] resolveSpawnEnv.test.ts / PTYManager 구조테스트
 [+] src/mcp/index.ts (resolver)
@@ -178,6 +179,7 @@
 ```
 
 ### 11.4 실패 모드
+
 | 모드 | 테스트 | 에러핸들링 | 사용자 체감 |
 |---|---|---|---|
 | env 미전파(WMUX_PTY_ID 부재) | weak='' 유닛 | degrade(현행 동일, 회귀0) | 데모 여전히 막힘 → **진단 로그가 노출** → dogfood가 잡음 → 사이드카/PROPER 에스컬레이트 |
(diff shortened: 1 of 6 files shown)


openwong2kim:
@codex review

PR #305, opened 2026-06-27
feat(channels): agent read, invite, discovery + message UX (Slack-for-agents)

openwong2kim:
## What

Turns channels from a one-shot @mention pager into something agents and humans can actually inhabit. The diagnosis (4-axis code sweep) was that channels were missing the whole "read / awareness" loop: an agent could post and be paged, but could not read the room, there was no way to assemble a room (no invite, no discovery, private channels unreachable), and the message view was raw text. The delivery model itself was fine — @mention push is exactly how Slack notifications work; what was missing was the *pull* side.

This branch fills that in. The target is "Slack for agents (+ humans)" with a pull + @mention-push attention model.

## What's in the branch

**Read — agents can catch up**
- `channel_read` MCP tool wraps the daemon's existing read path so an agent can read a channel it was never @mentioned in (the keystone). `limit` defaults at the tool layer and tail-slices to the most recent N to protect the agent's context; the daemon `getMessages` `limit` stays optional so the human view keeps full history.

**Get into the room**
- Discovery: a "Discover" group in the sidebar lists public channels you have not joined, with a Join button (the daemon already returned public channels to non-members — this is the renderer surface).
- Invite: a member can add another workspace, which is the only way into a private channel (you cannot self-join one). New `a2a.channel.invite` (daemon + RPC + capability + first-party allowlist) and `channel_invite` MCP tool. Any member may invite; the invitee gains the channel's history + live messages. Same-machine identity is forgeable (#113, accepted), so the authz is the intended model, not a hard cross-user boundary, but the grant itself is real.

**See who's there**
- The roster reads as agents + workspaces: the internal UI member id is hidden for human members (kept for real agents, for attribution), with a one-line note that membership is per-workspace.

**Sustain a conversation**
- A safe markdown subset in messages (fenced code blocks, inline code, bold) rendered as real React nodes, never an HTML sink. @mentions still highlight in plain runs; code is literal.
- Scrollback window: render the most recent page (200) with a "load earlier" affordance instead of mounting thousands of rows.
- In-channel message search: a header toggle filters the whole visible history (not just the loaded window).

**Already on the branch (earlier slices)**
- Pane-level mention addressing (mentions point at a specific agent/pane, fail-closed), P1 auto-response (the mentioned pane receives the nudge at its turn boundary), and pane self-naming (`w1-2(claude)` + GUI rename).

## Design notes

- **Pane-level *membership* was considered and rejected.** Membership/authz stays at the workspace. Per-pane authorization is theater under the same-OS-user ceiling (#113, accepted), and there is no durable per-agent identity to key it on (paneId is layout state, ptyId changes on restart). Sibling agents in one workspace are distinguished only for display + mention targeting, which already works.
- The remaining per-agent roster polish (live presence dot / MetadataStore label) is deferred because it needs a member-id ↔ pane mapping that does not exist yet.

## Verified

- Full suite: **4323 passing**, typecheck (root/daemon/mcp) 0, lint 0.
- Renderer interactions are covered by jsdom dynamic tests (functionally equivalent to app clicks): discovery Join, roster popover, markdown render, scrollback load-earlier, search type/filter.
- **Live daemon dogfood (real daemon bundle, real named pipe, token auth, 7/7):** create private channel, member invites a non-member, the invited (never-@mentioned) member reads the full history, `limit` tail-slices to the recent N, a non-member invite is rejected (private existence hidden), and a stranger read returns nothing. Run in an isolated instance (own pipe + data dir, shared token reused, not rotated) so the dev environment was untouched.

## Not in scope / follow-ups

- The one path not yet exercised live is an agent invoking `channel_read` / `channel_invite` as MCP tools from inside a pane (needs a host restart to refresh the MCP tool schema cache). The tools are thin pass-throughs over RPCs that the live dogfood proves end to end, so the residual risk is low.
- Per-agent presence / MetadataStore-label roster enrichment (needs the member-id ↔ pane mapping above).

## Summary by CodeRabbit

* **New Features**
  * Channel `@` mentions now target live panes with stable tokens and PTY-aware, pane-scoped delivery, including flush-on-stop behavior.
  * Added `a2a.channel.invite` and enhanced channel message reading with optional pagination.
  * Channels UI: **Discover** section and updated join/invite flow.
  * Channel view: richer message rendering (fenced code + inline), in-channel search, and **Load earlier**.
* **Bug Fixes**
  * Mention delivery is fail-closed; pane/PTY mismatches fall back safely and pane-scoped deduping is corrected.
  * Message history now follows `limit` tail semantics (incl. `limit=0`).
* **Tests**
  * Added/updated regression and dynamic UI coverage for pane-aware mentioning, routing, flushing, and search/scrollback.

---

## Stacked: A2A/channel reliability fixes (live-dogfooding follow-up)

Multi-agent dogfooding (agents cross-validating channel + A2A use) surfaced a cluster of mostly *silent-failure* defects — things that mis-routed or broke with no feedback to the sender. Each fix below was reviewed before it landed.

**Channels**
- Non-member @mentions were silently dropped; `post` now returns `droppedMentions` so the sender knows the ping didn't land (plus a per-post mention cap).
- Completed channel-mention tasks could resurrect — fixed in-session (idempotent create) and across a reload/restart (a durable, bounded handled-set keyed on the mention's targeted pane, so a boot replay can't re-create finished work).
- Greeting ping-pong between agents is now rate-bounded per pane (the mention stays pullable; only the auto-nudge is capped).
- A workspace no longer unread-badges its own posts (an agent posting via MCP stopped seeing itself as noise); the @-badge still fires for a genuine self-mention.
- `deliveryStatus` is real instead of stuck at `pending`: opening a channel acks receipt over the renderer-trusted, pipe-unreachable local path, flipping the recipient + message to `delivered`.
- Create-modal UX: no longer clipped, Enter creates, a failed create surfaces an error, the Create button disables while invalid/in-flight, a refresh button re-syncs the catalog, and the roster drops the meaningless "(you)" pseudo-member.

**A2A tasks**
- A task delegated to a workspace that then closes is failed (with a reason) and the sender is notified, instead of hanging in `working` forever — visible cross-process via the a2a.task event.
- Documented the task state machine on the `a2a_task_update` tool (submitted must pass through working; terminal states are final — no resurrection).
- `a2a_task_query` gained an `updated_since` cursor for incremental polling.

**Not in this branch (follow-ups)**
- Terminal copy / blank-on-restore / last-pane-close are focus-runtime issues that need live (dev) verification — tracked separately.
- A few channel UX polish items (in-flight disables on join/leave/archive, archived-composer affordance, a11y) remain.


openwong2kim:
@codex review

PR #316, opened 2026-06-29
fix(channels): catalog fan-out, delivery receipts, mention safety + humans-only kick

openwong2kim:
## What this does

Two commits on top of `main`:

1. **`feat(channels): humans-only member kick + line members icon`** — the
   roster gets a kick affordance that rides the renderer-only `channels:mutate-local`
   path (deliberately absent from the `a2a.channel.*` pipe router, so no agent
   can reach it), plus a members icon on the channel line.

2. **`fix(channels): catalog fan-out, real delivery receipts, mention + kick
   hardening`** — the batch that came out of the channels adversarial audit. The
   recurring theme was state that changed on the daemon but never reached the
   renderer, plus mentions that could land mid-turn or get lost.

## Highlights of the hardening

- **Catalog fan-out.** A new `channel.catalog` event fires on
  create/archive/join/leave/kick/invite, so every workspace's sidebar and roster
  re-sync instead of going stale. Six of seven mutations used to emit nothing.
  Public creates broadcast to all workspaces via a `'*'` recipient sentinel;
  private ones stay member-scoped, and `events.poll` re-imposes per-recipient
  scoping (same shape as `channel.message`).
- **deliveryStatus is real.** `ack` flips a message from pending to delivered
  once any recipient has it. It was vestigially pinned at pending before.
- **Mention safety.** Non-member mentions are dropped with explicit feedback to
  the sender; a busy pane holds the nudge until the agent's Stop instead of
  pasting mid-turn; boot no longer resurrects already-handled mentions; and the
  nudge no longer interpolates sender-controlled text (prompt-injection
  mitigation).
- **Kick correctness.** The actor must be a current member or the company CEO,
  and the catalog event reaches the kicked workspace so its mirror drops the
  channel.
- **Bounds.** Per-channel history is capped with tail eviction, idempotency keys
  are per-sender, and dead idempotency pointers are pruned on eviction.

## Verification

- Full suite green (`vitest run`, 4421 passed), `tsc` clean for daemon/mcp/root,
  eslint delta zero, `docs/api/reference.md` regenerated and drift-checked.
- Daemon side driven against the bundled daemon over the real pipe (create /
  post / ack / join / leave / kick / invite / archive, idempotency, sender-pin,
  mention drop): all assertions hold.
- Catalog projection, delivery receipts, and boot behavior verified live in a
  fresh build of the app over CDP. Mention routing and the busy-hold (no
  mid-turn paste) were also confirmed live with a real agent.

## Not in scope (follow-ups)

Kept deliberately out of this PR, tracked separately: binding caller identity to
the connection peer PID (raises the same-machine isolation ceiling), an
invite-accept model for non-consented membership, durable mention delivery
across a full renderer reload, and griefing rate-caps. These are design
decisions, not silent fixes.

## Summary by CodeRabbit

* **New Features**
  * Added channel roster “kick” support for authorized members (with renderer-side removal UI and toasts).
  * Introduced channel catalog activity events to improve roster visibility and event routing.
* **Bug Fixes**
  * Warn users when `@mentions` are dropped; ensure emitted mention data matches returned results.
  * Prevent cross-workspace post collisions using sender-scoped idempotency.
  * Cap per-channel message history growth and improve cache cleanup on refresh/resync.
  * Surfaced authorization failures consistently as `NOT_AUTHORIZED` (including archive eligibility and archived-channel join blocking).


openwong2kim:
Pushed one more on top (794be8c): operator-model membership.

The member panel was applying the agent self-join rule to the human GUI. From a workspace that wasn't a member you could only add yourself, and the creating workspace was hidden from the roster, so a channel you owned showed "0 members" and offered just the one workspace to add. Since the human operates every local workspace, the picker now offers every non-member workspace (public targets self-join, private ones get invited by the active member) and the roster shows all members with no privileged owner, so the count stops lying.

That left createdBy with exactly one job, the archive authz gate, so it's now metadata only. Archive becomes membership-gated and humans-only like kick: dropped from the a2a.channel.* pipe router, the channel_archive MCP tool, and the first-party allowlist. No agent can archive a channel anymore, which keeps the #280 griefing vector closed through the transport boundary. Daemon gate is member-or-CEO.

Verified: full suite 4422 passed, a fresh-daemon probe (member archive ok, non-member rejected), and a live CDP dogfood on a fresh build (owner-only channel shows 1 member, picker offers the other workspace, add works, member archive commits and moves to the archived group).

openwong2kim:
Worked through the prior-hardening findings on top (9c8ca67). Fixed 7:

- post rollback now rebuilds `state.idempotency` from the reverted map, so a `PERSIST_FAILED` can't orphan a composite key for the next save
- the A2 trim flushes immediately, so the durable cap survives a restart before the next post
- both catalog re-hydrate paths are generation-guarded against out-of-order async writes overwriting the roster with stale membership
- `postMessageOptimistic` now applies the render cap (was event-path only)
- the mention nudge re-validates the `#channel` label and falls back safely, so a malformed title can't paste raw text into the auto-submitted nudge (B7)
- documented the `'*'` broadcast sentinel in `ChannelCatalogEvent`
- `archive` + `kick` are flagged renderer-only in the generated API reference (fixed the generator, not the output, so it survives regen)

Skipped one: the CEO kick/archive wiring in `index.ts`. `ceoWorkspaceId` is intentionally `undefined` while Company mode (paid, gated off) is unwired, so the override only matters once that lands. The member gate covers the shipping default, and the same wiring lights up archive's CEO branch too whenever Company mode ships.

Full suite 4422 passed, tsc daemon+root 0, eslint 0, reference drift 0.

PR #326, opened 2026-07-03
fix(resume): Codex resume pill + exact-id session capture

openwong2kim:
## Problem

After a reboot, the resume pill over a recovered Codex pane typed `codex --continue`. Codex has no `--continue` flag, so it errors. Two causes:

- `resumeOfferForRecovered` returned `lastDetectedAgent` unfiltered, so Codex (and gemini/aider) panes got a pill even though wmux only knew how to resume Claude.
- `Pane.tsx` hardcoded the Claude `--continue` fallback in the pill.

Separately, Codex had no session-binding capture at all, so even a fixed pill could only ever do a best-effort `resume --last` — never resume the exact pre-reboot conversation.

## What this does

**Lane A — the fix (commit 1).** Generalize the resume grammar so Claude's flag form (`--continue` / `--resume <id>`) and Codex's subcommand form (`resume --last` / `resume <id>`) share one insertion path. Claude output is byte-identical. Gate the pill to agents wmux can actually resume, so no more broken-command pills for any agent.

**Lane B — exact-id capture (commit 2).** A self-contained notify bridge registered in `~/.codex/config.toml` captures the Codex resume binding (session id + cwd + rollout path) on each turn-complete, so the pill can resume the EXACT session. The daemon side is unchanged — `hooks.rpc` already builds the binding from `signal.agent` verbatim.

- `wmux-codex-notify.mjs` parses Codex's notify payload (last argv), builds an `agent.stop` AgentSignal, and sends `hooks.signal`. A 2s hard cap + always-exit-0 means it can never stall a Codex turn; it spools on RPC failure for the daemon to reconcile.
- Registration rides the existing MCP-config machinery: TOML-safe surgical write, only-if-installed, idempotent, and **skip-if-foreign** — a user's own `notify` is never clobbered (the skip is surfaced in `mcp` status, not silent).
- The bridge installs to `~/.wmux/hooks/` (version-free stable path, refreshed each boot).

## Testing

- 639 unit tests green (shared + mcp + resume durability); all 4 tsconfigs clean.
- `scripts/codex-resume-capture-probe.mjs`: 17/17 — runs the real notify script against a mock pipe and asserts the AgentSignal envelope, auth forwarding, spool-on-failure, and quiet-drop of a no-session payload.
- Live-verified against a packaged build: notify registered into the real `~/.codex/config.toml` with user content preserved, and a real notify → real wmux pipe → daemon persisted the exact-id binding.

## Follow-up

- Chain a foreign `notify` instead of skipping it (captured in TODOS.md) — deferred until a real user hits it.
- Confirm Codex's notify await/timeout model on a live interactive turn (the 2s hard cap already bounds the worst case).

## Summary by CodeRabbit

* **New Features**
  * Added Codex resume-capture via a notify-based bridge, enabling reliable session resumption and correct “resume <id> / --last” command behavior.
  * Updated resume offers to generate launcher-appropriate resume commands with safer fallbacks when exact-session bindings aren’t available.

* **Bug Fixes**
  * Made Codex notify configuration updates safer: wmux-owned installs/uninstalls are idempotent and preserve existing non-owned notify entries to avoid accidental overwrites.
  * Improved resilience when the notify bridge can’t connect immediately, so resume capture keeps working.


openwong2kim:
Thanks — both addressed in fc48c21:

- **Codex `-c` (Major):** Good catch. Scoped the SKIP_TOKENS + short-flag heuristic to the flag form (Claude); Codex keeps only its own `resume`/`exec`/`e` subcommand guard, so `codex -c model=o3` now rewrites to `codex resume --last -c model=o3` (or the exact-id form). Added a regression test. Claude output stays byte-identical.
- **`CODEX_NOTIFY_BASENAME` (nit):** Reused the shared constant in both `getCodexNotifySourcePath` and `installAndRegisterCodexNotify`.

99 unit tests green, tsc clean.

PR #351, opened 2026-07-07
fix(build): exclude MCP cache and skip native rebuild

snowyukitty:
## Summary
- Ignore generated local `mcps/` MCP tool descriptor caches.
- Keep Electron Forge/Vite packaging scoped to `/.vite` while explicitly excluding `/mcps`.
- Skip Electron native rebuild only for `node-pty`; the existing `postPackage` hook copies/prunes `node-pty` shipped prebuilds into `app.asar` and `daemon-bundle`.

## Rationale
`mcps/` is a local agent-harness cache, not source of truth. It is useful for local tool discovery, but should not be committed or shipped.

The local Windows package step was failing in Electron rebuild because `node-gyp` could not use the installed Visual Studio 18 BuildTools. `node-pty` is already handled by the app's prebuild copy path, so rebuilding it is unnecessary. This uses `ignoreModules: ['node-pty']` instead of disabling rebuild globally, so future native production modules still go through Electron rebuild.

## Verification
- PASS: `npm run package`
- PASS: `npm exec vitest run src/daemon/__tests__/StateWriter.test.ts src/daemon/channels/__tests__/ChannelStateWriter.test.ts src/daemon/__tests__/crashRestore.integration.test.ts src/renderer/stores/slices/__tests__/channelsSlice.test.ts`
- PASS: `node out\wmux-win32-x64\resources\cli-bundle\index.js mcp check --json`
- PASS: `powershell -NoProfile -ExecutionPolicy Bypass -File D:\AI_Projects\tools\guard-mcps.ps1 -Audit`
- NOTE: full `npm run test:parallel` was attempted after this amend and hit the existing parallel timeout in `src/renderer/stores/slices/__tests__/channelsSlice.test.ts` (`is reachable through the composed store via channels selector`). The focused rerun of that file passed.

## Native-module scan
Production dependencies currently expose only `node-pty` with `binding.gyp` / native install script. Other `.node` binaries found locally are dev-time bundler packages and are not production dependencies.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved app packaging to exclude the local MCP cache (`mcps/`) and avoid bundling unnecessary build/runtime items.
  * Refined packaging rules to keep only the intended build artifacts while filtering out other generated paths.
* **Chores**
  * Updated version control ignore settings to prevent local MCP cache files from being tracked.


openwong2kim:
Thanks a lot for this. The Windows packaging failure has been nagging me for a while and this is exactly the right call.

Ignoring `mcps/` in both git and the package is correct - it's a local discovery cache, never source of truth, so it has no business being committed or shipped. And scoping the native rebuild skip to `node-pty` alone is clean since the `postPackage` hook already copies/prunes the prebuilds into `app.asar` and the daemon bundle. Nothing else needs to touch the rebuild step.

CI is green across all lanes, so I'm merging. Appreciate you tracking down the node-gyp / VS toolchain side of it 🙏

PR #378, opened 2026-07-09
6g2: operator-join — human self-join into agent-made private channels

openwong2kim:
**DRAFT — trust-boundary change, needs a 3-model code review before merge.**

## Summary
Implements the operator-join trust path (design: `operator-join-design-2026-07-05.md` v1.1, already 3-model plan-reviewed): the human owner of the machine can pull themselves into a private channel that agents created, since agent-side invite of the human seat is (correctly) forbidden.

- `operatorJoin` / `operatorList` daemon methods on `ChannelService`, humans-only transport (renderer-local mutate IPC allowlist), **never** reachable from the pipe router / MCP / first-party grants — absence is pinned by boundary tests.
- `OperatorJoinParams` is exactly `{channelId, verifiedWorkspaceId}`; the human seat row is built from constants only (no caller-supplied field is ever consumed — P5 injection guard).
- On success the daemon appends **one server-issued system message** so the join is durably auditable (survives leave, visible to fan-out-missed members, and surfaces forged direct-socket joins to the human).

## Key implementation decision (worker report)
Append-only atomicity: a 2-envelope "seat then message" split can't be atomic (no compensating append if the message fails after the seat commits). So a new additive `operator-join` event **kind** carries both the seat row and the system message in **one envelope** — `1 commit = 1 envelope` invariant preserved, both effects under one fsync barrier. Legacy mode does explicit 3-part rollback (message → nextSeq → seat) mirroring `join()`/`post()`.

## Boundary tests (the security-critical ones)
- pipe-router non-registration (a forged senderPtyId still gets `Unknown method`, daemon rpc never called)
- `FIRST_PARTY_METHODS` exclusion, MCP tool-list absence (source-scanned)
- channelLocal allowlist pass + verifiedWorkspaceId stamp

## Verification (worker, then confirm in CI)
- `tsc --noEmit` clean; 376 targeted vitest passed (operator 19 + boundary/capability/events 97 + eventlog 17 + slices 243); eslint 0 new
- **Not run locally**: full `test:parallel` — the worktree had no `node_modules` (npx cache degraded mid-session). CI re-runs it.

## Review focus for the panel
1. The new `operator-join` envelope kind — additive-only compliance + replay idempotence
2. Atomic rollback (log mode single-envelope vs legacy 3-part)
3. Boundary tests actually pin agent-unreachability
4. System-message shape (`deliveryStatus: 'delivered'`, no `recipientSnapshot`) doesn't trip the wake-worker re-nudge

## Summary by CodeRabbit

* **New Features**
  * Added an operator-only view for discovering channels and joining private ones.
  * Private channels now show a clear lock state and join confirmation flow.
  * Joined-by-operator activity is now shown as a distinct system message in channel history.

* **Bug Fixes**
  * Unread counts no longer increase for system-generated operator join messages.
  * Duplicate join attempts now return a clearer “already a member” response.

* **Documentation**
  * Updated API references to include the new operator channel actions.

### DIFF
--- a/docs/api/reference.md
+++ b/docs/api/reference.md
@@ -26,7 +26,7 @@ returns `EPERM`. Wire framing: newline-delimited JSON, one object per line.
 
 ## RPC methods
 
-Total: **133** methods (`ALL_RPC_METHODS` in
+Total: **135** methods (`ALL_RPC_METHODS` in
 `src/shared/rpc.ts`). Capability and risk class are read from
 `src/main/mcp/methodCapabilityMap.ts`:
 
@@ -172,6 +172,8 @@ Total: **133** methods (`ALL_RPC_METHODS` in
 | `a2a.channel.nudgeRecorded` | `a2a.channel.send` | `a2a` |
 | `a2a.channel.unread` | `a2a.channel.read` | `a2a` |
 | `a2a.channel.purgeMembership` | `a2a.channel.send` | `a2a` |
+| `a2a.channel.operatorJoin` | `a2a.channel.send` | `a2a` |
+| `a2a.channel.operatorList` | `a2a.channel.send` | `a2a` |
 | `a2a.principal.upsert` | `wmux.internal` |  |
 | `a2a.principal.remove` | `wmux.internal` |  |
 | `a2a.principal.markStaleWorkspace` | `wmux.internal` |  |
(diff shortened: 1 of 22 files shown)


openwong2kim:
@codex review

PR #387, opened 2026-07-10
fix(security): scope events.poll private events to server-resolved identity (audit B3)

openwong2kim:
## What

Closes audit **B3** (P1): `events.poll` scoped private events (`channel.*`, `a2a.task`) purely by the client-supplied `workspaceId`, so any wire client could name a workspace and live-eavesdrop its private channels and delegated tasks — no pane identity required.

## Fix

- **Unforgeable trust signal**: new `RpcContext.firstParty`, set only via the `dispatch(request, { firstParty })` second argument — the renderer bridge and (post-review) the plugin host set it; the wire (PipeServer) never does, and a wire client putting `firstParty:true` in the request body is ignored (read from the argument, not the payload).
- **Split scope**: lifecycle events stay client-scoped (unchanged — an unscoped lifecycle poll is already a whole-machine firehose, so workspace scope was never the confidentiality boundary there). Private types (`a2a.task`/`channel.*`) are: first-party → operator/client scope (human operator model preserved); agent transport → scoped to the **server-resolved** workspace from the verified `senderPtyId` (client `workspaceId` ignored), and **fail-closed** (drop all private) when identity can't be resolved. Reuses the same identity anchor `a2a.channel.*` mutations already use.
- **MCP `wmux_events_poll`**: forwards its own PID-walk `senderPtyId`, so a legitimate agent still reads its own channels/tasks (Windows/Codex included, via server-side walk).

## Review (codex, 1R)

codex caught one regression the consumer sweep missed — **plugin-host polls** hit the agent-transport branch (no senderPtyId → all private dropped). Fixed by marking the plugin-host dispatch first-party (it's a human-approved, in-process entry point). Committed separately.

## Consumer sweep (no shipped regression)

Every `events.poll` consumer checked: MCP tool (forwards senderPtyId ✓), renderer channel subscription (first-party ✓), plugin host (first-party ✓ after review). What breaks is dev/dogfood scripts and the event-recorder example that polled channels/a2a by an arbitrary `workspaceId` — that **is** B3, so the break is the intended hardening.

## Gates

tsc 0 (changed files) · events.rpc suite 43/43 (7 new B3 cases: forged workspaceId ignored, resolved-member delivery, dual-party a2a, fail-closed, lifecycle client scope preserved, first-party operator preserved) · a2a.channel.rpc + RpcRouter suites green · test:parallel 5621 pass (2 pre-existing diff.handler Windows-symlink flakes, unrelated).

Note: true unforgeable identity (peer-PID via GetNamedPipeClientProcessId) stays deferred with the #113 ceiling track; this raises B3 from "name a workspace string" to "need a live pane ptyId", matching the B1/B2 write-forge bar and the `a2a.channel.*` precedent.

## Summary by CodeRabbit

* **Security**
  * Improved event confidentiality by resolving private event visibility from verified caller identity rather than caller-supplied workspace information.
  * Unresolved or unknown identities now fail closed for private events.
  * Lifecycle events retain their existing workspace-scoped behavior.
  * Trusted internal operations and legitimate agent event access continue to work as expected.


openwong2kim:
@coderabbitai review

openwong2kim:
Re the caching nitpick on `resolveCallerWorkspace`: keeping it uncached on purpose. The per-call renderer resolution mirrors the exact anchor the `a2a.channel.*` mutations use (same helper shape, same round-trip), so caching only the read path would give it a staleness class the write path doesn't have — pane close, ptyId reuse, and workspace moves would all need invalidation hooks inside a confidentiality boundary. The `wantsPrivate` guard already spares lifecycle-only polls, and agent poll cadence is seconds-scale (MCP tool invocations / scripted polls), so the extra hop is noise. If polling ever concentrates enough to matter, the right fix is a main-process pane→workspace mirror shared with the a2a.channel path, not a per-handler cache.

PR #389, opened 2026-07-10
fix(session): resolve session/metadata paths lazily so WMUX_DATA_SUFFIX isolation holds

openwong2kim:
## What

`SessionManager` resolves `session.json` / `metadata.json` paths **lazily** on first use instead of capturing them in its constructor.

## Why

`SessionManager` is instantiated as a module-level singleton (`session.handler.ts`), and module evaluation runs **before** `index.ts` applies `WMUX_DATA_SUFFIX` via `app.setPath('userData', …)`. The constructor-captured paths therefore pinned `session.json` and `metadata.json` to the **un-suffixed** userData dir.

Every other surface of a suffixed instance (pipes, auth tokens, daemon home, logs) is properly isolated — but its session state silently read from and wrote to another instance's directory:

- **dev (`-dev`) and packaged production share one `session.json`** — a dev run could overwrite the user's real layout, and vice-versa.
- **Any isolated instance's seeded session was invisible.** This surfaced while validating the perf-bench session seeding: a seeded `session.json` for a `-bench…`/`-probe…` suffix was never loaded (`[SessionManager] load … (no session file)` with the seed file present on disk). Prior bench session seeding (e.g. the scrollback A/B legs) was loading from the un-suffixed path, not the seeded file.

This is the same hardcoded-shared-path class as the daemon auth-token bug fixed in #325.

## Fix

Paths are resolved lazily on first use — always an IPC handler or save path, long after the suffix is applied — and cached per instance. **Unsuffixed production resolves the identical path either way, so this is a no-op for release builds.**

## Test

`src/main/session/__tests__` (reboot persistence / async save / metadata) all green; the tests mock `app.getPath('userData')` per-test, and lazy resolution reads it at call time as before.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved session file handling by resolving storage locations when needed.
  * Increased reliability when the application’s data directory changes during runtime.

### DIFF
--- a/src/main/session/SessionManager.ts
+++ b/src/main/session/SessionManager.ts
@@ -17,7 +17,23 @@ const DEBOUNCE_MS = 30_000;
 const QUEUE_KEY = 'session';
 
 export class SessionManager {
-  private filePath: string;
+  /**
+   * Resolved LAZILY (first use), never in the constructor. This class is
+   * instantiated as a module-level singleton (session.handler.ts) — module
+   * evaluation runs BEFORE index.ts applies WMUX_DATA_SUFFIX via
+   * app.setPath('userData', …), so a constructor-captured path pins
+   * session.json/metadata.json to the UN-suffixed userData dir. Every other
+   * suffixed surface (pipes, daemon home, logs, tokens) is isolated, so a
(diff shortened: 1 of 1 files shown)


openwong2kim:
@codex review

PR #390, opened 2026-07-10
perf(terminal): hidden-pane retention — zero renderer parsing for hidden workspaces (phase 3 PR-A)

openwong2kim:
## What

An opt-in setting (**Settings → "Skip hidden pane rendering (experimental)"**, default **off**, daemon sessions only) that stops the renderer from parsing hidden panes' PTY output entirely, and re-synchronizes a pane from the daemon session buffer when it's revealed or read.

This is **PR-A** of the phase-3 stutter work — the renderer half. It lands the whole correctness contract behind a flag while reusing the existing raw reconnect replay for re-sync; a follow-up (PR-B) swaps that replay payload for a daemon-side parsed snapshot over a live-pipe re-flush, without changing this protocol.

## Why

Even with the shared output scheduler (#388), hidden panes' bytes are still *parsed* eventually — and the new `hiddenFlood` bench shows that parsing total is what starves the visible pane. With 8 hidden flooders, the visible pane's frame cadence drops to ~15fps and typing echo p95 blows out past half a second.

## Measured (hiddenFlood A/B, same build, retention ON vs OFF)

| hidden 8 flood | retention ON | OFF |
|---|---|---|
| visible pane rAF p50 / p95 | **16.7 / 16.8 ms** (full 60fps, == idle) | 66.7 / 83.4 ms |
| typing echo p50 / p95 | **29.8 / 31.0 ms** | 101.6 / 526.6 ms |

hidden 4 improves rAF p50 33.4 → 16.7 ms and echo p95 208.9 → 30.8 ms similarly.

## How

- **Scheduler** (`terminalOutputScheduler.ts`): hidden bytes are queued but **never drained**. Past the 2 MB cap the backlog is discarded and the terminal marked **dirty** instead of being flushed into xterm; retained entries are invisible to the drain loop and its cadence. A foreground write releases the backlog into the normal priority catch-up (order preserved).
- **`useTerminal`**: a dirty pane re-synchronizes on reveal via the existing reconnect replay — incoming bytes are held out of xterm while the replay is in flight, then the flush-complete handler `reset()`s the stale buffer **first** and writes the held replay onto the clean one (no duplication, no early-parse race). Degrades to the last-known screen on failure/timeout; **never clears the ptyId** (a dead pane keeps its final output). A daemon replay arriving while hidden defers the same way (discard + dirty) instead of parsing the boot flood.
- **MCP reads hydrate first**: `pane.search` / `input.readScreen` re-sync a dirty pane and wait on an xterm parse barrier before scanning, so orchestrating agents never read a retention-stale buffer.

## Verification

- `scripts/phase3-retention-probe.mjs` — boots an isolated packaged instance (seeded flag), claims a hidden workspace over the main pipe, floods it, and checks the whole contract live: retention engages + hidden DOM stays unparsed, 2 MB overflow marks dirty, `input.readScreen` hydrates the dirty hidden pane, and revealing it re-syncs from the daemon and paints the true screen (asserted from the xterm buffer, since the revealed pane is WebGL-rendered).
- Unit suites: scheduler retention policy (never-drain / overflow-dirty / order), `useTerminal` wiring locks, hydrate parse-barrier contract against a real headless xterm.
- Three one-shot diagnostics mirrored into the main log (retention engaged / overflow→dirty / resync complete with byte counts) provide the fallback-rate observability the rollout (a later PR) will gate on.

## Notes

- Depends on #389 for **bench** verification (seeded session isolation); the retention code itself is independent.
- Known PR-A limits, addressed in PR-B: dirty-reveal replays the full ring through the renderer once, and the reveal-time reconnect briefly re-opens the session pipe.

## Summary by CodeRabbit

* **New Features**
  * Added an experimental setting to skip rendering output for hidden panes in daemon sessions.
  * Hidden-pane output is queued and automatically re-synchronized when the pane becomes visible.
  * Screen reads and searches now hydrate hidden panes first to return current content.
  * Added Korean and English translations, with the setting disabled by default.
  * The preference is saved and restored with session data.

* **Bug Fixes**
  * Improved recovery when hidden-pane output exceeds retention limits, preserving the last known screen content when re-synchronization fails.

* **Tests**
  * Added coverage for hidden-pane retention, hydration, resynchronization, and performance under background output.


openwong2kim:
All three addressed in e481e36:

- **workspaceSlice** — now validates `hiddenPaneRetentionEnabled` as a strict boolean, matching `a2aAutoApproveExecute`. Fair call for an experimental flag; a corrupt `"false"` shouldn't be able to flip it on.
- **en/ko copy + CHANGELOG** — reworded to "skip output parsing" (not rendering) and scoped to hidden *panes*, not workspaces. You're right that the contract is pane-level — a hidden tab inside the active workspace is retained too.

tsc clean, store suite 494/494 green.

PR #423, opened 2026-07-12
feat(diff): workspace git diff view — read-only DiffPanel for any repo

openwong2kim:
## What

A read-only git diff view for any workspace, one palette command away. **Show Git Diff** in the command palette resolves the active pane's cwd to its worktree toplevel and opens a diff tab showing every staged, unstaged, and untracked change against HEAD — reusing the existing task-review diff surface end to end.

## How

- **DiffPanel generalized, not forked**: props become a `source: { kind: 'task' } | { kind: 'workspace' }` union. All task couplings (hunk adoption, mission-channel comments, one-click PR, close) are gated to task mode; workspace mode is strictly read-only per the §6.J surface contract (read · comment · checkout only, no syntax highlighting).
- **Zero backend changes to `diff:read`**: on a regular repo with no `targetHeadOid`, `resolveTargetRepo` resolves to the repo itself and `merge-base HEAD HEAD` degenerates to HEAD, so the existing path yields exactly `git diff HEAD` + untracked synthesis. New handler tests lock this contract in (staged+unstaged+untracked, clean tree, rename+edit).
- **New `diff:resolveRepo` IPC**: normalizes an arbitrary cwd (subdirectory, linked worktree) to its worktree toplevel — required because untracked synthesis joins repo-root-relative paths against the passed path.
- Surfaces persist via the existing surface spread (new `diffRepoPath` field), dedupe per pane by repo path, and re-read on tab reactivation (derived data, no PTY).
- Drive-by: the comment 💬 emoji becomes a monochrome glyph per the design chrome rules.

## Verified

- Unit: 12 new cases across `diff.handler.test.ts` (workspace-mode reads, `resolveRepo` normalization incl. linked worktrees, fail-closed on non-git/empty args) and `surfaceSlice.test.ts` (persistence shape, per-pane dedupe, no cross-dedupe with task diffs). Full typecheck clean.
- Live dogfood on a dev instance: happy path (file tree + numstat + synthesized new-file hunk), read-only gates (no adopt button/checkboxes), non-git cwd toast, 2MB total-cap error path, and app+daemon restart restoring the tab and re-reading content.

Part 1 of the git/GitHub surface plan (workspace diff → worktree GUI → PR/comments → diff-based orchestrator questions).

## Summary by CodeRabbit

- **New Features**
  - Added a **Show Git Diff** command to the command palette.
  - Opens a **read-only** diff tab for the active pane’s Git repo, showing staged, unstaged, and untracked changes.
[...]

PR #424, opened 2026-07-12
feat(deck): Git tab with worktree GUI — list, create, open as workspace, remove

openwong2kim:
> Stacked on #423 (workspace diff view) — will retarget to main once that merges.

## What

The Command Deck grows a **Git** tab (next to Orchestrator) surfacing the worktrees of the repo behind the active pane: every worktree with its branch, folder, and locked/prunable state. Type a branch name to create a new worktree, **Open** drops it into a fresh workspace with its terminal already there, **Remove** deletes it — with git itself refusing dirty worktrees (no force option, deliberately).

## How

- New renderer-only IPC surface (`worktree:list/add/remove`, `worktree.handler.ts`) — pipe-unreachable, same trust basis as channelLocal/fanout. git is the on-disk source of truth: no cache, no persistence, daemon restarts are a non-event.
- Safety: branch names go through `validateGitRef` (flag injection, traversal, control chars); the worktree path is derived by the handler (sibling `<repo>-worktrees/<branch>` convention), never renderer-supplied; remove re-validates against the live worktree list, refuses the main worktree, and never passes `--force`.
- Repo context reuses `diff:resolveRepo` from #423. The `git()` execFile helper is promoted out of `diff.handler` into `main/git/git.ts` (no behavior change) and shared.
- New pure parser `shared/worktreeParse.ts` covers the full porcelain contract (attached/detached/bare/locked/prunable, CRLF, paths with spaces). The company WorktreeManager keeps its own parser untouched — that module is frozen and its types are module-local.
- "Open" creates a workspace whose `profile.startupCwd` is the worktree (spawn-time tolerant fallback already built in).
- Tab hides via Settings (default ON — informational surface); hiding while active snaps back to the orchestrator, mirroring the channels-tab guard.

## Dogfood catch

Live dogfooding from inside a linked worktree exposed a real bug: the calling worktree got the "main" badge (and lost its Remove) because the handler equated the caller's toplevel with the main worktree. Fixed by returning the porcelain first block as a separate `mainPath` (git contract: main is listed first) and keying both the badge and the remove guard on it — with a regression test running the whole flow from a linked-worktree context.

## Verified

- Unit: 12 parser cases (porcelain fixtures incl. locked/prunable/CRLF/spaces), 8 handler cases against real temp repos (add→list→remove roundtrip, existing-branch checkout, dirty-remove refusal with worktree preserved, arbitrary-path & main-worktree refusal, subdirectory normalization, linked-worktree mainPath regression). Typecheck clean; Deck/Channels/store suites green.
- Live dogfood on a dev instance: full list of 8 real worktrees, create at the owner-convention path, remove (confirm → gone from disk), Open-as-workspace landing the terminal in the worktree, and the main-badge fix re-verified from a linked-worktree context.

Part 2 of the git/GitHub surface plan (workspace diff → **worktree GUI** → PR/comments → diff-based orchestrator questions).


openwong2kim:
@codex review

PR #425, opened 2026-07-12
feat(deck): PR list + comments in the Git tab (gh CLI, provider-shaped)

openwong2kim:
> Stacked on #424 (Git tab + worktrees) — will retarget as the stack merges.

## What

The Git tab grows a **Pull Requests** section: every open PR of the active pane's repo with CI rollup (green/red/pending dot), draft/merged state, review decision, and relative freshness. Expanding a PR fetches its comments and reviews as one time-ordered stream — markdown rendered, review states labeled, 16KB body cap with a browser deep-link. Every row and comment is one click from the browser.

## How

- **Provider-shaped** (owner decision): a host-neutral contract — `shared/prSurface.ts` wire types + `PrProvider` interface + origin-host detection — with GitHub (gh CLI) as the v1 implementation. GitLab (`glab`, command-symmetric) becomes a later implementation, not a redesign.
- **`GhPrService` is deliberately NOT an extension of `PrStatusCache`** — that singleton serves the branch→one-PR metadata poll and its contract stays frozen. It reuses the gh conventions instead: non-interactive env, version≠auth two-step gate, ENOENT → quietly unavailable for the process lifetime, never-throw.
- **Freshness = coarse pull**: 30s TTL list cache in main + a 30s renderer poll only while the tab is mounted, plus a per-PR detail cache keyed by `updatedAt` so unchanged PRs never refetch comments. Net rate: ~2 gh calls per minute per visible tab.
- **Fail-closed surface**: gh missing / unauthenticated / non-GitHub origin / no remote each render an explicit explanation — never a silently empty section.
- Dogfood catch: bot reviewers (CodeRabbit) wrap bodies in HTML comment markers that rendered as raw text — stripped before the cap, locked by a test.

## Verified

- Unit: 14 cases — list mapping (state/draft/checks/reviewDecision, malformed filtering), comment+review merge ordering, body cap + truncation marking, HTML-comment stripping, ENOENT gate memoization, auth-failure gate, 30s TTL behavior, updatedAt detail-cache skip/refetch, stderr fail-soft, remote-host classification (https/ssh/scp forms).
- Live dogfood against this repository: both open PRs listed with CI dots, expanding #423 rendered the real CodeRabbit review comment.

Part 3 of the git/GitHub surface plan (workspace diff → worktree GUI → **PR/comments** → diff-based orchestrator questions).

PR #426, opened 2026-07-12
feat(diff): Ask the orchestrator about a hunk, context attached

openwong2kim:
> Stacked on #425 — final part of the git/GitHub surface plan. Will retarget as the stack merges.

## What

Every hunk in the diff surface (task review and workspace diff alike) gets an **Ask** action: an inline question form whose submission lands in the workspace orchestrator's chat as ONE message carrying a structured context block — repo, branch, file, hunk header, and the hunk body fenced as ```diff data — followed by the question. The deck flips to the Orchestrator tab so the answer streams in view.

## Why this shape

- **Not composePrompt**: the seam is for ambient state; context injected there never appears in the transcript, breaking the one-click-evidence rule. A single visible message keeps question and evidence together.
- **Not a channel ride**: mention→wake adds coalescer latency for what is a direct human question.
- **Fenced hunk body**: terminal/diff-derived text stays marked as data per the orchestrator grounding rules.

## Plumbing

A transient `deckSlice.pendingBrainPrompt` relay that CommanderView consumes into its existing `handleBrainSend` — fleet-context assembly, the optimistic bubble, and the model override stay in one place instead of being duplicated into DiffPanel.

Cap contract (`shared/diffAskContext.ts`, unit-tested): 8KB total; an oversized hunk drops the body entirely (paths + header survive) rather than shipping a silently half-cut diff.

Deferred by design: auto-reaction to new PR comments (EventBus → coalescer wake) — needs autonomy gating and cursor persistence, so it's a separate PR.

## Dogfood catch

The first cut used `window.prompt` (matching the existing diff-comment idiom) — live dogfooding showed Electron renderers don't support `prompt()` at all, so the question silently never happened. Replaced with the inline form and re-verified end-to-end: Ask → typed question → Orchestrator tab flip → the user bubble carries the full context block → a real orchestrator turn streamed the answer.

## Verified

- Unit: context-block assembly (fields, fencing, omission rules), cap behavior (oversized body dropped whole, paths/header/question survive), Deck suites green, typecheck clean.
- Live E2E on a dev instance as above.

Part 4 of the git/GitHub surface plan (workspace diff → worktree GUI → PR/comments → **diff-based orchestrator questions**).


openwong2kim:
@codex review

PR #442, opened 2026-07-13
feat(orchestrator): operator-assigned pane roles the brain reads

openwong2kim:
Give each agent in the Fleet roster a role (Builder / Reviewer / Tester / Planner) via a small dropdown on its row. The orchestrator sees the role in its workspace snapshot and prefers routing matching work to the matching pane. Soft guidance, not a hard gate.

## Behavior
- **Assigned role** → shows on the pane's snapshot line as `— role: Reviewer`; the system prompt tells the brain to prefer the matching existing pane, read live roles via `pane_list` before dispatching, and never change the role key itself. Precedence: explicit operator instruction > preferred role > any pane.
- **No role** → the snapshot line is unchanged (`- w1-1(claude) [Claude Code]`) and the orchestrator behaves exactly as before. Fully backward compatible.
- A role only matters on a pane running an agent; on a plain shell it is inert (the snapshot counts only agent panes).

## Design
Read path mirrors the existing `paneLabel` plumbing: main tees the role on the `pane.metadata.changed` relay + boot hydration → `useNotificationListener` → volatile `paneRole` mirror → `deckBrain.buildWorkspaceContextSummary` snapshot. Write path is a new `METADATA_SET_ROLE` IPC → `MetadataStore.set` custom deep-merge (never clobbers label or other tools' custom keys). Empty string is the unassigned sentinel (additive merge has no delete-one-key op). No mid-session push — freshness comes from the brain re-reading `pane_list` at the dispatch boundary.

`DeckFleet` row restructured so the jump button and the `<select>` are siblings (a `<select>` cannot nest in a `<button>`).

Scoped out (deferred): per-role CLI auto-launch, reusable saved team presets, custom-role combobox, hard enforcement. This design went through eng + GLM-5.2 + Codex review; see `plans/orchestrator-pane-role-presets-2026-07-13.md`.

## Verification
- typecheck clean; new tests for snapshot role injection (present / absent / empty-sentinel) and the `paneRole` mirror set/clear/close semantics.
- Live dogfood in the dev app: assigned "Reviewer" to a running `w1-1(claude)` pane via the dropdown, tasked the orchestrator with an ordinary prompt, and confirmed its actual model input carried `- w1-1(claude) [Claude Code] — role: Reviewer` and its reply described the pane as `Role: Reviewer — Code review and diff analysis` — read and used without being told to "use roles".

## Summary by CodeRabbit

- **New Features**
  - Added operator role selection for Fleet panes (Builder, Reviewer, Tester, Planner).
  - Pane roles persist with the pane and are reflected in the commander context.
  - The orchestrator preferentially routes work to an existing pane with a matching role; explicit instructions override routing, with fallback when no match exists.
  - Roles only apply when an agent is actively running in the pane.

- **Tests**
  - Added coverage for role display, persistence/clearing behavior, and role sanitization.


openwong2kim:
@codex review

openwong2kim:
Addressed the review in 999eaa9:

- **Security (found in adversarial review, not just the bots):** the role string is injected verbatim into the orchestrator's prompt snapshot, and `custom['orchestrator.role']` is writable by any `pane_set_metadata` caller (a worker pane can self-assign), with no per-value length cap on custom. A crafted role with newlines could forge fake pane lines/instructions and an oversized one could crowd out the snapshot budget. `readOrchRole` now sanitizes at the read boundary (strip control chars → spaces, cap 64) — the single chokepoint that feeds the snapshot mirror and hydration.
- **Role hydration after restart (CodeRabbit):** the mount-time snapshot pull now carries role, so a persisted role no longer stays invisible until the next metadata change. Includes panes with a label OR a role.
- **Omitted paneRole no longer clears the mirror (CodeRabbit):** only an explicit string acts ('' still clears).
- **METADATA_SET_ROLE added to handler cleanup (CodeRabbit).**

Added a `readOrchRole` sanitization test. Typecheck clean, tests green.

Issue #182, opened 2026-06-10
Toggle Zoom


Dzirik:
I may be misunderstanding the feature.

I have 5 panes.

Prefix mode works:
- Ctrl+B n
- Ctrl+B c
- Ctrl+B p

I rebound Toggle Zoom to several keys (z, m, Enter).

After pressing Ctrl+B and the bound key:
- the key is consumed
- nothing appears in the terminal
- no pane changes size

Is this expected behavior or a bug?

openwong2kim:
Thanks for the clear report — this is a **bug**, not expected behavior, and you diagnosed it precisely: the key is consumed but nothing happens.

**What was wrong:** Toggle Zoom was only half-implemented. Pressing the bound key flipped an internal "zoomed pane" flag, but *no rendering code ever read that flag* — so the keypress was swallowed (prefix mode exited) and the layout never changed. That's exactly why `n` / `c` / `p` worked (their actions are wired to actual behavior) while zoom did nothing, no matter which key you rebound it to.

**Fix** (#187): zoom now actually renders — the active pane expands to fill the whole workspace and the other panes are hidden (tmux-style), with a small `ZOOM` indicator so it's clear you're zoomed. Pressing the key again restores the exact previous split. It works with the default `z` and with any key you rebind it to.

Verified on a live build by driving the real `Ctrl+B` → `z` over a 3-pane layout: zoom fills the workspace 100% and un-zoom restores the split pixel-for-pixel. The fix is up in #187 and will go out in the next release. Thanks again for flagging it!

openwong2kim:
Fixed in #187, now merged to main. Will ship in the next release.

PR #447, opened 2026-07-14
feat(pane): per-pane header action buttons (new terminal, split, browser, zoom)

cheyras:
## What

Each pane's tab strip gains a right-aligned cluster of five icon buttons:

**new terminal · split right · split down · new browser · zoom**

Actions that were previously keyboard-only (`Ctrl+T`, `Ctrl+D`, `Ctrl+Shift+D`) become mouse-discoverable, sitting natively beside the existing per-tab close button. Tabs scroll on narrow panes while the cluster stays pinned right.

## Why

I came to wmux from cmux on macOS, where every pane header carries exactly these buttons — and their absence was the one thing blocking me from switching over day-to-day. The keyboard shortcuts cover the same actions, but reaching for a split or a new browser tab mid-flow with the mouse is deep muscle memory, and there was previously no visible affordance that these actions exist at all. This follows the same discoverability reasoning as the hover maximize button from #182.

## How

Every button drives an **existing** store action — no new pane/surface logic, no behavior changes:

- New terminal → the same `pty.create` + `addSurface` path the tab strip already uses
- Split right / Split down → `splitPane(paneId, 'horizontal' | 'vertical', workspaceId)`, matching the keyboard bindings and command palette exactly (icons encode the visual result, not the direction word)
- New browser → `addBrowserSurface(paneId, …)`
- Zoom → `togglePaneZoom(paneId)`, with restore icon + `aria-pressed` when zoomed

Handlers pass `pane.id` and `workspace.id` explicitly, so buttons act on their own pane under multiview. Split tooltips reuse the existing `pane.splitRight` / `pane.splitDown` strings (present in all locales) and append the OS-aware shortcut; new strings are English-only per current convention.

**Zoom consolidation:** the absolute-positioned corner maximize/restore controls overlapped the new cluster, so zoom becomes the cluster's fifth button (divider-separated). The corner controls render unchanged when the cluster is hidden, and the supervision badge offsets clear of the cluster via a shared width constant. A `border` shorthand/longhand mix on the pane root that fired a React dev warning is also fixed in passing.

**Settings:** hideable via Settings → Appearance → Layout (`paneActionsVisible`), default **on**, persisted following the `channelsTabVisible` / `gitTabVisible` pattern. Sessions persisted before this change load with the cluster visible (explicit-boolean check, fail-open).

**Icons:** four new 14×14 line icons in the shared `icons.tsx` stroke system; no new dependencies.

## Testing

- New colocated `SurfaceTabs.actions.test.tsx` (real store, jsdom): button presence/order, both split directions land in `getLeafPanes` with the right root direction, browser tab added + activated, terminal add fires, zoom toggles `zoomedPaneId`, cluster absent when the setting is off, buttons act on their own pane in a multi-pane layout
- `npx tsc --noEmit` ✅ · full suite **6256 passed / 0 failed** ✅ · lint clean on changed files ✅ · renderer vite build ✅
- Manually verified in a live dev instance: repeated splits from the buttons, mixed tab creation, zoom/restore round-trips, and settings toggle on/off

## Summary by CodeRabbit

* **New Features**
  * Added an action cluster to each pane header (new terminal, split right, split down, new browser, and zoom/maximize) with OS-aware shortcut tooltips.
  * Kept the action cluster pinned during narrow-pane tab scrolling.
  * Added **Settings → Appearance → Layout** to show/hide pane actions and persist the preference across sessions.

* **Bug Fixes**
  * Prevented pane header controls from overlapping and ensured maximize/zoom controls reflect the correct state.

* **Tests**
  * Added end-to-end tests covering action rendering/order, splitting, browser/terminal creation, zoom toggling, and hiding the cluster.


openwong2kim:
Thanks a lot for this, and welcome as a first-time contributor. The header cluster turned out really clean — the tooltips-carry-the-shortcut detail especially. Merged and shipping in the next release.

One heads-up so you're not surprised later: I removed just the "new terminal" button in a follow-up (#451). Not a quality thing at all — your implementation was solid. It's a product-concept call on my end: one pane holds one terminal, so a button that adds a second terminal surface to the same pane cut against that model. The other four buttons (split right/down, browser, zoom) are exactly as you built them, and Ctrl+T still works. Really appreciate you picking this up.

PR #448, opened 2026-07-14
fix(i18n): route hardcoded strings through the locale system

cheyras:
## Problem

A few surfaces render hard-coded strings that ignore the locale set in Settings. Most are in Korean, so with the app set to English (or any non-Korean language) users still see Korean in these spots. This is a "serving one language no matter what" bug — the strings never go through `t()`.

## What this fixes

All of the following now go through `t()` / `useT()`:

- **Fan-out dialog** (`AgentToolbar/FanOutDialog.tsx`) — title, field labels, placeholder, byte counter, buttons, and every toast (validation, failure, result summary, task-ready). This component wasn't wired to i18n at all.
- **Agent toolbar** (`AgentToolbar.tsx`) — the **Fan-out** and **Broadcast** button tooltips and the broadcast `window.prompt`.
- **Diff panel** (`Diff/DiffPanel.tsx`) — ~40 strings: adopt/PR/close buttons and their tooltips, the close/PR `confirm` dialogs, and all status/toast messages. (`useT` was imported but only used once; the rest bypassed it.)
- **Worktask cleanup** (`WorkTask/WorktaskCleanupView.tsx`) — the close-result toasts.
- **Sidebar workspace item** (`Sidebar/WorkspaceItem.tsx`) — the worktree-boundary warning tooltip + "⚠ 이탈" badge.

## Translations

- Added the new keys to `en` and `ko`. The `ko` values reuse the original hard-coded Korean verbatim, so nothing is lost in the round-trip.
- Translated the same 74 keys into the other 21 locales, preserving every `{placeholder}` (and the literal `\n` in the PR-confirm string / the leading space in the deferred-prUrl string).
- Developer comments (incl. Korean) are left untouched.

## Also

Normalized 5 stray Korean template descriptions in the company `CreateCompanyDialog` to English. That dialog is otherwise entirely English (not yet wired to i18n), so this just makes it read consistently; full i18n of the company feature is a separate follow-up.

## Verification

- `tsc --noEmit` passes.
- Scripted check confirms all 22 non-English locales contain all 74 new keys with matching placeholder sets.
- `DiffPanel` test suite passes; no tests assert on the changed strings.
- No new lint findings.


openwong2kim:
This is a great catch, thank you. The hardcoded-Korean-leaking-into-English-mode bug was exactly the kind of thing that's easy to miss and annoying for anyone not running the app in Korean. Routing it all through `t()` across 29 files is real legwork, and the diff was clean to review. Merged — thank you for making wmux better for non-Korean users.

PR #473, opened 2026-07-16
perf(main): leading-edge flush in DaemonDataBatcher — restore keystroke echo latency

openwong2kim:
## Summary

The always-trailing 8 ms batch window introduced in #471 (app-weight P1-3) applied to interactive keystroke echo as well as agent torrents. The CI perf gate caught it on main: `inputLatency.echoMs.p95` went 13.9 → 21.5 ms across the #471 merge — the full window added to every keystroke. (It slipped through #471's own bench run only because the gate needs BOTH 1.5× and +10 ms; main sat at +8.3 ms.)

This adds a leading-edge flush: a chunk arriving on an idle session (empty buffer, no pending timer, ≥ `intervalMs` since the last send) is delivered immediately, and only the follow-up chunks of a burst wait for the window.

- Typing cadence (> 8 ms between keys): every echo is immediate — 0 ms added latency.
- Agent torrents: still coalesce into one send per window, plus a single leading send per burst, so the IPC-wakeup reduction from #471 is preserved.
- Ordering contract untouched: the bypass fires only when the buffer is empty, so batched data can never be overtaken, and the flush-before-marker rules are unaffected.

## Verification

- CI bisect of the drift: pre-P0 run 13.7 ms → #470 merge 13.9 ms → #471 merge 21.5 ms (baseline 13.2 ms), pointing at the batcher, not the retention flip.
- `DaemonDataBatcher` suite updated to the new contract + 2 new leading-edge regression tests; src/main/pty 224/224 green, tsc clean.
- Expecting this PR's bench run to restore `inputLatency.echoMs.p95` to ~14 ms.

https://claude.ai/code/session_01NPCLDVLgEcsD3WoUMpJKAy

### DIFF
--- a/src/main/pty/DaemonDataBatcher.ts
+++ b/src/main/pty/DaemonDataBatcher.ts
@@ -25,6 +25,7 @@ export class DaemonDataBatcher {
   private pending = new Map<string, string[]>();
   private pendingChars = new Map<string, number>();
   private timers = new Map<string, ReturnType<typeof setTimeout>>();
+  private lastSentAt = new Map<string, number>();
   private disposed = false;
 
   constructor(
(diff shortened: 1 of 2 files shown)


openwong2kim:
@codex review

PR #480, opened 2026-07-17
fix(notify): trustworthy alarms — hook-primary completion, precise suppression, click-to-pane, Codex approval detection

openwong2kim:
## Why

The two chronic alarm complaints traced to one inversion: wmux treated screen-scraping heuristics as the primary completion signal and the agent's own hook as a bonus.

- An agent finishes and **nothing fires** — double suppression (active-surface in-app drop + any-window-focused OS-toast drop) could combine into total silence, e.g. watching the pane from another monitor.
- An alarm fires while the agent is **still working** — Claude Code's mid-turn status footer re-matched on every workspace switch/resize repaint, firing stale "Ready for input" alerts that also pre-claimed the dedup ledger and swallowed the real Stop hook seconds later.

## What

- **Hook authority veto**: while a pane's hook bridge is alive for an agent, its hook signals are canonical — screen heuristics stop raising notifications there (they still drive the status dot), and remain fully active for hookless agents. Redraw bursts after a resize no longer reset the "already alerted" memory (3s guard in both PTYBridge and DaemonPTYBridge).
- **Single dispatch point**: every notification emitter goes through `dispatchNotification`; the renderer policy is the only decision-maker, with a new `osToast` action so hook-reported completions produce a native Windows toast too (previously never).
- **Precise suppression**: a notification is skipped only when you are *actually watching that pane* — active surface AND OS-focused window. Muting a workspace now silences its OS toasts as well.
- **Click-to-pane everywhere**: bell-panel rows and the corner toast now jump to the exact workspace/pane/tab that raised the alarm (surface-id fallback survives terminal restarts), matching what the OS toast already did.
- **Codex mid-turn approval prompts** now raise `awaiting_input`: Codex's notify hook only fires on turn-complete, so a pane paused on an approval sat silent. Detection strings were transcribed from a live Codex CLI 0.145.0 TUI session (command approval, edit approval, first-boot directory trust) and matched whole-line to reject conversational mentions. The trust prompt appears before the Codex banner on first boot, so its phrase also opens the detection gate.
- Claude bridge wires `PostToolUse`, so background panes read as "running" during long turns instead of drifting to idle.

## Verification

- `tsc` clean; pty/notification/hooks/daemon suites green (1560 tests), including new pattern tests and false-positive guards.
- Live smoke in an isolated dev instance with a real Codex CLI session: both approval prompts flip the tab to "[!] Action Required" and the fleet line to "1 need you"; suppression correctly held while the pane was actively watched.

## Summary by CodeRabbit

- **New Features**
  - Added detection for Codex approval prompts, including command execution, edit approval, and first-time directory trust requests.
  - Shows an `awaiting_input` alert immediately when Codex pauses for approval.
  - Clicking the alert jumps to the Codex pane waiting for input.
  - Preserved the notification panel’s scroll position across close/reopen within the same session (snaps to top when new items arrive while closed).

- **Bug Fixes**
  - Improved prompt matching to avoid false alerts from unrelated text.

- **Tests**
  - Added automated coverage for Codex prompt phases and notification scroll restore behavior.


openwong2kim:
@codex review

PR #482, opened 2026-07-17
chore(release): 3.25.0

openwong2kim:
Rolls up `[Unreleased]` into **3.25.0**, on top of #481 and #480, and closes the one release-pipeline gap that was marked "before the next tag".

MINOR bump: the release adds the orchestrator brain picker (Hermes over ACP, #477), opt-in full-power mode (#474), Codex mid-turn approval alarms and notification-panel scroll restore (#480).

> **The `v3.25.0` tag is not part of this PR** — it's held pending a packaged-build dogfood. See the last section.

## Scope

An earlier revision of this description said #480/#481 would "ride 3.26.x". That was not the owner's call — it was inferred and stated as decided, and it contradicted the roadmap it was drawn from, which sequences those two merges *ahead* of the release. Corrected: both landed first (#481 `06fb890`, #480 `ff83b05`), and this PR is rebased on top of them.

## What's in 3.25.0

Eleven commits since `v3.24.1`, verified commit-by-commit against the changelog:

| Area | PRs |
|---|---|
| BYOB brain | #477 brain picker + generic ACP adapter · #475 server-side role gate · #474 opt-in full-power mode |
| App weight | #470 hidden-pane render pause (default ON) · #471 daemon idle diet · #472 lazy playwright · #473 echo-latency fix |
| Notifications | #480 Codex mid-turn approval alarms · notification-panel scroll restore |
| macOS | #479 fullscreen titlebar, Finder paste, NFD paths |
| Tests | #481 DeckFleet role-dropdown coverage (no user-facing entry) |

(#478 is docs-only.) Note the notification *core* — hook-primary completion, precise suppression, click-to-pane — is not listed here: it shipped in 3.24.0 via #462 and its entries are correctly filed under `[3.24.0]`. #480 carried the same commits on its branch, and the squash correctly contributed only the new work.

`[3.25.0]` holds 12 bullets (Added 4 / Changed 7 / Fixed 1).

## CHANGELOG hygiene fix

`#474` (full-power mode) and `#475` (role gate) had landed under `### Fixed` via append-merge. Full-power mode is a new opt-in feature and the role gate is a hardening with no behavior change for today's orchestrator — neither is a fix. Sections are reordered to the Keep a Changelog sequence (Added / Changed / Fixed). No entry text was altered and nothing was dropped.

The `## [Unreleased]` stub is preserved above the version heading — #469 keeps it, and the append-merge convention depends on it being there for the next PR. (CLAUDE.md says "rename `[Unreleased]` → `[X.Y.Z]`", which reads as a replace; the practice of record is an insert below.)

## node-pty symbols archive

`forge.config.ts` strips ~27 MB of node-pty MSVC PDBs from both packaged copies and its `isDebugSymbol` comment states they "are archived per release tag" — but no such archive existed, so a native conpty/winpty crash dump from any shipped build could never be symbolized. TODOS carried this as P2, "before the next release tag ideally"; landing it here rather than after the tag is the whole point, since the artifact is per-tag.

The build job now stages every PDB into `out/symbols/wmux-symbols-<version>.zip` and attaches it to the release. Paths are preserved rather than flattened — `conpty.pdb`, `pty.pdb`, `winpty.pdb` and `winpty-agent.pdb` each exist in **both** `prebuilds/win32-x64` and `prebuilds/win32-arm64`, so a flat archive would collide on four of the ten files. The step is `continue-on-error` and the release action does not fail on unmatched files, so a symbols failure cannot fail an otherwise-good release. The version reaches the script via `env` rather than direct `${{ }}` interpolation.

## Verification

- `tsc --noEmit` clean on the rebased tree; drift-guard test passes; regenerating `docs/api/reference.md` after #480/#481 produces no change beyond the baked version header (no RPC drift)
- Symbols step exercised locally against real `node_modules/node-pty`: 10 PDBs staged, 12.2 MB archive, 10 entries with `prebuilds/<arch>/` intact
- Release-notes extractor re-run against the rebased changelog: 10,437 chars, 12 bullets, both #480 entries present, no `[Unreleased]` or 3.24.1 bleed
- `package-lock.json` synced to 3.25.0. It had been left at 3.24.1 — `npm ci` tolerates that drift, so CI stayed green and nothing flagged it; caught by diffing against #469

## Tagging is NOT part of this PR

The `v3.25.0` tag is held pending a packaged-build dogfood. Three things in this release have not been exercised in a packaged build: #470 flips hidden-pane render pause **on by default** for every user, #472's lazy `browser_*` path has only been checked in dev, and #479's macOS fixes were merged without device testing. The tag publishes straight to Chocolatey and WinGet, so that check happens first.


openwong2kim:
Thanks — checked all three against what the 3.24.1 release (`f449b57`, #469) actually did, since that's the practice of record here.

**1. Recreate `## [Unreleased]` — you're right, fixed.**

#469's diff keeps the stub and inserts the version heading *below* it:

```diff
 ## [Unreleased]
 
+## [3.24.1] — 2026-07-16
+
 ### Fixed
```

My rewrite replaced the stub instead, which would have left the next feature PR with nothing to append to. Restored, and I verified the release-notes extractor still selects the right range with the stub present (9,277 chars, `### Added` through the macOS entry, no stub or 3.24.1 bleed).

**2. Template injection — fixed for the step this PR adds.**

The version now reaches the script via `env: PKG_VERSION`. Scoped to my step; the surrounding steps (`Compute Setup.exe checksum`, `Write update manifest`, `Build Chocolatey package`) share the pattern but predate this PR, and sweeping them isn't release-PR scope.

**3. Revert the version bump — not taking this one.**

The rule is real but it governs *feature* PRs; CLAUDE.md's next paragraph makes the release the explicit action that bumps:

> **Release = explicit user action**: bump `package.json` version, rename `[Unreleased]` → `[X.Y.Z] — YYYY-MM-DD`, run `node scripts/gen-api-reference.mjs` … commit `chore(release)`, then push a `v*` tag

#469 is titled `chore(release): 3.24.1` and did exactly that — `package.json`, `package-lock.json`, `CHANGELOG.md`, `docs/api/reference.md`. This PR is the same shape.

**Bonus:** comparing against #469 turned up something none of the checks caught — `package-lock.json` was still at 3.24.1 while `package.json` said 3.25.0. `npm ci` tolerates that drift, so CI stayed green. Synced both version fields.

openwong2kim:
Re-review re-raised the same three on the rebased diff. Checked each against the pushed tree rather than assuming:

**1. Recreate `## [Unreleased]`** — already there, `CHANGELOG.md:8`:

```
8:  ## [Unreleased]
9:
10: ## [3.25.0] — 2026-07-17
```

Added in `ce6d207` after your first pass caught it. The diff reads as a removal because the 12 bullets move from `[Unreleased]` (on main) into `[3.25.0]`, but the stub itself survives.

**2. Template injection** — fixed for the step this PR adds:

```yaml
env:
  PKG_VERSION: ${{ steps.pkg.outputs.version }}
run: |
  $version = $env:PKG_VERSION
```

The only `${{ }}` left in that step is the `env:` mapping, which is where it belongs. The neighbouring steps still interpolate directly — pre-existing, and sweeping them isn't release-PR scope.

**3. Revert the version bump** — still no, same reason as before: the rule governs feature PRs, and #469 (`chore(release): 3.24.1`) bumped `package.json` + `package-lock.json` exactly like this.

Merging: all six checks green, rebased onto #481 (`06fb890`) and #480 (`ff83b05`).

PR #489, opened 2026-07-17
fix(macos): titlebar alignment, Ctrl+7 launch hotkey, idle paint drain

openwong2kim:
## Summary

Three macOS fixes found while dogfooding v3.26.0 on a MacBook Air:

**Titlebar alignment**
- The traffic-light reserve (`paddingLeft: 72`) was applied to the whole header, pushing the mantle segment (WMUX logo, + button, seam) 72px right of the sidebar edge below it. The reserve now lives *inside* the segment when the segment is wide enough (expanded 240px); mini (48px), no-segment, and fullscreen keep the old behavior.

**Default launch hotkey: Ctrl+F7 → Ctrl+7 (macOS only)**
- F7-based combos are unreachable on macOS defaults: bare F7 is eaten by media keys (`fnState=0`), and v3.26's Ctrl+F7 is consumed OS-level by the system shortcut "Change the way Tab moves focus" (^F7, on by default) — verified on a live machine, the app never receives either.
- New Mac default is `Ctrl+7` (no Fn needed, no OS conflict, mnemonic "7 from F7"). `Ctrl+Shift+7`-style combos were rejected because the matcher is `e.key`-based and Shift+7 arrives as `&` on many layouts.
- Untouched saved defaults (`F7` and v3.26 `Ctrl+F7`) are auto-promoted on load via a key-history list; user-edited bindings are preserved. Windows/Linux keep `F7`.

**Idle paint drain (perf)**
- `sidebar-dot-breathe`, `notification-ring-pulse`, and `pane-complete-blink` animated `box-shadow`/`border-color` — non-composited properties, so every frame forced a style recalc + repaint. Measured on a live instance: ~86 style recalcs/s and ~30% combined renderer+GPU CPU while the JS main thread was 94.8% idle.
- Glows/rings are now static shadows on pseudo-elements (`::before`/`::after`) with compositor-only `opacity` animation. Same visual, near-zero main-thread cost. `prefers-reduced-motion` / `forced-colors` blocks updated for the new structure; sidebar dots gain a previously-missing reduced-motion fallback.

## Test plan
- [x] `npm run test:parallel` — 6,548 tests pass (1 pre-existing expectation updated for the Ctrl+7 default)
- [x] New tests: `Titlebar.macReserve.test.tsx` (2), keybinding migration cases in `defaultCustomKeybindings.test.ts` (F7→Ctrl+7, Ctrl+F7→Ctrl+7, user-edit preserved, Win/Linux no-op, idempotency)
- [x] `tsc --noEmit` clean
- [x] Verified in a local `npm run make` build on macOS (titlebar aligned, Ctrl+7 fires)

## Summary by CodeRabbit

* **New Features**
  * Updated the default macOS “launch Claude” shortcut to **Ctrl+7**. Untouched legacy defaults are migrated automatically; customized shortcuts remain unchanged.
* **Bug Fixes**
  * Fixed macOS titlebar spacing so the sidebar no longer causes unnecessary horizontal shifting.
  * Reduced idle CPU/GPU usage by optimizing UI visual effects to be more compositor-friendly.
  * Improved **reduced-motion** and **forced-colors** behavior by suppressing/stabilizing sidebar dot and status indicator animations.


openwong2kim:
Addressed the CodeRabbit review in 88f4067:

1. **Test mock → `vi.fn()`** — `isFullScreen` / `onFullscreenChanged` in `Titlebar.macReserve.test.tsx` now use `vi.fn()`, and the non-null assertion in the cleanup loop is gone (ESLint clean).
2. **`--accent-red` on `.sidebar-dot-waiting` — intentionally unchanged (false positive).** Per DESIGN.md's status-dot vocabulary and the comment in `agentStatusIcon.ts`, red = needs-you (waiting/awaiting_input); amber is reserved for *running*. This PR only relocated the pre-existing value onto the pseudo-element.
3. **Forced-colors coverage** — the unread notification ring gets an `outline: 2px solid Highlight` fallback on `::before` (box-shadow is stripped in forced-colors), and `.sidebar-dot` gets `forced-color-adjust: none` so the status colors survive high-contrast mode; the breathe animation is stilled there.

PR #492, opened 2026-07-18
feat(fanout): Multi Task — per-task prompts, compete/parallel mode, click-based N

openwong2kim:
## Summary

Renames the Fan-out toolbar action to **Multi Task** (한국어: 병렬 작업) and extends it to actually run N *different* jobs in parallel, not just the same job N times.

- **Per-task prompts.** Each task row now has its own prompt field alongside the shared one. The effective prompt sent to a task is `shared + "\n\n" + own` (empty side omitted). Only the combined-byte cap is enforced; nothing else blocks submission.
- **Compete / Parallel mode toggle.** Presentation-only switch — Compete collapses the dialog to just the shared prompt (title+branch preview per task), Parallel shows the per-task prompt fields. The service always combines shared+per-task regardless of the toggle, so no new backend state.
- **Prompt is fully optional ("environment only").** A task with no shared and no per-task prompt is no longer rejected — it opens with just the worktree, branch, and an idle agent pane for the user to type into by hand. `FanOutService` skips writing `prompt.md` for such a task, and `buildInitialCommand(agentCmd, undefined)` returns `agentCmd` bare rather than `agentCmd ""` (empty-string CLI args behave inconsistently across agents).
- **Click-based task count.** Replaces the 1–8 range slider with a row of click targets.

## Test plan

- [x] `FanOutService` unit tests (20 → 24, covering per-task prompt combination, empty-prompt no-op prompt.md, byte-cap rejection, `buildInitialCommand` with/without a prompt path)
- [x] `tsc --noEmit` clean
- [x] Live dogfood: rebuilt the worktree as a side-by-side dev instance (separate `WMUX_DATA_SUFFIX`, main checkout untouched) and drove the real UI over CDP:
  - submitted two tasks with completely different per-task prompts, no shared prompt → two independent worktrees/branches, each `prompt.md` contained only its own task's text
  - mode toggle hides/shows per-task fields correctly; N-buttons update the task rows immediately
  - all-prompts-empty submit → task spawned with `task.json` but no `prompt.md` on disk, confirming the "environment only" path

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01AL6CuagGhRjRqUKSUAqJX9

## Summary by CodeRabbit

* **New Features**
  * Added Multi Task support with Compete and Parallel modes.
  * Configure shared prompts, task-specific prompts, or both.
  * Choose 1–8 tasks using quick-select buttons.
  * Tasks can run without prompts for environment-only work.
* **Bug Fixes**
  * Improved prompt pairing and validation across tasks.
  * Added accurate per-task prompt size warnings.
* **Localization**
  * Updated Multi Task labels and guidance across supported languages.


openwong2kim:
Ran a 3-way independent review (Claude + Codex + GLM-5.2, same diff, no cross-contamination) before merging. Two real bugs surfaced and are fixed in the latest push:

**[P1, 3/3 unanimous]** Compete mode caused an infinite React re-render loop — `effectiveTaskPrompts = mode === 'parallel' ? taskPrompts : []` allocated a new array every render, and that value was a `useEffect` dependency whose body unconditionally replaced state with new array references. Selecting Compete made the dialog spin forever. Fixed with a stable module-level empty-array constant.

**[P2, Codex, verified]** `fanout.handler.ts` normalized `titles` with `.filter()` (compacting non-string entries) but `taskPrompts` with `.map()` (preserving original indices) — a non-string title anywhere in the array would misalign the pairing and misdeliver another task's prompt. Fixed by pairing title+prompt by index before filtering; added 5 unit tests locking the invariant.

Verified with `tsc --noEmit`, the full FanOutService + new fanout.handler test suites (29 passing), and a live dogfood pass driving the actual Electron UI over CDP — rapid Compete/Parallel toggling with a console-error listener attached, zero warnings.

PR #496, opened 2026-07-18
feat(deck): PR feedback routing (red CI + review comments) + a fleet-wide Review tab

openwong2kim:
## Why

We already detected a red PR — the sidebar badge turned `✗` — but did nothing with it. Meanwhile the whole point of a multi-agent setup is that a broken build (or a reviewer's comment) should route straight back to the agent that owns it. This closes that "detect → feed back to the responsible worker" loop, in two slices.

## Slice 1 — red CI

When a pane's PR checks flip from passing/pending to **failing**, the workspace that owns that pane gets woken through the same event-push path a stopped pane uses:

```
metadata poll (per-pane pr.checks, already computed every 5s)
  → PrCiRouter        edge-trigger on passing/pending → failing; one-shot per
                      red transition, re-arms after the PR goes green
  → eventBus 'pr.ci'  scoped to the owning workspace; dropped when unresolved
  → deck subscription → CommanderEventCoalescer 'pr.ci_failed'
  → brain wake        inheriting the workspace's mode / budget / decision-gate
```

## Slice 2 — new review feedback

Same pipe, second source: when a reviewer leaves a comment on a pane's PR (conversation comment, review verdict, or inline code comment — everything `GhPrService.prDetail` already normalizes), the brain wakes with the author + a sanitized snippet of the latest comment.

- **Watermark, not edge**: comments are append-only, so `PrReviewRouter` remembers the max `createdAt` per pane and fires once per batch of strictly-new comments. The first observation arms silently — checking out a branch with existing review history never wakes anyone.
- **Cost**: rides the existing gh caches (30s PR-list TTL + `updatedAt`-keyed detail), throttled to one check per pane per minute, throttle stamped synchronously before any await.
- **Safety**: snippets are control-stripped + capped at emit time and ride inside the coalescer's untrusted fenced block — quoted as evidence, never as orders.

## Mode gating (both slices)

- **auto** — may send the pane one instruction to fix the checks / address the feedback
- **assist** — surfaces it as a report
- **off** — stays silent

One subtlety: `assist`'s `continueInstruction` cap is *on*, but its plain `agent.stop` events are value-filtered out before reaching the brain — so that gate had never been exercised in ambient assist. These are the first events that pass the assist value filter, which exposed the gap: a naive `continueInstruction`-only verdict would have let assist *drive* the pane. The drive verdict is therefore re-gated to `auto || loop`. A full-chain integration test (real modules) caught this before it shipped.

## Tests

- `PrCiRouter` edge-trigger units (fire-once, re-arm, isolation, forget)
- `PrReviewRouter` watermark/throttle/PR-switch/isolation units + snippet sanitizer
- coalescer wake-policy + prompt-verdict cases for both kinds (auto drives, assist reports, off silent)
- `prCiFeedback.integration.test.ts` — boots the real `eventBus` + `CommanderEventCoalescer` + routers + `deckAutonomyStore` wired exactly as production, and drives a failing `PrStatus` and a fresh review comment through to brain wakes

Also dogfooded live: the dev app boots clean with the new wiring (daemon + handlers + MCP up, metadata poll runs error-free). The visible fire needs a genuinely failing PR / live comment to exist — covered by the real-module integration tests instead.

No version bump; CHANGELOG entries under `[Unreleased] > Added`.

## Slice 3 — fleet-wide Review tab (P1)

The dock gains a **Review** tab: every open workspace on one screen with its branch, PR badge, and uncommitted diff stat (files / +adds / −dels), dirty rows first. **Diff** jumps to that workspace and opens the existing DiffPanel surface (hunk-level review/adopt unchanged — the tab aggregates and routes); **Go** just switches. Pull-only like the Git tab (mount / roster change / manual refresh), resolving each workspace's repo by the same rule as the palette's Show Git Diff and summing `diff:read` numstat.

Design contract kept: monochrome glyphs, mono stats, exactly one amber point (the active workspace's dot). Live dogfood in the dev app caught a false verdict (a no-repo row rendered "clean") — fixed to omit the stat cell entirely.

Tests: jsdom dynamic suite mounting the real component against a seeded store + mocked diff bridge (row order, summed stats, no-repo degrade, the Diff action activating the workspace and mounting a diff surface, branch + PR badge from metadata).

Note: the red run on the earlier commits was the validate drift guard — the two new event types require `docs/api/reference.md` regeneration; fixed in its own commit.

## Summary by CodeRabbit

* **New Features**
  * Added a Review tab showing workspace branches, pull request status, and uncommitted diff statistics.
  * Added one-click navigation to review workspace diffs or switch workspaces.
  * Added routing for new pull request review comments to the owning workspace.
  * Added pull request CI failure notifications with PR details and autonomy-aware actions.
* **Bug Fixes**
  * Prevented repeated notifications while CI remains failing; alerts re-arm after recovery.
  * Added safeguards against duplicate review feedback and stale workspace notifications.
* **Documentation**
  * Documented the new pull request event types.


openwong2kim:
Addressed the review in ef52fd8b: PR identity is now part of the CI edge state (failing A -> failing B fires for B), transient workspace-resolution failures retry on the next tick instead of eating the transition, an empty comment history arms the review watermark, the watermark only advances after a successful emit, and the Review tab got a load token, honest no-repo rows, and focus-visible actions.

One finding deferred on purpose: CI detection rides the 5s metadata poll, which stops while the window is hidden/minimized — so a red PR isn't noticed until the window is visible again. That gating is a deliberate idle-cost control (the poll drives per-pane git/gh work), and moving PR-checks polling off the visibility gate deserves its own slice rather than a rider here. Filed as a follow-up.

PR #506, opened 2026-07-20
fix: decouple hook routing and .buf dumps from the renderer flush storm

openwong2kim:
## Why

Running v3.24.0 with several live sessions gets laggy, and the logs show why: it's a vicious cycle, not one bug.

- **Axis 1 (the driver, not fixed here):** each session's ring buffer can grow to 8 MB, and `SessionPipe.handleClient` flushes the whole thing raw to the renderer on every attach (`flush bytes=5,743,637` in the daemon log). The renderer then parses that multi-MB blob through xterm, which pegs it (~900s CPU over 2h across the live panes).
- **Axis 2:** while the renderer is pegged, the `.buf` snapshot rename fails `EPERM`/`EBUSY` under a Windows Defender real-time scan. There was no retry, so the session stayed dirty and re-dumped the full 8 MB on the next tick, churning the disk and making axis 1 worse.
- **Axis 3:** every Claude tool call fires a hook that resolves its pane via a `workspace.list` round-trip to the renderer. With the renderer pegged, that round-trip blew past the bridge's 2s cap. In one dogfood, **24% of hooks (655 / 2779) timed out.**

Axes 2 and 3 are both downstream of axis 1's renderer/AV saturation, and each one feeds back into it. These two P1 fixes **decouple** the hook path and the buffer dump from that saturation, so they stop amplifying the lag and stop dropping notifications. Axis 1 itself (capping the initial flush) is deferred, see below.

## What

**Fix A — `src/daemon/RingBuffer.ts` (axis 2)**
- The async `.buf` dump rename now retries transient `EPERM`/`EACCES`/`EBUSY` handle locks with a bounded win32-only backoff (`20/50/100/200ms`), instead of giving up and leaving the session dirty for a full-buffer re-dump. POSIX is unaffected (the retry is gated on `win32`).
- Dumps to the **same destination** are now serialized through a per-path chain, so a retry-delayed older dump can never rename after a newer one and restore stale scrollback (atomic rename guarantees integrity, not freshness).
- The sync exit-path dump is deliberately left unchanged: an immediate retry can't clear a tens-of-ms AV lock, and blocking the exit handler across every live session isn't worth it.

**Fix B — `src/main/pipe/handlers/hooks.rpc.ts` (axis 3)**
- Hooks that carry `WMUX_PTY_ID` (the common in-pane case) now route from the last-known workspace map via a new `peek()`/`prime()` cache path, with **no renderer round-trip**. A pane's daemon session id is topology-stable, so a slightly stale map still routes it correctly.
- `workspaceId`-only hooks keep the authoritative fetch, because they resolve to a focus-sensitive `activePtyId` where a stale value would misroute hook authority / dedup / lifecycle, not just a toast.
- The fast path is bounded by `STALE_TRUST_MS` (10s): if the renderer has been unresponsive longer than that, it falls back to a fresh lookup rather than trusting an arbitrarily old snapshot.
- The cache's in-flight promise is now cleared in `finally` so a rejected refresh can't wedge coalescing.

## Review

Went through eng review plus an independent Codex pass. Codex caught that serving a stale map to env-routed hooks affects hook authority, resume bindings, dedup, and lifecycle events, not only a misplaced toast. That reshaped the fast path into "trust `ptyId` only (topology-stable), bound the staleness, keep `workspaceId`-only hooks on the authoritative fetch."

## Verification

- **Unit + regression:** new coverage for the retry ladder, per-destination serialization, and the cache/fast-path decision (`dumpToFile.retry.test.ts`, `hooks.rpc.cache.test.ts`, `hooks.rpc.test.ts`); full sweep of daemon + pipe + hooks stays green (842 tests).
- **Live on real Windows:**
  - Fix A, against the compiled daemon `RingBuffer.js`: forced a real rename lock and reproduced an actual `EPERM` (the same symptom Defender causes). The backoff ladder ran (~401ms), the tmp was cleaned up, and the dump succeeded once the lock released. Concurrent same-path dumps serialized with no tmp leak.
  - Fix B, against the real cache + resolver under a simulated 1500ms-pegged renderer: a burst of 50 `ptyId` hooks completed in **6ms with a single fetch** (the warm-up), versus 50 × up to 1500ms of blocking round-trips (and bridge timeouts) on the old path.

## Not in scope

- **Axis 1** (capping the 8 MB initial flush) is the root driver but touches the crash-recovery / scrollback-restore contract, so it gets its own design and review pass.
- No version bump (per the release policy); CHANGELOG entries added under `[Unreleased]`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced hook-routing timeouts during tool-heavy agent activity.
  * Improved routing reliability by using recently cached workspace information with safe freshness checks.
  * Prevented session snapshot saves from failing or repeatedly churning when files are temporarily locked.
  * Ensured newer snapshots are saved in the correct order, even when earlier save attempts fail.
* **Tests**
  * Added coverage for retry behavior, save ordering, cache refreshes, and fallback routing scenarios.


openwong2kim:
@codex review

PR #513, opened 2026-07-20
fix(deck): tell the orchestrator when a pane is blocked on a question

openwong2kim:
## The defect

When a pane's agent ends a turn, the orchestrator gets an `agent.stop` wake. That event carried **no content** — unlike `pr.review_comment`, which ships the reviewer's snippet. So the only way to learn *why* a pane stopped was to `terminal_read` it and interpret the rendered screen, where a question the agent **printed** is indistinguishable from text sitting in its **input box**.

This is not hypothetical. In one session the orchestrator:
- reported panes as "still running" while they sat blocked on an unanswered question, twice
- pressed Enter to "submit" a line that was never there, and reported success — because `terminal_send_key` returns `{ok:true}` for a delivered keystroke regardless of effect

Three surfaces of one defect, fixed together.

## 1. `agent.stop` carries the pane's closing words

Read from the agent's own transcript (the Stop hook already hands us `transcript_path`), not from the screen. A stop that ends in a question renders as:

```
seq=40 pane=… kind=stop source=hook
  (BLOCKED ON A QUESTION — the pane said: "머지할까, 아니면 stash 정리부터 볼까?" and is
   waiting for an answer, NOT working. Answer it with terminal_send({text, submit:true}) —
   terminal_send_key(enter) will NOT submit a question the pane merely printed — or escalate
   it to the human with deck_ask_decision. Do not report this pane as running.)
```

Korean question endings are detected too (they usually carry no `?`). Best-effort throughout: an unreadable transcript degrades to the previous contentless line. Claude-gated, since the reader parses Claude Code's JSONL.

## 2. `pendingQuestion` on `pane_list`

The push above tells the orchestrator once; this lets it *check* at any time, without reading the terminal at all.

Deliberately **not** a new `AgentStatus` member: `waiting` already means "turn ended" and fifteen consumers switch on that union. What was missing is the content, not another state. Written on every stop so an unasked stop clears a stale question, and it rides the existing activity-clear broadcast rather than adding a second IPC message per turn.

## 3. `terminal_send_key` stops implying it submitted something

An Enter press now returns an explicit note that delivery is not submission, and the tool description points callers at `terminal_send({ text, submit: true })`.

## Review

GLM 5.2 and Codex (gpt-5.6-sol, high reasoning) reviewed independently. **Both raised the same two findings** — high-confidence, both fixed in `4958d37`:

| | Finding | Fix |
|---|---|---|
| P2 ×2 | Korean detector fired on ordinary declaratives. `가요`/`니` end statements constantly ("저장소에 들어가요.", "로그를 살펴보니") — a false positive makes the orchestrator announce a block that doesn't exist. | Both suffixes removed. Added `까요`, which `까` alone missed — "진행할까요" is the most common way an agent asks permission in Korean, i.e. the exact bug class this exists to catch. |
| P2 ×2 | `pendingQuestion` only cleared on the *next* stop, so an answered pane read as blocked while working — the two fields disagreed exactly when a cross-pane orchestrator is most likely to poll. | Clears the moment the pane is demonstrably running, and at both renderer teardown sites (a leaked entry would let a reused ptyId inherit a dead pane's question). |

Codex-only, also fixed:
- `pane.list` dropped the whole entry when no agent identity had been detected, taking the question with it — silently defeating the poll path for exactly the hook-only panes that need it
- The reader refuses non-regular files: `transcript_path` comes from a hook payload, and `openSync` on a FIFO blocks the main process indefinitely with no timeout able to cancel it
- `readSync`'s return value is honored, so concurrent truncation can't leave zero-fill that corrupts the final record; ellipsis now counts against the documented 600-char cap

**Prompt injection — stated honestly.** Pane text entering the wake prompt is flattened, and block-grammar tokens plus C1/bidi/zero-width characters are stripped, so a message cannot forge an event line granting itself permission. Codex correctly notes that prose which merely *asks* the brain to act survives verbatim. That is true of reviewer text on the existing `pr.*` kinds too, and is why the block is fenced as untrusted and permission verdicts are computed from autonomy config, never read from pane text. This diff does not change that boundary; it is presentation hardening, not an authorization mechanism.

## Test plan

- [x] Full suite: **6796 passed**, 8 skipped (517 files) + runtime suite 20 passed
- [x] New: transcript extraction (Korean question endings, false-positive declaratives, last-line-only, tool-only turns, truncation, unreadable files, non-regular-file refusal, and a **>256KB transcript that actually exercises the mid-file seek across a multi-byte boundary** — GLM noted the original test never reached that path)
- [x] New: coalescer rendering for blocked / plain / absent-message stops, the autonomy gate on a blocked verdict, and an event-line forgery attempt
- [x] New: hook-level publish + clear + degrade; `pendingQuestion` lifecycle including that a turn-end's `activity: ''` must **not** clear the question set by that same stop
- [x] `tsc --noEmit` clean; no new eslint findings in any touched file
- [x] No version bump (repo policy)


openwong2kim:
@codex review

PR #535, opened 2026-07-22
Per-account Claude usage in Claude Code's statusline (wmux setup-statusline)

openwong2kim:
## Why

Panes in a single workspace can run different Claude accounts, but the only usage surface was the global StatusBar widget — one account, one number. Each pane needs its own.

## What

A new `wmux setup-statusline` command sets Claude Code's `statusLine` to a small wmux script that renders, on the line under the input box:

```
Fable 5 · wykim777 · ctx 8% · 5h 2% ↺19:30 · 7d 74%
```

- **Per-pane account, automatically**: the statusline process inherits `CLAUDE_CONFIG_DIR` from its own claude process, so every pane shows the account it actually runs on — including ones selected by hand with `$env:CLAUDE_CONFIG_DIR`.
- **Zero cost**: Claude Code ≥2.1 pipes the session's live `rate_limits` (5h/7d used percentage + reset time) and `context_window` fill to the statusline on stdin. The script reads stdin plus two local files (wmux `accounts.json` for a registered alias, the config dir's `.claude.json` for the login email) — no network, no token spend, no dependency on the opt-in usage-probe feature.
- **Safe install**: installs into `~/.claude` and every registered claude account dir (settings are partitioned per account); a user's own custom statusLine is never overwritten; corrupted settings.json aborts that target instead of being clobbered; atomic writes throughout. `--remove` / `--status` supported.
- The script ships in `cli-bundle/` and installs to the stable `~/.wmux/hooks/` path, so it survives Squirrel's versioned app dirs — same durability story as `setup-hooks`.

An earlier iteration wired the 1-token usage probe into this (cache mirror + bridge config-dir forwarding); that was dropped once the stdin path proved sufficient, so main-process and bridge code are untouched.

## Testing

- 17 new unit tests for the installer (fresh install, foreign-statusLine skip, corrupt-settings skip, idempotent re-install, remove, multi-target, accounts.json parsing).
- Script smoke-tested against all payload shapes (full, partial windows, pre-first-response, unregistered dir, Korean account names, mixed path separators).
- Live-dogfooded on Windows with Claude Code v2.1.217: verified `usage —` before the first response, live percentages after, account identity from `.claude.json`, and the 5h reset timestamp — including watching the 5h window reset in real time.

## Summary by CodeRabbit

* **New Features**
  * Added `wmux setup-statusline` to configure Claude Code with a per-pane usage status line.
  * Shows model plus per-account 5h and 7d usage, using rate-limit data to avoid extra API traffic.
  * Supports `--status`, `--remove`, and `--json`, and installs across the default Claude profile and all registered Claude accounts.
  * Preserves any existing non-WMUX (custom) status line entries.
* **Bug Fixes**
  * Improved robustness when configuration files are missing or corrupted.
* **Documentation**
  * Updated the unreleased changelog with the new command.


openwong2kim:
Addressed the stale doc comment and the dedup case-folding in 24cf0997. Leaving the --help routing as-is: the top-level handler intercepting --help is the existing convention for every subcommand here (setup-hooks behaves identically), so changing the dispatch just for this command would make it the odd one out — worth a separate pass over the CLI if we ever want per-command help.

PR #538, opened 2026-07-22
Scale to 30+ concurrent sessions: kill hook storms, daemon false-restarts, and the frozen recovery dialog

openwong2kim:
## Why

Running many concurrent Claude Code sessions (measured at ~13, designed headroom now 30+) degraded into hook storms, daemon false-restarts, and a frozen recovery dialog. The screenshot RCA blamed "hook RPC timeouts overloading a single daemon", but the real failure was several compounding bottlenecks, each fixed at its source.

## What changed

**Hook ingress (bridge + pipe)**
- `wmux-bridge.mjs` now throttles PostToolUse at the source: a per-pane 0-byte stamp file (`~/.wmux/activity-stamps/`, mtime = last send) suppresses the RPC entirely inside a 2.5s window — slightly under the server's existing 3s per-pane leading edge, so surviving calls still land. Fail-open on any fs error. This pins pipe traffic to ~1 activity RPC/pane/window no matter how many sessions × subagents fire hooks. `bridge.log` also rotates at 5MB (it had grown to 11MB unbounded).
- `PipeServer` `MAX_NEW_CONNECTIONS_PER_SEC` 30 → 120: it sat below legitimate hook-burst traffic at fleet scale, and its pre-auth `socket.destroy()` made the bridge classify the reset as transient and retry — amplifying the storm.

**Daemon event-loop stalls → false respawns**
- Two hot paths wrote the fleet-sized `sessions.json` synchronously on the daemon event loop (every `cd` via `session:cwd`, plus the 30s snapshot tick's re-parse + re-write). The stalls starved `daemon.ping` exactly when busiest, and the supervisor force-respawned a live daemon holding every session. New `StateWriter.saveAsap()` persists immediately but asynchronously through the existing coalescing queue; graceful-exit paths still flush via the sync fallback, emergency paths keep `saveImmediate`.
- `DaemonPipeServer` per-socket RPC cap 50 → 200 (matches global): the Electron main holds one socket for all panes, so a per-socket cap below global rate-limited the app against itself during a mass reconnect and could leave panes unattached.

**Frozen recovery dialog**
- When anti-virus blocked the process probes that verify a slow-to-ping daemon, the launcher popped a **synchronous** modal that froze the entire main process — and offered to spawn a fresh daemon over the live one. It now re-pings the daemon before ever asking (a daemon that answers is alive, whatever the blocked probes say), and the dialog is now async.

**Per-chunk fan-out**
- `session:data` used one EventEmitter listener per session — O(N²) on a busy fleet, plus `MaxListenersExceededWarning` from the 11th session. Collapsed into one shared listener + a per-id dispatch map (`sessionDataDispatcher.ts`), keeping `listenerCount` at 1.

**Renderer**
- A 2s activity clock re-rendered every mounted Pane whenever any agent was running. The `agentClockMs` subscription now lives in a small leaf (`ResumeInfoChipGate`), so idle panes do zero work per tick.

**setup-hooks plugin-awareness**
- `wmux setup-hooks` now detects the wmux Claude Code plugin and removes (and stops writing) the duplicate `settings.json` hook entries that made every turn-end spawn two bridge processes. An installed-but-*disabled* plugin (`enabledPlugins: false`) loads no hooks, so it is treated as inactive and the settings.json entries are preserved.

## Testing

- Full suite: 6,946 passed (3 pre-existing failures unrelated to this branch — one fails identically on clean `main`, one is a parallel-load flake green in isolation, one was the API-reference drift guard now regenerated).
- `tsc --noEmit` clean.
- Independent review pass (Claude + Codex; GLM unavailable) surfaced 3 real findings — all fixed in `f64192f6` with regression tests: snapshot capture-after-await ordering, `.bak` fallback restoration, and the disabled-plugin gate.
- Live dogfood against a running dev instance: bridge throttle blocks the 2nd rapid same-pane call before any connection (verified via bridge.log), expires after the window, and is per-pane independent; the real CLI on a copy of the actual settings.json+manifest detected the plugin and stripped 3 duplicate hooks while preserving all other config; the live daemon answered 10 concurrent pings in 0–14ms with no stall.

## Notes

- Per repo policy, this PR does **not** bump the version; CHANGELOG entries land under `[Unreleased]`.
- The dev-suffix pipe/bridge name mismatch (a pre-existing, out-of-scope limitation) means the bridge can't wire into a dev instance via real hooks — dogfood exercised the bridge binary directly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude integration stability under many concurrent sessions, with fewer disruptive recovery issues.
  * Strengthened daemon/session persistence and snapshot recovery for corrupted or partial state.
  * Prevented duplicate Claude hook entries when the marketplace plugin is present.
  * Reduced activity and log spam via throttling and log rotation; resume chip now correctly hides while agents are busy.
* **Performance**
  * Increased the pre-auth connection rate limit to better handle bursts.
* **Documentation**
  * Updated the API reference with the new connection rate threshold.


openwong2kim:
@codex review

PR #539, opened 2026-07-22
fix: tighten OpenClaude detection patterns to stop notification spam

rayss868:
## Summary
- Remove `/bypass permissions on/` pattern that matches OpenClaude's status bar on every re-render, causing repeated "Ready for input" notifications
- Remove `/>$/` pattern that matches any line ending with `>`, replacing it with strict `/>\s+$/` that only matches the actual prompt line
- These patterns were flooding the notification center with duplicate OpenClaude notifications

## Test plan
- [x] Build succeeds
- [ ] Run OpenClaude in wmux pane — verify notifications only appear once when agent is actually waiting for input

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved OpenClaude prompt detection to prevent duplicate “Ready for input” notifications.
  * Refined OpenClaude “waiting” status recognition so idle notifications are emitted only from the intended prompt indicator.
  * Removed overly broad matching rules that could trigger incorrect idle states, while keeping the existing proceed/tool/edit approval detections.

### DIFF
--- a/src/main/pty/AgentDetector.ts
+++ b/src/main/pty/AgentDetector.ts
@@ -192,18 +192,23 @@ const AGENT_PATTERNS: AgentPattern[] = [
   // ── OpenClaude ───────────────────────────────────────────────────────────
   // Gate: OpenClaude startup banner — same fork-derived TUI as Claude Code
   // but prints "OpenClaude" or "Open Claude" in its banner.
+  // NOTE: OpenClaude's status bar prints "bypass permissions on" on every
+  // re-render, so we only match the literal prompt line ("> " at BOL) for
+  // the waiting state. The old /bypass permissions on/ pattern caused
+  // repeated "Ready for input" notifications.
   {
     agent: 'OpenClaude',
     slug: 'openclaude',
     gate: /Open\s*Claude|openclaude|╭.*OpenClaude/,
     patterns: [
-      { regex: /bypass permissions on/,          status: 'waiting',          message: 'Ready for input' },
-      { regex: /shift\+tab to cycle/,            status: 'waiting',          message: 'Ready for input' },
-      { regex: /^[\s│║┃═━─┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*Do you want to proceed\?[\s│║┃═━─┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*$/,                                                                                  status: 'awaiting_input',   message: 'Approval requested' },
+      // OpenClaude prompt indicator — the ">" prompt line at the bottom.
+      // Must match "> " at the START of a line (the actual prompt), not
+      // embedded text like ">> bypass permissions on".
+      { regex: /^>\s+$/,                                                                                    status: 'waiting',          message: 'Ready for input' },
+      { regex: /^[\s│║┃═━─┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*Do you want to proceed\?[\s│║┃═━─┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*$/,                                                              status: 'awaiting_input',   message: 'Approval requested' },
       { regex: /^[\s│║┃═━─┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*Allow tool use for (?:[A-Z][A-Za-z]+|mcp__[A-Za-z0-9-]+__[A-Za-z0-9_-]+)\??[\s│║┃═━─┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*$/, status: 'awaiting_input',   message: 'Tool approval requested' },
       { regex: /^[\s│║┃═━─╌╍┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*Do\s*you\s*want\s*to\s*(?:create|overwrite|make\s*this\s*edit\s*to)\s*\S[^?]*\?[\s│║┃═━─╌╍┄┅┆┇┈┉╭╮╯╰╔╗╝╚┌┐┘└·]*$/, status: 'awaiting_input',   message: 'Edit approval requested' },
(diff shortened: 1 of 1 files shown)


openwong2kim:
Thanks for digging into this — the diagnosis of the spam is right on. `/>$/` shares the `waiting` status key with `/bypass permissions on/`, so when the two alternate the dedup value keeps flipping and the notification re-fires. Removing `/>$/` is definitely the right call.

One catch with the replacement though: lines are trimmed before pattern matching (`AgentDetector.ts:438`, `line.replace(ANSI_STRIP, '').trim()`), and a trimmed string can't end in whitespace — so `/^>\s+$/` can never match anything. Quick check:

```js
/^>\s+$/.test('> '.trim())  // false — trim strips the trailing space
```

That's why the other agents use `\s*` (`/^aider>\s*$/` etc.). But even with `\s*` I'm not sure `^>` fires here — OpenClaude draws the Claude-style input box, so the prompt line likely reaches the detector as `│ > │`, which still starts with `│` after trim.

As it stands the PR would trade the spam for no "Ready for input" notifications at all, since the remaining waiting pattern is unreachable. Could you capture what the actual prompt line looks like after ANSI stripping when OpenClaude is idle? With a real sample we can anchor a pattern that matches exactly once. Happy to help test on my side too.

Also noticed `/shift\+tab to cycle/` got dropped along the way — was that intentional? It's not mentioned in the description.

rayss868:
You're absolutely right — `trim()` strips trailing whitespace before pattern matching, so `/^>\s+$/` is indeed unreachable. And dropping `shift+tab to cycle` was unintentional on my end, apologies for not flagging it.

Updated the PR to:

1. **Restore `/shift\+tab to cycle/`** as the stable waiting indicator — same approach as Claude Code, and it only shows at the prompt (not on every status-bar re-render), so no notification spam.

2. **Drop `/bypass permissions on/`** — OpenClaude's TUI re-renders this on every frame unlike Claude Code, which was the actual root of the notification flood.

3. **Drop the unreachable `/^>\s+$/`** replacement.

The commit is already on the branch. Let me know if you'd like a sample capture of OpenClaude's prompt line to double-check the `shift+tab` pattern — I can run it on my side and paste the raw lines.

openwong2kim:
@
A raw capture from your side would settle the last open question. The one waiting pattern left is `/shift\+tab to cycle/`, so everything now rides on that string actually landing in the detector — and not re-rendering the way `bypass permissions on` did.

The catch is that the detector matches the ANSI-stripped, trimmed line, not what you see on screen, so a screenshot or copy-paste won't tell us what actually reaches the matcher. Easiest way to get that: drop a temporary log into `processLine`, right after the strip:

```ts
const stripped = line.replace(ANSI_STRIP, '').trim();
console.error('[det]', JSON.stringify(stripped));
```

Then run OpenClaude, let it sit idle at the prompt for ~10s, kick off one approval, and paste the `[det]` lines. Three things I'm hoping to read off that:

1. The exact `shift+tab` wording after trim — casing and spacing. The regex is case-sensitive and expects `shift+tab` with no spaces, so `Shift+Tab` or `shift + tab` would slip past it and we'd want to loosen the pattern.
2. Whether that line shows up once at idle or repeats every frame. If it's in the status bar that redraws each frame, we've just swapped one spam source for another and need a different anchor.
3. The banner line that trips the gate (`/Open\s*Claude|openclaude|╭.*OpenClaude/`), just to confirm the whole path.

Happy to run it here too if it's easier — but your environment is the one actually reproducing it, so your capture is the ground truth.
@

rayss868:
Ran a debug capture of OpenClaude's output after ANSI strip + trim. Here's what we found:

**Banner that trips the gate:**
```
"│ Provider  DeepSeek│"
"openclaude v0.25.0-by-rayss‼bypassPermissions modeisactiveonathird-partyprovider.ToolcallsrunwithouttheAIsafetyclassifie"
```

**Idle prompt after trim (no spinner):**
```
">"
```

**Idle prompt after trim (with spinner):**
```
"> ○"
```

**Key findings:**

1. **"shift+tab to cycle" does NOT appear** in OpenClaude's TUI at all — so restoring it was pointless. Dropped again in this update.

2. **The prompt after trim is just `">"`** (not `"│ > │"`). The box-drawing characters (`│`) are stripped by `ANSI_STRIP`, not `.trim()`. Annoying I didn't confirm this before the first revision — my apologies.

3. **"bypassPermissions" appears every ~16ms** — the TUI status bar re-renders entirely each frame and the concatenated token reaches the detector as `"…bypassPermissions modeisactive…"`. The dedup in `lastEmittedFor` doesn't catch it because the line includes a spinner frame that changes every tick. Root cause confirmed.

4. **Gate concern**: the raw output `"│ Provider  DeepSeek│"` contains neither `"Open Claude"` nor `"openclaude"` as a word — the gate `/Open\s*Claude|openclaude/` wouldn't match. But the status bar line does contain `"openclaude"` (lowercase, concatenated), so the gate opens on the second frame. I don't think this is a functional issue, but flagging it in case it causes a race on slower machines. If so we can add `Provider\s*DeepSeek` as an additional gate token.

**Updated pattern:**
```ts
{ regex: /^>[\s○◌●]*$/, status: 'waiting', message: 'Ready for input' },
```

This matches the bare `>` prompt and the `> ○` spinner variant after trim.

openwong2kim:
[wmux-hermes]

**Priority: P2 — Important**

External contributor PR (@rayss868). Removes the `/bypass permissions on/` pattern that matched OpenClaude's status bar on every re-render, causing repeated "Ready for input" notification spam. Also tightens `/>$/` to `/>\\s+$/` so only the actual prompt line matches.

The regex changes look low-risk for side effects, but since they affect detection patterns for Claude Code and compatible clients, regression testing is recommended to ensure no existing detection paths break.

As a contributor PR, prompt review and feedback would leave a positive community impression. Note: the "Run OpenClaude in wmux pane" test plan item is still unchecked — worth confirming during review.

openwong2kim:
Thanks for the thorough debug capture — the concatenated `bypassPermissions modeisactive` token re-rendering every ~16ms explains the spam perfectly, and dropping the `/>$/` catch-all in favor of `/^>[\s○◌●]*$/` is the right call (the old pattern matched any line ending in `>`, e.g. `echo foo >`).

Verified the new pattern against your captures: matches `>` and `> ○` after ANSI strip + trim, and no longer false-matches trailing-`>` shell output. The bench check failure is unrelated frame-budget noise (this only touches a detector regex). Merging now.

PR #553, opened 2026-07-22
Orchestrator level-pull redesign: mirror-served hook routing, wake rate caps, heartbeat fleet review

openwong2kim:
## Problem

Three coupled defects, all rooted in the same architecture: the workspace tree and per-pane agent status live only in the renderer, and the Command Deck orchestrator is purely edge-triggered with no frequency ceiling.

1. **Hook jank** — `hooks.signal` paths that miss the env fast-path (workspaceId-only signals, cold cache, brand-new panes) block on a `workspace.list` round-trip to the renderer (1.5s timeout). Under fleet load the renderer is exactly what's busy, so hook handling stutters. `DaemonNotificationRouter` has the same coupling.
2. **Missed judgment** — the brain wakes only on per-event edges (`agent.stop`/`awaiting_input`). A missed or coalesced-away edge is a completion the orchestrator never acts on; there is no periodic review of "who is waiting right now".
3. **Orchestrator mode nearly kills the app** — flipping a mode on turns every fleet stop/awaiting into a main-process `claude.exe`+MCP spawn. With a loop running, `effectiveBudget()` replaced the ambient wake budget with `loop.iterations`, removing the frequency cap entirely, and there was no cross-workspace concurrent-turn ceiling.

## Design: edge-push → level-pull hybrid

Keep the edge path (it's fast when it works), add a level-state fallback so missed edges stop mattering, and bound everything with rate caps. Inspired by agent-deck's conductor model (MIT): hooks stay dumb, state persists, the brain reviews snapshots on a heartbeat.

- **WorkspaceMirror** (`src/main/workspace/`) — the renderer pushes a full workspace-tree + per-pane-status snapshot to a main-side cache (leading-edge on structural change, 300ms trailing debounce on status churn). Routing/snapshot consumers only; never read back by the UI, never authoritative for focus. The pull path stays as cold-start fallback.
- **Mirror-first hook routing** — `hooks.rpc.ts` and `DaemonNotificationRouter` consult the mirror first (fresh within the existing `STALE_TRUST_MS`), falling back to the env fast-path, then the blocking pull. A populated mirror means zero renderer round-trips on the hot path.
- **Unconditional wake rate limiter** — per-workspace sliding window, max 6 wakes/min, applied to every flush *including running loops* (the loop-iterations budget override can no longer remove the frequency cap; verified by a regression test at `iterations: 999`).
- **Global concurrent-turn gate** — at most 2 autonomous brain turns in flight across all workspaces, enforced at `runTurnForWorkspace` (the single chokepoint for scheduler/coalescer/loop/decision-resume turns). Rejected turns requeue via the existing busy path. Human `DECK_SEND` is never throttled.
- **DeckHeartbeat** — a separate unref'd timer (default 3 min, `deck-heartbeat.json`) that calls the new `coalescer.flushSnapshot()` per armed workspace: only when autonomy ≠ off, brain idle, no pending decision, ≥1 attention pane, not recently woken, and the rate limiter allows. The snapshot prompt (`[fleet-snapshot]` fence) lists every pane currently needing attention with the same fail-closed verdict grammar as edge wakes, so a completion whose edge was lost is recovered within one interval.
- **Fix: TUI agents stuck "running" forever** — an accepted `agent.stop` never broadcast `agentStatus: 'complete'`, and a live TUI's footer repaints keep the activity monitor from ever reaching idle — so OpenCode-style TUI agents showed "running" indefinitely in fleet badges, `pane_list`, and would have been invisible to the heartbeat's attention filter. Turn-end now broadcasts `'complete'` for all hook-bridged agents; the resume chip's typing-safety gate is deliberately unchanged (pinned by test).

## Verification

- Typecheck clean; ~1,500 tests green across the touched suites (new: mirror 21, hook-mirror routing, coalescer rate 5 + snapshot 17, gate/heartbeat 26, status fix 8). Only pre-existing environmental failures remain (`diff.handler` symlink EPERM on Windows).
- Live validation on the owner's machine: OpenCode's bridge `agent.stop`, previously rejected `no-workspace-match` on every fire, resolves `ok` with a ptyId through the mirror path on the new build.
- Loop-cannot-exceed-rate, cap-reject-requeues (not drops), heartbeat skip conditions, and snapshot/edge verdict parity are each pinned by dedicated tests.

## Notes for review

- `maxWakesPerMin` (6) and the heartbeat interval (3 min, 60s floor) are deliberate defaults, both adjustable (`CoalescerDeps`, `deck-heartbeat.json`).
- The mirror is a cache with last-writer-wins full-snapshot replacement — no delta merging, no second source of truth.
- No version bump; CHANGELOG entries under `[Unreleased]` per repo convention. `gen-api-reference` regen produced no content change (no new MCP/pipe surface).

## Summary by CodeRabbit

* **New Features**
  * Added periodic “level review” heartbeat for armed workspaces to surface missed attention.
  * Implemented a fleet-wide cap for concurrent autonomous turns, plus queued turn handling.
  * Added local workspace mirroring (renderer → main) and mirror-first routing for improved responsiveness.
* **Bug Fixes**
  * Fixed TUI agents incorrectly staying “running” after finishing a turn.
  * Improved signal/lifecycle handling so stop/session events transition pane state correctly.
* **Documentation**
  * Updated Windows install instructions to recommend package managers and clarify SmartScreen behavior.


openwong2kim:
Pushed fixes for everything a 3-way review round (Claude / Codex / GLM) surfaced:

- **Mirror fast path now requires an exact ptyId match** (`4bc1b2c5`). A hook from a just-created pane whose renderer push hadn't landed could resolve through the workspaceId fallback to *another pane's* active pty — the same cross-pane resume-binding clobber the env fast path already guards against. ptyId-carrying signals now accept the mirror only when it actually contains that pane; everything else falls through to the pull tiers.
- **Fleet snapshots are surface-accurate** (`ebee61d2`). The UI's pane-level rollup (background tab lights the pane card) was being reused verbatim for the heartbeat snapshot, which could direct the orchestrator at the active tab's pty when the *background* tab was the one awaiting input. Snapshots now emit one row per surface with a retained attention status; single-surface panes are byte-identical to before. Also dropped `agentClockMs` from the push-churn keys (it forced a full push every ~2s while any agent ran) in favor of a 30s bounded refresh.
- **Global turn gate hardened** (`7a1c6c9b`). Slots are token-tracked with a 15-min lease, so a wedged brain turn can no longer hold a fleet slot forever (two wedges used to deadlock all autonomous wakes). One-shot callers — loop kickoff, decision resumes, startup reconciliation — now queue for a slot (120s, FIFO, bounded) instead of silently dropping a `busy`, and startup reconciliation processes every resolved decision serially rather than the first two. Per-workspace busy is checked before the gate so a busy workspace never consumes a fleet slot. The scheduler/coalescer keep their fast reject-requeue path.
- Defensive copies on `WorkspaceMirror` read accessors, and the fleet tail line no longer claims "heartbeat will review" unconditionally.

Typecheck clean; 1,284 tests green across the touched suites. One known deliberate trade-off from the lease design: a genuinely-still-running turn past 15 min transiently exceeds the concurrency cap when its slot is reclaimed — unwedging the fleet won over strict cap enforcement.

openwong2kim:
@codex review

PR #562, opened 2026-07-23
perf: memory and reveal-latency pass — lazy overlays, shared-broker cost, attach snapshot, cold-park hidden workspaces

openwong2kim:
Performance pass across three sprints of the execution plan (`plans/wmux-perf-execution-v2-2026-07-22.md`), one commit per task.

## Sprint 1 — quick wins
- **Lazy overlays** (b398fe3f): the four always-mounted overlays now load and mount only when first opened.
- **Unread selector O(P×N×S) → O(S)** (bfe9577f): unread notifications indexed by surfaceId.
- **Shared MCP broker, Option A** (47a46ef5 + 08621f5b): flag-gated thin shim + one resident host; tool schema shapes hoisted to module scope, dropping the ~17.7 MB per-connection cost to ~0.

## Sprint 2 — attribution + polish
- **Per-pane agent RAM chips in Fleet View** (f8106e4a): process-tree attribution, refreshed on the existing liveness batch — no background polling while the panel is closed.
- **Glyph repaint cadence + WebGL release-on-hide** (68f558d8).

## Sprint 3 — structural
- **Attach snapshot, TASK-10** (5aa761f4): initial-attach flushes ≥256 KB replay through the daemon-side HeadlessSnapshot (same engine as the 3.29 resync path, global concurrency-1) and ship serialized screen state instead of up to 8 MB of raw ANSI — removes the multi-second reveal freeze on long-running agent panes. Same wire format; renderer untouched. Live-delta-during-parse, ring-wrap, and input-ordering guards; no-gain and alt-screen fail-open fallbacks keep the raw path bit-for-bit for anything the snapshot can't reproduce.
- **Cold-park hidden workspaces, TASK-9** (9784776c): workspaces hidden >5 min have their terminals unmounted (daemon PTY + store row survive); renderer RAM stops scaling with hidden-workspace count. Reveal remounts and replays via the attach snapshot above. Cross-pane search and agent screen-reads fall back to a new `daemon.readSessionText` RPC for parked panes, so nothing is silently skipped. Default ON with a persisted escape-hatch flag.

Ordering note: TASK-9 deliberately lands after TASK-10 — parking converts reveals into replays, which is only safe now that large replays are snapshot-fast.

## Verification
- `tsc --noEmit` clean; daemon bundle builds.
- Daemon suite 1275 passed (was 1271 + 4 new); renderer touched suites + 9 new tests green (1782 passed across the combined run).
- Attach-snapshot parity covered against a 1 MB CJK+SGR history; park/unpark state machine and daemon text-snapshot searchability unit-covered.
- Still needs a live pass for the two runtime ACs: reveal <200 ms on a full 8 MB ring, and measured RAM recovery vs the Sprint-0 baseline (requires a packaged build).

No version bump; CHANGELOG entries added under Unreleased.

## Summary by CodeRabbit

* **New Features**
  * Fleet View now displays per-pane agent memory usage (with process attribution).
  * Added “cold park” support for hidden workspaces, including a toggle to control whether terminals unload after inactivity.
  * Added daemon-backed plain-text session reading for parked sessions and large-history reveals.
  * Introduced an optional shared MCP broker mode for better multi-session isolation.

* **Performance**
  * Reduced idle/renderer memory use via deferred overlay loading, lazy WebGL hidden-pane disposal, and incremental notification unread updates.
  * Improved streaming repaint cadence and earlier GPU context release.

* **Bug Fixes**
  * Improved attach/replay behavior to prevent freezes and avoid missing live output during large reveals.


openwong2kim:
[wmux-hermes]

**Priority: P2 — Important**

Comprehensive performance pass across three sprints: lazy overlays, unread selector optimization (O(P×N×S) → O(S)), shared MCP broker (Option A), attach snapshot, and cold-park for hidden workspaces. One commit per task.

The scope is broad and includes architecture-level changes (shared broker, lazy mount). Even though each task is an independent commit, the aggregate regression surface needs careful evaluation. Specifically: verify the shared MCP broker's flag-gate behavior under concurrent access, and check for race conditions in lazy overlay first-mount timing.

If benchmark/profiling results are available, including them as merge evidence would strengthen the PR. Recommend thorough review before merge.

openwong2kim:
@codex review

Issue #557, opened 2026-07-23
[Bug]: Detached sessions are never reaped and are re-spawned on every daemon restart, leaking dozens of orphaned shells


YeongHun-Work:
### What happened?

On Windows, terminal sessions that fall into the detached state (no client attached, but the shell process is still alive) are never garbage-collected, and every daemon restart re-spawns a fresh shell for each of them. There is no TTL for live detached sessions and no config option to cap them, so they pile up indefinitely — the only limit is maxSessions: 200, which blocks new terminals instead of reaping old ones.

On my machine this left 40 orphaned powershell.exe processes holding ~3.3 GB of RAM, none of them referenced by any workspace, all re-created in a single ~3-second burst at the last daemon restart.

Environment: wmux 3.30.0 · Windows 11 Pro (10.0.26200) · Windows PowerShell

Evidence (snapshot)

|Metric|Value|
|---|---|
|Shells spawned by the daemon (powershell.exe, parent = wmux.exe)| 67 |
|Sessions in sessions.json|67 = attached 27 + detached 40|
|PTYs referenced by workspaces (workspace_list)|27 (exactly matches the 27 attached)|
|Live shells referenced by no workspace (detached)|40 (orphans)|
|Approx. RAM held by the 40 orphans|~3.3 GB (~82 MB/shell)|

All 40 detached sessions have a lastActivity equal to the last restart time and freshly assigned PIDs — i.e. the restart re-spawned them.

Root cause

1. The persistence/reaper filter gives detached no TTL (resources/daemon-bundle/index.js):

sessions = state.sessions.filter((s) => {
  const sinceMs = now - new Date(s.lastActivity).getTime();
  if (s.state === "dead")      return sinceMs < s.deadTtlHours * 60 * 60 * 1000;        // 24h
  if (s.state === "suspended") return sinceMs < this.suspendedTtlHours * 60 * 60 * 1000; // 7d
  return true; // attached & detached: kept forever, no TTL
});

A TTL only fires after the shell has already exited (dead) or crashed (suspended). A detached session whose shell is still running is retained unconditionally. deadSessionTtlHours (default 24) does not cover this case, despite the name.

[...]

PR #567, opened 2026-07-23
fix(daemon): restamp corrupted lastActivity instead of leaking sessions past TTL

### LINKED ISSUE #566: Daemon reaper keeps sessions with corrupt lastActivity forever (fail-open NaN)
## Summary

`StateWriter.load()` prunes sessions by TTL using:

```ts
const sinceMs = now - new Date(s.lastActivity).getTime();
```

If `s.lastActivity` is corrupted/malformed, `getTime()` returns `NaN`, so `sinceMs` is `NaN` and every `NaN < ttl` comparison is false. The dead/suspended/detached branches all fall through to keeping the session — meaning a record with a bad timestamp is **kept forever** (fail-open), defeating the #557 leak-prevention reaper for that record.

## Impact

Low. `lastActivity` is only ever written by our own code via `new Date().toISOString()`, and atomicWrite guards against partial writes, so a malformed timestamp requires on-disk corruption or manual editing. But it is a genuine hole in the reaper's leak guarantee.

## Proposed fix

When `sinceMs` is `NaN`, **restamp** `lastActivity = new Date(now).toISOString()` and keep the session for this load. The TTL clock restarts from now, so the record can never leak permanently, and we avoid killing a possibly-live session on a transient bad timestamp (fail-closed would risk reaping an active detached build/tail).

## Notes

Surfaced during an external mirror review of #560. The reviewers proposed a fail-closed NaN guard; we chose restamp-not-prune instead (asymmetric cost — wrongly killing a live session is worse than a slow-aging orphan).

Fix in progress.

openwong2kim:
## Problem

`StateWriter.load()` prunes stale sessions by TTL:

```ts
const sinceMs = now - new Date(s.lastActivity).getTime();
```

If a session's `lastActivity` is corrupted/malformed, `getTime()` returns `NaN`, so `sinceMs` is `NaN`. Every `NaN < ttl` comparison is then false, so dead/suspended/detached records with a bad timestamp are kept forever — a fail-open leak that quietly defeats the #557 idle-session reaper.

## Fix

When `sinceMs` is `NaN`, restamp `s.lastActivity = new Date(now).toISOString()` and keep the session for this load, handled once before the per-state switch.

Restamp rather than prune-on-NaN: pruning would fail-closed and could kill a possibly-live session on a single transient bad timestamp. Restamping restarts the TTL clock from now, so the record survives one bad timestamp but can still age out on a later load and can never leak permanently.

## Tests

Added to `StateWriter.test.ts`:
- A detached session and a dead session with `lastActivity: "not-a-date"` both survive `load()`, and their timestamps come back as valid, recent ISO strings.
- A session whose (real) timestamp is past the detached TTL is still pruned — proving the TTL clock works normally after a restamp and the corrupt path can't leak forever.

`StateWriter.test.ts` and `crashRestore.integration.test.ts`: 45/45 passing.

Closes #566

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed idle-session cleanup when sessions contain malformed or invalid activity timestamps.
  * Corrupted timestamps are now refreshed so sessions expire normally instead of being retained indefinitely.
  * During recovery, refreshed timestamps are saved when applicable so the corrected expiration behavior persists across restarts (including detached sessions).

* **Documentation**
  * Added a changelog entry describing the session expiration fix for the idle-session reaper.


openwong2kim:
[wmux-hermes]

**Priority: P3 — Low (correctness hardening for a narrow fail-open leak)**

This PR cleanly closes #566 by adding a single finite-check before the per-state TTL switch in `StateWriter.load()`: when `sinceMs` is `NaN` (malformed `lastActivity`), it restamps the timestamp to `new Date(now).toISOString()` and keeps the session for the current load, which restarts the TTL clock and guarantees the record can never leak permanently while avoiding the risk of pruning a possibly-live session on a transient bad timestamp. The restamp-not-prune choice is the correct asymmetric trade-off (wrongly killing a live detached build/tail is more costly than a slow-aging orphan), and localizing the guard to one place before the switch keeps the diff minimal and reviewable.

Risks are low: the change is additive and only affects the already-broken NaN path, so well-behaved timestamps are untouched, and the new tests cover both the survival+restamp behavior and the normal post-restamp TTL expiry (proving the record still ages out). Reviewers should confirm that the restamped timestamp is persisted during recovery for detached sessions — CodeRabbit's notes indicate this is handled — and that the `crashRestore.integration.test.ts` coverage exercises the recover-and-save path, not just the in-memory load. Overall a sound, well-scoped fix; safe to merge.

PR #568, opened 2026-07-23
Orchestrator decision authority: resolve rule-answered forks instead of freezing

openwong2kim:
## Why

An `auto`-mode orchestrator can raise a decision the human must settle — and then freeze its loop until someone clicks resolve. In a real incident it stayed frozen for a day on a fork a standing rule already answered: it recommended the correct option, quoted the rule in its own context, and still asked. The mode never reached the brain's prompt (`off`/`assist`/`auto` got byte-identical guidance), the escalation rule was a single vague sentence whose trigger *matched* the fork it should have resolved, and the only rule channel (orchestrator memory) is framed as non-binding background — so the brain had neither the authority, the procedure, nor a rule it was allowed to act on.

## What changed

- **Mode-aware decision authority.** Each turn now leads with an `[autonomy]` block: `auto` is told it has authority to resolve forks from standing rules and escalate only genuine residuals; `assist` reports and recommends; `off` gets no ambient instructions. Previously the mode never reached the prompt at all.
- **Resolve-first boundary.** The escalation guidance is now a procedure, not a proverb: check the binding policy → standing conventions → memory *before* raising a decision, and a choice a standing rule already answers is explicitly **not** a genuine fork.
- **Binding operator policy (`deck-policy.md`).** A new operator-editable file of authoritative standing rules the brain may act on and cite — deliberately distinct from orchestrator memory (which stays non-executable background). Rules can't grant tools or override safety; risky/irreversible actions still require a human.
- **Decision TTL + self-resolve.** A pending decision no longer blocks forever: after a TTL (default 30 min) the heartbeat wakes the brain for a *re-examine* turn — restate with better context, or, in `auto` only, resolve its own decision via the new `deck_resolve_decision` tool by citing the rule that settles it. Server-enforced (auto mode + TTL elapsed + a substance floor + a turn-scoped lease, all checked in the daemon RPC — not just the tool description).
- **Learnings loop.** When the operator's answer corrects the brain's judgment, it persists a memory fact so it stops re-raising that class of question.

## Safety properties

- Ordinary wakes stay blocked while a decision is pending — the TTL re-examine is the only wake past that gate, and it still honors the global auto-wake switch and the consecutive-wake budget.
- Self-resolve requires a live re-examine lease bound to `(workspace, decision id)`, so an ordinary turn (e.g. a human chat while a decision is pending) can't slip a self-resolution through.
- A human's resolution is never lost to a race: resolution provenance (`human`/`brain`), a compare-and-swap on stale-question replacement, and an errored-turn guard together ensure a mid-turn human answer always survives to a resume turn rather than being consumed unseen.
- A lost decision timestamp fails closed (human-only). Risky/irreversible decisions are never self-resolvable.

## Verification

- Full test suite green; new coverage for every gate and race above, plus a **named regression** that replays the exact `deck-decisions.json` record that froze the orchestrator and asserts the turn now carries all three previously-absent ingredients (auto authority, the binding worktree rule, the resolve-first system prompt) — and that `off` mode still receives none of it.
- The diff went through six independent adversarial review rounds (Codex + GLM-5.2); the self-resolve gate in particular surfaced a chain of "lose the human's answer" races that are each closed and tested.

## Verified in a packaged build

Resolved the seed question from a live dogfood: a **packaged build** (`npm run package`, run isolated via `WMUX_DATA_SUFFIX`) seeds `deck-policy.md` correctly on a fresh profile with the exact binding worktree rule, and boots clean with no errors. The earlier non-appearance was a dev-server (electron-forge/Vite) main-process eval-timing artifact, not a product defect — the `mkdir -p` hardening is still in and unit-tested. Feature is fail-open either way.

## Not in scope

A soft, non-freezing `NEED:`-style escalation tier (agent-deck's model) — deferred until this lands and a real "needed it to proceed without me" case is recorded. The hard gate stays for irreversible actions.

## Summary by CodeRabbit

* **New Features**
  * Command Deck now resolves questions from standing policy and known context before escalating.
  * Pending decisions automatically become stale after a configurable TTL (30 minutes by default).
  * In `auto` mode, stale decisions may be safely self-resolved; other modes continue to require human action.
  * Added policy guidance and a new decision-resolution command.
  * Added decision status and resolution provenance indicators.

* **Bug Fixes**
  * Prevented automatic decision-making from freezing on resolvable questions.
  * Improved visibility when a command turn fails or ends unexpectedly.
  * Added safeguards against conflicting or outdated decision updates.


openwong2kim:
Addressed the CodeRabbit review:

- **Backslash mangling in the frozen-fork fixture** (real catch — thanks): the Windows paths used single backslashes, so JS parsed `\t`/`\w` as invalid escapes and silently dropped the separators. Doubled them so the fixture matches the actual `deck-decisions.json` record it reproduces.
- **Empty fake stubs**: annotated `interrupt()`/`dispose()` with a no-op comment to match this file's existing `FakeAdapter` convention.
- **Empty `.catch(() => {})` handlers in deck.handler.ts**: keeping these as-is — it's the established fire-and-forget style used throughout this file (and the wider codebase) for best-effort cleanup, and lint passes on it. Diverging just these sites would break consistency.

eslint is clean (0 errors) and the touched suites are green.

PR #574, opened 2026-07-24
fix(cli-shim): fail closed on an untrusted PATH read instead of wiping the user PATH

### LINKED ISSUE #573: CLI shim PATH edit wipes the entire user PATH under ConstrainedLanguage PowerShell (read fails, write proceeds anyway)
## Summary

Installing wmux can replace the user's **entire** `HKCU\Environment\Path` with a single entry — the wmux `bin` directory — destroying every pre-existing PATH entry. Any npm-global CLI (`%APPDATA%\npm`), language toolchain, or hand-added path is silently gone, and the previous value is unrecoverable.

```
PS> claude
claude : 'claude' is not recognized as the name of a cmdlet, function, script file, or operable program.

PS> reg query "HKCU\Environment" /v Path
    Path    REG_EXPAND_SZ    C:\Users\<user>\AppData\Local\wmux\bin
```

The `claude.exe` binary is still on disk at `%APPDATA%\npm\claude.exe`. It is not a deletion — it is PATH loss.

## Root cause (confirmed by repro, **not** `setx`)

The obvious first guess is that the installer wrote PATH with `setx` / a blind overwrite. **That is not what happens.** There is no `setx` anywhere in the repo, and `buildPathEditScript` in `src/main/cliShim.ts` is a correct read-append-write.

The real cause is that the script **reads the registry with a .NET method call but writes it back with a cmdlet**, and the two have different failure conditions:

```powershell
$key  = [Microsoft.Win32.Registry]::CurrentUser.OpenSubKey('Environment', $true)   # (1)
$cur  = [string]$key.GetValue('Path', '', [...]::DoNotExpandEnvironmentNames)      # (2)
$parts = @($cur -split ';' | Where-Object { $_.Trim().Length -gt 0 })              # (3)
if (-not $hit) { $parts += $bin; $changed = $true }                                # (4)
Set-ItemProperty -Path 'HKCU:\Environment' -Name 'Path' -Value ($parts -join ';') -Type ExpandString   # (5)
```

Under **ConstrainedLanguage** PowerShell — how AppLocker/WDAC lock down enterprise machines — arbitrary .NET method invocation is blocked, but cmdlets keep working. So:

1. `(1)` throws `MethodInvocationNotSupportedInConstrainedLanguage`; `$key` stays `$null`.
2. `(2)` throws `InvokeMethodOnNull`; `$cur` stays `$null`.
3. There is no `$ErrorActionPreference = 'Stop'` and no `try`/`catch`, so **execution continues**. `(3)` resolves the user's current PATH to an empty array.
4. `(4)` makes that array `@($binDir)`.
5. `(5)` is a cmdlet, so it is **not** blocked. It persists `<binDir>` as the user's entire PATH, as `REG_EXPAND_SZ` — matching the observed registry evidence exactly.

The defect class is *"a read-modify-write whose read can fail silently, and whose write proceeds anyway."* ConstrainedLanguage is the trigger seen in the field; a null key or a denied ACL would produce the same wipe.

### Reproduction

Running the **real** generated script (via the real `execFileSync` spawn path) against a throwaway HKCU key seeded with `C:\Program Files\Git\cmd;%USERPROFILE%\AppData\Roaming\npm;C:\Program Files\nodejs;C:\Windows\System32`:

| Mode | Result |
| --- | --- |
| FullLanguage | appended correctly, nothing lost |
| **ConstrainedLanguage** | **`Path` = `C:\...\wmux\bin`, 4/4 pre-existing entries destroyed** |

First error emitted before the destructive write:

```
$key = [Microsoft.Win32.Registry]::CurrentUser.OpenSubKey(...)
    + FullyQualifiedErrorId : MethodInvocationNotSupportedInConstrainedLanguage
$cur = [string]$key.GetValue('Path', ...)
    + FullyQualifiedErrorId : InvokeMethodOnNull
```

To force ConstrainedLanguage for the repro, prefix the script with
`$ExecutionContext.SessionState.LanguageMode = 'ConstrainedLanguage'`.
(Note: the `__PSLockdownPolicy` environment variable is ignored on Windows 10+ and does **not** work for this.)

## Environment

- Windows 10/11, any machine where PowerShell runs in ConstrainedLanguage (AppLocker / WDAC / Defender Application Control).
- Reachable from every install channel that runs the Squirrel install/update hooks: Setup.exe, winget, and app self-update — `installCliShim()` runs on `--squirrel-install` and `--squirrel-updated`.
- Present on `main` as of 3.32.0.

## Why the "obvious" fix is wrong

The natural-looking remedy — switch to `[Environment]::GetEnvironmentVariable` / `SetEnvironmentVariable` — is a regression on two counts:

- `SetEnvironmentVariable` demotes the value from `REG_EXPAND_SZ` to `REG_SZ`, permanently baking in any `%VAR%` entries the user had.
- `System.Environment` is *also* a .NET type, so both calls are blocked under ConstrainedLanguage anyway (verified). It would fail identically.

`Get-ItemProperty` is not a valid reader either: it works under ConstrainedLanguage but **expands** `REG_EXPAND_SZ`, so `%USERPROFILE%\...` would be rewritten as a literal path.

## Fix

`reg.exe` is an external process, so it is immune to language mode, and unlike `Get-ItemProperty` it returns the **raw, unexpanded** value. Using it as a read fallback makes the edit work correctly under ConstrainedLanguage instead of merely failing safely. On top of that, the write must be gated:

1. **Fail closed** — `$ErrorActionPreference = 'Stop'` plus explicit `try`/`catch`; if *both* readers fail, exit without touching the registry.
2. **Structural invariant** — assert between read and write that no pre-existing entry other than `binDir` disappears; abort otherwise. Any future bug that mis-resolves the current PATH aborts here instead of persisting.
3. **Backup** — copy the pre-edit raw value to `HKCU:\Software\wmux\UserPathBackup` before writing (not into `HKCU:\Environment`, where a stray value would become a real environment variable).
4. **Isolate the WM_SETTINGCHANGE broadcast** — `Add-Type` is blocked under ConstrainedLanguage; a cosmetic notification failure must not be reported as an edit failure (today it makes the process exit non-zero even when the write succeeded).

## Recovery for already-affected users

The old code kept no backup, so the previous value cannot be restored from wmux. Options, best first:

1. **System Restore** — if a restore point predates the install, it carries the user hive.
2. **Reconstruct manually.** Most entries are re-derivable; the common one is npm globals:
   ```powershell
   $p = [Environment]::GetEnvironmentVariable('Path','User')
   $add = "$env:APPDATA\npm"
   if (($p -split ';') -notcontains $add) { [Environment]::SetEnvironmentVariable('Path', "$p;$add", 'User') }
   ```
   Machine PATH (`HKLM`) is untouched, so system tools still resolve — only user-scoped entries were lost.
3. Re-running an installer for a given tool generally re-adds its own entry.

## Verified NOT affected

`install.ps1` (source-build path) uses `[Environment]::GetEnvironmentVariable` / `SetEnvironmentVariable`, which look like the same read-then-write shape. They are **not** destructive here: both are `System.Environment` statics, so under ConstrainedLanguage the read and the write are blocked *together* and no write occurs. It is the only other PATH-writing site; `src/` contains exactly one registry write (`cliShim.ts`).

## Follow-up (separate issue)

`wmux doctor` could detect the wiped state (user PATH consisting solely of the wmux bin dir, or missing `%APPDATA%\npm` while npm globals exist) and surface the recovery steps above.

openwong2kim:
Closes #573.

Installing wmux on a machine where PowerShell runs in **ConstrainedLanguage** (AppLocker/WDAC — standard enterprise lockdown) replaced all of `HKCU\Environment\Path` with a single entry, the wmux `bin` directory. Every pre-existing user PATH entry was destroyed, unrecoverably.

## Root cause

Not an overwrite, and not `setx` — there is no `setx` in the repo. The bug is an **asymmetry between how the script reads and how it writes**:

```powershell
$key   = [Microsoft.Win32.Registry]::CurrentUser.OpenSubKey('Environment', $true)   # .NET  — blocked
$cur   = [string]$key.GetValue('Path', '', [...]::DoNotExpandEnvironmentNames)      # .NET  — blocked
$parts = @($cur -split ';' | Where-Object { $_.Trim().Length -gt 0 })               # → @()
if (-not $hit) { $parts += $bin; $changed = $true }                                 # → @($bin)
Set-ItemProperty -Path 'HKCU:\Environment' ... -Type ExpandString                   # cmdlet — NOT blocked
```

ConstrainedLanguage blocks arbitrary .NET method invocation but leaves cmdlets working. With no `$ErrorActionPreference = 'Stop'` and no `try`/`catch`, execution sails past `MethodInvocationNotSupportedInConstrainedLanguage` and the follow-on `InvokeMethodOnNull`, resolves the current PATH to an empty array, and persists `$bin` as the user's entire PATH — as `REG_EXPAND_SZ`, matching the reported registry evidence exactly.

The defect class is *a read-modify-write whose read can fail silently and whose write proceeds anyway*. ConstrainedLanguage is the field trigger; a null key or a denied ACL produces the same wipe.

## Approach

`reg.exe` is an external process, so it is immune to language mode, and unlike `Get-ItemProperty` it returns the value **raw and unexpanded**. Using it as the read fallback means the edit now *works correctly* under ConstrainedLanguage rather than merely failing safely.

Two alternatives were evaluated and rejected:

- `[Environment]::Get/SetEnvironmentVariable` — the natural-looking fix, but `SetEnvironmentVariable` demotes `REG_EXPAND_SZ` → `REG_SZ`, permanently baking in the user's `%VAR%` entries; and `System.Environment` is itself a .NET type, so both calls are blocked under ConstrainedLanguage anyway (verified). It would fail identically.
- `Get-ItemProperty` as the reader — works under ConstrainedLanguage, but **expands** `REG_EXPAND_SZ`, rewriting `%USERPROFILE%\...` as a literal path.

## Changes

`src/main/cliShim.ts` — every guard below corresponds to a rule the original violated:

1. **Fail closed.** `$ErrorActionPreference = 'Stop'`, explicit `try`/`catch` around each reader, and if *both* fail the script exits `READ_FAILED` (10) without touching the registry. The whole-key `reg query` (rather than `/v Path`) keeps "value absent on a fresh profile" distinguishable from "key unreadable".
2. **Structural invariant.** Between read and write, assert that no pre-existing entry other than `binDir` disappears; otherwise exit `INVARIANT_VIOLATED` (11). A future bug that mis-resolves the current PATH aborts here instead of persisting.
3. **Backup.** The pre-edit raw value is copied to `HKCU:\Software\wmux\UserPathBackup` first — deliberately *not* under `HKCU:\Environment`, where a stray value would become a real environment variable for the user.
4. **Isolated broadcast.** `Add-Type` is blocked under ConstrainedLanguage, and the `WM_SETTINGCHANGE` call was making the process exit non-zero even when the write had succeeded. It is now best-effort and cannot mask a good write.
5. `explainPathEditExit()` turns a deliberate bail-out into an actionable warning ("add `<binDir>` to your PATH manually") instead of an opaque `execFileSync` throw; unexpected failures still propagate.

## Testing

`buildPathEditScript` gains test-only `subKey` / `backupSubKey` seams so the regression test can run **the real generated script through the real `execFileSync` spawn path** against a throwaway HKCU key — proving the property that matters, not just that a guard string is present.

Seeded with `C:\Program Files\Git\cmd;%USERPROFILE%\AppData\Roaming\npm;C:\Program Files\nodejs;C:\Windows\System32`:

| Mode | before this PR | after |
| --- | --- | --- |
| FullLanguage | appended correctly | appended correctly |
| **ConstrainedLanguage** | **`Path` = `…\wmux\bin`, 4/4 entries destroyed** | appended correctly, `%USERPROFILE%` preserved unexpanded |

Live tests cover, under **both** language modes: add appends without loss, add is idempotent, remove strips only the bin entry — plus an unreadable key exiting `READ_FAILED` while creating nothing, and the backup being written. `describe.skipIf(process.platform !== 'win32')` keeps the macOS/Linux CI baseline green.

```
Tests  23 passed | 8 skipped (31)     # 8 skipped = existing darwin suite
tsc --noEmit: clean
eslint: 0 errors (4 pre-existing duplicate-import warnings, unchanged)
```

## Scope

`src/` contains exactly one registry write, so the blast radius is this file. `install.ps1` (source-build path) looks like the same shape but was verified **not** destructive: its `GetEnvironmentVariable` / `SetEnvironmentVariable` are both `System.Environment` statics, so under ConstrainedLanguage they block together and no write occurs.

Users already hit by this cannot be repaired by wmux — the old code kept no backup. Recovery guidance is in #573; a `wmux doctor` detector is noted there as follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows installer PATH updates in PowerShell constrained environments by failing safely when PATH cannot be reliably read.
  * Prevents existing PATH entries from being overwritten or truncated; ensures only the wmux shim entry can be changed.
  * Adds a backup of the prior PATH before applying updates.
  * Corrected installer status/reporting behavior tied to the environment-change broadcast.
* **Tests**
  * Expanded unit coverage for generated PATH-edit behavior and added a Windows integration-style validation for constrained and normal modes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Installing wmux can no longer wipe your entire user PATH.** On a machine where PowerShell runs in ConstrainedLanguage — how AppLocker/WDAC lock down enterprise Windows — the CLI shim's PATH edit could replace all of `HKCU\Environment\Path` with a single entry, the wmux `bin` directory. Every other entry was destroyed and unrecoverable: npm-global CLIs (`%APPDATA%\npm`), toolchains, everything user-scoped. The cause was an asymmetry rather than an overwrite: the script *read* the registry with a .NET method call (blocked in ConstrainedLanguage) but *wrote* it back with a cmdlet (not blocked), and with no `$ErrorActionPreference = 'Stop'` it sailed past both read errors, resolved the current PATH to empty, and persisted the bin directory as the whole thing. The edit is now fail-closed and, more importantly, actually works under ConstrainedLanguage: the raw value is read via `reg.exe`, which is immune to language mode and — unlike `Get-ItemProperty` — does not expand `%VAR%` entries and bake them in. If every read strategy fails the registry is left untouched, a structural invariant aborts any edit that would drop an entry other than the bin directory, the pre-edit value is backed up to `HKCU:\Software\wmux\UserPathBackup`, and the cosmetic `WM_SETTINGCHANGE` broadcast can no longer report a successful write as a failure. The source-build `install.ps1` path was verified unaffected. (#573)
+
 ## [3.32.0] — 2026-07-24
 
 ### Added
--- a/src/main/__tests__/cliShim.test.ts
+++ b/src/main/__tests__/cliShim.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { buildShimCmd, buildPathEditScript, deriveShimPaths } from '../cliShim';
+import { buildShimCmd, buildPathEditScript, deriveShimPaths, explainPathEditExit } from '../cliShim';
 
 describe('buildShimCmd', () => {
   it('discovers app-* dynamically, scopes ELECTRON_RUN_AS_NODE, and forwards args + exit code', () => {
@@ -40,7 +40,7 @@ describe('buildPathEditScript', () => {
     expect(script).toContain("'Environment'");
     // Idempotency: only writes when membership actually changes
     expect(script).toContain('if (-not $hit) { $parts += $bin; $changed = $true }');
-    expect(script).toContain('if ($changed) {');
+    expect(script).toContain('if (-not $changed) { exit 0 }');
   });
 
   it('remove: filters only the exact bin entry', () => {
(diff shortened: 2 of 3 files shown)


openwong2kim:
## [wmux-hermes] Code Review

**Priority: P1 (Critical) — fixes a P1 data-loss bug (#573), recommend merge after review.**

### What this PR changes

This is a ground-up rewrite of the PATH edit logic in `src/main/cliShim.ts` that converts a silent-data-loss path into a fail-closed one. The core fix is sound and well-reasoned: the read is now a two-strategy waterfall — `.NET` raw read first (fast path), `reg.exe` fallback second (immune to ConstrainedLanguage and returns the value unexpanded, unlike `Get-ItemProperty`). If both fail, the script exits `READ_FAILED` (10) without touching the registry. Between read and write, a structural invariant asserts that no pre-existing entry other than `binDir` disappears (exit `INVARIANT_VIOLATED` / 11). The pre-edit raw value is backed up to `HKCU:\Software\wmux\UserPathBackup` (correctly placed outside `HKCU:\Environment` to avoid creating a bogus env var). The `WM_SETTINGCHANGE` broadcast is isolated in `try/catch` so its failure (e.g. `Add-Type` blocked under ConstrainedLanguage) can no longer mask a successful write. `runPathEdit` now catches the deliberate bail-out codes and warns instead of throwing, while unexpected failures still propagate. The CHANGELOG entry is thorough and accurate.

### Strengths

The test suite is the standout. The string-assertion tests prove the guards are *present*, but the live registry tests (`describe.skipIf(process.platform !== 'win32')`) actually run the **real generated script through the real `execFileSync` spawn path** against a throwaway HKCU subkey, in **both** FullLanguage and ConstrainedLanguage modes — proving the property that matters (no PATH destruction) rather than just that a regex matches. Seeding with a `%USERPROFILE%` entry and asserting it survives unexpanded is exactly the right edge case. The `subKey`/`backupSubKey` test seams are clean and well-documented ("Production callers must never pass them").

### Potential risks and feedback

1. **reg.exe output parsing fragility (minor):** the regex `'^\s+Path\s+(REG_\S+)\s{4}(.*)$'` assumes exactly 4 whitespace characters between the type token and the value. This matches the stable `reg.exe` output format (empirically validated on `windows-latest`), but it is the kind of assumption that could break on a future Windows version or a localized build. Consider making the separator `\s+` instead of `\s{4}` — there's no ambiguity since `REG_*` type names contain no spaces, so any run of whitespace cleanly delimits type from value. Non-blocking.

2. **Backup is best-effort (defensible, worth noting):** the backup write is wrapped in `try { ... } catch { }`, so the main write proceeds even if the backup fails. This is a reasonable trade-off — the structural invariant already prevents destructive writes, so the backup is a belt-and-suspenders safety net rather than the primary guard — but it means a user whose backup silently fails has no recovery if a future bug somehow bypasses the invariant. Acceptable for this PR; just flagging the design decision.

3. **Test-only params on a production function (cosmetic):** `subKey`/`backupSubKey` on the exported `buildPathEditScript` is a minor design smell. A stricter pattern would use a separate internal builder or a branded parameter type, but given the function has a single production caller (`runPathEdit`) that never passes them, and the JSDoc is explicit, this is fine.

4. **Exit code collision surface (negligible):** codes 10 and 11 are arbitrary but unlikely to collide with PowerShell's own exit codes in this context. No action needed.

### Recommendation

**Approve and merge.** This is the correct fix for a critical bug, the approach (reg.exe fallback + fail-closed + invariant + backup) addresses every failure mode identified in #573, and the test coverage — especially the live ConstrainedLanguage regression test — is exemplary. The nitpicks above are all non-blocking. The only real follow-up is the `wmux doctor` detector for already-affected users (noted in #573), which is out of scope here.

PR #578, opened 2026-07-24
wmux web: phone-usable browser terminal, grouped by workspace, started from the GUI

openwong2kim:
## What this is

`wmux web` began as a read-only browser view of a single pane. This turns it into a phone-usable surface for watching and steering the fleet, started from the GUI, with the read-only default preserved and enforced server-side.

Everything runs **inside the daemon** — it survives the desktop app closing, and it never contends with the GUI for a pane (it tees the existing non-exclusive PTY fan-out rather than opening its own).

## Highlights

**Reach it without a keyboard fight**
- A **6-character pairing code** (`/pair`, single-use, 10-min TTL) replaces typing a 36-character token on a phone. The connect popover is split into two labelled paths — *open on this machine* (a clickable token URL) and *on your phone* (a token-free `/pair` address + the code) — with a copy button per value and a "New code" action when a code is spent.

**Actually usable on a touch screen**
- A **touch key bar**: `Esc Tab Ctrl Alt ← ↑ ↓ → | / ~ - _ Home End PgUp PgDn`, with tap-to-arm / double-tap-to-lock modifiers, plus an agent row (`Shift+Tab`, `Ctrl+C`, `yes`, `continue`, `/compact`, `/clear`, `/resume`).
- **Font size + fit/zoom**: fit-to-width for a glance, or 1:1 with sideways scroll to read.

**See the whole fleet, organised**
- The session sheet is **grouped by workspace** (header + count, panes indented, fleet order preserved). Panes from before the workspace stamp fall into an honest "workspace unknown" bucket; a legacy-only setup renders flat with no header.
- Panes are named **agent → program → cwd** through one shared label helper, so a pane can't be called different things in the switcher, the fleet chips, the sheet, and the split tiles.
- **Split view** (1/2/4) opens one stream + one xterm per tile, staying inside the selected pane's workspace. Below 700px it forces 1-up — two 100-col terminals on a phone are unreadable.

**In-app notifications**
- The server tees `session:critical` (approval requests) and `session:notification` onto every SSE client, so a viewer on pane A still hears that pane B needs an answer. Tap the banner to jump; the tab title carries a badge.

**Start it from the GUI**
- A quiet **`web` toggle** on the sidebar rail beside Agent and Git (amber while running) opens a popover that drives `daemon.web.*`. `src/shared/web.ts` is the one contract shared by daemon RPC, the main IPC handler, and the renderer.

## Security

Read-only is **not** a UI state: `POST /api/input` returns 403 unless the server was started with `--allow-input`, so a tampered page or a raw `curl` still cannot type — the trust boundary is the `pty.write` gate. On top of that, from the security review (`docs/web-security-review-2026-07-24.md`, P1+P2): frame protection + CSP `frame-ancestors 'none'` + `nosniff` + `Referrer-Policy: no-referrer` on every response, a `Host` allowlist (DNS-rebinding guard), and pairing-code regeneration. Only the workspace **name** and the program **basename** are surfaced from a pane — never its full env or command line (tested: credentials and arguments never reach the wire).

## Verification

- **322 tests** pass (web server, IPC handler, WebToggle, resolveSpawnEnv); `tsc --noEmit` clean; changed files lint clean.
- **Live end-to-end dogfood against a real daemon and a real PTY**: paired from an emulated phone with the 6-char code, confirmed workspace labels, drove the key bar to run `echo <marker>` and saw the real PowerShell answer, exercised split view and the phone guard, verified the grouped sheet and same-workspace split fill.

Bugs the dogfood caught and this PR fixes: fit/zoom was cancelled by CSS-scaling the terminal; the service worker cached the shell forever so an installed PWA could never update; `shellLabelOf` returned "Program" for every pane on the real (unquoted, spaced) Windows path; split view mixed workspaces.

## Not covered

- Real phone **hardware** (all mobile checks were a 390×844 emulated viewport).
- **Web push** (needs HTTPS; in-app notifications already cover the tab-open case). Follow-ups W1 (`--expose` TLS) and W4 (full CSP) remain, and the orchestrator is out of reach from the daemon-hosted surface by design.

See `docs/wmux-web-handoff-2026-07-24.md` for the full state and the rebase/verification notes.

## Summary by CodeRabbit

- **New Features**
  - Added `wmux web` for viewing terminal panes from a browser or phone.
  - Includes read-only-by-default access, secure token and pairing-code authentication, session switching, notifications, touch controls, adjustable text sizing, and split-pane views.
  - Optional keyboard input and network exposure can be enabled explicitly.
  - Added titlebar/sidebar controls to start, stop, pair, copy access links, and refresh pairing codes.
  - Web assets are now included in packaged builds.

- **Documentation**
  - Added security review and handoff documentation for the web experience.


src/daemon/web/WebTerminalServer.ts
```diff
@@ -0,0 +1,753 @@
...
+    const p = url.pathname;
+
+    // Reject anything addressed to a host we do not serve (DNS-rebinding guard).
+    // Checked before routing so it also covers the unauthenticated /api/pair.
+    const hostHeader = req.headers.host ?? '';
+    // Strip the port; keep IPv6 literals ("[::1]:7681") intact.
+    const hostname = hostHeader.startsWith('[')
+      ? hostHeader.slice(0, hostHeader.indexOf(']') + 1).toLowerCase()
+      : hostHeader.split(':')[0].toLowerCase();
+    if (!this.allowedHosts.has(hostname)) {
```
openwong2kim, line 330: **Attacker note: this is a DNS-rebinding guard only — it does not block the direct-IP localhost-CSRF class.**

A malicious page on any public origin can embed ``. The browser sends `Host: 127.0.0.1:7681`, this check passes (`127.0.0.1` is in `allowedHosts`), and `/api/pair` — the one unauthenticated route — processes the request. Five such image loads burn the pairing code. No DNS rebinding is needed; the attacker uses the loopback literal directly.

The web token itself stays safe: CORS blocks the cross-origin page from reading the `{token}` response body, so this vector **cannot exfiltrate the credential**. The realizable impact is burning the pairing code, which the W5 regen cooldown (`PAIR_REGEN_COOLDOWN_MS = 30s`) bounds to a ~30s phone-pairing DoS. Firefox and Safari do not enforce Private Network Access, so the vector is live on them today (recent Chrome's PNA rollout may block the subresource fetch).

The PR description and the in-repo security review frame W3 as "DNS-rebinding pairing DoS" and ship the Host allowlist as the fix — but the allowlist only covers the rebind variant (Host = attacker domain). The direct-IP variant (Host = `127.0.0.1`) walks straight through.

Two acceptable resolutions:
- **Close it:** add an `Origin` / `Sec-Fetch-Site` check on `/api/pair` (accept same-origin, or `null`/absent for non-browser clients like the CLI). ~5 lines, kills the cross-origin burn on every browser regardless of PNA.
- **Document the residual:** if the ~30s DoS is acceptable, note in the W3 writeup that the Host check is narrowly a DNS-rebinding control, not a general CSRF defense, so future readers don't assume `/api/pair` is unreachable from arbitrary web content.

openwong2kim (commented):
Attacker-perspective review. One residual worth flagging on the unauthenticated `/api/pair` route. (Note: the SSE security-headers gap was already raised by CodeRabbit — not duplicating.)

No findings on PR #579 from an attacker angle — the boot-time refresh is narrowly scoped (only acts when `settings.json` already references the wmux-owned file), never writes `settings.json`, and the bundled source is already part of the trusted app image. The tmp+rename write is atomic and actually *removes* any locally-planted malicious bridge as a side effect.

openwong2kim:
@codex review

openwong2kim (commented):
**[wmux-hermes] Code review — P2 (significant new feature; merge after review)**

This is a large (5,729-line) but well-structured PR that turns `wmux web` from a read-only single-pane view into a phone-usable fleet surface served *inside the daemon* (so it survives the desktop app closing and never contends with the GUI — it tees the non-exclusive `DaemonPTYBridge` fan-out rather than opening its own PTY). The core additions are: a 6-char single-use pairing code (`/pair`, 10-min TTL, 5-attempt burn) to avoid typing a 36-char token on a phone; a touch key bar + agent row; workspace-grouped session sheet; split view (1/2/4) with a <700px phone guard; in-app notifications (SSE-tee'd `session:critical`/`session:notification`); and a quiet `web` toggle on the sidebar rail.

**Security model is the strongest part of this PR and was clearly reviewed** (`docs/web-security-review-2026-07-24.md` is included). The read-only default is enforced *server-side* (`POST /api/input` returns 403 unless started with `--allow-input` — not a UI state), so a tampered page or raw `curl` cannot type. Token handling is solid: a fresh random web token is minted per start (the daemon master token never touches the network), compared timing-safe via `crypto.timingSafeEqual`, and the SSE stream is the only route allowed to carry `?token=` in the query (EventSource can't set headers) — every other endpoint is Bearer-only, keeping the token out of URL/proxy logs. The DNS-rebinding guard (Host allowlist of loopback + concrete bound addresses, checked before routing including the unauthenticated `/api/pair`) is the right defense. Baseline headers on every response (`X-Frame-Options: DENY`, CSP `frame-ancestors 'none'`, `nosniff`, `Referrer-Policy: no-referrer`) match the hardening Docker/VS Code/Electron apply. Input is capped at 64KB to prevent memory exhaustion. Only the workspace *name* and program *basename* are surfaced from a pane — never full env or command lines (which can carry credentials), and this is explicitly tested. Pairing-code regeneration is rate-limited (30s cooldown) so an attacker who can burn codes can't spin a regen loop.

**Implementation observations and potential risks:**
- The `bridge.setMaxListeners(0)` per SSE client is the right call (each viewer adds a listener), but worth confirming there's no unbounded memory growth if many viewers connect to one pane (each holds a `res` with buffered SSE data). The `close`/`detach` cleanup on disconnect looks correct.
- `workspaceLabelOf` reads `WMUX_WORKSPACE_NAME` from the persisted env — an honest limitation noted in the code: panes spawned before the env key existed have no name (frontend falls back to cwd), and a post-spawn rename isn't re-stamped. This is acceptable (better than fabricating), but worth a release note so users aren't surprised.
- All mobile verification was on an emulated 390×844 viewport — no real phone hardware. The PR is honest about this. The service-worker-cached-forever bug (caught in dogfood) and the CSS-scaling-canceling-fit/zoom bug are both fixed.
- Follow-ups W1 (`--expose` TLS) and W4 (full CSP) are explicitly left open — the current CSP is `frame-ancestors 'none'` only, not a full `default-src` policy, which is fine for an MVP but should be tightened before any `--expose` path ships.

322 tests pass, `tsc --noEmit` clean, lint clean. Recommendation: **merge after review** — this is a significant, security-conscious feature with thorough test coverage and honest scope documentation.

openwong2kim:
[wmux-hermes]

**Priority: P2 — Important**

**What the PR changes.** This is a substantial feature PR (~5,700 lines across 28 files) turning `wmux web` from a read-only single-pane view into a phone-usable fleet surface hosted *inside the daemon* (so it survives the desktop closing and never contends with the GUI for a pane — it tees the non-exclusive `DaemonPTYBridge` fan-out). Highlights from the diff: a full `WebTerminalServer.ts` (753 lines, Node `http` + SSE, no new deps) with a 6-char single-use pairing code (`/api/pair`, 10-min TTL, 5-attempt burn, rate-limited regeneration), a touch key bar + agent keystroke row in `app.js` (1,173 lines), workspace-grouped session sheet via a shared label helper (`workspaceLabelOf` reading only `WMUX_WORKSPACE_NAME`, `shellLabelOf` surfacing only the program basename), split view (1/2/4) with a <700px phone guard, in-app notifications teed from `session:critical`/`session:notification`, and a `WebToggle` popover in the sidebar. The security posture is solid by default: read-only enforced server-side at the `pty.write` boundary (403 unless `--allow-input`), loopback-only bind unless `--expose`, per-start rotating UUID token (master token never networked), `timingSafeEqual` comparisons, Bearer-only auth on non-SSE routes, 64 KB input cap, and a `Host` allowlist + `frame-ancestors 'none'`/`nosniff`/`no-referrer` headers addressing the review findings (W2/W3) inline.

**Risks / edge cases worth noting.** The security review doc (`docs/web-security-review-2026-07-24.md`) is included and is refreshingly candid — it inventories the residual risks honestly: **W1** `--expose` sends token + full scrollback in cleartext over HTTP (the warning could be stronger, and auto-TLS/`tailscale serve` is a follow-up); **W4** a full CSP `connect-src` is still a follow-up (currently only `frame-ancestors 'none'`); **W5** pairing-code regeneration after burn/expiry is now handled (rate-limited), which was an availability gap. The biggest live risk is operator-misuse of `--expose` on untrusted networks — the scrollback routinely contains secrets, and a sniffed token grants full read. The PR is correct to keep read-only as the enforced default and to document this loudly, but the cleartext warning tier (A) should probably land before this ships broadly. A correctness edge in `shellLabelOf`: the Windows path parsing heuristic (quoted span → exe-extension boundary → first whitespace token) is reasonable but could mislabel an unusual invocation; it only affects a display label, so impact is cosmetic. The `service worker` caching bug (PWA could never update) is noted as fixed.

**Implementation feedback.** The server code is careful and idiomatic: `clientError`/`closeAllConnections` prevent daemon crashes/hangs, SSE heartbeat is `.unref()`'d, the `Host` allowlist correctly includes `[::1]` IPv6 literals and the bound host, and the strip-port logic handles both IPv4 and IPv6. The pairing implementation is sound — `crypto.randomBytes`, single-use burn on success, attempt-budget decrement on miss, and a cooldown on lazy regeneration to prevent an oracle loop. Test coverage: 322 tests pass per the body, including `WebTerminalServer.test.ts` (374 lines) and `WebToggle.test.tsx`. The honest "Not covered" list (real phone hardware, web push, W1/W4 follow-ups, orchestrator out-of-reach by design) is the right framing. One minor structural note: `WebTerminalServer.ts` is a big single file (753 lines) — the SSE/input/pairing/helpers concerns are well-commented but could be split if it grows further; not a blocker now. Naming (`shellLabelOf`/`workspaceLabelOf` as pure module functions) is clean.

**Recommendation: merge after review.** This is a well-engineered feature with strong safe-by-default defaults and an unusually honest accompanying security review. The read-only + loopback defaults, server-side enforcement, and Host/CSP headers mean the default config is safe. The main pre-broad-release action item is strengthening the `--expose` cleartext warning (W1 tier A — ~10 lines) since that's the highest-impact operator-awareness gap; the rest (full CSP, auto-TLS) are reasonable follow-ups. The 322-test suite and live dogfood verification give reasonable confidence. Given the size, a second pass on `app.js` (the largest single file, 1,173 lines) for any DOM-XSS sinks beyond the xterm.js `textContent` path would be prudent, though the review already flags xterm byte rendering as the XSS guard.

openwong2kim:
Pushed b508e7d2 addressing all review findings — the 12 CodeRabbit items, the 8 Codex P2s, and the attacker note on /api/pair.

Highlights beyond the mechanical fixes:

- **/api/pair cross-site burn** (attacker note): the route now refuses requests stamped `Sec-Fetch-Site: cross-site`, so a hostile page embedding `` can no longer touch the attempt budget. Same-origin fetches from the pairing page, direct navigation, and non-browser clients are unaffected (tested).
- **Tailscale Serve host** (Codex): rather than honoring forwarded headers (which would soften the rebinding guard), there's a new `--allow-host <name>` / `allowedHosts` start option to explicitly allowlist the MagicDNS name. The CLI hint under the tailscale-serve suggestion now mentions it.
- **Expired pairing codes**: status() never advertises a code the phone would be refused for — it drops an expired one and mints a replacement once the 30s regen cooldown allows.
- **Input ordering**: browser input POSTs are now chained per session, so fast typing / paste chunks can't reach the PTY out of order.
- Failed binds no longer leak session-manager listeners (registration moved after a successful listen), and the SSE response now carries the same baseline security headers as everything else.

7 new tests cover these (SSE headers, cross-site refusal, expired-code replacement, allowedHosts, IPv6 URL bracketing, bind-failure listener cleanup, null IPC payload). tsc clean, changed files lint clean, `build:daemon-web` still builds.

PR #579, opened 2026-07-24
fix(hooks): keep ~/.wmux/hooks scripts current on app update, trim statusline label

openwong2kim:
## What / why

The statusline and hook-bridge scripts under `~/.wmux/hooks/` are copied outside the app dir on purpose, to survive Squirrel's `app-x.y.z` swap — which is exactly why **an app update never refreshed them**. New code shipped in the bundle while the stale copy kept its old behavior until the user re-ran `wmux setup-statusline` / `wmux setup-hooks` by hand.

Two regressions this caused:

- **The statusline's `1M` note disappeared.** wmux derived it from a `" (1M context)"` suffix Claude Code used to bake into `model.display_name`; ≥2.1.218 sends the clean name and reports the window under `context_window.context_window_size`, so the suffix regex was permanently false. Rather than resurrect the note, the window is **dropped from the label on purpose** — it rarely differs pane to pane, and the live `ctx N%` beside it is the part that actually moves. The label now reads `Opus 4.8 (xhigh)` (model + effort). The legacy suffix is still stripped so older Claude Code renders identically.
- **The bridge missed the "30+ concurrent sessions" scaling fix (#538):** log rotation + a per-session activity-stamp throttle that tames hook storms. On an already-installed bridge the log grew without bound and the throttle wasn't there at all (observed: `~/.wmux/bridge.log.1` at 11 MB).

## The fix — the update path itself

`refreshStatuslineScript()` / `refreshHookBridge()` reconcile each stable copy against the bundled version shortly after boot (packaged app only, 3 s deferred, best-effort). Reconcile means **repair, not just refresh**: an entry that still references a script whose file was deleted (manual cleanup, partial reinstall) is rewritten rather than left pointing at nothing.

**Neither can enroll anyone** — each is a no-op unless a wmux-owned entry (a `statusLine`, or hook groups referencing the bridge) is already installed, and neither ever writes `settings.json`. The "never auto-run at boot" constraint (opting a user *in*, 2026-07-17) is untouched. The bridge refresh owns only the plugin-LESS copy — the marketplace plugin's own versioned bridge under `~/.claude/plugins/cache/…` is left to Claude Code's plugin system. Both writes are pid-scoped tmp+rename, so a tick/hook landing mid-update never reads a half-written script and two instances racing at boot can't collide.

## Tests

The label rendering had no tests at all — which is how this regressed unnoticed. Added:
- **statusline render (8)** — feeds the script the exact JSON Claude Code pipes on stdin. Locks that the window size never leaks into the label from *either* source (1M / 200k / legacy suffix / 2M window / no effort / malformed context_window).
- **statusline refresh (7) + bridge refresh (11)** — stale→refreshed, identical→up-to-date, never-installed & foreign-only→not-installed, missing-but-referenced→restored, no source→no-source, unreadable→failed, `settings.json` never written.

`tsc --noEmit` clean · related tests 69 passing · no new lint.

## Review

3-way independent review (Claude / Codex / GLM-5.2), no P1. Codex+GLM converged on a P2 — the refresh returned `not-installed` for a referenced-but-deleted file, so a broken statusLine/hook was never repaired; fixed in `bce76ff4` (check ownership before reading the file). P3s (pid-scoped temp, log levels, test cleanup/timeout) also applied.


openwong2kim (commented):
**[wmux-hermes] Code review — P3 (regression fix + auto-refresh improvement; merge after review)**

This PR does two things: (1) trims the statusline model label to `Opus 4.8 (xhigh)` — model + effort only, dropping the context-window size that never actually surfaced (Claude Code ≥2.1.218 moved it to `context_window`, and the old `" (1M context)"` suffix regex is now just stripped for parity); and (2) adds a boot-time reconcile that keeps `~/.wmux/hooks/` copies of `wmux-statusline.mjs` and `wmux-bridge.mjs` in sync with the bundled versions, fixing a real class of silent staleness where app updates (Squirrel's `app-x.y.z` swap) never refreshed the external copies until the user re-ran `wmux setup-statusline`/`setup-hooks` by hand. The bridge staleness had a concrete operational impact (the "30+ sessions" scaling fix — log rotation + per-session activity-stamp throttle — never reached already-installed bridges, so `bridge.log` grew without bound).

The implementation is careful and well-scoped. The refresh functions (`refreshStatuslineScript`, `refreshHookBridge`) correctly check ownership *before* reading the file (so a settings entry pointing at a deleted script is *repaired*, not written off as uninstalled), never write `settings.json` (preserving the "never auto-enroll at boot" constraint), use pid-scoped tmp+rename for atomicity (so a statusline tick or hook firing mid-update can't read a half-written script, and two boot-racing instances can't collide), and degrade to typed outcomes (`refreshed`/`up-to-date`/`not-installed`/`no-source`/`failed`) rather than throwing. The bridge refresh deliberately owns only the plugin-less copy and leaves the marketplace plugin's versioned bridge untouched — a clean separation. Test coverage is strong: 8 statusline render tests (feeding the exact stdin JSON Claude Code pipes), 7 statusline refresh cases, and 9 bridge refresh cases covering stale→refreshed, identical→up-to-date, not-installed, foreign-only→skip, no-source, unreadable→failed, and corrupted-settings→skip.

A couple of minor observations (none blocking): the 3-second boot delay is `unref`'d and best-effort, which is appropriate, but it's worth confirming that a machine under heavy startup load won't consistently lose the refresh window (the `failed`/`no-source` warn-level logging will surface this if it happens in the wild). The statusline label change is a deliberate UX decision (window size rarely differs pane-to-pane, `ctx N%` is the live signal) — reasonable, and the legacy suffix stripping ensures older Claude Code renders identically. No new lint issues, `tsc --noEmit` clean, 65 related tests passing.

Recommendation: **approve / merge after review** — this is a focused, well-tested fix for a real staleness problem plus a cosmetic label cleanup.

PR #584, opened 2026-07-24
feat(deck): pin an agent+model to a pane role and enforce it on wmux-assembled launches

openwong2kim:
## What

Bind a pane role (Builder / Reviewer / Tester / Planner) to an agent + model in Settings → Orchestrator, and the agent launches **wmux itself assembles** for a pane carrying that role are transparently rewritten to use the bound model. `terminal_send("claude")` from the orchestrator becomes `claude --model haiku`. This is the token-optimization move power users hand-code on competitor platforms (the "worker issuance" thread in `docs/orca-power-user-gap-analysis-2026-07-24.md`); here it is a Settings row.

Coverage is exactly three wmux-owned assembly points: **orchestrator-issued launches** (`input.send`), **project-seeded commands** (a `wmux.json` leaf that declares a role), and the **resume command** (so a bound model survives a reboot). **A launcher you type into the terminal yourself is not rewritten** — your keystrokes go straight to the shell — and the Settings copy says exactly that.

## How it works

- Pure transform `applyRoleBinding()` in `src/shared/orchestratorRole.ts`. Two load-bearing gates: the launcher stem must be a known agent CLI, **and** the rest of the line must look like arguments rather than prose — so a shell command or a natural-language instruction to a running agent comes through untouched even when its first word is `claude`.
- Backbone enforcement at the `input.send` chokepoint (`input.rpc.ts`), which every orchestrator-issued submit flows through. Fail-open on a role-lookup miss (this is a cost/policy control, not a security boundary).
- A model is injected only when the binding also names its agent (`--model` grammar is per-agent); the Fleet chip and pane badge appear only when the binding is really being applied, never for one that is configured but inert.

## Security: the enforcement role is operator-only

The role selects which binding applies, so it must not be agent-writable. `custom['orchestrator.role']` is now **operator-only on the pipe RPC**: a non-first-party `pane.setMetadata` (or `pane.clearMetadata`) can no longer set or clear it — closing a path where a prompt-injected agent could retarget the operator's policy onto a pane of its choosing, or unbind its own role to escape a cheap-model pin. The operator's Fleet dropdown is unaffected (separate `metadata:set-role` IPC). `ctx.firstParty` cannot be forged from the external wire.

## Review

Two rounds of independent review (Claude + Codex + GLM + UX + security). Round 1 found `args` appended with no launcher gate (every submitted line mutated), a model forced onto the wrong agent, false "enforced" copy, and the agent-writable role key; round 2 found the args field couldn't be typed (only pasted), badges claiming enforcement that wasn't happening, and project-seeded panes silently losing their role. All fixed. GLM specifically hunted the launch heuristic and the role guard for bypasses and found none.

## Verification

- Unit + handler + component tests across the touched families: all passing, three tsconfigs typecheck clean. Security tests pin the operator-only role guard on both `setMetadata` and `clearMetadata`, under every merge mode.
- **Live dogfood** (packaged build, isolated data dir), the decisive comparison: two panes both bound Reviewer → `claude`/`haiku`. The orchestrator path (`input.send`) launched **Haiku 4.5** and returned `enforcedModel: "haiku"`; typing `claude` by hand into the *same-bound* pane launched **Opus 4.8 (default) — not rewritten**. Same binding, different input path — exactly what the Settings copy promises.

No version bump; CHANGELOG under `[Unreleased]`. No new MCP tool/capability, so no `gen-api-reference` change.

## Note for the owner

A hand-typed launcher in a bound pane still shows the `claude · haiku` chip in the Fleet roster (the chip is the pane's *policy*, not its running model). The copy explains this, but if you'd rather the chip read as "configured" vs "enforced" there, say the word.

## Summary by CodeRabbit

* **New Features**
  * Added role-to-agent/model bindings for orchestrated panes, with automatic enforcement during launches, project rebuilds, and resumes.
  * Added Orchestrator settings to configure each role’s agent, model, and optional arguments.
  * Added visible indicators showing enforced launches in pane and fleet views.
  * Added support for declaring valid pane roles in `wmux.json` layouts.
* **Security**
  * Pane roles are now operator-controlled and cannot be assigned or removed by agents.
* **Bug Fixes**
  * Preserved role bindings across session restoration and restarts.
  * Avoided rewriting commands with explicit model options or unsupported command formats.


openwong2kim:
Fixed in eb82cf8b: the reboot-recovery pill now re-asserts the bound model too, not just the persistent chip. Extracted a pure `planRecoveryPillType` helper that routes each launcher-prefixed variant through `applyRoleBinding`; for the two-stage path the model rides the stage-0 base (`claude --model haiku --permission-mode plan`) so the appended `--resume <id>` fragment lands on an already-bound line. 13 new tests cover it. The stage-2 continuation fragment is deliberately not rewritten (no launcher stem). Thanks for catching this — it's the exact reboot case the CHANGELOG promises.

Issue #580, opened 2026-07-24
browser.cdp.info exposes every workspace's live browser targets to any browser.read caller (sibling of #554); composes with global browser.close


snowyukitty:
## Summary

`browser.cdp.info` returns **every** registered CDP target — across all workspaces — with its `surfaceId`, `targetId`, and owning `workspaceId`, unfiltered by the caller. Its capability is `browser.read`, an ordinary plugin-declarable capability (not `wmux.internal`). So a plugin the user has approved for browser reads can enumerate which workspaces currently hold live browser guests and learn their surface IDs — the same cross-workspace disclosure that #554 set out to close for the browser *tools*, still open at the RPC beneath them.

It composes with a second, deliberate behavior: `browser.close` resolves an explicit `surfaceId` across **every** workspace (the "surface.close lesson" — scoping an unambiguous ID only manufactures false not-founds). A caller holding `browser.read` + `browser.navigate` can therefore read a foreign `surfaceId` out of `browser.cdp.info` and hand it straight to `browser.close`, tearing down another workspace's browser.

This is a **sibling of #554 / follow-up to #565**, not a regression in either. Neither leg is new: #563 (for #554) scoped the read *tools* (`browser_snapshot`, `browser_evaluate`, …) through `PlaywrightEngine.getPage()`, and #575 (for #565) scoped `browser_tabs`, but both sit *above* `browser.cdp.info`, which still answers globally. Filing so the gap is tracked outside the design doc that currently records it.

Surfaced by two independent reviews of PR #575 (Grok found the global `browser.close`; Antigravity found the global `browser.cdp.info`); it took both halves to see the composition. It is written up in that PR's merged contract at `docs/internal/browser-tabs-workspace-contract.md` §8.2, but a doc note is not a tracked item.

## Severity / threat model

**P2/P3, within the same-OS-user trust ceiling — not remotely exploitable.** The attacker is a plugin the user has already approved for `browser.read` **and** `browser.navigate`. It cannot read page content, URLs, or titles through this path (see "What is *not* exposed"). But it is a genuine workspace-isolation gap: #554 is a P1 precisely because "one workspace's agent observing/acting on another's browser" is a boundary this project enforces, and this is the same boundary, one layer down.

### Leg 1 — global disclosure

`src/main/pipe/handlers/browser.rpc.ts:420` — `browser.cdp.info` takes no params and filters by nothing:

```ts
router.register('browser.cdp.info', async () => {
[...]

Issue #565, opened 2026-07-23
browser_tabs is still workspace-blind (follow-up to #554)


openwong2kim:
Follow-up from #554 (PR #563), which scoped the browser **read** tools (`browser_snapshot`, `browser_evaluate`, `browser_extract_*`, `browser_wait`) to the calling session's workspace. `browser_tabs` was deliberately left out of that fix because it's a different code path with fuzzier semantics.

## What's still workspace-blind

`browser_tabs` (src/mcp/playwright/tools/navigation.ts) doesn't go through `PlaywrightEngine.getPage()`. Its `list` action calls `engine.getBrowser()` and enumerates **every** page across all CDP contexts:

```ts
const contexts = browser.contexts();
const allPages = contexts.flatMap((ctx) => ctx.pages());
```

So from any workspace it returns the whole CDP browser — the app shell (`127.0.0.1:5173`), DevTools targets, and every other workspace's guest page — not just the caller's own surface. `select` / `close` then act on a `tabId` index into that global list, so an agent in workspace A can enumerate (and potentially close/select) workspace B's page.

Reported as a secondary symptom in #554 ("`browser_tabs` lists the app shell and DevTools targets instead of the surface's page").

## Why it wasn't folded into #563

The read-path fix has a clean per-call answer: "the caller's one surface." `browser_tabs` has no obvious scoped meaning yet:
- Is a "tab" a surface? A workspace has at most one browser surface today, so `list` would usually return a single entry — is that the intended model?
- Should `list` be scoped to the caller's workspace surface only, or is a cross-workspace view sometimes legitimate (e.g. an operator/commander role)?
- `select`/`close` by global index is unsafe regardless — at minimum they should be constrained to the caller's own surface(s).

## Proposed direction

- Filter `browser_tabs list` to the caller's workspace using the same `workspaceId`-tagged `browser.cdp.info` targets #563 added (surfaces now carry their owning workspace), and drop the app-shell / DevTools targets.
- Make `select`/`close` fail-closed to the caller's own surface(s).
- Decide the tab model explicitly (per-surface vs per-workspace) before wiring it.

Lower priority than #554 — the read tools were the P1 exposure; `browser_tabs` is used far less by agents.

openwong2kim:
[wmux-hermes]

**Priority: P2 — Important**

[...]

Issue #554, opened 2026-07-22
browser_snapshot / browser_evaluate can read another workspace's browser page when multiple browser surfaces are open


openwong2kim:
## What happens

With two workspaces each holding a live browser surface, an agent's read-side browser tools can return the OTHER workspace's page.

Repro (dev build off current main, v3.31.0):

1. Workspace A: agent calls `browser_open` with `https://www.google.com/`
2. Workspace B: agent calls `browser_open` with `https://example.org/`
3. Agent in A calls `browser_snapshot` → it reports **https://example.org/** (B's page)

Ground truth via CDP `/json/list` at the same moment shows the state itself is correct — A's webview is on google.com, B's on example.org. Only the read is misrouted.

## Scope

- **Write path is fine.** `browser_open` / `browser.navigate` route by the caller's resolved workspace; each surface loads its own URL.
- **Read path is workspace-blind.** `PlaywrightEngine.getPage()` picks the first non-shell page from the CDP target list without filtering by the caller's workspace/surface. `browser.cdp.info` already returns targets tagged with `surfaceId`, so the data needed to scope the selection is available — it just isn't used at pick time.
- Likely the same selection layer: `browser_tabs` lists the app shell (`127.0.0.1:5173`) and DevTools targets instead of the surface's page, and `browser_wait` with a URL pattern timed out even though the page was already loaded.

## Impact

Any fleet running 2+ browser surfaces: `browser_snapshot` / `browser_extract_*` / `browser_evaluate` from one pane can read a different pane's page — cross-workspace information exposure (read-only, but still the wrong pane's content), plus confusing agent behavior (an agent 'verifies' a navigation against somebody else's page).

Verified against both a pane-spawned MCP child and an externally configured one — it is the page-selection heuristic, not the spawn topology.

openwong2kim:
[wmux-hermes]

**Priority: P1 — Critical**

In a multi-workspace setup, an agent's browser read tools (`browser_snapshot`, `browser_evaluate`) can return another workspace's page. This is a workspace isolation breach — the agent reads and acts on unintended context, raising data integrity and security concerns.

CDP `/json/list` confirms both surfaces exist simultaneously. The agent appears to be selecting a target that doesn't belong to its own workspace, indicating missing ownership validation in the browser surface ↔ agent binding or target selection logic.

[...]

PR #586, opened 2026-07-24
fix(browser): scope browser.close by explicit surfaceId to the caller's workspace (#580 action leg)

snowyukitty:
## Summary

Closes the **destructive half** of #580. `browser.close` resolved an explicit `surfaceId` across **every** workspace, so a caller in workspace A could close workspace B's browser once it learned B's surface id — which it can still read out of the unscoped `browser.cdp.info` (the disclosure half of #580). This is Option 3 from the issue, which you flagged as the welcome self-contained stopgap.

An explicit `surfaceId` is now scoped to the caller's own workspace and fails closed when caller identity is absent. The `browser.cdp.info` scoping (Options 1/2) remains tracked in #580.

## What changed

- **`browser.close` explicit-`surfaceId` path is workspace-scoped.** It routes through the same `closeBrowserTabInWorkspace` helper `browser_tabs` already uses, so a foreign or missing id fails identically and never mutates, and the #143 last-surface pane cascade is preserved.
- **Fails closed without caller identity.** An explicit `surfaceId` with no `workspaceId` is rejected rather than falling back to the UI-active workspace. Per your review and contract §5 ("unresolved identity MUST NOT fall back to the UI-active workspace"), UI activity is not caller identity — otherwise a plugin with no workspace could still destroy whichever surface the user happens to be looking at.
- **The surface-id-less "close the browser pane" convenience is unchanged** (CLI `wmux browser close`): it still resolves the first browser surface in the routed workspace, explicit `workspaceId` else UI-active.
- Ownership policy extracted to a pure `decideBrowserClose(params, activeWorkspaceId)` in `browserTabs.ts`; the handler just executes its decision.

## Why this does not reintroduce the "surface.close lesson"

The old comment warned that scoping an explicit id "manufactures false not-founds." That lesson was specifically about scoping to the **UI-active** workspace, which broke background closes. This scopes to the **caller's** workspace, which is a different thing:

- **MCP `browser_close`** always sends a verified `workspaceId` via `requireWorkspaceId` — a background workspace's own agent closing its own surface still works (covered by a runtime regression test).
- **CLI `browser close`** never sends a `surfaceId` at all (`src/cli/commands/browser.ts`), so it never reaches this branch.
- I enumerated every `browser.close` caller to confirm no legitimate one depends on the old cross-workspace shape.

## Test plan

**Runtime tests** (real store fixture, `browserTabs.test.ts`), the piece you required:

- Workspace A **cannot** close workspace B's surface — `closeBrowserTabInWorkspace(state, 'ws-a', 'browser-b')` returns null with **no** `closeSurface`/`closePane` call.
- Workspace A **can** close its own surface while B is the active workspace.
- A **background** workspace can still close its own surface — the guard proving no active-workspace regression (the fixture's `activeWorkspaceId` is `ws-b`, so `ws-a` is background throughout).

**Unit tests** for the pure `decideBrowserClose`: explicit+workspace → scoped; explicit+no-workspace → **reject** (and the decision provably does not contain the active workspace id); surface-id-less → routed/active; empty-string ids treated as absent.

**Source-structural** (`browserClose.test.ts`): the test that pinned the old global search is replaced with one asserting the handler delegates to `decideBrowserClose` + the scoped helper and **no longer** does `for (const ws of store.workspaces)`. Verified this negative assertion is a real lock — it fails against the old code and passes against the new.

**Checks:** focused suites (26 + 67 passing), `tsc --noEmit` clean, eslint on touched files (no new findings), `git diff --check` clean, full `npm test` green (the intermittent `diff.handler` timeout is the unrelated parallelism flake, never these files).

## Stability tier impact

- [x] Change to an `experimental` surface (note in release notes) — the browser tools. Also an `internal` behavior change to the `browser.close` RPC's explicit-id resolution.

Behavior change, scoped to a security tightening: an explicit foreign `surfaceId` that previously closed a foreign browser now fails. No legitimate caller relied on that (see caller enumeration above).

## Substrate contract impact

`docs/internal/browser-tabs-workspace-contract.md` §8.2 updated: the destructive leg is marked closed with the reasoning above; the `browser.cdp.info` disclosure leg is explicitly still open and pointed at #580.

## Related

Closes the destructive half of #580 (Option 3). Sibling of #565 / #554. `browser.cdp.info` scoping (Options 1/2) awaits your input on what `PlaywrightEngine`'s attach path requires from the global target list.

## Reviewer checklist

- [x] CHANGELOG entry filled in (under `### Fixed`, next to the #565 sibling).
- [x] Stability tier impact classified.
- [x] Substrate contract doc updated (§8.2).
- [x] Tests cover the new behavior and the regression case (runtime + unit + wiring).
- [x] No secrets, tokens, or unrelated changes — 6 files, all intended.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser closing security by restricting explicit surface closures to the caller’s workspace.
  * Requests without a valid workspace identity now fail safely instead of potentially affecting another workspace.
  * Preserved the existing behavior for closing the active browser pane when no surface is specified.

* **Documentation**
  * Clarified browser target visibility and workspace-scoped closing behavior in the internal browser contract documentation.


openwong2kim:
[wmux-hermes]

**Priority: P2 — Important**

**What the PR changes.** This closes the destructive ("action") half of #580: `browser.close` previously resolved an explicit `surfaceId` by iterating over *every* workspace (`for (const ws of store.workspaces)`), so a caller in workspace A could tear down workspace B's browser once it learned B's surface id (which the still-unscoped `browser.cdp.info` readily discloses). The fix extracts the ownership policy into a pure, unit-testable `decideBrowserClose(params, activeWorkspaceId)` in `browserTabs.ts` (three outcomes: `bySurface` / `reject` / `byWorkspace`), and the handler now routes an explicit `surfaceId` through the same workspace-scoped `closeBrowserTabInWorkspace` helper that `browser_tabs` already uses. Crucially, an explicit `surfaceId` *without* a caller `workspaceId` now **fails closed** rather than falling back to the UI-active workspace — per contract §5, UI activity is not caller identity. The surface-id-less "close the browser pane" convenience (CLI `wmux browser close`) is intentionally unchanged.

**Risks / edge cases worth noting.** The author carefully distinguishes this scope-to-*caller* from the old "surface.close lesson" about scope-to-*active*, and that distinction is the crux: MCP `browser_close` always supplies a verified `workspaceId` via `requireWorkspaceId`, and the CLI never sends a `surfaceId`, so no legitimate caller depended on the cross-workspace shape. The risk to watch is a future caller that issues `browser.close` with an explicit `surfaceId` but relies on the active-workspace fallback — that will now get a `reject`, which is the correct (safer) behavior but is a behavior change worth flagging in release notes (the CHANGELOG entry does this). The contract doc §8.2 is updated to mark the destructive leg closed and the `browser.cdp.info` disclosure leg explicitly still open — honest and well-scoped.

**Implementation feedback.** Extracting `decideBrowserClose` as a pure function is exactly the right move: the ownership logic is now testable without standing up the renderer, and the handler becomes thin glue. The test plan is genuinely strong — runtime tests on a real store fixture proving (a) A cannot close B's surface (no `closeSurface`/`closePane` call), (b) A closes its own surface while B is active, and (c) a *background* workspace closes its own surface (the active-workspace-regression guard, since the fixture's `activeWorkspaceId` is `ws-b`). The source-structural test replacing the old global-search pin with a negative assertion (`not.toMatch(/for \(const ws of store.workspaces\)/)`) is a clever and valid lock — it fails against the old code. Empty-string-id handling is covered. One tiny note: `decideBrowserClose` uses `params.surfaceId.length > 0` which correctly treats `''` as absent — good.

**Recommendation: approve.** This is a clean, well-reasoned security tightening with no legitimate-caller regression, thorough tests (runtime + unit + structural lock), and an honest contract-doc update. The disclosure leg (`browser.cdp.info`) remains correctly tracked in #580 since it needs its own design around `PlaywrightEngine`'s attach path.

PR #587, opened 2026-07-24
fix(mcp): register the wmux server from a stable path so it survives app updates

openwong2kim:
## What's wrong

wmux registers itself as an MCP server in each agent's config (`~/.claude.json`, Codex's `config.toml`) by writing the path to its bundled server script. That path pointed straight into the versioned install directory — on Windows, `%LocalAppData%\wmux\app-{version}\resources\mcp-bundle\index.js`.

So the moment Squirrel cleaned up the old `app-*` after an update — or, while dogfooding, a build's `out/` or a git worktree got pruned — the registered path pointed at a file that no longer existed, and any agent reading that config could no longer launch the wmux MCP server. Every `wmux` tool silently went missing until the app booted again and re-registered.

Found live while investigating: the registered `args` on my machine pointed at a `D:\wmux-worktrees\…\out\…\mcp-bundle\index.js` that had already been deleted.

## Fix

Copy the MCP bundle to a stable, version-free location under `~/.wmux/mcp/` and register THAT path — the same cross-update durability the Codex `notify` bridge and the hook/statusline scripts already have.

- The whole bundle **directory** travels together — `index.js` resolves its siblings (`playwright-chunk.js`, `playwright-core-package.json`) via `path.join(__dirname, …)`, so a single-file copy wouldn't work.
- The copy is **version-marker gated**, so a warm boot pays nothing (the bundle is ~6MB).
- It **repairs** itself when the stable file is later deleted.
- It **fails open** to the old versioned path if the copy can't be written (locked file mid-upgrade, disk full), so registration is never worse than before.
- Dev mode is unchanged — it already points at `dist/`, which is stable.

Your own MCP servers in those configs are left untouched (foreign-key preservation was already in place).

Scope: only `McpRegistrar` (the every-boot registration in the main process) is fixed here — that's the path that produced the observed breakage. The `wmux mcp register` CLI computes its own bundle path too and could get the same treatment as a follow-up; it only runs on a manual invocation, so it's lower priority.

## Verification

- New unit tests for the copy/gate logic (cold, warm, version change, repair, fail-open, sibling copy, broker shim) plus the existing McpRegistrar suite: 17 passing. `tsc --noEmit` and eslint clean.
- Live run: drove the real `McpRegistrar` in a packaged context against an isolated HOME with the real 6.4MB bundle. Confirmed the registered path is the stable one (not the versioned resources path), the whole bundle copies, warm boots skip the copy, a version bump re-syncs, and — the point of the fix — the registered path still resolves after the versioned source directory is deleted.


openwong2kim:
[wmux-hermes]

**Priority: P2 — Important**

**What the PR changes.** This fixes a genuine cross-update durability bug: `McpRegistrar` registered the MCP server command pointing straight into the versioned `resources/mcp-bundle/index.js`, so once Squirrel cleaned up the old `app-*` (or a dogfood worktree/`out/` was pruned), the registered path in `~/.claude.json` / Codex `config.toml` dangled and every wmux MCP tool silently disappeared until the next boot re-registered. The fix introduces `stabilizeMcpBundle` (`src/main/mcp/stabilizeBundle.ts`, pure fs+path), which copies the *whole bundle directory* into `~/.wmux/mcp/` (siblings like `playwright-chunk.js` must travel with `index.js` since the entry resolves them via `__dirname`), gated on a version-marker file so a warm boot pays nothing, with repair-on-missing-entry and fail-open to the versioned source on any copy failure. `McpRegistrar.getMcpScriptPath()` is the only wiring change, applied to both the full-bundle and broker-shim branches.

**Risks / edge cases worth noting.** The tmp+rename-per-file strategy (pid-suffixed `.tmp` then atomic `renameSync`) correctly prevents a half-written script from being spawned mid-copy and avoids two-boot races — good. The marker is written *last*, so a crash mid-copy leaves `marker !== version` and the next boot re-syncs rather than trusting a partial directory — also correct. One edge worth keeping in mind: `fs.readdirSync(sourceDir, { withFileTypes: true })` copies every regular file in the bundle dir, so if the source layout ever grows a stray non-runtime file (a `.map`, a test artifact) it travels too — harmless for correctness but slightly wasteful; not a blocker. The `entry-absent-after-copy` fallback is defensive and correct.

**Implementation feedback.** The split into a pure `stabilizeMcpBundle` function + a thin Electron-aware `stabilizePackaged` method on `McpRegistrar` is the right shape — the security-relevant copy/gate logic gets direct unit coverage without an Electron harness. The 7 new test cases (cold boot, survives-source-deletion, warm-boot no-op, version-change re-sync, repair-on-deletion, fail-open-on-missing-source, no-`.tmp`-leftovers, shim entry) are exactly the cases that matter. One minor gap: there's no test for the `EACCES`/`ENOSPC` copy-failure path (the `catch` arm that returns `sourceScript` with a reason) — the `entry-absent-after-copy` case covers the structural fallback but not an actual `copyFileSync` throw; a mock-able `fs` or a read-only `stableDir` would close it, though this is non-blocking since the behavior is "identical to pre-fix." CHANGELOG is excellent.

**Recommendation: approve.** A clean, well-scoped fix for a real "tools vanished after update" bug, mirroring the durability pattern already used for the Codex notify bridge and hook scripts. Fail-open guarantees it's never worse than before. The follow-up to apply the same treatment to the `wmux mcp register` CLI path (noted in the PR body) is correctly deferred.

PR #588, opened 2026-07-24
harden the MCP broker for default-on: readiness fallback, pipe reclaim, broker-aware CLI

openwong2kim:
Hardens the shared MCP broker (`WMUX_MCP_BROKER=1`, still default OFF) so it's safe to flip default-on later. Implements W1, W2, and W4 from `plans/mcp-broker-enable-plan-2026-07-24.md`. The broker itself already exists behind the flag; this closes the three lifecycle edges that made default-on strictly worse than the single-child bundle in failure cases.

## What changed

**W1 — Registration falls back to the full bundle when the broker can't serve.** With the broker enabled, each agent CLI is pointed at a thin shim that connects to one resident broker. If that broker never comes up (missing script, crash loop, permissions), the old path left every agent on a shim that gives up after ~3s and takes all wmux tools (browser, terminal, a2a, channels) down with it. Boot now awaits `BrokerSupervisor.whenReady()` (bounded, never rejects, per-probe socket timeout so a half-open Windows named pipe can't hang startup) and registers the self-contained bundle when the broker isn't reachable. `register()` is self-correcting: the re-register IPC and first-run onboarding call sites omit opts and probe the broker themselves, so clicking "re-register" while the broker is down writes the working bundle rather than re-installing a dead shim.

**W2 — A stood-down supervisor reclaims an orphaned pipe.** When a second instance finds the pipe already owned (dev hot-reload, two windows) it exits 75 and stands down. Previously that was permanent — if the owner later exited, nothing took over. The supervisor now probes periodically and restarts once the other broker is gone; `stop()` clears the timer.

**W4 — `wmux mcp register` keeps the broker topology.** The CLI used to always rewrite configs to the full bundle, so running the exact command the app's troubleshooting text recommends silently reverted every agent off the broker. It now probes the broker pipe directly (not the env flag — the CLI shell env isn't the app env) and keeps the shim wiring when a broker is listening, falling back to the bundle otherwise. `WMUX_MCP_BROKER=0` remains an escape hatch that skips the probe.

Flag-off behavior is byte-identical: no probe, no new logs, full bundle.

## Verification

- 37 tests across the 5 broker/CLI test files pass; `tsc --noEmit` (root + cli) clean.
- New dynamic harness spawns the real `broker.js` and validates readiness flip, the exit-75 already-running contract, and the connect-hang guard against real processes.
- Live end-to-end dogfood: real `shim.js` → real `broker.js` over an isolated pipe completes the MCP `initialize` + `tools/list` handshake with the correct `ws+pty` identity hints and exposes all 84 wmux tools.
- Full suite failures (21) are pre-existing Windows git-spawn timeouts / EBUSY, unrelated to this diff (pass in isolation, zero code overlap).

## Not in this PR

Does not flip the default (stays OFF), and does not run the W3 packaged-build matrix — both are separate gates before default-on. A minor follow-up: the connect-probe helper is duplicated between `src/main/mcp/brokerProbe.ts` and the CLI (intentional, to avoid coupling the CLI bundle to a main-process module).

## Summary by CodeRabbit

* **New Features**
  * MCP registration now detects broker availability and selects the appropriate lightweight or full MCP bundle.
  * MCP operations wait for broker readiness and registration completion before reporting status.
  * Broker recovery can reclaim abandoned connections and restart safely when ownership is released.

* **Bug Fixes**
  * MCP now falls back to the full bundle when the broker is unavailable, missing, or times out.
  * Added safeguards to prevent connection checks from hanging during startup or shutdown.

* **Tests**
  * Added coverage for broker detection, registration paths, readiness, recovery, and lifecycle behavior.


openwong2kim:
@codex review

openwong2kim:
[wmux-hermes]

**Priority: P2 — Important**

**What the PR changes.** This hardens the shared MCP broker (`WMUX_MCP_BROKER=1`, still default OFF) along three lifecycle edges so it's safe to flip default-on later — implementing W1, W2, W4 from the broker-enable plan. **W1:** boot now `await`s `BrokerSupervisor.whenReady(timeoutMs)` (bounded, never rejects, per-probe socket timeout) and registers the full self-contained bundle when the broker can't serve, so agents keep their tools instead of stranded on a dead shim; the self-correction extends to re-register IPC and first-run onboarding (both omit `opts` and probe the broker themselves). **W2:** a supervisor that stood down on exit-75 (another instance owns the pipe) now periodically probes and reclaims the orphaned pipe once the other broker dies, instead of standing down permanently. **W4:** `wmux mcp register` (CLI) now probes the broker pipe directly and keeps the shim topology when one is listening, rather than always rewriting to the full bundle. A new `canConnectBrokerPipe` helper (with an explicit socket timeout guarding against half-open Windows named pipes) is shared between the supervisor, registrar, and CLI.

**Risks / edge cases worth noting.** The `register()` signature change from sync `void` to async `Promise<void>` ripples through every call site (boot `app.on('ready')`, `FirstRunOrchestrator`, `mcp.handler` re-register IPC) — all are correctly `await`ed, and since `ready` is already async this doesn't introduce a new boot-blocking hazard beyond the bounded `whenReady(4000)`. The reclaim probe (`setInterval` 15 s) is properly cleared in both `reclaimTick` (before `start()`) and `stop()`, and the defensive clear in `start()` prevents an interval leak if both instances stood down — the unit test "start() while a reclaim timer is armed clears it" pins exactly this. One thing to watch: if two instances both stood down and both reclaim simultaneously, one wins and the other gets exit-75 again and re-arms — converging to one owner, which the PR body calls out correctly. The `WMUX_MCP_BROKER=0` escape hatch in the CLI correctly skips the probe entirely.

**Implementation feedback.** The testing strategy is strong and layered: unit tests for `BrokerSupervisor.whenReady` / exit-75 reclaim (fake timers, mocked `canConnectBrokerPipe`), unit tests for `canConnectBrokerPipe` itself (driven socket lifecycle: connect/timeout/error, post-settle event absorption), `McpRegistrar` broker-topology selection (opts override, self-correcting probe, flag-off byte-identical), and a **dynamic harness** (`brokerLifecycle.dynamic.test.ts`) that spawns the real `broker.js` to pin the three contracts that mocks can't: the readiness signal actually flips on a real listener, a second broker really exits 75 on `EADDRINUSE`, and the probe settles (never hangs) against accept-never-write and missing pipes. The isolation discipline in the dynamic harness is good — unique `WMUX_DATA_SUFFIX` per test (no collision with a live wmux), `WMUX_AUTH_TOKEN` env (never touches the user's real token file), and `afterEach` SIGKILL of every child. The duplicated `canConnectBrokerPipe` between `src/main/mcp/brokerProbe.ts` and the CLI is called out as intentional (avoiding coupling the CLI bundle to a main-process module) — acceptable. The `socket.on('error')` (not `once`) choice to absorb stray post-destroy errors is a thoughtful detail that prevents unhandled-exception crashes.

**Recommendation: approve.** This is a well-structured hardening pass that closes three real failure edges (dead-shim stranding, permanent stand-down, CLI topology revert) without changing default-off behavior. The bounded `whenReady` never blocks boot past its timeout, fail-open is consistent, and the test coverage — including a real-process dynamic harness — is exactly what a change like this needs. The note that it does *not* flip the default and defers W3 (packaged-build matrix) is the right sequencing.

openwong2kim:
Both Codex findings addressed in cde5478c: the dynamic broker harness now skips itself when dist/mcp-bundle/broker.js isn't built (clean-CI runs skip it deliberately; it's a local/dogfood lane), and tsconfig.cli.json now excludes every __tests__ dir under src/cli so the CLI build no longer picks up test sources. Verified both lanes: suite passes with the bundle present and skips with exit 0 without it; dist/cli contains no test output.

Issue #529, opened 2026-07-21
Page.captureScreenshot hangs for a hidden (display:none) webview guest


openwong2kim:
While testing hidden-workspace browser panes I hit this: `Page.captureScreenshot` never resolves for a `<webview>` guest whose pane is `display:none` (hidden workspace). It's not our RPC layer — the same command sent over a direct devtools websocket to the guest target also hangs (tried `fromSurface:false` and `captureBeyondViewport:true` too), while the identical call returns in ~80ms as soon as the workspace is made visible again.

Repro (dev build):

1. Open a browser pane, then switch to another workspace so the guest is hidden.
2. Connect to the guest's `webSocketDebuggerUrl` from `/json` and send `Page.captureScreenshot {format:"png"}`.
3. The command never returns; switch the workspace back and it completes immediately.

Likely the compositor stops producing frames for the hidden guest, and capture waits on a frame forever. Impact: `browser_screenshot` / `browser.screenshot` silently time out for guests in hidden workspaces — exactly the background-automation case #353 cares about. `Runtime.evaluate` and input dispatch are unaffected.

Worth deciding whether to (a) force a frame before capture (e.g. briefly re-enable compositing or use an offscreen strategy), or (b) fail fast with a clear error instead of a hang.

openwong2kim:
Investigated this properly today. Short version: the hang is mitigated in #534, but real pixels for a non-compositing guest are not obtainable from our side — leaving this open to track that limit.

Findings, all measured over a raw devtools socket to the guest target (so none of this is our RPC layer):

- The hang is the capture waiting on a compositor frame the guest never produces. A healthy capture returns in under 100 ms.
- **Which visibility state hangs flips between GPU sessions on the same machine.** I reproduced both "hidden hangs, visible is fast" (as originally filed) and the exact inverse ("visible hangs, hidden returns instantly") on different boots. So don't trust any fix that branches on visibility — I almost shipped one.
- Levers that do NOT unblock a pending or fresh capture, all tried live: `fromSurface:false`, `captureBeyondViewport:true`, `Page.bringToFront`, `Page.setWebLifecycleState("active")`, and CSS reveal (overriding every `display:none` ancestor of the webview to an off-screen fixed position — a capture started before the reveal stays pending, and a fresh one during the reveal also hangs).
- `webContents.capturePage()` was the promising fallback and it hangs on exactly the same guests — same surface-copy machinery underneath.

[...]

PR #591, opened 2026-07-24
fix(browser): scope browser.cdp.info target list to the caller's workspace (#580 disclosure leg)

snowyukitty:
## Summary

Closes the **disclosure leg** of #580 — Option 1, implemented to your spec on the issue. `browser.cdp.info` no longer volunteers other workspaces' live browser targets to a caller that identifies itself.

Together with #586 (the destructive leg), this addresses both halves of #580.

## What changed

- **`browser.cdp.info` filters `targets` to the caller's workspace server-side.** It accepts an optional `workspaceId`; when present it returns only that workspace's targets and sets `targetsScoped: true`. `cdpPort` and `shellUrl` stay unscoped — they're workspace-agnostic and load-bearing for shell recognition. A param-less call keeps the legacy unscoped shape, so an older engine still works.
- **`PlaywrightEngine` passes its resolved workspace on every target-reading call** — `resolveCallerSurface`, `findViaTargetDomain`, `findViaJsonEndpoint` — threaded through `resolveSelectionContext` → `_getPageImpl`. `ensureConnected` only needs `cdpPort`/`shellUrl`, so it's left unscoped.
- **The `targetsScoped` flag disambiguates an empty list**, exactly as you flagged: with it, an empty response means "caller owns no live target" (`kind: 'none'`); without it (legacy main) the engine keeps its `anyTagged` leniency. This also means the no-`surfaceId` `targets[0]` fallback in the page finders can no longer resolve to another workspace's guest.

## Defense-in-depth, not a hard seal (as you noted)

Since `cdpPort` is still returned, anything holding it can enumerate targets by hitting the CDP `/json` endpoint directly — the engine itself does that during discovery. Sealing that would mean brokering CDP attach instead of sharing the port, which is a larger change and out of scope. This PR removes the casual/accidental cross-workspace read through the well-behaved RPC path, within the same-OS-user trust ceiling. The contract doc §8.2 and the CHANGELOG both say this plainly.

## Backward compatibility

The `workspaceId` param is additive and optional. A caller that doesn't pass it (or an older main that ignores it) gets identical behavior to before. The pre-existing #554 engine tests, which simulate an old main returning the full unfiltered list, still pass unchanged — they exercise the legacy path, and the new tests exercise the scoped path.

## Test plan

**Handler** (`browser.rpc.test.ts`):
- A `workspaceId` filters `targets` to that workspace and sets `targetsScoped: true`; neither the foreign nor an untagged target leaks.
- An owner-of-nothing gets `{ targetsScoped: true, targets: [] }` — the signal the engine reads as "own none".
- A param-less call returns all targets and no flag (legacy).

**Engine** (`PlaywrightEngine.test.ts`, new `#580` block):
- `resolveCallerSurface` passes the resolved `workspaceId` to `browser.cdp.info`.
- It reports its own surface from a scoped response.
- It reads an **empty scoped** response as `'none'`, not `'unscoped'` — the `targetsScoped` disambiguation, so the caller never falls through to another workspace's page.
- **The attach path under scoping:** on an empty scoped response, `getPage` auto-opens the caller's **own** browser — `browser.open` and `browser.cdp.info` are both called scoped to the caller (`ws-B`), never a foreign page. (The auto-open→returned-page mechanics are covered independently by the existing #554 auto-open test; this asserts the isolation-critical calls on the scoped path.)

**Checks:** `tsc --noEmit` clean; `git diff --check` clean; eslint on touched files adds **no** new findings (the 3 pre-existing errors — `new Promise(() => {})` in #529 tests, an empty constructor, an unused import — are on the base too, verified by stashing); full `npm test` green apart from two unrelated parallelism flakes (`diff.handler`, `McpRegistrar`) that both pass in isolation and touch nothing here.

## Stability tier impact

- [x] Change to an `internal` surface (no public contract impact) — `browser.cdp.info` is an internal RPC. Additive optional param; response gains an optional `targetsScoped` field.

## Substrate contract impact

`docs/internal/browser-tabs-workspace-contract.md` §8.2 updated: the disclosure leg is marked scoped-as-defense-in-depth, with the `/json`-via-port caveat spelled out; §5's cross-reference updated to match.

## Related

Closes the disclosure leg of #580 (Option 1). Sibling of #586 (destructive leg, merged) and #575 / #554. Not attempted here: brokering CDP attach to seal the port-based `/json` path — a separate, larger design.

## Reviewer checklist

- [x] CHANGELOG entry filled in (`### Fixed`, next to the #586 sibling).
- [x] Stability tier impact classified.
- [x] Substrate contract doc updated (§8.2, §5 cross-ref).
- [x] Tests cover the new behavior (scoped path) and the compatibility case (legacy unscoped path still covered by existing #554 tests).
- [x] No secrets or unrelated changes — 6 files, all intended.

## Summary by CodeRabbit

* **Security**
  * Browser target discovery is now restricted to the caller’s workspace, preventing foreign workspace surfaces from being revealed through browser operations.
  * Browser page resolution and automatic opening remain within the caller’s workspace.

* **Documentation**
  * Clarified workspace disclosure boundaries and the limitations of CDP target protection.

* **Tests**
  * Added coverage for workspace-scoped target discovery, empty results, legacy behavior, and workspace-safe page opening.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **`browser_close` can no longer close another workspace's browser by surface id.** Closing by an explicit `surfaceId` searched every workspace, so a caller in workspace A could tear down workspace B's browser if it learned B's id (composing with the cross-workspace target ids `browser.cdp.info` still exposes — [#580](https://github.com/openwong2kim/wmux/issues/580)). An explicit `surfaceId` is now scoped to the caller's own workspace and fails closed when caller identity is absent, rather than falling back to whichever workspace is on screen — matching the boundary `browser_tabs` already enforces. The surface-id-less "close the browser pane" convenience is unchanged. This closes the destructive half of #580; scoping the `browser.cdp.info` disclosure itself is tracked separately. (#580)
 
+- **`browser.cdp.info` no longer volunteers other workspaces' live browser targets.** The internal RPC that the browser automation engine uses to locate a guest returned every registered CDP target — with its surface id and owning workspace — to any `browser.read` caller, which is how a foreign surface id could be discovered in the first place. It now filters that list to the caller's own workspace server-side (the port and app-shell URL, which are workspace-agnostic, are unchanged). This is defense-in-depth within the same single-OS-user trust boundary, not a hard seal: anything holding the shared CDP port can still enumerate targets directly, so sealing that off is left as a larger change. Closes the disclosure half of #580. (#580)
+
 ## [3.32.0] — 2026-07-24
 
 ### Added
--- a/docs/internal/browser-tabs-workspace-contract.md
+++ b/docs/internal/browser-tabs-workspace-contract.md
@@ -317,7 +317,7 @@ Security rules:
 - an older main without workspace ownership support MUST NOT trigger legacy global enumeration; and
 - no action may acquire `withAutomationLease(undefined)`.
 
-These rules bound what **this tool** discloses. They are not a system-wide guarantee that workspace ownership is unobservable: §8.2 records a sibling RPC that still exposes live-target metadata across workspaces.
+These rules bound what **this tool** discloses. They are not a system-wide guarantee that workspace ownership is unobservable: §8.2 records a sibling RPC (`browser.cdp.info`) whose cross-workspace metadata is now scoped as defense-in-depth, but which — like any holder of the shared CDP port — cannot hard-seal enumeration within the same-OS-user trust ceiling.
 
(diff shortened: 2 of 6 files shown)


openwong2kim:
Merged — thanks for taking on both halves of #580 and following the Option 1 spec so closely. The targetsScoped flag handling and the fail-closed treatment of untagged targets came out exactly right, and the honest framing of the boundary in the docs is appreciated too.

One non-blocking thing if you ever feel like a small follow-up: CodeRabbit pointed out the empty-targets retry checks the unscoped list, so a scoped caller can miss the 1.5s grace when another workspace already has targets. Narrow edge case, no pressure — happy to take it myself otherwise.

PR #592, opened 2026-07-24
fix(terminal): defer dispose during mouse drag to stop remount TypeError

### LINKED ISSUE #582: Uncaught TypeError in xterm mouseup/mousedrag ('dimensions' of undefined) after terminal remount churn; renderer missing CSP
## Summary

Two distinct problems surfaced in the Electron renderer's DevTools console during normal use (multiple workspaces with reattached daemon PTYs, mouse-drag text selection in terminals):

1. A recurring **uncaught `TypeError`** thrown from inside `@xterm/xterm`'s own mouse-tracking code, triggered by terminal mouse drag/selection.
2. Electron's **"Insecure Content-Security-Policy" security warning** (dev-only, but worth tracking so a real CSP gets added before it matters).

Filing both together since they came from the same session's console capture; happy to split if preferred.

## 1. Uncaught TypeError: Cannot read properties of undefined (reading 'dimensions')

```
Uncaught TypeError: Cannot read properties of undefined (reading 'dimensions')
    at get dimensions (@xterm_xterm.js:4641:33)
    at Xt.getMouseReportCoords (@xterm_xterm.js:4493:107)
    at r (@xterm_xterm.js:8643:31)
    at HTMLDocument.mouseup (@xterm_xterm.js:8667:102)
```
(and the `mousedrag` variant at `@xterm_xterm.js:8668:20`)

This fires repeatedly — dozens of times per session in the captured log — always from `HTMLDocument.mouseup` / `HTMLDocument.mousedrag`, i.e. xterm's **document-level** drag-tracking listeners (registered so a selection drag can be released outside the terminal's own element), not a listener scoped to a specific terminal's container.

**Observed pattern in the console capture:** the errors cluster right after workspace reconciliation cycles that re-mount/reattach terminals. The log shows `[AppLayout] Daemon active PTYs...` → `Reconciling workspace: ...` → `[Terminal] Using existing ptyId` → `[useTerminal] daemon reattach ptyId=... (active-at-mount)` repeating **identically four times in a row** on a single load, and the TypeErrors start appearing once `[wmux:reveal] ptyId=... mechanism=retained-catchup` / hidden-pane reveal events start firing on those reattached panes.

**Working hypothesis (not yet confirmed):** `useTerminal.ts` fully disposes and recreates the underlying `Terminal` instance on every mount ([useTerminal.ts:1892](src/renderer/hooks/useTerminal.ts:1892) `terminal.dispose()`, effect keyed on `[ptyId, containerRef]` at [useTerminal.ts:1897](src/renderer/hooks/useTerminal.ts:1897)). `dispose()` nulls internal services (`_renderService`, `_bufferService`), which is what `get dimensions()` reads. If xterm's own document-level `mouseup`/`mousedrag` listeners aren't torn down in the same tick as `dispose()` — or a stale closure over an already-disposed `Terminal` instance still fires — a drag that spans a remount will hit the getter on a half-torn-down instance. The existing teardown comment at [useTerminal.ts:1868-1872](src/renderer/hooks/useTerminal.ts:1868) already documents that "unmount churn" from repeated remounts caused a related class of bug for WebGL contexts (#191 / #197); this may be the same churn surfacing through xterm's core mouse-tracking instead.

The redundant back-to-back reconciliation runs visible in the log (4 identical cycles before any user interaction) are themselves worth checking independently — if `AppLayout`'s reconciliation effect is re-running more than once per actual state change, that's the churn amplifying this and other remount-sensitive bugs.

**Suggested investigation:**
- Confirm why reconciliation logs the same PTY set 4x in a row on load ([AppLayout.tsx](src/renderer/components/Layout/AppLayout.tsx)) — dedupe or guard against redundant re-runs if so.
- Check whether `@xterm/xterm`'s `CoreMouseService`/`SelectionService` removes its `document`-level `mouseup`/`mousedrag` listeners synchronously inside `Terminal.dispose()`, and whether our disposal in [useTerminal.ts](src/renderer/hooks/useTerminal.ts:1892) runs before or after that teardown is safe to interleave with a remount.
- If it's an upstream xterm.js ordering issue rather than something we control, consider a defensive check (e.g. a global `mouseup` capture-phase guard that no-ops if the event's originating terminal instance is already disposed) rather than patching xterm internals.

Not user-blocking per se (terminal still functions), but it's a real uncaught exception firing on ordinary mouse-drag selection, and worth fixing since these can mask other errors and add noise that hides real regressions.

## 2. Electron Security Warning: Insecure Content-Security-Policy

```
Electron Security Warning (Insecure Content-Security-Policy)
This renderer process has either no Content Security Policy set or a policy
with "unsafe-eval" enabled. This exposes users of this app to unnecessary
security risks.
https://electronjs.org/docs/tutorial/security
```

Dev-only (Electron suppresses it in packaged builds), so no user impact today, but it means we currently have no CSP guarding the renderer, or one permitting `unsafe-eval`. Worth adding a proper `Content-Security-Policy` (via `session.defaultSession.webRequest.onHeadersReceived` or a meta tag) scoped tightly enough to not need `unsafe-eval`, so a real policy exists before packaging assumptions change.

## Not actionable (noise, mentioned for completeness)

- `Download the React DevTools for a better development experience` — standard React dev-mode notice, not something we control.

## Execution

- Start a new branch
- Implement the task
- Start a sub-agent (opus level), pass the original GitHub issue as context and ask it to review the implementation
- Fix all feedback from previous step
- Start a sub-agent (opus level), pass the original GitHub issue as context and ask it to design (design only) the e2e tests
- Implement the e2e tests as designed until they pass
- Start opus sub-agent for e2e review
- Open draft PR

openwong2kim:
## What

Fixes the recurring uncaught `TypeError: Cannot read properties of undefined (reading 'dimensions')` that fires from xterm's own mouse-tracking code during terminal remount churn (workspace switch, reconnect). Closes #582.

## Why

xterm's `CoreMouseService` registers document-level `mouseup`/`mousemove` listeners on `mousedown` so a selection drag can be released outside the terminal element. `Terminal.dispose()` nullifies `_renderService` **before** removing those listeners. A `mouseup` landing in that gap runs `getMouseReportCoords`, which reads `_renderService.dimensions` on a half-torn-down instance and throws — dozens of times per session. Not user-blocking (the terminal still works) but a real uncaught exception on ordinary text-selection drags, and the noise masks real regressions.

## Changes

- **`src/renderer/hooks/useTerminal.ts`** — Track whether a mouse drag is active on any `.xterm` element (capture-phase document listeners, installed once). In the cleanup path, if a drag is active, defer only the internal `terminal.dispose()` until `mouseup` (2-second backstop for a drag abandoned outside the window). All PTY listeners are already torn down at that point, so no new data arrives while disposal waits. No xterm internals are patched.
- **`src/renderer/components/Layout/AppLayout.tsx`** — Number the reconciliation cycles in the console so "one cycle walking N workspaces" is distinguishable from "N cycles" (the second part of #582's diagnosis).
- **`src/main/window/createWindow.ts`** — Suppress the dev-only Electron "Insecure Content-Security-Policy" warning in development. Vite's HMR requires `unsafe-eval`; the production CSP remains strict with no `unsafe-eval`, so this is dev-only noise reduction, not a security trade-off (the third part of #582).
- **`src/renderer/hooks/__tests__/terminalDragGuard.test.ts`** — New: pins the drag-detection logic (activate on `.xterm` mousedown, clear on document mouseup / window blur, single listener install).

## Verification

- `tsc --noEmit -p tsconfig.json` — clean
- `vitest run terminalDragGuard.test.ts` — 7/7 pass
- eslint on changed files — no new findings (pre-existing findings untouched)
- Live dogfood: booted the dev app (isolated `wmux-dogfood582` profile) with `ELECTRON_ENABLE_LOGGING=1`, drove the live renderer over CDP, and replayed the #582 drag→document-`mouseup` race on the real `.xterm` element — no exception thrown, zero uncaught exceptions during the session, the modified `useTerminal` module confirmed loaded, and the real daemon-reattach path (the remount churn that triggered the bug) ran error-free on boot.

## Versioning

No version bump (per repo policy). CHANGELOG entry under `[Unreleased]`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a workspace-switching issue where dragging text across a terminal could trigger recurring errors.
  * Improved terminal teardown behavior during active mouse drags to avoid crashes and stray input during the transition.
* **Chores**
  * Enhanced diagnostic logging for workspace/terminal reconciliation activity.
  * Reduced an Electron security-policy warning in development (production protections remain enabled).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Dragging your mouse across a terminal during a workspace switch no longer throws a recurring uncaught error.** Selecting text in a terminal registers document-level mouse listeners (so a drag that leaves the terminal can still be released), and on a remount — a workspace switch, a reconnect — `Terminal.dispose()` nullified xterm's internal render service before those listeners came down. A `mouseup` landing in that gap read `dimensions` off a half-torn-down instance and threw `TypeError: Cannot read properties of undefined (reading 'dimensions')`, dozens of times per session. wmux now tracks whether a drag is active on any terminal and defers only the internal `dispose()` until the mouse is released (with a 2-second backstop for a drag abandoned outside the window), closing the race without patching xterm internals; all PTY listeners are already torn down at that point, so no new data arrives while disposal waits. The redundant back-to-back reconciliation runs visible on load are now numbered in the console — so "one cycle walking four workspaces" is distinguishable from "four cycles" — and the dev-only Electron "Insecure Content-Security-Policy" warning is suppressed in development (Vite's HMR requires `unsafe-eval`; the production CSP remains strict with no `unsafe-eval`).
+
 ## [3.33.0] — 2026-07-24
 
 ### Added
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -1,3 +1,16 @@
+// #582: Suppress Electron's dev-only "Insecure Content-Security-Policy"
+// warning at the earliest possible point — before `app` ready and before any
+// BrowserWindow is created. Vite's HMR requires `unsafe-eval`, so the warning
+// is unavoidable in dev; the production CSP (enforced in createWindow) is
+// strict with no `unsafe-eval`, so this is dev-only noise reduction, not a
+// security trade-off. This MUST run before createWindow()/loadMainRenderer()
+// so the override is active before the first renderer navigates — setting it
+// inside createWindow (after loadMainRenderer) leaves a window where Electron
+// reads the flag before the override lands.
+if (MAIN_WINDOW_VITE_DEV_SERVER_URL) {
+  process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true';
+}
+
 process.on('unhandledRejection', (reason) => {
   console.error('[Main] Unhandled rejection:', reason);
 });
--- a/src/main/window/createWindow.ts
+++ b/src/main/window/createWindow.ts
@@ -141,11 +141,20 @@ export function createWindow(opts: { deferLoad?: boolean } = {}): BrowserWindow 
   }
 
   // CSP header — production only.
-  // In development, Vite serves scripts from localhost with inline module loaders,
-  // which are incompatible with strict CSP. We only enforce CSP in production builds.
+  // In development, Vite serves scripts from localhost with inline module
+  // loaders and eval-based HMR, which are incompatible with strict CSP.
+  // We only enforce CSP in production builds.
   // 'unsafe-inline' in style-src is required because Tailwind CSS and xterm.js
   // inject inline styles at runtime; removing it breaks UI rendering.
-  if (!MAIN_WINDOW_VITE_DEV_SERVER_URL) {
+  //
+  // #582: In dev, we suppress Electron's "Insecure Content-Security-Policy"
+  // warning via ELECTRON_DISABLE_SECURITY_WARNINGS. Vite's HMR requires
+  // unsafe-eval, so any dev CSP we set would still trip the warning. The
+  // production CSP below is strict (no unsafe-eval), so this suppression is
+  // dev-only noise reduction, not a security trade-off.
+  if (MAIN_WINDOW_VITE_DEV_SERVER_URL) {
+    process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true';
+  } else {
     const cspPolicy = "default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self'; font-src 'self'; frame-src 'self' https: http:";
 
     mainWindow.webContents.session.webRequest.onHeadersReceived((details, callback) => {
--- a/src/main/window/createWindow.ts
+++ b/src/main/window/createWindow.ts
@@ -147,14 +147,14 @@ export function createWindow(opts: { deferLoad?: boolean } = {}): BrowserWindow 
   // 'unsafe-inline' in style-src is required because Tailwind CSS and xterm.js
   // inject inline styles at runtime; removing it breaks UI rendering.
   //
-  // #582: In dev, we suppress Electron's "Insecure Content-Security-Policy"
-  // warning via ELECTRON_DISABLE_SECURITY_WARNINGS. Vite's HMR requires
-  // unsafe-eval, so any dev CSP we set would still trip the warning. The
-  // production CSP below is strict (no unsafe-eval), so this suppression is
-  // dev-only noise reduction, not a security trade-off.
-  if (MAIN_WINDOW_VITE_DEV_SERVER_URL) {
-    process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = 'true';
-  } else {
+  // #582: The dev-only "Insecure Content-Security-Policy" warning is suppressed
+  // via ELECTRON_DISABLE_SECURITY_WARNINGS — but that is now set at the very
+  // top of the main entry (src/main/index.ts), before this function runs, so
+  // the override is live before the first renderer navigates. Vite's HMR
+  // requires unsafe-eval, so any dev CSP we set would still trip the warning;
+  // the production CSP below is strict (no unsafe-eval), so this is dev-only
+  // noise reduction, not a security trade-off.
+  if (!MAIN_WINDOW_VITE_DEV_SERVER_URL) {
     const cspPolicy = "default-src 'self'; script-src 'self'; style-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self'; font-src 'self'; frame-src 'self' https: http:";
 
     mainWindow.webContents.session.webRequest.onHeadersReceived((details, callback) => {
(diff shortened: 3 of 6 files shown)


openwong2kim:
@codex review

openwong2kim:
## Review follow-up (`3c473634`)

Addressed every actionable item from CodeRabbit and Codex. All changes are hardening of the same #582 guard — no scope creep, no version bump.

### Is this a root-cause fix?
Not in the deepest sense, and that is deliberate. The true root cause is in **xterms own dispose order**: `Terminal.dispose()` nullifies `_renderService` *before* removing its document-level mouse listeners, so a `mouseup` in that gap throws from `getMouseReportCoords`. Patching bundled xterm internals would be fragile across updates, so this guard instead **avoids tripping the race** by deferring our `dispose()` until the drag releases. That trade is sound only if the guard is airtight — and the reviews found four gaps, now closed:

### CodeRabbit
- **createWindow.ts (security): set `ELECTRON_DISABLE_SECURITY_WARNINGS` before the renderer loads.** ✅ Moved to the very top of `src/main/index.ts` (before `app` ready / before `createWindow`/`loadMainRenderer`). `createWindow` now only branches on the production CSP. The override is live before the first navigation.
- **useTerminal.ts (critical): install drag listeners before first interaction.** ✅ `_ensureDragListeners()` is now called in the **terminal mount effect**, before the user can start a drag. The lazy install from the cleanup path missed the *first* drag→remount (the `mousedown` fired before the listener existed → flag read false → immediate dispose). Idempotent, so only the first mount pays.
- **useTerminal.ts (major): release deferred-disposal resources on either path.** ✅ The `mouseup` listener and backstop timer are now stored and cleared inside `safeDispose`, so neither leaks when the other path wins.
- **terminalDragGuard.test.ts (nitpick): assert registration counts.** ✅ Spies on `document`/`window.addEventListener` before the first call and asserts exactly 2 + 1 registrations, unchanged across repeated calls.

### Codex
- **(P2) Listener installed too late — same as the CodeRabbit critical above.** ✅ Fixed via the mount-effect install.
- **(P2) Detach stale `onData` before deferring dispose.** ✅ `terminal.onData(...)` is now captured into `onDataDisposable` and disposed in the cleanup **before** the defer block, so xterm→PTY input (incl. mouse-tracking reports in DECSET modes) cannot flow into an already-detached PTY during the ≤deferral window.
- **(P2) Keep disposal deferred while the drag is active.** ✅ The 2s backstop now re-checks `isTerminalDragActive()` and keeps waiting (bounded retries → leak guard) instead of cutting down a genuine long selection mid-drag, which would have reopened the exact race.

### Verification
- `tsc --noEmit -p tsconfig.json` — clean
- `vitest run src/renderer/hooks/__tests__/` — 529/529 pass (drag-guard suite now 8/8, +1 regression test for a drag started before any cleanup read)
- `eslint` on changed files — no new findings (pre-existing findings untouched)
- No CHANGELOG change: user-facing behavior is unchanged (same "drag during workspace switch no longer throws"); these are internal hardening refinements.

openwong2kim:
**[wmux-hermes] Code Review - Priority: P2 (Important bugfix, closeable after review)**

This PR closes #582 with a clean, well-contained workaround for a genuine xterm-internal race: it introduces a module-level drag guard (isTerminalDragActive()) with capture-phase document listeners on .xterm mousedown/mouseup plus a window blur clear, and in the useTerminal cleanup path it defers only the internal terminal.dispose() while a drag is active - a 2s x 5-retry backstop force-disposes as a leak guard for a stuck button. All PTY-side listeners (including onData, refactored to a disposable) are torn down before the defer, so no stray input flows into a half-disposed instance during the wait window. The CSP-warning suppression (ELECTRON_DISABLE_SECURITY_WARNINGS set at the top of src/main/index.ts, gated on MAIN_WINDOW_VITE_DEV_SERVER_URL) and the reconcile-cycle numbering in AppLayout.tsx are both dev-only diagnostics and are correctly scoped; the production CSP is untouched and remains strict. Test coverage is good - terminalDragGuard.test.ts pins the detection logic across 8 cases including the key regression (drag starting before the first getter read) and single-install guarantee - though the useTerminal defer/dispose path itself is not unit-covered (hard to test without mounting the hook, but worth a note).

A few observations for the maintainer:

- Concurrent multi-terminal unmount is handled correctly - e.g. a workspace switch tearing down N terminals. Each cleanup closure registers its own mouseup listener; a single mouseup fires all of them, each guarded by its own _disposed flag, and the try/catch around terminal.dispose() absorbs the StrictMode double-cleanup case. Worth confirming live with 3+ terminals during a workspace switch, since the unit tests cannot exercise this.
- The backstop's isTerminalDragActive() re-check means a genuinely long selection (user holding the button over 10s, e.g. scroll-selecting a large buffer) will still be force-disposed after 5 retries. This is an unlikely but real edge; the terminal.dispose() at that point is correct (PTY listeners are already gone), so it is a UX nit (selection ends early) not a correctness issue - acceptable as-is.
- Naming and structure are clear (safeDispose, scheduleBackstop, _retriesLeft). The only minor nit is that _terminalDragActive/_dragListenersInstalled are module-global singletons - fine for a single Electron renderer, but if useTerminal is ever used in a second window/process they would be shared. Not a concern today.
- CHANGELOG entry is thorough and accurate. No version bump per repo policy; CHANGELOG under [Unreleased].

Recommendation: approve and merge after review. The fix matches the severity and diagnosis in #582, the workaround avoids patching xterm internals, the regression test directly targets the original failure mode, and the dev-only changes carry no production security surface. No blocking issues found.

Issue #517, opened 2026-07-20
[Feature]: Choice of browser backend (built-in vs external Chrome) + lightweight mode for the built-in browser


p-poppe:
### Problem / motivation

The built-in wmux browser panel is heavy and sluggish — slow to respond, and resource-intensive when several workspaces are open (each embedded Chromium instance multiplies memory/CPU). For everyday research/reading it's often not worth using.

At the same time, agents running in wmux (e.g. Claude Code) can drive the user's **real Chrome** via a browser extension. So there are effectively two possible browser backends, but no way to tell wmux which one to use.

### Proposed solution

1. A config option (global, ideally also per-workspace) to select the **browser backend**:
   - built-in wmux browser (embedded Chromium), or
   - **delegate to external Chrome / the Claude Chrome extension** — so browser tasks open in the user's real Chrome instead of the embedded panel.
2. A **lightweight mode** for the built-in browser when it is kept:
   - throttle / suspend the panel when it isn't focused,
   - suspend background tabs,
   - a lower-resource rendering profile.

The goal: users who find the embedded browser too heavy can offload to real Chrome, and those who keep it can cap its resource use.

### Alternatives considered

Currently the embedded panel is the only option; there's no toggle to disable it or route browser work elsewhere. Manually using a separate Chrome window loses the integration.

### Additional context

Motivated by heavy multi-workspace usage (many agent panes open at once), where the embedded browser is a noticeable resource cost relative to how much it's used.

### Confirmation

- [x] I searched existing issues and Discussions and this is not a duplicate.

openwong2kim:
@p-poppe partial progress on this: #528 just landed a lightweight mode for the built-in browser (Settings → Terminal → "Lightweight background browsers", off by default while it bakes — will ship in the next release).

While it's on, a browser pane that isn't actually visible (hidden workspace, zoomed-away pane, unselected tab, minimized window) gets its CPU throttled, and goes back to full speed the moment you look at it or any automation touches it. Measuring this confirmed your report, by the way — a hidden pane's Chromium keeps running rAF at 60fps and full-speed timers even when its workspace has been hidden for hours.

Two honest caveats on the rest of the request:

- **Memory doesn't move yet.** Throttling keeps the renderer resident, so this is CPU relief only. Real memory relief needs page discard/suspend (unload the renderer, reload on revisit), which has state-loss tradeoffs we want to spec separately. Leaving this issue open for that.
[...]

PR #593, opened 2026-07-24
fix(deck): retire a heartbeat-surfaced complete pane instead of re-reviewing it every interval

### LINKED ISSUE #561: Heartbeat re-surfaces idle-complete panes every interval until budget/rate exhausts
Follow-up from #553 (orchestrator level-pull redesign). Surfaced in the 3-way review (GLM), accepted as a deliberate follow-up rather than a blocker because the rate limiter and wake budget already bound the cost.

## What

When a hook-bridged TUI agent (OpenCode, etc.) finishes a turn, #553 marks its pane `complete` — an *attention* status — so the fleet badge, `pane_list`, and the heartbeat level review all see the completion (fixing the "stuck running forever" bug). But `complete` is retained on the surface until a genuinely new turn clears it (fresh output / awaiting-input). If the user just leaves the finished pane sitting there, it stays `complete`.

The level-review heartbeat (`DeckHeartbeat` → `CommanderEventCoalescer.flushSnapshot`) scans exactly the attention statuses (`awaiting_input` / `waiting` / `complete` / `error`). So an armed (autonomy ≠ off) workspace with an idle-but-`complete` pane is re-surfaced to the brain on every heartbeat interval (~3 min) until something changes — spending ambient tokens re-reviewing a pane the orchestrator already saw finish.

## Why it's not urgent

The wake is not unbounded — it still passes through the full gate stack:
- the sliding-window rate ceiling (6 wakes/min/workspace),
- the consecutive wake budget,
- `lastWakeAt` freshness (a review within `intervalMs` of the last wake is skipped).

So the worst case is a slow ambient token drip on an armed workspace with a finished-and-forgotten pane, not a storm.

## Options to consider

1. **Retire `complete` after the heartbeat surfaces it once** — clear the attention entry (or mark it "seen") after the first level review folds it in, so a subsequent heartbeat doesn't re-raise the same completed pane. New activity re-arms it exactly as today.
2. **Age out `complete`** — expire the attention entry after some idle TTL (like the activity heuristic decay), so a long-finished pane settles to idle and drops out of the review set.
3. **Exclude `complete` from the heartbeat's attention filter** (but keep it for the fleet badge / `pane_list`) — the level review would then only chase `awaiting_input` / `waiting` / `error` (panes actually *blocked* on something), which is arguably the truer "needs a look" set. `complete` is informational, not blocking.

Option 3 is the cleanest split conceptually (the heartbeat is a missed-*judgment* net; a completed pane needs no judgment), but it changes what the level review catches — worth thinking through against the "recover a dropped stop edge" goal that motivated the snapshot path.

## Pointers

- `src/main/pipe/handlers/hooks.rpc.ts` — the `agent.stop` → `agentStatus: 'complete'` broadcast
- `src/main/deck/DeckHeartbeat.ts` `isAttentionStatus` — the heartbeat's scan filter
- `src/main/deck/CommanderEventCoalescer.ts` `flushSnapshot` — the attention filter on the snapshot side
- `src/renderer/stores/slices/paneSlice.ts` `setSurfaceAgentStatus` — where `complete` is retained / cleared

snowyukitty:
## What

Closes #561. On an armed (autonomy ≠ off) workspace, the level-review heartbeat re-surfaced the **same** idle-and-`complete` pane to the brain on every interval (~3 min), spending ambient tokens re-judging a completion the orchestrator had already seen. This retires a `complete` pane after the heartbeat surfaces it once, and re-arms it on new activity.

## Root cause

When a hook-bridged agent (Claude Code, OpenCode, …) ends a turn, #553 marks its pane `complete` — an *attention* status — and retains it on the surface until genuinely new activity clears it.

The heartbeat (`DeckHeartbeat` → `CommanderEventCoalescer.flushSnapshot`) scans the attention statuses (`awaiting_input` / `waiting` / `complete` / `error`), and its `lastWakeAt` gate only skips a review that lands *within* `intervalMs` of the last wake — a review exactly one interval later is not skipped. So a workspace holding an idle-but-`complete` pane got that pane folded into a fresh brain turn on every heartbeat, forever, until something changed. The rate ceiling and wake budget bound the cost to a slow drip rather than a storm (which is why #553 accepted this as a follow-up), but it is still ambient tokens spent re-reviewing a pane the orchestrator already saw finish.

Scanning `complete` is deliberate and load-bearing, though: it is exactly how the heartbeat recovers a **dropped `agent.stop` edge** — a dropped stop shows up as `complete` in level state, and the snapshot path exists to catch it (`DeckHeartbeat.ts:1-11`). So the fix has to keep surfacing a *newly*-complete pane while stopping the *repeat*.

## Fix — Option 1 (retire-after-surface)

`CommanderEventCoalescer` tracks, per workspace, the ptyIds whose `complete` level-state has already been folded into a snapshot wake (`snapshotSurfacedComplete`). In `flushSnapshot`:

- **Re-arm first** — `reconcileSurfacedComplete` drops any tracked ptyId that is no longer `complete` in the current snapshot (new activity moved it on, or it left the fleet). The snapshot is the ground truth of current state, so this runs before every gate — including the early-returning ones (pending decision, auto-wake off, policy `none`) — and regardless of whether the flush wakes.
- **Filter** — a `complete` pane already in the set is dropped from the attention set, so it is not re-reviewed.
- **Retire synchronously at the commit point** — the `complete` ptyIds going into the prompt are added to the set right before the dispatch, not when the async `runTurn` settles. At that instant they are exactly the currently-complete panes, so the add cannot resurrect a pane that a *newer* snapshot has since re-armed (a settle-time add can, and did).
- **Un-retire if the turn is then refused as busy** — the sync `isBusy(workspaceId)` gate only sees this workspace's manager mid-turn, but `runTurnForWorkspace` also rejects `busy` for the fleet-wide concurrent-turn slot gate and for a manager that is non-idle without being `busy`. Those land *after* the commit point, so the `busy` branch rolls this dispatch's retirement back.

The add/remove asymmetry is deliberate: a late **add** suppresses a completion a newer snapshot already re-armed — a review silently lost, and a finished pane never re-arms itself because it only moves when the brain drives it. A late **remove** at worst re-arms a pane a newer flush retired for real, costing one redundant review that the next reconcile settles. Fail open toward surfacing. The non-busy failure branch keeps retiring, matching how it advances the watermark to avoid a poison loop.

I chose Option 1 over the issue's Option 3 (exclude `complete` from the heartbeat's filter entirely). Option 3 is conceptually cleaner but would defeat the snapshot path's documented reason to exist — recovering a dropped stop edge. Option 1 keeps that recovery (first heartbeat still surfaces a newly-complete pane) while killing the repeat.

## Properties held

- **Dropped-stop recovery preserved** — a newly-`complete` pane is still surfaced once.
- **Scoped to `complete` only** — genuinely-blocked panes (`awaiting_input` / `waiting` / `error`) still re-surface every heartbeat, because those are still blocked on something; `complete` is informational.
- **Re-arms on new activity** — a pane that runs again and completes again surfaces fresh.
- **Nothing dropped when the brain never looked** — a pane is retired only for a turn that actually ran; every `busy` rejection (sync gate, fleet slot gate, or a manager gone non-idle) leaves it armed for the next heartbeat.
- **Assist mode is a deliberate no-op** — under `value-filtered` (assist), `snapPanes` is narrowed to `awaiting_input`, so no `complete` pane ever reaches the retire and nothing is retired there. That is correct rather than a gap: assist already drops a lone finished pane before the worthiness gate, so there is no repeat review to suppress. Noted in a comment at that branch.
- **Bounded** — the set is pruned each flush to only ptyIds still `complete` in the latest snapshot, so it can never outgrow the live fleet.
- **Single source of truth** — the change is entirely in the coalescer (the documented authority on whether a wake fires); `DeckHeartbeat`'s cheap pre-check is unchanged, and a flush it triggers for an already-retired pane simply drops without a turn.

## Tests

Added to `CommanderEventCoalescer.snapshot.test.ts`:

- surfaces a complete pane **once**, then stops re-reviewing it on later heartbeats (asserts only one wake spent);
- **re-arms** once the pane shows new activity, so its next completion surfaces again;
- a **sync busy reject** leaves the pane armed — it re-surfaces once the brain frees up;
- an **async busy reject** (idle brain, fleet slot gate full) leaves the pane armed, surfaces the completion on the next heartbeat, and only then retires it;
- a **non-busy failure** still retires, so a poison prompt cannot loop the brain every interval;
- **two panes completing at once** retire and re-arm independently;
- re-arms even when the re-observing flush **early-returns** (auto-wake toggled off between the completion and the re-observation);
- retiring one complete pane never **suppresses another** attention pane in the same workspace;
- **`error` panes are not retired** — a still-broken pane keeps surfacing (scope guard).

`vitest` on the deck subsystem: **395/395 green**. `tsc -p tsconfig.json --noEmit` clean. `eslint` clean on the changed files (the one warning in that test file is a pre-existing `for (const _ of …)` in the untouched rate-ceiling test).

Per the repo's release policy this branch does not bump `package.json`; the user-facing note is under `## [Unreleased]` in `CHANGELOG.md`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented already-completed Fleet Snapshot panes from being repeatedly surfaced for re-review on every heartbeat.
  * Completed panes are surfaced once for recovery, then retired until they show new activity.
  * Recovery is preserved for dropped “turn ended” signals and transient busy conditions.
  * Blocked and error panes continue to receive recurring prompts as before.


snowyukitty:
Addressed both CodeRabbit findings in `493eaf72` — both were valid:

1. **Reconcile before the gates.** `reconcileSurfacedComplete` was running *after* the decision / auto-wake / policy early-returns, so a retired pane seen as `running` on a flush that early-returns (workspace `off`, auto-wake disabled, decision pending) was never re-armed, and its next real completion got wrongly suppressed. Moved it to immediately after `ensureState()`, so every observed snapshot updates the re-arm state regardless of whether that flush can wake.

2. **Retire synchronously at commit, not at settle.** The stale-add race was real: `surfacedCompleteIds` captured before `runTurn`, added in the async `.then`/`.catch`, could re-add a pane a newer snapshot had already re-armed. The panes are now retired synchronously at the commit point (after the sync busy gate), so a retire can never re-add a pane a later snapshot re-armed, and a future `reconcile` still drops any that move off `complete`. The mid-turn (sync-busy) rejection still returns *before* the retire, so a pane the brain never reviewed stays armed exactly as before — the "busy leaves it armed" test still passes.

Added a test that a retired pane re-arms across an early-returning flush (auto-wake toggled off between the completion and the re-observation). Deck suite 392/392 green; `tsc` + `eslint` clean on the changed files.

openwong2kim:
## [wmux-hermes] Code Review — Priority: P3 (Minor bugfix / optimization, merge after review)

### What this PR changes
This PR closes #561 by retiring an idle-`complete` pane from the heartbeat's attention set after it has been surfaced once, so an armed workspace no longer spends ambient tokens re-reviewing a finished pane every ~3-min interval. The implementation is entirely contained in `CommanderEventCoalescer`: a new per-workspace `snapshotSurfacedComplete: Set<string>` tracks ptyIds whose `complete` level-state has already been folded into an accepted snapshot wake. Three moving parts — `reconcileSurfacedComplete` (re-arm, runs *before* any gate so an early-returning flush still re-arms), the attention-filter exclusion (drops already-surfaced `complete` panes), and `retireSurfacedComplete` (adds surfaced panes synchronously at the commit point, after the busy gate, so a busy-rejected pane stays armed). The design deliberately preserves dropped-stop-edge recovery (a newly-`complete` pane still surfaces once) while scoping retirement to `complete` only — `awaiting_input`/`waiting`/`error` keep re-surfacing.

### Implementation feedback
The reasoning is tight and the execution matches the PR description precisely. Three details I want to call out positively: (1) `retireSurfacedComplete` runs at the commit point **synchronously** rather than after `runTurn` settles, which correctly closes the stale-add race where a later snapshot might have re-armed the pane in the async gap; (2) `reconcileSurfacedComplete` iterates over a spread-array copy (`[...set]`) while mutating the set, which avoids the classic concurrent-modification footgun; (3) the reconcile runs before the gate stack, so the early-return paths (pending decision, auto-wake off, policy `none`) still re-arm — and there is a dedicated test pinning exactly that wedge case.

### Edge cases / risks
One subtle point worth confirming: under the `value-filtered` (assist) policy, `snapPanes` is restricted to `awaiting_input` panes, so `snapPanes.filter(p => p.agentStatus === 'complete')` is always empty and **no complete pane is ever retired in assist mode**. I believe this is benign — in assist mode a lone `complete` pane never enters `snapPanes` and (per existing worthiness logic) does not force a wake on its own, so there is no ambient drip to suppress there. But it is worth a one-line confirmation in the PR description so a future reader doesn't read the non-retirement in assist mode as a bug. The other concern — whether a `complete` pane lingering in the `attention` array affects the worthiness gate even when value-filtered out of `snapPanes` — is pre-existing behavior this PR does not touch. Multi-workspace isolation is sound (the set is keyed on `WsState`).

### Tests
Six new cases cover the important invariants: surface-once-then-stop, re-arm-on-new-activity, busy-reject-stays-armed, no-cross-suppression, re-arm-through-an-early-returning flush, and `error`-not-retired. Coverage is thorough; the only gap I'd note is no test for *multiple* simultaneously-`complete` panes retiring and re-arming independently (the cross-suppression test pairs one `complete` with one `awaiting_input`), but that is a minor addition rather than a blocker. Reported `vitest` result of 391/391 green on the deck subsystem is consistent with the scope of the change.

### Recommendation
**Approve and merge.** Low-risk, well-scoped, well-tested, and it correctly trades off the competing constraints (recover dropped stops vs. stop re-reviewing finished panes). The only follow-up worth considering is whether to surface the assist-mode non-retirement behavior explicitly in the PR body or a code comment near the `value-filtered` branch.

openwong2kim:
Thanks for taking this on, and for picking Option 1 over Option 3 — keeping the first surface so a dropped `agent.stop` is still recovered is the right call, and the write-up on why is clear.

One thing I think needs fixing before this lands.

## A pane can get retired without the brain ever reviewing it

`493eaf72` moved the retire to the synchronous commit point. Its message says the mid-turn rejection still returns before the retire, so a pane the brain never reviewed stays armed. That holds for the sync gate — but there's a second busy source it doesn't cover.

The coalescer's `isBusy` only looks at this workspace's own manager:

```ts
// deck.handler.ts:779
isBusy: (workspaceId) =>
  managers.get(workspaceId)?.manager.getStatus().status === 'busy',
```

`runTurnForWorkspace` has another one — the fleet-wide slot gate:

```ts
// deck.handler.ts:597
const token = runOpts.queued ? await ... : globalTurnGate.tryAcquire(workspaceId);
if (!token) {
  return { ok: false, code: 'busy' as const };
}
```

(`preMgr.getStatus().status !== 'idle'` on :588 is a third — it also reports `busy` for a `disposed` manager, which `=== 'busy'` doesn't catch.)

So when the global slots are saturated, a workspace whose own brain is idle passes the sync gate, commits, retires the pane at :401 — and *then* the turn gets rejected. The snapshot drops, exactly as the `r.code === 'busy'` branch intends, except the pane is already in `snapshotSurfacedComplete`.

What makes this worse than one missed review: the re-arm condition is "the pane is next observed non-`complete`", and a finished pane doesn't move on its own. It moves when the brain hands it something — and the brain is now filtering it out. So it doesn't self-heal, it sits there until a human notices.

It takes a dropped `agent.stop` *and* a saturated gate to bite, so it won't be common. But that combination is precisely what the snapshot path exists to catch, on precisely the busy fleet where it matters.

## Repro

Drop this next to the existing snapshot suite. It uses the harness already in the file, just with `setRunResult` instead of `setBusy` — the brain is idle, the fleet gate is what rejects:

```ts
it('an async busy reject (fleet gate full) leaves the pane ARMED', async () => {
  const h = mk();
  // Workspace brain is idle, so the flush commits and dispatches — but the
  // fleet-wide slot gate rejects the turn. The brain never reviewed it.
  h.setRunResult({ ok: false, code: 'busy' });
  h.c.flushSnapshot('ws-1', snap([pane({ ptyId: 'ptyA', agentStatus: 'complete' })]));
  await settle();
  expect(h.prompts).toHaveLength(1);
  expect(h.c.getWakeBudgetRemaining('ws-1')).toBe(100); // no wake spent

  // Gate frees up. The completion must still surface — this is the
  // dropped-stop recovery the whole snapshot path exists for.
  h.setRunResult({ ok: true });
  h.c.flushSnapshot('ws-1', snap([pane({ ptyId: 'ptyA', agentStatus: 'complete' })]));
  await settle();
  expect(h.prompts).toHaveLength(2);
});
```

On the branch as it stands this fails with `expected 2, got 1`. The rest of the suite is green — nothing currently exercises this path.

## Suggested fix

Keep the synchronous retire (it does close the stale-add race CodeRabbit found), and un-retire in the busy branch:

```ts
const retiredThisFlush = snapPanes
  .filter((p) => p.agentStatus === 'complete')
  .map((p) => p.ptyId);
this.retireSurfacedComplete(st, retiredThisFlush);
```

```ts
} else if (r.code === 'busy') {
  // The turn never ran (fleet slot gate, or the manager left idle under us),
  // so the brain never reviewed these — put them back so the next heartbeat
  // surfaces the completion. Safe to do late, unlike the add: removing only
  // ever re-arms, so a stale removal costs one extra surface, never a lost one.
  for (const id of retiredThisFlush) st.snapshotSurfacedComplete.delete(id);
  st.phase = st.buffer.size > 0 ? 'buffering' : 'idle';
  if (st.buffer.size > 0) this.restartDebounce(workspaceId, st);
}
```

The asymmetry is the point: a late *add* can suppress a completion a newer snapshot already re-armed (the race you fixed), a late *remove* can only cause one redundant review. Fail-open toward surfacing.

Retiring on the non-busy failure branch I'd leave as is — that matches how the same branch advances the watermark to avoid a poison loop, and it's consistent.

## Smaller things

- `CHANGELOG.md` conflicts with main now (#517 and #592 both landed entries under `[Unreleased]`). Just needs the entry moved under the existing `### Fixed`.
- The description still describes the pre-`493eaf72` behaviour — "the panes surfaced this flush are added to the set only where the edge watermark also advances (accept + non-busy failure + catch), never on a busy reject". Worth updating once the above is in, since that's the property people will read it for.
- Both commits carry `Co-Authored-By: Claude` / `Claude-Session:` trailers. We don't keep those in this repo's history — could you drop them before we squash?

Happy to look again once the busy path is covered.

snowyukitty:
Thank you — that's a genuinely good catch, and the write-up saved me the diagnosis. You're right on every point; all four are fixed in `89324251`.

## The busy path — confirmed, and it was worse than one missed review

I traced it against the handler and the gap is exactly where you said. The coalescer's sync gate is `isBusy` (`deck.handler.ts:779`), which tests `=== 'busy'` on **this** workspace's manager. `runTurnForWorkspace` rejects `busy` for two things that gate cannot see: the fleet-wide slot gate (`:599`) and `preMgr.getStatus().status !== 'idle'` (`:588`), which also catches a `disposed` manager that `=== 'busy'` misses. Both land *after* the commit point where `493eaf72` retires, so the pane goes into `snapshotSurfacedComplete` for a turn that never ran.

And your point about it not self-healing is the part that makes it a bug rather than a missed interval: re-arming needs the pane observed non-`complete`, a finished pane only moves when the brain drives it, and the filter is what's stopping the brain from seeing it. It sits there.

I dropped your repro in as-is. On `493eaf72` it fails with `expected 2, got 1`; with the fix it passes — I verified both directions rather than just the green one.

Implemented as you suggested: the retire stays synchronous at the commit point (that's what closes the stale-add race), and the `busy` branch un-retires that dispatch's ptyIds via a new `rearmSurfacedComplete`. I put your asymmetry argument in the doc comment there, because it's the non-obvious part — a late *add* silently loses a review, a late *remove* costs one redundant one, so late removal is the safe direction and a plain `delete` needs no observation-guard. Left the non-busy failure branch retiring, for the watermark-consistency reason you gave.

Three more tests around it: the non-busy-failure counterpart (pins that the branches deliberately differ), two panes completing together retiring and re-arming independently — the multi-pane gap you flagged in the first review — and the sync-busy case that was already there.

## The smaller things

- **CHANGELOG** — rebased onto `9929833c`; the #561 entry now sits under the existing `### Fixed` alongside #592/#594, with #517's `### Added` above it. Append-merge, nothing of yours touched. Also broadened that entry's last sentence, since "the brain was mid-turn" was the same incomplete claim as the description.
- **Description** — rewritten. The retire section now describes the commit-point retire plus the busy rollback and the asymmetry, instead of the pre-`493eaf72` watermark story.
- **Assist mode** — confirmed benign, and documented both places. Under `value-filtered`, `snapPanes` is narrowed to `awaiting_input`, so no `complete` pane reaches the retire; and since assist drops a lone finished pane before the worthiness gate, there's no repeat review to suppress there. There's now a comment at that branch saying so, plus a bullet in the description, so the non-retirement doesn't read as an oversight.
- **Trailers** — dropped from all commits; the branch is force-pushed with clean messages. Sorry about those, noted for next time.

Also picked up CodeRabbit's minor one on the test harness: a supplied `isAutoWakeEnabled` predicate now stays live across flushes instead of being read once at construction, with `setAutoWake` as an explicit override.

Deck suite **395/395** green, `tsc --noEmit` clean, `eslint` clean on the changed files (the one warning there is the pre-existing `_` in the untouched rate-ceiling test). Still confined to `CommanderEventCoalescer.ts` and its snapshot suite — `DeckHeartbeat` untouched.

openwong2kim:
Reviewed and merging. Thank you — this is a genuinely careful piece of work.

What stood out: putting `reconcileSurfacedComplete` ahead of every gate (so a flush that early-returns at the pending-decision / auto-wake / policy-none branch still re-arms), retiring synchronously at the commit point rather than at settle time, and rolling the retirement back on the async `busy` reject that the sync `isBusy()` gate cannot see. Those are exactly the three places this would have quietly lost a review, and the tests pin each one down.

Also appreciated the call on Option 1 over the issue's Option 3 — keeping the dropped-`agent.stop` recovery alive while killing the repeat is the right trade, and scoping the retire to `complete` only (with the `error` scope-guard test) keeps genuinely-blocked panes surfacing.

CI is green across all three platforms and the release-policy bits are correct (no version bump, entry under `[Unreleased]`). Merging.

PR #595, opened 2026-07-24
feat(browser): choose browser backend — built-in panes vs external OS browser (#517)

openwong2kim:
## Summary

Closes the backend-choice half of #517. Adds a global **Browser backend** setting (Settings → Terminal): `Built-in` (default, unchanged) or `External`, which routes browser opens to your real OS default browser and mounts **no embedded Chromium at all** — so a workspace that opens pages costs wmux zero browser memory.

The lightweight/discard half of #517 already shipped (#528/#530); this is the "delegate to real Chrome" half the issue also asked for.

## What changed

- **New setting, main-owned.** `BrowserBackendStore` persists the value and reads it **synchronously at boot**, so the very first agent/RPC open after launch honors it — no renderer-push settled-gate, no startup race that could fail open to built-in. Settings UI mirrors it over IPC (`browser:get/set-backend`) and the control is disabled until boot hydration lands.
- **External mode governs every open, not just agents.** An agent's `browser_open`/`browser_navigate` **and** user-facing affordances — the pane browser button, command palette, keyboard shortcut, sidebar port badge, terminal-link clicks, project-seeded browser leaves — all delegate to `shell.openExternal` via a single renderer guard (`maybeDelegateExternalBrowser`). (Owner dogfooding caught that an agent-only scope left the pane button still spawning the embedded Chromium the user opted out of.)
- **Fire-and-forget, honestly.** wmux gets no handle on the OS tab, so tabs aren't tracked/listed/closable, and deep-automation tools (click, screenshot, extract, …) fail closed with one shared `EXTERNAL_BACKEND_UNSUPPORTED` contract error routed through a **single target-resolution choke point** — never a generic target-miss, never a silent fallback onto another workspace's pane.
- `browser_tabs new` returns a `BrowserTabsResult` external variant the MCP consumer validates (so a delegated open isn't mis-reported and retried into duplicate tabs); the `browser_wait` tool treats the contract error as a permanent setup error instead of polling to timeout.

## Trust model

External mode routes every workspace's opens into your one shared, signed-in OS browser session — that reach is the point. It's global and opt-in, and the Settings copy says so plainly.

## Testing

- Unit: main RPC fork + fail-closed + tabs contract + store persistence; PlaywrightEngine external short-circuit; renderer delegation guard; i18n contract. **577 affected tests green, tsc clean.**
- Live dogfood (isolated daemon + real RPC pipe, userData pre-seeded to `external`): `cdp.info` reports `external`, `browser.open` delegates, **wmux process tree stays at 6 — zero webview guest spawned**, `browser.screenshot` returns the verbatim contract error, `browser.tabs new` returns the external envelope. All checks passed.

Reviewed via 3-way independent review (Claude + Codex + GLM-5.2); 3 P1s and a hydration-race P2 they surfaced are fixed in this branch.

## Summary by CodeRabbit

* **New Features**
  * Added a Terminal setting to choose between the built-in browser pane and the operating system’s default browser.
  * External mode opens pages without embedding Chromium and reports delegated opens clearly.
  * Browser navigation and new-tab actions support external browsing, while advanced automation shows a clear unsupported message when unavailable.
  * The selected backend is saved and restored between sessions.
* **Bug Fixes**
  * Prevented external mode from incorrectly targeting embedded browser panes.
  * Improved error handling for unsupported automation and failed external opens.
* **Documentation**
  * Added release notes and localized settings descriptions.


openwong2kim:
Thanks — addressed both:

**Hydration race (Major):** fixed in 45339a70. The renderer mirror is now seeded synchronously from main's persisted value at store-module load (`ipcRenderer.sendSync('browser:get-backend-sync')`), so it's correct before the first render. That closes the window entirely — a session-restored browser leaf or an early open can no longer spawn an embedded webview while 'external' is persisted. The async `getBackend()` path stays as a fallback/refresh. Agent opens were never affected (main reads its own store synchronously).

**Korean strings in ko.ts (Minor):** respectfully skipping — `ko.ts` is the Korean *locale* file, so Korean translation values are correct there by definition (`en.ts` is the SSOT; `ko.ts` mirrors it in Korean, same as the sibling `browserLightweight`/`browserDiscard` keys). The repo's English-only rule covers commits, comments, docs, and PR text — not localization values.

PR #599, opened 2026-07-24
wmux web: survive a daemon restart, and keep the phone's token working

### LINKED ISSUE #596: wmux web does not survive a daemon restart — phone access dies silently and can only be restored at the desktop
## What happens

`wmux web` serves live panes to a phone browser. If the daemon restarts — a crash, a
reboot, or the one-click "check for updates" flow that shipped in 3.33.0 — the web
server does not come back. The port stops answering, the phone gets nothing, and
nothing on the phone indicates anything changed. Recovery requires a human at the
desktop to run `wmux web` again or click the titlebar control.

Sessions survive a daemon restart by design. The server that serves them does not.

## Why it matters

The 3.33.0 note for `wmux web` says it "keeps working even after you close the desktop
app — the agents are still supervised by the daemon." That holds for closing the GUI.
It does not hold for restarting the daemon, and 3.33.0 also shipped a manual update
path that restarts the app into the installer.

The situation where phone access earns its keep is the one where you are away from the
desktop. That is exactly the situation where you cannot fix this.

## Repro

Isolated daemon (`WMUX_DATA_SUFFIX=-webrestart`), so no real session data is touched.
Start the daemon, start `wmux web`, confirm it serves, kill the daemon, start a fresh
one against the same data directory:

```
STEP 1 — start daemon #1
  ready. status: {"running":false}

STEP 2 — start wmux web
  started. running = true  port = 7699
  HTTP probe: {"status":200}

STEP 3 — kill daemon #1 (simulates crash / reboot / upgrade restart)
  HTTP probe while down: {"error":"ECONNREFUSED"}

STEP 4 — start daemon #2 (same data suffix, like a normal restart)
  daemon #2 ready. web status: {"running":false}
  HTTP probe after restart: {"error":"ECONNREFUSED"}
```

## Where

- `src/daemon/index.ts:53` — "Instantiated lazily in `registerRpcHandlers`; nothing
  listens until a `daemon.web.start` RPC arrives."
- Only two callers of `daemon.web.start`, and both need a human at the machine:
  `src/cli/commands/web.ts:64` (the CLI) and `src/main/ipc/handlers/web.handler.ts:79`
  (the titlebar popover).
- `src/daemon/index.ts:3224` stops the server on shutdown. Nothing starts it on boot.
- The enabled state is not persisted anywhere.

## This is not the lazy-init default

"Nothing listens until asked" is intentional, documented in three places, and should
stay. The gap is narrower: an operator's explicit "yes, serve this" is forgotten across
a restart the operator never asked for.

## Adjacent, worth deciding together

The web token is re-minted on every start (`WebTerminalServer.ts:144`), and the pairing
code with it. So even a server that did come back would hand a previously paired phone
a token that no longer works. Restoring the listener without addressing credential
lifetime would produce a server that is up but unreachable from the device that was
using it — arguably worse than the current honest failure. Same applies to the
`tailscale serve` path, where adding `--allow-host` requires a restart that rotates the
token.

## Environment

- wmux 3.33.0
- Windows 11
- Reproduced against `dist/daemon-bundle/index.js` with an isolated data directory

openwong2kim:
Closes #596

`wmux web` serves live panes to a phone. Sessions already survive a daemon restart; the server that serves them did not. A crash, a reboot, or the one-click update flow that shipped in 3.33.0 left the port dead, the phone showing nothing, and the only fix was to walk back to the desktop — which is the one situation phone access exists for.

Two halves, because doing either alone is worse than the honest failure it replaces.

### Restore the listener

Nothing recorded that the operator had turned it on. `daemon.web.start` was the only path to a listening server and the boot sequence had no restore step.

A new `web-state.json` (0600, next to `daemon-auth-token`) records the exact options an explicit start used, and the daemon replays it at boot.

- **Its own file, not `config.json`.** It carries the bearer token, and `config.json` is hand-edited and routinely pasted into bug reports. A malformed web state also can't take config.json's core structure down with it.
- **Kicked off before the control pipe starts listening**, gated behind a `webRestore` promise that every web RPC and `shutdown()` awaits. Placed after, an operator `daemon.web.start` arriving during boot would interleave with the restore's own bind, or the restore would stomp the fresh options with the persisted ones.
- **No state file → nothing listens.** The lazy-init default is untouched for anyone who never ran `wmux web`.

### Carry the token

`start()` minted a fresh UUID every time. A restored server would have 401'd the very device that was using it — a server that is up but unreachable from the paired phone, which the issue rightly calls worse than the current failure.

The token now survives restarts *and* operator re-starts, which also fixes the smaller version of this: `wmux web --allow-host <magicdns>` no longer locks out a phone that already paired.

The reuse seam is `WebTerminalServer.start({ token })`, set only by the daemon from its own 0600 file — **never from RPC params**, so no pipe client can choose the token a browser is handed. `--stop` revokes it; a new `--new-token` rotates on demand.

### Phone side

`EventSource` reports a revoked token and a restarting daemon identically, so a dead session sat on "reconnecting…" forever. A rate-limited probe of one authenticated route now tells them apart and routes a 401 to the pairing screen.

### Security posture

Restoring replays the operator's full choice, `--expose` and `--allow-input` included. A half-restore would bring back a server that isn't doing what they asked, and loopback-only defeats the whole point. A restore that puts a writable terminal back on all interfaces logs at `warn` — they asked for it, but "it came back on its own" should be findable without knowing to look for it.

A corrupt or unreadable state file degrades to off rather than blocking daemon boot, and `enabled` is fail-closed twice: it must be literally `true`, and a record with no token counts as disabled.

### Verification

The issue's own repro, against the real bundled daemon:

```
STEP 4 — start daemon #2 against the SAME data dir
  web status: {"running":true,"port":7699,...,"token":"f6904405-…"}
  HTTP probe after restart: {"status":200}
STEP 5 — the phone's saved token
  token identical to the one the phone has? true
  old token vs /api/sessions: {"status":200}
```

Tests:

- `webRestart.runtime.test.ts` — three cases against the real bundled daemon: SIGKILL, then a fresh daemon on the same data dir. Restore + token continuity, stop-is-remembered + revocation, and `--new-token` rotation.
- `webStateStore.runtime.test.ts` — 10 cases on coercion, fail-closed, prototype-pollution, atomic write.
- Three new `WebTerminalServer` cases for the token seam.

**The restart tests were checked against a pre-fix bundle and fail there** (`expected false to be true`), so they discriminate rather than just pass.

Full suite: 7775 passing. Two files (`main/git/mergeSession`, `ipc/diff.handler`) fail only under full parallelism with Windows temp-dir `EBUSY` and pass 42/42 in isolation — unrelated paths, pre-existing flake.

### Not in scope

The GUI popover has no `--new-token` equivalent; its Stop button revokes, and the CLI covers on-demand rotation. Say the word if it should be there too.

## Summary by CodeRabbit

- **New Features**
  - `wmux web` now remembers its serve decision, listener settings, and access token across daemon restarts/crashes, so paired devices can reconnect automatically.
  - Added `--allow-host` and `--new-token`, and updated status/stop behavior messaging accordingly.
  - Pairing recovery is improved: pairing codes rotate even when tokens persist.

- **Bug Fixes**
  - Stopping `wmux web` now persistently disables access and revokes the current token (including after restart).
  - Improved handling for corrupt/invalid/empty saved state: web auth now fails safely (disabled) instead of partially working.


openwong2kim:
[wmux-hermes] triage — **P1 / review: approve with one thing to confirm**

## What this PR changes (from the diff)
This closes #596 by making `wmux web` survive a daemon restart, which it accomplishes in two coordinated halves. The first is `web-state.json`, a new 0600 sibling of `daemon-auth-token` that records the operator's exact start options (port, host, `allowInput`, `allowedHosts`, token) and is replayed by `restoreWebServer()` at boot, *before* the control pipe starts listening. The second is token continuity: `WebTerminalServer.start({ token })` now reuses a caller-supplied token instead of always minting a fresh `randomUUID`, so a paired phone reconnects on its own. The token seam is deliberately daemon-internal only — it comes from the 0600 state file, never from RPC params, so no pipe client can choose the token a browser receives. On the phone side, `diagnoseStreamError()` probes one authenticated route to distinguish a revoked token (-> pairing screen) from a restarting daemon (-> keep retrying), since `EventSource` reports both identically.

## Security and correctness — this is the part that matters
The threat model here is real and the PR handles it with the right instincts. Persisting the operator's full choice means a restore replays `--expose` and `--allow-input` unconditionally — bringing a writable terminal back on all interfaces with nobody at the desktop. The code logs that case at `warn` (good), and the `restoreWebServer` comment explains why a half-restore would be worse than the honest failure. The fail-closed discipline in `coerceWebState` is exactly right: `enabled` must be literally `true`, an enabled record with no token counts as disabled, corrupt JSON degrades to off, and the `__proto__`/`constructor`/`prototype` reviver guard prevents prototype pollution from a hostile state file. The atomic write (tmp + rename) mirrors `saveConfig`, and a write failure returns `false` rather than failing the start RPC.

## The serialization design is the strongest part
The `webRestore` promise, awaited by every web RPC and by the graceful teardown path, is the load-bearing invariant that makes the whole thing race-free. Booting the restore *before* the control pipe listens means an operator `daemon.web.start` arriving during boot can never interleave with the restore's own bind, and the restore can never stomp fresh operator options with persisted ones. The comment at `main()` step 7a documents this explicitly. `--stop` revokes the token (clears the state file), `--new-token` rotates on demand, and a plain re-start (e.g. adding `--allow-host`) deliberately carries the token forward so it doesn't lock out an already-paired phone — that last point fixes a smaller, separate version of the issue.

## Feedback and things to confirm
1. **The runtime test body has a redaction artifact.** In `webRestart.runtime.test.ts` the `probe()` helper contains `Authorization: *** ${token}` — the `Bearer ` prefix appears to have been stripped/redacted, leaving a malformed header value (`*** <token>`). If this is how it is on disk it would 401 every probe; if it's a display artifact in the diff then no action needed, but please confirm the committed code reads `Authorization: \`Bearer ${token}\``.
2. **`persistWebState` re-persist on restore.** After a restore, `persistWebState(info, state.allowedHosts)` re-writes the file with `info.port ?? 7681`. The comment notes the bind only differs from requested when port `0` was asked, which "never happens here" — correct, since `coerceWebState` clamps to 7681 default. Fine, but worth a one-line assertion or comment that `state.port` is never `0` by the time it reaches `restoreWebServer`, so the re-persist is genuinely a no-op mirror.
3. **`diagnoseStreamError` rate limit is global, not per-tile.** `AUTH_PROBE_MS` is module-scoped across all four split tiles, which is the right call (one probe in flight total), and the comment explains it. Just confirming this is intentional — it is, and it correctly avoids a 4x probe burst on a full reconnect.
4. **GUI popover has no `--new-token` equivalent.** The PR body flags this explicitly as out of scope; the popover's Stop button revokes and the CLI covers on-demand rotation. Reasonable scope boundary, just noting it's an acknowledged gap rather than an oversight.

## Risk assessment
Low. The change is additive (new state file, new optional `token` param defaulting to fresh-UUID behavior), the lazy-init default is explicitly preserved for anyone who never ran `wmux web`, and the regression tests are checked against a pre-fix bundle and fail there. The 7775-passing suite with two unrelated pre-existing `EBUSY` flakes under full parallelism is consistent with a clean change.

## Recommendation
**Approve.** This is a well-scoped fix for a P1 issue (#596) with strong security posture, correct race-avoidance design, and discriminating tests. Confirm item 1 (the `Authorization` header in the runtime test) before merge — if that's a real redaction rather than a diff artifact it needs fixing; everything else is optional polish.

PR #600, opened 2026-07-24
wmux web: show what notifications actually say, and stop losing attention events on reconnect

### LINKED ISSUE #597: wmux web: every OSC notification on the phone renders as the literal string "Notification" and drops the message body
## What happens

Every OSC notification shown on the `wmux web` phone surface renders as the literal
string **"Notification"** with an empty subtitle. The agent's actual message is dropped.

An agent emitting `ESC ] 9 ; Build finished, 3 tests failed BEL` produces a banner that
says "Notification" and nothing else. The desktop toast for the same event is correct —
only the phone loses the text.

## Cause

The frontend reads a field the pipeline never sends.

`src/daemon/web/frontend/app.js:700`:

```js
title = data.message || 'Notification';
sub = '';
```

`src/main/pty/oscNotification.ts` returns `{ source, title, body }` — there is no
`message` field. OSC 9 gives `{source:'osc9', title:null, body}`, OSC 777 gives
`{source:'osc777', title, body}`. `DaemonPTYBridge.ts:184` forwards that shape unchanged
plus `ts`.

So `data.message` is always `undefined`, the fallback fires every time, and `body` — the
part the user actually wants — is never read at all.

## Why it wasn't caught

The unit test fabricates the payload instead of using the production shape.
`src/daemon/web/__tests__/WebTerminalServer.test.ts:262-264`:

```js
(sessionManager as unknown as EventEmitter).emit('session:notification', {
  sessionId: 's3',
  event: { message: 'done', ts: 123 },
});
```

That `{message}` shape does not exist anywhere in the product. The test asserts the event
reaches the wire, which it does — the bug is one layer further on, in rendering.

## Suggested fix

Render `title`/`body` as the parser produces them, with `body` as the required part and
`title` as the optional heading (OSC 9 has no title; OSC 777 may). Then update the test
to emit the real `{source, title, body, ts}` shape so the fake stops diverging from
production.

Worth checking whether anything else consumes `session:notification` off the web wire
with the same assumption.

## Impact

This blocks measuring anything about the notification channel on mobile — you cannot
assess signal quality on a surface that throws the content away. Relevant to any future
work that routes attention events to a phone.

## Environment

- wmux 3.33.0
- Found by reading the source, not reproduced on a device (the phone surface has no
  automated coverage; `app.js` is ~1200 lines with no tests).

### LINKED ISSUE #598: wmux web: attention events fired during an SSE reconnect are lost permanently (terminal bytes replay, attention does not)
## What happens

On the `wmux web` phone surface, terminal output survives a dropped connection but
**attention events do not**. If a pane raises an approval request or emits a notification
while the browser's SSE stream is reconnecting, that event is gone. The terminal catches
up on reconnect and looks completely normal, so there is no sign anything was missed.

On a phone this is not an edge case. `EventSource` reconnects on every network blip,
every backgrounded tab, every transition between wifi and cellular.

## Why bytes survive and attention doesn't

They take different paths on reconnect.

**Terminal bytes** are re-sent in full: `handleStream` writes a `snapshot` event from
`ringBuffer.readAll()` on every stream open (`WebTerminalServer.ts:444-445`). Whatever
you missed is in the snapshot.

**Attention events** are pure fan-out with no history: `broadcastEvent` iterates the
currently-connected clients and writes to each (`WebTerminalServer.ts:541-552`). A client
that is not connected at that instant is simply not in the set. There is no buffer, no
cursor, and nothing re-sends on reconnect.

`writeSse` (`WebTerminalServer.ts:715-717`) emits only `event:` and `data:` lines — no
`id:` field — and the request handler never reads `Last-Event-ID`. So even a client that
wanted to resume from where it left off has nothing to resume from.

## Related, smaller

`broadcastEvent` iterates `this.clients`, which is populated only in `handleStream`
(`WebTerminalServer.ts:484`), and `handleStream` 404s without a valid `?session=`.
So attention delivery requires at least one open pane stream.

In practice the app auto-selects the first pane on load (`app.js:661`, `:1099`), so this
mostly does not bite today. It does mean there is no path to a fleet-level surface that
receives attention events without subscribing to some pane's byte stream — worth knowing
before anything is built on top.

## Suggested direction

Not prescribing a fix, but the shape that would resolve it:

- Give attention events monotonic `id:` values and honour `Last-Event-ID` with a bounded
  replay window, **or**
- Keep a short server-side list of recent/unacknowledged attention events and expose
  `GET /api/attention` (or similar) that the client fetches on every connect, the same way
  it gets a snapshot for bytes.

The second is probably simpler and composes better with anything that later needs
"what is currently waiting on me."

## Impact

Any future work that routes attention to a phone depends on this. A notification channel
that silently drops events during exactly the connectivity conditions phones live in is
not something you can build a "the agent needs you" workflow on.

Related: #596 (the web server does not come back after a daemon restart) — same theme,
different layer. Together they mean an away-from-desk user can miss events both because
the server is gone and because the stream reconnected.

## Environment

- wmux 3.33.0
- Found by reading the source during a design/architecture review; not reproduced on a
  device. The phone frontend has no automated test coverage.

openwong2kim:
Closes #597
Closes #598

Two defects in the phone surface's attention channel, fixed together because they share a path and neither is verifiable without the other — a replay that delivers events the UI then renders as the bare word "Notification" proves nothing.

### #597 — the body was dropped at the last hop

`app.js` built the banner from `data.message`. The daemon has never sent that field: the parsed OSC notification is `{source, title, body, ts}`, with `title: null` for OSC 9 (the common case). So `data.message` was always undefined, the fallback fired every time, and `body` — the part the user wanted — was never read.

It shipped green because the unit-test fake emitted `{message, ts}`, a shape that exists nowhere in the product. The fake now emits the real wire shape, and the test asserts **all four fields** survive to the wire verbatim — not just the body (the push payload and any future client need `source`/`ts` too).

Formatting moved into a pure `formatAttention` (`frontend/attentionFormat.js`, inlined by the build script ahead of `app.js`), so the rule that OSC 9's body-only notification becomes the headline is now unit-tested without a DOM — `app.js` had zero test coverage, which is how a one-line field mismatch survived.

### #598 — fan-out with nobody listening is a drop

`broadcastEvent` wrote to whatever SSE clients were connected at that instant. Terminal bytes replay on reconnect (ring-buffer snapshot); attention events had no buffer, no ids, no `Last-Event-ID`. On a phone — where EventSource reconnects on every network blip, backgrounded tab, and wifi↔cellular hop — an approval raised at the wrong moment was gone, and the terminal caught up looking perfectly normal.

Now:

- Every attention event is recorded in a bounded in-memory log (100 entries / 30 min TTL) with a monotonic id, keyed by a **per-process epoch** minted in the constructor. Not the OS boot id — two daemon restarts inside one boot would share that, and a stale cursor would silently swallow the new id space.
- New `GET /api/events`, one route, two shapes: an SSE stream (`id:` lines, `Last-Event-ID` resume, replay-then-live) for EventSource, and a JSON backlog (`?since=epoch:id` → `{epoch, headId, reset, events}`) for fetch-on-connect. A cursor from a foreign epoch gets `reset` plus the full window, so a client always knows when it must resync rather than guessing.
- Auth is exactly the existing rule set: Bearer for JSON, the `?token=` query exception only for the SSE variant (EventSource cannot set headers), same as `/api/stream`. Same security headers, same Host allowlist.
- The frontend opens the stream once per page (not per pane — a pane switch used to construct a fresh EventSource, which would have discarded `Last-Event-ID`), dedups on `epoch:id`, persists the cursor for reloads, and collapses a reset replay into one "N events while you were away" summary plus fleet-chip marks instead of a banner storm.
- Attention keeps riding the pane streams for one release, so a cached PWA frontend against a new daemon misses nothing; new frontends dedup the double delivery by id. This also decouples attention from having a pane stream open at all — `/api/events` needs no `?session=`, which the approvals work (#roadmap M2) sits on top of.

### Proof the tests bite

With the server fix stashed and the tests kept, 11 fail — including the two regression cases that define these bugs:

```
FAIL ★ delivers an attention event raised while NO client was connected  (expected 404 to be 200)
FAIL tees session:critical and session:notification …                    (expected … to match /"id":\d+/)
```

Full run with the fix: `src/daemon/web/` 40/40, `attentionFormat` 74 lines of cases (OSC 9 body-only, OSC 777 title+body, garbage payloads), suite-wide 7795 passing. The two files that fail only under full parallelism (`ipc/diff.handler`, `main/git/mergeSession`) are the known Windows temp-dir EBUSY flake and pass in isolation.

### Notes

- TTL eviction is tested with an injected clock (`now()` through deps), not sleeps.
- No RPC surface change — `/api/events` is HTTP-only, so no API-reference regen.
- Related: #596 / #599 persist the web server and its token across daemon restarts. Together with this, a phone that was disconnected through a daemon restart reconnects on its own **and** learns what it missed — the epoch reset marker is what makes that honest.

## Summary by CodeRabbit

- **Bug Fixes**
  - Desktop notifications now display the real notification content (title/body/timestamp) rather than a generic label.
  - Attention/approval events are no longer lost across reconnects or when no browser is actively watching.
  - Reconnecting or revisiting restores missed alerts without duplicates, using cursor-based replay.
  - Missed alerts are summarized after a return/idle period, while live critical/notification banners continue to show normally.


openwong2kim:
[wmux-hermes] triage — **P2 / review: approve**

## What this PR changes (from the diff)
Closes #597 and #598 together because neither bug is verifiable in isolation — a replay that delivers events the UI then renders as the bare word "Notification" proves nothing. The changeset spans four areas:

1. **`WebTerminalServer.ts` — durable attention log.** `broadcastEvent()` now records every event into a bounded in-memory log (`ATTENTION_CAP=100`, `ATTENTION_TTL_MS=30min`) *before* fan-out, so an event raised while nobody is connected lands in the log and is replayed on next connect. Each entry gets a monotonic `id` within a per-instance `attentionEpoch` (minted once in the constructor, deliberately not the OS boot id so two daemon restarts in one boot don't share an id space). Two new routes share `/api/events`: an SSE variant (`id:` lines, `Last-Event-ID` resume, replay-then-live) for `EventSource`, and a JSON backlog (`?since=epoch:id`) for fetch. A cursor from a foreign epoch yields `reset: true` plus the full window, so a client always knows when it must resync. Auth mirrors `/api/stream`: `?token=` only for the SSE variant (EventSource can't set headers), Bearer for everything else.
2. **`attentionFormat.js` — pure formatter.** Fixes #597 by rendering `title`/`body` from the real `{source, title, body}` payload instead of a nonexistent `data.message`. Extracted as a dependency-free module (inlined ahead of `app.js` by `build-daemon-web.mjs`) precisely so the rule is unit-testable without a DOM, since `app.js` has no test harness.
3. **`app.js` — page-lifetime attention stream + dedup.** Opens `/api/events` once per page (not per pane — a pane switch would otherwise discard `Last-Event-ID`), dedups on `epoch:id` via a FIFO-bounded seen set (cap 200 > log cap 100, so the whole window is covered), persists the cursor in `sessionStorage` for reloads, and collapses a `reset` replay into a single "N events while you were away" summary plus fleet-chip marks instead of a banner storm.
4. **Tests.** 11 new `WebTerminalServer` cases (attention log survives no-listener drop, foreign-epoch resync, auth gating, cap+TTL eviction with injected clock, SSE id/reset/replay ordering, `Last-Event-ID` resume, log survives `stop()`) plus a dedicated `attentionFormat.test.ts` (74 lines covering OSC 9 body-only, OSC 777 title+body, empty/garbage payloads). The PR claims the regression tests fail against a pre-fix bundle — a good signal they discriminate.

## Potential risks / edge cases
- **Replay-end heuristic is time-based.** `armReplayEnd()` treats a 200ms quiet window as "backlog done." If a slow/throttled connection spaces backlog chunks >200ms apart, a single logical replay can be split into multiple "N events while you were away" summaries. Low likelihood (the full 100-entry window is small), but worth a one-line note if you've seen it in manual phone testing.
- **Double-delivery for one release is intentional and handled.** The old pane-stream tee and the new `/api/events` both carry attention; the `epoch:id` dedup makes the overlap harmless, and the tee is explicitly time-boxed ("remove after one release"). Back-compat reasoning for a cached PWA frontend is sound.
- **`seenAttention` cap (200) vs log cap (100):** correctly larger, so a full replay can't slip a duplicate through. Good.
- **No unit coverage for the frontend dedup/coalescing logic in `app.js`** (DOM-bound, no harness). The testable rule (`attentionFormat.js`) is covered; the untested part is the replay-summary state machine. Acceptable given the PR's scope, but the `beginReplay`/`armReplayEnd`/`endReplay` flow is the most logic-dense untested code in the diff — worth a manual smoke test of the reset path on a real phone.
- **Epoch/log surviving `stop()` is correct** (a web-server restart is not an event-loss boundary; the daemon kept running). The `stop()` path properly detaches `eventClients` and clears the set without touching the log.

## Specific feedback
- Naming is clear throughout (`ATTENTION_CAP`, `attentionWireBody`, `headId`, `parseCursor` splitting on `lastIndexOf(':')` to future-proof the epoch format).
- The injected `now()` clock seam for TTL testing is the right call — no 30-minute sleeps.
- The `MERGE FOLLOW-UP` note on `app.js:openAttentionStream` flagging that #596's `diagnoseStreamError` 401-probe should be wired in once both branches land is exactly the kind of cross-branch note that prevents a silent regression — good.
- The test-fake correction (emitting the real `{source, title, body, ts}` shape instead of a nonexistent `{message}`) is the root-cause fix for how #597 shipped green; asserting all four fields survive to the wire closes that gap.

## Recommendation
**Approve.** Both bugs are well-understood, the design (per-process epoch, bounded replay, fetch-on-connect + SSE resume, back-compat tee) is sound, and the test coverage genuinely discriminates. The 200ms replay heuristic and untested frontend coalescing are observations rather than blockers — neither is worth blocking merge. P2: significant fixes to a core remote-access feature, but not a P1 (no crash/data-loss on the desktop app, no security boundary crossed — the auth model is unchanged from `/api/stream`).

Issue #570, opened 2026-07-23
Perf gate flakes on ram.panes8.workingSetBytes — working-set + software-render runner + PPID tree contamination


openwong2kim:
## Summary

The Perf gate (`ram.panes8.workingSetBytes`) flakes between pass and fail on `main` with no code change to explain it. On 2026-07-23 alone `main` recorded Perf success ×6 / failure ×3, and PR #569 (a tests-only change touching two `.test.ts` files) tripped it with `ram.panes8.workingSetBytes 1695.1 MiB → 4442.8 MiB (+162.1%)` — a regression the diff cannot possibly cause.

This isn't a real regression. It's a measurement-design problem in how `measureRam()` samples memory, amplified by the CI runner environment. Filing so the gate stops producing false reds.

## Why it swings (root cause)

The gated metric sums `WorkingSetSize` over a PPID-derived process tree, sampled once on a shared `windows-latest` runner (`scripts/perf-bench.mjs:968` `measureRam`, gate config in `scripts/perf-compare.mjs:73`). Three structural factors each add volatility:

1. **`WorkingSetSize` tracks system memory pressure, not demand.** Working set is the pages currently resident in physical RAM. Windows only trims it under pressure and lets it balloon when RAM is free, so the *same* app reports 1.7 GB or 4.4 GB depending on unrelated runner load and the OS's lazy trim decisions. `commitBytes` (PageFileUsage) is the stable demand metric; the gate uses the volatile one.

2. **No real GPU on the runner → software rendering.** `windows-latest` has no GPU, so the 8 WebGL-backed xterm panes fall back to SwiftShader software rasterization. The GPU-process working set for software-rendered WebGL is large and unstable run-to-run.

3. **The PPID tree walk is contamination-prone on Windows** (`perf-bench.mjs:980`, `byParent.get(pid)` BFS). Windows recycles PIDs aggressively and does not reparent orphans to a stable ancestor. If a prior scenario's Electron helper isn't fully reaped before the `panes8` sample, or a recycled PID coincidentally matches a tree PID, the walk pulls in an extra subtree. The observed ≈2.6× jump (1.7 → 4.4 GB) matches a whole extra set of Electron helper/renderer processes being double-counted — plain jitter (±10%) does not produce binary multiplier jumps like this.

Secondary: `measureRam()` has no forced GC or settle barrier before sampling, so it snapshots 8 renderers + GPU + daemon at an arbitrary GC moment.

[...]

PR #606, opened 2026-07-25
fix(bench): publish the perf trend where CI can actually write it

### LINKED ISSUE #602: Perf trend history has recorded nothing since 2026-06-12 — the append commit is rejected by branch protection, skipped on red runs, and omits hiddenFlood entirely
## Summary

`bench/history.ndjson` — the main-branch perf trend documented in `bench/README.md` as "one NDJSON line per push, appended by CI" — has recorded **nothing since 2026-06-12**. The last line is `2026-06-12T22:56:49Z` (`e058264`) and the last `chore(bench): append perf history [perf-history]` commit is from the same day.

The workflow step still runs on every push. It fails every time, warns non-fatally, and nobody sees it. Three independent defects stack on top of each other, and the net effect is that the project has had no perf trend data for ~6 weeks — which is exactly the data #570 needs to decide whether a baseline is still descriptive.

## Defect 1 — the history commit cannot be pushed to a protected `main`

From the latest green `main` run ([30122487773](https://github.com/openwong2kim/wmux/actions/runs/30122487773), `2bfce3ce`):

```
[main b93ce75] chore(bench): append perf history [perf-history]
 ! [remote rejected] main -> main (protected branch hook declined)
 ! [remote rejected] main -> main (protected branch hook declined)
 ! [remote rejected] main -> main (protected branch hook declined)
##[warning]perf history push failed after 3 attempts (non-fatal)
```

The `github-actions[bot]` identity the step configures cannot push to a protected `main`, so all three retries are rejected for the same reason — the retry loop exists for a *racing* push, and a permission rejection is not a race, so retrying cannot help. I checked every `main` Perf run in the current 40-run window (2026-07-24 08:10Z → 2026-07-25 01:05Z): **11 of 11 green runs were rejected this way.** I can't see when branch protection was enabled, but the 2026-06-12 cutoff and the rejection are consistent with each other.

The step is deliberately non-fatal (`exit 0`, `::warning::`) so a lost trend line can't fail the gate — that part is right. What makes it invisible is that a warning on an otherwise-green run is nothing anyone reads, and there is no other signal that the trend is dead.

## Defect 2 — the history step is skipped on exactly the runs worth recording

```yaml
- name: Commit perf history
  if: github.event_name == 'push'
```

No `always()`. The gate (`Compare against baseline`) exits 1 on a failure, so the job stops there and this step never runs.

`perf-compare.mjs` is not the problem — it appends the line *before* returning its exit code ("Best-effort; never gates"), so on a red run the trend line is correctly written to the working copy. It is then thrown away with the runner, because nothing ever commits it. Neither red `main` run in the same window ([30081357631](https://github.com/openwong2kim/wmux/actions/runs/30081357631), [30118238142](https://github.com/openwong2kim/wmux/actions/runs/30118238142)) reached the step at all.

So even with Defect 1 fixed, the resulting dataset would be **biased by construction**: it would contain only runs that passed the gate, and systematically exclude the outliers. For a heavy-tailed metric that is the difference between seeing the noise profile and not seeing it. Since the append already happens, this is a one-line `if:` fix.

## Defect 3 — `historyLine()` never recorded the `hiddenFlood` family

`scripts/perf-compare.mjs::historyLine()` writes these trend fields:

`coldFirstPtyDataMs`, `echoP95Ms`, `frameP95Ms`, `frame8P95Ms`, `ramIdleBytes`, `ram8Bytes`, `frameBudgetP95Ms_N4/N8/N16`, `imePass`, `webglContextLossPass`.

There are **15 gated metrics**, and the four `hiddenFlood` gates (`N4/N8 × echoMs.p95/frameDeltaMs.p95`) are not among the recorded fields. `frameBudget` was added to `historyLine` as "W2 trend fields" when its gates landed; `hiddenFlood` gates were added later (baseline blessed 2026-07-10) without the matching trend fields.

That is the *noisiest* gated family — its own gate comment in `perf-compare.mjs` calls `echoMs.p95` "the noisiest gated metric", observed spread 2.3× — and it is the one family with no trend record. It is also the family currently producing false reds (#570).

## Impact

- No trend data since 2026-06-12, silently. `bench/README.md` still documents the trend as live.
- `baseline-ci.json` was blessed on 2026-07-10 from a single run and there is no way to check it against a distribution — which is directly why the `hiddenFlood` baselines are now ~4× the metric's central tendency (detail in #570).
- The bias in Defect 2 means the dataset would stay misleading even after Defect 1 is fixed.

## Options

Defects 2 and 3 are small and uncontroversial (`if: always()` on the step; add four fields to `historyLine`). Defect 1 is a repo-permissions decision and yours to make:

1. **Push the trend to an unprotected branch** (e.g. `bench-history`) instead of `main`. No secret, no protection change, and it drops the `paths-ignore` dance in `perf.yml` since the trend commit never lands on `main`. Costs: trend lives off `main`, so reading it is one extra fetch.
2. **Give the workflow a token that may bypass protection** (GitHub App token or PAT). Keeps the file on `main`; costs a credential with write access to a protected branch.
3. **Upload the line as a workflow artifact** and aggregate on demand. No permissions at all; costs a real accumulation story (artifacts expire, and there is no single readable file).
4. Whichever route: **make the failure loud** — `::error::` or a job-summary line — so a dead trend cannot go unnoticed for six weeks again.

I'd suggest (1) plus (4), with 2 and 3 folded in, as one small PR.

Minor and adjacent, mentioned only so it isn't lost: the gate table in `bench/README.md` still lists 6 metrics and predates both the `frameBudget` and `hiddenFlood` gates, so it documents 6 of the 15 gates that actually run.

Happy to open the PR for any subset of this — say which of the four options you want for Defect 1 and I'll build to it.

snowyukitty:
Built to your steer on #602: option **1 + 4** for the push, defects 2 and 3 folded in, and the README gate-table drift while I was in there.

## What changed

- **The trend is published to the unprotected `bench-history` branch** as `history.ndjson`, instead of committed to `main`. No new credentials, and `paths-ignore` for the history file is gone from `perf.yml` since no CI commit lands on `main` any more.
- **A failed publish is loud**: an `::error::` annotation plus a `> [!CAUTION]` job-summary line. Still `exit 0` — a lost trend line must not fail the gate.
- **Red runs are recorded** (defect 2). `perf-compare.mjs` already wrote the line before returning its exit code; only the commit step was skipped.
- **Every gated metric has a trend field** (defect 3), including the four `hiddenFlood` ones.
- `bench/README.md`: the gate table listed 6 of the 15 gates that actually run; the file inventory and CI section now describe where the trend lives and why.

## Five decisions worth your eye

**1. The trend fields are derived from the gate tables rather than extended by four.** Every pre-existing history field name was already exactly a gate `key`, reading exactly that gate's `path` — `ram8Bytes`, `frameBudgetP95Ms_N4`, `imePass`, all eleven of them. The list was a hand-made copy of something that already existed, which is why it could drift and did. So `historyLine()` now spreads `GATES` and `BOOL_GATES` directly. The emitted field names and values are unchanged, the four `hiddenFlood` fields appear with the names they would have had, and a gate can no longer be added without its trend field — because they are now the same list. The result is shorter than what it replaces.

**2. The commit is built with git plumbing, not a checkout.** `hash-object` → `update-index`/`write-tree` → `commit-tree` → `push origin <sha>:refs/heads/bench-history`. The working tree holds the packaged Electron build, and swapping branches in it to append one line would be slow and needlessly destructive. It also makes the race trivial: the new commit's parent *is* the tip just fetched, so the push is a fast-forward by construction, a concurrent `main` push rejects it, and the retry re-reads the new tip and re-appends onto it.

**3. `!cancelled()` rather than the `always()` you suggested.** Same effect for the red-run case you asked for, minus one: a *cancelled* run may have been interrupted mid-measurement, and that is not a sample worth publishing. Say the word if you'd rather have plain `always()`.

**4. `bench/history.ndjson` is deleted from `main`.** With the trend on the branch, a file in `bench/` that the README used to call "appended by CI" is exactly the kind of thing that looks alive and isn't — the trap this whole issue is about. The 11 pre-break lines stay readable at `c3e18eda:bench/history.ndjson`. Easy to veto: say so and I'll restore it as a frozen archive with a header note instead.

**5. Reading the existing trend is fail-closed.** My first draft treated "could not read `history.ndjson` from the branch" as "no trend yet" and carried on, which would replace the entire series with the single new line. Since that is precisely the silent-data-loss shape this PR exists to remove, the step now refuses to publish, retries, and fails loudly instead. See below — I did not reason my way to that, a live run did.

## A second pass over my own step

Nothing below changes what the step does; it is what a hard re-read turned up.

- **A failed `fetch` lands on the bootstrap path too**, and the comment now carries the argument for why that is safe rather than leaving a reviewer to reconstruct it: it yields a parentless commit, pushing one of those at an existing branch is not a fast-forward, so the server rejects it and the next attempt re-reads. Nothing here ever forces — that is what makes "could not read the tip" structurally unable to erase it.
- **`[Console]::OutputEncoding` is pinned to UTF-8.** The existing trend is read back through PowerShell as text before being rewritten, and native-command output is otherwise decoded with the runner's code page. Every field written today is ASCII, so this changes nothing now; it is what stops a future non-ASCII one from being silently mangled by a step whose whole job is to round-trip a data series faithfully.
- **The bootstrap commit says what the branch is** — `chore(bench): start the perf trend — one line per push to main, see bench/README.md` — since a first commit reading "append perf history" is both untrue and a wasted chance to orient anyone who lands on a branch holding one bare `.ndjson`.
- Small: no sleep after the final attempt, the retry logs *why* it is retrying, and a successful publish logs the resulting trend length.

## Testing

The publish step is `push`-only, so **this PR's own Perf run cannot exercise it** — it first runs on merge, on the bootstrap path. So I ran it against a real remote instead: a shallow clone (matching `actions/checkout@v4` with `fetch-depth: 1`) pointed at my fork, driving the step against a throwaway branch.

To keep that honest, the runnable copy is **extracted from `perf.yml` itself** by a generator that parses the YAML and makes exactly one substitution — the branch name — then asserts the result differs from the shipped body on exactly one line. The race variant is the same file with an interloper push injected before the read. So what ran is the step you are reviewing, not a paraphrase of it.

| path | result |
| --- | --- |
| bootstrap (branch absent) | parentless commit, explanatory subject; tree is exactly one file at mode `100644` |
| append (branch present) | appended in order; ran twice more later, ending at 5 lines |
| race — an interloper pushed to the branch between our fetch and our push | attempt 1 rejected `(non-fast-forward)`, retry re-read the **new** tip and re-appended: every line present, in order, none lost |

Final file: 0 CR bytes, tree entry exactly `history.ndjson`. The step body also parses clean under `[Parser]::ParseFile` (both variants). Then the throwaway branch was deleted.

One thing the simulation does **not** cover: it pushed with my own credentials, not the workflow's `GITHUB_TOKEN`. That path is evidenced instead by the failure this PR fixes — the old step's push reached the server and was refused by *branch protection*, not by authentication, and `permissions: contents: write` (already declared) covers creating and updating an unprotected ref.

**That run caught two real defects in my own step**, neither of which was visible by reading it:

- `git mktree` reads its entry from stdin, and PowerShell terminates piped stdin with **CRLF** — so the tree was created with the filename `"history.ndjson\r"`. Everything "worked": the push succeeded, the branch looked right in the GitHub UI. It is now built via `update-index --cacheinfo`, which takes the path as an argument, so no stdin parsing can smuggle a line ending into a filename. (`hash-object` also gained `--no-filters`, so attribute-driven CRLF conversion can't touch the blob either.)
- Because of that, the *next* run's `cat-file` of the previous file failed — and the `$existing = @()` fallback quietly truncated the history to one line. That is decision 5 above.

Unit side, on `historyLine()`:

- 7 tests: coverage over every gate, exact field order, the pre-#602 names pinned (a gate-key rename would otherwise silently fork the series), the four `hiddenFlood` values, run identity, an absent scenario recorded as `null` rather than a missing key, and the boolean tri-state.
- **Proven both ways**: restoring the hand-written field list fails exactly 3 of them, with `no trend field for gate hiddenFlood.N4.echoMs.p95`. The other 4 pass in both directions by design — they pin behaviour that was already correct.
- `npx vitest run scripts/__tests__/` → 121 passed. `npx eslint` on both changed script files → clean. The workflow parses under `js-yaml`, and the step body parses clean under the PowerShell parser (I extracted it from the YAML and ran it through `[Parser]::ParseFile`). No `.ts` changed, so a type-check would not have covered any of this and I did not run one.

## What this does not fix

The six-week gap stays gone, as you said. The branch starts empty on the first push after merge, so the `hiddenFlood` baseline question on #570 only becomes answerable once a few weeks of lines have accumulated — at which point the four fields that were missing will finally be in them.

Closes #602

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored automated performance-trend recording after a period of missing or incomplete history.
  * Ensured trend records reliably include all configured gate metrics, including previously omitted `hiddenFlood` values.
  * Improved reporting for skipped checks and missing measurements so absent vs false is handled consistently.
* **Documentation**
  * Clarified where trend history lives and how CI publishes to the separate trend-history location, including how failures are surfaced.
* **Tests**
  * Added coverage to validate the generated trend-history record’s completeness, ordering, and boolean/nullable behavior.


openwong2kim:
[wmux-hermes]

This PR reworks how the CI performance trend is published, moving it off the protected `main` branch — where the `github-actions[bot]` append commit was being silently rejected on every push since 2026-06-12 — onto an unprotected `bench-history` branch, and assembling the trend commit with git plumbing (`hash-object` → `update-index`/`write-tree` → `commit-tree` → `push`) so the packaged build occupying the working tree is never disturbed. It touches `.github/workflows/perf.yml` (the publish step is rewritten end-to-end), `scripts/perf-compare.mjs`, `scripts/__tests__/perfCompare.test.mjs` (seven new tests for `historyLine`), `bench/README.md`, `CHANGELOG.md`, and deletes the now-redundant `bench/history.ndjson` from `main` (its 11 frozen lines remain readable in git history). The `historyLine()` output is derived directly from the `GATES` and `BOOL_GATES` tables rather than a hand-copied second list, which restores the four `hiddenFlood` trend fields that had been missing since their gates shipped, and adds a unit test that fails the moment a gate and its trend field drift apart again. The publish step now runs on `!cancelled()` — so red runs (the outliers a noise investigation actually needs) are recorded too, where the old plain `if:` skipped them — and a failed publish raises an `::error::` annotation plus a job-summary caution instead of a warning nobody reads, while remaining non-fatal so a lost line cannot break the perf gate itself. This resolves the three defects catalogued in #602: the branch-protection rejection, the systematic exclusion of failed runs, and the missing trend fields for the noisiest gated metric family.

snowyukitty:
Force-pushed a hardening pass (`1caa13ba` → `92f6e12c`) before you spend time on this — **no behaviour change**, it is comments plus four small robustness edits, written up in the PR body under "A second pass over my own step". The one worth knowing about if you already read the summary: a `git fetch` that fails for a reason *other* than "branch absent" lands on the same code path, and the reasoning for why that cannot erase the branch (a parentless commit is not a fast-forward onto an existing ref, and nothing here ever forces) now lives in the step instead of only in my head.

The live run against a real remote was redone against the amended body, and the runnable copy is now extracted from `perf.yml` by a generator that asserts it differs from the shipped step on exactly one line — the branch name — so the testing claims describe the step you are reviewing rather than a hand-copy of it.

openwong2kim (approved):
This is a great PR — the simulation harness that extracts the step from perf.yml itself, and the two real defects it caught (the CRLF-in-filename one is nasty), is exactly the kind of verification this step needed. Answers on the five decisions:

1. Deriving the trend fields from the gate tables — yes. This is the real fix for defect 3; the hand-copied list is how it drifted in the first place.
2. Plumbing commit — yes. Not touching the packaged working tree and getting fast-forward-by-construction out of it is the right shape.
3. `!cancelled()` — agreed, keep it. A cancelled run is a half-finished measurement; putting it in a series whose whole purpose is telling regressions from noise would be self-defeating. No `always()`.
4. Deleting `bench/history.ndjson` from main — agreed, delete it. A frozen archive in `bench/` would be another file that looks alive and isn't, which is the exact trap this issue is about. The pre-break lines stay readable in history, and the field names are compatible if we ever want to prepend them to the branch.
5. Fail-closed on an unreadable trend — yes, and glad the live run forced it.

One thing to watch after merge: the simulation pushed with your credentials, so the `GITHUB_TOKEN` path first runs for real on the first push to main — the bootstrap path. I'll keep an eye on that run; if the publish fails it'll be loud now, which is the point.

Thanks for the thorough writeup — it made this easy to review.

snowyukitty:
The bootstrap ran for real on the merge commit, so the `GITHUB_TOKEN` path you flagged is confirmed. From the Perf run on `12fe18a`:

```
 * [new branch]      57880299142f62502202fcd542cafa36513b76dc -> bench-history
Published the trend line to 'bench-history' (57880299142f62502202fcd542cafa36513b76dc). Trend length: 1.
```

`bench-history` now holds five lines from five pushes to `main`: one parentless bootstrap commit carrying the explanatory subject, four appends each with a single parent, one tree entry at `100644 history.ndjson`, zero CR bytes. All 15 gated metrics are in every line, including the four `hiddenFlood` fields that had never been recorded once.

I put the early numbers on #570 — one of them is a knife-edge the trend surfaced on its literal first day, which is a fair argument for the whole exercise.

Thanks for the review, and for the steer that shaped it.

PR #610, opened 2026-07-25
feat(updater): macOS (Apple Silicon) in-app auto-update

### LINKED ISSUE #609: macOS in-app auto-update (signed-ZIP self-update, Phase E)
## Problem

On macOS, Settings → "Check for updates" does nothing: `AutoUpdater.ts` gates the whole in-app updater on `process.platform === 'win32'` (AutoUpdater.ts:36). macOS users must manually download the DMG even though a newer release exists.

Additionally, the renderer's `handleCheck` (SettingsPanel.tsx UpdateStatus) ignores the handler's direct `not-available` return, so unsupported platforms get stuck in the 'checking' state.

## Plan

Keep the fail-closed SHA-256 manifest pinning philosophy and install atomically via Squirrel.Mac (ShipIt):

1. **Detect** — poll `update.electronjs.org/openwong2kim/wmux/darwin-arm64/<version>` (infra already exists: MakerZIP darwin artifact `wmux-darwin-arm64-<ver>.zip` is signed + notarized + stapled in CI).
2. **Download + verify** — CI publishes a new `update-manifest-darwin-arm64.json` (version / file / sha256 / url) release asset; reuse the existing `downloadAndVerify` SHA-256 pinning path.
3. **Install** — serve the verified local ZIP over a loopback HTTP server (random port + random token path) as a Squirrel.Mac JSON feed, then `autoUpdater.setFeedURL` + `quitAndInstall()` (same pattern electron-updater uses; Squirrel.Mac does not trust file:// feeds).

### Changes
- `src/main/updater/verifyUpdate.ts` — generalize manifest schema (`setupExe` | `file` → `fileName`); Windows manifest stays byte-compatible.
- `src/main/updater/LocalUpdateFeed.ts` (new) — loopback feed server, Electron-free.
- `src/main/updater/AutoUpdater.ts` — darwin-arm64 support, platform-specific feed/manifest URLs, darwin install path, unsigned-build fail-closed error message.
- `.github/workflows/release.yml` — macos job publishes `update-manifest-darwin-arm64.json`.
- `src/renderer/components/Settings/SettingsPanel.tsx` — fix 'checking' stuck state on unsupported platforms.
- Tests: platform gating, manifest schema, LocalUpdateFeed, download filename generalization.

### Out of scope
- CLI shim refresh after update (macos-compat-plan #3) — follow-up.
- Intel (x64) builds — not produced.

Daemon survives the quit/relaunch (detached), so sessions persist across the update, matching the Windows UX.

openwong2kim:
Closes #609.

On macOS, Settings → "Check for updates" did nothing: the whole updater was gated on `process.platform === 'win32'`, so Mac users had to notice a release, download the DMG and drag it over the old app. This adds the darwin-arm64 path, keeping the existing fail-closed SHA-256 pinning philosophy end to end.

## How it works

1. **Detect** — poll `update.electronjs.org/openwong2kim/wmux/darwin-arm64/<version>`. Same code path as win32; only the platform segment differs.
2. **Verify** — the release pipeline now publishes `update-manifest-darwin-arm64.json`, and the existing `downloadAndVerify` pins the ZIP's SHA-256 against it. Any mismatch or transport error refuses the install outright.
3. **Install** — Squirrel.Mac does not trust `file://` feeds, so the verified ZIP is served from `127.0.0.1` on an ephemeral port under a random token path (the pattern electron-updater uses), handed to Electron's built-in `autoUpdater` as a JSON feed, and installed on `quitAndInstall()`. The daemon is detached, so sessions survive the relaunch exactly as on Windows.

## Changes

- `src/main/updater/verifyUpdate.ts` — `UpdateManifest` generalized to `{version, fileName, sha256, url}`. `validateManifest` accepts `setupExe` (the Windows manifest, byte-unchanged so already-shipped clients keep working) or `file` (darwin) and normalizes both. Every other check — https + `github.com` allowlist, 64-hex sha256, version match — is unchanged and applies to both shapes.
- `src/main/updater/LocalUpdateFeed.ts` (new) — loopback feed server, pure `node:http`, electron-free so it is unit-testable. Two token-scoped routes, everything else 404s; lives only for the duration of one install.
- `src/main/updater/AutoUpdater.ts` — `isUpdaterSupported` now `win32 || (darwin && arm64)`; per-platform feed segment and manifest file; temp download name from the manifest instead of a hardcoded `.Setup.exe`. `performInstall`'s win32 branch is untouched. Any Squirrel error tears down the feed, clears the install guard and surfaces `UPDATE_ERROR` — a code-signature failure (an unsigned local build can never self-update) becomes an actionable message pointing at the DMG rather than a silent no-op.
- `.github/workflows/release.yml` — the macOS job hashes the signed+notarized darwin-arm64 zip and uploads the manifest. Separate file from the Windows manifest, so the two legs never race and a failure in one cannot damage the other. The Windows job is untouched.
- `src/renderer/components/Settings/SettingsPanel.tsx` — `handleCheck` ignored the handler's direct `not-available` reply, leaving the row stuck on "checking" forever on any platform without an in-app updater. Fixed (existing i18n keys; no new strings).

## Tests

- `LocalUpdateFeed.test.ts` (new) — feed JSON shape, ZIP served byte-for-byte with Content-Length, 404 on any path outside the random token, unreachable after `stop()`.
- `AutoUpdater.platform.test.ts` — win32 invariants kept byte-identical; new darwin-arm64 coverage (own feed + manifest URLs, loopback JSON feed handed to Squirrel, `quitAndInstall` on `update-downloaded`, code-signature error fails closed); the "unsupported" cases are now linux and Intel macOS.
- `verifyUpdate.test.ts` — `setupExe`/`file` both normalize to `fileName`, reject when neither is present, and every other validation still applies to the darwin shape.

`npm run test:parallel`: 7845 passed / 22 skipped, 579 files. Typecheck clean; lint reports no new errors on touched files. (`bootMarker.runtime.test.ts` fails identically on a clean `main` checkout — pre-existing timing flake, unrelated.)

## Out of scope

- CLI shim refresh after an update (macos-compat-plan #3) — follow-up.
- Intel (x64) macOS — no build is produced.

Verified end to end only at the unit level; the live path needs a real signed release tag with `ENABLE_CROSS_PLATFORM_RELEASE=true`.

## Summary by CodeRabbit

- **New Features**
  - Enabled in-app updates for Apple Silicon macOS using a locally served, verified update feed and macOS-appropriate installation flow.
  - Normalized update manifest artifact names across platforms for consistent download/verification.
- **Bug Fixes**
  - Prevented update UI from remaining stuck in “Checking for updates…” and ensured correct “Up to date” / “not-available” behavior.
  - Added fail-closed handling for macOS update errors and improved cleanup for superseded/stale downloads.
- **Documentation**
  - Updated the changelog with macOS auto-update behavior, verification, and DMG guidance on self-update limitations.
- **Tests**
  - Expanded macOS Apple Silicon coverage plus loopback feed and verification scenarios (including failure paths).


openwong2kim:
[wmux-hermes] This PR adds in-app auto-update support for Apple Silicon macOS (darwin-arm64), closing the gap where Settings → "Check for updates" was a no-op on Mac and users had to manually download and install the DMG. Detection still polls update.electronjs.org and pins the downloaded ZIP's SHA-256 against a per-platform manifest published by CI, but the final install hands off to Electron's built-in Squirrel.Mac `autoUpdater` via a new loopback `LocalUpdateFeed` (bound to `127.0.0.1` on an ephemeral port under a random token path) because Squirrel.Mac refuses `file://` feeds. The manifest schema in `src/main/updater/verifyUpdate.ts` is generalized (`setupExe` | `file` → `fileName`) while staying byte-compatible with already-shipped Windows clients, `src/main/updater/AutoUpdater.ts` gains the darwin platform branch plus fail-closed error handling that surfaces an actionable message (with a DMG link) when a locally-made unsigned build cannot self-update, `.github/workflows/release.yml` publishes the separate `update-manifest-darwin-arm64.json`, and `src/renderer/components/Settings/SettingsPanel.tsx` fixes a stuck "Checking for updates…" row on unsupported platforms by honoring the handler's direct `not-available` reply. The daemon is detached, so sessions survive the quit/relaunch exactly like the Windows update path.

Issue #608, opened 2026-07-25
wmux web: ship a full Content-Security-Policy for the terminal page


openwong2kim:
From the security review of the web terminal surface (docs/web-security-review-2026-07-24.md, finding W4).

The web terminal currently sends `Content-Security-Policy: frame-ancestors 'none'` only (shipped with the clickjacking fix). There is no script/connect policy, so if a future regression ever rendered session metadata as HTML, the injected script could read the token from `sessionStorage` and exfiltrate it anywhere.

There is no XSS today — xterm.js renders raw bytes and metadata goes through `textContent` — so this is purely a containment layer. The key directive is `connect-src 'self'`: even with script execution, the token couldn't leave the origin.

Plan: `scripts/build-daemon-web.mjs` already inlines everything into one self-contained page, so it can compute `sha256-...` hashes of the inlined script blocks after substitution and inject them into the policy. Target:

```
default-src 'none';
script-src 'self' 'sha256-<xterm>' 'sha256-<app>';
style-src 'self' 'unsafe-inline';
img-src 'self' data:;
connect-src 'self';
manifest-src 'self';
frame-ancestors 'none';
base-uri 'none';
form-action 'self';
```

Hashing is deterministic over the built blob, so it's stable per release. Ship as the full policy in `securityHeaders()` (`src/daemon/web/WebTerminalServer.ts`).

Acceptance: no CSP violations in devtools on a working session (terminal stream, pairing, PWA manifest all fine), and a test asserting `connect-src` is present on `GET /`. Estimated ~half a day, mostly build-script work.

openwong2kim:
[wmux-hermes]

This issue proposes shipping a complete Content-Security-Policy for the wmux web terminal page, which currently sends only `frame-ancestors 'none'` and has no script or connect policy. It originates from a security review of the web terminal surface (finding W4) and notes there is no active XSS today — xterm.js renders raw bytes and session metadata goes through `textContent` — so the policy is a containment layer rather than a fix for a live exploit. The change touches `src/daemon/web/WebTerminalServer.ts` (the `securityHeaders()` function) and `scripts/build-daemon-web.mjs`, which inlines all script/style into a single self-contained page and would compute `sha256-...` hashes of those inlined blocks to permit in the `script-src` directive. The key directive is `connect-src 'self'`, which would ensure that even in a hypothetical script-injection regression the access token could not be exfiltrated off-origin.

Issue #607, opened 2026-07-25
wmux web: --expose warning doesn't mention cleartext sniffing; no TLS option


openwong2kim:
From the security review of the web terminal surface (docs/web-security-review-2026-07-24.md, finding W1).

`wmux web --expose` binds 0.0.0.0 over plain HTTP. The per-start token and the full scrollback travel unencrypted, so anyone who can sniff the network (open Wi-Fi, ARP spoofing, corporate proxy) can grab the token and read every served pane — and type into them if `--allow-input` is on.

The current warning (`src/cli/commands/web.ts`) says the token is the only gate and to treat the URL as a secret, but never says the token itself is sniffable on untrusted networks. The `tailscale serve` mention is framed as a PWA convenience rather than the security floor.

Two parts:

**Quick fix (~10 lines):** when the bind is exposed, print an explicit cleartext warning, something like:

```
⚠ EXPOSED OVER PLAIN HTTP. The access token and your full scrollback travel
  UNENCRYPTED. Anyone on this network who can sniff traffic can steal the
  token and read every pane. Use this only on a trusted network, or front it
  with HTTPS (`tailscale serve`, caddy, nginx) before opening it remotely.
```

Acceptance: the `--expose` report contains "UNENCRYPTED" plus a concrete HTTPS fronting instruction, with a test asserting the wording.

**Longer term (open to discussion):** actually remove the risk instead of warning about it — `--tls-cert`/`--tls-key` flags, or auto-suggesting `tailscale serve --bg` when a Tailscale adapter is detected (`collectIpv4` already recognizes 100.64.0.0/10). Warning wording only mitigates operator awareness; TLS is the only real fix.

Not exploitable in the default config (loopback + read-only) — this only matters after an explicit `--expose` opt-in.

openwong2kim:
[wmux-hermes]

[...]

PR #611, opened 2026-07-25
fix(web): warn that --expose serves the token and scrollback in cleartext

openwong2kim:
## What

`wmux web --expose`'s report warned that the access token was "the only thing gating it" and to treat the URL as a secret. True, but it missed the sharper point: the server speaks plain HTTP on all interfaces, so anyone who can sniff the network (open Wi-Fi, ARP spoofing, a compromised switch) captures the token **and the full scrollback** straight off the wire — they never need the URL. This is finding W1 (Tier A) from `docs/web-security-review-2026-07-24.md`.

The exposed-bind branch now prints:

```
⚠ UNENCRYPTED. Anyone on this network who can sniff traffic (open Wi-Fi,
  ARP spoofing, compromised switch) can read the access token and the
  full scrollback. Use `tailscale serve` (HTTPS) or wait for native TLS.
```

## Also in this PR

- A source-invariant test asserting the exposed branch contains the UNENCRYPTED wording, the sniffing statement, and the `tailscale serve` instruction (the acceptance criteria from the review doc).
- The review doc itself overstated this wording as already shipped; corrected the W1/W4 status rows and linked them to #607 / #608.
- CHANGELOG entry under Unreleased.

## Not in this PR

Native TLS (`--tls-cert`/`--tls-key` or a `tailscale serve` wrapper) stays tracked in #607 — wording only mitigates awareness; TLS is the real fix. Full hashed CSP is #608.

Part of #607.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a clearer warning when exposing the web interface over unencrypted HTTP.
  * Warns that network observers could capture the access token and terminal scrollback.
  * Recommends using `tailscale serve` to provide an HTTPS front.

* **Documentation**
  * Corrected security review statuses to accurately reflect the current warning and TLS implementation progress.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`wmux web --expose` now says out loud that the connection is unencrypted (#607, partial).** The old warning told you the access token was "the only thing gating it" and to treat the URL as a secret — true, but it skipped the sharper point: over plain HTTP on all interfaces, anyone who can sniff the network (open Wi‑Fi, ARP spoofing, a compromised switch) can read the token *and* the full scrollback off the wire, no URL required. The exposed-bind report now states exactly that and points at `tailscale serve` for an HTTPS front. Wording only — native TLS stays tracked in #607.
+
 ## [3.34.0] — 2026-07-25
 
 ### Added
--- a/docs/web-security-review-2026-07-24.md
+++ b/docs/web-security-review-2026-07-24.md
@@ -68,10 +68,10 @@ Boundary strengths already in place (no action needed):
 
 | ID | Severity | Finding | Location | Status |
 |---|---|---|---|---|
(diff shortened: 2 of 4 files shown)


openwong2kim:
[wmux-hermes] This PR sharpens the warning printed by `wmux web --expose` when the server binds to all interfaces over plain HTTP. The previous message noted that the access token was the only gate and should be treated as a secret; the new wording states explicitly that the token and the full terminal scrollback travel **UNENCRYPTED** and can be captured by anyone able to sniff the network (open Wi-Fi, ARP spoofing, a compromised switch), and it points users at `tailscale serve` for an HTTPS front. It touches `src/cli/commands/web.ts` (the warning string), adds a source-invariant test in `src/cli/commands/__tests__/web.exposeWarning.test.ts` asserting the `UNENCRYPTED`/`sniff`/`tailscale serve` wording, corrects the W1/W4 status rows in `docs/web-security-review-2026-07-24.md` (the doc had overstated this wording as already shipped), and records a CHANGELOG entry. This is the Tier A wording mitigation from security review finding W1; it improves operator awareness on an opt-in `--expose` config but does not change the default loopback/read-only behavior, and native TLS remains tracked separately in #607.

PR #612, opened 2026-07-25
fix(web): serve the terminal page with a full Content-Security-Policy

### LINKED ISSUE #608: wmux web: ship a full Content-Security-Policy for the terminal page

openwong2kim:
## What

Finding W4 from `docs/web-security-review-2026-07-24.md`. The web terminal page previously shipped `Content-Security-Policy: frame-ancestors 'none'` only. `GET /` now carries the full strict policy:

```
default-src 'none'; script-src 'self' 'sha256-…' 'sha256-…' 'sha256-…';
style-src 'self' 'unsafe-inline'; img-src 'self' data:; connect-src 'self';
manifest-src 'self'; worker-src 'self'; frame-ancestors 'none';
base-uri 'none'; form-action 'self'
```

There is no XSS today (xterm renders raw bytes, metadata goes through `textContent`) — this is containment: with `connect-src 'self'`, even a future regression that executes injected script cannot send the token anywhere off-origin.

## Design choice

The review doc suggested computing the hashes in `build-daemon-web.mjs`. I put them in the daemon instead: `loadAssets()` hashes each inline `<script>` block of the exact `terminal.html` bytes it is about to serve. The build script and the policy can then never drift apart — a rebuilt page automatically gets matching hashes, and there is no manifest file to forget.

Notes:
- `style-src` keeps `'unsafe-inline'`: xterm manages styles dynamically, and a style injection cannot reach the token.
- `worker-src 'self'` admits the `/sw.js` service worker that `default-src 'none'` would otherwise block.
- API/JSON responses keep the minimal `frame-ancestors 'none'` baseline — hashes are meaningless there.

## Verification

- New test: `GET /` (synthetic two-script page) returns the full policy with exactly one hash per inline block and no `'unsafe-inline'` in `script-src`; API responses keep the baseline. All 38 web server tests pass, `tsc --noEmit` clean.
- Audited the real built page (577 KB): 3 inline script blocks, zero external requests (the only `http://` strings are license comments and SVG namespaces), no inline event handlers, no `style=` attributes, no `eval`/`new Function`.

Closes #608

## Summary by CodeRabbit

* **Security**
  * Added a strict Content Security Policy to the `wmux web` terminal page.
  * Inline scripts are protected with exact, hash-based allowlists, without relying on unsafe inline execution.
  * Restricted connections to the application itself to help prevent access-token exfiltration.
  * Permitted service-worker functionality while maintaining other security protections.

* **Documentation**
  * Updated security review and changelog documentation to reflect the CSP hardening as shipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **The `wmux web` page now ships a full Content-Security-Policy (#608).** The page previously carried only frame protection. It now declares a strict policy — every inline script is allowed by its exact hash (no `'unsafe-inline'`), and `connect-src 'self'` means that even if a rendering bug ever reintroduced an XSS, the injected script could execute but could not send the access token anywhere. The hashes are computed by the daemon from the exact page bytes it serves, so the policy can never drift out of sync with a build. Purely defense-in-depth: there is no known XSS today, and nothing changes for a working session.
+
 ### Fixed
 
 - **`wmux web --expose` now says out loud that the connection is unencrypted (#607, partial).** The old warning told you the access token was "the only thing gating it" and to treat the URL as a secret — true, but it skipped the sharper point: over plain HTTP on all interfaces, anyone who can sniff the network (open Wi‑Fi, ARP spoofing, a compromised switch) can read the token *and* the full scrollback off the wire, no URL required. The exposed-bind report now states exactly that and points at `tailscale serve` for an HTTPS front. Wording only — native TLS stays tracked in #607.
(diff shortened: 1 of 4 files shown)


openwong2kim:
## [wmux-hermes] PR Summary

This PR implements finding W4 from the web security review, serving the `wmux web` terminal page with a full strict `Content-Security-Policy` instead of the previous `frame-ancestors 'none'`-only header. The policy is computed by the daemon at asset-load time via a new `buildHtmlCsp()` helper in `src/daemon/web/WebTerminalServer.ts`, which hashes each inline `<script>` block of the exact `terminal.html` bytes being served so that the hashes and the build can never drift apart; the served `GET /` then carries `default-src 'none'; script-src 'self' <hashes>; ...; connect-src 'self'; frame-ancestors 'none'`. There is no known XSS today (xterm renders raw bytes and metadata goes through `textContent`), so this is purely a defense-in-depth containment layer: with `connect-src 'self'`, even a future regression that managed to execute injected script could not exfiltrate the access token from `sessionStorage` to any off-origin endpoint.

The change is isolated to `src/daemon/web/WebTerminalServer.ts` (new `buildHtmlCsp` export plus the `htmlCsp` field attached to the HTML route only; API/JSON responses keep the minimal baseline policy) with a corresponding unit test in `src/daemon/web/__tests__/WebTerminalServer.test.ts` asserting one hash per inline block, absence of `'unsafe-inline'` in `script-src`, and that API responses still carry only the baseline. Documentation (`docs/web-security-review-2026-07-24.md`) and `CHANGELOG.md` are updated to mark W4 as resolved.

This hardens the web terminal surface against the worst outcome of a hypothetical future XSS — token theft — and closes #608. It is a security-hardening change with no behavioral impact on working sessions.

PR #614, opened 2026-07-25
feat(onboarding): offer Claude Code statusline install in the first-run wizard

openwong2kim:
## What

The first-run wizard now shows an opt-in **Enable statusline** step when Claude Code is detected. One click installs the per-account usage statusline (model · account · context % · 5h/7d rate limits under Claude Code's input box) — the same thing `wmux setup-statusline` does, which until now was CLI-only and invisible to app-only users. Works on macOS and Windows; the existing installer already handles both.

## Why

The statusline backend, IPC handlers, and preload bridge all existed, but no UI ever called them — an app-only user (winget/Setup.exe) had no way to discover the feature.

## Design constraints respected

- **Strictly opt-in** (no-auto-install-at-boot decision, 2026-07-17): nothing runs without a click.
- The block is **hidden** when the statusline is already installed, or when every settings target is foreign-owned or corrupt (a button that can only no-op is noise).
- A user-authored `statusLine` is never overwritten (existing installer guarantee).
- Success is reported only when a target actually took the install — `ok:true` alone also covers the all-skipped case if settings changed between the status probe and the click.
- Install failures surface the underlying error message; the dialog gained `max-height`/`overflow-y` so the extra block can't push the footer off small screens.

## Tests

- `decideStatuslineOffer` offer/hidden matrix (installed, foreign-only, corrupt, missing/none targets)
- `StatuslineBlock` render variants (offer / installing disabled / installed / error)
- Full suite for the file: 30 passing; `tsc --noEmit` and eslint clean.

## Summary by CodeRabbit

- **New Features**
  - Added an optional Claude Code statusline setup step to the first-run wizard.
  - Users can enable the statusline when eligible, with clear installing, success, and error states.
  - Existing statuslines are preserved, and installation can be retried from the terminal if needed.
  - Updated English and Korean translations for the new wizard experience.

- **Documentation**
  - Documented the statusline opt-in flow and installation behavior in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **The first-run wizard now offers to enable the Claude Code statusline.** `wmux setup-statusline` — which shows model, context usage, and 5h/7d rate limits under Claude Code's input box — existed only as a CLI command, so app-only users never discovered it. The onboarding wizard now shows an opt-in "Enable statusline" step (macOS and Windows) when Claude Code is detected and at least one settings file can accept the install. It stays strictly opt-in: nothing installs at boot, an already-installed statusline shows nothing, and a user-authored statusline is never overwritten.
+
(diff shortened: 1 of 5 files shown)


openwong2kim:
[wmux-hermes]

This PR adds an opt-in "Enable statusline" step to the first-run wizard that surfaces the existing Claude Code usage statusline (model · account · context % · 5h/7d rate limits shown under Claude Code's input box) to app-only users who never see the CLI `wmux setup-statusline` command. The changes touch the wizard UI in `src/renderer/components/FirstRunWizard.tsx` (new `StatuslineBlock` component and `decideStatuslineOffer` gating logic), add bilingual i18n strings in `src/renderer/i18n/locales/en.ts` and `ko.ts`, expand the unit tests in `FirstRunWizard.test.ts`, and add a CHANGELOG entry. It improves feature discoverability for winget/Setup.exe-installed users on both macOS and Windows while preserving the strictly opt-in posture decided on 2026-07-17 — nothing installs at boot, the block is hidden when already installed or when no target can accept the install, and a user-authored statusline is never overwritten.

Issue #605, opened 2026-07-25
session:critical is a pattern match, not an approvable request — rework needed before any remote approve/deny surface


openwong2kim:
## What this is

`session:critical` looks like an approval request, but it is a pattern match on
whatever the terminal prints. Before any surface grows an "Approve / Deny" button —
the web page, or a phone — the signal underneath needs rework. Filing this as its own
issue so the approval-surface work can track it as a dependency.

## The three problems

**1. The event fires on output, not on a pending action.**

`CRITICAL_PATTERNS` (`src/main/pty/AgentDetector.ts`) are tested against every
completed line of PTY output. The event fires when an agent *prints*
`git push --force` — in a `git log`, in a diff hunk, in a README it is reading out
loud. Nothing is blocked, nothing is waiting, and there is no addressee for a "Deny".
Answering one of these with a keystroke would type into a *running* command.

The orchestrator already treats this class of signal as too weak to act on:
`CommanderEventCoalescer` forbids auto-answering a regex-sourced `awaiting_input`
and requires reading the pane first. A human on a phone deserves at least the same
bar.

**2. The payload discards the one thing an approver needs.**

The detector computes the matched line, uses it for dedup only, and emits just a
category label:

```ts
const value = clean.slice(0, 80);          // the actual matched line
if (this.lastEmittedFor.get(key) === value) return;
this.lastEmittedFor.set(key, value);
cb({ action: cp.label, riskLevel: cp.riskLevel });   // label only — line discarded
```

`cp.label` is one of 11 hardcoded strings, so `git push --force origin main` and
`git push -f scratch` produce byte-identical events. No command, no cwd, no
surrounding context, no request identity.

**3. The dedup silently drops repeats.**

Same label + same line within a cycle → no event at all. For a notification that is
tolerable; for anything approval-shaped it means a lost request with no trace.

## Direction

- Demote `session:critical` to notify-only, permanently. It stays useful as a
  heads-up ("the pane printed something spicy — look"), and its payload should carry
  the matched line and basic context so the heads-up is actually informative.
- Anything actionable gates on the hook-sourced `awaiting_input` signal instead —
  that one corresponds to a real pending prompt, arrives with the agent's own
  envelope, and is the only signal with a defensible request lifecycle.
- Replace drop-on-repeat dedup with per-request identity on whatever becomes
  actionable, so a second attempt is a second request rather than silence.

## Why now

The web surface just gained durable state and replayable attention events
[...]

PR #616, opened 2026-07-25
Reach your agents from a phone: hooks in the daemon, remote answers, per-device credentials

openwong2kim:
Everything a phone app needs from the daemon, built and verified on real hardware. The iOS app itself is not here — this is the server side it will talk to, and the browser surface gets all of it too.

## What this does

**Agent hooks reach the daemon, so signals survive the desktop closing.** The bridge every agent CLI calls used to deliver to the GUI's pipe; close the window and the authoritative signal path died, leaving only screen-scraping heuristics. It now targets the always-on daemon first and falls back to the desktop only when the daemon is unreachable, with the hook-beats-detector arbitration moved along with it so both sources are settled in one process. An old bridge keeps working through a one-hop relay, and `WMUX_HOOKS_TO_MAIN=1` restores the old targeting.

**Answer an agent's question from anywhere.** When a pane raises an AskUserQuestion prompt the daemon records it as a pending request carrying the question text and option labels, and any authenticated surface can answer it. One pending per pane, racing answers settle cleanly, a restart invalidates everything pending, and the daemon re-reads the pane before writing a byte — refusing rather than typing into a screen that moved on. Only hook-delivered signals can create a request; heuristics never can, which is the bar the orchestrator already holds itself to. Deliberately allowed on a read-only server, because answering one daemon-raised request is strictly narrower than `--allow-input`.

**Pair a phone by name, revoke that one phone.** Pairing used to hand over the server-wide token, and since #596 made that token durable, every paired device shared one secret. Each device now gets its own credential, stored as a hash, revocable on its own, with revocation cutting live streams immediately and failing closed. Credentials are header-only — a long-lived secret must never sit in a URL — so a browser trades its credential for a short-lived, device-bound ticket to open a stream.

**`wmux web --tailscale`** collapses tailnet setup into one command and removes the serve config again on `--stop`. It refuses rather than half-works, and now prints what tailscale actually said when it can't.

**A reconnecting phone stops re-downloading the whole scrollback** — the first paint is capped without ever cutting a character or escape sequence in half.

**Groundwork for push:** a stateless relay that cannot read what it forwards (notifications are sealed with a key derived from the pairing secret before they leave the machine, and the log helper's type makes a payload leak a compile error), plus the byte-exact envelope spec the iOS Notification Service Extension will implement, with known-answer vectors so the Swift side can prove compatibility without a device.

## Dogfooding found six things the test suite could not

Run against a packaged build on a real phone over a real tailnet. Every one of these passed unit tests and harnesses first:

1. **The plugin manifest version never moved**, so Claude Code would have kept serving the old bridge from its version-pinned cache forever — correct code, green tests, and a change that could never arrive.
2. **Home-screen launches lost their login.** `start_url` is `./`, so an installed app opens with no token in the URL, and the credential lived in sessionStorage, which iOS drops when it evicts a PWA.
3. **The pairing refusal explained itself and the page threw it away**, rendering "Pairing failed."
4. **The page had no cache headers**, so a phone kept running a build that had already been fixed.
5. **`--tailscale` swallowed tailscale's own answer** — "Serve is not enabled on your tailnet" plus a one-click URL — because it read stderr (tailscale writes it to stdout) and then printed only the first line, which is the exec wrapper.
6. **Rejected keystrokes vanished silently.** fetch resolves on 401 and 403, so a lone `.catch()` saw neither.

Four of the six are the same shape: the server answered correctly and the client discarded it.

## One thing to know about the merge

main shipped its own CSP in #612 and reached the same conclusions independently. Two differences were resolved on their merits: its hashes were computed over the file's raw bytes, and the HTML parser rewrites CRLF to LF before the tokenizer runs — so a page built from a Windows checkout produced hashes matching nothing the browser computes and **all three inline blocks were refused, giving a blank terminal**. Measured here: 3 of 3 blocks differ between raw and normalized hashing. The normalizing builder supersedes it and the old one is removed rather than left as a second, wrong way to do the same thing. Placement went the other way — main puts the full policy on the HTML and a baseline elsewhere, which is right, because script hashes would otherwise add ~250 bytes to every JSON reply and a phone typing pays that per keystroke.

## Verification

Four live-daemon harnesses, all green against the final merged build, plus 2579 unit tests and four typecheck configs:

- `scripts/hooks-daemon-dynamic.mjs` — 6/6. Hook authority with no GUI, arbitration driven by real ConPTY output through the real detector, a 100-signal flood with the daemon staying responsive, SIGKILL restart.
- `scripts/approvals-dynamic.mjs` — 10/10. Question and options verbatim over HTTP and the pipe, approve on a read-only server while `/api/input` stays 403 on that same server, the real `❯` glyph through ConPTY, prompt-gone refusing with zero bytes written.
- `scripts/m3-csp-device-stream-dynamic.mjs` — 7/7. CSP verified in a real browser with zero violations, device credentials and tickets end to end, revocation cutting streams and tickets while sparing the neighbour.
- `scripts/bridge-dogfood.mjs` — 4/4. Drives the real bridge the way Claude Code drives it, which nothing else had ever done.

On the owner's machine, over HTTPS on the tailnet: hooks arriving with `target: daemon` at p50 6ms, a phone pairing and typing into a pane, device credentials issued and revoked.

## Not here

The iOS app (M6), and the daemon-side push sender, which needed the device records this adds. `session:critical` is still notify-only and unsuitable for a remote approve button — that rework is #605.

## Summary by CodeRabbit

* **New Features**
  * Added per-device phone pairing and credentials, per-device revocation, and improved phone reconnect behavior.
  * Added phone-based approval answering with safer prompt/race handling and daemon-backed lifecycle replay.
  * Added an encrypted APNs push relay and documented relay deployment.
  * Enhanced `wmux web --tailscale` with simplified one-flag setup/teardown, stronger security headers/CSP, SSE keep-alives, stream-ticket auth, and improved session resume.
* **Bug Fixes**
  * Improved daemon-first hook/bridge routing to avoid duplicate or ambiguous endpoint firing.
* **Documentation**
  * Added phone-client contract documentation and relay API/threat-model guidance.
* **Tests / Chores**
  * Expanded CI and Vitest coverage for web, relay, hook, and approval flows.


openwong2kim:
## [wmux-hermes] Summary

This is a large feature PR (+17,860 / −1,199 across 74 files) that brings the server-side groundwork for accessing wmux agents from a phone — the iOS app itself is not included, only the daemon and browser surfaces it will talk to. The four pillars are: (1) agent hooks now target the always-on daemon first with fallback to the GUI (`HookIngest.ts`, `hooks.rpc.ts`, and the Claude/Codex/OpenClaude bridges in `integrations/*/bin/`), moving hook-beats-detector arbitration into the daemon so signals survive the desktop closing; (2) a remote-answers path where an AskUserQuestion prompt becomes a pending daemon request carrying question text and options that any authenticated surface can resolve, with one-pending-per-pane, restart invalidation, and a pre-write pane re-read (`src/daemon/approvals/*`, gated to hook-sourced signals only); (3) per-device pairing credentials (hashed, individually revocable, header-only with a short-lived device-bound ticket for streams) replacing the shared server-wide token, in `src/daemon/web/DeviceStore.ts` and `WebTerminalServer.ts`; and (4) a new `wmux web --tailscale` command (`src/cli/tailscale.ts`, `src/cli/commands/web.ts`) that collapses tailnet setup into one step and tears it down with `--stop`.

Additional changes include a stateless push-notification relay under a new `relay/` package (Cloudflare Worker with sealed payloads and known-answer vectors for the future iOS Notification Service Extension), a normalizing CSP builder (`webCsp.ts`) that fixes a CRLF-vs-LF hashing bug that produced blank terminals on Windows checkouts, scrollback snapshot windowing (`snapshotWindow.ts`), SSE heartbeats, and four live-daemon dogfood harnesses under `scripts/`. The hook signal router, flood meter, and latency meter are relocated from `src/main/hooks/` to `src/shared/hooks/`, and the PR resolves a CSP collision with main (#612) by adopting the normalizing builder and main's HTML-first policy placement.

openwong2kim:
@codex review

openwong2kim:
@codex review

openwong2kim:
[wmux-hermes]

This PR adds the server-side infrastructure a phone app needs to reach the wmux daemon, verified on real hardware, touching 78 files across the daemon, CLI, integrations, and a new `relay/` component. The headline changes are: agent hooks now target the always-on daemon first (falling back to the desktop GUI only when the daemon is unreachable), with hook-beats-detector arbitration moved into the daemon so signals survive the desktop closing; the daemon records AskUserQuestion prompts as pending requests carrying question text and option labels that any authenticated surface can answer, with one-pending-per-pane, race-safe resolution, restart invalidation, and a pane re-read before writing so it refuses rather than typing into a screen that moved on; per-device phone pairing replaces the shared server-wide token with individual hashed credentials that are revocable on their own and cut live streams immediately; `wmux web --tailscale` collapses tailnet setup into one flag and tears it down on `--stop`; and a stateless, end-to-end-sealed APNs push relay (`relay/`) lays groundwork for notifications, with the byte-exact envelope spec and known-answer vectors the iOS Notification Service Extension will implement.

The files span `src/daemon/` (hooks, approvals, push, web server, device store, CSP), `src/cli/commands/web.ts` and `src/cli/tailscale.ts`, the Claude/Codex/OpenClaude integration bridges under `integrations/`, the new `relay/` Cloudflare Worker, and shared signal types moved under `src/shared/`. The PR also merges main's independently-developed CSP work (#612), resolving a CRLF-vs-LF hash mismatch that produced a blank terminal on Windows checkouts, and lands six dogfood findings the test suite could not catch (plugin manifest version pinning, home-screen login loss, pairing error surfacing, cache headers, `--tailscale` stderr/stdout, and silent 401/403 rejection). Verification is four live-daemon harnesses plus 2579 unit tests across four typecheck configs.

PR #621, opened 2026-07-25
chore(release): 3.35.0

openwong2kim:
Version bump, CHANGELOG promotion, regenerated API reference. Four files, the same shape as every previous release commit.

## What ships in 3.35.0

Everything under `[Unreleased]`, which is the phone-access line from #616 plus the security work that landed alongside it:

- **Reach your agents from a phone.** Hooks arrive at the always-on daemon rather than the desktop, so an agent signal survives closing the window. An `AskUserQuestion` prompt becomes a request any paired surface can answer — allowed on a read-only server, because answering one daemon-raised request is strictly narrower than `--allow-input`.
- **Per-device credentials.** Pairing hands each phone its own revocable credential instead of the server-wide token; revoking one cuts its live streams immediately and fails closed. `wmux web --new-token` revokes every paired device, which it had stopped doing once credentials became per-device.
- **`wmux web --tailscale`** sets up tailnet HTTPS in one command, and `--status` now prints the address a phone can actually use rather than the loopback bind.
- **A full CSP** on the served page, with hashes computed from the exact bytes the daemon serves.
- **Groundwork for push:** a relay that cannot read what it forwards, and an envelope sealed to a key that never leaves the phone.

The iOS app itself is not in this release; the contract it implements is documented in `docs/phone-client-contract.md`.

## Verification

`main` is green at 84e751d5 (Baseline on all three platforms, Perf, CI). The API reference regenerates byte-identical apart from the version line the drift guard checks.

## Summary by CodeRabbit

* **Documentation**
  * Updated the API reference to reflect version 3.35.0 sources.
* **Release**
  * Published version 3.35.0.
  * Added the v3.35.0 release entry to the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [3.35.0] — 2026-07-25
+
 ### Security
 
 - **The `wmux web` page now ships a full Content-Security-Policy (#608).** The page previously carried only frame protection. It now declares a strict policy — every inline script is allowed by its exact hash (no `'unsafe-inline'`), and `connect-src 'self'` means that even if a rendering bug ever reintroduced an XSS, the injected script could execute but could not send the access token anywhere. The hashes are computed by the daemon from the exact page bytes it serves, so the policy can never drift out of sync with a build. Purely defense-in-depth: there is no known XSS today, and nothing changes for a working session.
--- a/docs/api/reference.md
+++ b/docs/api/reference.md
@@ -7,7 +7,7 @@
 
 # wmux API Reference (generated)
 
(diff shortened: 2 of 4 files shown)


openwong2kim:
[wmux-hermes]

This PR cuts the **3.35.0 release** — a version bump plus its supporting bookkeeping, no source logic changes. `package.json` and `package-lock.json` move from `3.34.0` to `3.35.0`, `CHANGELOG.md` promotes the existing `[Unreleased]` block into a dated `[3.35.0] — 2026-07-25` entry, and `docs/api/reference.md` is regenerated by `scripts/gen-api-reference.mjs` so the "Generated from wmux v3.35.0 sources" header matches the new version. The shipped scope of 3.35.0 is the phone-access line (per-device credentials, `wmux web --tailscale`, durable attention events) together with the security work (full CSP, CDP audit) that landed on `main` since 3.34.0; the iOS app itself is not in this release. Four files, identical in shape to every prior release commit.

Issue #622, opened 2026-07-25
coldStart.firstPtyDataMs varies 3.4x on CI, so the perf gate red-lights at random


openwong2kim:
`coldStart.firstPtyDataMs` swings far enough on CI runners that the gate cannot tell a regression from a quiet runner. It cost five re-runs during the 3.35.0 release today, and a gate that red-lights at random teaches people to re-run rather than look — which is the state where a real regression walks through.

## The clearest evidence: one commit, two runs

`d6c5e981` (the 3.35.0 release commit, which changes four version strings and a CHANGELOG heading — no code at all):

| Attempt | firstPtyDataMs | vs baseline | Result |
| --- | --- | --- | --- |
| 1 | 3038 ms | +151.7% | **FAIL** |
| 2 | 1385 ms | +14.7% | PASS |

Byte-identical tree, same hour, 2.2× apart. Nothing in that commit can reach a cold-start path.

## The spread across today

Every one of these ran on effectively the same code — the deltas between them were security fixes to daemon web/push paths, none of which are on the measured path:

| Commit | firstPtyDataMs | vs baseline (1207 ms) |
| --- | --- | --- |
| `179e4a08` | 889 ms | **−26.3%** |
| `695207cc` | 1367 ms | +13.3% |
| `d6c5e981` (rerun) | 1385 ms | +14.7% |
| `eb67d267` | 2339 ms | +93.8% (FAIL) |
| `d6c5e981` | 3038 ms | +151.7% (FAIL) |

Low to high is **3.4×**. The gate fires above ~2207 ms (`ratio: 1.5` **and** `absMargin: 1000` against a 1207 ms baseline, `scripts/perf-compare.mjs`), which sits inside that range rather than outside it.

Worth noting the metric is *already* median-of-3 within a run (`coldStart run 1/3…3/3` → `coldStart median`). The variance survives the median.

## It is specific, not general

On those same runs the frame metrics were rock solid — `frameBudget.N4/N8` and `hiddenFlood.*.frameDeltaMs` all reported `+0.0%` repeatedly. `inputLatency.echoMs.p95` is milder but also moves (13.6 → 25.1 ms across runs). So this is not "CI is noisy", it is cold start specifically: process spawn, disk, and first-paint timing on a shared runner.

`hiddenFlood.N4/N8.echoMs` are also permanently ~70-80% *below* baseline and have been printing "improved — consider refreshing baseline" for a while, which is its own smaller staleness problem.

## Options

[...]

PR #625, opened 2026-07-25
test(daemon): wait for the pipe server to register a socket, not for 50ms

openwong2kim:
`rotateToken invalidates old token, issues new one, drops connected sockets` timed out on `windows-latest` during the 3.35.0 release, with 8311/8333 otherwise passing, and went green on a plain re-run. Noted as a loose end in #622.

## What was actually wrong

The test, not the server.

`rotateToken()` destroys the sockets in `connectedSockets`, and the server adds a socket to that set inside its own `connection` handler. That is a different event, on the other side of the pipe, from the client's `connect`. The test bridged the two with `setTimeout(50)` — a guess about how far apart they land.

Named pipes have a longer accept path than unix sockets and the Windows leg is the slowest of the three-OS matrix, so that is where the guess loses. When it does: rotation iterates a set that does not yet contain the socket, the connection is never dropped, `await liveClosed` never resolves, and vitest kills the test with an anonymous "timed out in 5000ms" that names nothing.

## The change

Wait for the condition instead of for a duration. The server already exposes everything needed — `getConnectionCount()`, `isConnected`, and in one case the input buffer the assertion itself reads.

Six sites in `DaemonPipeServer.test.ts` had this shape. Only one had failed so far; the rest are the same defect and would fail the same way.

Second commit consolidates: four test files had each defined a byte-identical `waitFor`, so it now lives in `src/test-utils/waitFor.ts` and the three copies are gone. The label defaults to the predicate's own source, which is why the ~21 existing call sites did not have to change — the second argument is still a timeout, and the message improves anyway:

```
waitFor: () => server.getConnectionCount() === 1 did not hold within 4000ms
```

The three copies all threw a bare `waitFor: deadline exceeded`, which put their failures in the same position as the flake that started this.

## Deliberately not done

The same fixed-sleep pattern appears ~68 times across 20 test files. Most are legitimate — waiting out a debounce window, where the time *is* what is under test. Converting them without diagnosing each one would be a speculative rewrite of tests that have never failed. This covers the files that already had the helper, plus the one that needed it.

## Verification

- The four touched suites: 43/43. `DaemonPipeServer.test.ts` run five times, stable.
- Faster, not just sturdier: test time 2.75s → 2.36s locally, matching the ~410ms of fixed sleeps removed.
- Three typecheck configs, and the helper has its own tests — including that a timeout names the condition.

One `no-empty-function` error remains in `SessionPipe.attachSnapshot.test.ts`; it is on `main` already and is left alone.

## Summary by CodeRabbit

* **Tests**
  * Improved asynchronous test reliability by waiting for expected conditions instead of using fixed delays.
  * Added coverage for timeout handling, custom labels, immediate resolution, and readable error messages.
* **Refactor**
  * Consolidated repeated test polling logic into a shared utility with configurable timeouts.

### DIFF
--- a/src/daemon/__tests__/DaemonPipeServer.test.ts
+++ b/src/daemon/__tests__/DaemonPipeServer.test.ts
@@ -11,30 +11,8 @@ import {
   getDaemonAuthTokenPath,
   getLegacyDaemonAuthTokenPath,
 } from '../../shared/constants';
+import { waitFor } from '../../test-utils/waitFor';
 
-/**
- * Wait until `predicate` holds, polling instead of sleeping a fixed span.
- *
- * The server registers a socket inside its OWN `connection` handler, which is a
- * different event on the other side of the pipe from the client's `connect`. A
- * fixed sleep is a guess about how far apart those two land, and on a loaded
- * Windows runner the guess loses — named pipes have a longer accept path than
- * unix sockets, and the Windows leg is the slowest of the three-OS matrix. When
- * it loses, the socket is not in `connectedSockets` yet, so `rotateToken()`
- * iterates a set that does not contain it, the connection is never dropped, and
- * the test hangs until vitest kills it with a bare "timed out in 5000ms".
- *
- * Waiting for the condition itself is both sturdier and faster: no fixed cost on
- * a quick machine, and a named failure instead of an anonymous timeout on a slow
- * one.
- */
-async function waitFor(predicate: () => boolean, label: string, timeoutMs = 4000): Promise<void> {
-  const deadline = Date.now() + timeoutMs;
-  while (!predicate()) {
-    if (Date.now() > deadline) throw new Error(`timed out waiting for ${label}`);
-    await new Promise((r) => setTimeout(r, 5));
-  }
-}
 
 // Helper: generate unique pipe name for each test to avoid conflicts
 function testPipeName(suffix: string): string {
@@ -282,7 +260,7 @@ describe('DaemonPipeServer', () => {
       c1.on('connect', onConn);
       c2.on('connect', onConn);
     });
-    await waitFor(() => server.getConnectionCount() === 2, 'both sockets to register');
+    await waitFor(() => server.getConnectionCount() === 2);
     // While a client is still connected, the disconnect anchor must remain
     // untouched (otherwise the idle window would start during active use).
     expect(server.getLastDisconnectAt()).toBeNull();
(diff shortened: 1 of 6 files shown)


openwong2kim:
## [wmux-hermes] Summary

This PR replaces fixed-duration `setTimeout` sleeps with a condition-based `waitFor` helper in the daemon test suites, fixing a flaky `DaemonPipeServer` test (`rotateToken … drops connected sockets`) that timed out on `windows-latest` during the 3.35.0 release. The root cause was that `setTimeout(50)` raced ahead of the server-side socket registration on a slow runner, so `rotateToken()` iterated a set that did not yet contain the socket and the test hung until the 5000ms vitest kill.

The changes touch six files under `src/daemon/__tests__/` and `src/test-utils/`: a new shared `waitFor` helper (`src/test-utils/waitFor.ts`) with its own test suite, plus `DaemonPipeServer.test.ts`, `SessionPipe.attachSnapshot.test.ts`, `SessionPipe.clientgone.test.ts`, and `SessionPipe.reflux.test.ts` migrated to use it, removing four previously byte-identical inline copies of the same helper.

The improvement matters because the new `waitFor` resolves as soon as the predicate holds instead of paying a fixed delay on every run, and on timeout it names the condition it was waiting for (e.g. `waitFor: () => server.getConnectionCount() === 1 did not hold within 4000ms`) rather than failing with an anonymous "timed out in 5000ms".

openwong2kim:
[wmux-hermes]

This PR replaces fixed `setTimeout` sleeps in daemon pipe-server tests with a condition-based `waitFor` helper, addressing a timeout flake in `DaemonPipeServer.test.ts` where `rotateToken` iterated `connectedSockets` before the server had registered the client socket — on `windows-latest` (the slowest leg of the three-OS matrix) the 50 ms guess lost the race, the socket was never dropped, and the test hung until vitest killed it at 5000 ms with an anonymous error. The changes span `src/daemon/__tests__/DaemonPipeServer.test.ts` (six sites converted) and three `SessionPipe` test files, and a new shared helper at `src/test-utils/waitFor.ts` consolidates four byte-identical `waitFor` copies into one location whose default label is derived from the predicate's own source text so that timeout failures name the condition rather than reporting a bare "deadline exceeded". The PR also adds `src/test-utils/__tests__/waitFor.test.ts` covering immediate resolution, explicit labels, timeout message readability, and bounded label length, and reports the affected suites passing 43/43 with `DaemonPipeServer` stable across five local runs and approximately 150 ms faster from the removed fixed sleeps.

PR #626, opened 2026-07-25
docs(banner): rebuild the banner with OFL fonts instead of the macOS system faces

openwong2kim:
A licence fix for the banner that landed in #624.

That render declared `'Inter', -apple-system, …`, and Inter turned out not to be installed on the machine that produced it — so the headline was set in **San Francisco** and the mono chips in **SF Mono**. Apple's font licence covers using those faces on Apple platforms, not shipping a raster of them in a public repository.

Re-rendered with the two faces embedded at render time instead of resolved from the system:

- **Inter** — SIL OFL 1.1
- **JetBrains Mono** — SIL OFL 1.1

Both licences permit the rasterised output, so nothing further needs to travel with the image. The banner is visually identical; only the glyph source changed (451 KB, down from 459 KB).

For the record, the rest of the fonts in play:

| Where | Face | Licence |
|---|---|---|
| Landing page UI | Geist / Geist Mono via `next/font/google` | SIL OFL 1.1 |
| Arabic OG card | Almarai (committed to the site repo, OFL.txt alongside) | SIL OFL 1.1 |
| Screenshots | whatever the app renders (Cascadia Code, system UI) | screenshots of running software, not font redistribution |


openwong2kim:
## [wmux-hermes] Summary

This PR replaces `docs/banner.png` with a re-rendered version that embeds OFL-licensed fonts (**Inter** and **JetBrains Mono**, both SIL OFL 1.1) at render time instead of resolving them from the system. The previous banner, landed in #624, declared `Inter, -apple-system, …` but Inter was not installed on the rendering machine, so the headline was set in **San Francisco** and the mono chips in **SF Mono` — Apple font licences cover use on Apple platforms, not shipping a raster of those faces in a public repository.

The only changed file is the binary asset `docs/banner.png` (451 KB, down from 459 KB); the banner is visually identical, with only the glyph source changed. Both embedded font licences permit the rasterised output, so no further licence attribution needs to travel with the image. This resolves a licensing exposure introduced by #624 and keeps the repository's shipped assets on permissive terms.

[wmux-hermes]

PR #627, opened 2026-07-25
Reach your phone from the app: restore the GUI pairing path, and name what you pair

openwong2kim:
The app's `web` popover offered two ways to serve panes and neither could pair a phone.

`#616` gated device credentials on a loopback bind — a device secret never expires, so it is not handed across plain HTTP. But `WebStartArgs` only carried `allowInput` and `expose`, and the TLS front is fed by `allowedHosts`, which only the CLI could set. So the GUI's choices were loopback (an address meaningless on a phone) or expose (reaches the phone, then answers `403 insecure-transport` at redemption).

The popover advertised a fresh pairing code every ten seconds in exactly that state, because `ensurePairCode` never consulted the refusal. You read six characters onto a phone and learned it was hopeless at the last step.

## What changed

**The tailnet transport is reachable from the app.** The main process now runs the same ordered sequence `wmux web --tailscale` runs and hands the daemon the resulting `allowedHosts`. `tailscale` stops at the main process — it owns the serve registration — so it is an IPC field, not an RPC one.

**That sequence lives in one place now.** It used to sit bare inside `handleWeb`, where only the CLI could reach it. The steps a second copy drops are the two rollbacks, the ones that only run on failure paths nobody exercises by hand: a serve registered in front of a server that then failed to start is a permanent 502, and a stop that removes the front before confirming the server is gone takes the front door off a running server. `startWebTransport` / `stopWebTransport` own the ordering and take the server start/stop as a callback, so a caller cannot forget the cleanup.

**Refusals are shown instead of a code.** `status` reports `pairRefusal: { reason, detail }` and withholds `pairCode` when it fires. The reason is what the UI switches on; the detail is operator prose for logs. It mirrors the shape `/api/pair` already answers a refused redemption with.

**A vanished front is caught before a human uses it.** The daemon replays a persisted `allowedHosts` across a restart and cannot tell the serve behind it is gone. Verification runs on deliberate moments — popover open, toggle on, new code — and never on the 10s poll, which would spawn Tailscale six times a minute for a fact that changes only when someone acts.

**Devices are named before they are paired.** The daemon already required this; the GUI simply never called `pairStart`, showing the nameless code `start()` mints instead. The live roster here was five "Unnamed device" rows out of eight. The name is taken on the desktop, where there is a keyboard.

## Verification

Dogfooded end to end on a real tailnet from a packaged build: `/pair` answers 200, a wrong code is refused with the attempt budget decremented, a correct one mints a device credential, that credential lists sessions, and reusing the code reports expired. A phone paired successfully, and the roster now carries its name.

`tsc --noEmit` clean; 455 tests across the touched suites.

Two things worth reviewing closely:

- The front check is gated on `allowedHosts` being non-empty rather than on the `tailscale` flag. The first version used the flag and dogfooding caught it immediately: a `web-state.json` written before that field existed replays its hosts and parses as `tailscale: false`, so the dead address sailed straight past the check. There is a regression test for that exact shape.
- `registerWebHandlers` takes an exec seam. Writing the tailnet test without one registered a real `tailscale serve` on my machine, pointing at a port nothing was listening on.

## Not in this PR

Tailscale onboarding for users who do not have it installed — the feature is a quiet no-op for them today, which is worth fixing but is its own change. QR pairing is stacked on top of this branch.

## Summary by CodeRabbit

* **New Features**
  * Added Tailnet HTTPS transport mode, including front verification during status refresh.
  * Pairing improvements: name-first pairing and “pairing as …” support.
* **Bug Fixes**
  * Clear pairing refusal when pairing isn’t possible (e.g., insecure transport or missing front).
  * Prevent stale HTTPS/allowed-host info from persisting after the front disappears.
  * Improved rollback behavior when starting transport fails.
  * Persist Tailnet transport preference across restarts.
* **Documentation**
  * Updated the changelog with the above behavior changes.
* **Tests**
  * Expanded coverage for transport start/stop sequencing and UI pairing states.


openwong2kim:
@codex review

openwong2kim:
[wmux-hermes]
This PR restores the GUI web popover's ability to pair a phone, which was previously broken because the two transports it offered — loopback (unreachable from a phone) and network-expose over plain HTTP (refused for pairing since a device credential never expires) — could never redeem a code, even though the popover advertised a fresh one every ten seconds in exactly that state. It adds an "HTTPS over my tailnet" option to the popover that registers the same `tailscale serve` front the CLI uses, extracts the start/stop ordering (register front before server start, read port before stop, only remove our own serve config, roll back on failure) into shared `startWebTransport`/`stopWebTransport` functions so the CLI and GUI run an identical sequence, and surfaces a `pairRefusal` state that explains why pairing cannot succeed instead of showing a code that would only be answered with a 403. The PR additionally requires naming a device before minting its code, addressing a roster that was filling with indistinguishable "Unnamed device" entries; verifies a persisted tailnet front still exists at deliberate moments (popover open, toggle on, code mint) rather than on the background 10-second poll; and enforces mutual exclusivity between the tailnet and expose transports. The diff spans the CLI web command and tailscale module, the daemon web server and state store, the main-process web handler, the preload bridge, the shared web types/constants, and the renderer popover with comprehensive tests for ordering regressions and refusal rendering.

openwong2kim:
All three addressed in `492a0ba0`.

**CLI transport flag** — correct, and it was the worse half of the bug I had already fixed on the GUI side. `wmux web --tailscale` is where a phone actually gets paired from, so persisting `tailscale: false` there meant the restart-drops-you-to-loopback case survived in the more common path. Forwarded now.

**`error` reuse** — right, and chasing it turned up something worse: the pairStart failure was rendered nowhere. A refused mint left the panel looking untouched, with no code and no reason. It has its own `pairStartError` field now (always alongside `running: true`, so the offline check stays honest) and the popover shows it.

**Vacuous assert** — correct. I went further than the suggested string swap: asserting `not.toContain('482913')` against a fixture that blanks `pairCode` would still only prove a missing code renders nothing. The case now keeps the code LIVE and asserts the refusal hides it, which is the actual invariant — `start()` always mints a code, so a plaintext-bound server is always in exactly that state.

tsc clean, 459 tests across the touched suites.

PR #628, opened 2026-07-25
Pair a phone by scanning, with nothing to type

openwong2kim:
Stacked on #627 — that PR makes the address a phone can reach real; this one removes the last thing you have to type.

The QR carries the address **and** the code, so a scan finishes the pairing. A QR that only opened the pairing screen and still wanted six characters typed would have removed the easy half: the address was always the miserable part on a phone, because a raw tailnet hostname gets no autocomplete and no search suggestion.

## On putting a code in a URL

This repo forbids durable credentials in query strings and built `StreamTicket` to satisfy that rule — "the narrow thing a URL can safely carry: it grants opening a stream, it expires in two minutes, it is bound to one device." A pairing code is strictly narrower: single use, ten minutes, five attempts. This applies the rule rather than carving an exception out of it, and the reasoning is written where the payload is built rather than left to a reviewer to reconstruct.

`Referrer-Policy: no-referrer` is already set, so the address bar was the last write-down surface — `/pair` strips the code via `replaceState` before anything else can fail. The camera app's own scan history cannot be reached from here; the code being single-use and short-lived is what bounds that, not the strip.

## Shape

Encoding lives in `qrPath.ts` and returns one SVG path. Pure, so it is tested without a DOM — the popover's whole test strategy is `renderToStaticMarkup` against controlled props, and a component that encoded its own QR would need a hook and break that. The parent memoises on the payload string: the popover re-renders every 10s from the status poll while the thing being encoded changes only when a human mints a code. One path rather than ~1000 `<rect>` elements in a 288px box.

`pairQuery.js` is split out of `app.js` for the same reason `attentionFormat.js` was — the parsing decides whether a scanned code is auto-submitted or lands on the manual form, and it is tested against the exact bytes the phone runs.

## Verification

A wrong QR does not look wrong, it just fails to scan, so the tests assert what a scanner keys on: three finder patterns, module count, and the path compared cell by cell against an independently encoded grid.

Real-browser CSP check on the live tailnet: zero violations, terminal page renders, manual form still there, and the scanned path strips the code from the URL. The served header carries four script hashes. The owner confirmed QR login works from their phone.

The build refused this change at first — `build-daemon-web.mjs` pins the inline `<script>` count and says to re-read the gate rather than guess. Raised 3 to 4 deliberately; the policy is derived from the served bytes, so the new block is hashed like the others.

## Behaviour preserved

Junk in `?code=` lands on the manual form instead of auto-submitting — five bad attempts lock a code out, and a half-copied link should not spend them. A `/pair` with no query string is byte-for-byte the path it always was, which is what every typed-by-hand address still takes.


openwong2kim:
@codex review

openwong2kim:
[wmux-hermes]
This PR adds QR code scanning to complete phone pairing with a single camera scan, encoding both the pairing address and the code into one QR so the phone operator types nothing on a raw tailnet hostname that offers no autocomplete. The change touches the renderer's web popover (`WebToggle.tsx`) with a memoized QR render, a new pure `qrPath.ts` encoder that reduces the matrix to a single SVG path string, the daemon web frontend (`app.js` plus a new `pairQuery.js`) which reads the scanned code and strips it from the address bar via `history.replaceState` on load, the build script's inline-script-count gate (raised from 3 to 4 deliberately), and the `en.ts` i18n strings. It introduces the `qrcode-generator` runtime dependency and ships structural tests asserting finder patterns, exact dark-module coverage against an independent encode, and URL-parsing edge cases (junk, malformed input, missing code). The pairing code carried in the URL is framed as consistent with the repo's rule against durable credentials in query strings, since it is single-use, ten-minute, and five-attempt limited — strictly narrower than the existing `StreamTicket` the same rule produced.

PR #634, opened 2026-07-26
fix(web): durably revoke state on stop

### LINKED ISSUE #620: wmux web --stop can report success while leaving the server enabled, and web-state.json is unhardened on Windows
Two problems in `web-state.json` handling, both on `main`. They came out of a review pass on #616 and belong to #596/#599's code rather than that branch, so they are split out here.

## 1. `wmux web --stop` can report success and leave the server enabled

`clearWebState` swallows every `unlink` failure:

https://github.com/openwong2kim/wmux/blob/c99143bf9254776ea63ff4a42b0b93d5cec6df80/src/daemon/web/webStateStore.ts#L161-L167

The comment is right about `ENOENT` — already gone is the end state we wanted. It is wrong about `EPERM` and `EBUSY`, which is what an AV scanner holding the file, or a permissions problem, actually produces on Windows. In that case:

1. `daemon.web.stop` calls `clearWebState`, which returns `void` and cannot fail.
2. The RPC succeeds. The operator is told the server is stopped and its token revoked.
3. `web-state.json` survives with `enabled: true` and the old token still in it.
4. The next daemon boot hits `restoreWebServer`, which gates only on `state.enabled`:

https://github.com/openwong2kim/wmux/blob/c99143bf9254776ea63ff4a42b0b93d5cec6df80/src/daemon/index.ts#L102-L104

So the server the operator explicitly stopped comes back serving, on the token they believed was revoked, and nothing ever said otherwise. Stop is the one operation whose whole point is "do not bring this back", so failing open here is the wrong direction.

Suggested fix: have `clearWebState` distinguish `ENOENT` from a real failure and report it, and have `daemon.web.stop` either fail or write a durable `enabled: false` before acknowledging. Writing the disabled state is probably better than failing — it works even when the unlink cannot, and the file is small.

## 2. On Windows the state file gets whatever ACL the directory hands it

`saveWebState` writes a temp file at mode `0o600` and renames it over the target:

https://github.com/openwong2kim/wmux/blob/c99143bf9254776ea63ff4a42b0b93d5cec6df80/src/daemon/web/webStateStore.ts#L136-L146

The comment already notes that `mode` is a no-op on Windows, and it also says this file "is as sensitive as daemon-auth-token" — but `daemon-auth-token` gets `secureWriteTokenFile`, and `devices.json` gets that plus `scheduleTokenFileReHarden`. This one gets neither, so the renamed inode keeps the temp file's inherited DACL. On a machine whose profile or data directory carries a broad inherited ACE, another local account can read a long-lived bearer token for a server that may be reachable over the tailnet.

Worth noting the contrast with `devices.json`, which deliberately defers its hardening because the file holds only salts and scrypt output — nothing usable. `web-state.json` holds the token itself, so the same reasoning does not carry over.

Suggested fix: route this write through the same helper the other two use.

## Not affected

`devices.json` (`DeviceStore.persist`) already hardens on create and re-hardens after each rename, and `approvals.json` holds no secret.

snowyukitty:
## Summary

Fix the two fail-open paths in #620: a durable web stop no longer acknowledges success unless persisted state is revoked, and reusable web tokens are written only to a hardened inode.

## Changes

- Persist disabled web state before stopping the live listener, report durable revocation failures precisely, and reconcile the actual listener state across the CLI, GUI, and Tailscale front.
- Harden a credential-free placeholder before populating its token through an existing-only file descriptor, repair POSIX mode `0600`, and fail closed if the hardened inode changes or disappears.
- Avoid rewriting identical state during daemon restore; repair existing permissions asynchronously so cold start does not block the daemon event loop.
- Add regression coverage for filesystem failure sequences, the daemon RPC seam, live-stop failures, GUI error rendering, CLI output, and Tailscale cleanup.

### Fixed

- **A successful `wmux web --stop` now means the listener is off now and stays off after a daemon restart.** If Windows refuses to delete `web-state.json`, wmux securely replaces it with a disabled record that contains no bearer token; if neither operation succeeds, it still stops the live server but reports that persisted state could not be revoked instead of acknowledging a durable stop. The popover shows that real error instead of claiming the daemon is offline, and a Tailscale front is still removed once a fresh status check confirms the listener did stop. A reusable web token is now written only after its inode has been synchronously hardened, POSIX overwrites repair mode `0600`, and boot restore no longer rewrites an identical record merely to re-harden it — existing permissions are repaired asynchronously without freezing the daemon event loop. (#620)

## Stability tier impact

n/a — no stable RPC method, event envelope, permission contract, or Named Pipe contract changed.

## Test plan

- Affected unit and component tests: 6 files, 163 tests passed.
- `webStateStore.runtime.test.ts`: 19 tests passed.
- Bundled Windows daemon restart/stop coverage: 4 tests passed with no skips, including durable-stop write failure after the live listener is down.
- `npm run build:daemon`
- `npm run build:cli`
- ESLint on every changed TypeScript file; existing `src/daemon/index.ts` baseline remains 21 problems, with zero findings in the changed regions.
- `npm run test:parallel`: 605/609 files passed. Three unrelated Git/worktree test files hit the existing Windows 5-second timeout followed by EBUSY cleanup failures; rerunning those files with one worker passed 53 tests with 1 skipped.

## Related issues / PRs

Fixes #620

## Reviewer checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `wmux web --stop` durability and correctness across daemon restarts by verifying shutdown via a fresh status check.
  * Prevented web access from reappearing after stop failures and ensured persisted web state is revoked/neutralized when possible.
  * Updated JSON/exit behavior to reflect “listener removed” accurately, and fixed front-removal only after confirmed shutdown.
  * Strengthened token-file permission hardening (including POSIX `0600`) and improved related error reporting in the UI popover.

* **Tests**
  * Expanded runtime and CLI coverage for durable stop, restart persistence, and hardened revoke semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to wmux are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **A successful `wmux web --stop` now means the listener is off now and stays off after a daemon restart.** If Windows refuses to delete `web-state.json`, wmux securely replaces it with a disabled record that contains no bearer token; if neither operation succeeds, it still stops the live server but reports that persisted state could not be revoked instead of acknowledging a durable stop. The popover shows that real error instead of claiming the daemon is offline, and a Tailscale front is still removed once a fresh status check confirms the listener did stop. A reusable web token is now written only after its inode has been synchronously hardened, POSIX overwrites repair mode `0600`, and boot restore no longer rewrites an identical record merely to re-harden it — existing permissions are repaired asynchronously without freezing the daemon event loop. (#620)
+
 ## [3.36.0] — 2026-07-26
 
 ### Added
--- a/src/cli/__tests__/tailscale.test.ts
+++ b/src/cli/__tests__/tailscale.test.ts
@@ -698,6 +698,32 @@ describe('startWebTransport / stopWebTransport ordering', () => {
     expect(out.reason).toBe('skipped');
   });
 
+  it('removes the front when a failed reply still confirms the listener stopped', async () => {
+    const events: string[] = [];
+    const exec: TailscaleExec = async (_cmd, args) => {
+      if (args[0] === 'serve' && args[1] === 'status') {
+        events.push('tailscale:serve-status');
+        return { stdout: serveOurs(), stderr: '' };
+      }
+      events.push('tailscale:serve-off');
+      return { stdout: '', stderr: '' };
+    };
+    const out = await stopWebTransport({
+      exec,
+      readPort: async () => WEB_PORT,
+      stopServer: async () => ({
+        failed: true,
+        liveStopped: true,
+        value: 'durable revoke failed',
+      }),
+    });
+
+    expect(events).toEqual(['tailscale:serve-status', 'tailscale:serve-off']);
+    expect(out.value).toBe('durable revoke failed');
+    expect(out.serveRemoved).toBe(true);
+    expect(out.reason).toBe('removed');
+  });
+
   it('leaves a foreign serve config alone on stop', async () => {
     const foreign = JSON.stringify({
       TCP: { '443': { HTTPS: true } },
--- a/src/cli/__tests__/web.test.ts
+++ b/src/cli/__tests__/web.test.ts
@@ -1,5 +1,8 @@
 import { describe, it, expect } from 'vitest';
-import { deviceCredentialWarning } from '../commands/web';
+import {
+  annotateWebStopJsonResponse,
+  deviceCredentialWarning,
+} from '../commands/web';
 
 /**
  * `wmux web --status` and the per-device credential posture (M3).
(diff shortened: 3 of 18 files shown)


openwong2kim:
[wmux-hermes] Triage summary for **PR #634 — fix(web): durably revoke state on stop**.

**What it is.** This closes #620 by eliminating the two fail-open paths in `web-state.json` handling. It introduces a new `stopWebServerDurably` coordinator that persists the disabled intent *before* stopping the live listener, so a process death mid-stop can never leave an enabled record that resurrects the server on the next daemon boot. When neither deletion nor a disabled overwrite succeeds, the RPC now rejects with a precise error instead of acknowledging a stop that did not durably happen.

**What it touches.** Eighteen files across the daemon (`webStateStore.ts` rewrites its save path to a harden-placeholder-then-populate strategy using `secureWriteTokenFile`, plus a new `webStop.ts`), the CLI (`web.ts` annotates stop responses with front-removal state even on failure, `tailscale.ts` reconciles the Tailscale front against a confirmed live-stopped status), the GUI handler and `WebToggle.tsx` (popover now surfaces the real control error rather than a generic "daemon offline"), and the shared security layer (`secureWriteTokenFile` now also repairs POSIX `0600` on overwrites, not just on inode creation). The inode-populate step opens with `r+` and verifies file identity via `fstat`/`stat`, so an external deletion in the harden→populate window fails the save rather than silently recreating an unhardened token file.

**Why it matters.** Stop is the one operation whose entire contract is "do not bring this back," so failing open there is the wrong direction — and the unhardened `web-state.json` held the long-lived bearer token on whatever ACL the temp file inherited, a contrast with `daemon-auth-token` and `devices.json` which were already hardened. The implementation is fail-closed by construction (a credential is never written to an inode that could not be hardened first) and the test plan is thorough: 6 files / 162 tests plus 18 in `webStateStore.runtime.test.ts`, runtime daemon restart coverage including a same-name-directory write-failure fixture, and regression cases for `EPERM`/`EBUSY` deletion, placeholder disappearance, and dual-failure reporting.

**Priority: P2.** This is a security-correctness fix for an important issue and is ready for review; flagged P2 rather than P1 because there is no remote path and the prior behavior required an explicit operator action to expose. One thing worth a glance during review: `scheduleTokenFileReHarden` runs asynchronously on daemon restore while every *new* token-bearing write stays synchronous — that asymmetry is intentional (documented as avoiding a multi-second Windows shell-out on the control pipe at boot) but worth confirming it matches the threat model for an already-exposed existing credential.

openwong2kim:
Merged — thank you for this one, @snowyukitty. 🙏

This closes both fail-open paths in #620 cleanly: durable stop no longer acknowledges success unless persisted state is actually revoked, and the reusable token now only lands on a hardened inode. The placeholder→harden→populate design with the `r+` inode-identity + content read-back checks is a genuinely careful approach, and the regression coverage across the FS-failure sequences, the daemon RPC seam, and the CLI/GUI/Tailscale surfaces is excellent.

It went through a three-model review panel given it touches the auth/token surface. The one flagged concern — a claim that `statSync(path).ino` is always 0 on Windows and would break restore — was empirically refuted by this PR's own green windows-latest CI, which runs the new `webStateStore.runtime.test.ts` through the real `populateExistingWebState` path. Modern libuv populates the file id via `GetFileInformationByHandle`, so `fstat`/`stat` ino match.

A few optional hardening ideas came out of the panel (O_EXCL+rename to erase the residual same-user window, fsync-on-revoke for power-loss durability, start/stop RPC serialization). All non-blocking and tracked on our side — nothing needed from you. Thanks again for the thorough work and test plan!

PR #635, opened 2026-07-26
web: photo upload route for the phone client (--allow-upload)

openwong2kim:
A phone has a camera and a desk does not. There was no way to get a whiteboard, an error on a screen, or a broken part in front of an agent without walking it through the Mac by hand — `POST /api/input` writes to a PTY and an image cannot ride it.

So: `POST /api/upload` takes raw JPEG/PNG bytes, writes them to disk, and answers `201 {path, expiresAt}`. The **client** puts that path in a composer draft; the operator sends it through the existing input route. Nothing here types anything at anyone.

## The three decisions

**The grant is its own flag.** `--allow-upload`, not a rider on `--allow-input`. Typing into a pane the operator is watching is a smaller thing than writing a file into their home directory, so one must never imply the other. Without it: `403 {error: 'uploads-disabled: server started without --allow-upload'}` — an exact string, because the iOS client prefix-matches the `uploads-disabled:` tag to show "ask your Mac to enable uploads" instead of an HTTP code. `allowUpload` is exposed on `GET /api/config` and on `status()`, and a daemon predating this route has no such key, which the client reads as false.

`allowUpload` is **required** on `WebTerminalStartOptions` rather than optional, so a future start site cannot inherit "uploads on" by forgetting the field. Persisted state coerces with `=== true`, so an old `web-state.json` restores a server that cannot write.

**The bytes decide the format.** JPEG (`FF D8 FF`) and PNG (8-byte signature) only, sniffed from the body; `Content-Type` is ignored entirely. It is a claim by a client about a file the client did not necessarily produce, and the one thing that must not be client-controlled is the extension we write to disk. Anything else, including an empty body, is 415. 10 MB cap → 413 + `req.destroy()` (same trade-off `/api/input` makes).

**The name is ours.** `photo-<ISO timestamp with ":" and "." replaced by "-">-<8 hex>.<jpg|png>`, `0600`, `wx` so a collision is a 500 rather than a silently replaced photo. A client-supplied filename is a traversal primitive and buys nothing — nothing reads these by name.

## Expiry

24h TTL, swept **before each write and once at start** — no timer. A sweeper firing while the operator is reading a path would be a race we invented for ourselves, and the contract says plainly that the path is a consumable. The sweep only unlinks names matching `/^photo-.*\.(jpg|png)$/`: `~/.wmux/uploads` is also where an operator stages files for `browser_file_upload`, and those are not ours to delete however old they are.

`uploadsDir` is an injected dep, optional like `approvals` — a server without one answers `503 uploads-unavailable`, which is distinct from 403 on purpose (the operator *did* grant the permission; we simply have nowhere to put the bytes).

## Tests

Nine new cases in `WebTerminalServer.test.ts`, real fetches against an ephemeral bind and a real temp directory — the route's whole job is the file it leaves behind, so the filename pattern, the absolute path and the sweep are only observable on disk. 401 unauthenticated · 403 with `--allow-input` but not `--allow-upload` (the grants are separate, not merely spelled differently) · 413 over cap · 415 for text, empty, and a truncated signature · 201 JPEG (absolute path, name pattern, byte identity, `expiresAt`) · 201 PNG announced as `image/jpeg`, stored `.png` · TTL sweep at +25h with a `keep.txt` surviving · `allowUpload` on `/api/config` · 503 with no `uploadsDir`.

Full suite green: 8641 passed, 28 skipped. `tsc --noEmit` clean after every commit.

Docs: new §8 in `docs/phone-client-contract.md` (Push → §7, later sections shift down one), CHANGELOG under `[Unreleased]`. No version bump.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in photo uploads from a paired phone via `wmux web --allow-upload`.
  * Supports JPEG/PNG up to 10 MB, with server-generated paths returned for the phone composer.
  * Upload retention: photos are automatically deleted after 24 hours.
* **Documentation**
  * Updated CLI help and phone-client API contract to describe the new `allowUpload` capability and the photo upload endpoint behavior, including permission gating and detailed error handling.

## Follow-up: bounding the sum (review round 2)

The first cut bounded one request and not the total. Three fixes:

- **Disk quota.** After the sweep, at most 100 files / 200 MB of the route's own output; over it, `507 {error:'uploads-full: quota exceeded, try again later'}`. 507 rather than 403 — the permission is there and the request is fine, the server has no room, and the TTL will make some.
- **Concurrency bound.** Each in-flight upload holds its body in RAM, so the disk quota does nothing for heap. Four may buffer at once, checked **before** the body is read (refusing afterwards costs exactly what the bound exists to avoid); a fifth gets `429 {error:'too-many-uploads: try again in a moment'}`. The slot is released on every terminal path and, as a backstop, on the request's `close` — release is idempotent, because a leaked counter would lock the route out until a daemon restart.
- **Narrower sweep.** `photo-*.jpg` was a deletion predicate loose enough to eat an operator's `photo-vacation.jpg`. Now every segment is anchored: `/^photo-\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}-\d{3}Z-[0-9a-f]{8}\.(jpg|png)$/` — the exact shape `handleUpload` writes. The same pattern decides the quota count: files that are not ours to delete are not ours to charge for either.

Four more tests (quota 507, quota ignores foreign files, 429 with a held-open request that then completes 201 to prove the slot comes back, sweep spares `photo-vacation.jpg` and an uppercase-hex lookalike). Limits are injectable for tests only — the honest way to test a 200 MB ceiling is to fill it, and that is not a test. Docs §8 and the changelog entry updated.


openwong2kim:
@codex review

openwong2kim:
[wmux-hermes] Triage summary

**What it is:** a new opt-in `POST /api/upload` route that lets a paired phone push a JPEG/PNG (up to 10 MB) onto disk under `~/.wmux/uploads/phone/` and returns the absolute path plus a 24-hour expiry. The phone then drops that path into the composer for the operator to send through the existing `/api/input` channel — the route never types anything on its own. It is wired through the CLI flag (`--allow-upload`), the daemon RPC and restore path, `web-state.json`, `GET /api/config`, and `wmux web --status`.

**What it touches:** 9 files, +497/−32. The core logic lives in `WebTerminalServer.handleUpload` plus two helpers (`sniffImageExt`, `pruneUploads`); the rest is plumbing the new `allowUpload` grant through `src/daemon/index.ts`, `src/cli/commands/web.ts`, `src/daemon/web/webStateStore.ts`, and the phone-client contract doc (new §8). `allowUpload` is a REQUIRED field on `WebTerminalStartOptions` (not optional), persisted state coerces with `=== true`, and a daemon predating the route restores with uploads off.

**Why it matters / design review:** the security posture is deliberately careful and worth calling out, because this is the first authenticated route that writes arbitrary bytes into the operator home directory. The three refusal axes are sound: (1) the grant is its own flag and never implied by `--allow-input`; (2) the format is decided by magic bytes, ignoring `Content-Type` entirely, so a client cannot lie its way to a `.jpg` extension on a PNG (or anything else); (3) the filename is server-generated (`photo-<ts>-<8 hex>.<ext>`, `0o600`, `wx` flag), which closes the path-traversal primitive that a client-supplied name would open. The 10 MB cap destroys the socket like `/api/input` does, and the TTL sweep is gated on the `/^photo-.*\.(jpg|png)$/` name pattern so operator-staged files in the same directory survive. `503 uploads-unavailable` correctly distinguishes "no grant" (403) from "grant present, nowhere to write".

**Status:** mergeable, all CI green (windows-latest, macos-14, ubuntu-22.04 baselines + validate + perf bench), CodeRabbit summary attached. Nine new tests cover the 401/403/413/415/201 paths, the magic-byte-over-header case, the TTL sweep with a `keep.txt` survivor, the `/api/config` exposure, and the 503 path. No blocking concerns; assigning P3 as a standard additive feature with no contract impact flagged by the author. The one thing to keep an eye on during merge: confirm the uploads directory is created under the same hardened-write discipline the rest of `~/.wmux` uses, since `writeFileSync(full, body, { mode: 0o600, flag: 'wx' })` relies on inherited ACLs on Windows (same gap called out in #620).

Issue #615, opened 2026-07-25
security: audit 2026-07-25 follow-ups (F2-F6)


openwong2kim:
The 2026-07-25 audit turned up six findings. F1 (CDP remote-debugging port) got
its own issue — #613 — because it is the one worth fixing first. The other five
are small enough that a separate issue each would just be noise, so they are
collected here.

None of them is a fresh hole in the `docs/SECURITY.md §3` sense. F2 is a
consistency gap that sits inside the same-user ceiling we already accept. F4 and
F5 are CI hardening. F3 and F6 are optional. Listed in the order I would do them.

Full analysis: [`docs/security-audit-2026-07-25.md`](https://github.com/openwong2kim/wmux/blob/main/docs/security-audit-2026-07-25.md)

---

### F2 — `git` / `gh` execFile calls accept unconfined `repoPath` (MEDIUM, ~2h)

- [x] Confine renderer-supplied paths before they reach `git -C` / `gh` (done in #637)

**Where:**
- `src/main/ipc/handlers/worktree.handler.ts:494, 503, 516, 528, 542, 554, 566`
- `src/main/ipc/handlers/diff.handler.ts:446, 473, 490` — `diff:apply-hunks` writes files
- `src/main/ipc/handlers/toolbar.handler.ts:23` — `git:status` with renderer-supplied `cwd`
- `src/main/ipc/handlers/github.handler.ts:51, 62`

Arguments are validated as non-empty strings, then `repoPath` / `cwd` goes to
`git -C <path>` verbatim. No confinement to a project or workspace root, no
sensitive-path blocklist.

Compare `pty:create` (`pty.handler.ts:48-68`), which runs an `isAllowedShell()`
basename allow-list plus `validateCwd()`, and `fs:readFile` / `fs:writeFile`
(`fs.handler.ts:19-77`), which go through `resolveAccessiblePath()` with a
blocklist (`.ssh`, `.aws`, `.gnupg`, `.kube`, `.env`, `.npmrc`, `.netrc`,
`.wmux-auth-token`) and a `realpath` symlink re-check. The shell-spawn surface is
guarded; the more frequently called git surface is not.

Reuse `resolveAccessiblePath()`, or confine to registered workspace roots.
`diff:apply-hunks` modifies files inside the supplied directory and deserves the
strongest guard of the three.

Verification: existing worktree/diff handler tests still pass, plus rejection
cases for arbitrary paths (`~/.ssh/`, `/etc/`).

---

[...]

PR #637, opened 2026-07-26
security: confine renderer repo paths before git/gh runs in them (F2)

openwong2kim:
Closes F2 of #615.

## Problem

The worktree, diff, and GitHub-PR IPC handlers validated a renderer-supplied `repoPath` / `worktreePath` / `cwd` as a non-empty string and then handed it straight to `git -C <path>` (or `gh` in that directory). No confinement, no sensitive-path blocklist. A request could make a git command operate inside a secret directory (e.g. `~/.ssh`).

This sits inside the same-user ceiling `docs/SECURITY.md §3` already accepts. It is a consistency fix: `fs:readFile` / `fs:writeFile` (`fs.handler.ts`) and `git:status` (`toolbar.handler.ts`) already route through the `resolveAccessiblePath` guard — these three handlers did not.

## Fix

Route the path through the same `resolveAccessiblePath` guard the filesystem handlers use: canonicalize with `realpath`, reject anything that resolves into a blocked location (`~/.ssh`, `~/.aws`, `~/.gnupg`, credential stores, the auth-token files).

- **`worktree.handler.ts`** — all seven entry points (list/add/remove/merge-start/status/land/discard) funnel through `resolveToplevel(cwd)`, so one guard there covers them all.
- **`diff.handler.ts`** — confine at `DIFF_READ`, `DIFF_RESOLVE_REPO`, `DIFF_APPLY_HUNKS` (the last also writes files).
- **`github.handler.ts`** — confine at `GITHUB_PR_LIST`, `GITHUB_PR_DETAIL`.
- **`toolbar.handler.ts`** — already guarded, unchanged.

Scoped to the `-C` working directory only. `git worktree add <new-path>` passes a not-yet-existing path as a subcommand argument (`realpath` would fail on it), so it is deliberately left unconfined — that is the precise vector F2 describes.

## Tests

- New regression test in `worktree.handler.test.ts`: a valid git repo created under `<home>/.ssh` is still rejected (uses `vi.spyOn(os, 'homedir')` to relocate home deterministically, cross-platform).
- `npm run test:parallel`: 8579 passed, 0 failed.
- `tsc --noEmit`: clean. `eslint`: 0 errors.

## Scope

Additive input validation on the IPC boundary. No behavior change for legitimate repo paths (subdirectory resolution and canonicalization preserved). No dependency, schema, or config changes.

## Summary by CodeRabbit

* **Security**
  * Added stronger confinement for repository paths used by Git, GitHub, diff, and worktree actions, including canonicalization and blocking access to sensitive credential/secret directories.
  * Unsafe or inaccessible paths now fail safely instead of proceeding with Git-related operations.
* **Tests**
  * Added an end-to-end test to verify worktree handling rejects repos located under a mocked sensitive home directory subtree (for example, under `~/.ssh`).
* **Documentation**
  * Updated the Unreleased changelog with the new security behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **A successful `wmux web --stop` now means the listener is off now and stays off after a daemon restart.** If Windows refuses to delete `web-state.json`, wmux securely replaces it with a disabled record that contains no bearer token; if neither operation succeeds, it still stops the live server but reports that persisted state could not be revoked instead of acknowledging a durable stop. The popover shows that real error instead of claiming the daemon is offline, and a Tailscale front is still removed once a fresh status check confirms the listener did stop. A reusable web token is now written only after its inode has been synchronously hardened, POSIX overwrites repair mode `0600`, and boot restore no longer rewrites an identical record merely to re-harden it — existing permissions are repaired asynchronously without freezing the daemon event loop. (#620)
 
+### Security
+
+- **A repository path from the Git tab is confined before any `git`/`gh` command runs in it.** The worktree, diff, and GitHub-PR IPC handlers used to validate a renderer-supplied `repoPath`/`worktreePath`/`cwd` as a non-empty string and then hand it straight to `git -C <path>` (or `gh` in that directory). The path now passes through the same sensitive-path guard that `fs:readFile`/`fs:writeFile` and `git:status` already use — it is canonicalized with `realpath` and rejected if it resolves into a blocked location (`~/.ssh`, `~/.aws`, `~/.gnupg`, credential stores, the auth-token files, …), so a request cannot make a git command operate inside a secret directory. This sits inside the same-user ceiling `docs/SECURITY.md §3` already accepts — it is a consistency fix that brings these handlers up to the confinement the filesystem handlers had, not a new boundary. (F2, #615)
+
 ## [3.36.0] — 2026-07-26
 
 ### Added
--- a/src/main/ipc/handlers/__tests__/worktree.handler.test.ts
+++ b/src/main/ipc/handlers/__tests__/worktree.handler.test.ts
@@ -180,11 +180,11 @@ describe('worktree.handler — list/add/remove 왕복', () => {
     expect(r.worktrees!.length).toBeGreaterThanOrEqual(1);
   });
 
-  it('F2 (#615) — 홈 .ssh 하위의 유효한 repo도 confine 거부', async () => {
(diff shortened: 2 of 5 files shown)


openwong2kim:
[wmux-hermes] Triage summary

**What it is.** Security consistency fix landing finding F2 of #615: the worktree, diff, and GitHub-PR IPC handlers now route renderer-supplied `repoPath` / `worktreePath` / `cwd` through the existing `resolveAccessiblePath` guard before that path reaches `git -C <path>` or `gh` in that directory. Previously these handlers validated the path as a non-empty string and passed it through verbatim, so a request could make git operate inside a sensitive location such as `~/.ssh`. The filesystem handlers (`fs:readFile` / `fs:writeFile`) and `git:status` already had this guard; this PR closes the three handlers that did not.

**What it touches.** `worktree.handler.ts` adds a single choke point in `resolveToplevel()` that covers all seven worktree entry points (list/add/remove/merge-start/status/land/discard). `diff.handler.ts` confines paths at three call sites: `DIFF_READ`, `DIFF_RESOLVE_REPO`, and `DIFF_APPLY_HUNKS` (which also writes files). `github.handler.ts` confines `GITHUB_PR_LIST` and `GITHUB_PR_DETAIL`. The guard canonicalizes via `realpath` and rejects anything resolving into the existing sensitive-path blocklist (`~/.ssh`, `~/.aws`, `~/.gnupg`, credential stores, auth-token files). A new regression test in `worktree.handler.test.ts` deterministically relocates home via `vi.spyOn(os, "homedir")` and asserts that a valid git repo created under `<home>/.ssh` is rejected. The CHANGELOG gains an Unreleased Security entry.

**Scope and limits — worth noting on review.** The confinement is scoped to the `-C` working directory only. `git worktree add <new-path>` deliberately stays unconfined because the path does not exist yet and `realpath` would fail on it — the PR body calls this out as the precise vector F2 describes, so it is a known residual rather than an oversight. For `diff:apply-hunks`, confinement runs before the request body is inspected, so a well-formed but malicious body still cannot escape because the directory it would write into is already blocked. On failure the handlers fail-closed (return `{ok:false}`) rather than falling back to the raw path, which is the correct posture for a path-validation guard.

**Why it matters.** This is a consistency fix, not a new security boundary — it sits inside the same-user ceiling `docs/SECURITY.md §3` already accepts. Its value is defense-in-depth parity: the most frequently called git surface now matches the guards the shell-spawn and filesystem surfaces already had, and it ships with a regression test that pins the behavior. Priority assigned as **P2**, matching #615.

Verification noted in the PR: 8579 tests pass, `tsc --noEmit` clean, eslint clean.

PR #638, opened 2026-07-26
Recognise another agent host from config, and find out what it is calling itself

### LINKED ISSUE #636: feat(mcp): make first-party client list configurable via config.json
## Motivation

The first-party client allowlist (`FIRST_PARTY_CLIENT_NAMES` in `src/main/mcp/firstParty.ts:64`) is hardcoded to `{ claude-code, codex-mcp-client, opencode }`. Any other MCP client that connects to the bundled wmux MCP server lands at `unconfirmed` forever under enforce mode, with no path to unblock — even for the curated method set the bundled server already calls.

Adding a new first-party agent today means: edit source → rebuild → ship a release. Heavy for a one-line list change.

## What the Hermes case actually showed

*(This section revises the original framing of this issue, which was wrong on one point.)*

Hermes Agent connects through the official **Python MCP SDK** and never passes `client_info`. The SDK falls back to its own default:

```
mcp/client/session.py:18    DEFAULT_CLIENT_INFO = Implementation(name="mcp", version="0.1.0")
mcp/client/session.py:134   self._client_info = client_info or DEFAULT_CLIENT_INFO
```

So Hermes does **not** report `hermes`. Our own trust store recorded it as:

```json
"mcp": { "name": "mcp", "version": "0.1.0", "status": "unconfirmed" }
```

Two consequences that change this issue's design:

**1. `mcp` is a shared default, not an identity.** Allowlisting it would grant first-party recognition to every Python-SDK client that never set a name. And it is precisely what a user would paste: the natural way to use this feature is to read the observed name out of the trust store and put it in config — which produces the worst possible setting. A config key alone is therefore not safe to ship.

**2. There is no supported way to discover the observed name.** The rejection reads `surface.list: plugin is unconfirmed` and says nothing about what name was seen. Learning it required reading `~/.wmux/plugin-trust.json` by hand. A config key the user cannot correctly fill is half a feature.

Upstream fix in flight: [NousResearch/hermes-agent#70074](https://github.com/NousResearch/hermes-agent/pull/70074) makes Hermes report `hermes-agent`. Until that merges and ships, **no correct value exists** for the Hermes case.

## Proposed solution

Three parts, all required:

### 1. Config-driven allowlist

```json
// ~/.wmux/config.json
{
  "mcp": {
    "firstPartyClients": ["hermes-agent"]
  }
}
```

- `firstParty.ts` unions the config list with the compiled `FIRST_PARTY_CLIENT_NAMES`.
- Loaded once at boot, mirroring `resolveEnforcementMode` (`src/main/index.ts:765`). Consequence: changes need a restart — document it.
- Fail-closed: corrupt or missing config → compiled defaults only.

### 2. Generic-name denylist (hard, not overridable from config)

Reject and log names that are SDK defaults or otherwise non-identifying: `mcp` (Python SDK default), `unknown`, `client`, empty/whitespace. The trust store already carries both `mcp` and `unknown` rows, so this is observed behaviour, not a hypothetical.

### 3. Observed-name discoverability

Surface the `clientName` wmux actually saw — either in the rejection message or as a `wmux mcp clients` listing over the trust store. Without it users guess; in the Hermes case the guess (`hermes`) would have been wrong.

## Invariants preserved

The scoped method allowlist (`FIRST_PARTY_METHODS`) stays compiled — only *who* is recognized becomes configurable, never *what* they may call. The spec §2.4 argument holds: an impersonator gains recognition but stays inside the curated method set. The denylist is what keeps "recognition" from silently degrading into "any anonymous client".

## Non-goals

- **Not `mcp.mode: shadow`.** That disables enforcement globally — wrong tool, and it removes the security surface for every client.
- **Not the approval-dialog path (§4).** `wmux.*`-reserved methods (including `surface.list`, #285) are grammar-forbidden in any declaration, so first-party recognition stays the only route for the surface-lifecycle methods a supervisor agent needs.

## Scope

- `src/main/mcp/firstParty.ts` — config union + denylist
- config read helper, sibling to `enforcementMode.ts`
- `src/main/mcp/__tests__/firstParty.test.ts` — merge path, fail-closed, denylist rejection
- discoverability surface (rejection message or CLI listing)
- `docs/api/mcp-plugin-spec.md` §2.4 — config key, restart semantics, denylist rationale, and the fact that empirical name capture moves to the user on the config path

No protocol change, no RPC surface change. Main-process config plumbing plus one read-only surface.

## Priority

P2. Note the Hermes unblock is gated on NousResearch/hermes-agent#70074 landing regardless of what we ship here.

openwong2kim:
Closes #636.

## Summary

**Config-driven first-party recognition.** `FIRST_PARTY_CLIENT_NAMES` was compiled into the bundle, so recognising a new agent host meant edit source → rebuild → release. Anything outside the built-in three sits at `unconfirmed` forever under enforce mode, and the only escape that exists today is `mcp.mode: shadow`, which disables enforcement for every client at once. `mcp.firstPartyClients` in `~/.wmux/config.json` is now unioned with the compiled defaults at boot.

It changes *who* is recognised, never *what* they may call. `FIRST_PARTY_METHODS` stays compiled, so a configured host gets exactly the curated set the built-in three get, and the three existing guards are untouched: an explicit `denied` still wins, a failed trust read still declines, a method outside the set still falls through. Every malformed config shape resolves to no additions rather than to everybody.

**Non-identifying names are refused, and the refusal is not overridable from config.** This is the part the issue as filed did not have, and it is why the config key alone was not safe to ship.

Hermes was the trigger for #636, and the issue assumed its `clientInfo.name` was `hermes`. It is not. Hermes connects through the Python MCP SDK and never passes `client_info`, so the SDK falls back to `DEFAULT_CLIENT_INFO = Implementation(name="mcp", version="0.1.0")` (`mcp/client/session.py`, verified against mcp 1.26.0). Our own trust DB recorded it as `mcp` / `0.1.0`. Both proposals floated at the time — hardcode `hermes`, or config-add `hermes` — would have matched nothing.

That makes `mcp` a *shared* name, not a missing one: every Python-SDK client that never named itself reports it. And it is exactly the string an operator would copy out of the trust DB to unblock the client in front of them. Without a denylist, the most natural use of this feature produces its worst possible configuration. `wmux-cli` is refused for a different reason — it is deliberately allowed a *narrower* set than first-party and is checked after it, so configuring it would silently widen the CLI from 29 methods to ~80.

The gate lives in the setter, not in the config reader, so no caller — loader, test helper, or a future Settings UI — can reach recognition through a different path.

**Discoverability.** Finding the observed name previously meant reading `~/.wmux/plugin-trust.json` by hand; the rejection said only "plugin is unconfirmed". Identity rejections now echo the `clientName` wmux actually saw, and `wmux mcp clients` lists every client in the trust DB with `NOT CONFIGURABLE` marked on names that can never be added. The CLI reads the file directly rather than going through the daemon, because the moment you need it is the moment you are editing `config.json` with the app closed.

Upstream: [NousResearch/hermes-agent#70074](https://github.com/NousResearch/hermes-agent/pull/70074) adds `client_info` on the Hermes side and will make its name `hermes-agent`. Until that lands and ships there is no correct value for the Hermes case — which is why nothing here is urgent.

### Commits

| Commit | What |
|---|---|
| `feat(mcp)` | config-driven list + denylist gate + boot wiring |
| `feat(cli)` | `wmux mcp clients` |
| `feat(mcp)` | identity rejections name the observed client |
| `docs(mcp)` | spec §2.4 + CHANGELOG |

## Test Coverage

34 tests added across 4 files (2 new, 2 extended).

```
CODE PATHS                                              GUARDS
[+] src/main/mcp/firstPartyConfig.ts                    [+] Denylist (the reason this PR exists)
  └── readConfiguredFirstPartyClients()                   ├── [TESTED] `mcp` refused end-to-end through the enforcer
      ├── [TESTED] well-formed list                       ├── [TESTED] `wmux-cli` refused (CLI-widening path)
      ├── [TESTED] trims entries                          ├── [TESTED] case-insensitive
      ├── [TESTED] drops non-string / blank               └── [TESTED] good entries survive a mixed list
      ├── [TESTED] fail-closed x7 (missing, bad JSON,
      │            non-object root, no mcp, non-object   [+] Scope containment
      │            mcp, no key, non-array)                 ├── [TESTED] configured name gets the SAME curated set
      └── [TESTED] __proto__ stripped                      └── [TESTED] still rejected for daemon.shutdown / workspace.new

[+] src/main/mcp/firstParty.ts                          [+] Three original guards preserved
  └── setConfiguredFirstPartyClients()                     ├── [TESTED] explicit `denied` still wins
      ├── [TESTED] extends without touching defaults       └── [TESTED] failed trust lookup declines (fail-closed)
      ├── [TESTED] replaces, does not accumulate
      ├── [TESTED] trims / dedupes / drops blanks        [+] Discoverability
      ├── [TESTED] empty list restores defaults-only       ├── [TESTED] rejection echoes observed clientName
      └── [TESTED] exact match only (no fuzzy/prefix)      ├── [TESTED] rejection points at `wmux mcp clients`
                                                          └── [TESTED] control characters stripped from the echo
[+] src/cli/commands/mcp.ts
  └── readObservedClients()                             COVERAGE: 34/34 paths tested (100%)
      ├── [TESTED] sorts most-recently-seen first
      ├── [TESTED] flags SDK default + wmux-internal
      ├── [TESTED] falls back to record key
      ├── [TESTED] null on missing/corrupt, [] on empty
      ├── [TESTED] skips malformed entries
      └── [TESTED] __proto__ stripped
```

Three of these are the regression guards that matter: `mcp` refused end-to-end through the real enforcer, a configured name still blocked from `daemon.shutdown`/`workspace.new`, and the `denied` / failed-lookup guards intact for configured names.

## Verification

```
tsc --noEmit (tsconfig.json)       clean
tsc --noEmit (tsconfig.cli.json)   clean
eslint (all touched files)         clean
npm run test:parallel              610 files / 8613 tests passed, 1 file + 20 tests skipped
wmux mcp clients (real trust DB)   lists 4 clients; `mcp` and `unknown` correctly marked NOT CONFIGURABLE
```

`src/main/index.ts` reports 4 pre-existing eslint errors (`no-var-requires` at lines 162/163/549/881 pre-change). Confirmed unchanged by stashing this branch's edit and re-running — same set, shifted line numbers only. No new lint problems introduced.

## Notes for review

- **`NON_IDENTIFYING_CLIENT_NAMES` lives in `src/shared/rpc.ts`, not in `firstParty.ts`.** The CLI needs it to explain the refusal, and `tsconfig.cli.json` has an explicit include allowlist that does not cover `src/main/**`. Duplicating it would create two sources of truth for a security-relevant list. `WMUX_CLI_CLIENT_NAME` already lives there for the same reason.
- **The denylist is deliberately short.** Only names with evidence behind them: `mcp` (verified SDK default), `unknown` and `wmux-cli` (wmux's own, both present in real trust DBs), plus `client` / `default` as obvious non-identities. Speculative additions were left out.
- **Restart semantics.** Read once at boot, matching `mcp.mode`. Documented in spec §2.4 rather than left to be discovered.
- **No version bump** — per repo policy, CHANGELOG entry is under `## [Unreleased]`.

## Test plan

- [x] `npm run test:parallel` — 8613 passed
- [x] `tsc --noEmit` on both main and CLI configs
- [x] `eslint` on every touched file
- [x] `wmux mcp clients` exercised against the real `~/.wmux/plugin-trust.json`


openwong2kim:
[wmux-hermes] Triage summary

This PR closes #636 by making the first-party MCP client allowlist operator-extensible via `mcp.firstPartyClients` in `~/.wmux/config.json`, unioned at boot with the compiled `FIRST_PARTY_CLIENT_NAMES`. The key design insight — and the reason the issue as originally filed was incomplete — is that the config key alone is unsafe to ship: Hermes connects through the Python MCP SDK and never sets `clientInfo`, so it reports the SDK default `mcp`/`0.1.0`, which is a *shared* name every anonymous Python-SDK client inherits. Allowlisting it would grant first-party recognition to every such client, and it is precisely the string an operator would copy out of the trust DB. The PR therefore adds a hard, non-overridable denylist (`NON_IDENTIFYING_CLIENT_NAMES` in `src/shared/rpc.ts`) refusing `mcp`, `unknown`, `client`, `default`, and `wmux-cli` (the last because the CLI has a deliberately *narrower* allowlist checked after the first-party branch, so naming it would silently widen it). The gate lives in `setConfiguredFirstPartyClients` (the setter) rather than the config reader, so no caller — loader, test helper, or future Settings UI — can route around it.

Discoverability is the other half and it is done right. Identity rejections now echo the `clientName` wmux actually saw instead of the bare "plugin is unconfirmed", and a new `wmux mcp clients` subcommand lists every client in the trust DB sorted most-recently-seen-first, marking non-configurable names with a `NOT CONFIGURABLE` explanation. The CLI reads `plugin-trust.json` directly rather than going over RPC, which is the correct choice because the moment an operator needs it is when the app is closed and they are editing `config.json`. Scope is well-contained: `firstPartyConfig.ts` (new config reader, fail-closed across 7 malformed shapes with `__proto__` stripped), the setter in `firstParty.ts`, the CLI listing in `mcp.ts`, spec §2.4 documentation, and a CHANGELOG entry under `## [Unreleased]` with no version bump per repo policy. Invariants are preserved — `FIRST_PARTY_METHODS` stays compiled, so configuring *who* is recognised never widens *what* they may call, and the three existing guards (explicit `denied` wins, failed trust read declines, method outside the set falls through) are explicitly tested as intact for configured names.

Test coverage is thorough: 34 new tests across 4 files (2 new, 2 extended), with a coverage table in the PR body mapping every code path to its guard. The three regression guards that matter most are all present and tested — `mcp` refused end-to-end through the real enforcer, a configured name still blocked from `daemon.shutdown`/`workspace.new`, and the `denied`/failed-lookup guards intact — and the author reports 8613 tests passing with both tsc configs and eslint clean (4 pre-existing `no-var-requires` errors in `src/main/index.ts` confirmed unchanged by stashing). One upstream note worth keeping visible: the Hermes unblock this was triggered by is gated on NousResearch/hermes-agent#70074 landing and shipping, so nothing here is urgent on that axis.

Priority assigned: **P2**, matching the parent issue #636. It is an important security-adjacent improvement that unblocks a real class of clients without a release, but it is not a live vulnerability or regression — the denylist specifically prevents the worst-case misconfiguration, and enforce mode continues to work exactly as before for every client not explicitly added.

openwong2kim:
@codex review

PR #639, opened 2026-07-26
Run the orchestrator on Claude Code's own interactive terminal (claude-pty brain)

openwong2kim:
## What

A third orchestrator-brain option — **Settings → Orchestrator brain → "Claude Code (terminal)"** — that runs the user's own `claude` binary as a real interactive TUI in a daemon pty, embedded live in the Orchestrator tab. Hedge for the scenario where headless (`-p`/Agent SDK) sessions stop being covered by a Claude subscription: an interactive terminal session is what a subscription unambiguously pays for. The default SDK brain and the Hermes ACP brain are unchanged; the seam is the existing `BrainAdapter` interface.

## How it works

- `ClaudePtyBrainAdapter` spawns `claude` as a daemon pty session (cwd pinned to the wmux dir for `--resume` stability) and types prompts into it; the deck embeds the live terminal via the same `useTerminal` wiring panes use, with the bubble log kept below it (collapsible, sticky terminal header, scrollable history).
- No TUI parsing: structured turn signals ride Claude Code's own hooks (`Stop`/`SessionStart`) through the existing `wmux-bridge.mjs` → `hooks.signal` RPC. A claim-first bus intercepts the brain's own signals above the fleet lane — otherwise the brain's `agent.stop` would wake the brain on its own turn end, forever.
- Final assistant text is recovered from the transcript (`readLastAssistantMessage`) and emitted as one `text-delta` before `turn-end`, so the bubble log stays the durable report surface.
- Sandbox parity without `canUseTool`: a generated `--settings` profile denies `Agent/Task/Bash/Edit/MultiEdit/NotebookEdit/Write` (with a PreToolUse exit-2 backstop) and pre-approves exactly the commander MCP surface; commander trust token minted per spawn.
- Spawn env scrubs every `CLAUDE*`/`ANTHROPIC*` var plus `AI_AGENT` — an inherited `CLAUDE_CODE_CHILD_SESSION` silently disables transcript persistence, which kills both the Stop hook's `transcript_path` and `--resume` (empirically verified).

## Dogfood-hardening in this PR

Found live and fixed here: prompt/Enter written as separate pty writes with one retry (paste detection swallowed a same-chunk Enter); hook bridge executed as node, not as a GUI Electron app; bridge pipe/token paths honour the instance suffix (a dev brain's signals were being delivered to the production instance); commander preamble not re-injected into resumed conversations; ambient autonomy/policy blocks sent to the TUI brain only when their content changes; the brain excluded from its own fleet roster (both the session-listing and per-ptyId metadata lanes, closing a fail-open filter); dialog deadlocks hand the turn back with a pointer at the now-typable embed; and a delegation contract + stop-event authority baked into the shared commander prompt.

## Tests

- New: `ClaudePtyBrainAdapter` unit suite (20) — settings profile snapshot, hook-signal → event mapping (exactly one `turn-end`, duplicate Stop ignored, dispose mid-iteration), env scrub, stale-resume soft-fail; `deck.handler` vendor plumbing (5); fleet exclusion selectors.
- All four tsc projects clean; targeted vitest sweep 1520 passed post-rebase.

## Out of scope (follow-ups)

Live tool chips for the TUI brain, `limit` events, approval-dialog detection for non-Claude worker panes, and infra-enforced (rather than prompt-enforced) dispatch contracts.

## Summary by CodeRabbit

* **New Features**
  * Added an option to run the orchestrator in an interactive Claude Code terminal.
  * Added embedded terminal views for orchestrator sessions, with workspace-specific sessions.
  * Added phone photo uploads to panes via `wmux web --allow-upload`.
  * Added support for preserving and displaying critical approval risk indicators.

* **Bug Fixes**
  * Prevented shutdown while approvals are pending.
  * Fixed permission-prompt deadlocks and stale workspace listings.
  * Improved instance isolation for authentication and terminal connections.

* **Documentation**
  * Updated security and usage documentation for relay encryption and new orchestrator capabilities.


openwong2kim:
[wmux-hermes] Triage summary

This PR adds a third orchestrator brain option (`claude-pty`) that runs the user's own `claude` binary as a real interactive TUI inside a daemon pty session, embedded live in the Orchestrator tab — a hedge against the headless `-p`/Agent SDK path one day falling outside Claude subscription coverage. The design is deliberate about not parsing the TUI screen: all structured turn signals (turn boundaries, session id, final assistant text) ride Claude Code's own hooks through the existing `wmux-bridge.mjs`, with the embedded terminal itself serving as the conversation view rather than re-synthesising streamed bubbles. Scope is broad but coherent — the new `ClaudePtyBrainAdapter.ts` (833 lines) is the centrepiece, supported by a claim-first hook bus to stop the brain from waking itself on its own `agent.stop`, env scrubbing to prevent `CLAUDE_CODE_CHILD_SESSION` from silently disabling transcript persistence, and a generated `--settings` profile that denies `Bash/Edit/Write/Agent/Task` twice over (allow-list plus a `PreToolUse` exit-2 backstop) while pre-approving exactly the commander MCP surface.

The PR also lands two correctness fixes found during dogfooding that are worth calling out on their own: the bundled `wmux-bridge.mjs` now honours `WMUX_DATA_SUFFIX` for its auth-token path and pipe name (without this a dev brain's signals were silently delivered to the production instance), and `Watchdog` gains a `pendingApprovals` keep-alive signal so a pending approval holds the daemon up even with zero connections — wired defensively today since `sessions` already covers the steady state, but explicitly enforcing the M4 invariant so a future phone surface cannot regress it. Test coverage is solid: 20 new `ClaudePtyBrainAdapter` unit tests (settings-profile snapshot, hook-signal → event mapping with exactly-one-`turn-end`/duplicate-Stop-ignored/dispose-mid-iteration, env scrub, stale-resume soft-fail), 5 deck-handler vendor-plumbing tests, fleet-exclusion selectors, plus new Watchdog cases for the `pendingApprovals` path; the author reports 1520 targeted vitest tests passing post-rebase and all four tsc projects clean.

Priority assigned: **P2**. It is a substantial, strategic feature rather than a regression or security fix, and it leaves the default SDK brain and the Hermes ACP brain entirely unchanged — no existing path is altered, so there is no urgency to merge ahead of review. The out-of-scope follow-ups (live tool chips for the TUI brain, `limit` events, approval-dialog detection for non-Claude workers, infra-enforced dispatch contracts) are correctly noted for separate tracking.

openwong2kim:
@codex review

PR #643, opened 2026-07-26
fix(updater): let a manual check run with auto-update switched off

openwong2kim:
## What

The auto-update toggle fed the single early-return that every update check passes through, so turning background polling off also turned the manual "check for updates" button into a no-op: the IPC handler had already replied `checking`, then `check()` bailed silently. With the toggle off, the only way to update was reinstalling.

## Fix

`check()` now bails on `!enabled` only for background polls. A manual press (`oneShot`) is an explicit user request and always checks — one line in `AutoUpdater.ts`.

## Tests

Two regression tests in `AutoUpdater.platform.test.ts`:
- toggle off → neither the 15s first check nor a 30-min interval touches the network, but a manual `UPDATE_CHECK` still hits the feed (fails without the fix)
- toggle back on → background polling resumes at the next interval

`tsc --noEmit` clean; updater suite 46/46.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the **Check for updates** button so it still performs an explicit update check when automatic updates are disabled (instead of acting like a no-op).
  * Disabling automatic updates now silences only scheduled background polling, not user-initiated checks.
  * Re-enabling automatic updates correctly resumes the next scheduled update checks.
* **Tests**
  * Added coverage to verify manual checks work while auto-update polling is gated and that polling resumes after re-enabling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to wmux are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **"Check for updates" now works with auto-update switched off.** The toggle was wired through the one gate every check passes, so turning background polling off also turned the manual button into a no-op that reported "checking" and went silent — leaving reinstalling as the only way to update. The toggle now silences background polls only; pressing the button is an explicit request and always checks.
+
 ## [3.37.0] — 2026-07-27
 
 ### Added
--- a/src/main/updater/AutoUpdater.ts
+++ b/src/main/updater/AutoUpdater.ts
@@ -121,7 +121,11 @@ export class AutoUpdater {
     // Defense in depth: never poll the update feed on an unsupported platform,
     // even if a caller invokes check() directly.
     if (!isUpdaterSupported) return;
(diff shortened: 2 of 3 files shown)


openwong2kim:
[wmux-hermes]

## Summary

This is a focused bug fix for the updater: when the auto-update toggle was switched off, the single early-return that every update check (`check()`) passed through also silently disabled the manual "Check for updates" button — the IPC handler already replied `checking`, then `check()` bailed, leaving reinstall as the only way to update. The fix widens the guard to `if (!this.enabled && !oneShot) return`, so the toggle silences background polls only while a manual press (`oneShot`) is always honored as an explicit user request.

**What it touches:** `src/main/updater/AutoUpdater.ts` (1 production line, +5/-1 incl. comment), two regression tests in `AutoUpdater.platform.test.ts` pinning both directions (toggle off → no scheduled polls but a working button; toggle back on → polling resumes at the next interval), and a CHANGELOG entry under `[Unreleased]`.

**Why it matters:** It restores the only in-app update path for anyone who runs with auto-update off — a real UX trap rather than a cosmetic issue. The change is minimal, the intent is documented inline, and the tests directly assert the regression (the manual-check case fails on the old code). No surface beyond the updater is touched, and there is no behavior change when the toggle is on.

**Priority: P2.** User-facing update-path bug with a workaround (reinstall) — important to ship soon but not a hotfix. Clean diff, well-tested; looks safe to merge after CI is green.

PR #647, opened 2026-07-27
security: widen web pairing codes to eight characters

snowyukitty:
## Summary

- replace the proxy-collapsed TCP-source cooldown with an 8-character pairing code (40 bits over the existing unambiguous base-32 alphabet)
- retain the 10-minute TTL, single-use redemption, five-attempt burn, regeneration cooldown, cross-site refusal, and timing-safe comparison
- align the browser form, scanned-QR parser, desktop QR fixtures, phone-client contract, changelog, and generation/regeneration tests with the new shape
- remove the now-unused request-host parameter from the pairing exchange

## Design notes

The original source limiter did not survive wmux's supported deployment topology: successful remote pairing is normally fronted by Tailscale Serve or another loopback reverse proxy, so every client reaches Node through the same proxy source. The one mode that preserves client IPs (`--expose`) is also the mode where `mintRefusal()` deliberately prevents durable credential minting. A source cooldown was therefore a shared lock where it mattered and ineffective where it was identifiable.

The browser QR parser now accepts the exact server alphabet and refuses impossible `0/O/1/I` candidates locally, so a malformed scanned URL cannot spend an attempt.

The desktop wrong-attempt signal and a Tailscale identity gate remain separate follow-ups: both are useful, but each has a distinct trust boundary and review surface.

#648 introduced phone protocol v1 under `[Unreleased]`; no release has shipped that contract yet. This PR therefore defines the initial v1 pairing shape as eight characters rather than creating a v2 that never had a released v1 predecessor.

## Verification

- focused pairing/UI/QR suite: 5 files, 201/201 passed
- impossible-glyph QR parser cases: 8/8 passed
- `npx tsc --noEmit -p tsconfig.daemon.json` — passed
- `npx tsc --noEmit -p tsconfig.json` — passed
- `npm run build:daemon-web` — CSP gate passed
- touched-file ESLint — 0 errors (15 pre-existing warnings)
- `git diff --check origin/main` — passed

The default local `npm run test:parallel` completed 610/614 files and 8,675 tests; seven tests in three untouched Windows Git/registry files exceeded their 5–15 second defaults. Re-running those files with one worker and a 30-second timeout passed all 67 runnable tests (9 skipped). Repository-wide `npm run lint` remains blocked by pre-existing baseline errors, including generated `out/` artifacts; the touched-file lint above is clean.

Refs #615 (F3). F6 remains separate.

## Summary by CodeRabbit

* **Security**
  * Increased web pairing codes from **6 to 8 characters** using an unambiguous base-32 alphabet, expanding the code space to **40 bits** while keeping the same **ten-minute lifetime**, **single-use redemption**, and **attempt/burn limits**.

* **User Experience**
  * Updated the phone pairing flow to instruct and validate **8-character** entry, with matching on-screen wording for both QR and manual pairing.

* **Documentation / Tests**
  * Updated pairing documentation, QR/manual code parsing behavior expectations, and test fixtures to use the new **8-character** format.


openwong2kim:
[wmux-hermes] **Summary**

This PR implements the **F3 hardening** item from the 2026-07-25 security audit (#615): per-TCP-source rate limiting on web pairing attempts under `wmux web --expose`. After one wrong code from a given source, a 30-second cooldown returns `429` with `Retry-After` and refuses further submissions *before* the code comparison — closing both the budget-burning path (an attacker on the LAN could previously exhaust the 5-attempt global budget) and a response-code oracle (throttled correct and wrong codes return identical 429s). Source identity is taken from the TCP socket only; `Forwarded`/`X-Forwarded-For` are deliberately ignored since a direct client can forge them, and IPv4-mapped IPv6 (`::ffff:`) is normalized to its IPv4 form. Cooldowns reset on server restart and the pairing UI surfaces the wait time.

**Review notes on the diff (5 files, +266/−29)**

The core change in `WebTerminalServer.ts` is tight and well-commented: a `pairSourceBlockedUntil` map with opportunistic pruning, `pairSourceRetryAfter()`/`blockPairSource()` helpers, and the cooldown gate placed after the expiry/budget checks but before `timingSafeEquals`. The `handlePair` signature cleanly swaps `requestHost: string` (now unused for security decisions) for `source: string` derived from `normalizeRemoteAddress(req.socket.remoteAddress)`. The existing test that burned 5 attempts in a tight loop was correctly refactored to space attempts 30s apart (`spacedWrongPairAttempts`), and three new tests cover the per-source limit, forwarded-header forgery resistance, IPv4/v6 normalization, correct-code-during-cooldown (oracle closure), and restart clearing — 135/135 passing per the author. The frontend (`app.js`) adds a 4-line branch for the `'rate limited'` error, and `CHANGELOG.md` + `docs/phone-client-contract.md` document the new `429` response.

**Priority**

Labeled **P3** — defense-in-depth hardening of an already-six-ways-defended pairing path, reachable only under explicit `--expose`. The implementation is careful and matches the threat model in #615 (F3). No concerns blocking merge; the `unknown` sentinel for a missing `remoteAddress` is the only minor nit (a loopback client with no peer address would share a single `unknown` budget, which is harmless here).

openwong2kim:
Thanks for taking this on — the test work here is genuinely strong (the pre-comparison oracle ordering, the socket-level source mocking, the restart coverage), and I learned something from the oracle discipline in particular. Before we go further I want to think through the deployment side with you, because I hit something while mapping this against how wmux is actually run, and I'd like your read on it.

Here's the table that's bothering me:

| Deployment | Can a stranger reach `/api/pair`? | What `remoteAddress` sees |
|---|---|---|
| default (loopback bind) | no — same machine only | `127.0.0.1` for everyone |
| `wmux web --tailscale` (the recommended path) | tailnet members | `127.0.0.1` for everyone — `tailscale serve` fronts the loopback bind |
| `--expose` (plaintext 0.0.0.0) | the LAN | real IPs — but `mintRefusal` already refuses pairing on this bind |

If I'm reading this right, in every deployment where pairing can actually succeed, all clients collapse into the proxy's single `127.0.0.1` bucket — including a self-hosted reverse proxy (`--allow-host` + caddy/nginx), which has the same shape. That would turn the per-source cooldown into a de facto global lock: one wrong code every 30s from anyone on the tailnet keeps the bucket at 429, and the operator's phone holding the *correct* code waits with it. And on the one bind where real IPs are visible, pairing is already refused, so the limiter has nothing to protect there. Am I missing a deployment shape where the source key survives?

Two smaller things I noticed while testing that line of thought, in case they're useful either way:

- The 429 body carries the global `attemptsLeft`, and the cooldown check runs after the expiry/budget checks — so a blocked source can still poll pairing state (code live? budget left?) for free. The correct-code oracle is closed, but I think a state oracle opens.
- `pairSourceBlockedUntil` isn't cleared on code regeneration, so a source that trips once and retries under 30s stays blocked across regenerations until a restart.

If the table above holds, here's the direction I keep landing on, and I'd value your opinion before anyone writes code:

1. **Widen the code from 6 to 8 characters** (32⁸ ≈ 1.1 × 10¹²) — the other option #615 lists. With the existing 5-attempt burn and timing-safe compare, brute force stops being a conversation, and QR pairing carries the code in the URL so the common path types nothing.
2. **Surface wrong attempts instead of blocking them**: a wrong guess raises a visible "someone is guessing the pairing code (N left)" signal on the desktop. Pairing is a short, operator-supervised ceremony, so detection may buy more here than prevention built on an identity we can't trust.
3. Possibly on top, later: when `tailscale: true`, `tailscale serve` injects `Tailscale-User-Login` on the loopback forward — the one client identity in this topology that a remote caller can't forge. Gating `/api/pair` on it would remove the guessing surface for non-operators entirely. That feels like its own PR, though.

The parts of this PR I'd hate to lose are the tests — the oracle assertions and the source-mocking pattern would carry almost directly into (2), or into the identity gate. Would you be up for reworking in that direction, or do you see a way to keep the per-source approach viable that I'm not seeing? Genuinely open to either — you've clearly thought hard about this surface.

snowyukitty:
Your deployment table holds. I do not see a supported pairing topology where the TCP source key remains useful:

- the default bind is loopback-only;
- `--tailscale` terminates at Serve and forwards to the loopback listener, collapsing clients into the proxy source;
- a conventional loopback Caddy/nginx front has the same shape;
- `--expose` preserves client IPs, but `mintRefusal()` intentionally prevents credential minting on that plaintext non-loopback bind.

So the current cooldown is effectively a shared lock exactly where pairing can succeed. I also confirmed both secondary observations: the ordering plus `attemptsLeft` creates a pairing-state oracle for a blocked source, and the source map survives code refresh/regeneration. I do not think either is worth patching around when the source identity itself is wrong.

I am happy to rework this. My proposed split is:

1. Make this PR the narrow F3 fix: remove the TCP-source limiter and widen the existing base-32 code from 6 to 8 characters (30 to 40 bits), retaining the five-attempt burn, TTL, regeneration cooldown, and timing-safe comparison. I will preserve the useful generation/redemption/oracle regression coverage where it still applies.
2. Put the visible wrong-attempt signal in a follow-up. It is a good operator-supervised defense, but it crosses the daemon/main/renderer notification surface and is independently reviewable; bundling it would make this small hardening PR much wider than the issue requires.
3. Keep the Tailscale identity gate separate as suggested. Serve's identity headers are the first trustworthy per-caller identity in this topology, but tagged-device behavior (no user identity header) and the loopback-only trust boundary deserve an explicit design and tests of their own.

I will remove the limiter before resolving the #648 conflicts, so the rebase follows the agreed design rather than spending review effort preserving code we no longer want. If you would prefer the desktop signal in this PR, say so; otherwise I will proceed with the narrow split above.

openwong2kim:
The narrow split is exactly right — proceeding as you laid it out is approved: this PR stays the F3 fix (limiter removed, 8-character codes, existing burn/TTL/timing-safe compare retained), the visible wrong-attempt signal goes to a follow-up, and the Tailscale identity gate gets its own design and PR. Preserving the oracle and source-mocking test work through the rework is appreciated — that coverage is the part I most wanted to keep.

One housekeeping item: the PR is showing a merge conflict, likely from #649 landing just now — it added entries under `## [Unreleased]` in CHANGELOG.md. Conflicts there are append-merges by convention (keep both sides, order doesn't matter). Note that CI silently skips conflicted PRs here, so the green checks above are from the pre-conflict head — after the rebase a fresh run will queue automatically.

Once it's rebased and green this is ready to land from my side.

openwong2kim:
Thank U very much 🫡

PR #648, opened 2026-07-27
feat(web): phone-API protocol version handshake

openwong2kim:
## Problem

The phone API is unversioned. Once a TestFlight iOS build ships, the HTTP surface
is hard-coupled to binaries the daemon cannot update — and a mismatch would have
surfaced as routes failing one at a time, with no way for the app to say "you are
too old" rather than "something went wrong".

## What this adds

`src/daemon/web/protocolVersion.ts`:

- `PHONE_PROTOCOL_VERSION = 1` — bumped only when the phone-facing wire changes
  incompatibly (route removed/renamed, field removed or re-meaninged). Additive
  changes never bump it.
- `MIN_PHONE_PROTOCOL_VERSION = 1` — the hard cutoff; bumped only when support
  for an older shape is actually deleted, since every client below it is stranded.
- `daemonServerVersion()` — reads the same `WMUX_SPAWNED_BY_VERSION` the B′
  stale-daemon auto-replacement reads, rather than a second source that would
  eventually disagree with it (and rather than a `package.json` a bundled daemon
  does not have beside it).

`GET /api/config` now carries `protocolVersion`, `minProtocolVersion` and
`serverVersion` alongside the existing flags. No new route: `/api/config` is
already the first authenticated call a client makes, so a dedicated
`/api/version` would be a second round trip that only pre-handshake daemons
could fail — precisely the daemon the handshake exists to recognise.

## Compatibility

- **Old daemon, new client**: the three keys are simply absent. The client reads
  that as protocol 0 and behaves exactly as it does today — the same way a
  missing `allowUpload` reads as `false`.
- **Old client, new daemon**: the fields are ignored; nothing that shipped
  changes shape. `MIN_PHONE_PROTOCOL_VERSION` stays at 1, so nobody is cut off.
- **Client below the floor** (future): show an explicit "update required" state
  and stop, rather than retrying into worse errors.
- A newer `protocolVersion` than the client's is not a warning — the floor is
  what decides whether a client is still served.

`serverVersion` is display/bug-report only and may be the literal `unknown`;
clients gate on the two numbers, never on that string.

## Tests

Two new cases in `src/daemon/web/__tests__/WebTerminalServer.test.ts`:

- `/api/config` carries all three fields with the exported values, and
  `minProtocolVersion <= protocolVersion` (a floor above what the server speaks
  would lock out even a freshly built client).
- The handshake survives a bind that refuses to pair (`0.0.0.0` plaintext,
  `pairRefusal: insecure-transport`) — an already-paired phone still reaches
  `/api/config` there, and a client that could not read the version fields would
  report "update required" for what is actually a transport problem.

Two existing exact-match assertions on the `/api/config` body were relaxed to
`toMatchObject`, since the body is now additive.

## Docs

`docs/phone-client-contract.md` gains a "Protocol version" subsection under §1
documenting the three fields, the missing-key case, and the required client
behaviour on each comparison outcome. `CHANGELOG.md` gets an `[Unreleased]`
entry. No version bump.

## Gates

`npx tsc -p tsconfig.daemon.json --noEmit` clean; `npx vitest run src/daemon/web`
301 passed / 4 skipped; eslint on the touched files reports pre-existing warnings
only.

## Summary by CodeRabbit

* **New Features**
  * Added phone protocol version details to the `/api/config` response.
  * Older phone clients can now detect incompatible daemon versions and display an update-required message instead of failing silently.
  * Protocol information remains available even when secure pairing is refused.

* **Documentation**
  * Documented the phone protocol handshake, supported version fields, and client behavior for incompatible versions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to wmux are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **A phone that is too old to talk to your daemon can now say so.** The phone API had no version on it, so once a mobile build ships it is pinned to whatever HTTP surface the daemon happens to serve — and a mismatch would have surfaced as routes failing one by one with no explanation. `GET /api/config`, the call a client already makes at connect, now reports the protocol the daemon speaks, the oldest one it still accepts, and the release it was spawned from. A client below the floor can show "update the app" instead of a broken screen. Nothing that already works changes: a daemon predating this simply has no version fields, which reads as the pre-handshake protocol, and a phone that ignores them behaves exactly as it does today.
+
 ## [3.37.1] — 2026-07-27
 
 ### Fixed
--- a/docs/phone-client-contract.md
+++ b/docs/phone-client-contract.md
@@ -35,6 +35,39 @@ refused before routing — a DNS-rebinding guard. Send the host you dialed.
 This is **not** treated as evidence of a secure transport anywhere. It used to
 be, for minting; that was removed, because the caller writes the header.
 
+### Protocol version
+
+`GET /api/config` is the first authenticated call a client makes, and it carries
+the handshake:
+
+| Field | Meaning |
+| --- | --- |
+| `protocolVersion` | the phone contract this daemon speaks |
+| `minProtocolVersion` | the oldest client contract it still accepts |
+| `serverVersion` | the release the daemon was spawned from — display and bug reports only, never compared |
+
+Read it once at connect, before anything else on the screen depends on a route
+answering.
+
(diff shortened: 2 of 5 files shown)


openwong2kim:

This PR introduces a dedicated version handshake for the phone-facing web API: `GET /api/config` now reports `protocolVersion`, `minProtocolVersion`, and `serverVersion`, so a shipped iOS client can detect an incompatible daemon and surface an explicit "update required" state instead of failing route-by-route. The new `src/daemon/web/protocolVersion.ts` deliberately separates the wire protocol number from the release version (additive changes never bump it, breaking ones do), reuses the same `WMUX_SPAWNED_BY_VERSION` source the stale-daemon check reads to avoid drift, and the tests cover both the happy path and the insecure-transport bind where pairing is refused. The change is purely additive on the wire — old clients ignore the new fields, and a pre-handshake daemon's missing keys read as protocol 0 — which keeps it low-risk ahead of the TestFlight build. Marking P3 as it is forward-looking groundwork rather than a regression or urgent fix; safe to land.

PR #649, opened 2026-07-27
fix(daemon): reap phantom PTY exits instead of tombstoning live shells

### LINKED ISSUE #646: [Bug]: pane tombstoned 'dead' while its shell is still running — null exitCode is a phantom exit, not a real one
### What happened?

**A pane is tombstoned `dead` while its shell is still running.** The pane prints
`[Process exited with code -1]` and goes inert — no keystrokes, no Esc. Minutes later it "recovers"
as a fresh prompt, which is actually a *brand new shell in the wrong directory*, while the original
`powershell.exe` and the `claude.exe` under it keep running, orphaned and invisible.

I verified the process is alive **after** the daemon declared it dead:

```
wmux sessions.json : daemon-ddc7f1e5  state=dead  pid=20516
Get-Process 20516  : ALIVE, powershell.exe, started 07:41:13
  child            : PID 56488  claude.exe --resume  (running)
```

The daemon reported the death at 01:18:15Z — 37 minutes *after* that process started, and it was
still running afterwards. Not pid reuse: same process, survived its own "exit".

This is very likely the root of the long-standing **"powershell exits -1 under claude"** report that
`DaemonSessionManager.ts` already names in a comment ("Silent PTY deaths (no log, no recorded
exitCode) made [it] undiagnosable"). The forensics added there always record `null`, so they never
narrowed anything — because **there was never an exit to record**.

---

#### 1. Every recorded death has a null exit code

29/29 `session:died` events across 14 days of `~/.wmux/logs/daemon-*.log`:

```
reason=pty-exit exitCode=null signal=none
```

Not one real exit code, ever — including sessions closed normally.

#### 2. node-pty is not the reason the code is missing

I probed node-pty 1.1.0 directly (same version as `resources/daemon-bundle/node_modules/node-pty`),
8 scenarios:

| scenario | exitCode |
|---|---|
| `exit 0` / `exit 42` / `exit 1` | `0` / `42` / `1` |
| `taskkill /F` on the shell | `1` |
| external kill, consumer attached throughout | `1`, detected 1559 ms |
| external kill, **no consumer ever** | `1`, detected 1461 ms |
| external kill, consumer attached 5 s / 10 s late | `1`, detected ~1500 ms |

A real code every time. A null code is therefore a distinct failure mode, not normal behaviour.

#### 3. Where the null comes from

`node-pty/lib/windowsTerminal.js:97`

```js
this._socket.on('close', function () {
  this.emit('exit', this._agent.exitCode);
  this._close();
});
```

`_agent.exitCode` is assigned **only** by the native process-exit callback
(`windowsPtyAgent.js:221-223`, wired at line 89). `emit('exit')` fires on **conout socket close** and
reads a value written by a different path. Socket closes first → `undefined` → wmux logs `null`.

So node-pty cannot distinguish "the process exited" from "the socket went away", and wmux inherits
a phantom death.

#### 4. The OS pipes are intact — nothing actually broke

Enumerated `\\.\pipe\` while the tombstoned session was still alive:

```
conpty-<a>  -in  -out  -out-worker     complete
conpty-<b>  -in  -out  -out-worker     complete
conpty-<c>  -in  -out  -out-worker     complete   <- includes the "dead" session
conpty-<d>       -out                  fragment
conpty-<e>       -out                  fragment
conpty-<f>       -out                  fragment
conpty-<g>             -out-worker     fragment
```

Three complete triples for three live shells, including the tombstoned one — conin, conout and
worker pipes all present. The OS-level pipe is fine; what disappeared is node-pty's Node-side
`_outSocket`. The only path that destroys it without an OS break is
`_$onProcessExit` → `_flushDataAndCleanUp` → `_cleanUpProcess` → `_outSocket.destroy()`, and
`_$onProcessExit` assigns `_exitCode` unconditionally.

**Conclusion: the native ConPTY exit callback fired, with no exit code, for a process that had not
exited.** That defect is below the JS layer, in node-pty's native `connect()` exit watcher.

(Four orphaned pipe fragments are also visible — partial-teardown leaks from earlier incidents.
`wmux-session-daemon-*` pipes existed only for sessions wmux considered live, so wmux's own teardown
ran correctly; the leak is on the ConPTY side.)

#### 5. Deterministic 260 ms timing after attach

| | attach | died | Δ | flush |
|---|---|---|---|---|
| day 1 | 15:12:16.118 | 15:12:16.378 | **260 ms** | 2 097 152 B, `durationMs=229` |
| day 2 | 01:18:15.152 | 01:18:15.412 | **260 ms** | 611 077 B, `durationMs=192` |

Identical to the millisecond across a 3.4x difference in ring size → a deterministic code path, not
a race. Across 14 days: 15 of 27 deaths fall within 10 s of that session's own `attachSession`;
P(death | attach) = 4.19% against a 0.022% random-window base rate (~190x enrichment). But note the
same session survived three earlier attaches and died on the fourth, so attach alone is not
sufficient.

---

### Three damages per incident

1. **Original shell + agent orphaned.** Invisible to the UI, holding RAM, and still consuming API
   quota if the agent was mid-task. Never reaped — the tombstone says they are dead.
2. **Replacement spawns in the wrong cwd.** After the freeze, a new session appeared with
   `cwd=C:\Users\<user>` instead of the project directory. This is exactly the self-create path
   `shutdownKill.ts` documents for the reboot case ("renderer's saved ptyIds matched nothing →
   self-create → terminal reset"), but triggered by a phantom death during normal use. No
   `PaneSupervisor` restart entry appears in the log, so it is the renderer self-creating.
3. **A tombstone holding a live pid persists in `sessions.json`**, surviving daemon restarts.

Observed accumulation on my machine: daemon had **5 `powershell.exe` children** while wmux accounted
for 4 sessions (one a tombstone with a live pid); 15 `claude.exe` running for 3 visible panes.

---

### Why the pane is unrecoverable

`DaemonSessionManager.ts` classifies the exit via `shutdownKill.ts`:

```js
if (opts.shuttingDown) return true;
return opts.platform === 'win32' && exitCode === SHUTDOWN_KILL_EXIT_CODE; // 0x40010004
```

`null === 0x40010004` is false → every null-code death is filed as a **voluntary** exit and
tombstoned `dead`. This contradicts `DaemonPTYBridge.ts`, which documents a null code as the
involuntary-teardown signature — the two files disagree and the classifier wins.

Once `dead`: `resizeSession` throws `Session '<id>' is dead` (this is the `pty:resize` error spam in
the main log — 24 and 40 occurrences on two days), the session is excluded from
`listLiveSessions()`, and recovery skips it. The pane has no input target left.

---

### Hypotheses I tested and ruled out

Recording these so nobody re-treads them:

- **ConPTY resize storms.** Attach publishes real geometry → `resizeSession` → `ResizePseudoConsole`.
  6 rounds × 219 resizes each, cycling five real pane geometries under ~1.7 MB of streaming output:
  **6/6 survived.**
- **Deferred exit detection** (attach merely being when the daemon notices an earlier death).
  Killing the shell with no consumer attached, and with consumers attached 5 s and 10 s late, still
  detected in ~1.5 s with a real exit code: **not deferred, and lateness does not null the code.**
- **Main-thread stall.** `attachSession` → `SessionPipe.flush` replays the ring through a headless
  terminal synchronously on the daemon's main thread (the 192/229 ms above), while node-pty drains
  conout on a Worker thread precisely to avoid ConPTY deadlocks. 5 rounds × 17 hard event-loop
  stalls of 230 ms with ~2.6 MB streaming: **5/5 survived, no phantom exit.**
- **Double-attach.** `attachSession` is logged twice ~50–190 ms apart, but it also happens on a
  brand-new healthy session, so it is a general pattern, not the trigger.

---

### Suggested fix

**Validate the claim before acting on it.** On a PTY exit with no exit code, confirm the recorded pid
is actually gone before tombstoning:

```js
// null exitCode is not evidence of an exit — node-pty emits it on conout socket
// close too, which happens while the shell is still running.
function processGone(pid) {
  try { process.kill(pid, 0); return false; } catch { return true; }
}
```

Alive → phantom exit → reconnect rather than bury. This is cheap, needs no native change, and does
not depend on first understanding the native-layer defect, because it checks the *claim* rather than
the *cause*. It prevents all three damages at once.

Two companions worth considering:

- **Reconcile existing tombstones** that hold live pids, otherwise already-orphaned sessions stay
  orphaned across restarts.
- **Do not reclassify null as `suspended`.** That was my first instinct and it is wrong: suspending
  assumes the PTY is gone and must be respawned, which would abandon a live shell and its running
  agent exactly like the current `dead` path does.

Separately, it would help to record what node-pty actually delivered (raw payload shape, and which
of the two paths fired first). Today a null is a dead end; with that, it becomes evidence for
chasing the native `connect()` exit watcher.

I am happy to open a PR for the liveness guard if you want it in that shape.

### Steps to reproduce

I do not have a deterministic trigger — the underlying native callback misfire is not something I
could reproduce in isolation (see the four ruled-out hypotheses). What reproduces reliably *on my
machine* is the consequence:

1. Run `claude` in a pane and use it normally for a while (long agent turns, tab switching between
   several workspaces).
2. Eventually a pane prints `[Process exited with code -1]` and stops accepting input.
3. While it is in that state, read the pid from `~/.wmux/sessions.json` for that session id and run
   `Get-Process -Id <pid>` — the shell is still alive, usually with a `claude.exe` child.
4. Wait; the pane self-replaces with a new shell in the wrong cwd, and the original stays orphaned.

Step 3 is the part that matters and it is checkable on any occurrence.

### Expected behavior

A pane whose shell is still running should stay attached, or reattach. It should never be tombstoned
`dead`, and the renderer should never self-create a replacement shell over a live one.

### wmux version

3.36.0 (also reproduced on 3.35.1 and, per the logs, back through 3.x for at least 14 days)

### How did you install wmux?

`setup.exe` installer (Squirrel auto-update to 3.36.0)

Windows PowerShell 5.1 (`C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe`)

### Windows version

Windows 10 Pro 19045

Claude Code

### Confirmation

- [x] I am on Windows 10 or 11.

openwong2kim:
## The bug

On Windows, node-pty emits `exit` from two different places. One is the real process exit. The other fires when the ConPTY conout socket closes: the agent's exit handler runs with `_agent.exitCode === undefined`, which reaches us as a `null` exit code with no signal — while `powershell.exe` and whatever agent it is hosting are still running.

`isShutdownKillExit` only treats `0x40010004` (or our own `shuttingDown` flag) as involuntary, so `null` fell through to the **voluntary** path and the session was tombstoned `dead`. Three damages followed from that single false tombstone:

1. the live shell + agent were orphaned — RAM and API quota kept burning, and because recovery skips `dead` records nothing ever reaped them (the reporter found shells outliving their daemon by 11–12 days);
2. `sessions.json` persisted a `dead` record holding a **live** pid, which survived daemon restarts and hid the orphan from every census;
3. the renderer, seeing its binding die, self-created a replacement pane in the wrong cwd.

`DaemonPTYBridge.ts` documented a null exit code as the involuntary-teardown signature, directly contradicting the classifier — and the classifier is what runs.

## The guard

New `src/daemon/phantomExit.ts` holds the pure predicates, in the style of `shutdownKill.ts`:

- `isPidAlive(pid)` — `process.kill(pid, 0)`, with `EPERM` counting as alive.
- `isPhantomExit(exitCode, signal, pid, alive)` — true only for the code-less, signal-less, known-pid, still-alive shape. Platform-agnostic on purpose: on posix a genuine exit likewise always carries a code or a signal, so that shape is anomalous there too, and gating on win32 would only hide the same bug elsewhere.
- `shouldReconcileTombstone(session, { rebooted, alive })` — the boot-time eligibility policy (below).

`DaemonSessionManager`'s exit handler consults the guard *before* the dead/suspended classification. On a phantom it logs the raw node-pty payload verbatim (the native-layer investigation asked for this), cleans up the bridge, and emits a new `session:phantomExit` event carrying `{ id, pid, exitCode, signal, cmd, lastActivityMsAgo }`. It sets no state and announces no death — the manager stays mechanism-only, matching how the `session:interrupted` policy lives in `index.ts`.

## Reap and die

`index.ts` subscribes to `session:phantomExit` next to `session:interrupted`, logs a `[lifecycle]` line explaining that the pid is alive after a code-less exit, kills the process **tree** (`taskkill /pid <pid> /T /F` on Windows, `SIGKILL` to the process group with a single-process fallback elsewhere), and then runs the ordinary death path with `reason: 'phantom-exit'` so the log distinguishes it from a real exit. Observable outcome: session `dead`, pid actually gone, log says phantom.

Two roads not taken:

- **Not `suspended`.** That path respawns the session on recovery, which would abandon the very shell we are trying not to orphan.
- **Not a new session state.** A pane whose transport is gone but whose process lives is not a state users can act on — node-pty has already destroyed its `_outSocket`, so nothing can reattach. Parking the session there would keep a recorded orphan instead of an unrecorded one.

## Reconciliation

`recoverSessions` now inspects `dead` records instead of skipping them outright. When the boot id is unchanged (so the recorded pid cannot have been recycled), the pid is still alive, and `isOurShellProcess` confirms it really is the shell we spawned, the tree is reaped and logged as `[recovery] reconciled tombstone <id>: pid <pid> was still alive`. Anything uncertain — rebooted, no pid recorded, or a pid that is not our shell — is left alone. Persisted sessions already carried `pid` (`DaemonSession.pid` is required and `buildState` persists the meta verbatim), so no state-shape change was needed; the predicate treats a missing pid as ineligible for old files anyway.

## Out of scope (follow-ups)

- **Stream reattachment** to the still-live ConPTY. Realistically an upstream node-pty fix.
- **The renderer's wrong-cwd self-create.** The guard removes the trigger for it in this path, but the self-create behaviour itself is untouched.

## Tests

`src/daemon/__tests__/phantomExit.test.ts`, 14 cases across `isPhantomExit`, `isPidAlive`, and `shouldReconcileTombstone` — liveness is injected, so no real processes are involved. Full suite: 8693 passed, `tsc --noEmit` clean.

Fixes #646

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal sessions being incorrectly marked as exited when the underlying output stream closes without an exit code or signal.
  * Improved session recovery by cleaning up orphaned terminals using safer, start-time–verified process-tree shutdown and enhanced tombstone reconciliation.
  * Adjusted dead-session retention so expired dead records are reconsidered if their process is still alive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **A phone that is too old to talk to your daemon can now say so.** The phone API had no version on it, so once a mobile build ships it is pinned to whatever HTTP surface the daemon happens to serve — and a mismatch would have surfaced as routes failing one by one with no explanation. `GET /api/config`, the call a client already makes at connect, now reports the protocol the daemon speaks, the oldest one it still accepts, and the release it was spawned from. A client below the floor can show "update the app" instead of a broken screen. Nothing that already works changes: a daemon predating this simply has no version fields, which reads as the pre-handshake protocol, and a phone that ignores them behaves exactly as it does today.
 
+### Fixed
+
+- **A pane no longer goes dead while the shell inside it keeps running.** On Windows, node-pty sometimes reports that a terminal exited when in fact only its output socket closed — the code it hands back is empty, and `powershell.exe` and whatever agent it was hosting are still very much alive. wmux believed the report, marked the pane dead, and walked away, which meant an agent kept burning memory and API quota with nothing left to reach it or shut it down, and the pane came back in your home directory instead of your project. wmux now asks the operating system whether the process is actually gone before believing any exit that arrives without a code. If it is still there, wmux shuts the whole tree down itself — the shell and everything it started — and closes the pane properly, saying so in the log. The exact report from the terminal layer is logged verbatim, which is what the underlying investigation needs. Real exits are untouched: an exit that carries a code, or a signal, still behaves exactly as before. (#646)
+
+- **Shells orphaned by that bug are now cleaned up at startup.** Because the pane had already been written off, nothing ever reaped those processes — one report found shells still running eleven and twelve days after the daemon that spawned them had gone. On startup wmux now checks its own records of closed panes: if one still points at a running process, and the machine has not rebooted since (so the process id can't belong to something else), and it really is the shell wmux started, it is shut down and noted in the log. Anything uncertain is left alone.
+
 ## [3.37.1] — 2026-07-27
 
 ### Fixed
--- a/src/daemon/DaemonPTYBridge.ts
+++ b/src/daemon/DaemonPTYBridge.ts
@@ -260,10 +260,20 @@ export class DaemonPTYBridge extends EventEmitter {
     this.dataDisposable = () => onDataDisposable.dispose();
 
     // PTY exit handler. Capture `signal` alongside exitCode: a clean shell
-    // exit carries a numeric exitCode and no signal, whereas a terminated
-    // process (ConPTY torn down, killed) shows up as a signal or a null
-    // exitCode. That distinction is what the silent-death investigation needs
-    // to tell "the shell exited on its own" from "something killed it".
+    // exit carries a numeric exitCode and no signal, whereas a killed process
+    // reports the signal that killed it. That distinction is what the
+    // silent-death investigation needs to tell "the shell exited on its own"
+    // from "something killed it".
+    //
+    // A NULL exitCode with no signal is neither (#646). It is node-pty's
+    // conout-socket-close path: the Windows agent's exit handler runs with
+    // `_agent.exitCode === undefined` when the socket drops, and the shell may
+    // well still be running. This comment used to call that shape the
+    // involuntary-teardown signature, which contradicted the classifier in
+    // shutdownKill.ts (only 0x40010004 / our own shutdown flag qualify) — and
+    // the classifier is what runs, so null was treated as a voluntary exit and
+    // buried a live shell. Consumers must not infer a death from a null code
+    // alone; see phantomExit.ts for the liveness check that settles it.
     const onExitDisposable = ptyProcess.onExit(({ exitCode, signal }) => {
       this.emit('exit', { sessionId, exitCode, signal });
     });
--- a/src/daemon/DaemonSessionManager.ts
+++ b/src/daemon/DaemonSessionManager.ts
@@ -16,7 +16,7 @@ import { isMac } from '../shared/platform';
 import { getWindowsDefaultShell, resolveBareShellName, resolveLaunchableWindowsExe } from '../shared/shellResolution';
 import { ENV_KEYS } from '../shared/constants';
 import { createDefaultConfig } from './config';
-import { isPhantomExit, isPidAlive } from './phantomExit';
+import { getProcessStartTime, isPhantomExit, isPidAlive } from './phantomExit';
 
 const DEFAULT_COLS = 80;
 const DEFAULT_ROWS = 24;
@@ -119,7 +119,8 @@ const clampRows = (rows: number): number => Math.max(MIN_SAFE_ROWS, rows);
  *      A PTY exit classified as involuntary (OS shutdown killing children —
  *      see shutdownKill.ts). The session is SUSPENDED, not dead: daemon/index
  *      dumps the buffer + persists so post-reboot recovery replays the same id.
- *  - 'session:phantomExit'  → { id, pid, exitCode, signal, cmd, lastActivityMsAgo, raw }
+ *  - 'session:phantomExit'  → { id, pid, pidStartTime, exitCode, signal, cmd,
+ *                               lastActivityMsAgo, raw }
  *      A PTY exit that reported no exit code and no signal while the shell pid
  *      is STILL ALIVE (node-pty's ConPTY socket-close path — see
  *      phantomExit.ts / issue #646). Mechanism only: no state is set and no
@@ -592,6 +593,8 @@ export class DaemonSessionManager extends EventEmitter {
           signal: payload.signal,
           cmd: meta.cmd,
           lastActivityMsAgo,
+          // Identity of the pid we are about to ask the policy layer to kill.
+          pidStartTime: meta.pidStartTime,
           // Carried, not logged here: this class has no access to the daemon's
           // file logger, and a console.log would never reach the daemon log
           // the native-layer investigation actually reads. index.ts prints it
@@ -633,6 +636,17 @@ export class DaemonSessionManager extends EventEmitter {
       bridge.setMuted(true);
     }
     bridge.setupDataForwarding(ptyProcess, ringBuffer, params.id, promptLog);
+
+    // #646: stamp the pid's OS creation time so later reaping paths can tell
+    // OUR shell from whatever process inherits the pid after recycling.
+    // Deliberately not awaited — createSession is synchronous and the probe
+    // shells out. The value lands on the live meta well before any tombstone
+    // could be written, and every sessions.json write reads meta at write
+    // time, so persistence picks it up. If the probe fails the field stays
+    // absent and the reaping paths fall back to their weaker check.
+    void getProcessStartTime(meta.pid).then((startTime) => {
+      if (startTime) meta.pidStartTime = startTime;
+    });
 
     this.emit('session:created', { session: { ...meta } });
     return { ...meta };
--- a/src/daemon/DaemonSessionManager.ts
+++ b/src/daemon/DaemonSessionManager.ts
@@ -119,13 +119,15 @@ const clampRows = (rows: number): number => Math.max(MIN_SAFE_ROWS, rows);
  *      A PTY exit classified as involuntary (OS shutdown killing children —
  *      see shutdownKill.ts). The session is SUSPENDED, not dead: daemon/index
  *      dumps the buffer + persists so post-reboot recovery replays the same id.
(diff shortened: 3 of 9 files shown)


openwong2kim:

This PR fixes #646 — the phantom-PTY-exit bug where a Windows pane was tombstoned `dead` while `powershell.exe` and the agent under it kept running. It adds `src/daemon/phantomExit.ts` with three pure, platform-agnostic predicates (`isPidAlive`, `isPhantomExit`, `shouldReconcileTombstone`), then consults `isPhantomExit` in `DaemonSessionManager` *before* the dead/suspended classification so a code-less, signal-less, still-alive exit no longer buries a live shell. The transport is irrecoverable either way (node-pty has already destroyed its `_outSocket`), so the policy in `index.ts` deliberately reap-and-dies the orphaned process tree and runs the normal death flow with a `phantom-exit` reason rather than suspending — which would respawn the session and abandon the very shell this fix exists to protect.

The diff is well-scoped: it reconciles persisted tombstones at boot under a conservative policy (same boot id, pid still alive, confirmed to be our shell via `isOurShellProcess`), so a reboot cannot trick it into killing an unrelated recycled pid; it logs the raw node-pty payload verbatim, which is what the underlying native-layer investigation asked for; and the 14 new unit cases cover the incident signature, the live/dead and signal/code negations, the unknown-pid guards, and the reboot exclusion. Reviewing the actual diff, the classification ordering is correct (phantom check gated on `!involuntary`), the `session:phantomExit` event is mechanism-only with the policy owned by `index.ts` (matching the `session:interrupted` split), and the `reapProcessTree` helper is best-effort with swallowed failures so a kill error cannot trip the three-strikes fatal handler. Worth a careful look at the Windows `taskkill /T /F` path and the `isOurShellProcess` confirmation before merge, since that is the one place this fix does an irreversible thing.

PR #651, opened 2026-07-27
perf(gate): judge cold start on the fastest boot, not the median

openwong2kim:
## Why

A perf red on `main` (f6644ac) that no code could explain. That commit's entire diff widens web pairing codes from six characters to eight, and `WebTerminalServer` is only constructed when web is enabled — nothing in that path runs during a cold boot.

The runner was the cause, and the evidence is in the run's own daemon marks:

| | healthy run (582eb50, 07:19) | red run (f6644ac, 09:36) |
| --- | ---: | ---: |
| daemon process spawn → `main-start` | 615 ms | **2177 ms** |
| `main-start` → `lock-acquired` | 5 ms | **2043 ms** |
| the three boots (`firstPtyDataMs`) | 1254 / 1259 / 1176 | **1442 / 9154 / 2470** |
| median vs 1207 ms baseline | 1254 (PASS) | 2470 (**FAIL**) |

The confirmation re-run (#570, #632) could not catch this, because it re-measures on the **same runner**. It answers "is this code deterministic", not "is this machine fit to measure" — so a host degraded for the whole job reproduces the red by construction, which is exactly what the log says happened.

## What changed

Gate the **fastest** of the N boots instead of the median. Interference is one-sided: a busy host makes a boot slower and never faster, so the minimum is the sample least polluted by the machine.

- `scripts/perf-bench.mjs` — records `coldStart.best` alongside `coldStart.median` and prints both.
- `scripts/perf-compare.mjs` — the cold-start gate reads `scenarios.coldStart.best.firstPtyDataMs`. A gate may now name a `baselineFallbackPath`, so a baseline blessed before this change is read from its median rather than reported `NEW` — a gate that silently stops gating is worse than a conservative one, and the old median is never below the new best.
- `tailRegressionNote()` — when the median trips both thresholds while the gated fastest boot does not, the summary says so as a **note, never a failure**. A slowdown only some boots hit is a startup race or a partly busy runner: worth a look, not worth a red build on a machine nobody owns.
- `bench/baseline-{ci,local}.json` — a `best` block computed from the runs each baseline already records, so the bless stays descriptive rather than aspirational. CI baseline: `firstPtyDataMs` 1134 ms.
- `bench/README.md` — gate table, the estimator rationale, what the re-run cannot answer, and a note that the `coldFirstPtyDataMs` trend series carries the fastest boot from here on.

## Verification

Both result artifacts from the red run (30254659860) were replayed through the real CLI against the updated baseline:

| input | result |
| --- | --- |
| the red run itself (best 1442 ms) | **exit 0** — gate green, tail note printed |
| every boot 2.5x slower (3000 ms) | **exit 1** — detection preserved |
| that regression vs a pre-`best` baseline | **exit 1** — fallback keeps it gated |

Seven regression tests added in `scripts/__tests__/perfCompare.test.mjs`, using the real 1442 / 9154 / 2470 numbers. `npm run test:parallel`: 8716 passed, 0 failed.

## Trade

Best-of-N needs all N boots to regress before it fails, where median-of-3 needed two. That is the coverage this buys immunity to a bad runner with, and the tail note is what keeps the difference visible instead of silent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated cold-start performance gating to reduce false failures from slow CI machines by using the fastest observed boot while still capturing median performance.
  * Added conservative fallback behavior when older baselines lack the gated metric, marking results appropriately instead of misclassifying them as new.
  * Enhanced tail-regression reporting when median regresses past thresholds but the gated “best” value remains within limits, including during confirmation handling.
* **Documentation**
  * Clarified cold-start “best vs median” semantics and updated gating/metric mapping notes in the bench README and changelog.


openwong2kim:

This PR reworks the cold-start perf gate to judge the **fastest** of the N boots instead of the median, fixing a false red that blocked the 3.35.0 release. The motivating incident is well-evidenced: on 2026-07-27 a degraded `windows-latest` runner needed 2.2s just to spawn the daemon process and 2.0s to acquire a file lock, which pushed the median to 2470ms against a 1207ms baseline on a commit whose only diff widened web pairing-code length — and the existing confirmation re-run could not save it because it lands on the *same* runner, so a host degraded for the whole job reproduces the red by construction.

The core insight is sound: runner interference is one-sided (a busy host makes boots slower, never faster), so the minimum is the sample least contaminated by the machine, while a genuine regression slows every boot and still trips both thresholds. The implementation touches `scripts/perf-compare.mjs` (gate now reads `scenarios.coldStart.best.firstPtyDataMs`, with a `baselineFallbackPath` so a pre-`best` baseline falls back to its median instead of silently going `NEW` and ungate the metric), `scripts/perf-bench.mjs` (records `best` alongside `median`), both `bench/baseline-*.json` files (adds descriptive `best` blocks computed from existing runs), and adds `tailRegressionNote()` so a median-only regression the gate let through is still surfaced as a NOTE rather than swallowed.

The verification is the strongest part: both result artifacts from the false-red run (30254659860) were replayed through the real CLI — the red run reads 1442ms and passes, an all-boots-2.5× regression still fails, and the pre-`best` baseline keeps it gated via fallback — and seven regression tests using the actual 1442 / 9154 / 2470 numbers lock the behavior in, with 8716 passing overall. The trade is acknowledged honestly (best-of-N needs all boots to regress before failing, where median-of-3 needed two), and the tail note is what keeps that difference visible instead of silent. This is a P2: a flaky gate trains people to re-run rather than investigate, which is exactly the state a real regression walks through undetected, so making the signal trustworthy is worth fixing soon.

PR #653, opened 2026-07-27
fix(web): find the Tailscale CLI from the GUI process on macOS

openwong2kim:
## Summary

The "Serve panes to a browser" Start button failed with `spawn tailscale ENOENT` on macOS even with Tailscale installed and the CLI working in every terminal. An app launched from the Dock inherits launchd's minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin`), so the GUI process could not resolve a `tailscale` that lives in Homebrew, `/usr/local/bin`, or `~/.local/bin`. `wmux web --tailscale` in a terminal was never affected.

Two-layer fix:

- **`tailscaleCandidates()`** now falls back to the CLI inside `/Applications/Tailscale.app/Contents/MacOS/Tailscale` on macOS — a stock desktop install never puts it on any PATH (Tailscale's own docs tell users to alias it by hand) — mirroring the existing Windows Program Files fallback.
- **The tailscale exec now runs with `getExecEnv()`**, the PATH-corrected env that git spawns already used.

Structural change: `getGitExecEnv` is renamed to `getExecEnv` and extended with `~/.local/bin`, because the correction was never git-specific — any external binary the GUI spawns hits the same launchd PATH wall. This is the second recurrence of the same defect class (git branch-sync badge 2026-07-19, tailscale 2026-07-27); the module now states the rule once: every GUI spawn of an external binary passes `env: getExecEnv()`.

## Test coverage

- `execEnv.test.ts`: asserts `~/.local/bin` is appended to PATH on macOS (the per-user CLI shim case that triggered this report).
- `tailscale.test.ts`: `tailscaleCandidates` now has a darwin branch asserting the bare-name-first, app-bundle-fallback order.
- Existing coverage already locks the ENOENT-only fallback walk in `runTailscale` (a real failure like "not logged in" never advances to the next candidate) and the Homebrew/system PATH augmentation.

## Test plan

- [x] `tsc --noEmit` clean
- [x] Vitest parallel suite: 8719 passed, 20 skipped
- [x] Vitest runtime suite: 89 passed, 9 skipped
- [x] Touched files lint clean


openwong2kim:

This fixes `spawn tailscale ENOENT` on macOS when the "Serve panes to a browser" Start button fires from the GUI: an app launched from the Dock inherits launchd's minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin`), so the process cannot resolve a `tailscale` CLI that lives in Homebrew, `/usr/local/bin`, or `~/.local/bin`. The CLI terminal path (`wmux web --tailscale`) was never affected because it has a full shell PATH. Two concrete changes close it: `tailscaleCandidates()` gains a darwin branch that tries the bare name first and falls back to the CLI inside `/Applications/Tailscale.app/Contents/MacOS/Tailscale` (mirroring the existing Windows Program Files fallback), and the tailscale exec now runs with `getExecEnv()`.

The structural part is the right call: `getGitExecEnv` is renamed to `getExecEnv` and extended with `~/.local/bin`, because the launchd-PATH defect was never git-specific — every external binary the GUI spawns hits the same wall. The diff propagates the rename across every spawn site (git, gh, npm, worktree, mergeSession, metadata, task close/scan) and the module now states the rule once: every GUI spawn of an external binary passes `env: getExecEnv()`. This is the second recurrence of the same defect class (git branch-sync badge 2026-07-19, tailscale 2026-07-27), so folding it into one named helper rather than re-discovering it per spawn site is the correct fix.

Test coverage is appropriate: `execEnv.test.ts` asserts `~/.local/bin` is appended on mac and the non-mac identity-pass-through still holds; `tailscale.test.ts` locks the bare-name-first / app-bundle-fallback order on darwin; existing coverage already pins the ENOENT-only fallback walk in `runTailscale` so a real failure like "not logged in" never advances to the next candidate. The PR body reports `tsc --noEmit` clean, the parallel vitest suite at 8719 passed / 20 skipped, the runtime suite at 89 passed / 9 skipped, and touched files lint clean.

**Priority: P2** — owner-reported functional break on a shipped feature, correct and well-tested fix that also retires the underlying defect class.

PR #655, opened 2026-07-27
feat(chat): Chat View — a conversation lens over agent panes

openwong2kim:
## Summary

Chat View renders a Claude pane's conversation — human turns and agent prose, with tool calls, code, and diffs collapsed to one-line chips — as an opt-in view mode on the same pane surface. The terminal stays the default rendering and the ground truth: the PTY runs untouched underneath, and a persistent trust seam links back to it. Toggle per pane via a Terminal|Chat segment in the surface tabs or `Cmd/Ctrl+Shift+J`.

### How it works

- **Daemon `TranscriptProjector`** (`src/daemon/transcript/`): tails the Claude Code transcript JSONL the hook bridge already registers per pane, normalizes entries into a shared `TurnEvent` model, and serves snapshot/append/code-block RPCs. Bounded reads (256KB tail, 128KB serialized budget per payload), byte-accurate offsets computed on the Buffer, directory-level fs.watch with a polling fallback, per-client subscriptions with backpressure protection, and first-party-only access to conversation content.
- **Main relay** (`chat.handler`): Electron IPC bridge (`CHAT_*` channels) between the daemon projector and the renderer; approval-gate state rides a dedicated channel.
- **Renderer** (`src/renderer/components/Chat/`): store-driven chat list with hand-rolled row windowing, collapsed-by-default code/diff/tool chips, and a composer wired to the same PTY.
- **Composer safety**: the composer locks while an approval prompt is open — lock sources are split (hook/registry vs screen-derived), unlock happens only on approval resolution or turn-end, the gate seeds from the daemon's open-approval list on mount/reconnect (fail-closed), and the paste payload itself is guard-checked at write time.
- **Memory policy**: optional keep-warm LRU (Settings ▸ Terminal, default off) unmounts hidden terminals beyond the 4 most recently toggled chat panes; an in-app probe (`__wmuxChatMemory`) measures the 30-pane comparison that decides whether the default flips on.

### Notes

- Claude-only in v1: Codex/Gemini panes show a disabled toggle (no structured transcript to project). Chat View for a fresh session becomes available after its first turn completes (the transcript path is only known then).
- A multi-pass adversarial review ran before this PR; all confirmed critical findings are fixed in the three `fix(…)` commits (reachability bootstrap, seq resync loop, oversized/multibyte transcript handling, transcript RPC trust boundary, composer gate fail-closed semantics, pipe backpressure, transcript-path validation).
- Design contract recorded in DESIGN.md (terminal-first identity unchanged; Chat is a per-surface lens, not a pivot).

## Test plan

- `tsc --noEmit`, `npm run build:daemon`, and the full vitest suites pass (~6,300 tests incl. new coverage: transcript parser/reader fixtures — malformed, oversized, multibyte, Windows paths — projector watch/runtime tests, IPC relay, composer gate invariants, reachability-without-state-injection regression, keep-warm LRU/windowing).
- Pending owner dogfood on a live fleet: end-to-end projection, approval-gate behavior, and the 30-pane memory measurement (procedure documented on the probe).

## Follow-ups (tracked, intentionally out of scope)

- keep-warm slot cleanup on pane teardown, per-consumer subscription refcount, row-height re-measure on chip expand, stale-path guard on session restart, marker-boundary-aware prose cap, uuid dedup hardening — review INFO items deferred by owner decision.
- P3: the same projection over `wmux web` for phones.

## Summary by CodeRabbit

* **New Features**
  * Added opt-in Chat View for agent panes with a terminal↔chat toggle, live transcript updates, folded tool calls, code-block expansion, diff chips, safe Markdown (including tables), and a “jump back to terminal” trust seam.
  * Added optional keep-warm/unload behavior and composer input locking aligned to approval prompts.

* **Bug Fixes**
  * Improved Chat View reliability and UX: correct newest-turn opening, scrolling, turn formatting, and composer newline/submit behavior (no stuck sessions after exit).
  * Hardened transcript handling and improved correctness across multiple app instances.


openwong2kim:

This PR introduces **Chat View**, an opt-in per-pane view mode that renders a Claude pane's on-disk transcript JSONL as a structured conversation — human turns and agent prose with tool calls, code blocks, and diffs collapsed to one-line chips — layered over the same pane surface while the terminal stays mounted underneath as the ground truth. It is the largest single piece of surface work in the queue (81 files, ~10k additions) and spans a new daemon-side `TranscriptProjector` (bounded, crash-proof tail reads via `readTail.ts`, `fs.watch` with polling fallback), a main-process IPC relay (`chat.handler.ts`), and a full renderer component tree under `src/renderer/components/Chat/` plus a keep-warm LRU memory policy in Settings.

The security posture is the part I want on record, because it is unusually careful for a feature this size and is what makes it safe to land: `transcriptPathGuard.ts` confines the hook-supplied transcript path to the Claude projects root with a `realpath` symlink check and a basename-equals-`${agentSessionId}.jsonl` rule (turning a would-be arbitrary-file-read into a refused binding); the `daemon.transcript.*` RPCs are gated behind a `firstPartyOnly` classification so a pane's full conversation never reaches CLI/MCP pipe clients; `sendTo` enforces per-subscriber backpressure (`SUBSCRIBER_BACKPRESSURE_BYTES`) to stop a stalled reader from growing daemon heap unbounded; and the composer gate locks on a hook-authoritative approval and unlocks only on hook-derived turn-boundary signals — never on byte-silence `idle`, which would re-open the arrow-key-menu hazard on a timer.

**Priority: P2.** This is flagship feature work rather than a regression, and the owner has already run a multi-pass adversarial review (the three `fix(…)` commits address reachability bootstrap, seq resync, oversized/multibyte handling, the transcript RPC trust boundary, composer fail-closed semantics, and pipe backpressure), with ~6,300 tests passing. The remaining gate before merge is the owner's live dogfood across a real fleet (end-to-end projection, approval-gate behavior, the 30-pane memory probe that decides the keep-warm default) — that is the right bar for a change that touches daemon, main, preload, and renderer at once.

openwong2kim:
@codex review

openwong2kim:
## Live dogfood

Ran this on a live instance, not a harness: dev build of `feat/chat-view` @ b4f759fb on Windows 11, launched side-by-side with the production app under `WMUX_DATA_SUFFIX=-chatview` (own pipe `wmux-chatview-rizz`, own data dir), Claude Code v2.1.220 in a real pane, everything driven with real clicks and keystrokes over CDP.

Short version: once a transcript path reaches the daemon, Chat View is good — projection, chips, code fetch, composer round-trip, unlock, and the trip back to the terminal all behave. Getting the transcript path there is where it falls over, and it falls over in exactly the configuration you'd dogfood in.

### 1. Hooks never reach a suffixed instance, so Chat View can never open there

`wmux-bridge.mjs` honours `WMUX_DATA_SUFFIX` for the **main** endpoint (`instanceSuffix()`, `getAuthTokenPath`, `getPipeName`) but hardcodes the unsuffixed `~/.wmux/` for the **daemon** endpoint (`getDaemonAuthTokenPath`, `getDaemonPipeName`). The comment above them says the bridge "cannot see `WMUX_DATA_SUFFIX` (a reserved `WMUX_*` var, stripped from the pane env)" — that isn't true any more, and the top of the same file already says so. `envFilter` only strips `WMUX_AUTH*`, and I confirmed it live: the pane's env carries `WMUX_DATA_SUFFIX = -chatview`.

Running the real bridge with the pane's exact env:

```
bridge.log: {"bridge":"0.2.0","hook":"Stop","outcome":"rpc-rejected","target":"daemon","reason":"no-workspace-match"}
```

It connected to `\\.\pipe\wmux-daemon-rizz` — the **production** daemon — which of course doesn't know that pane, so the binding got spooled into production's `~/.wmux/resume-spool` and the `-chatview` daemon never learned anything. `daemon.transcript.status` stays `no-binding`, the Chat segment stays greyed out forever, and the tooltip promises "Chat View opens after this agent's first turn completes" while no turn will ever make it open.

Net effect on this PR's own test plan: "owner dogfood on a live fleet" is only possible on an unsuffixed packaged build, which can't run while a production wmux holds the single-instance lock.

Two more instance-blind spots on the same path, same root cause:
- the hook installer wrote into `~/.wmux/hooks/` and rewrote the global `~/.claude.json`, `~/.codex/config.toml`, `~/.gemini/settings.json` to point at the dev worktree's MCP entry when the `-chatview` app booted. Those registrations dangle as soon as the worktree is deleted — I restored them to `~/.wmux/mcp/index.js` by hand afterwards.
- `getResumeSpoolDir()` is unsuffixed too, so the fallback lands in the wrong instance's spool as well.

Fix looks small: run the daemon dir, the spool dir and the log through `instanceSuffix()` the way the main endpoint already does.

Workaround I used to get past this and actually exercise the feature: an isolated `USERPROFILE` home with this instance's `daemon-pipe` + `daemon-auth-token` planted at the path the bridge reads, then invoke the real bridge — `outcome: ok, target: daemon`. Everything below is from that setup, so daemon → main → renderer is real; only "who spawned the bridge" is simulated.

### 2. The locked composer never shows the question

The gate itself is correct — `data-chat-composer="locked"`, textarea and send button **removed** from the DOM, "needs you" badge, "Answer in terminal" CTA. But the question is missing.

The daemon has it:

```json
{ "id": "78a3dd65…", "sessionId": "daemon-18403677", "agent": "claude",
  "kind": "awaiting_input", "question": "Dogfood: which deploy target?",
  "options": ["Staging", "Production"], "state": "pending" }
```

and the UI renders no `[data-chat-pending-question]` at all — the user just gets "This pane is waiting on a terminal-only answer." The lock rides the PR's approval-gate channel, but `pendingQuestion` only ever gets set from `METADATA_UPDATE` (`useNotificationListener` → `setSurfacePendingQuestion`), which the daemon approval path doesn't feed. The PR body says the question is shown above the locked composer, so this is a wiring gap rather than a design call.

### 3. The speaker label follows the heuristic detector, and it disagrees with the transcript

Mid-session the pane chip read `w1-1(codex)` and Chat View attributed Claude's turns to **"Codex CLI"** — same pane, Claude Code v2.1.220, screenshot has both. It flipped back to "Claude Code" once the hook binding landed, then flapped again later.

Chat View is Claude-only by construction (`binding.agent !== SUPPORTED_AGENT → not-claude`), and the bridge hardcodes `agent: 'claude'`, so the label should come from the projection or the binding, not from `lastDetectedAgent`. As it stands the one surface that *knows* it is rendering a Claude transcript is the one showing another vendor's name.

### 4. Prose renders raw markdown

`**bold**`, backticks and `-` bullets show up literally in Chat View, while the same turn renders formatted in the terminal underneath it. For a surface whose pitch is "the conversation, minus the noise", it currently reads noisier than the ground truth it projects. Either render a minimal subset or strip the markers.

### 5. The locked composer clips its own primary CTA

With the gate engaged, "Answer in terminal" is cut off by the bottom edge and the trust seam scrolls out of view — 1080p window, single pane, nothing exotic. The locked block grows by three elements and the list doesn't yield the space.

### 6. A pane that inherits `CLAUDE_CODE_CHILD_SESSION` silently can never open Chat View

First pane I opened printed `Transcript saving is off — inherited CLAUDE_CODE_CHILD_SESSION marker`. No transcript file is ever written, so Chat View is unreachable, and the only thing the UI says is that it "opens after this agent's first turn completes". You hit this whenever wmux is launched from inside a Claude session (agent-driven launches, dogfood rigs). `CLAUDE_CODE_*` isn't in `INTERNAL_PATTERNS`, so it rides all the way into the pane. Worth either stripping the marker for panes or saying "this pane has transcript saving off" instead of promising a turn that will never help.

### What worked

- Projection of a real transcript: user turn, collapsed `✓ 1 tool call (Read)`, agent prose, `code · 6 lines` chip, `● complete` status, trust seam `projection of 9d739e47-….jsonl · terminal is ground truth`.
- Tool-run expand → `✓ Read → readTail.ts`. Code chip on-expand fetched the real body from the daemon (`export function readTranscriptPage(` … 6 lines) — bodies really do stay off the wire until asked for.
- Composer → PTY round trip: typed in Chat, the agent answered, and the same turn shows in the terminal. Rows went 5 → 9 live, no resnapshot flicker.
- Unlock on turn end: `Stop` → approval goes `expired` → composer back to `open`.
- `daemon.approvals.resolve` fails closed with `prompt-gone` when the pane's screen no longer shows the menu.
- `daemon.transcript.*` refuses a non-first-party pipe client (`not-authorized`) — the trust boundary holds against a hand-written client.
- Toggle back to the terminal repainted the whole session in ~2 s, no corruption, alt-screen caveat not hit.

### Not done: the 30-pane measurement

The probe itself works — one live sample, 1 pane in Chat: mean 961.9 MB over 5 reads, and it reports `panes` / `chatPanes` / `keepWarmLru` alongside. But the §8 criterion needs 30 panes each holding a registered transcript binding, and with finding 1 in place every one of those bindings has to be hand-injected, so the number would be measuring the harness rather than the feature. This one should be the first thing run after the bridge fix — it's still the gate on flipping the setting's default.

openwong2kim:
## Follow-up: Codex is closer than the body says, and a warning for the P3 phone surface

Two things I looked into after the dogfood run above.

### Codex isn't blocked by "no structured transcript"

The body says Codex/Gemini get a disabled toggle because there's no structured transcript to project. That's true for Gemini; for Codex it isn't:

- Codex writes JSONL rollouts — `~/.codex/sessions/YYYY/MM/DD/rollout-<ts>-<uuid>.jsonl`, real files on this machine (45 KB – 1.8 MB). The shape is a clean envelope: `{timestamp, type, payload}` with `session_meta` / `turn_context` / `event_msg` (`task_started`, …) / `response_item` (`payload.type` = `message` | `function_call` | `reasoning`, `role` = user/assistant/developer, `content: [{input_text|output_text, text}]`).
- We already ship the capture half. `integrations/codex/bin/wmux-codex-notify.mjs` (v0.2.0) sends the same canonical envelope as the Claude bridge — `agent: 'codex'`, `agentSessionId`, and `payload.transcript_path` — and `HookIngest` builds the binding agent-agnostically, as its own comment says.

So the part that's usually painful (pinning a transcript path to an exact pane) already works for Codex. What actually blocks it:

1. `TranscriptProjector.resolve()` — the one-line `binding.agent !== SUPPORTED_AGENT → 'not-claude'` gate.
2. `transcriptPathGuard.claudeProjectRoots()` — only `~/.claude/projects` (+ `CLAUDE_CONFIG_DIR`). A `~/.codex/sessions/…` path is refused as "outside the Claude projects root", so the binding loses the path before the projector ever sees it. Needs a per-agent root allowlist.
3. `parseEntry.ts` — a second mapper. This is the only substantial piece, and it's bounded: `TurnEvent` already exists as the normalization target, and the fiddly part (dropping Codex's injected `developer`-role messages — permissions instructions, plugin lists) is the same class of filtering the Claude parser already does for meta/sidechain entries, fixtures included.

Two constraints that shape the first cut:

- Codex's notify only fires on `agent-turn-complete`. No `PreToolUse` ⇒ no approval-gate lock for a Codex pane.
- The composer's DECSET 2004 bracketed-paste assumption is Claude-specific per the comment in `ChatComposer.tsx`; unverified against Codex's TUI.

Both are avoided by shipping **Codex read-only** first: the lens works, the composer stays hidden. 1 and 2 are hours, 3 is the real work, 4 and 5 are zero if input stays off.

### If P3 (the same projection over `wmux web`) is the destination

Worth being explicit about what transfers, because it isn't the whole PR:

**Transfers as-is** — `src/daemon/transcript/*` (parseEntry 19 KB + readTail 17 KB + TranscriptProjector 25 KB): bounded reads, byte-accurate offsets, watch + polling fallback, per-client subscriptions with backpressure, the first-party gate, the path guard. Plus the `TurnEvent` contract in `src/shared/transcript/`. This lives in the daemon — the same process that serves `wmux web` — and it's the hard half.

**Doesn't transfer** — the `CHAT_*` Electron IPC relay (the phone needs HTTP/SSE), and `src/renderer/components/Chat/*` (React + Tailwind + our tokens, whereas the web client is a single vanilla `app.js`, 68 KB, no React and no imports). The UI gets written twice. keep-warm LRU, row windowing and the memory probe are desktop-only concerns — a phone shows one pane.

For reference, `WebTerminalServer.ts` (130 KB) currently has **zero** references to transcript/TurnEvent, so none of this is wired to the phone path yet.

### ⚠ The composer gate has to move server-side before that surface exists

Today the gate is enforced in the renderer: `needsInput` prop, `submitGuard`, and the textarea removed from the DOM. That is a client-side check. A phone client that POSTs to the daemon bypasses it completely — the "no byte reaches the PTY while an approval is open" property would silently stop holding for the surface where it matters most.

On a phone the composer matters *more* than on desktop (typing into a terminal on a phone is miserable), so the fail-closed work is aimed right; it just needs its enforcement point inside the daemon's write path rather than in the component. Cheaper to put in the contract now than to retrofit once a web client is talking to it.

### Suggested order

1. The bridge suffix fix from the comment above — until then nobody can dogfood any of this without a packaged, unsuffixed build.
2. Land this as-is, default off, and freeze desktop polish. Markdown rendering and the §8 memory default flip have no bearing on the phone.
3. Expose the projector over `WebTerminalServer` via SSE, reusing `TurnEvent` verbatim, with the PTY write gate enforced in the daemon.
4. Codex read-only. On a phone, "which agent am I looking at" matters more than it does on desktop.

openwong2kim:
## Memory: an 8-pane run, and why the §8 procedure can't be trusted as written

Followed up on the measurement I left open above. I couldn't do the full 30, so this is 8 panes — but the interesting result isn't the number, it's that **the documented procedure measures the instrument, not the feature**.

Setup: dev build of this branch, 8 panes, one real Claude transcript bound to each so every pane could actually enter Chat, `__wmuxChatMemory` for the tree total plus per-process working set from the OS side.

### First pass — looked alarming, and was wrong

| state | tree total |
|---|---|
| 8 terminals (baseline) | 879.2 MB |
| 8 chat panes, LRU off | 1013.9 MB |
| 8 chat panes, LRU **on** (4 xterms unmounted) | 1055.6 MB |
| LRU on, post-GC | 1051.8 MB |
| LRU off again, post-GC | 1040.7 MB |
| **back to 8 terminals**, post-GC | **1097.2 MB** |
| 8 terminals, idle +2.5 min | 1092.2 MB |

Reading that top-down gives "chat costs ~135 MB for 8 panes" and "the LRU costs rather than saves". Both conclusions are artefacts. The tell is the second-to-last row: returning to the *exact* starting state left the number 213 MB higher, and it then stayed flat while idle. So the growth tracked toggling activity, not the state being sampled — which means an A→B sample can attribute it to whatever B happens to be.

### Second pass — where the growth actually lived

Per-process, over 15 more round trips:

- **GPU process plateaus**: 224 → 248 → 249 → 249 MB. A bounded cache, not a leak.
- **Renderer JS heap is flat**: 24 MB at round 0 and still 24 MB at round 20. Nothing is being retained on the JS side by the chat rows, the event arrays, or the composer.
- There were **two** renderer processes, and the volatile one was **DevTools**, which the dev build opens for you. Closing it dropped the tree from 1278 MB to 912 MB — that one process was holding 333 MB of accumulated console output, and this app logs a lot into it.

### Third pass — clean, DevTools closed, LRU off, 15 rounds (120 chat mounts + 120 restores)

| round | gpu | renderer | main | daemon | total |
|---|---|---|---|---|---|
| 0 | 256 | 248 | 142 | 115 | **906 MB** |
| 5 | 238 | 250 | 138 | 113 | 884 MB |
| 10 | 237 | 252 | 138 | 113 | 885 MB |
| 15 | 237 | 257 | 138 | 115 | **892 MB** |

Net −14 MB. The renderer moves +9 MB across 15 rounds (~0.6 MB/round, allocator noise territory) and the JS heap doesn't move at all. **The growth disappears once DevTools is out of the picture.**

### What this means for §8

1. **Chat View's steady-state cost is at the noise floor** on 8 panes — no measurable JS-heap cost, single-digit MB in the renderer, GPU bounded. Whatever blocks flipping the default, it probably isn't the projection itself.
2. **The keep-warm LRU shows no measurable saving.** It really does unmount the terminals (8 xterms → 4, verified), but the working set doesn't follow. Caveat that matters: my panes were idle shells, so the terminals being unmounted were cheap ones. This under-measures the LRU's upside — it does *not* mean the LRU is useless, it means 8 idle panes can't demonstrate it. Real agent panes with an 8 MB ring each are the case worth measuring.
3. **The procedure in `chatMemoryProbe.ts` needs two amendments before its result can gate anything.** As written — one dev instance, sample terminals, toggle, sample again — it measures DevTools accumulation and mount churn as if they were the feature. Suggest: run it on a **packaged** build with **DevTools closed**, and make it **A/B/A** (return to terminals and re-sample). If the closing sample doesn't come back near the opening one, the run is invalid and no comparison from it should be believed.

Numbers above are from a dev build, so treat them as a shape rather than absolutes; the packaged tree is lighter. The 30-pane run is still the one that decides the default, and it should now be cheap: the binding trick (one transcript bound to every pane) makes it a fleet of shells rather than a fleet of agents.

openwong2kim:
## Owner dogfood — three findings, all fixed on this branch

This PR listed "pending owner dogfood on a live fleet" as the remaining gate. That dogfood happened today against a real fleet, and produced three findings. All three are now on the branch.

### 1. Quitting the agent left the lens projecting a dead conversation

`fix(chat): leave Chat View when the pane's agent process exits`

Quitting `claude` does not end the pane — the shell it ran in survives — so the lens kept rendering a finished conversation with no way back to an agent. The composer wrote into that shell, which ignored it, and nothing on the surface offered a relaunch.

The terminal is the ground truth; when the ground truth has lost its agent there is nothing left to project, so the surface flips back to it and says why.

**"Gone" is deliberately narrow.** It is the daemon `AgentProcessTracker`'s process-truth edge (`agentProcessAlive === false`) and nothing softer:
- **PTY exit is a different event**, and node-pty has been observed reporting a bogus one (#646) — a pane whose shell is very much alive.
- **Activity heuristics cannot distinguish** a long thinking gap from a dead agent.
- **An unattributed pane (`undefined`) is never-observed**, not observed-dying.

The liveness map refreshes on a 15 s poll, so entering Chat takes one first-hand reading of the daemon session list before arming the switch — otherwise a stale `false` would eject the user the instant they re-entered Chat on a just-relaunched agent.

### 2. The toggle now lives on the surface tab's context menu

`feat(chat): toggle Chat View from the surface tab context menu`

Right-click a surface tab to switch that pane between Terminal and Chat. Two things this buys that the segment could not:

- **Panes other than the active one** can be toggled directly. `Cmd/Ctrl+Shift+J` only ever reached the focused pane.
- Chat View is Claude-only in v1, so Codex and Gemini panes previously carried a permanently disabled segment. A menu entry can simply not offer what does not apply.

It rides the existing right-click handling rather than introducing a second menu system — the same surface that already does selection→copy / empty→paste / link menus.

### 3. The tab strip was spending its width on a control most panes cannot use

`refactor(chat): drop the Terminal|Chat segment from the pane tab strip`

With the toggle on the context menu, the segment is pure cost: at four to six panes the strip was wider than the content it labelled. Removing it is a deletion, not a redesign — which is why it lands as its own commit after the menu exists rather than alongside it.

---

Nothing in the Chat View contract moved: the terminal is still the default rendering and the ground truth, and Chat is still a per-surface lens rather than a pivot. Transcript projection, the approval gate and the memory policy are untouched.

openwong2kim:
Two follow-ups pushed on top of this branch, both from dogfooding:

- `0536795e` — the composer's Send button is gone. Enter already sent and Shift+Enter already newlined, so the button was a second way to do the same thing, and it narrowed the field beside it while taking the surface's one warm primary. Soft keyboards have a return key, so nothing is lost on a phone. The key is named in the placeholder now.
- `e7defebc` — agent turns are drawn against a hairline left rail. The human's message is right-aligned in a raised card and the agent's reply had no treatment at all, so a conversation read as a styled message next to unstyled page text. The rail spans the whole turn (prose, folded tool runs, diff chips), so ChatView draws it — it is the only place that sees the sequence. Not a card: an agent turn is long-form markdown and boxing it would squeeze it.

Checks run: `tsc --noEmit` clean, Chat suites 123 passed (3 new for `agentRunFlags`).

**Not verified visually.** A dev instance cannot start while an installed wmux holds the single-instance lock, so neither change has been seen rendered — the rail's weight (`text-main` 12%, against the card border's 10%) is the value most likely to need a nudge on real pixels.

openwong2kim:
## Tool bodies land — plus the 3-way review that reshaped them

Chat View could say a tool call happened but never what it was. `ToolUseEvent`
carried `argSummary` cut at 80 chars; `ToolResultEvent` carried `ok` + `bytes`.
For an agent session that IS the content — one real session here is 473 tool
calls, 315 of them Bash.

The data was never missing. `parseEntry.ts` extracts every input and output
correctly and then drops them. Two commits now carry them.

### Measured on four real transcripts (largest 47 MB)

```
                inline   needs-fetch   no click
tool_use         98.7%       8            98.7%
tool_result      95.0%      30            95.0%
                 (99.2 / 100 / 96.8% across the other three)

widest entry              6,066 B
entries over the PAGE budget (dropped whole)   0
entries over the ENTRY budget (cap failed)     0
fetchable bodies missing srcOffset             0
```

No new RPC, no new IPC channel, no cache: over-cap bodies ride the
`daemon.transcript.codeBlock` path that was already wired end to end, which
re-reads the single transcript line by offset. The existing `eventId` check
still refuses a stale offset after rotation.

### What the three-way review changed

Ran Codex and GLM-5.2 independently against the diff. Three release blockers,
all verified against source before touching anything.

**1. The 4 KB cap did not bound a frame.** Codex and GLM landed on this
separately. One assistant line carries every parallel `tool_use` of a turn, so
a 64-wide fan-out is 256 KB from one line — twice the projector's 128 KB page
budget. And a JSONL line cannot be split: `TranscriptProjector` fits a page by
reading *less*, so an entry that busts the budget alone is skipped entirely and
the whole turn disappears with nothing said. That is a worse version of the
exact silent loss the cap existed to prevent. Bodies now draw from a per-entry
budget and degrade to head+fetch once it is spent.

**2. Nothing over the cap could ever be expanded.** `onFetchBody` resolved refs
only on `assistant_text.codeBlocks`, so every tool body failed with "no source
offset". Now resolves by event kind.

**3. `Buffer` does not exist in the renderer.** I derived "is this only a head?"
with `Buffer.byteLength` in `ToolRunLine`; with `nodeIntegration: false` that is
a `ReferenceError` on expanding any run, not a wrong number. The producer now
states it as `ToolBody.truncated`.

Worth recording: **GLM disputed #3**, arguing `Buffer` was already used in the
renderer. It is not — the hits it cited are a local `inputBuffer` variable and
xterm's `Buffer.line` API. The only real Node `Buffer` under `src/renderer` was
the line this removes. Cross-model disagreement resolved by reading the code,
not by vote.

Also from that pass: the side-table body is now cut on a UTF-8 boundary rather
than by UTF-16 character count (a fetched body could otherwise end in half a
surrogate pair), and a test that claimed to cover an unserializable tool input
built a cyclic object it never used and then parsed ordinary JSON — replaced
with cases that actually run.

### Verification

- 4 real transcripts through the real parser, largest 47 MB — table above
- Projector end-to-end: an over-cap body fetched back, and the rotation guard
  confirmed to refuse a mismatched `eventId`
- Per-entry budget is **mutation-verified** — removing it turns the test red
- Packaged build: artifacts grepped, daemon bundle confirmed to carry the new
  parser, app boots and answers RPC on the real `~/.wmux` daemon
- 419 tests green across the touched surface; tsc and lint clean

### Not verified, and it needs you

The final visual pass — expand a tool run in the running app and read a bash
output — needs a real agent turn in the packaged build. The dev build cannot do
it (the hook bridge is hardcoded to the unsuffixed `~/.wmux`, so hooks never
reach a `-dev` instance), and the packaged build exposes no CDP to drive. The
build is running now if you want to look.

### Still open

**Approvals.** The acceptance bar was to drive a session from Chat View alone,
and that still fails at every permission prompt. My earlier plan to wire
approve/deny buttons to `daemon.approvals.resolve` was wrong and Codex caught
it: the registry only ever acts on hook-sourced `AskUserQuestion`, and Claude's
tool-permission prompts have no hook at all — `approvalKeystrokes.ts` says so
in as many words, and they are deliberately excluded.

The way through is real and unused: Claude Code has a **`PermissionRequest`
hook** that fires exactly when that prompt is shown, carries `tool_name` +
`tool_input`, and can return `allow`/`deny`. wmux wires zero of it today and
screen-scrapes instead. The blocker is that `wmux-bridge.mjs` caps every hook at
`HOOK_TIMEOUT_MS = 2000` "so we never slow Claude" — which cannot wait for a
human. That rationale does not apply to a prompt where Claude is *already*
blocked on one, so the fix is a bounded wait that falls through to the TUI
prompt on timeout. Filed separately rather than bolted on here.

openwong2kim:
@codex review

Focus on the last two commits (`5be1a414`, `ff7a5850`) — tool call input/output bodies in Chat View.

Specifically worth a hard look:
- The per-entry inline budget (`INLINE_ENTRY_MAX_BYTES`). It exists because a per-body cap does not bound a frame, and an entry that busts the projector's 128 KB page budget is dropped whole. Is 32 KB actually safe once ids, names, summaries and the JSON envelope ride alongside, and is the degrade-to-fetch path correct when the budget runs out mid-entry?
- `onFetchBody` in `Pane.tsx` now resolves refs by event kind. Any kind that can carry a fetchable body and is still missed?
- `ToolBody.truncated` is produced daemon-side because the renderer has no `Buffer`. Is it set correctly for every path — over-cap, budget-exhausted, and `MAX_BODY_BYTES`-clipped?
- UTF-8 boundary slicing in `sliceToBytes`, and whether the side-table body and the inline head can ever disagree about what the reader is being shown.
- Renderer retention: fetched bodies live in component state and inline bodies live in the 2000-row store cap. Is there a pane-count scenario where that is a problem?

openwong2kim:
## Live dogfood — packaged build

Ran the packaged build (`out/wmux-win32-x64/wmux.exe`, 213 MB, daemon bundle
verified to carry the new parser) against the real `~/.wmux` daemon.

**Boots clean, recovers 5 sessions, Chat View renders prior conversation.**

### What the daemon would not let me automate, and why that is correct

I tried to drive the check from a CLI client and every pane answered:

```
pane daemon-864cc9c7: unavailable — reason=not-authorized
pane daemon-5ec541bb: unavailable — reason=not-authorized
… 5/5
```

That is `firstPartyOnly(ctx.clientId, 'status')` doing its job — transcript
content is main-only, and a CLI socket is not main. Worth stating plainly
because it is also the reason a scripted end-to-end pass over this surface is
not possible by design, so the visual confirmation has to be a human at the
window.

### What I could verify without the GUI

**Four real transcripts through the real parser** (largest 47 MB):

```
                     inline   fetch    no click
tool_use              98.7%     8        98.7%
tool_result           95.0%    30        95.0%
   other three:  99.2 / 100 / 96.8%  ·  95.3 / 97.8 / 91.9%

widest single entry                            6,066 B
entries over the PAGE budget (dropped whole)        0
entries over the ENTRY budget (cap failed)          0
fetchable bodies missing srcOffset                  0
```

**Projector end-to-end**: an over-cap body fetched back through
`daemon.transcript.codeBlock`, and the rotation guard confirmed to refuse a
mismatched `eventId`.

**Packaged artifacts**: `called with` / `returned` / `show all` / `truncated`
all present in the asar; `INLINE_ENTRY_MAX_BYTES` and the `<tool_use_error>`
unwrap present in the daemon bundle.

### One false alarm worth recording

A full local run showed 12 failures in `diff.handler.test.ts` and
`mergeSession.test.ts` — neither touched by this work. Cause was
`EBUSY: resource busy or locked, rmdir …\wmux-diffh-…\repo`: the packaged wmux
I had just launched was holding the temp git repos those tests create. With the
app stopped, 42/43 pass. CI on this exact head is 646/648 green.

Filing it because the failure looks like a real regression in an unrelated
subsystem, and the next person to run the suite with wmux open will see the
same thing.

### Left for you

Expand a tool run in the window and read a bash command with its output. That is
the one assertion no script here can make.

openwong2kim:
Both P1s from your pass are fixed in `f69024f3` — thanks, the budget one was the real bug.

**Serialized-cost budget.** You were right that raw bytes was the wrong meter. Worth adding one thing I found chasing it: the 6x blow-up only happens on the tool **result** path. A result body is raw program output so its control characters survive to the wire, but a `tool_use` input is `JSON.stringify`d before it is stored and therefore arrives pre-escaped. My first regression test used the input path and **passed under the bug** — mutating the fix back to raw bytes did not turn it red. Corrected to the result path with your exact eight-NUL case, and it now fails without the fix.

**Starved fetch handle.** Confirmed and fixed: `ToolBodyBlock` returned null on empty content, which took the expand button with it, so every later call in a wide fan-out was unreachable despite the daemon still holding the body. Now only the `<pre>` is conditional; the label and fetch control render whenever the body is truncated.

On the rest — most are pre-existing in this PR rather than from these three commits, and I have not touched them:

- `ChatComposer.tsx:68` (deferred-submit gate reads a ref updated in a passive effect) is a real P1 and is the same A7 window the composer lock exists for. Separate fix; it predates this work.
- `TranscriptProjector.ts:341` (unvalidated `resumeBinding.transcriptPath` via main's `hooks.signal` fallback and the direct `setResumeBinding` RPC) — also pre-existing and worth its own change, since the guard belongs on every write path, not just daemon-side HookIngest.
- `foldToolRuns.ts:121` (diff-shaped result flushes the run before recording it, so the matching call stays `running` forever) is adjacent to what I touched and I can fold it in if you want it here rather than separately.
- Scroll-to-tail, keep-warm order on close, `hasImage`, oversized code-block truncation, and subscription teardown on renderer death are all pre-existing.

Happy to take any of those in this PR if that is the preference — flag which.

openwong2kim:
## RAM: what the inline tool bodies actually cost

Measured, not estimated. Parsed six real transcripts, took the last 2000 events
exactly as the renderer store windows them (`MAX_ROWS_IN_MEMORY`), and measured
the retained payload with bodies against the same events with bodies stripped —
i.e. the shape before these commits.

```
                                events   before    after     delta
59bbbc2b…                        1186    0.24 MB   1.29 MB   +1.05 MB
7ee5807a…                         601    0.15 MB   0.40 MB   +0.24 MB
4e9eca98…                         934    0.24 MB   1.13 MB   +0.89 MB
d20f65ac…                        2000    0.48 MB   2.15 MB   +1.67 MB
cd250981…                         238    0.04 MB   0.24 MB   +0.20 MB
c1a6222a…                         990    0.19 MB   1.24 MB   +1.05 MB

per pane (avg)                           0.22 MB   1.07 MB   +0.85 MB
30 panes                                 6.7  MB   32.2 MB   +25.5 MB
```

Strings counted at two bytes per char (V8's upper bound, and the one that
matters for a CJK-heavy session).

### Correcting my own number

The eng review estimated **+0.4 MB/pane**. The real figure is **+0.85 MB** —
about double. I used that estimate to argue the 30-pane measurement should not
gate this work, so half the basis for that call was wrong. The conclusion still
holds at 30 panes (+25 MB is affordable), but the reasoning was weaker than I
presented it.

### The open one: the cap counts rows, not bytes

`MAX_ROWS_IN_MEMORY` is 2000 **rows**. Nothing bounds their size, so the
theoretical ceiling is now:

```
2000 rows × 4 KB inline × 2 bytes/char  =  15.6 MB/pane  ·  0.5 GB across 30
```

The measured worst real session is **2.15 MB** — about a seventh of that — so
this is a ceiling, not a present problem. But there is no structural reason a
session of nothing but large `Bash` output could not walk toward it. Codex
flagged the same thing as P2 on `chatSlice.ts:30`.

A per-pane byte budget of ~4 MB would not touch any session measured here and
would bring the 30-pane worst case from 0.5 GB to 120 MB.

**Deferred by owner decision for now** — recorded here so the ceiling is on the
record rather than rediscovered later.

### Still unrun

The 30-pane **real Electron RSS** measurement. The numbers above are the store
payload, which is the honest delta for this change, but total process RSS also
carries GPU and xterm instances. That measurement is the baseline the keep-warm
LRU ships default-OFF until it passes, and it remains open independently of
this work.

openwong2kim:
## The §8 memory measurement — 20 panes, macOS, packaged build

The remaining merge gate was the multi-pane RSS measurement. Ran it on macOS (arm64) against a packaged build of this branch (`f69024f3`), under `WMUX_DATA_SUFFIX=-cvrss` so the production instance kept running untouched. **20 panes, not 30 — the app itself caps a workspace at 20 panes** (`pane.split` answers "pane cap reached (max 20 per workspace)"), so 20-in-one-view is the largest state the product can reach; a 30-pane single view does not exist to measure.

Setup: 20 panes in one workspace, each session bound to a **real** Claude transcript (1–47 MB `.jsonl`, via `daemon.setResumeBinding`, all 20 probing `available: ok`), toggled through the real `setSurfaceChatMode` path. Each state settled 60 s before sampling; `__wmuxChatMemory.sample` (5 reads, 1 s apart); no DevTools open — this run inherits the earlier finding that DevTools was the fake leak.

| state | mean | median | min–max | chatPanes |
|---|---|---|---|---|
| 20 terminals (baseline) | 410.0 MB | 409.9 | 409.9–410.3 | 0 |
| 20 chat, LRU **off** | 491.1 MB | 474.9 | 471.8–517.3 | 20 |
| 20 chat, LRU **on** | 456.9 MB | 456.7 | 454.4–459.7 | 20 |
| 20 terminals (return) | 387.5 MB | 384.8 | 384.7–391.7 | 0 |

### Reading it

- **No leak.** Returning to terminals lands *below* the original baseline (387.5 vs 410.0). Whatever chat mode allocates, it gives back.
- **Chat with the keep-warm LRU on costs ~47 MB over baseline at 20 panes** (~2.3 MB/pane). The LRU-on row is also the tightest spread (454–460), i.e. genuinely settled.
- **The LRU earns its keep: −34 MB vs LRU off** (456.9 vs 491.1), and the LRU-off row's wide spread (max 517) is the toggle burst still draining — its true settled cost is closer to its median 475.
- **§8 verdict as written: chat mean (456.9) > terminal mean (410.0) → the criterion does not pass, so `chatKeepWarmLru` ships default-OFF stays correct** — but note the criterion's sign is doing the opposite of what the numbers suggest: the LRU *reduces* chat cost, it just doesn't get chat all the way under the terminal baseline. At ~2.3 MB/pane the delta is bookkeeping, not a blocker.

### Caveats

- 20 panes in one workspace means ~10-col terminals — smaller xterm buffers than a realistic fleet, so per-pane terminal cost is understated, which *flatters* the terminal baseline. A realistic-geometry fleet would narrow the chat-vs-terminal gap.
- LRU-on was sampled after LRU-off in the same session (order effect possible, though the return-to-baseline row bounds it).
- macOS this time; the earlier 8-pane run was Windows. Same shape of result: bounded cost, no retention.

Automation note for reproducing: the store isn't reachable from the console, so this run drove `setSurfaceChatMode` via a throwaway local patch exposing the store (not committed). The `⌘⇧J` chord path was also verified live — handler fires and gates on `chatStatus` exactly as documented.

openwong2kim:
## Owner decision — parking the surface, landing the pipeline

Live dogfood verdict on the surface: for a short session the lens is mostly machinery rows around two lines of conversation, and the local-command reality (/usage renders only in the TUI, its output never reaches the transcript) makes the composer feel half-connected. The desktop Chat surface does not currently earn its 15k lines.

So this PR splits:

- **The daemon half lands separately as #771** — TranscriptProjector / parseEntry / TranscriptDiscovery, transcriptPathGuard, the DaemonPipeServer unicast + first-party + backpressure work, the daemon.transcript.* RPCs, HookIngest vetting + nudge, and all their tests, verbatim from this branch (including the slash-command label and Skill-summary fixes from defebe2c). No consumers, no behavior change; it stops this work rotting against main.
- **The renderer half stays parked on this branch.** If a conversation surface earns its way back — most likely as the phone's native chat, for which the projection pipeline is the foundation — it restarts from here rather than from zero.

Leaving this open as the parked renderer track; #771 carries the rest forward.

openwong2kim:
Closing by owner decision. The daemon half landed as #771; the renderer surface did not earn its size in live dogfood (machinery rows outweighed conversation on short sessions, and TUI-only command output can never reach the projection). If a conversation surface earns its way back — most likely as the phone's native chat on top of the #771 pipeline — it restarts from this branch.

PR #656, opened 2026-07-27
test,ci: make the git context watcher tests deterministic and retry apt

openwong2kim:
## Why

macOS Baseline has been failing intermittently on `gitContextWatch > emits again when HEAD changes (branch switch)` and passing on rerun. The watcher itself is correct — `arm()` runs before `readAndEmit`, so there is no arming race. The failure is inherent macOS FSEvents delivery latency plus the 50 ms re-read debounce racing a fixed 2 s timeout in the test helper. On a contended shared runner the event simply arrives late.

Reruns are not a fix: they hide the signal and cost a full cross-platform matrix every time.

## What changed

**Deterministic watcher tests.** `GitContextWatcher` takes an optional watch factory (defaulting to `fs.watch`), so the tests can drive the watch callback directly and use fake timers for the debounce. Production behavior is unchanged — one call site now goes through the injected factory.

Coverage went up rather than down. The suite now asserts the debounce boundary on both sides, burst coalescing, `HEAD.lock` acceptance with `config`/`index`/`refs`/`ORIG_HEAD` filtered out, identical-rewrite dedup, repo→non-repo transitions, `.git` appearance re-resolve including the old watch being closed and the new target re-armed, and debounce cleanup on `remove()`/`dispose()`. The two wall-clock sleeps that stood in for "nothing else was emitted" are gone.

One real-filesystem end-to-end test remains, proving the default `fs.watch` wiring re-emits on a real HEAD rewrite. It lives under `vitest.runtime.config.ts`, where the other real-I/O serial tests already are, with a generous timeout.

**apt retry on Linux runners.** Both apt call sites are wrapped in a bounded retry with backoff. `Acquire::Retries` only retries a fetch inside a single apt run, so it cannot survive a mirror returning a hash-sum mismatch or resetting the connection. Existing hardening (pinned actions, permissions, third-party source removal) is untouched.

**Two census fixes.** A 2 s "safety" timer in `DaemonPipeServer.test.ts` was actually the assertion's deadline, and a 1700 ms sleep in `deck.handler.loop.test.ts` gated a positive assertion behind a 1500 ms debounce. Both now wait on the condition.

## Not changed, deliberately

The Windows `saveDebounced` flakes in `StateWriter`/`ChannelStateWriter` were already fixed on 2026-07-18 by the `waitForCondition` helper — last occurrence 2026-07-14, no recurrence since. `SessionPipe.attachSnapshot` was likewise fixed on 2026-07-23. Re-touching those files would be churn.

## Test plan

- `tsc --noEmit`, `vitest run src/main src/shared` (3561 passed), targeted runs per commit, YAML parse plus `bash -n` on the changed workflow blocks.
- Determinism: the hardened files run **20/20** consecutive green on both the parallel and runtime configs.

## Follow-ups

A ranked census of the remaining load-sensitive sleeps is in the branch discussion — the largest clusters are `DaemonClient` pipe-settle sleeps, `DeckBriefingCard` real sleeps inside `act()`, and an 18-sleep cluster in `deck.handler.loop`. Converting those is a per-file pass, intentionally out of scope here.

Separately, two genuine cross-platform bugs surfaced while sampling failure logs (Windows 8.3 path handling in `diff.handler`/`worktree.handler`, and `execEnv`). They are unrelated to flakiness and are not addressed here.

## Summary by CodeRabbit

* **Reliability**
  * Improved Linux build and release reliability by retrying package installation steps with backoff when network/package sources are temporarily unavailable.
  * Improved Git context monitoring so branch/`HEAD` changes are detected more consistently, including updates to `HEAD` content.

* **Tests**
  * Reduced timing-related flakiness by switching from fixed delays to condition-based waiting for connection caps and concurrent turn dispatch.
  * Expanded Git context watcher coverage with both deterministic unit tests (controlled watch behavior) and a runtime end-to-end filesystem test.

### DIFF
--- a/.github/workflows/ci-cross-platform-baseline.yml
+++ b/.github/workflows/ci-cross-platform-baseline.yml
@@ -84,8 +84,10 @@ jobs:
               if sudo apt-get -o Acquire::Retries=3 "$@"; then
                 return 0
               fi
-              echo "apt-get $* failed (attempt ${attempt}/3); retrying after $((attempt * 15))s"
-              sleep $((attempt * 15))
+              if [ "$attempt" -lt 3 ]; then
+                echo "apt-get $* failed (attempt ${attempt}/3); retrying after $((attempt * 15))s"
+                sleep $((attempt * 15))
+              fi
             done
             echo "apt-get $* failed after 3 attempts" >&2
             return 1
--- a/.github/workflows/ci-cross-platform-baseline.yml
+++ b/.github/workflows/ci-cross-platform-baseline.yml
@@ -69,9 +69,30 @@ jobs:
           # azure-cli.list, …), so a filename glob misses some. Matching the
           # host also survives a runner bump to deb822 (24.04 keeps the Ubuntu
           # archives in sources.list.d/ubuntu.sources, which must NOT be dropped).
+          #
+          # Dropping the third-party sources removes the deterministic 403, but
+          # the Ubuntu/Azure mirrors themselves still fail transiently (Hash Sum
+          # mismatch, connection reset, 5xx from a single mirror behind the load
+          # balancer). Acquire::Retries only retries the individual fetch inside
+          # one apt run; a mirror serving a stale index fails the whole run. Wrap
+          # both apt operations in a bounded retry with backoff so a slow/soggy
+          # mirror costs seconds instead of a red X on an unrelated PR.
+          set -euo pipefail
+          apt_retry() {
+            local attempt
+            for attempt in 1 2 3; do
+              if sudo apt-get -o Acquire::Retries=3 "$@"; then
+                return 0
+              fi
+              echo "apt-get $* failed (attempt ${attempt}/3); retrying after $((attempt * 15))s"
+              sleep $((attempt * 15))
+            done
+            echo "apt-get $* failed after 3 attempts" >&2
+            return 1
+          }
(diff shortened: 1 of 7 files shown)


openwong2kim:

This PR hardens CI and test timing on three fronts with no production behavior change: it makes the `GitContextWatcher` tests deterministic by injecting a fake `fs.watch` factory (defaulted to real `fs.watch`, with one real-fs end-to-end test relocated to the serial runtime config), wraps Linux apt operations in a bounded 3-attempt retry with linear backoff in both the cross-platform baseline and the release Linux build, and replaces two load-sensitive fixed sleeps — the `DaemonPipeServer` connection-cap 2s safety timer and the `deck.handler.loop` 1700ms wake — with condition-based polling. It directly removes the macOS Baseline flake on "emits again when HEAD changes" (inherent FSEvents delivery latency racing a fixed 2s timeout, not a watcher bug), absorbs transient Ubuntu/Azure apt mirror failures that would otherwise drop a `.deb/.rpm/.AppImage` from a release, and clears two census-flagged timing hot spots. The scope is tight and well-justified: the only production-path edit is the clean constructor-injected test seam in `gitContextWatch.ts`, coverage went up rather than down, and the hardened files reportedly run 20/20 consecutive green; the deferred items (additional sleep clusters, two unrelated cross-platform bugs) are correctly left out. Priority P3 — valuable CI/test hygiene with no user-facing or runtime impact.

PR #660, opened 2026-07-27
test: deflake brick-recovery TTL prune (fixture pid could be alive on CI)

openwong2kim:
## Summary

`v281BrickRecovery.integration.test.ts` broke two consecutive main pushes today (validate on the #653 merge, Windows Baseline on #656) with the same assertion: an expired dead session that should have been pruned was still there.

Root cause: the fixture pins `pid: 4242`. Since #646, `StateWriter.load()` keeps an expired dead tombstone alive as long as its recorded pid still runs, so the test's outcome depended on whether the CI runner happened to have a live pid 4242 — a coin flip that came up tails twice in a row today and will keep recurring.

- The pruned fixture now uses `pid: -1`, which `isPidAlive` rejects unconditionally, making the prune deterministic.
- Added the counterpart test locking #646's intended behaviour: an expired tombstone with a live pid (`process.pid`) survives the prune.

No production code changes.

## Test plan

- [x] `vitest run v281BrickRecovery.integration.test.ts` — 4 passed (was 3)
- [x] `tsc --noEmit` clean, touched file lint clean

## Summary by CodeRabbit

* **Tests**
  * Improved regression coverage for cleanup of expired dead sessions.
  * Added validation that expired records are retained while their associated process is still running.
  * Reduced test flakiness by explicitly setting process identifiers in test scenarios.

### DIFF
--- a/src/daemon/__tests__/v281BrickRecovery.integration.test.ts
+++ b/src/daemon/__tests__/v281BrickRecovery.integration.test.ts
@@ -174,7 +174,11 @@ describe('v2.8.1 brick recovery — full Bug 1 flow', () => {
     writer.saveImmediate({
       version: 1,
       sessions: [
-        session('dead-tight', 'dead', new Date(now - 25 * HOUR).toISOString()),
+        // pid -1 can never be alive. The fixture's default pid (4242) is a
+        // real pid on some CI runners, and #646's keep-expired-tombstones-
+        // while-the-pid-lives rule then retains this session and flakes the
+        // assertion below.
+        { ...session('dead-tight', 'dead', new Date(now - 25 * HOUR).toISOString()), pid: -1 },
         {
           ...session(
             'suspended-tight',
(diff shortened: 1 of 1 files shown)


openwong2kim:

This PR deflakes `v281BrickRecovery.integration.test.ts`, which broke two consecutive main pushes (validate on the #653 merge, Windows Baseline on #656) because its fixture pinned `pid: 4242` — a real pid on some CI runners. Since #646 made `StateWriter.load()` retain an expired dead tombstone as long as its recorded pid is still alive, the prune assertion became nondeterministic, flipping between pass and fail depending on whether the runner happened to have a live process with that id. The fix sets the to-be-pruned fixture to `pid: -1` (which `isPidAlive` rejects unconditionally, making the prune deterministic) and adds the counterpart regression test locking in #646's intended behaviour using `process.pid` so an expired tombstone with a genuinely live pid survives the prune. No production code is touched; the change is confined to one test file, the comment explaining the rationale is accurate, and the two new assertions exercise both sides of the #646 rule directly.

**Priority: P3** — test-only reliability fix, no runtime impact, but worthwhile to merge promptly so it stops red-lighting main on coincidental pid collisions.

Label: `P3`.

Issue #659, opened 2026-07-27
control pipe multiplexes broadcast events with RPC responses, with no subscription and no warning — naive clients report phantom failures with empty error messages


Kbooher2:
## Summary

The daemon control pipe (`\\.\pipe\wmux-daemon-<user>`) delivers **unsolicited broadcast events down the same socket as RPC responses**, with no subscription step and nothing in the docs warning a client about it. A client that writes a request and reads one line back — the obvious implementation — intermittently reads a broadcast instead of its response.

Because a broadcast frame carries neither `ok` nor `error` nor `result`, the naive client reports a **spurious failure with an empty error message**, and silently discards the real response when it arrives.

## Why this is worth a fix rather than a shrug

Two independent teams, working on separate hosts, hit this within hours of each other while pairing over `lanlink` (#658). **Both published incorrect root-cause analysis off the back of it** and had to retract — one of them attributed an empty error string to a fault "below the lanlink code" that did not exist. Between them it burned several live pairing windows with people waiting.

Note the asymmetry that makes it nasty: this failure mode can **fabricate failures but never successes**. So it does not look like a client bug. It looks like the daemon intermittently failing whatever you just asked it to do, with no error text to go on — which is exactly the wrong direction to send someone debugging.

## What the daemon does today

Responses are correlatable — every path echoes the request id:

```js
JSON.stringify({ id: request.id, ok: false, error: "unauthorized" })
JSON.stringify({ id: request.id, ok: false, error: "rate limited" })
this.dispatch(request).then((response) => …)
```

Broadcasts are not, and are pushed to every connected client unconditionally:

```js
broadcast(event) {
  const msg = JSON.stringify(event) + "\n";
  this.connectedSockets.forEach((socket) => {
    if (!socket.destroyed) { try { socket.write(msg); } catch {} }
  });
}
```

So an RPC-only client is pushed `{"type":"title.changed","sessionId":…}`, `lanlink.remote.received`, and similar frames it never asked for, interleaved with its own replies. Correlation *is* possible — the id is there — it is simply undocumented and unenforced, so the correct client is the non-obvious one.

## Suggested fixes, cheapest first

1. **Document it, prominently.** "The control pipe is multiplexed. Clients MUST read lines until `resp.id` matches the id they sent, and MUST ignore frames without an `id`." One paragraph would have saved both teams hours.
[...]

PR #662, opened 2026-07-27
fix(lanlink): commit a pairing atomically so a failed save cannot half-pair

### LINKED ISSUE #658: lanlink pairing leaves a permanent half-pair: responder persists the peer, initiator always fails with `connection closed` awaiting `AEAD_RECORD`
## Summary

`lanlink` pairing never completes on the **initiator** side. The responder completes the PAKE, persists a peer record, and sends its confirming AEAD record; the initiator dies at `await stream.next(AEAD_RECORD)` with `connection closed` and never reaches `upsertPaired`.

The result is a permanent, silent **half-pair**: the responder holds a peer the initiator has no secret for. Afterwards:

- initiator → responder: fails locally at `sendToPeer: unknown or burned peer <uuid>` (never opens a socket)
- responder → initiator: the initiator's `onReconnectHello` hits `secretOf()` → `null` → silent `destroy(conn)`

So peer-to-peer messaging is unusable in both directions, and the state is unrecoverable from the initiator side — the long-term secret only ever existed on the responder.

Reproduced independently by two operators on separate hosts, in both role directions, on two builds.

## Affected versions

**3.29.0** and **3.31.0** (Windows). Not tested on 3.37.1 — noting only that nothing in the 3.32.0 → 3.37.1 release notes appears to touch `lanlink`.

## Reproduction

Two hosts on a LAN, A and B. On both: `lanlink.configure { enabled: true, port: 45651 }` bound to the LAN NIC, host firewall permitting the peer.

1. On A: `lanlink.pair.begin` → 6-digit PIN. Confirm `lanlink.pair.status` shows `active: true`.
2. On B, within the 120s TTL: `lanlink.pair.join { host: <A>, port: 45651, pin: <PIN> }`
3. `lanlink.peers.list` on both.

**Expected:** both hosts hold a peer record.
**Actual:** A (responder) holds a record. B (initiator) holds none; its RPC returns `{"ok": false, "error": "connection closed"}`.

Swapping roles reproduces it with the roles reversed — whichever host is *responder* persists, whichever is *initiator* does not.

## Evidence

Four pairings across both role permutations and both version pairings:

| initiator | responder | responder persisted | initiator persisted |
|---|---|---|---|
| 3.29.0 | 3.31.0 | yes (×2) | **no** |
| 3.31.0 | 3.29.0 | yes (×2) | **no** |

Four responder commits, zero initiator commits. Fully symmetric, direction- and version-independent.

Initiator RPC elapsed: **~716ms** on a fresh window (both sides complete scrypt), ~98ms against an expired window. Responder-side `failCount` stayed `0` throughout — `noteWindowFailure()` never fired, i.e. **the PAKE itself succeeds**. The orphaned peer records show `burned: false` and `sendSeq: 0`, and `sendSeq` never advances, confirming `RECONNECT2` never comes back.

## Analysis

Initiator, `pairWithPeer()`:

```js
socket.write(encodeFrame(CONFIRM, confirm));
const firstAead = await stream.next(AEAD_RECORD);   // <-- throws "connection closed"
const respMac   = new AeadOpener(pending.sessionKeys.s2cKey, 2).open(firstAead);
const result    = ini.verifyRespMac(respMac, pending);
opts.peers.upsertPaired(result);                    // <-- never reached
} finally { socket.destroy(); }
```

Responder, `onConfirm()`:

```js
let r;
try { r = conn.pairing.onConfirm(body); }
catch { this.noteWindowFailure(); this.destroy(conn); return; }
this.deps.peers.upsertPaired(r.result);                       // commits BEFORE confirming to peer
this.establishAead(conn, r.result.peerUuid, r.sessionKeys);   // clears the 5s handshake deadline
this.send(conn, AEAD_RECORD, conn.sealer.seal(r.respMac));    // then sends; no destroy on this path
```

### The frame never arrives, and the socket is actively closed

`FrameStream.next(expected)` has three distinct failure texts, which lets us discriminate:

```js
done = (f) => f.type !== expected
  ? reject(new Error(`expected frame 0x${expected.toString(16)}, got 0x${f.type.toString(16)}`))
  : resolve(f.body);
setTimeout(() => reject(new Error("LanLink client: frame timeout")), FRAME_TIMEOUT_MS);
socket.on("close", () => this.fail(new Error("connection closed")));
```

We get **`connection closed`** — not a type mismatch, and not a frame timeout. So the AEAD record never reached the initiator's reader *and* the socket closed, well inside the frame timeout.

### It is not the network

On the same socket, one round-trip earlier, **`PAKE2` travels responder → initiator successfully** — the initiator can only produce a valid `CONFIRM` because it received and processed `PAKE2`. The very next frame in the same direction on the same connection never arrives. That rules out a one-way block, a firewall/MTU/overlay issue, or a routing asymmetry, and localizes the fault to the responder's post-commit send path.

### Suspect: any throw after the commit closes the connection silently

Two things on that path swallow failures into a close:

- `send()` catches a write throw and calls `this.destroy(conn)`:
  ```js
  send(conn, type, body) {
    if (conn.state === "dead" || conn.socket.destroyed) return;
    try { conn.socket.write(encodeFrame(type, body)); } catch { this.destroy(conn); }
  }
  ```
- `conn.sealer.seal(r.respMac)` is evaluated as an **argument**, i.e. *outside* that `try` — so a throw there propagates out of `onConfirm` to the frame dispatcher, which also destroys the connection.

Either produces exactly the observed signature: peer persisted, socket closed, nothing on the wire. Worth instrumenting that path rather than the crypto, which verifiably succeeds.

## Two design issues worth fixing regardless of root cause

1. **The responder commits before the handshake is mutually confirmed.** Because `upsertPaired` precedes the send, *any* failure after it leaves a one-sided record with no way to detect or recover it. Committing after mutual confirmation — or reconciling on the first `RECONNECT_HELLO` miss — would make this failure recoverable instead of permanent.
2. **Every rejection path is a silent `destroy(conn)` with no error frame.** Expired window, wrong PIN, unknown peer, admission-cap rejection, and version mismatch are all indistinguishable to the initiator; so is this bug. Diagnosing it took two operators several hours, almost entirely because of that. A typed error frame before `destroy` would have made it minutes.

## Ruled out

- **Version mismatch.** Every pairing constant matches across 3.29.0/3.31.0 (`WIRE_VERSION`, `MIN_WIRE_VERSION`, `CIPHER_ID`, accepted versions/ciphers, `MAX_FRAME`, `LEN_BYTES`, `PIN_LEN`, `PAIR_TTL_MS`, `PAIR_FAIL_BURN`, `SCRYPT_PARAMS`), as does the entire AEAD leg (chacha20-poly1305, key/nonce/tag lengths, nonce layout, AAD bytes, counter semantics, direction ids, record layout, HKDF hash/salts/labels, respMac label). `pairWithPeer()` is byte-identical between the two builds, and the responder path quoted above was read independently out of both.
- **Peer-uuid derivation.** Both derive `sha1(LANLINK_NS || initiator rawJPub32)`, `LANLINK_NS = "lanlink-peers-v5"`, same `0x50`/`0x80` variant bits.
- **The 5s handshake deadline.** `establishAead()` clears `conn.deadline` before the send, and the observed failures land at ~716ms.
- **Window timing, `pakeInFlight`, and the per-IP connection cap.** Verified `active: true` immediately before firing, with a >60s idle cooldown and zero live connections to the peer.
- **The peer store and its ACL hardening.** It writes correctly whenever it is reached.

## Environment

Two Windows Server hosts on the same private LAN segment (routed over a WireGuard-based overlay), direct TCP on 45651 verified in both directions, no proxy or NAT between them. Node bundled with the app.

Disclosure: on one host the `[lanlink-firewall]` `netsh advfirewall` rule application failed and was logged (`Public/Domain BLOCK rule could not be applied — relying on Public-category refusal + pairing window`). That host still accepted inbound pairing connections and completed the PAKE as responder, and `PAKE2` traverses the same direction on the same socket immediately before the failing frame — so it is disclosed for completeness, not offered as a cause.

openwong2kim:
Closes #658. Also lands the docs half of #659.

## The bug

The responder added the peer to its in-memory map and *then* persisted it. When `persist()` threw, the record stayed in memory, the throw unwound out of `onConfirm` into `pump()`'s broad catch, and the connection was destroyed before the confirming AEAD record went out.

`listPeers()` reads memory, so the responder reported a healthy pairing over a file that was gone, while the initiator died at `await stream.next(AEAD_RECORD)` with `connection closed`, never reached its own `upsertPaired`, and held no secret at all. Unrecoverable from that side — the long-term secret only ever existed on the responder. Because the responder closes first, it presents identically in both role directions, which is why the reporters saw it symmetrically.

**On the reported root cause:** the report suspected `conn.sealer.seal(r.respMac)` being evaluated outside `send()`'s `try`. That is structurally true and is fixed here, but `AeadSealer.seal` can only throw on counter overflow and the counter is 1 — it is not the cause. The throw is in the peer-store commit one line earlier. `persist()` has more than one throw branch on win32 (the C12 fail-closed unlink; a failed rename in `atomicWriteJSONSync`, which leaves no primary file because the previous one was already moved to `.bak`). **Which branch fired on the reporting hosts is not established, and this PR does not claim it** — the fix covers all of them.

## The fix

`upsertPaired` deep-snapshots the map, mutates, persists, and restores on any throw. Deep because other mutators edit records in place.

**Only `upsertPaired` rolls back.** Rolling back `revoke()`, `noteSteadyStateAuthFail()` or `bumpHighWater()` would be fail-*open*: it would restore access the caller asked to remove, stop a host that cannot persist from ever reaching `PEER_BURN_THRESHOLD`, and reopen the C8 replay gate for an authenticated-but-hostile peer. Granting access is the only direction where memory running ahead of disk is the dangerous one. `revokePeer()` now cuts live connections in a `finally` so a store throw cannot skip the C13 half that takes effect immediately.

The joiner's `connection closed` at that exact point now names the two real possibilities. **Deliberately not a wire error frame** — the responder has no authenticated channel there, and a cleartext one would be a PIN oracle and forgeable by anyone on the segment.

## Scope

Fixes the *silent, permanent half-pair*, not whatever made the write fail. On a host where the peer file cannot be written, pairing still fails — but now it fails on both sides and says so. Making the write itself succeed is a follow-up that needs the reporters' daemon logs to pick the branch.

Known limits and pre-existing issues found while reviewing this are filed separately: revoke/burn not surviving a restart via the `.bak` fallback, the mirror-image half-pair when the *joiner* fails to save, and `noteSeen()` persisting on every inbound record.

## Verification

- `tsc` on the root, daemon and cli projects
- `eslint src/daemon/lanlink` — 0 errors
- 8726 + 88 tests pass
- New: rollback across both the ACL branch and a failed atomic write; non-rollback for revoke and burn; end-to-end proof that a refused commit pairs neither side and reports something other than `connection closed`

## Summary by CodeRabbit

* **Bug Fixes**
  * Pairing now completes atomically, preventing partially saved pairings when persistence fails.
  * Failed pairings provide clearer connection-closed messages and likely causes.
  * Revoked or unsuccessfully paired connections are reliably terminated.
  * Improved handling of control-stream messages prevents pushed events from being mistaken for replies.

* **Documentation**
  * Added protocol guidance explaining how to distinguish responses from unsolicited events on multiplexed connections.


openwong2kim:
@codex review

openwong2kim:

This PR closes the permanent half-pair from #658 by making `PeerStore.upsertPaired` atomic — it deep-snapshots the in-memory map, mutates, persists, and restores the snapshot on any throw, so a failed write can no longer leave one side holding a peer the other has no secret for. The scope discipline is the right call: it fixes the *symmetry break* (memory running ahead of disk) rather than guessing at which `persist()` throw branch fired on the reporters' hosts, and it deliberately does not roll back `revoke()`/`noteSteadyStateAuthFail()` where fail-open would be worse. The joiner-side `connection closed` rewrite into a human-readable message, the `revokePeer` `finally` block (so a store throw cannot skip C13 connection teardown), and the `seal()`-in-try fix are all correct hardening on the same path.

The test coverage is thorough — it exercises both the ACL branch and the failed-rename branch independently, plus the negative case proving revoke/burn are intentionally *not* rolled back, and an end-to-end server test asserting neither side holds a record. The PROTOCOL.md §2.9 addition for #659 (multiplexed control connection) and the three CHANGELOG entries are clean and accurately scoped. One note for merge: the CHANGELOG line claiming "neither side is left holding a pairing the other cannot honour" is true for the responder-fails case this PR fixes but does not hold for the mirror-image joiner-fails case filed as #664 — worth correcting the wording when that lands, as the issue itself calls out.

**Priority: P2** — resolves a P1 correctness bug with a focused, well-tested fix that is ready to merge; the remaining follow-ups (#663, #664, #665) are tracked separately and explicitly out of scope here.

PR #667, opened 2026-07-28
fix(daemon): retry the atomic rename when antivirus briefly holds the target (win32)

openwong2kim:
Follow-up to #658 / #662. This addresses one of the two candidate root causes of the write failure itself — the half-pair symptom was fixed in #662; this PR is about making the save *succeed* on affected Windows hosts.

## Problem

Windows real-time antivirus opens a handle on a file the moment it is written. While that handle is held, a rename whose destination is that file fails with `EPERM`/`EACCES`/`EBUSY`. The window is tens of milliseconds, which makes the failure rare and unreproducible on demand.

The write ordering makes it destructive rather than merely annoying: by the time `tmp → target` runs, the previous primary has already been moved to `.bak`, so a failure at that exact point leaves **no primary file at all**. That is one of the two ways `lanlink-peers.json` can vanish mid-pairing (#658); the same layer backs the inbox and every other durable store in the daemon, so any of them could lose their primary the same way.

## Fix

A bounded retry with backoff around the final rename only:

- **win32 only**, and only on `EPERM`/`EACCES`/`EBUSY` — any other error, or any other platform, fails immediately exactly as before.
- 5 attempts, 10→160 ms backoff, ~310 ms worst-case total — bounded so the sync path cannot stall the daemon event loop meaningfully. The sync variant sleeps via `Atomics.wait` (no CPU burn); the async variant awaits a timer.
- On final failure the **original error is rethrown unchanged** (callers depend on error semantics), plus one warn line with the attempt count so an affected host is visible in `daemon.log`.
- Retry policy is injectable via `AtomicWriteOptions.renameRetry` — tests pass fake sleeps; production callers pass nothing.

## Tests

- transient `EPERM` × N then success → write succeeds, content correct
- persistent `EPERM` → original error after bounded attempts, no spin
- non-transient (`ENOENT`) → immediate throw, no retry
- non-win32 → single attempt (platform stubbed per repo convention)

## Verification

- `tsc` root + daemon + cli: clean
- `eslint src/daemon/util/atomicWrite`: touched files 0 problems (the 11 pre-existing errors in old test files are identical on main)
- Full suite: 8733 + 88 pass

If the reporters' daemon logs (requested on #658) point at the ACL branch instead, that fix is a separate security-policy decision and will come as its own PR — this one is correct regardless, since the rename hazard exists for every durable store.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved state persistence reliability on Windows when antivirus software temporarily locks newly written files.
  * Automatically retries eligible file commits briefly before reporting an error.
  * Preserves existing behavior on non-Windows platforms and for unrelated errors.

* **Documentation**
  * Added the Windows persistence fix to the unreleased changelog.


openwong2kim:

This PR adds a bounded retry around the final `tmp → target` rename in `atomicWriteJSON`/`atomicWriteJSONSync`, targeting the Windows antivirus scenario where a real-time scanner briefly holds a handle on the destination file and the rename fails with `EPERM`/`EACCES`/`EBUSY`. It is one of the two candidate root causes behind the `lanlink` write failure in #658 — the half-pair symptom was fixed in #662, and this addresses making the save actually succeed on affected hosts.

The fix is tightly scoped and correct: win32-only, only for the three transient errno codes, five attempts with 10→160ms backoff (~310ms worst-case total), original error rethrown unchanged on final failure so callers's `err.code` branching is preserved, and a warn line for visibility in `daemon.log`. The ordering argument is the key insight — by the time the commit rename runs, the previous primary has already moved to `.bak`, so a failure here leaves no primary file at all, which is how state loss propagates to every durable store in the daemon, not just the peer file.

Priority is **P2**: it directly unblocks the #658 reporter population on Windows, the primitive it fixes underpins the inbox and the daemon auth token (not just lanlink), and the test suite covers transient-then-success, persistent-failure, non-transient, and off-win32 paths. Worth landing regardless of whether the logs point at this branch or the ACL branch, since the rename hazard is independent of root cause.

openwong2kim:
@codex review

PR #668, opened 2026-07-28
feat(fanout): expose fan-out on the pipe and MCP surface

openwong2kim:
Fan-out — one prompt split into N isolated parallel tasks, each with its own git worktree on a fresh branch, its own workspace with an agent already launched, and its own mission channel — was reachable only from the GUI modal. `fanout.handler.ts` is a renderer-only `ipcMain` surface with no `RpcRouter` registration, so an MCP client could not open one. Given that "wmux exposes itself as MCP" is one of the two differentiators still standing, the marquee journey being absent from it is the gap this closes.

## What lands

- `task.fanout.start` on the pipe router.
- `fanout_start` MCP tool.
- `channel_mission_list` MCP tool — a wrapper over the **existing** `task.mission.list` RPC, which was already pipe-registered. **No new query RPC was added**: fan-out tasks are born owned by the caller and `task.mission.list` is owner-scoped, so a caller's fan-out tasks are exactly its mission list, including the materialised `branch` / `worktreePath` / `paneGroupId`. J0 had deliberately deferred this tool "to J1"; this is J1.
- `FanOutService` is now constructed once and injected into both front doors.

Out of scope, as agreed: diff adoption, PR creation, task close from the wire.

## Security requirements

The wire contract is deliberately narrower than the renderer IPC one. The IPC path trusts every field because a human typed it into the modal; the wire does not. The handler therefore builds the `FanOutRequest` field by field from a strict subset of caller input plus server-derived values — `params` is never spread, so a field added to the wire later cannot leak through by accident.

| # | Requirement | How it is met | Test |
|---|---|---|---|
| 1 | `agentCmd` never accepted from the wire | Not an input. `FANOUT_WIRE_AGENT_CMD` is set unconditionally. It is interpolated verbatim into `` `${agentCmd} "$(cat '<path>')"` `` and written to a PTY, so accepting one would be arbitrary command execution. A sent value is ignored rather than rejected (no field-existence oracle, no break for clients mirroring the IPC shape); the tool schema has no such field. The renderer IPC path is untouched. | `R1` — ignores an injection payload, and ignores a plausible-looking one |
| 2 | `verifiedWorkspaceId` derived from ctx, not params | Resolved from a verified `senderPtyId` via the renderer's `input.findOwnerWorkspace` — the same D5 anchor `a2a.channel.*` mutations use. An unresolvable caller is refused outright. `memberId` is likewise not read (it defaults to the resolved workspace). `ws-human` is rejected for symmetry with the channel handler. | `R2` — caller-supplied ws id ignored; unresolvable / absent ptyId / reserved workspace all refused |
| 3 | `repoPath` confined to the caller's own repo | The fan-out repo is `git rev-parse --show-toplevel` of the caller workspace's own `metadata.cwd`, realpath'd. A supplied `repoPath` may only re-state that same repository, compared as realpath'd **toplevels** — not a prefix match, which `..` and symlinks defeat, and which would reject a legitimate subdirectory. The caller's string is never forwarded; the normalised root is. No fallback to "the active workspace". | `R3` — foreign repo, non-repo, non-string and flag-shaped paths refused; own subdirectory accepted but normalised; missing cwd and non-repo cwd refused |
| 4 | origin allowlist, fail-closed | `ctx?.origin === 'local'` or reject, verbatim from the a2a execute precedent. `origin` is a required `RpcContext` field, so a future LAN transport cannot inherit fan-out by forgetting to classify itself. | `R4` — remote and context-less refused, **and** refused before any identity or repo resolution runs |
| 5 | N capped server-side | Same `FANOUT_MAX_TASKS` the GUI enforces, checked at the handler before the service is entered: the rejection is wire-shaped, and the raw array length is bounded before any per-element work. `taskPrompts` capped too. | `R5` — over-cap, 5000-entry array, over-cap `taskPrompts`, zero usable titles, non-array |
| 6 | prompt / title caps server-side | Each task's *effective* prompt (shared + per-task, the exact rule the service applies) capped at `FANOUT_PROMPT_MAX_BYTES` in **UTF-8 bytes**; each title at `CHANNEL_TOPIC_MAX`, the daemon's own limit, so N-1 tasks are not spawned before task N is rejected downstream. All-or-nothing, matching the service's existing contract. | `R6` — oversized shared prompt, oversized combination, byte-vs-character, over-long title |
| 7 | approval gate | **Not added, deliberately.** Reasoning below. | — |

### On requirement 7

Fan-out spawns `claude "$(cat <prompt>)"` into a fresh PTY — an ordinary interactive session with its own downstream permission prompts. That is not the a2a execute path, which spawns a headless worker in `bypassPermissions`; the existing gate exists precisely because that mode has no downstream prompts.

With requirements 1–6 applied, the only primitive fan-out adds over what a caller reaching this method already holds on the same router (`input.send`, which types *any* command into a pane; `mcp.claimWorkspace`; `pane.split`; `task.mission.start`) is `git worktree add` on the caller's own repo, ≤8 per call, into `~/.wmux/worktrees/`, on a fresh `wtask/*` branch that must not already exist. No commits, no ref rewrites, preserved rather than deleted on failure. Gating that while `input.send` stays ungated would be incoherent.

There is also a concrete cost: the call is asynchronous (below), so a prompt would fire *after* the tool returned, and the existing gate auto-denies after 30s — an unattended fleet would silently lose every fan-out. And reusing `requestExecuteApproval` verbatim would fold fan-out under the user's existing `a2aAutoApproveExecute` toggle, silently widening a setting agreed to for a different action.

Fan-out is loud rather than silent: N workspaces in the sidebar, N missions in the Missions section, one audited mission channel each, bounded by `WORKTASK_MAX_OPEN_PER_WORKSPACE`. **If you want a gate anyway**, the reuse path is a renderer round-trip into `requestExecuteApproval` with a `kind: 'fanout'` discriminator so the dialog copy stays truthful — deliberately left out because it touches the store, the dialog, the approval inbox and i18n, and should be a decision rather than a side effect.

## Accept-then-poll

`wmux-client.ts` sets a hard 10s RPC deadline, and a single task's renderer spawn alone is allowed 30s. A synchronous fan-out would time out by construction and the client's retry would re-fire it. So the call accepts, runs detached, and the caller polls by re-sending the same `idempotencyKey` — which falls straight out of the idempotency LRU the service already maintains (`accepted` → `running` → `completed` with the full result). `idempotencyKey` is therefore required on the wire.

Two small service additions serve this: `statusOf(key)`, and `start()` recording a failed result instead of releasing the key when `run()` throws — releasing it would let a poll **restart** a fan-out that had already spawned tasks. The GUI is unaffected (it mints a fresh key per submit).

## One service, two front doors

The idempotency LRU and in-flight set are **instance** fields, so two instances mean one key accepted twice and the fan-out running twice. The port assembly moves to `createFanOutService.ts`, `main/index.ts` constructs it once, and both registrars receive the same binding. A source-level test pins this, since no unit test can exercise the wiring.

## Not touched

`COMMANDER_RPC_METHODS` — the commander brain's allow lane stays as-is (it supervises; it does not need to spawn worktrees). `COMMANDER_TEARDOWN_DENY` — fan-out has no teardown effect.

## Verification

`tsc --noEmit` clean; `npm test` green (8777 passed / 23 skipped, plus 88 runtime). `docs/api/reference.md` regenerated for the new method. Design notes in `plans/fanout-mcp-surface-design.md`.

## Summary by CodeRabbit

- **New Features**
  - Added the `fanout_start` MCP tool for initiating parallel tasks across workspaces.
  - Added `channel_mission_list` to view active and completed mission tasks.
  - Fan-out requests now return immediately with an accepted status and support progress polling.
  - Added safeguards for caller identity, repository scope, request size, and duplicate submissions.

- **Documentation**
  - Updated the API reference and changelog with the new fan-out capability and usage flow.
  - Added design documentation describing the fan-out workflow and security controls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Agents can now fan out.** Splitting one job into N isolated parallel tasks — each with its own git worktree on a fresh branch, its own workspace with an agent already running, and its own mission channel — required a human to open the GUI modal. It is now a tool, `fanout_start`, so an agent supervising a fleet can open one itself. A companion tool, `channel_mission_list`, lists the tasks you own so you can follow them; it reads the mission RPC that already existed. Fan-out takes far longer than one call can wait, so `fanout_start` returns as soon as it has accepted the work and you poll it by calling again with the same idempotency key.
+
+  What a caller may ask for is deliberately narrower than what the modal accepts, because the modal's fields were typed by a human and a tool call's are not. The agent command is not an input at all — it is interpolated into a shell line, so accepting one would be arbitrary command execution. Which workspace owns the tasks is resolved from the caller's verified terminal, not from anything the caller says, so no one can create tasks owned by someone else. The repository is the one the calling workspace is already in; naming any other is refused. Task count and prompt and title sizes are capped where the request arrives, not only deeper in. And like the existing background-execution path, fan-out is refused outright to anything that did not arrive over the local machine's own socket.
+
 ### Changed
 
 - **The protocol docs now state that the daemon control connection is multiplexed.** Replies and pushed events share one stream, with no subscription step, and are told apart only by whether they carry the `id` of the request you sent. Clients that wrote a request and read exactly one line back therefore worked until an event arrived at the wrong moment, then reported a failure with an empty error message and dropped the real reply — a failure mode that can invent failures but never successes, which sent two teams debugging in the wrong direction. `docs/PROTOCOL.md` §2.9 now spells out the correlation rule. No behaviour changes; correctly-written clients were never affected. (#659)
--- a/docs/api/reference.md
+++ b/docs/api/reference.md
@@ -26,7 +26,7 @@ returns `EPERM`. Wire framing: newline-delimited JSON, one object per line.
 
 ## RPC methods
 
-Total: **144** methods (`ALL_RPC_METHODS` in
+Total: **145** methods (`ALL_RPC_METHODS` in
 `src/shared/rpc.ts`). Capability and risk class are read from
 `src/main/mcp/methodCapabilityMap.ts`:
 
@@ -270,6 +270,7 @@ Total: **144** methods (`ALL_RPC_METHODS` in
 | `task.mission.close` | `a2a.channel.send` | `a2a` |
 | `task.mission.list` | `a2a.channel.read` | `a2a` |
 | `task.mission.update` | `a2a.channel.send` | `a2a` |
+| `task.fanout.start` | `a2a.channel.send` | `a2a` |
 
 ---
 
--- a/plans/fanout-mcp-surface-design.md
+++ b/plans/fanout-mcp-surface-design.md
@@ -1,0 +1,337 @@
+# Fan-out on the MCP / pipe surface — design (2026-07-28)
+
+## 0. Why
+
+`plans/as-is-to-be-2026-07-28.md` §0.3 concludes that two moats survived: **reboot
+survival** and **wmux exposing itself as MCP**. Fan-out — the marquee journey (J1)
+— is missing from the second one. `src/main/ipc/handlers/fanout.handler.ts` is a
+renderer-only `ipcMain.handle` surface; it is not registered on the pipe
+`RpcRouter`, so an MCP client cannot run a fan-out. A human has to open the GUI
+modal. This PR closes that gap.
+
+The interesting part of this change is not the plumbing — `pane.rpc.ts` already
+imports `sendToRenderer`, so there was never a technical barrier. The barrier was
+a **trust decision**: the IPC path treats every field of the request as
+renderer-authored, i.e. human-authored. The wire is not. So the wire handler is a
+*different* handler with a *narrower* input contract that happens to reuse the
+same service.
+
+## 1. Scope
+
+**In**
+
+- One new pipe RPC: `task.fanout.start` (+ `RpcMethod` union, capability map,
+  first-party allowlist, generated API reference).
+- One new MCP tool: `fanout_start`.
+- One new MCP tool for the status read: `channel_mission_list` — a thin wrapper
+  over the **existing** `task.mission.list` RPC (already registered on the pipe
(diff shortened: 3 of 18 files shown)


openwong2kim:

This PR exposes fan-out — splitting one job into N isolated parallel tasks with per-task git worktrees, workspaces, and mission channels — on the pipe RPC router and MCP surface, where previously it was reachable only from the GUI modal. It adds `task.fanout.start` and `fanout_start` (plus `channel_mission_list` over the existing `task.mission.list`), makes `FanOutService` a single shared instance injected into both front doors, and adopts an accept-then-poll contract forced by the client's 10s RPC deadline.

The trust-boundary work is the substantive part: the wire handler is a deliberately narrower surface than the IPC path, never reading `agentCmd` (which is interpolated into a shell line), deriving `verifiedWorkspaceId` from the caller's verified pty, confining `repoPath` to the caller's own repo root via realpath comparison, and enforcing origin-local fail-closed plus server-side caps on task count, prompt bytes, and title length. The documented decision to skip an approval gate is reasoned rather than hand-waved — fan-out spawns an interactive `claude` session with its own downstream prompts (unlike the headless bypassPermissions execute path), and its marginal capability over what the caller already holds (`input.send`, `task.mission.start`) is `git worktree add` on the caller's own repo.

Priority is **P2**: this closes a real gap in a stated differentiator (wmux exposing itself as MCP), the security design is thorough and test-backed, and the scope is well-contained with explicit out-of-scope items (diff adoption, PR creation, wire-side task close). The idempotency-shared-instance refactor and the fail-closed posture on unresolvable callers are both correct.

openwong2kim:
Closing as superseded, not rejected.

This branch was produced by a duplicate run: the same specification was implemented twice in parallel by mistake, and this one was pushed before the duplication was noticed. The two implementations diverged on a real design question — whether the wire surface needs its own approval gate — and each caught something the other missed.

Both are being folded into a single branch, which keeps this branch's accept-then-poll contract (the MCP client's RPC deadline is shorter than a single task's spawn allowance, so a synchronous call reports failure for a fan-out that actually succeeded) and the other's approval gate, with fan-out separated from the existing A2A auto-approve toggle rather than folded under it.

The design note in `plans/` here was committed past `.gitignore` with `-f`; whether that directory should be tracked is a separate decision and is not being settled by this branch.

PR #669, opened 2026-07-28
docs: state what diff adoption actually does, and stop leading with the browser

openwong2kim:
## What

Two README claims did not match the code, in opposite directions.

**Hunk adoption was undersold.** The README said `adopt hunks all-or-nothing`, which reads as "the whole diff lands or nothing does". The code does something better: you tick individual hunks, per file and across files, and *that selection* is combined into one patch and applied atomically (`diff.handler.ts`). Individual per-hunk `--check` is deliberately not used as a gate, because dependent hunks make it produce false rejections. The wording now describes the selection as the atomic unit, and names the three paths that refuse a selection outright (target moved, worktree dirty, hunk no longer matches).

**The browser was oversold.** `Agents drive a *real* browser` claimed uniqueness that no longer holds — an integrated browser is now common in this category. The row moves to the end of the feature table and states integration as a fact rather than a differentiator. Nothing about the feature itself changed, and the detailed docs and MCP tool list are untouched.

A draft of this PR also claimed the browser tools run "under the same approval gates as everything else". They do not, so that sentence was removed before commit rather than shipped.

## Adoption semantics are now pinned by tests

The suite had two gaps that were exactly the two things the README now asserts:

- **Partial selection within a file** — every existing multi-hunk test selected *all* hunks, so per-hunk granularity was never exercised.
- **Atomicity across files** — existing tests asserted that a bad selection is refused, but never that the *other* files stayed untouched.

Both are covered now, plus a third case selecting a different hunk in each of two files. Fixtures pin `diff.context=3` / `interHunkContext=0`: without the pin, a developer whose global gitconfig widens diff context merges the two edits into one hunk and the tests fail for an unrelated reason (reproduced, then pinned).

## Not changed

`THIRD_PARTY_NOTICES` was checked and left alone. The installed `@modelcontextprotocol/sdk` declares MIT in its `package.json`, ships an MIT `LICENSE`, carries no `NOTICE` file, and the registry metadata agrees — so there is no Apache NOTICE obligation to inherit and the current entry is accurate.

## Review

Reviewed by a three-model panel. Consensus finding: the adoption guarantee still read as unconditional, because `applyHunks` re-reads the source diff at apply time. The wording was narrowed to what the code guarantees. The underlying re-read behaviour is a real defect and is being fixed separately — it is not introduced by this PR.

One finding was raised and **not** acted on: the atomicity test stops at the combined `--check`, so a mutant that replaces the single `git apply` with a per-file loop still passes. Verified by mutation (28/28 passed under the mutant). Pinning "apply is called once" needs mocking the git module, and this file runs against real git by design. Recorded here rather than silently left.

## Verification

- `tsc --noEmit` clean
- `test:parallel`: 8,729 passed, 0 failed
- `diff.handler.test.ts`: 28/28
- Mutation-checked: reverting the implementation turns the new granularity tests red

## Summary by CodeRabbit

* **Documentation**
  * Updated README and changelog wording to describe the in-workspace browser experience.
  * Clarified per-file/per-hunk selection and that adoption is performed as a single all-or-nothing `git apply` gated by a target snapshot.
  * Documented stricter refusal behavior when the diff is outdated, files are dirty, or selected hunks no longer apply.
* **Tests**
  * Added coverage for per-hunk selection granularity and request-wide atomicity, including multi-file scenarios and selection rejection when any selected hunk becomes not-applicable.


openwong2kim:

This is a documentation-only PR that corrects two README inaccuracies in opposite directions, each verified against the code. The hunk-adoption row was undersold: "adopt hunks all-or-nothing" read as if the entire diff landed or nothing did, when the code actually lets you tick individual hunks per file and across files, and only that selection is combined into one atomic `git apply` — the all-or-nothing guarantee is on the apply of the selection, not the scope of the selection. The browser row was oversold: "Agents drive a *real* browser" led with a differentiator that no longer holds now that vendors ship their own integrated browser, so the row is restated as integration rather than uniqueness and moved to the end of the feature table. What makes this PR worth more than a passing glance is that it pins both corrected README claims with new tests in `diff.handler.test.ts`: partial selection within a file (previously every multi-hunk test selected all hunks), atomicity across files (existing tests only asserted a bad selection was refused, never that other files stayed untouched), and a third case selecting a different hunk in each of two files. Fixtures pin `diff.context=3`/`interHunkContext=0` to keep the hunk split stable against a developer's global gitconfig. One honestly-recorded gap: the atomicity test stops at the combined `--check`, so a mutant replacing the single `git apply` with a per-file loop still passes — noted rather than silently left.

Priority: **P3** — documentation wording plus the tests that pin it; no runtime behaviour changes.

openwong2kim:
@codex review

openwong2kim:
## Live dogfood — adoption claims

Docs PR, so the check is whether each claim survives contact with the code.

**The README is now accurate.** `5240a185` fixed both overclaims Codex raised:
the hunk-selection row says "in any **text** file — renames, binaries,
mode-only and over-cap files stay display-only", which matches
`hunkSelectable === false` and the `truncated` set in `applyHunks`, and the
detail paragraph now names the three refusal paths and the re-read caveat.

**The changelog was not.** The entry describing that fix still read "you tick
individual hunks, **in any file** and across files" — the exact sentence being
corrected, in the file announcing the correction. Fixed on this branch; the two
files in one PR no longer disagree about what adoption supports.

**One claim is true only once #672 lands.** The README's headline —
*"one atomic `git apply` that lands whole or is refused whole"* — does not hold
on `main` today. `applyHunks` builds `selectedFiles` only from paths present in
the fresh read (`diff.handler.ts:333-353`), so a selected file that left the
diff between the read and the click is **skipped, not refused**, and the
remainder applies with `ok: true`. The user's displayed selection neither lands
whole nor is refused whole — quietly.

#672 is what makes that sentence true: it refuses the whole adoption if any
selected file changed, left, or lost its hunks. Merging this one first would
ship a promise ahead of the code that keeps it. Either order the two, or land
them together.

The new tests are the good part of this PR — partial selection within a file,
atomicity across files, and a different hunk chosen in each of two files were
genuinely uncovered before, and pinning `diff.context` / `interHunkContext` in
the fixture is the right call (a developer's global gitconfig was silently
deciding the hunk split). The honestly-recorded gap — the atomicity test stops
at the combined `--check`, so a per-file-loop mutant still passes — is worth
keeping visible rather than papering over.

64 passed / 1 skipped across the diff suites on this branch. CI green.

PR #670, opened 2026-07-28
feat(diff): generate diffs with the histogram algorithm, not Myers

openwong2kim:
## What

Every path that produces hunks now asks git for the histogram algorithm instead of the default Myers.

- `diff.handler.ts` — the `diff:read` patch **and** its `--numstat` companion. Both, because the file tree's +/- counts sit directly above the hunks they are supposed to describe; letting them come from different algorithms means the tree can disagree with what is rendered under it. `applyHunks` re-invokes `readDiff`, so the patch a user selected and the patch that gets applied stay on the same algorithm by construction.
- `sessionDiff.ts` — the staged/unstaged patches the phone renders before approval.

Agent-authored changes contain a lot of moved code, which is where Myers produces the most fragmented hunks. Histogram describes the same change in fewer, larger ones — directly fewer checkboxes to reason about in the adoption UI.

## Hunk anchors are not affected

Checked before changing anything, since re-hunking could in principle invalidate stored selections:

- **Adopted-hunk selections are never persisted.** They live in `DiffPanel`'s `useState` and are rebuilt from `diff:read` each time the panel opens. There is no stored hunk index to invalidate.
- **Comment anchors (`hunkHeader`) are persisted**, but `DiffPanel` already demotes any comment whose header no longer matches into a "moved" group. Line drift triggers that path constantly today; an algorithm change routes through the same handling. No loss, no crash.
- The algorithm is fixed at build time, so a panel session can never mix algorithms between its read and its re-read.

No migration needed.

## Tests assert argv, not output

Modern git enables `indentHeuristic` by default, which makes Myers and histogram produce identical output for most inputs — including both candidate fixtures, verified against real git 2.50.1. An output-comparison test would therefore pass with the flag removed, which is worse than no test. The tests assert the flag reaches the command line for both the patch and `--numstat` calls, in both task and workspace modes.

Three existing tests were updated because this change broke them: two identified staged vs unstaged by argv *position* (`args[1] === '--cached'`), which stops being true once a flag is inserted ahead of it.

## Review

Reviewed by a three-model panel. Two consensus findings, both fixed:

1. **A third positional check was missed.** The unborn-repo test injected its failure the same way, and after the flag insertion the injection silently stopped landing — while the test kept passing, because every stub returned an empty patch and the assertion happened to hold. It was asserting nothing. Fixed with a content check plus an assertion that the repair itself is observable. Mutation-verified: restoring the positional stub turns it red (1 failed/37 passed), reverting gives 38 passed.

2. **The original security rationale was wrong and has been withdrawn.** This change was first justified as preventing a cloned repository from choosing its own diff algorithm. `diff.algorithm` is a `.git/config` key and `git clone` does not transfer config — verified empirically. The comments, changelog and commit message now describe the actual benefit, which is determinism. It is deliberately *not* added to `GIT_HARDENING_CONFIG`: that list defends against config keys which turn a checkout into code execution, and adding a formatting preference would overstate what it covers.

Three single-model findings were rejected with evidence, including adding the flag to the `--no-index` untracked synthesis — the two argvs produce byte-identical output there, because a `/dev/null` baseline never consults the algorithm.

One **pre-existing** defect was found and deliberately left out of scope: `diff.handler.ts` passes neither `--no-ext-diff` nor `--no-textconv`, so a user with `diff.external` configured gets a patch that parses to zero files while `--numstat` still reports counts — a file tree showing changes with no hunks beneath it. Present on `main`, unrelated to this change, and being fixed alongside the related adoption defect instead.

## Verification

- `tsc --noEmit` on both app and daemon configs: clean
- `test:parallel`: 8,729 passed, 0 failed · `test:runtime`: 88 passed
- eslint on changed files: 0 errors

## Summary by CodeRabbit

- **Changed**
  - Diff views now use Git’s histogram algorithm for more consistent hunk placement across task and workspace diffs.
  - Diff and numstat results use the same explicitly selected algorithm, preventing configuration differences from causing mismatches.
  - Diff hunk numbering may change when the histogram algorithm produces different groupings.
  - Patch application behavior remains unchanged.


openwong2kim:

This PR pins git diff generation to the histogram algorithm (`--histogram`) instead of the default Myers, applied consistently across the task diff surface (`diff.handler.ts` — both the patch and its `--numstat` companion) and the phone-side working-tree patch (`sessionDiff.ts`). The motivation is practical: agent-authored changes are movement-heavy, and Myers tends to anchor on incidental repeated lines (`}`, blanks), fragmenting one logical relocation into several interleaved hunks — histogram weights rare lines as anchors and emits the same change in fewer, more readable hunks, which is exactly the unit a reviewer ticks a checkbox next to. Stating the flag on the argv rather than reading `diff.algorithm` gives determinism across config variations, and the author correctly declines to file it under `GIT_HARDENING_CONFIG` (that list defends against config keys that turn a checkout into code execution; `diff.algorithm` lives in `.git/config`, which `git clone` does not transfer, so the original security rationale was withdrawn). Tests assert the flag reaches the command line in both task and workspace modes — the right choice, since modern git enables `indentHeuristic` by default and Myers/histogram produce identical output on most inputs, making an output comparison vacuous. Three existing tests were updated because the flag insertion broke positional argv checks (`args[1] === '--cached'`), and a previously-vacuous unborn-repo test was strengthened to assert the injected failure actually lands. The pre-existing `--no-ext-diff`/`--no-textconv` gap that lets a `diff.external` config produce a file tree with no hunks is correctly scoped out as unrelated.

Priority: **P2** — a focused, well-tested DX improvement to diff readability with no behaviour risk beyond hunk renumbering (which adoption is unaffected by, since selections are session-scoped).

openwong2kim:
@codex review

openwong2kim:
## Live dogfood — histogram diff

The suites that back this PR drive **real git in real temp repos** (not stubs),
so running them is itself the live test: 64 passed / 1 skipped across
`diff.handler`, `diff.handler.algorithm` and `sessionDiff`. The `--histogram`
flag reaches the command line on all four sites (task patch, task numstat,
workspace mode, phone-side working-tree patch).

I then went looking for the readability win on real inputs, in a scratch repo
with git 2.55:

| case | Myers | histogram |
|---|---|---|
| 4-line block relocated past 8 similar blocks | 2 hunks / 7 lines | 2 hunks / 7 lines |
| 5 near-identical handler blocks moved to end of a 60-block file | 2 hunks / 70 lines | 2 hunks / 70 lines |
| new block inserted mid-file among 60 near-identical ones | identical output | identical output |

Byte-identical every time. Which is what the description already predicts —
`indentHeuristic` has been on by default since git 2.14 and it absorbs most of
the brace-anchoring that made Myers fragment moves — so this is a confirmation,
not a contradiction.

Worth being explicit about in the changelog, though, because the entry reads as
a readability improvement and a reviewer who goes looking (as I did) will not
find one on ordinary input. **The value here is determinism, not fewer hunks**:
the flag pins the algorithm on the argv so a developer's `diff.algorithm` config
cannot change what the panel renders, and adoption resolves selections by
positional hunk index — so "the same diff for everyone" is load-bearing for
correctness, not just for reading comfort. That is a better claim than "fewer,
more readable hunks", and it is one I could verify.

Declining to file it under `GIT_HARDENING_CONFIG` is right: `diff.algorithm`
lives in `.git/config`, which `git clone` does not transfer.

### Reviewer item, fixed

CodeRabbit flagged Korean `describe`/`it` labels in the new test file — the repo
rule is that artifacts are English. Renamed the three. Checked the rest of the
diff: no other new Korean lines; the existing Korean comments in
`diff.handler.ts` and `sessionDiff.test.ts` predate this branch and are left
alone.

CI green.

PR #671, opened 2026-07-28
fix(terminal): measure emoji by grapheme cluster, not codepoint

openwong2kim:
## What

Swaps `@xterm/addon-unicode11` for `@xterm/addon-unicode-graphemes` (0.4.0, MIT) so width is computed per user-perceived character instead of per codepoint.

Measured against `@xterm/headless` 6:

| | before | after |
|---|---|---|
| ZWJ family 👨‍👩‍👧 | 6 | **2** |
| skin tone 👍🏽 | 4 | **2** |
| VS16 ❤️ | 1 | **2** |
| keycap 1️⃣ | 1 | **2** |

**No CJK regression.** Hangul (composed and decomposed), Han, kana, fullwidth, halfwidth katakana, CJK punctuation, ambiguous-width, box drawing, braille, powerline and nerd-font ranges all measure identically to Unicode 11 across 15 ranges.

The renderer and the daemon's headless snapshot terminals are switched together — a split there would make the screen and the snapshot disagree about where text wraps. Three test harnesses and a probe script that mirrored the same setup were updated for the same reason.

## The width model is now one thing

The version string used to appear as a literal at three call sites with nothing binding them. Reverting any single one left the whole suite green, so "switch both places" was enforced by a comment rather than by the code. There is now a single `applyUnicodeWidthModel()` entry point in `src/shared/terminalUnicode.ts`, and the version string exists exactly once in the tree.

This matters more than it looks: the ANSI snapshot path **cannot be covered by any black-box test**. The serialize addon re-emits the characters, so the payload is byte-identical under either width model. Structure is the only available defence there.

Two guards were added: one asserts the addon actually registers the version being requested (a rename or removal upstream would otherwise fall back silently), and the CJK no-regression suite compares against `Unicode11Addon` *in the same process* rather than against hardcoded numbers, so a future addon bump that disturbs CJK widths fails here.

## Corrections found in review

A three-model panel found three real defects in the first version of this work:

1. An `it.each` title bound `%d` to the wrong tuple element, so every case printed "Unicode 11 said NaN" — the comparison number never appeared in the report.
2. **The flag test passed vacuously.** U+1F1F0 U+1F1F7 already measures 2 under Unicode 11, and only the post-swap width was asserted, so it passed under either model and caught nothing. Worse, the changelog listed flags as a *fixed* case — a user-facing claim that was simply false. Both assertions are now made and flags are removed from the changelog.
3. **"Both places" was not true.** A third grid-owning site — the `wmux web` terminal — loads no unicode addon at all and remains on Unicode 6. Closing that gap needs the addon inlined and the web CSP script hashes regenerated, which is out of scope here; the claim is corrected rather than the gap quietly left implied.

Also recorded: widening VS16 to 2 is a **trade-off, not a pure win**. Applications on the far side of the PTY that count with classic `wcwidth` (ncurses, Go runewidth's default) still count 1, so wmux is one column ahead of them. And `UnicodeGraphemesAddon.dispose()` resets the active version to `'6'`, unlike the Unicode 11 addon — the helper does not return the instance, so that mistake cannot be made through it.

## Known limitation, accepted

Grapheme clustering joins a ZWJ chain of any length into a **single cell**, and the core has no cap on a cell's combined-character string. Scrollback trimming is row-granular, so it does not account for what a row actually costs — the `O(1)` term in `rows × cols × O(1)` stops holding.

A standalone reproduction is included at `scripts/repro-grapheme-cell-growth.mjs` (no wmux dependencies). With a 500k-emoji ZWJ chain:

| | unicode11 | graphemes |
|---|---|---|
| max chars in one cell | 3 | **1,499,999** |
| chars per cell | 0.6 | **382.8** |

This is an upstream property of the addon, not something this change introduces — it is the substance behind the package's "experimental" warning. It is documented rather than mitigated: bounding it requires reimplementing the core's private property-value bit encoding, which is a worse long-term liability than the limitation itself. It does not arise from natural text, and closing the pane recovers the memory.

## Throughput

Measured at 200 columns, 20k rows, after warmup: ascii 44.2 → 55.6 MB/s (noise), CJK 87.1 → 54.7 MB/s (−37%), emoji 44.5 → 35.4 MB/s (−21%). No regression gate is added: the daemon's snapshot budget works out to ~2 MB/s, leaving roughly 27× headroom, and picking a threshold now would be arbitrary. The numbers are recorded so a later gate has a baseline.

## Verification

- `tsc --noEmit` across all six tsconfigs: 0 errors
- `test:parallel`: 8,758 passed, 0 failed · `test:runtime`: 88 passed
- eslint on changed files: 0 errors · renderer `vite build`: succeeds
- `test:harness` reports 11 pre-existing failures in this sandbox (`posix_spawnp` — PTY spawn is blocked here), confirmed identical on a clean tree

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved terminal alignment for multi-codepoint emojis, including family emojis, skin-tone variants, flags, keycaps, and variation-selector symbols.
  - Prevented complex emojis from shifting following text or splitting across rows near the terminal edge.
  - Kept CJK character width behavior unchanged.
  - Synchronized terminal display and screen snapshots for consistent wrapping.

- **Documentation**
  - Updated Unicode support documentation to reflect Unicode 15 grapheme-cluster width handling.
  - Documented known compatibility and memory-use limitations for exceptionally long emoji sequences.


openwong2kim:

This PR replaces `@xterm/addon-unicode11` with `@xterm/addon-unicode-graphemes` (Unicode 15) so terminal width is computed per user-perceived grapheme cluster rather than per codepoint — fixing the cases where a ZWJ family (👨‍👩‍👧), a skin-tone modifier, a VS16 sequence, or a keycap occupied the wrong number of columns and dragged the rest of the line sideways. The renderer (`useTerminal.ts`) and the daemon headless snapshot (`HeadlessSnapshot.ts`) are switched together through a single new `applyUnicodeWidthModel()` helper in `src/shared/terminalUnicode.ts`, which is the right structural choice: the snapshot is restored onto the renderer grid, so a width-model split between the two would paint cell-shifted, and centralizing the addon choice plus version string in one module makes that divergence impossible to reintroduce silently. Coverage is thorough — grapheme-cluster widths, a differential CJK no-regression assertion measured against the old Unicode 11 tables in-process, wrap-boundary parity between snapshot and renderer, and a guard that the addon actually registers the claimed version. The accepted limitation (an unbounded ZWJ chain collapses into a single cell, so a row-counted scrollback stops bounding buffer memory) is documented honestly with a standalone reproduction, and the throughput regression on CJK is recorded as baseline without an arbitrary gate. The only gap left open is the `wmux web` browser terminal, which still renders at the xterm Unicode 6 default — correctly called out rather than implied away.

Priority: **P2** — a correctness fix to character width that materially improves rendering for multi-codepoint emoji and keeps the snapshot/renderer in lockstep, but CJK (the larger user base) is unchanged and the trade-offs are well-contained.

openwong2kim:
@codex review

openwong2kim:
Closing: the grapheme swap is declined, and the part worth keeping ships separately as #TBD.

**Why declined.** Counting what actually appears in this repository: VS16 symbols (⚠️ ✔️ ❤️) in 26 files, ZWJ families in 2 — and both of those are test fixtures we wrote *for this change*. Skin tones and keycaps: zero. So the swap's only real-world effect on our content is VS16, which is the most contested class of all: the programs running inside wmux panes — ncurses, Go's runewidth default, older Python wcwidth — still measure it as one column. Making the terminal more correct per spec while disagreeing with the applications drawing inside it is a bad trade, and misalignment is worse than being consistently wrong in both places.

Against that, four costs: upstream mis-reads its own compressed trie when Node hands it a pooled Buffer (filed as https://github.com/xtermjs/xterm.js/issues/6079 — three symptoms, including silently wrong widths with nothing thrown); one grapheme cluster occupying one cell removes the memory bound a row-counted scrollback implies (13 MB of input retained 214 MB); CJK throughput measured ~37% lower; and the VS16 disagreement above. This also began as a suggestion out of a survey rather than a reported problem, so there was no user-visible defect paying for any of it.

**What survives.** The review of this PR found something independent of the addon: the renderer and the daemon's two snapshot terminals each named the width model themselves, with nothing tying them together — and reverting any single site left the whole suite green, because a snapshot only paints cell-shifted once it is restored onto a grid wrapping at a different column. That consolidation, plus a version guard that asserts the id is *registered* rather than merely assigned, ships on its own branch against Unicode 11 with no width changes at all. The module records why it is Unicode 11 and not grapheme clustering, so the next person to propose this swap finds the evaluation instead of repeating it.

Nothing here was wasted: the upstream defect is filed with a one-line reproduction, and the CI failure that started this — four suites dying on macOS and Windows while Linux passed the same commit — turned out to be that upstream bug rather than anything in this branch.

PR #672, opened 2026-07-28
fix(diff): adopt only what the user reviewed, and render what git itself produces

openwong2kim:
Two defects in the same `diff:read` / `diff:applyHunks` path, both of which split **what the user sees** from **what actually gets adopted**. They are shipped together because they are the same failure mode in the same read, and reviewing them apart means reading the same code twice.

Neither is introduced by recent work: both are present on `main`. The first was found by a three-model panel reviewing an unrelated documentation change, which is also why the adoption wording in that PR had to be narrowed.

## 1. The selection was mapped onto a diff the user never saw

`applyHunks` re-read the task diff at apply time and mapped the stored selection onto that fresh read by `(path, positional hunk index)`. Nothing tied the fresh read to the diff on screen. Anything the task agent wrote to its worktree between the read and the adopt was silently adopted **in place of** what was reviewed.

The same mapping dropped selections instead of refusing them: a selected path that had left the diff was skipped, an out-of-range hunk index was dropped inside `reassembleFile`, and adoption proceeded with the remainder and reported success.

That window is not exotic — it is the **normal path** for the fan-out journey, where agents keep writing while a human reviews their output.

The existing gates never covered it. Drift, dirty and the probes all describe the *target* repo; none describes the *source* diff. `git apply` being all-or-nothing does not help either, because the patch is already the wrong patch by the time git sees it.

**Fix.** `diff:read` returns an adoption fingerprint per file — sha256 over a canonical serialization of the entry (path, kind, adoptability, header block, and every hunk's header and body; *all* hunks, because a selection is a positional index into that list). The adopt request echoes it back per selection, and the handler re-derives it from the fresh read before assembling anything. A selected file that changed, left the diff, or no longer holds the picked hunks **rejects the whole request** — partial adoption is not reachable.

Rejections ride the existing failure shape rather than a new channel: one added value on the code union (`'stale'` — the source moved, as distinct from `'drift'` where the target moved) plus `failedProbes` naming the refused hunks, which the panel already renders. A request without a fingerprint is refused; that needs no compatibility window because the channel is in-process IPC and selections are not persisted anywhere.

This also covers a selection left over from a Reload, which previously pointed at hunk positions in a re-parsed diff. The three existing refusals are unchanged.

## 2. The rendered patch could come from something other than git

`git diff <base>` was called with no engine pin, so two user-level settings could replace what the panel renders:

- **`diff.external`** (difftastic being the common case) substitutes the tool's output wholesale. `parseUnifiedDiff` finds no `diff --git` header and returns zero files, while the `--numstat` call in the same read never consults `diff.external` and reports the real counts. **The file tree showed additions and deletions for a file with no hunks under it, and nothing in the response said why.**
- **textconv** is on by default for `git diff` and rewrites file content, so the rendered hunks belong to a patch that cannot be applied to anything — including by adopt, which reassembles exactly those bytes.

**Fix.** The patch call now passes `--no-ext-diff --no-textconv`.

`--numstat` deliberately gets neither flag. It was measured to be unaffected by both — with a line-doubling textconv driver bound to the file, the patch went from 3 lines to 6 while `--numstat` kept reporting the raw 1/1 — so adding them there would be inert.

## Tests

- A read with `diff.external` configured still parses hunks for the file `--numstat` lists.
- A read under a content-rewriting textconv driver carries the real content and adopts to the real bytes.

Neither configured command is ever spawned once the flags are in place, so both tests are portable.

**Mutation-verified**: dropping either flag turns its own test red; reverting the fingerprint check turns the staleness tests red.

## Notes

- Refusal is expressed in the result, not as a thrown IPC rejection: `handleAdopt` has no catch, so rejecting at the IPC layer would leave the panel stuck in `applying`. Refusals belong to the layer that owns them.
- `DiffPanel.load()` still does not clear the selection on reload. With this change that is no longer a correctness issue — a stale selection is now explicitly refused rather than wrongly applied — so whether to auto-clear it is left as a UX decision.
- The `'stale'` reason string renders in English; the other messages from this handler are equally unlocalized today, and wiring a code→i18n mapping is separate work.

## Verification

- `tsc --noEmit` clean · `test:parallel` and `test:runtime` green · eslint on changed files clean

## Merge order

Touches `diff.handler.ts` and `diff.handler.test.ts`, which also change in #669 and #670. Those two do not conflict with each other; this branch should rebase after both land.


openwong2kim:

This PR closes a silent-corruption window in the hunk adoption flow (`diff.handler.ts` / `DiffPanel.tsx`) and pins the rendered diff to git's own engine. Today, `applyHunks` re-reads the task worktree diff at apply time and maps the user's selection onto it by `(path, hunk index)` with nothing tying that fresh read back to the diff on screen — so anything the agent wrote between review and adopt lands in place of what was ticked, and out-of-range or vanished selections were dropped rather than refused, with adoption reporting success on the remainder. That path is the normal one in the fan-out journey, where agents keep writing while a human reviews, so it is worth fixing rather than documenting around. The fix introduces a per-file adoption fingerprint (`sha256` over a length-delimited canonical serialization of path, kind, adoptability, header block, and every hunk header+body), echoed back on each selection and re-derived from the fresh read before any patch is assembled; any mismatch now rejects the **whole** request (`code: 'stale'`, with `failedProbes` naming the refused paths/hunks), and a fingerprint-less request is refused outright — safe because the channel is in-process IPC and selections are never persisted. The second commit passes `--no-ext-diff --no-textconv` to the patch call so a user-level `diff.external` (difftastic) or default-on `textconv` driver can no longer hand the panel a zero-file parse or an unappliable patch while `--numstat` keeps reporting real counts; `--numstat` is intentionally left unflagged after measurement showed it is unaffected by either.

The contract change is additive and contained: one new field on `DiffReadFile` / the selection type, one new value on the `DiffApplyResult.code` union (`'stale'`, distinct from `'drift'` which is target-side), and the renderer already renders `failedProbes`. The rejection rides the existing failure shape rather than a thrown IPC rejection, which matters because `handleAdopt` has no catch — refusing at the IPC layer would leave the panel stuck in `applying`. Tests are thorough and mutation-verified: five regressions cover the edited-between-read-and-apply case, a selected path leaving the diff (asserting the still-valid file stays untouched — closing a pre-existing hole where all-or-nothing was only pinned at the combined `--check`), an unresolvable hunk index, a fingerprint-less request, and refused-hunk reporting; a pass-through spy on the git module now pins multi-file adoption to exactly one write invocation. Both flag additions on the engine pin were separately mutation-verified to turn their tests red.

One thing worth noting on merge ordering: this branch touches `diff.handler.ts` and `diff.handler.test.ts`, which also change in #669 and #670, and the description states it should rebase after both land — the conflicts are expected and localized to the same read/apply path. The `stale` reason string is rendered in English with no i18n wiring, matching the rest of this handler today; localizing the code→message map is called out as separate work. Beyond that the change is well-scoped to the defect and I see nothing that would block merge once the rebase order is observed.

Priority **P2**: correctness of a shipped review-gated feature where the failure mode is silent adoption of unreviewed bytes, reached on the normal fan-out path — important to land, but not a live outage or data-loss fire, so it does not need to jump the P1 lane. Once it lands, the CHANGELOG note in #669 about adoption being unconditional should already be aligned with this fix (the panel-side narrowing there anticipated it).

openwong2kim:
@codex review

openwong2kim:
## Live dogfood — adoption integrity

Dogfooded against a **running dev build of this branch** (`WMUX_DATA_SUFFIX=-dev`,
CDP on 18841), importing the module the live renderer is actually serving rather
than a test copy:

```
PASS  file unchanged after reload       -> []
PASS  agent rewrote it under selection  -> ["a.ts"]
PASS  file left the diff                -> ["a.ts"]
PASS  only the moved one is dropped     -> ["b.ts"]

PASS  adopt sends the digest recorded at tick time
PASS  no fresh-read digest remains at adopt
PASS  reload prunes selections whose file moved
```

### One blocking finding, fixed on this branch

The gate was defeated on the one path it most needed to hold — the Reload path
Codex flagged, and I could confirm the whole chain in the source rather than
just the symptom:

`handleAdopt` built each selection's digest from `filesByPath` — *the read being
displayed right now* — while `load()` deliberately keeps the selection across a
reload (there is a comment saying so: 채택 selection이 refetch로 날아가면 안 됨).
So after a reload the digest **agreed with itself by construction**, and hunk
indices picked from the previous rendering adopted whatever now sat at those
positions. That is exactly the silent substitution this PR exists to refuse, and
the changelog claimed it was already covered.

Fixed by recording the digest when a hunk is ticked and sending *that*. On
reload, a path whose entry is gone or no longer matches what it was ticked
against loses its ticks; the rest keep theirs, so the manual-reload contract
still does what it is for. The decision is a pure function
(`staleSelectionPaths`) so it is pinned by tests instead of by rendering the
panel, and the changelog line claiming this case was already handled now
describes what actually happens.

### One thing worth deciding before merge

The README claim on #669 — *"lands whole or is refused whole"* — depends on this
PR. On `main` today, `applyHunks` builds `selectedFiles` only from paths present
in the fresh read, so a selected file that left the diff is **skipped, not
refused**, and the remainder applies successfully with `ok: true`. This branch
is what makes that sentence true. If #669 lands first it ships a promise the
code has not made yet; worth merging in the other order, or noting it.

Typecheck, lint and the diff suites are green on the branch (49 passed / 1
skipped across `components/Diff` + `diff.handler`).

PR #673, opened 2026-07-28
feat(fanout): expose fan-out on the MCP surface, accept-then-poll, on its own consent

openwong2kim:
Fan-out — one prompt split into N isolated tasks, each with its own git worktree, branch, workspace, agent pane and private mission channel — was reachable only from the in-app dialog. `fanout:start` is renderer-only by the same trust basis as `channelLocal` (the Electron process boundary), so the MCP surface wmux exposes to other agents could not reach its flagship journey.

This adds `task.fanout.start` to the pipe router, plus the `fanout_start` and `channel_mission_list` MCP tools.

> **Provenance.** This surface was implemented twice against the same requirements, in parallel, by mistake. The two implementations disagreed on whether the wire needs its own approval gate, and each caught something the other missed. This branch reconciles them; the first is closed as superseded (#668).

## The trust model is respected, not bypassed

Everything the renderer was trusted to *state* is re-derived server-side for a wire caller.

- **Identity** — `senderPtyId` resolves to the owning workspace through the renderer and is stamped over any client-supplied `verifiedWorkspaceId`, the same D5 discipline `a2a.channel.*` and `task.mission.*` already use. An unresolvable sender fails closed.
- **Repo scoping** — `repoPath` is **not a wire parameter**. It is the git top level of the calling workspace's own directory, realpath'd on both sides so a subdirectory normalises for free and a symlink cannot alias another repo. An explicit `repoPath` is *rejected* rather than dropped: dropping it would let a caller believe it fanned out over a different repo than it did.
- **`agentCmd`** — not on the wire. The command builder interpolates it unquoted into a shell command, which is only safe because a human types it into the dialog.
- **`memberId`** — likewise. It is the caller's coordinate in the mission-channel roster and reaches `task.mission.start` unmediated.
- Local origin only. The reserved human workspace is refused. Commander brains are unchanged — those are allowlists, and they were not widened.
- Titles and effective prompts are capped where the request arrives, not only inside the service.

## Asynchrony was not optional

The MCP client's RPC deadline is 10s; one task's renderer spawn alone is allowed 30s. A synchronous handler therefore timed the caller out **by construction** — with N worktrees and N agents actually created, the caller believing it had failed, and the client's retry re-firing it.

The call now returns as soon as it has accepted the work, and the caller polls with the same idempotency key: `awaiting_approval`, `running`, `completed`, `denied`.

That makes a key's terminal state load-bearing in a way it was not when only the GUI called this:

- `FanOutService.start` records a **thrown** run as a failed result instead of releasing the key. Releasing it would let the next poll restart a fan-out that had already spawned tasks.
- **The poll is namespaced by the resolved caller workspace.** Keys are caller-chosen strings, so an unscoped poll would hand a neighbour the full result — task ids, branches, worktree paths — for a guessed key, and would let one caller squat another's.

## The approval prompt stays, but not on the A2A toggle

Agreeing that an agent may spawn a background agent is not agreeing that it may create N git worktrees and branches in your repository. So fan-out never auto-approves, and the dialog no longer offers a checkbox that would not apply to it.

Because the call is async, the prompt blocks nobody — so the 30s auto-deny costs only the fan-out. But not silently: the verdict carries its outcome, and a poll on a refused key reports `declined`, `timeout` or `unavailable` **with the reason**, rather than going quiet on an unattended machine.

The in-app dialog path stays ungated: clicking it is the approval. Approval is asked **before** the idempotency cache is consulted, so replaying a key cannot skip the gate.

## One service instance

`FanOutService` is now built in one place and shared by both front doors. Its idempotency LRU and its worktree serial queue are instance state, and a second instance is exactly the concurrent `git worktree add` that the queue exists to prevent.

## Tests

Each boundary requirement has a test that asserts the **refusal**: forged `verifiedWorkspaceId` ignored in favour of the derived one; explicit `repoPath` rejected; explicit `agentCmd` rejected; unresolvable `senderPtyId` fails closed; unresolvable pane cwd fails closed; approval denial starts nothing; task count over the cap rejected; cross-workspace poll refused. Plus the index-alignment regression for `titles`/`taskPrompts` when a blank title is present.

## Verification

- `tsc --noEmit` clean · `test:parallel` 8,794 passed · `test:runtime` 88 passed · eslint on changed files clean
- `scripts/gen-api-reference.mjs` regenerated — 145 methods

## Open questions

Recorded rather than decided here:

1. **Approval window length.** Still 30s. Now that the call is async, a longer window (or a non-expiring inbox item) would be kinder to someone away from the machine.
2. **Unattended fan-out.** If it should be possible, it wants its own opt-in — not a widening of the A2A toggle.
3. **`senderPtyId` vs peer PID.** Unchanged from the #113 same-user ceiling: a caller whose PID walk does not hit a known pane cannot fan out. Worth fixing across every D5 mutation at once, not here.
4. **Whether `plans/` should be tracked.** The design note for this work is untracked because of `.gitignore`. Not settled by this branch.

## Summary by CodeRabbit

* **New Features**
  * Added the `fanout_start` MCP tool to launch multiple isolated tasks from one request.
  * Added the `channel_mission_list` MCP tool to view the caller’s missions.
  * Fan-out requests now provide progress and completion status through polling.
  * Added clear approval prompts for fan-out operations; approval is always required.

* **Bug Fixes**
  * Improved mission-list access controls to prevent viewing another workspace’s missions.

* **Documentation**
  * Updated tool and API references to reflect the expanded MCP and RPC surfaces.


openwong2kim:
@codex review

openwong2kim:

This PR exposes fan-out — one prompt split into N isolated worktree tasks — on the MCP/pipe surface (task.fanout.start, plus the fanout_start and channel_mission_list tools), where it was previously reachable only from the in-app dialog. The change spans 23 files: a new pipe RPC handler with its own trust boundary (fanout.rpc.ts), a refactor that consolidates FanOutService into a single shared instance (createFanOutService.ts), the MCP tool wiring, the renderer approval dialog, and the shared RPC method/capability maps.

The trust discipline is the substance of this PR and it is done correctly. Everything the renderer was trusted to *state* is re-derived server-side for a wire caller: repoPath, agentCmd, verifiedWorkspaceId, and memberId are all rejected or resolved from the caller's verified senderPtyId rather than read from params, and an explicit repoPath/agentCmd is refused outright (not silently dropped) so a caller cannot act on a false picture of what it fanned out over. The accept-then-poll contract is forced by the 10s RPC deadline vs. 30s-per-task spawn, and the approval gate is deliberately decoupled from the A2A a2aAutoApproveExecute toggle — fan-out creates N worktrees and branches, which is not what that consent covers. The test suite pins every refusal path (forged identity, hostile agentCmd, cross-repo repoPath, unattended timeout, key squatting), and the single-instance invariant is asserted at the source level since no unit test can exercise the main/index.ts wiring.

One thing worth noting during review: the senderPtyId-based identity carries the same #113 same-user ceiling as the rest of the A2A surface — a process on the same machine can forge it — and the PR is explicit that this handler is no weaker and no stronger than the channel surface it mirrors. The four open questions recorded in the description (approval window length, unattended fan-out opt-in, senderPtyId vs peer PID, plans/ tracking) are the right things to defer. Assigning P2: this is a security-sensitive feature surface that should merge soon, but it is not a live breakage.

openwong2kim:
## Not mergeable yet — 5 blocking findings from a two-round, three-model panel

Reviewed by Claude, Codex and GLM 5.2, twice: once against the predecessor branch and once against this diff (1,925 lines). 20 findings, 11 after dedup. Every item below was re-verified against the source rather than taken from a model's word. Converting to draft.

**Command injection: not found.** Six independent reviews across both rounds followed `prompt`/`titles`/`task_prompts` into `buildInitialCommand` and none reaches unquoted interpolation — the prompt body is written to `prompt.md` and only the *path* enters the shell, single-quoted with POSIX `'\\''` / PowerShell `''` escaping. Titles flow through slug→branch→worktreePath→promptPath and terminate inside that escaping. Keeping `agentCmd` off the wire is sufficient, and that judgement is supported by all six reviews.

### Blocking

1. **The approval dialog does not show what the agents will actually be told.** `fanout.rpc.ts:447` builds the preview from the shared prompt plus titles; `task_prompts` — the per-task instructions that are actually injected — are absent entirely, and a 500-char slice is applied on top with no truncation marker. An attacker fills the shared prompt with 500 benign characters and puts everything else in `task_prompts`; the operator approves N autonomous agents without seeing a single character of their instructions. The cap allows 8 KB per task. This substantially cancels the benefit of separating fan-out from the A2A auto-approve toggle: the prompt appears, but its content does not.

2. **An evicted key lets an already-spawned fan-out be re-approved and re-run.** `setGate` evicts the oldest entry once `gates.size` exceeds the 1000 cap, regardless of state — `awaiting`, `denied` and `started` alike — and `FanOutService.results` has the same cap. When a key is evicted from both, `statusOf` returns `unknown` and the poll falls through as a *new* request: fresh approval prompt, full re-execution. The `expired` guard only fires while a `started` gate entry survives. The changelog line "a poll can never restart tasks that already spawned" is false on this path. Terminal states need a tombstone that outlives result-body eviction.

3. **The repository is derived from workspace metadata, not from the calling pane.** `resolveWorkspaceCwd()` reads `entry.metadata.cwd` from the renderer's workspace projection. That value is workspace-scoped and tracks the last `cd` of the *active* surface (OSC 7 / prompt scrape). So the calling agent can retarget it with one `cd`, and — new in this branch, which widened the predecessor's per-surface lookup to workspace scope — a *sibling* pane can retarget it for a caller that never moved. Rejecting an explicit `repoPath` does not constrain anything the caller cannot already set. Minimum: go back to the sender ptyId's per-surface cwd and re-verify the resolved root at approval time against preflight, rejecting on mismatch.

4. **The approval queue can be starved.** Each distinct key spawns a git subprocess and a detached approval RPC with no per-workspace or global in-flight ceiling, no rate limit and no post-denial cooldown (denial pins to a key, so a caller simply picks a new one). `pendingExecuteApprovals` is an unbounded record, the queue is FIFO, and the 30s timer starts at enqueue — so flooding fan-out prompts auto-denies ordinary A2A execute approvals before they are ever rendered. A security control degrading into a denial-of-service is not a low-severity outcome.

5. **`channel_mission_list` promises a scoping it does not enforce.** `task.mission.list` is registered as a read, so an unresolved `senderPtyId` leaves the caller-supplied `verifiedWorkspaceId` in place, and the MCP resolver falls back to the spoofable `WMUX_WORKSPACE_ID` env hint on a PID-map miss — a path where no `senderPtyId` is sent at all. A process launched with another workspace's id can read that workspace's mission titles, branches, absolute worktree paths, pane group ids and channel ids. The tool description ("Only your own missions are returned"), the code comment, and a test all assert the opposite. Either raise the method to fail-closed like the mutating ones, or correct all three to describe a documented same-user residual.

### Non-blocking, recorded

- `src/mcp/fanout.ts:103-108` collapses a denial into `Error [code]: message`, discarding `status`, `reason` and `idempotencyKey`. An unattended caller cannot distinguish "the user declined" from "the 30s window expired". Half of the async contract evaporates at the tool boundary.
- `verifiedWorkspaceId` is silently ignored while `repoPath`, `agentCmd` and `memberId` are explicitly rejected — the file's own stated principle is that silently overwriting leaves the caller acting on a false picture. Reject it too; better, whitelist-validate and reject unknown parameters.
- `idempotencyKey` has no length bound. Near-frame-size keys retained across 1000 gate entries and their detached closures are a memory amplifier, echoed again into daemon logs.
- Concurrent calls on one key can raise two prompts: `setGate` runs after two renderer round-trips and a git call, so both callers observe no gate. Double spawn is prevented by service idempotency, but with finding 1 the two prompts are visually indistinguishable while carrying different payloads.
- Fan-out from inside a linked worktree resolves `--show-toplevel` to that worktree while `git worktree add` and branch creation land in the parent's shared ref store — the approved path is not the repository actually mutated.

### Ruling on the identity ceiling

`senderPtyId` arrives as a parameter, not as a connection binding, so a same-user process can present a victim's ptyId (enumerable via `a2a.discover`) and land in that workspace's key namespace. This is the documented #113 same-user ceiling and is not introduced here — but the blast radius moves from "a message" to "git worktrees plus autonomously spawned agents", so it is accepted **explicitly and in writing** rather than by omission. It will be recorded as a known residual with the peer-PID work as its resolution, and the tool documentation must not claim protection it does not provide.

Findings 1, 2, 3 and 5 are required before this leaves draft. 4 and the recorded items may ship separately.

openwong2kim:
## Live dogfood — fan-out on the wire

Ran a dev build of this branch and hit `task.fanout.start` directly on the live
main pipe — real handler, real router, no test double:

```
no senderPtyId (unverifiable caller)        -> NOT_AUTHORIZED
forged repoPath                             -> NOT_AUTHORIZED
agentCmd on the wire                        -> NOT_AUTHORIZED
9 titles (cap is 8)                         -> NOT_AUTHORIZED
oversized idempotencyKey (4096 bytes)       -> INVALID_ARGUMENT
missing idempotencyKey                      -> INVALID_ARGUMENT
non-array taskPrompts                       -> NOT_AUTHORIZED
```

The ordering is visible in the answers and it is the right ordering: the two
argument-shape guards that precede identity fire on their own, and **everything
past that point is gated on a verifiable caller**. A wire caller that owns no
pane cannot reach the repo derivation, the approval prompt, or the service at
all. Fail-closed, live.

### Two things fixed on this branch

**1. CI was red for a reason that had nothing to do with the feature.** 34 of 65
tests in `fanout.rpc.test.ts` failed, in `validate` and `Baseline
(windows-latest)` alike, and they read as if the trust boundary itself were
broken — every repo derivation came back "not inside a git repository", so the
happy path, the approval preview, R1, R2 and R3 all went red at once.

The production code was fine. The handler resolves the caller's cwd with
`path.resolve()` before handing it to git — correct, it is about to be a
child-process cwd — and on win32 that turns the fixture `/repo/packages/app`
into `D:\repo\packages\app`, so the git stub keyed on
`dir.startsWith('/repo')` never matched. Fixtures are now built with
`path.resolve()` and the stub matches on the resolved roots. 65/65.

Worth noting because of how it presented: a POSIX-only path fixture in a
Windows-first repo fails *loudly but misleadingly* — it accuses the code under
test.

**2. `fanout_start` could not work as the first tool called on a fresh MCP
server** (Codex P1, confirmed). `MY_PTY_ID` is a **side effect** of the
workspace lookup — nothing walks the PID map until something asks who the caller
is. Every channel tool asks (`registerChannelTools` gets `resolveWorkspaceId`
and calls it first); `registerFanOutTools` was wired with the getter alone. So
it read `''`, sent no `senderPtyId`, and got the `NOT_AUTHORIZED` above. It
worked only if some unrelated identity-resolving tool happened to have run
earlier in the session.

Now wired the same way as the channel tools. The resolved workspace id is
discarded on purpose — the handler derives the owner from the ptyId and rejects
a stated one. Covered by a cold-server test; I mutated the warm-up out and
watched it go red.

### The other blocking findings, re-checked against the code

- **Per-task prompts in the approval preview** — in. `buildFanOutPreview(shared,
  titles, taskPrompts)` labels every task and reports truncated bytes.
- **Evicted key cannot restart a spawned fan-out** — in. `pending` and
  `terminal` are two maps now, and that split is the whole point; the `started`
  tombstone outlives the result body.
- **Repo from the calling pane, not workspace metadata** — in, and the sibling
  test pins it.
- **`channel_mission_list` scoping** — in (`b9a29df5`), fail-closed without a
  resolvable senderPtyId, and the tool description no longer claims protection
  past the #113 ceiling.

Finding 4 (approval-queue starvation: no per-workspace in-flight ceiling, no
rate limit, no post-denial cooldown) is untouched, as recorded. Worth a tracking
issue rather than leaving it in a PR comment — a security control that degrades
into a DoS on the *other* approval queue is not something to rediscover later.

### What I could not exercise live

The approved happy path. It needs a caller whose `senderPtyId` resolves to a
real pane, which means driving the GUI to spawn an agent pane and calling from
inside it. Everything up to and including the refusal is verified live; the
spawn side is unit-tested only.

Still draft — the CI fix and the identity fix are pushed, so the tree is green
whenever you want to take it out.

PR #675, opened 2026-07-28
feat: make agent-to-agent delegation expressible, and say what starts a turn

openwong2kim:
An orchestrating agent posted instructions into a mission channel for four workers. Every post returned success with a valid mention. **None of the four ever saw them.** The silence read as progress, and about two and a half hours went into waiting on work that had never started.

Nothing was broken. Three things were missing.

## A mention could not name a pane

`channelMentionFlush` delivers a mention by pasting into a pane, and it requires a pane pin to do so — **by design**: a workspace-level mention may have been meant for the human, and pasting one into "the one live agent" is a mistake that an earlier adversarial review caught in the field.

But the MCP `channel_post` schema let a caller name a `workspace_id` and a `member_id` and **not a pane**. So an agent-to-agent instruction posted to a channel was undeliverable *by construction*: it raised an unread badge and waited to be polled, which an agent that has just finished its work is not doing. Nothing errored on either side.

`mentions[].pane_id` is the missing vocabulary. A pinned mention rides the **existing** delivery path — no new transport, no new route. Unpinned mentions are untouched and stay badge-only, so the rule that caught the earlier bug is intact.

**The pin is proven before it is persisted.** The principal registry keys panes as `pane:${workspaceId}/${paneId}`, so a lookup hit *is* the proof that the pane belongs to the mentioned workspace. Without that check the new field would be a way to aim a paste at a pane in a workspace the caller merely shares a channel with. An unprovable pin is refused, the mention degrades to the workspace-level badge it would have been anyway, and the refusal returns in `droppedMentions` as `pane_not_in_workspace` — this path does not drop things quietly.

Liveness is deliberately *not* checked here: a stale record still proves ownership, and only the receiving renderer can prove a pty is alive.

## Nothing said that a post does not start a turn

From the outside, an idle worker and a working worker are indistinguishable. That is what made the failure cost hours instead of minutes — there was no way to tell "not delivered" from "delivered, still thinking".

Three surfaces now carry the contract:

- **`channel_post`** says a post does not start a turn, and names what does.
- **`send_message` / `a2a_task_send`** say they are the delivery that does.
- **Fan-out workers** get the same paragraph appended to their opening prompt. It goes after the caller's text and after the byte-cap check, so it cannot displace or reorder what the caller wrote. This is the one place a worker is guaranteed to read before it first goes idle — a doc it must choose to open would not be.

`docs/how-to/delegate-to-agents.md` is the long form.

## A waiting call could not be expressed

The MCP client's transport constants — 10 s timeout, 3 retries — make a long wait impossible: it is cut at 10 s, and the retry ladder would reopen the same wait up to three times over. **Raising the global timeout alone would be strictly worse than the status quo.**

`sendRpc(method, params, { longPollMs })` raises that one call's ceiling (capped at 15 minutes) and switches its retries off. Both effects hang off a single option on purpose: *"long timeout, keep retrying"* is the bad state and must not be reachable. Endpoint fallbacks stay, because they fire only on a connect-class failure — which means no wait was ever opened.

**Plumbing only.** Nothing calls it yet; the server-side wait RPC it exists for is separate work. The tests are its current consumer and cover both halves against a local fake daemon.

## Tests

The pin path is covered three ways, each mutation-verified: a pinned mention is delivered; an unpinned mention still stays badge-only; a pin naming a pane in another workspace is refused rather than silently dropped.

## Known follow-up

`droppedMentions` now carries a new reason, and the renderer toast renders every reason as "did not land". For `pane_not_in_workspace` that is not quite true — the badge *did* land; only the paste was refused. Splitting the copy per reason belongs with the wider status-vocabulary work, not here.

The guard is applied uniformly, including to the human composer path. Narrowing it to "only the new shape" would carry zero regression risk, but then sending a raw RPC with a ptyId would route around it and the guard would be decorative.

## Summary by CodeRabbit

* **New Features**
  * Channel mentions can target a specific agent pane using `pane_id`.
  * Added long-polling options for extended message operations, with waits up to 15 minutes.
  * Fan-out workers now receive clear instructions for handling delegated tasks and follow-ups.
* **Bug Fixes**
  * Invalid, unavailable, or cross-workspace pane targets are safely rejected and reported.
  * Improved message delivery behavior prevents unpinned notifications from unexpectedly interrupting agents.
* **Documentation**
  * Added guidance covering delegation workflows, pane-targeted mentions, delivery behavior, and failure reporting.


openwong2kim:
@codex review

openwong2kim:

This PR closes a silent failure in agent-to-agent delegation: an orchestrating agent could post instructions into a mission channel for four workers, every post would return success, and none of the workers would ever see them — because `channel_post` mentions had no way to name a specific pane, and pane-pinning is exactly what the existing delivery path keys on. The fix adds an optional `pane_id` to mentions (delivered via the principal registry's ownership proof, with unprovable pins refused and reported in `droppedMentions` rather than dropped silently), spells out the delegation contract in the tool descriptions and a new `docs/how-to/delegate-to-agents.md`, and appends a `WORKER_DELIVERY_PREAMBLE` to every fan-out prompt so a worker knows before it goes idle that a workspace-level post will not wake it. The third piece — `sendRpc(..., { longPollMs })` on the MCP client — is plumbing only, but it is designed correctly: raising the timeout and disabling retries hang off a single option on purpose, so the strictly-worse "long timeout, keep retrying" state is not expressible.

The trust model is respected rather than bypassed: `senderPtyId` resolves to the owning workspace and overrides any client-supplied `verifiedWorkspaceId`, `repoPath` is rejected on the wire and re-derived from the caller's own directory, and `agentCmd`/`memberId` never reach the wire at all. Tests cover the three load-bearing properties (pinned mention is pasted; unpinned mention stays badge-only; a cross-workspace pin is refused) plus the mutation-verified refusals, and the long-poll path is exercised against a real socket fixture. The known follow-up — `droppedMentions` reasons all rendering as "did not land" in the toast — is correctly scoped out. This is the kind of change that looks small on the diff and prevents hours of "is it working?" debugging; worth landing.

Priority: **P2** — fixes a real, observed, hours-lost failure mode in the flagship fan-out journey, with a clean trust boundary and no wire break.

openwong2kim:
## Dogfooding checklist

This PR is the fix for a failure we hit while running this repository's own agents, so the useful verification is behavioural rather than unit-level. Three things worth exercising against a real build:

1. **A pane-pinned mention actually starts the recipient's turn.** Post to a mission channel with `mentions[].pane_id` set to an idle agent's pane and confirm it begins work — not just that an unread badge appears. Before this change that was unreachable by construction, and the symptom was silence rather than an error.

2. **An unpinned mention still stays badge-only.** This is the safety rule an earlier adversarial review put there: a workspace-level mention may have been meant for the human, and pasting one into whichever agent happens to be live is how a message reached a PTY it was never addressed to. The default must not have moved.

3. **A pin naming a pane in another workspace is refused**, and the refusal arrives as `pane_not_in_workspace` in `droppedMentions` rather than vanishing. This guard is new — without it the field would be a way to aim a paste at a pane in a workspace the caller merely shares a channel with — so it is the part most worth confirming outside the test suite.

Note for whoever reads the toast: `droppedMentions` renders every reason as "did not land", which for `pane_not_in_workspace` is not quite accurate — the badge did land, only the paste was refused. Splitting that copy per reason belongs with the wider status-vocabulary work rather than here.

openwong2kim:
## Live dogfood — pane-pinned delegation

Ran a dev build of this branch and drove the **live daemon over its control
pipe** (`\.\pipe\wmux-daemon-dev-*`) — real `ChannelService`, real persistence,
not a test double.

```
daemon.ping            -> {"status":"ok","sessions":1,…}
a2a.channel.create     -> {"ok":true,"channel":{"id":"ch-1668f00c-…"}}

pin an unknown pane
  droppedMentions      -> [{paneId:"pane-does-not-exist", reason:"pane_not_in_workspace", …}]
  persisted mentions   -> [{workspaceId:"ws-probe-…", name:"probe"}]     <- badge-only, no pane coordinate
```

The ownership gate behaves exactly as documented end to end: the pin is
stripped, the mention still lands at workspace level, and the refusal comes back
to the sender instead of vanishing. App and daemon boot clean, 0 plugin
failures.

**A second gate confirmed by accident, and worth calling out as working:** I
tried to register a synthetic pane principal to stage the dead-pane case, and
`a2a.principal.upsert` refused it — `record.ptyId` must resolve to a real
session *of that workspace*. So the registry cannot be seeded with a pane that
does not exist, which is the right hardening and it is live.

### One finding, fixed on this branch

`resolvePanePrincipal` answered from `PrincipalService.find()`, which returns a
**stale** record and its last-known `ptyId` — and that ptyId was filled in as the
route-time snapshot for an MCP pin.

A stale snapshot does not fail loudly. The receiving renderer's fail-closed
match misses, the mention degrades to workspace level, and the workspace-level
flush pastes into whatever agent the workspace still has. So an instruction
addressed to a **departed** worker starts a *sibling* worker's turn, and the
sender is told nothing. That is strictly worse than the unproven-pin case this
gate already refuses — and silent where that one is reported.

The comment said liveness was deliberately not gated because "a stale record
still proves ownership". True, but the two answers have different consequences,
and `PrincipalService.livePtyId()` two hundred lines up already draws exactly
this line for the wake worker: *"a stale record's ptyId may be a dead (or
reused) session id, so it is never [handed out]"*. Ownership and liveness are
now answered separately, and a missing ptyId becomes a reported refusal —
`pane_not_live`, alongside `pane_not_in_workspace`. The tool description and
changelog name both; the test registry models all three answers instead of
collapsing the middle one.

### What I could not stage live, and it is yours to close

The stale transition itself. `markStaleByPtyId` fires on `session:died` /
`session:destroyed`, and the upsert hardening above (correctly) blocks
synthesizing a principal to get there — so reproducing it needs a **real agent
pane that dies**. The one dogfood step worth doing by hand before merge:

1. Two panes in one workspace, both agents.
2. From another workspace, `channel_post` with `pane_id` pinned to pane A.
   Confirm A's prompt receives it.
3. Kill A's agent. Pin A again.
4. Expect `droppedMentions: [{reason: "pane_not_live"}]` and **pane B silent**.
   Before this fix, B takes the turn.

Everything else: 946 tests green across `src/daemon/channels` + `src/renderer/hooks`,
typecheck clean, CI green.

PR #676, opened 2026-07-28
chore(licenses): fail the build on drifted notices and undetermined licenses

openwong2kim:
`THIRD_PARTY_NOTICES` had gone **two months and 18 packages** out of date, because nothing ever ran the generator. This regenerates it and adds the guards that would have caught it.

Two guards on the existing `npm test` gate — no new CI workflow. The same arrangement that already keeps `docs/api/reference.md` from drifting.

## Drift

`generate-notices.mjs` gains `--check` (byte-compare against the committed file) and `--stdout-digest`; a test runs both.

A second test asserts that **two runs hash identically**. Determinism is what the guard rests on, and *a flapping guard gets deleted rather than fixed* — so it is pinned before anything is built on top of it.

## An undetermined verdict fails

`check-licenses.mjs` vets the same lockfile closure the notices file documents, so we cannot publish attribution for one package set while vetting another. It is deny-by-default: a package passes only if its declared license is on an explicit permissive allow list.

**The part that matters is that "could not determine" is a violation, not a pass.** `NOASSERTION`, `"other"`, `UNKNOWN`, empty, and npm's non-SPDX `"SEE LICENSE IN <file>"` all fail.

A deny-list alone does not hold. Automated license detection reports `"other"` for projects whose `LICENSE` file, read by a person, is the GNU General Public License — we hit exactly that case while surveying prior art for unrelated work. Anything treating "could not determine" as "no hits found" waves precisely that case through green. The failure message says so, and tells the reader to open the file.

## The manifest is what lies

A third check reads the `LICENSE` text each package actually ships and rejects a copyleft body **regardless of the declared identifier** — because in this failure mode the manifest is the part that is wrong.

## Exceptions are written down

`license-allowlist.json`, keyed on exact name + version + declared license, each entry carrying a written reason recording what was read. A version bump ships a new artifact and retires the exception.

The nine current entries are the `@anthropic-ai/claude-agent-sdk` family, whose `LICENSE.md` was read directly: Anthropic proprietary terms, not copyleft.

Allow list is MIT, ISC, BSD-2-Clause, BSD-3-Clause, Apache-2.0 and Unlicense — every license in the current closure — plus 0BSD, MIT-0, CC0-1.0 and Python-2.0, which are equally permissive and common enough transitively that omitting them would hard-fail a routine bump.

## Mutation-verified

Each reverted afterwards:

- a one-character edit to `THIRD_PARTY_NOTICES` → drift guard fails, green again on restore
- fake lockfile entries declaring `NOASSERTION`, `"other"`, `GPL-3.0-or-later` and no license field → all four reported, correctly classified, exit 1
- GPL text dropped into a package declaring MIT → caught by the text scan

## Note

The guards run inside `ci-cross-platform-baseline.yml`, which executes `npm test` on three operating systems. Determinism was demonstrated locally; the Windows and Linux runners confirm it here for the first time.

Why this matters for this repository specifically: we ship a signed, notarized Electron binary. A copyleft dependency arriving unnoticed is not a lint failure — it is a source-disclosure obligation that cannot be walked back after release.

## Summary by CodeRabbit

- **Compliance**
  - Added an automated release gate that blocks shipping when dependency licenses are unapproved, undetermined, or copyleft.
  - Added an additional check that rejects copyleft content found in bundled license texts, regardless of metadata claims.
  - Added a CLI license audit with clear pass/fail output and optional JSON reporting.
- **Documentation**
  - Regenerated third-party notices deterministically from the production dependency set and updated notice blocks and metadata.
- **Quality Improvements**
  - Added drift guards to ensure third-party notices remain byte-for-byte reproducible and aligned with the current dependency closure.
  - Added tests covering license classification, copyleft detection, and allowlist validation.


openwong2kim:
@codex review

openwong2kim:

This PR builds a deny-by-default license gate on the production dependency closure and brings `THIRD_PARTY_NOTICES` back in sync after it had drifted 18 packages and two months out of date. The two new pieces — `check-licenses.mjs` (SPDX expression evaluation plus a copyleft text scan against each shipped LICENSE, undetermined verdicts failing by default) and the determinism fixes to `generate-notices.mjs` (lockfile closure instead of `npm ls`, no wall-clock stamp, LF normalization, `--check`/`--stdout-digest` modes) — are the right shape: the guard rests on byte-for-byte reproducibility, and the policy closes the specific hole where automated detection reports `other` for what is, on inspection, the GPL. The allowlist/waiver file is version-pinned with a written reason per entry, which is the only honest way to record a human reading, and the tests are mutation-verified against each axis they claim to defend.

The "why it matters" is specific to this repo: wmux ships a signed, notarized Electron binary, so a GPL/AGPL dependency reaching the bundle is a source-disclosure obligation that cannot be walked back after release — a lint failure is the best-case outcome of missing it. Resolving the package set from `package-lock.json` rather than the installed tree also makes the notices file correct across all three build platforms at once, which `npm ls` could never have been. Worth landing before the next release tag; nothing here changes runtime behavior, so the risk surface is the build scripts and the regenerated file itself.

Priority: **P2** — compliance infrastructure for the shipping artifact, important to land before any release, but not a live runtime defect.

openwong2kim:
Amended after reading the linked terms.

The nine `@anthropic-ai/claude-agent-sdk` exceptions were justified on one axis — not copyleft, therefore no source-disclosure obligation. That reading is correct, and it is also the half of the question that does not apply to a project shipping a signed installer.

`LICENSE.md` reserves all rights. The Anthropic terms it links grant permission to **use the Services** and are silent on redistributing, sublicensing or bundling SDK packages. Both were read; neither permits nor forbids it. wmux does not merely use this package — it ships it inside the installer, which is redistribution rather than use.

Silence is not permission, and an entry that reads "not copyleft" invites the next reader to treat the matter as settled when it was never asked. Each entry now carries `redistributionStatus: unresolved` and records what was read, what was absent, and that removing the package from the bundle in favour of resolving it on the user's machine would retire the question outright.

**Nothing about the check changes** — these packages were already passing by explicit exception and still do. This is a correction to the written justification, not to the policy.

The file header now asks for both axes on every future entry. The failure here was not a wrong answer; it was a question that never got asked — which is the same shape as the `NOASSERTION` case this PR exists to close.

openwong2kim:
## On the CI failure

Expected axis, flagged before the push: determinism of the notices generator was demonstrated locally, and these runners are the first time it is exercised on Windows and Linux.

**If the drift guard is what fails, the fix is the generator, not the guard.** If `generate-notices.mjs` emits different bytes per platform — path separators, sort order, line endings — then the guard is correctly reporting a real problem, and loosening it would leave the drift it exists to catch. A flapping guard gets deleted rather than fixed, which is why the two-runs-hash-identically test is in this PR at all.

If instead the license policy check is what fails, that is more interesting: it would mean the lockfile closure resolves differently on another platform, and the declared licenses of a package set we ship differ from the ones we vetted. That is worth understanding rather than allowlisting.

openwong2kim:
## Live dogfood — license gate

Ran both gates against a **clean `npm ci` of this branch's own lockfile** (not a
reused tree — the previous install was three releases stale and was missing a
package the lockfile lists, which is its own small lesson about testing a
lockfile change against the lockfile).

```
$ node scripts/check-licenses.mjs
[licenses] checked 128 production packages from package-lock.json.
[licenses] waived  @anthropic-ai/claude-agent-sdk@0.3.206 — Proprietary Anthropic terms, not copyleft.
… 8 more platform siblings …
[licenses] OK — every production dependency is permissively licensed or explicitly waived.
exit=0

$ node scripts/generate-notices.mjs --check
[notices] 128 unique production packages in the lockfile closure.
[notices] THIRD_PARTY_NOTICES is up to date (128 packages).
exit=0
```

Both clean. 128/128, nine waivers, no drift.

### On the CI failure

The prediction in the thread was that a drift/determinism axis would break on a
non-Windows runner. It was neither — and worth recording, because the actual
cause is invisible from the summary.

`scripts/__tests__/checkLicenses.test.mjs` imported `scripts/check-licenses.mjs`
directly, and that file carries `#!/usr/bin/env node`. vitest cannot parse a
module with a shebang, so the whole suite failed to load with a bare
`SyntaxError: Invalid or unexpected token` — **the gate never ran at all**, on
any platform. A green-looking split (615 passed / 1 failed) hid the fact that
the one failure was the new gate itself.

Fixed by splitting along the line the rest of `scripts/` already uses:
`scripts/lib/*.mjs` is importable and has no shebang, `scripts/*.mjs` is the CLI
entry point. The audit moved to `scripts/lib/license-audit.mjs`.

### Reviewer findings — four holes, all reproduced before changing anything

1. **Precedence (P1).** The expression was flattened by splitting on `/OR|AND/`,
   so `(MIT AND GPL-3.0) OR (GPL-2.0 AND AGPL-3.0)` found MIT among four loose
   operands and returned **allowed** — although every branch you could elect
   carries a copyleft obligation. Replaced with a parser for the real SPDX
   grammar (AND binds tighter than OR) and a three-valued evaluation over the
   tree. An expression that does not parse as a whole is now undetermined
   rather than judged on the prefix that did.

2. **`BSL-1.0` was on the deny prefixes (P2).** That is the Boost Software
   License — OSI-approved, permissive. Business Source is `BUSL-*`, already
   covered by the next entry. Because a denied verdict cannot be waived, a
   Boost dependency could not have been approved at all.

3. **The allowlist could waive a denial (P1).** The lookup ran for a `denied`
   verdict as well as an undetermined one, so an entry naming `EUPL-1.2`,
   `OSL-3.0` or `CPAL-1.0` would have passed the production gate — against this
   file's own rule that copyleft is removed, never waived. The invariant test
   only caught a hand-written subset that omitted exactly those three; it now
   checks against the policy's own prefix list.

4. **The empty-declaration path was unreachable (CodeRabbit, Major).** A package
   with no license field keyed as `name@version|`, and `loadAllowlist` rejects
   an empty `declaredLicense` — so the case the header names as reviewable could
   never be reviewed. Both sides now go through one helper with a `(none)`
   sentinel.

Also stated rather than implied: a package **absent from disk** (an os/cpu
variant on a runner that is not its target) is judged on its declaration alone,
so the shipped-text check cannot see it. The comment claimed those were covered
by allowlist review; only the undetermined ones are.

Two items left as findings rather than changes, because both are policy calls:

- A dual-licensed package shipping `LICENSE-MIT` **and** `LICENSE-GPL` trips the
  text scan before its `(MIT OR GPL-2.0)` election is considered, and the
  `continue` runs before the allowlist, so there is no waiver path. No such
  package is in the closure today. Loosening it means letting a copyleft *body*
  through on the strength of a manifest, which is the thing this PR exists to
  distrust — your call.
- Forcing an allowlist entry for every platform-constrained package with an
  allowed declaration would pull in every `@esbuild/*`-shaped optional dep. That
  is a real tightening, but a large one, and not mine to make.

The production closure is unaffected by all of the above: same 128 packages,
same nine waivers, exit 0.

PR #677, opened 2026-07-28
refactor(terminal): bind the renderer and daemon to one width model

openwong2kim:
The renderer (`useTerminal.ts`) and the daemon's two snapshot terminals (`HeadlessSnapshot.ts`) each named the unicode addon and the version string themselves. Nothing tied them together.

That is worse than it sounds. An addon bump that updated one site and missed another would leave the two grids measuring differently — **and every test would still pass**, because a snapshot only paints cell-shifted once it is restored onto a grid that wraps rows at a different column. The failure is invisible until a restored pane draws one column off from the live one.

All three now go through `applyUnicodeWidthModel` in `src/shared/terminalUnicode.ts`: one addon choice, one version string, in one place.

**No width changes.** The model stays Unicode 11 and every measured width is identical to before.

## Two guards, failing for different reasons

- **The version guard asserts the id is *registered*, not merely assigned.** An addon that renames or drops its version does not throw on assignment — the terminal quietly falls back to xterm's Unicode 6 default and every width changes.
- **The width table pins absolute numbers** rather than diffing against a second addon, because the differential form only means anything while two width models are in the tree. Its CJK / box-drawing / block / braille rows measure the same under Unicode 6, so they do not catch a *missing* width model — they catch a future width-model **change**, which is what tears TUI frames drawn over Korean text. Its emoji rows are what Unicode 11 buys over the default and do catch a missing one, with a premise test pinning that difference.

The known cost of per-codepoint measurement — a ZWJ family at 6 cells, a skin-tone thumbs-up at 4 — is asserted too, so the trade stays visible and a future change to it reads as a decision rather than a surprise.

## Why Unicode 11 and not grapheme clustering

The module records this, because it is the answer to the next person who proposes the swap — and it was evaluated in full before being declined (#671).

- **VS16 is the only class it changes for our content**, and the programs running inside our panes still measure it as one column. Counting this repository: VS16 in 26 files; ZWJ families in 2, both fixtures written for that change; skin tones and keycaps, zero.
- **Upstream mis-reads its own compressed trie** when Node hands it a pooled `Buffer` — https://github.com/xtermjs/xterm.js/issues/6079. Three symptoms depending on what the process allocated first, the worst being silently wrong widths with nothing thrown.
- **One grapheme cluster in one cell removes the memory bound a row-counted scrollback implies**: 13 MB of input retained 214 MB.
- **CJK throughput measured ~37% lower.**

## Verification

`tsc --noEmit` clean · `test:parallel` and `test:runtime` green · eslint on changed files clean. Both guards mutation-verified: removing the addon load, and renaming the version string, each turn their own test red.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency in how terminal content measures Unicode characters, including CJK text, combining marks, box-drawing symbols, and emoji.
  * Aligned rendered terminal output and text snapshots to use the same Unicode width behavior.

* **Tests**
  * Added coverage for Unicode character widths and multi-codepoint emoji sequences to help prevent display and cursor-position regressions.

### DIFF
--- a/src/daemon/HeadlessSnapshot.ts
+++ b/src/daemon/HeadlessSnapshot.ts
@@ -29,7 +29,7 @@
 
 import { Terminal } from '@xterm/headless';
 import { SerializeAddon } from '@xterm/addon-serialize';
-import { Unicode11Addon } from '@xterm/addon-unicode11';
+import { applyUnicodeWidthModel } from '../shared/terminalUnicode';
 import {
   PartialSequenceTracker,
   MarginTracker,
@@ -188,8 +188,7 @@ async function generateTextInner(req: SnapshotRequest): Promise<TextSnapshotOutc
     logLevel: 'off',
   });
   try {
-    terminal.loadAddon(new Unicode11Addon());
-    terminal.unicode.activeVersion = '11';
+    applyUnicodeWidthModel(terminal);
 
     let utf8Carry: Buffer = Buffer.alloc(0);
     let bytesIn = 0;
@@ -287,11 +286,12 @@ async function generateInner(req: SnapshotRequest): Promise<SnapshotOutcome> {
   const serializer = new SerializeAddon();
   try {
     terminal.loadAddon(serializer);
-    // Width parity with the renderer (useTerminal sets the same): without
-    // Unicode 11 tables, CJK/emoji rows advance the cursor differently here
-    // than on screen and the snapshot paints cell-shifted tears.
-    terminal.loadAddon(new Unicode11Addon());
-    terminal.unicode.activeVersion = '11';
+    // Width parity with the renderer (useTerminal calls the same helper):
+    // without the Unicode 11 tables, CJK/emoji rows advance the cursor
+    // differently here than on screen and the snapshot paints cell-shifted
+    // tears. Going through the shared helper is what keeps the two sides from
+    // drifting — do not inline the addon here.
+    applyUnicodeWidthModel(terminal);
 
     const partialTail = new PartialSequenceTracker();
     const margins = new MarginTracker();
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -3,7 +3,7 @@ import { Terminal } from '@xterm/xterm';
 import { FitAddon } from '@xterm/addon-fit';
 import { WebglAddon } from '@xterm/addon-webgl';
 import { SearchAddon } from '@xterm/addon-search';
-import { Unicode11Addon } from '@xterm/addon-unicode11';
+import { applyUnicodeWidthModel } from '../../shared/terminalUnicode';
 import { WebLinksAddon } from '@xterm/addon-web-links';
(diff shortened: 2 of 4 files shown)


openwong2kim:
@codex review

openwong2kim:

This PR consolidates the terminal Unicode width model into a single shared helper (`applyUnicodeWidthModel` in `src/shared/terminalUnicode.ts`) so that the renderer's xterm grid (`useTerminal.ts`) and the daemon's two headless snapshot grids (`HeadlessSnapshot.ts`) can no longer drift independently — previously each site named the addon and version string itself, and a future addon bump that touched one and missed the other would have left the two grids measuring differently while every test still passed green. The change introduces no width changes (Unicode 11 is retained), and ships two complementary regression guards: a version-registration assertion (catching a renamed/dropped version silently falling back to xterm's Unicode 6 default) and an absolute width table pinning CJK, box-drawing, and emoji rows that catches a future width-model change as well as a missing one. The rationale for declining grapheme clustering is documented thoroughly in the module header, including the upstream xterm.js trie-decoding bug, the unbounded memory risk, and the measured 37% CJK throughput cost. This is sound, low-risk hygiene work that makes a latent and hard-to-diagnose class of snapshot-renderer divergence structurally impossible. Assigning P3 — it is a pure refactor of existing behavior with no active defect to fix, but it is a clear correctness improvement worth landing.

openwong2kim:
## Live dogfood — one width model

Built a **real `@xterm/headless` terminal the way `HeadlessSnapshot` builds one**
(`allowProposedApi: true`, same package), applied the shared helper, and read
widths back off the actual grid via the cursor column — rather than asserting on
a table.

```
helper registers version   : 11
activeVersion  (no helper) : 6
activeVersion  (helper on) : 11

    hangul   가나다      default=6  unicode11=6
    han      漢字漢      default=6  unicode11=6
    box      ─│┌┐        default=4  unicode11=4
  * emoji    👍          default=1  unicode11=2
    ascii    abc         default=3  unicode11=3
```

Two things this confirms live rather than by inspection:

1. **The drop-to-default failure mode is real and silent.** A terminal that does
   not run `applyUnicodeWidthModel` reports `activeVersion: '6'` — it does not
   throw, it does not warn, it just measures differently. That is exactly the
   divergence the version-registration assertion in this PR exists to catch, and
   it is invisible in any test that only checks widths the two versions agree on.

2. **The rows that actually differ are the emoji ones.** CJK and box-drawing
   measure identically under 6 and 11, so a width table built only from those
   would pass with the addon missing. `👍` at 1 cell vs 2 is the observable
   divergence — one cell of drift per emoji between the renderer grid and the
   daemon's snapshot grid, which is precisely the "same session, two different
   screens" bug that has no obvious cause when you hit it.

The absolute width table in this PR does include emoji rows, so it would have
caught it. Good.

No behaviour change confirmed: Unicode 11 is retained on both sides, and the
helper's refusal to hand back the addon instance is the right call — it is the
only thing stopping a third site from managing the model separately.

Test suites green locally on this branch; CI green.

PR #679, opened 2026-07-28
fix(mcp): fail closed when browser page selection cannot be scoped to the caller

openwong2kim:
`resolveCallerSurface()` answered `{ kind: 'unscoped' }` on **every** path where the calling session's workspace could not be established — and an unscoped selection context reaches any workspace's live guest: `findViaTargetDomain` falls back to `cdp.info`'s `targets[0]`, and `_getPageImpl` falls back to "the first non-shell page in any context".

So an agent that could not be identified got read/write access to whatever browser happened to be open.

## No spoofing is required

Every sibling on the same identity miss already refuses: `browser_open` and `browser_close` throw from `requireWorkspaceId`, `browser_tabs` returns `BROWSER_TABS_WORKSPACE_UNRESOLVED`, and the engine's own auto-open issues no RPC at all. **Page selection was the one path that stayed lenient.**

And reaching it needs no forgery. An agent sandboxed away from process-tree inspection *legitimately* cannot resolve its own identity — the case is documented in `lookupPidMapWorkspace`. The crossing happened for a well-behaved agent under normal conditions, which is why the known same-user identity ceiling does not absorb it.

## The state is removed, not merely blocked

`'unscoped'` is gone from the union. With it gone, the two workspace-blind selection sites are **unreachable by construction**: `resolveSelectionContext` can no longer produce a context with neither a `surfaceId` nor `callerHasNoSurface`. A guard can be routed around later; a state that no longer exists cannot come back.

Each refusal throws `WORKSPACE_SCOPE_UNRESOLVED` carrying the specific reason — identity unresolved, `cdp.info` unavailable, or a main too old to tag targets — mirroring the shape of `EXTERNAL_BACKEND_UNSUPPORTED`. It logs as well: several tools swallow the throw via `getPage().catch(() => null)`, so without the log the reason would simply vanish.

## One case stays functional

A legacy main reporting **no** live targets is unambiguous no matter whose they would be, so it resolves to `'none'` and the caller auto-opens its own surface. That is what the old leniency was actually protecting — a single-workspace setup with nothing to cross into.

## Tests

Refusal instead of returning a foreign live page (asserting no CDP connect and no `browser.open`), per-branch reason text, the legacy-untagged refusal, and the legacy-empty `'none'`.

Two existing tests had encoded the leniency in their mocks — page discovery with no resolver wired, and auto-open re-resolving against a main that never reports the opened target. Both now model a main that scopes and tags, so they exercise the paths that actually run.

## Provenance

Found by a survey of every workspace-scoping decision in the codebase (164 sites), which flagged this one on the asymmetry criterion: sibling surfaces failing closed on an identical failure while this one failed open. Verified against source before being acted on.

## Summary by CodeRabbit

* **Security**
  * Page selection now fails closed when the caller’s workspace cannot be verified, preventing accidental access to another workspace’s browser page.
  * Clear `WORKSPACE_SCOPE_UNRESOLVED` errors identify unresolved identity, unavailable browser control, or unsupported target metadata.

* **Bug Fixes**
  * Preserved safe auto-opening when no browser is open.
  * Improved page routing for scoped browser targets and legacy browser environments.
  * Strengthened browser session cleanup and shell-page selection behavior.


openwong2kim:
@codex review

openwong2kim:

This PR closes a fail-open gap in workspace-scoped browser page selection inside `PlaywrightEngine.ts`. When the calling session could not resolve its own workspace identity — an entirely legitimate state for an agent sandboxed away from process-tree inspection — `resolveCallerSurface()` returned `{ kind: "unscoped" }`, and that unscoped selection context fell through to "the first live browser guest in any context." The result was read/write access to whichever workspace happened to have a browser open, which is exactly the cross-workspace boundary scoping exists to prevent. Every sibling path (`browser_open`, `browser_close`, `browser_tabs`, auto-open) already refused in this state; page selection was the one route that stayed lenient, and it required no spoofing to reach.

The fix removes the `unscoped` variant from the union entirely rather than bolting on a guard, which makes the two workspace-blind selection sites (`findViaTargetDomain` falling back to `targets[0]`, and the "first non-shell page" heuristic) unreachable by construction — `resolveSelectionContext` can no longer produce a context with neither a `surfaceId` nor a resolved `callerHasNoSurface`. Every now-unreachable path throws `WORKSPACE_SCOPE_UNRESOLVED` carrying the specific reason (identity unresolved, control connection down, or a main too old to tag targets), and logs it as well because several callers swallow the throw via `getPage().catch(() => null)`. One legitimate case stays functional: a legacy main reporting zero live targets is unambiguous, so it resolves to `none` and the caller auto-opens its own surface — that is the single-workspace setup the old leniency was actually protecting.

Scope is well-bounded: 291 additions / 71 deletions across `PlaywrightEngine.ts`, its test file, and a CHANGELOG entry, and all CI (validate, three-platform baseline, perf) is green. The CHANGELOG calls out, correctly, that tool paths which fall back to the app's own screenshot/evaluate handlers still pick a default target the way they always have — that is a separate gap and explicitly out of scope here. Priority P2: it is a real isolation defect that sits outside the #113 same-user ceiling (no file or process access was needed, only an unresolved identity), the fix is cheap and well-tested, and the asymmetry criterion the PR was found under — siblings failing closed while this one failed open — is the right signal for shipping it promptly.

openwong2kim:
## Dogfooding note

This change makes a previously lenient path refuse, so the risk it carries is **over-locking**, not under-locking. Worth confirming against a real build that ordinary browser use is untouched:

- `browser_open` / `browser_snapshot` / `browser_evaluate` / `browser_extract_*` all still work from a normal agent pane whose identity resolves.
- A workspace with no browser open still auto-opens its own surface rather than erroring — that is the `'none'` path this PR deliberately preserved, and it is the case the old leniency was actually protecting.

If something that used to work now throws `WORKSPACE_SCOPE_UNRESOLVED`, the reason string says which branch refused (identity unresolved, `cdp.info` unavailable, or a main too old to tag targets), and it is logged as well — several tools swallow the throw via `getPage().catch(() => null)`, so stderr is where to look.

One thing this PR deliberately does **not** cover: a caller that supplies an explicit `surfaceId` rather than relying on resolution. The scoping survey raised the question of whether that path validates ownership; it was not traced to a conclusion before the survey ended. If it turns out to be unguarded it is the same defect through a different door, and it should be a separate change rather than widened into this one.

openwong2kim:
## Open before merge: does the explicit-`surfaceId` path validate ownership?

**This PR closes one door. There may be a second one, and it has not been traced to a conclusion.**

What this PR fixes is the *identity-resolution-failure* path: when the caller's workspace cannot be established, page selection used to fall back to any live guest, and now it refuses.

What it does **not** address is a caller that supplies an explicit `surfaceId` rather than relying on resolution. There, identity may resolve perfectly well — the question is whether the named surface is checked to belong to the caller's workspace before it is used. If it is not, that is the same defect through a different door, and merging this as "fixed" would leave it half-closed while reading as complete.

**Where to pick it up:** `PlaywrightEngine.ts` — `resolveSelectionContext` and its explicit-surface branch. The survey that found the original defect was inspecting `gh pr diff 679` for `explicitSurfaceId` / `t.surfaceId === surfaceId` / `assertSurface` / `ownsSurface` handling when it ran out of context, so no verdict was reached either way.

**How to judge it,** using the same criteria that classified the original finding as real rather than accepted:
- A same-user process forging identity is the documented ceiling and does **not** qualify on its own.
- It qualifies if a normal agent in another workspace can cross the boundary, or a prompt-injected one can cross it without human approval, or sibling surfaces refuse an identical input while this one accepts it.

If it turns out to be unguarded, it belongs in its own change rather than widened into this one — same reasoning that kept this PR to a single asymmetry.

Context for the original finding: it came out of a survey of all 164 workspace-scoping decisions in the codebase, and was verified against source before being acted on.

openwong2kim:
## Live dogfood — fail-closed page selection

Ran a dev build of this branch (`WMUX_DATA_SUFFIX=-dev`). App and daemon boot
clean, browser surfaces open, and the refusal path carries the code and reason
as described. Verified against the source the running renderer serves:

```
PASS  surfaceId is read off browser.open's reply
PASS  the opened surfaceId is pinned directly
PASS  a re-derive refusal no longer aborts the retry
PASS  the 'unscoped' selection variant is gone from the union
```

Removing `unscoped` from the union rather than guarding it is the right shape —
it makes the two workspace-blind selection sites unreachable by construction
instead of unreached by discipline.

### One finding, fixed on this branch

The **one legacy case this PR promises stays functional did not work end to
end** — the panel opened and its page could never be returned.

Against a main too old to tag targets, an empty target list is unambiguous, so
`resolveCallerSurface()` answers `none` and `getPage()` auto-opens the caller's
own surface. It then **re-derived** the surface to pin it. But the target it had
just created carries no `workspaceId` either, so the re-derivation found nothing
tagged and threw `WORKSPACE_SCOPE_UNRESOLVED`. That throw landed in the
auto-open `catch`, `callerHasNoSurface` stayed true, Strategies 1–3 stayed
skipped, and `getPage()` returned `null` after its retries — then latched
`playwrightFailed` for 10s.

So the sentence in the description — *"a legacy main reporting zero live targets
is unambiguous… the caller auto-opens its own surface"* — described an intent the
code did not reach.

`browser.open` already replies with the `surfaceId` it created. Taking it from
there needs no tagging and no proof: we asked for that surface. The
re-derivation stays only for a main whose reply does not name one, and a refusal
there is now caught rather than aborting the loop.

Covered by a `getPage()` test that runs the full loop against an untagging main.
I mutated the fix out and watched it go red before keeping it.

### What I could not exercise live

The legacy path itself. Reproducing it for real needs an *actual* older main
build to connect to, which is also why the test drives a fake one. The
modern-main paths (scoped targets, explicit `surfaceId`, refusal on unresolved
identity) all ran against the real app.

126/126 in `src/mcp/playwright` green; typecheck clean. The two ESLint
complaints in `PlaywrightEngine.ts` (`BrowserContext` unused, empty constructor)
predate this branch.

Issue #680, opened 2026-07-28
Panes spawn without TERM/COLORTERM/TERM_PROGRAM — Node CLIs start with colors disabled


moqing123456:
## Problem

wmux panes spawn with **no `TERM`, no `COLORTERM`, and no `WT_SESSION`/`TERM_PROGRAM`** in their environment. Every Node-based CLI that does terminal capability detection (supports-color and friends) therefore concludes the terminal has no color support and starts with colors disabled.

Measured inside a fresh wmux pane (Windows, PowerShell):

```
echo $env:TERM        # (empty)
echo $env:COLORTERM   # (empty)
node -e "console.log(process.stdout.isTTY, process.stdout.hasColors())"
# true false
```

The same CLI in Windows Terminal gets colors for free because WT sets `WT_SESSION`; VS Code's terminal injects `TERM_PROGRAM=vscode` plus `COLORTERM`. Since wmux's xterm.js front end fully supports 256 colors and truecolor, the missing advertisement is purely a spawn-env gap.

User-side workaround is to set `TERM=xterm-256color` / `COLORTERM=truecolor` manually, but every wmux user on Windows currently hits this out of the box (observed with Kimi Code CLI; affects any supports-color-based tool).

## Proposal

Inject at PTY spawn (both `PTYManager` and daemon `DaemonSessionManager` paths, which already share `envFilter.ts`):

- `TERM=xterm-256color` (xterm.js's advertised capability set)
- `COLORTERM=truecolor`
- optionally `TERM_PROGRAM=wmux` + `TERM_PROGRAM_VERSION` — the conventional identity pair other Electron terminals set (VS Code sets `TERM_PROGRAM=vscode`), which also helps apps distinguish where they run

`envFilter.ts` already whitelists `COLORTERM` in `SAFE_PASSTHROUGH`, so the filter won't fight the injection. Only set values the user hasn't already defined, so a user override still wins.

Split out from #678's side note per the triage suggestion.

openwong2kim:
## [wmux-hermes] Triage Summary

[...]

PR #681, opened 2026-07-28
fix(terminal): clip xterm overflow so IME composition can't drag the pane (#678)

### LINKED ISSUE #678: TUI redraw desync: whole pane shifts while typing in Kimi Code CLI input box, snaps back on submit
## Symptom

In a wmux pane running **Kimi Code CLI** (an interactive TUI agent, similar in nature to Claude Code), the **entire pane content visibly shifts/jitters on every keystroke** while typing a message into the input box — most noticeably as the input box grows taller (long input wrapping to more lines). After pressing Enter to submit, the whole layout **snaps back** to the correct position.

The content is not corrupted or overlapping (unlike #318) — it is displaced as a whole during input growth and self-corrects on submit. It looks like a cursor-positioning desync between the TUI's redraw (cursor-up + repaint as the input box height changes) and xterm's viewport, specific to the input phase rather than the streaming phase.

## Environment

- wmux **3.37.2** (Windows, Squirrel install)
- Windows 11 22H2 (build 22621)
- Pane shell: Windows PowerShell 5.1 over ConPTY
- Kimi Code CLI **0.29.2** on Node v24.17.0
- `TERM=xterm-256color`, `COLORTERM=truecolor` (set manually — see note below)
- `process.stdout.isTTY` = `true`, `hasColors()` = `true` in the pane

## Repro

1. Open a wmux pane (PowerShell).
2. Run `kimi`.
3. Type a long message into the input box (enough to wrap / add lines).
4. Observe: the whole window content shifts around as the box grows, per keystroke.
5. Press Enter — layout immediately snaps back to normal.

## Notes

- Possibly the same repaint-scheduler family as #318 / #333 / #525, but the trigger here is **interactive input-box growth**, not agent streaming output, and there is no stuck or overlapping output — just whole-screen displacement that resets on submit.
- Kimi Code is not on wmux's tested-agent list (Claude Code / Codex / Gemini / Aider / OpenCode / Copilot CLI), so this may be a redraw pattern those agents don't emit. Happy to capture a screen recording, a raw PTY trace, or test a nightly build if that helps.

## Side note (separate minor issue)

wmux panes spawn with no `TERM` / `COLORTERM` / `WT_SESSION` at all, so every Node-based CLI that does capability detection (supports-color) starts with colors disabled — `hasColors()` returns `false`. Setting `TERM=xterm-256color` and `COLORTERM=truecolor` user-side fixes it, but injecting these by default at pane spawn (like VS Code injects `TERM_PROGRAM`) would make colored TUIs work out of the box.

moqing123456:
## Summary

Fixes the horizontal tearing a pane suffers during CJK IME composition with TUIs that park the buffer cursor at the right edge (reported with Kimi Code CLI): xterm anchors its hidden IME helper textarea at that cursor, the preedit-holding textarea overflows the screen's right bound, and Chromium's scroll-into-view on the IME caret drags every `overflow:hidden` ancestor sideways. The fix is a one-rule CSS invariant: `overflow: clip` on `.xterm` and `.xterm-screen` — identical painting to `overflow:hidden`, but no scroll container, so a misplaced textarea is clipped at the edge instead of dragging the pane.

## Changes

- `src/renderer/styles/globals.css` — add the IME composition overflow guard rule (with full RCA comment).
- `src/renderer/__tests__/imeOverflowClip.test.ts` — new source-level regression lock: the clip rule exists, no `.xterm`/`.xterm-screen` rule may use a scroll-container overflow, and the scrollbar/viewport rules stay untouched (vertical scrollback must survive).

### Fixed

- Panes no longer tear horizontally during CJK IME composition with TUIs that leave the buffer cursor at the right edge (e.g. Kimi Code CLI): the IME helper textarea is now clipped at the pane edge instead of Chromium scroll-into-view dragging the whole pane sideways until the composition commits (#678).

## Stability tier impact

n/a — renderer CSS only.

## Root cause (for the reviewer)

CDP probe against a dev instance (`WMUX_DATA_SUFFIX`-isolated, real Kimi Code session, synthetic composition via `Input.imeSetComposition`):

- Kimi Code's input row ends with a right-aligned hint/border, so after its redraw the **buffer cursor sits at the last column** (unlike Claude Code, which restores the cursor to the text caret — which is why Claude Code panes don't tear).
- xterm's CompositionHelper positions the hidden textarea at `left: 1784px` (the right edge); holding the preedit it extends past the screen (`scrollWidth 1833 > clientWidth 1792`).
- Chromium scrolls the pane container (`scrollLeft` 31px → 190px, growing per keystroke) to reveal the IME caret — the whole-pane left shift that snaps back on commit.

`overflow: clip` was first hot-patched into the live instance to confirm the mechanism (scroll drops to 0), then verified as the shipped CSS after a full reload.

## Test plan

- [x] New unit test `imeOverflowClip.test.ts`: 3/3 pass.
- [x] `npx tsc --noEmit` clean.
- [x] Full `npm run test:parallel`: 8739–8742 pass; the only failures (`diff.handler.test.ts`, `cliShim.test.ts` — live-registry/git-worktree tests) **pass standalone** and are the documented Windows parallel-suite flakes, unrelated to this change.
- [x] Live E2E via CDP: `scrollLeft` stays 0 through a 45-char composition with the patch; reporter confirmed visually in the dev build (before/after evidence attached to #678).

## Related issues / PRs

- Closes #678
- Refs #680 (pane-spawn env `TERM`/`COLORTERM` — separate concern, intentionally not in this PR)

## Reviewer checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved IME composition behavior in the terminal by preventing unintended horizontal pane movement and unwanted caret scrolling.
  * Added a regression test to ensure terminal overflow styling consistently uses `overflow: clip` where required and does not introduce conflicting overflow behaviors.


openwong2kim:

This PR fixes the horizontal pane tearing reported in #678 with a minimal, well-reasoned one-rule CSS change: `overflow: clip` on `.xterm` and `.xterm-screen`. The root-cause analysis is sound — xterm anchors its hidden IME helper textarea at the buffer cursor, so a TUI that parks the cursor at the right edge (Kimi Code's input row) lets the preedit-holding textarea overflow the screen, and Chromium's scroll-into-view then drags every `overflow:hidden` ancestor sideways on each keystroke. `overflow: clip` paints identically to `hidden` but creates no scroll container, so the stray textarea is clipped at the edge instead of dragging the pane, and the accompanying regression test locks the invariant at the source level (since jsdom cannot observe the tearing) while also asserting the scrollbar/viewport rules are left untouched. The scope is tight — renderer CSS and one test file, no runtime or wire changes — and it closes #678 cleanly without touching the separate spawn-env concern tracked in #680.

moqing123456:
@coderabbitai review addressed:

1. **Per-rule clip assertion** — adopted in 4f0d6c8: the first test now asserts *every* targeted `.xterm`/`.xterm-screen` rule block individually declares `overflow: clip`, instead of checking the aggregate text. Thanks, that's strictly better.

2. **Positive overflow-y assertion for the scrollbar/viewport** — not adopted, deliberately: `globals.css` declares no `overflow-y` for `.xterm-viewport` or the Monaco scrollable element — the vertical scrolling behavior comes from `xterm.css` (node_modules), which this repo's stylesheet intentionally leaves alone. The invariant this test guards is purely negative (the guard must never *clip* the viewport's scroll container), so asserting a positive value would mean vendoring third-party CSS internals into the test and pinning more than the fix owns.

openwong2kim:
Merged as `0320c144`. Thank you — and welcome, this is your first contribution here.

Worth saying what made it easy to take: you did not just report the tearing in #678, you found the mechanism. The chain from "xterm anchors the helper textarea at the buffer cursor" through "a TUI that parks the cursor at the last column pushes it past the screen bound" to "Chromium's scroll-into-view then drags every `overflow:hidden` ancestor" is the actual root cause, and `overflow: clip` is the right lever precisely because it paints like `hidden` while creating no scroll container. One rule, at the layer where the problem starts.

The source-level regression lock is the part I appreciated most. You could have skipped the test — jsdom genuinely cannot observe this — and instead you locked the invariant where it lives, per-rule rather than in aggregate, with the scrollbar/viewport rules explicitly asserted untouched. That is the harder and more useful choice.

## One thing to watch when you next run a build

`.xterm` and `.xterm-screen` carry no `overflow` in xterm's own stylesheet, so this introduces clipping where there was none rather than swapping `hidden` for `clip`. Two things overflow those boxes today:

- `.xterm-helper-textarea` — parked at `left: -9999em` with `opacity: 0`, so clipping it changes nothing visible.
- `.composition-view` — the black preedit box xterm paints during IME composition. It is `position: absolute` at the cursor with `white-space: nowrap`, so with the cursor at the last column a long preedit used to spill past the pane edge and will now be cut at it.

Clipping is the more correct behavior — spilling over a neighbouring pane was never right — but it is a visible change for CJK input, and no one has yet looked at it in a real build. If you would check it on your Windows + Kimi setup once this ships (current release is 3.37.2, so the next tag), two things would close it out: that the original tearing is actually gone, and that a Korean/Chinese preedit at the right edge of a pane is still readable rather than truncated to nothing.

Your #685 is in the queue and will get the same look.

PR #683, opened 2026-07-28
feat(channels): give channels a trash, and bind missions to their workspace

openwong2kim:
## Summary

Fan-out made the sidebar worse the more you used it. Five completed missions sat permanently expanded in MISSIONS with no way to dismiss or collapse them, and eight mission channels held their slots in the channel list. `task.mission.close` archives the channel, but archiving only makes it read-only — the row stays. There was no way to remove one, and no way to fold them away. README sells fan-out on the promise that "leftovers land in a cleanup list, never as mystery folders"; worktrees had that list, missions and channels did not.

**Channel trash.** A new `trashedAt`/`trashedBy` marker layered on top of `status: 'archived'` — deliberately *not* a third `ChannelStatus` value. Every read-only / join / comment gate in the tree keys on `status === 'archived'`, so a trashed channel inherits all of them unchanged and an older build sees a correct (merely un-hidden) archived channel. A third enum value would turn every `!== 'archived'` check into a permission hole on a durable, sender-verified record. Three replay kinds (`trash`, `restore`, `destroy`), all idempotent per the applier invariant.

`destroy` refuses any channel not already in the trash (`CHANNEL_NOT_TRASHED`). That precondition is the safety property: there is no path from active or archived straight to gone, so permanent deletion always has a separate, reversible step in front of it.

**Authz.** Lifecycle ops gate on member OR CEO OR the local operator. Archive's member-or-CEO gate cannot express this case: mission channels are private agent rooms the human only *observes* (W1), so under that gate the operator could not clear a single one of the rooms cluttering their own sidebar. Sound for the same reason `operatorJoin`/`operatorList` are — these ride the renderer-only `channels:mutate-local` IPC and are absent from the pipe router, so no agent/MCP caller can hide or destroy a channel.

**Retention.** `config.json` gains a `channels` section with two knobs defaulted apart on purpose:

| Knob | Default | Why |
|---|---|---|
| `trashTtlHours` | `720` (30d) | Ships **on**. Nothing reaches the trash except an explicit human action, so the sweep only ever finishes a deletion the operator started. |
| `autoTrashArchivedHours` | `0` (off) | Ships **off**. The only knob that would discard records nobody chose to discard. Even on, it only *moves* to the trash, so the full undo window still applies. |

The sweep runs at boot and hourly rather than at `load()` like the existing empty-channel reaper (KTD8), because a load-time-only reaper never fires on a daemon that stays up for weeks. Garbage config values coerce to `0` rather than to the default — a knob the operator visibly set to nonsense must fail toward keeping data.

**Missions bound to workspace lifetime.** A mission row is visible while its workspace exists and disappears with it. The visibility rule reads ONE signal: whether the workspace id is in the session's workspace list. No derived state is consulted — `agentStatus`, `deliveryStatus` and `pane_list.gitBranch` have each reported false in this repo, so "is this mission still live" is answered by existence, not inference. A task whose `paneGroupId` has not materialized yet (fan-out in flight) is kept: it has no workspace yet, which is not the same as having lost one.

Deleting a task workspace now closes its mission, which archives the mission channel into the collapsed group. **The channel survives** — closing a mission archives, it never deletes.

**Sidebar.** Trash group below Archived, collapsed by default, so N leftovers cost one line instead of N. MISSIONS section collapses, and completed-but-still-live missions sit under a collapsed "Done" disclosure.

## Review

Three models reviewed this diff independently with identical instructions and fresh context (Claude subagent, Codex `reasoning=high`, GLM 5.2), then findings were cross-scored by fingerprint. 21 raw findings merged to 17 distinct. Every one was verified by hand before acting.

Consensus findings, all fixed:

- **(2-model, verified)** `task.mission.close` never reached the daemon. It rode `rpc.invoke` → the pipe RpcRouter, where `task.mission.close` is registered `mutating` and the forwarder rejects any mutating call with no resolvable `senderPtyId`. Renderer calls have none, so it failed 100% of the time — and the response was never checked, so it failed silently. Now routed through the renderer-only IPC, with the response checked. A regression test pins the transport.
- **(2-model, verified)** The 7-day empty-channel reaper knew nothing about `trashedAt`. A zero-member trashed channel (routine for mission channels after workspace deletion → `purgeMembership`) was pruned at 7 days, making the promised 30-day undo window nominal. Trashed channels are now exempt; only the retention sweep deletes them.
- **(2-model, verified)** `destroy` emitted its catalog event *after* removing the row, so the ws-human recipient injection was skipped and a zero-member channel notified nobody — a permanent ghost row. Recipients are now computed before the commit, for all three ops.
- **(3-model, verified)** `destroy` cleared the persisted idempotency map but not the runtime LRU, leaking up to 1000 entries per destroyed channel for the daemon's lifetime.
- **(2-model)** `destroy` is reachable from a direct daemon-pipe client — the same documented `#113`/F1 same-user residual as `archive`, but irreversible where archive is not. Accepted and now stated explicitly in the code and in `docs/api/stability.md`.

Solo findings, all fixed: a self-inflicted regression where deleting a fan-out *parent* wiped the mission cache and hid live *child* missions; a one-click irreversible trash on live channel rows (now the same two-step armed confirm `ChannelView` already uses for archive); ghost unread badges on trashed channels; a silent sweep that swallowed `PERSIST_FAILED`; the sweep destroying a channel an open mission still anchors; sub-hour retention values flooring to "never"; and four smaller correctness items.

## Test Coverage

Deletion paths are where data is lost, so the three properties that matter are pinned directly: a trashed channel restores with its history intact; permanent deletion hits only its target and cannot skip the trash; and **a channel survives its workspace being deleted** — the core of the policy — asserted on the channel row, its messages, and its members after the close that workspace removal triggers.

16 mutations were run against the deletion path. Two survived initially and both were real gaps:

- Making the sweep purge any aged channel (falling back to `createdAt`) failed no test — untrashed channels were saved by `destroy`'s precondition, so asserting on the outcome could not distinguish "selected correctly" from "selected everything and was refused". The sweep's *selection* is now pinned by spying on the call, independent of the guard behind it.
- Dropping the `status === 'archived'` check from the auto-trash pass, which would hide channels people are still posting in. The test now includes a live channel of the same age.

All 16 now fail against their mutation. The other fourteen covered the trash gate, the appliers, the authz gate, the idempotency guard, the mission visibility rule, and the renderer thunks.

## Test plan

- [x] `npx tsc --noEmit` clean on all six tsconfigs
- [x] `npm run test:parallel` — 8805 passed, 23 skipped (619 files)
- [x] `npm run test:runtime` — 88 passed, 11 skipped
- [x] `npx eslint` on every touched file — zero new errors or warnings against the pre-change baseline
- [x] `node scripts/gen-api-reference.mjs` regenerated (144 → 147 methods), no drift after merging `main`
- [x] Suites re-run against `main` merged in, not just the branch tip

## Not in scope

- Chat View (#655) is being built on another branch. No surface-tab or Chat file is touched here.
- Message durability, sender verification, and inbox semantics are untouched.
- No version bump — `package.json` stays at the last released version, CHANGELOG entries land under `[Unreleased]`.

## Summary by CodeRabbit

* **New Features**
  * Added channel Trash support, including restore and permanent deletion after a confirmation step.
  * Added configurable retention periods and optional automatic archiving-to-Trash cleanup.
  * Protected channels linked to open missions from automatic deletion.
  * Made the Channels Trash group collapsible and added related notifications.
  * Improved mission sidebar organization with collapsible active and completed sections.

* **Bug Fixes**
  * Removed missions tied to deleted workspaces from the live mission list while preserving their underlying channel history.
  * Prevented trashed channels from being removed by unrelated empty-channel cleanup.


openwong2kim:
@codex review

openwong2kim:

This PR lands two related sidebar-hygiene improvements: a reversible channel trash lifecycle (soft-delete → restore → permanent destroy) with a configurable 30-day retention sweep, and a binding of mission rows to the lifetime of their fan-out workspace so finished missions stop occupying the sidebar permanently. It touches 28 files across the daemon (`ChannelService` trash/restore/destroy + `sweepRetention`, config, RPC registration, method-capability map), the renderer (`channelsSlice`, `ChannelsPanel` Trash UI, `MissionsSection`, `workTaskSlice`/`workspaceSlice` cleanup-on-removal), and shared protocol/constants, and adds regression tests for the lifecycle transitions, retention selection, mission close-on-workspace-removal, and the panel UI.

The design choices look sound and are worth calling out during review: the destructive surface is **humans-only at the transport layer** (the three lifecycle RPCs ride the renderer-only `channels:mutate-local` IPC and are deliberately absent from the pipe router, matching archive/kick), permanent deletion is unreachable without going through the trash first, and the retention knobs are oriented to fail toward keeping data — `autoTrashArchivedHours` ships off precisely because it is the only one that would discard records nobody explicitly chose to. Two follow-ups worth confirming before merge: the `sweepRetention` loop should re-validate each candidates `trashedAt` under the lock immediately before calling `destroy` (a flagged concern, since the expired list is computed before the lock is taken), and the live `this.idempotency` Map should be dropped on destroy so it does not leak per-channel LRU entries on a long-running daemon.

**Priority: P2** — this directly fixes the "fan-out makes the sidebar unusable" pain the owner hit dogfooding, so it should land soon, but it is not blocking a release or a live defect. Tests, CHANGELOG, stability docs, and API reference are all updated; ready for review once the sweep re-validation and idempotency-cleanup nits are resolved.

PR #684, opened 2026-07-28
feat(sidebar): give channels a direct entry point in the footer

openwong2kim:
## What

Adds a **Channels** button to the sidebar footer. The cluster is now Agent / Git / Channels / web.

## Why

Channels had no button of their own. The Agent button opens the deck on whichever tab it was last on, so reaching channels meant opening the deck and then hunting for the tab. Worse, `channelsTabVisible` defaults to **off**, so for a user who had never been into Settings there was no path to channels at all — the surface that shows what the agents are saying to each other was the least reachable one in the window.

## How

The button mirrors the existing Git one: opens the deck on the `channels` tab, closes the deck on a second press. Steel-blue while open, amber with a count while anything is unread (reusing the same `channelUnread` sum the Agent button already shows).

One judgement call worth flagging: pressing the button also sets `channelsTabVisible` to true. Without it the button would open a tab `ChannelDock` refuses to render, which reads as a dead control. Pressing a button labelled Channels *is* the intent that setting gates. The Settings toggle still turns it back off.

## Verification

- `tsc --noEmit` clean
- `vitest run src/renderer/components/Sidebar src/renderer/i18n` — 19 passed
- `eslint` on the touched files — 0 errors (7 pre-existing unused-var warnings in `Sidebar.tsx`, untouched)

No new test was added for the button itself; say the word if you want one before merge.

## Summary by CodeRabbit

* **New Features**
  * Added a **Channels** button to the sidebar footer control cluster.
  * The button toggles the Channels tab/panel on and off (it won’t open an empty view).
  * Channels shows an amber unread-count badge when applicable.
  * Added new English and Korean labels/tooltips for the Channels control.
* **UI Enhancements**
  * Updated the sidebar icon set with a new hash-style icon for the Channels button.
* **Documentation**
  * Updated the unreleased changelog entry and refreshed the task status overview.


openwong2kim:

This PR adds a dedicated **Channels** button to the sidebar footer, making the cluster Agent / Git / Channels / web. The change is small and well-scoped: a new `IconHash` glyph, a toggle button in `Sidebar.tsx` that mirrors the existing Git button (opens the deck on the Channels tab, closes on a second press, steel-blue while active, amber with an unread count), and the matching `en`/`ko` i18n strings. The button also flips `channelsTabVisible` on, since that setting defaults to off and `ChannelDock` refuses to render the tab without it — so without this the button would appear dead on first press.

The motivation is sound: Channels had no direct entry point. The Agent button opens the deck on whatever tab was last active, and for anyone who had never enabled the Channels tab in Settings there was no path to it at all — making the surface that shows inter-agent communication the least reachable one in the window. The implementation respects existing patterns (focus ring, theme tokens, the same `channelUnread` sum the Agent button already shows), and the decision to set `channelsTabVisible` on press is the right call given the alternative is a non-rendering tab.

One thing worth a glance before merge: the `TODOS.md` diff adds ~80 lines of unrelated planning notes (orchestration dogfood, iOS release readiness, workspace-scoping survey follow-ups, `plans/` tracking policy). None of it is tied to the Channels button. If those notes belong in the repo, a separate commit keeps this feature PR's history clean; if they are scratch/planning notes, `TODOS.md` is tracked and will carry them into every diff going forward.

Priority: **P3** — a UX discoverability improvement, low risk, no external surface touched.

PR #685, opened 2026-07-28
feat(terminal): advertise TERM/COLORTERM/TERM_PROGRAM at PTY spawn (#680)

### LINKED ISSUE #680: Panes spawn without TERM/COLORTERM/TERM_PROGRAM — Node CLIs start with colors disabled

moqing123456:
## Summary

Implements #680: advertise terminal capabilities (`TERM=xterm-256color`, `COLORTERM=truecolor`, `TERM_PROGRAM=wmux`) as **defaults** in both `envFilter` spawn-env builders, so capability-probing CLIs (supports-color — most Node TUIs) get colors out of the box in wmux panes. User/system/profile overrides always win.

## Changes

- `src/shared/envFilter.ts` — new `TERMINAL_CAPABILITY_DEFAULTS` + `applyTerminalCapabilityDefaults()` applied in `buildInteractiveShellEnv` and `buildGatedAutomationEnv`, so the PTYManager and DaemonSessionManager spawn paths stay in lockstep (both funnel through these builders via `resolveSpawnEnv` / the daemon `buildSafeChildEnv` fallback). Presence check is case-insensitive because win32 env keys are — a user-set `Term` must suppress the `TERM` default instead of surviving as a nondeterministic duplicate.
- `src/shared/__tests__/envFilter.test.ts` — 5 cases: injection in both builders, base-env override wins, case-variant key blocks the default, filtered-out keys can't be resurrected.
- `src/main/pty/__tests__/resolveSpawnEnv.test.ts` — 4 funnel-level cases: injection reaches the canonical merge point under both `gated` and `passthrough` policies, base-env TERM wins, workspace-profile TERM overrides the default.

`COLORTERM` was already in `SAFE_PASSTHROUGH`, so the credential stripper does not fight the injection.

### Fixed

- Panes now advertise `TERM=xterm-256color`, `COLORTERM=truecolor`, and `TERM_PROGRAM=wmux` at spawn, so Node-based TUIs (Kimi Code CLI and anything else using supports-color) render in color out of the box instead of falling back to monochrome; user-defined values always take precedence (#680).

## Stability tier impact

Env values added to spawned panes only; no RPC/wire/protocol change.

n/a — spawn-env construction only. `resolveSpawnEnv`'s merge order is unchanged (the defaults land inside the baseline step, so overlays and forced identity still apply exactly as before).

## Test plan

- [x] New unit tests: `envFilter.test.ts` (+5) and `resolveSpawnEnv.test.ts` (+4) — 52/52 pass in the two files.
- [x] `npx tsc --noEmit` clean.
- [x] Full `npm run test:parallel` — exit 0.
- [x] Live check on a `WMUX_DATA_SUFFIX`-isolated dev instance (Windows 11): the spawn env chain was traced end-to-end; note that on this Windows build conhost already injects `COLORTERM=truecolor` itself, and a restored daemon session replays its create-time env verbatim (expected behavior) — fresh panes get the new defaults.

## Related issues / PRs

- Closes #680
- Refs #681 (the IME overflow-clip fix — separate concern per triage)

## Reviewer checklist

## Summary by CodeRabbit

* **New Features**
  * Terminal sessions now receive default terminal capability settings, including terminal type, color support, and terminal program information.
  * Existing environment values and workspace profile settings continue to take precedence over defaults.

* **Bug Fixes**
  * Prevented default terminal settings from restoring environment variables that were intentionally removed.
  * Applied consistent terminal capability behavior across interactive shells and gated automation.


openwong2kim:

PR #685 implements issue #680 by injecting `TERM=xterm-256color`, `COLORTERM=truecolor`, and `TERM_PROGRAM=wmux` as defaults into both `envFilter` spawn-env builders (`buildInteractiveShellEnv` and `buildGatedAutomationEnv`), so capability-probing Node CLIs (e.g. Kimi Code, anything using supports-color) render in color out of the box in wmux panes on Windows instead of silently degrading to monochrome. The change is small and well-contained: a new `TERMINAL_CAPABILITY_DEFAULTS` table plus `applyTerminalCapabilityDefaults()` applied inside the two builders, with a case-insensitive presence check so a user- or system-set `Term` is never clobbered on win32. Coverage is solid — 5 new cases in `envFilter.test.ts` (injection in both builders, base-env override wins, case-variant key blocks the default, stripped credentials are not resurrected) and 4 funnel-level cases in `resolveSpawnEnv.test.ts` (injection reaches the canonical merge point under both gated and passthrough policies, base-env and workspace-profile overrides still apply), and the PR notes that `tsc --noEmit` and `npm run test:parallel` are green. One small thing worth confirming on review: the change touches the shared `envFilter` baseline step, so it is worth a quick sanity check that any daemon automation path that intentionally strips `TERM` (headless/CI shells, for example) is not unexpectedly advertising a color-capable terminal to a process that should not believe it has one — the defaults are applied unconditionally in both builders rather than gated on a "this is a TTY-rendered pane" signal.

**Priority: P3** — quality-of-life polish that removes a recurring first-run papercut for Node TUI users on Windows; no protocol, RPC, or external surface change.

moqing123456:
On the triage note about the defaults being applied unconditionally in both builders — considered, and I believe it's safe for two reasons:

1. **Gated spawns are still xterm-rendered panes.** `buildGatedAutomationEnv` feeds agent/automation panes, which in wmux are real, visible terminal surfaces rendered by the same xterm.js front end — the advertisement is accurate for them too. There is no spawn path through these builders whose output is *not* ultimately displayed in an xterm pane.

2. **The env hint alone can't force colors onto a non-TTY consumer.** supports-color (and its analogues in other ecosystems) gate on `stdout.isTTY` first; the `TERM`/`COLORTERM` hints only discriminate *between* TTYs. A process whose stdout is a pipe or file — the genuine headless/CI case — still resolves to no-color regardless of these variables. The only lever that overrides that is `FORCE_COLOR`, which we deliberately do *not* set.

So the failure mode the note worries about (a non-TTY process believing it has a color terminal) would require the consumer to honor `TERM` while ignoring `isTTY`, which is exactly backwards from how the mainstream capability probes work. Happy to gate the injection behind a spawn-kind flag instead if you'd rather have the invariant spelled out in code — but with the current spawn taxonomy the unconditional form is both accurate and the simpler invariant.

openwong2kim:
Merged as `a4348930`. Second one in a day — thank you.

The part that made this reviewable quickly: you put the injection in `envFilter.ts` rather than at either spawn site, so both builders get it and the `PTYManager` / `DaemonSessionManager` paths stay in lockstep the way the rest of that module already insists on. And the case-insensitive presence check is the detail most people would have missed — a user-set `Term` on win32 has to suppress the `TERM` default or both keys survive and the OS picks one nondeterministically. You wrote the test for it too.

One note for the record, since it narrows what this actually changes rather than taking anything away from it.

`node-pty`'s unix path does:

```js
// lib/unixTerminal.js:61
var name = opt.name || env.TERM || DEFAULT_NAME;
env.TERM = name;
```

`opt.name` wins unconditionally, and both spawn sites already pass `name: 'xterm-256color'` (`PTYManager.ts:220`, `DaemonSessionManager.ts:412`). So macOS and Linux panes already had `TERM=xterm-256color`, and the `TERM` default is a no-op there — ConPTY is the one path that ignores `name`, which is exactly why #680 reproduced only on Windows. Outside Windows the real delta is `COLORTERM=truecolor` and `TERM_PROGRAM=wmux`, and `COLORTERM` alone is what Node's `getColorDepth()` keys on, so the reported symptom is covered either way.

The other thing worth knowing about: `TERM_PROGRAM` is a default, not a force. A dev build started from iTerm inherits `TERM_PROGRAM=iTerm.app` into Electron's env, the filter is a blocklist so it passes straight through, and the pane then advertises itself as iTerm. That is today's behavior and your change does not make it worse — but it does mean `TERM_PROGRAM` is not yet a reliable "this is wmux" signal for tools that branch on it. Forcing that one key while leaving `TERM`/`COLORTERM` as defaults would close it. Not something I want folded in after the fact; raising it in case you want to pick it up.

Same verification ask as #681, whenever the next tag ships: `echo $env:TERM` / `$env:COLORTERM` in a fresh pane, and `node -e "console.log(process.stdout.hasColors())"` reading `true` this time.

moqing123456:
Picked up the `TERM_PROGRAM` follow-up in #696 — forced to `wmux` (with case-variant cleanup, same nondeterminism guard as the presence check), while `TERM`/`COLORTERM` stay user-overridable defaults. One deviation worth flagging: a workspace profile overlay still applies after the baseline, so a deliberately configured profile `TERM_PROGRAM` can still override the forced value — that seemed in the spirit of the existing override chain (profile `TERM` already beats the injected default), but easy to tighten if you'd rather have the identity win there too.

PR #687, opened 2026-07-28
chore(changelog): assemble release notes from one file per PR

openwong2kim:
## The problem, measured

Every PR adds its entry at the same line of `CHANGELOG.md` — directly under `### Added` inside `[Unreleased]`. Git sees two different insertions at one position and cannot order them, so it reports a conflict. That is not a mistake anyone is making; it is what a shared insertion point means.

Today it cost real time. Four merges left most of the open queue `DIRTY`, and merging one further docs-only PR (#669) re-dirtied five branches that had just been fixed by hand. Two of those had green check marks displayed from a base they were no longer on, because **a conflicted PR silently stops queueing CI** — so the queue looked healthier than it was.

## The change

A PR no longer edits `CHANGELOG.md`. It adds `changelog.d/<pr-number>.md` with plain Keep a Changelog headings. Separate files cannot collide, so the conflict is designed out rather than resolved over and over.

`node scripts/collect-changelog.mjs` folds the fragments into `CHANGELOG.md` at release time and deletes them. Ordering is by PR number, so the same fragments always produce the same file. Entries already written directly under `[Unreleased]` are kept and new ones appended after them — this can be adopted without moving anything that is there today (the eleven entries currently sitting in `[Unreleased]` stay put).

`--check` reports what would be folded without writing.

## Refusals

The fold deletes the fragments it consumed, so anything it skipped silently would be unrecoverable. Two inputs are therefore errors, not warnings:

- a heading that is not one of the six Keep a Changelog sections (a typo like `### Improvements`)
- prose sitting outside any heading

## Verification

- 9 unit tests covering multi-line entries, section ordering, numeric-not-lexical file ordering (`9.md` before `100.md`), preservation of released sections, and both refusals
- mutation-checked: replacing the numeric sort with a string sort turns the ordering test red
- dry-run against the real `CHANGELOG.md` with this PR's own fragment — `[Unreleased]` keeps its existing Added/Changed/Fixed sections and the fragment lands under Changed

`CLAUDE.md`'s release procedure is updated to run the fold before the version bump. No version bump here.

## Note on the other open branches

This does not retroactively fix them — the five currently `DIRTY` branches still each need their `CHANGELOG.md` conflict resolved once. It stops the next merge from creating a sixth round.

## Summary by CodeRabbit

* **New Features**
  * Added automated tooling to collect release-note fragments, organize entries by pull request number, and merge them into the unreleased changelog.
  * Added a check mode to preview available release-note fragments without modifying files.

* **Documentation**
  * Clarified the release workflow and required formatting for changelog entries.
  * Documented supported changelog categories and fragment conventions.

* **Tests**
  * Added coverage for parsing, ordering, validation, merging, and release-section handling.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -38,9 +38,14 @@ In QA/design-review mode, flag any code that doesn't match DESIGN.md.
 - **PRs never bump the version.** `package.json` stays at the last released
   version on every feature branch. Do NOT let /ship (or any workflow) bump
   MAJOR/MINOR/PATCH, claim version slots, or prefix PR titles with `vX.Y.Z`.
-- CHANGELOG: each PR adds its user-facing entries under `## [Unreleased]`
-  at the top (Keep a Changelog). Merge conflicts there are append-merges.
-- **Release = explicit user action**: bump `package.json` version, rename
+- CHANGELOG: a PR does **not** edit `CHANGELOG.md`. It adds one fragment,
+  `changelog.d/<pr-number>.md`, holding its user-facing entries under plain
+  Keep a Changelog headings. Separate files cannot conflict — editing the one
+  shared insertion point meant every merge left every other open PR dirty.
+  See `changelog.d/README.md`.
+- **Release = explicit user action**: run
+  `node scripts/collect-changelog.mjs` to fold the fragments into
+  `CHANGELOG.md`, bump `package.json` version, rename
   `[Unreleased]` → `[X.Y.Z] — YYYY-MM-DD`, run
   `node scripts/gen-api-reference.mjs` (the generated header bakes the
   version — the CI drift guard enforces this), commit `chore(release)`,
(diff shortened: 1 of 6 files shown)


openwong2kim:

This issue proposes replacing direct `CHANGELOG.md` edits with one fragment file per PR (`changelog.d/<pr-number>.md`) plus a `scripts/collect-changelog.mjs` script that folds them at release time. The motivation is concrete and measured: every PR currently inserts at the same line under `[Unreleased]`, so Git cannot order two simultaneous edits and reports a conflict — a structural problem that has already dirtied five branches and masked a stale CI queue. The fragment approach designs the conflict out entirely, and the `--check` dry-run plus strict refusal of malformed fragments make the fold safe to run unattended.

The proposal is well-specified: ordering by numeric PR number (not lexical), preservation of existing `[Unreleased]` entries, and two hard-error refusals (unknown headings, prose outside any heading) that prevent silent data loss during the fold. This is a developer-workflow and CI-hygiene improvement rather than a product change — low individual severity, but it removes a recurring friction point that costs real merge time and can make the queue look healthier than it is.

Labeling **P3**. It is worth doing (the conflict loop is actively painful during release windows), but it does not block any feature or fix and can land whenever the tooling work is ready.

PR #688, opened 2026-07-28
fix: release DECSTBM scroll region leaked by dead-TUI replay

### LINKED ISSUE #318: [Bug]: Terminal render corruption: old Claude output overlays new output (messages appear cut/frozen)
### Description of the bug
In wmux terminal panes, Claude Code output gets visually corrupted during long conversations:
- old output segments stay “stuck” on screen
- new messages are partially rendered or clipped
- lines look overlaid/mixed with previous content
This makes full responses unreadable.

### Environment
- OS: Windows (please fill exact version)
- wmux version: (fill from app/about or commit)
- Shell: (PowerShell/CMD/Git Bash/etc.)
- Claude Code version: (fill exact version)
- GPU acceleration: I tested (On and Off) and still the issue happen

### Frequency
(Always)

### Workarounds tried
- Clearing terminal/scrollback
- Restarting pane/session
- Toggling hardware acceleration
- Avoiding pane resize during stream
- Move the scrollback to several limit 500 - 1000 - 10000

Looks related to terminal render/reflow/viewport sync under streaming output in long-running AI sessions.

### Reproduction steps
1. Open wmux
2. Start Claude Code in a pane
3. Ask for long multi-paragraph responses repeatedly
4. Resize/split/focus panes while streaming output
5. Observe stale old segments overlaying new text; response appears truncated/frozen

### Expected behavior
Terminal should repaint cleanly. New output should fully replace/append correctly with no stale overlays.

### Actual behavior
Old lines remain on screen and overlap with new output; latest response is incomplete/garbled visually.

### wmux version

3.11.1

### How did you install wmux?

`setup.exe` installer 

PowerShell 7+ (pwsh)

### Windows version

windows 10 and 11

None / not relevant

```shell

```

### Confirmation

AdgDean:
## Summary

A TUI agent killed mid-run can leave a DECSTBM scroll region armed (`ESC[<top>;<bottom>r`) with no matching `ESC[r` to release it. When the daemon's ring buffer is replayed into a fresh `xterm` instance on reconnect, that sequence re-executes — confining all future output, and scrollback navigation, to the narrow region while everything outside it stays frozen on screen. Scrolling up doesn't reach real history; it reveals whatever was frozen there when the region was armed.

This adds `STALE_REPLAY_DISPLAY_RESETS` (`ESC[r` + `ESC[?6l`) alongside the existing `STALE_REPLAY_INPUT_MODE_RESETS`, at both replay call sites in `useTerminal.ts` (`paintDeadSnapshot` and `resetStaleReplayModes`). `ESC[r` resets the scroll region to full-screen without erasing any cell, so — unlike the `?1049`/`?25` display state intentionally left alone by the existing reset — it cannot alter restored scrollback content; it only un-narrows where future writes are allowed to scroll.

The codebase already tracks this exact class of leaked state via `MarginTracker` in `ansiStreamScan.ts`, and `HeadlessSnapshot.ts` already documents that "DECSTBM scroll margins are not exposed by the public API" — this closes the last gap: the margin was tracked but never reset on the replay path.

## Related

Possibly related to #318 ("Terminal render corruption: old Claude output overlays new output"). That issue was closed by switching the affected pane's TUI mode from `fullscreen` to `default` as a workaround — which sidesteps this bug rather than fixing it, since it's `fullscreen` TUIs that tend to arm a DECSTBM scroll region in the first place. The final symptom reported there ("cannot scroll back to see everything already discussed", "only what is currently being responded stays visible") matches the scroll-region leak this PR fixes closely enough that I want to flag it, but I can't confirm 100% it's the same root cause without re-testing in `fullscreen` mode against this fix. Not using "Fixes #318" so it won't auto-close — @openwong2kim, if useful I'm happy to re-verify in `fullscreen` mode once this lands.

## Test plan

- [x] New behavioral tests against a real headless `xterm` instance (`STALE_REPLAY_DISPLAY_RESETS` describe block): arms the exact region observed in reproduction (`ESC[5;20r`), confirms it releases back to full-screen, confirms no cell is erased, confirms no response bytes are provoked.
- [x] Source-level regression lock: both `paintDeadSnapshot` and `resetStaleReplayModes` call sites are pinned to apply `STALE_REPLAY_DISPLAY_RESETS` alongside `STALE_REPLAY_INPUT_MODE_RESETS`.
- [x] `npx vitest run src/renderer/terminal/__tests__/staleReplayModeReset.test.ts` — 9/9 passed.
- [x] Verified live against a dev build: reconnecting to a pane where a TUI died mid-scroll-region no longer traps output in a narrow window; scrollback is fully reachable again.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed terminal recovery after mid-run reconnects to ensure stale-replay display state is fully cleared, not just input-mode state.
  * Un-pins narrowed “frozen” scroll regions so future output scrolls across the full display without erasing restored content.
  * Preserves the restored cursor position and avoids emitting unintended response data during reset.
* **Tests**
  * Expanded coverage to validate display-state reset behavior, including scroll-region clearing and no data emitted.


openwong2kim:

This PR fixes a correctness bug in the stale-replay path: when a TUI agent is killed mid-run with a DECSTBM scroll region armed (`ESC[<top>;<bottom>r`) and no matching `ESC[r`, the daemon ring buffer replays that sequence into the fresh xterm instance on reconnect — confining all subsequent output to the narrow region while everything outside it stays frozen on screen, and scrolling up no longer reaches real history. The fix adds `STALE_REPLAY_DISPLAY_RESETS` (`ESC[r` + `ESC[?6l`) alongside the existing `STALE_REPLAY_INPUT_MODE_RESETS` at both call sites in `useTerminal.ts` (`paintDeadSnapshot` and `resetStaleReplayModes`), releasing the scroll region back to full-screen without erasing any restored cell. It is a minimal, surgical change — two write sites plus a new export — and the new behavioral tests against a real headless xterm (armed region → released, no cell erased, no response bytes provoked) and source-level regression locks at both call sites give it solid coverage. Worth landing because this is the normal reconnect path after a TUI agent death, the codebase already tracks this exact class of leaked state via `MarginTracker` but never reset it on replay, and the fix closes that last gap without touching restored scrollback content. Flagged as possibly related to #318 but intentionally not auto-closing it.

openwong2kim:
Merged as `62f64ec8`. Thank you — and welcome, this is your first contribution here.

I checked the three load-bearing claims against this repo's xterm (4.1.0) rather than taking them on trust, because the whole fix rests on them. All three hold:

- A bare `ESC[r` does snap the cursor to `(0,0)`, so the DECSC/DECRC bracket is doing real work rather than being defensive noise.
- The full sequence releases `ESC[5;20r` back to `0..rows-1`, keeps the cursor at `(9,14)`, and leaves the SGR attributes untouched.
- `ESC[?6l` survives the `ESC 8` — which is the one I expected to catch you out, since the VT spec has DECSC save origin mode along with the cursor. xterm.js does not implement that part, so the `?6l` sticks.

That last one is worth writing down somewhere other than a PR comment: the reset leans on an xterm.js deviation from the spec. If xterm ever makes DECSC spec-correct, `ESC 8` would restore origin mode and undo the `?6l`. Harmless when it happens — the margins are full-screen by then, so DECOM-relative and absolute addressing agree — but the reasoning in the docblock would quietly stop being true.

The part I liked: you scoped the reset to exactly what cannot alter restored content. `ESC[r` erases no cell, which is what makes it safe to apply unconditionally, and you said so explicitly rather than leaving the next reader to work out why `?1049`/`?25` are deliberately left alone while this one is not. Same for applying it at both call sites and pinning that with a source-level lock.

**One thing I added before merging**: `changelog.d/688.md`, pushed to your branch as `73480204`. This PR opened at 21:30 and the one-fragment-per-PR arrangement landed at 19:39 the same evening, so it was live but easy to miss — see `changelog.d/README.md`. Nothing for you to do; flagging it so the next one starts with a fragment.

On #318: yes please, and thank you for not writing `Fixes` on a hunch. It was closed by moving the pane from `fullscreen` to `default`, which sidesteps the class of TUI that arms a scroll region in the first place, so your read that the two are connected is plausible. If you re-test in `fullscreen` against this fix once the next tag ships and it holds up, say so on #318 and we will close it properly rather than leaving a workaround standing in for a fix.

PR #692, opened 2026-07-28
fix(channels): let the daemon prove a pane pin itself, on every path

openwong2kim:
Follow-up to #675. Two holes, one root: the daemon was asking someone else whether a pane was alive, and only asking on some paths.

## The check skipped the path that matters

The pin gate's liveness half sat inside `else if (panePtyId.length === 0)`. The local composer attaches a `ptyId` to every mention it sends, so that branch never ran for a human's mention.

A person mentions a pane. The agent in it exits. The stale pty rides through, the receiving renderer's fail-closed match misses, `mentionPaneId` is still stamped, and the degraded workspace-level paste lands in whatever agent that workspace still has. **An instruction addressed to a departed worker starts a sibling's turn**, and the sender is told nothing — the failure #675 refused for MCP, still live on the path that produces most mentions.

The caller-supplied pty is now ignored outright. A mention leaves the gate carrying the pty the daemon proved, or carrying no pin at all. A sender does not get to pick which pty the receiver matches against.

## Proving it meant asking a different authority

Liveness came from the principal registry's `liveness` field. Only a renderer can refresh that field, and the daemon backfills every pane-agent to `stale` on its own restart. The two paths back to `live` are a reconnect re-seed that filters to panes the renderer does *not* already know, and an activity broadcast that requires the pane to emit output.

An **idle** agent hits neither. So after any daemon restart, a `pane_id` mention aimed at an idle agent was refused `pane_not_live` — permanently, until that agent happened to print something. That is the exact case pane-pinned mentions were built for, and it shipped in #675.

The daemon owns the PTYs. It answers from its own session table now, and the registry is what it should have been all along: the ownership proof and the pty coordinate, nothing more. No renderer changes, and the documented "only a renderer re-registration brings a pane back to live" invariant is left intact rather than drilled through.

`suspended` is not live. The shutdown-kill classifier demotes an involuntarily-killed session to it and `attachSession` rejects it because no ptyProcess remains. Live is `attached` or `detached` — the same pair `listLiveSessions()` filters on.

## The rules are now testable

`resolvePanePin` is pure: both lookups arrive as functions. The gate used to be an anonymous closure inside `daemon/index.ts`, so a test could only inject a fake and the production pairing of registry and session table was never exercised. The daemon-restart regression — stale principal, live session, still delivers — could not have been written against the old shape.

## An absent resolver refuses, rather than passes

The dependency is optional in the type so channel unit tests need no registry fixture, and that used to mean an absent resolver let every pin through unchecked, with a test cementing it. Nothing had proven the pane belonged to that workspace, which is the one thing the gate establishes. A gate that opens when its dependency is missing is decorative, so absence is now treated as a failed lookup.

Two fixtures relied on the old behaviour. Rather than weaken the gate, they inject a resolver and go on testing the dedup and memberId mapping they were written for.

## What this does not fix

Stated in the contract instead of left to be found:

- **The window is narrower, not closed.** The daemon checks at post time, the renderer re-checks at delivery. A pane that dies between the two still meets the degraded workspace-level fallback.
- **Live means a live PTY, not a live agent.** An agent can exit back to its shell while its session stays attached; such a pin is accepted and the text lands at a shell prompt. Nothing in the delivery path models agent-exit today, so `pane_not_live` does not promise more than it can check.

The renderer's 2b degraded delivery is deliberately untouched — it is an intentional, tested decision, and reversing it is a separate call.

## Tests

`resolvePanePin` covers ownership (missing pane, other workspace's pane, non-pane principal) and liveness (`attached`/`detached` prove; `suspended`/`dead`/no-session/no-ptyId refuse), plus the daemon-restart pair: a stale principal with a live session still delivers, and a stale principal whose session is genuinely gone is not resurrected.

At the service level: a composer-shaped mention (paneId **and** ptyId) with a dead pane is refused rather than routed, a caller-supplied ptyId is overwritten with the proven one, and an absent resolver refuses. Existing pin, dedup, and 2b coverage stays green.


openwong2kim:

This PR fixes two correctness defects in the channel pane-pin gate that shipped in #675. First, the liveness check was gated behind `panePtyId.length === 0`, but the local composer attaches a ptyId to every mention it sends — so for any mention typed by a person the check never ran, and a stale pty from a dead agent could ride through and cause the degraded workspace-level paste to start a sibling agent's turn. Second, liveness was read from the principal registry's `liveness` field, which the daemon backfills to `stale` on restart and only a renderer can restore — so an idle agent (the exact case pane-pinned mentions exist to serve) was permanently refused after any daemon restart. The fix moves liveness to the daemon's own session table, extracts a pure `resolvePanePin` so the real rules are testable without a daemon, ignores caller-supplied ptyIds entirely, and treats an absent resolver as a refusal rather than a pass-through. The test coverage is strong: ownership, liveness per session state, the daemon-restart regression pair, and the composer-shaped mention that exposed the original gap. Assigning P2 — this closes a real cross-agent delivery bug and a post-restart usability regression in a recently shipped feature, and the diff is clean and well-scoped.

PR #696, opened 2026-07-29
feat(terminal): force TERM_PROGRAM=wmux instead of default-injecting it

moqing123456:
## Summary

Follow-up to #685, picking up the point raised in [this review note](https://github.com/openwong2kim/wmux/pull/685#issuecomment-5111799328): `TERM_PROGRAM` is now **forced** to `wmux` instead of default-injected.

- `TERM` / `COLORTERM` remain user-overridable defaults (a user/system-set value still wins, case-insensitively on win32) — unchanged.
- `TERM_PROGRAM` is now always set to `wmux`. The filter is a blocklist, so a dev build launched from another terminal inherits e.g. `TERM_PROGRAM=iTerm.app` straight through; default-only injection would leave the pane advertising the wrong host terminal to tools that branch on `TERM_PROGRAM`.
- Before forcing, any case-variant key (`Term_Program`, …) is deleted so a stray win32 variant cannot survive alongside the canonical key and get picked nondeterministically — same failure mode the #685 case-insensitive presence check guards against, applied to the force path.

## Tests

- `envFilter.test.ts` (+2): forcing replaces an inherited identity in both builders; a case-variant key is removed rather than shadowed. The existing override test now covers only `TERM`/`COLORTERM`.
- `resolveSpawnEnv.test.ts` (+1): funnel-level — a base-env `TERM_PROGRAM` is forced to `wmux` there too.
- `vitest run` on both files: 55/55 pass. `tsc --noEmit`: clean.

## Note

A workspace profile overlay still applies after the baseline, so a deliberately configured profile `TERM_PROGRAM` can still override the forced value — the override chain stays intact, matching how profile `TERM` already beats the injected default.

## Summary by CodeRabbit

- **Bug Fixes**
  - Enforced consistent terminal identity by setting `TERM_PROGRAM=wmux` for all spawned panes, after any account/workspace environment overlays.
  - Prevented workspace profiles or inherited values (e.g., from other terminal apps) from overriding `TERM_PROGRAM`.
  - On Windows, removed any case-variant `TERM_PROGRAM` keys so only the canonical value remains.
  - Continued to preserve overridable `TERM` and `COLORTERM` defaults.
- **Tests**
  - Expanded coverage to verify `TERM_PROGRAM` precedence and casing behavior.
- **Documentation**
  - Added a changelog entry describing the terminal-capability behavior change.


openwong2kim:

This PR forces `TERM_PROGRAM=wmux` on every pane spawn instead of default-injecting it, while keeping `TERM` and `COLORTERM` as overridable defaults. The motivation is correct: `envFilter.ts` is a blocklist, so a dev build launched from another terminal inherits the host identity (e.g. `TERM_PROGRAM=iTerm.app`) straight through, leaving the pane advertising the wrong host terminal. Forcing the value closes that gap, and the case-variant cleanup (`Term_Program` → removed) is a sensible win32 hardening since Windows env keys are case-insensitive and a stray variant would otherwise survive alongside the canonical key.

The implementation is clean: `applyTerminalCapabilityDefaults()` runs after filtering in both `buildInteractiveShellEnv` and `buildGatedAutomationEnv`, so the two spawn paths stay in lockstep as the module intends. The new tests cover the override hierarchy (base-env TERM wins over default, profile overlay still applies), the force semantics for `TERM_PROGRAM`, case-variant removal, and a regression confirming the capability injection never resurrects a stripped credential key. One review note worth resolving before merge: the case-insensitive presence check (`k.toUpperCase()`) is applied on all platforms, but POSIX env keys are case-sensitive, so a `Term` key would suppress the `TERM` default on macOS/Linux where it should not — consider gating the case-insensitive match to `process.platform === "win32"`. The PR is also currently `CONFLICTING` with main and will need a rebase.

Priority: **P3** — a well-scoped terminal-compatibility improvement that does not block anything, but is worth merging once the case-sensitivity review point and the conflict are addressed.

openwong2kim:
Merged — thanks for following up on the #685 review note!

A couple of things I folded in while resolving the conflict with main, so you know what changed on top of your branch:

- Main had already picked up the capability-defaults half of this (TERM/COLORTERM/TERM_PROGRAM as defaults landed there after you branched), so most of the file rewrite melted away in the merge and your actual delta — promoting TERM_PROGRAM from default to forced — is what shipped.
- Two independent reviews flagged that a workspace profile applying after the builders could still override the forced identity. Since the whole point of forcing is that a pane never advertises the wrong host, I moved the re-assert to the end of the spawn funnel (after the account/profile overlays) and kept your case-variant cleanup as a shared helper. TERM and COLORTERM stay overridable, exactly as you had them.
- Added a changelog fragment (repo convention: changelog.d/<pr>.md instead of editing CHANGELOG.md directly).

The persisted-session replay path in the daemon still bypasses this (recovered sessions replay their saved env verbatim) — that's a pre-existing gap we'll take as a follow-up, not something your change introduced.

Issue #695, opened 2026-07-29
browser tools' RPC fallback still selects a webview target workspace-blind (the gap #679 left open)


openwong2kim:
## Summary

PR #679 closes the workspace-blind **page selection** path in `PlaywrightEngine`: a caller whose workspace cannot be proven is now refused with `WORKSPACE_SCOPE_UNRESOLVED` instead of being handed "the first live guest anywhere". Its own body flags the boundary of that fix:

> Tools that fall back to the app's own screenshot/evaluate handlers when no page is available still pick the default target the way they always have, which is a separate gap and not one this change touches.

This issue is that separate gap, so it doesn't evaporate the moment #679 merges.

## The remaining path

Most browser tools swallow a page-selection failure and fall back to a main-process RPC:

```ts
// src/mcp/playwright/tools/inspection.ts (browser_snapshot; same shape in
// interaction/extraction/state tools)
const page = await engine.getPage(surfaceId).catch(() => null);
if (page) { /* scoped Playwright path */ }
// Fallback: extract page structure via RPC evaluation.
const result = await sendRpc('browser.evaluate', {
  expression: buildDomSnapshotExpression(),
  ...(surfaceId && { surfaceId }),   // ← no surfaceId, no workspaceId
});
```

On the main side, target selection for these RPCs is first-registered-wins:

- `src/main/pipe/handlers/browser.rpc.ts` — `browser.evaluate` (`:684`), `browser.screenshot` (`:616`), and the other `registerLeased` handlers all do `webviewCdpManager.getTarget(surfaceId)` with `surfaceId` optional.
- `src/main/browser-session/WebviewCdpManager.ts` — with no `surfaceId`, `getTarget()`/`ensureAwake()` return `this.sessions.values().next()` — the **first live guest in the process, whichever workspace owns it** (`:372-381`; `ensureAwake` will even wake a discarded surface).

So the exact caller #679 turns away — identity unresolvable, no explicit `surfaceId` — still reaches a foreign workspace's page whenever the tool has an RPC fallback: `getPage()` now throws, the `.catch(() => null)` converts the refusal into "no page", and the fallback RPC selects workspace-blind. After #679, the fallback is not a rare packaged-build path for these callers; it is the path.

## Why it survived #679

[...]

PR #697, opened 2026-07-29
fix(browser): decide the cdp.info registration grace after workspace scoping (#591 follow-up)

snowyukitty:
## Summary

- decide `browser.cdp.info`'s registration grace period **after** the caller's workspace scope has been applied, so a foreign workspace's already-live target can no longer suppress the wait for the caller's own in-flight target
- skip that wait when the backend is `external` — the one backend that never registers a builtin target
- read the reported `workspaceBackend` after the wait rather than before it, so a mid-wait backend change is not reported stale

## The defect

`browser.cdp.info` waits up to 1.5s for an in-flight target registration rather than answering "no targets" while one is still arriving. It decided that wait on the *global* target list, and applied the caller's workspace filter afterwards.

With two workspaces, that ordering reads a stranger's target as proof: B is registered, so the list is non-empty, so A skips the grace period — while A's own target is still registering. A receives `targetsScoped: true, targets: []`. `PlaywrightEngine` correctly reads that as "A owns no target" and may auto-open another browser surface for it. The result is two panes for one workspace, and afterwards a nondeterministic choice between them once registration finishes.

Nothing crosses a workspace boundary here: the scoped response and the downstream workspace guards stay fail-closed either way. This is the narrow follow-up you called optional on #591 — same shape as the fix, one layer up.

Filtering first and waiting on the caller's own set is the whole change.

## Deliberate choices, cheap to overrule

- **A scoped caller that genuinely owns nothing now waits 1.5s before being told so**, where it previously answered immediately whenever another workspace had a target. That is the price of not accepting a foreign target as evidence about the caller. The same wait already existed on the globally-empty path, so this is an existing cost applied to one more case, not a new mechanism.
- **The backend read is split.** The value read at entry decides whether to wait — you cannot retroactively un-wait. The value in the response is read after the wait, because `browser:set-backend` runs on the same event loop and a Settings flip can land mid-wait; reporting a stale `builtin` with zero targets would send the caller into target-miss retries when the honest answer had become `EXTERNAL_BACKEND_UNSUPPORTED`. The second commit is that fix.
- **The wait is gated on `!== 'external'`, not `=== 'builtin'`.** `BrowserBackend` is a closed union, so today these are the same condition. It is written in terms of the backend the exception is actually about, and if the union ever grows it fails toward waiting — which costs a bounded delay, where skipping costs a duplicate surface.
- **External mode (#517/#595) skips the wait entirely.** It has no builtin target by design, so waiting could only add latency to a guaranteed unsupported-error.
- **Unscoped callers are unchanged.**

## Scope

The grace period stays bounded and rechecks once: a registration slower than 1.5s still reports nothing, and the changelog says so rather than promising the race away.

#695 is the separate workspace-blind selection in the RPC fallback path. This change does not touch it.

## Verification

- `npx vitest run src/main/pipe/handlers/__tests__/browser.rpc.test.ts src/main/pipe/handlers/__tests__/browser.rpc.backend.test.ts src/mcp/playwright/__tests__/PlaywrightEngine.test.ts --maxWorkers=1` → 95/95
- mutation-checked in both directions: the three tests from the first commit fail against unmodified `browser.rpc.ts`, and the backend-staleness test fails against the first commit — none of them passes for the wrong reason
- targeted ESLint on the changed files: clean (two pre-existing `no-empty-function` errors elsewhere in `browser.rpc.test.ts` predate this branch and are untouched)

Not run: the full suite and `npm run lint`, both of which are red on `main` already.

Refs #591.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved browser target discovery across workspaces, preventing unnecessary additional browser launches.
  * Browser information now waits briefly for targets belonging to the current workspace before reporting none.
  * External browser mode returns immediately without an unnecessary wait.
  * Browser status now reflects backend changes that occur during the discovery window.


openwong2kim:

This PR fixes a cross-workspace race in `browser.cdp.info`'s target-registration grace period — the narrow follow-up to #591 that was left optional at the time. The defect: the 1.5s wait for an in-flight builtin target was decided against the *global* target list, and the caller's workspace scope was applied only afterwards. So if workspace B already had a live target while workspace A's was still registering, A saw a non-empty global list, skipped the wait, and received an empty scoped result — which `PlaywrightEngine` correctly read as "A owns nothing" and responded to by auto-opening a second browser pane. Two panes for one workspace, and afterwards a nondeterministic choice between them once registration completed.

The fix reorders the logic so the workspace filter runs first and the wait is decided on the caller's own scoped set: only the caller's own target can end the grace period early, and a foreign workspace's target no longer suppresses it. Two further refinements land in the same change. External-browser mode (#517/#595) now skips the wait entirely, since it never registers a builtin target and waiting could only add latency to a guaranteed unsupported-error. And `workspaceBackend` is read *after* the wait rather than before it, so a `browser:set-backend` flip that lands mid-wait over IPC is reported honestly instead of handing the caller a stale `builtin` + zero-targets response that would send it into pointless retries.

The change is well-contained — three test cases added across `browser.rpc.test.ts` and `browser.rpc.backend.test.ts`, mutation-checked in both directions (each fails against the unmodified or intermediate code for the right reason), plus `tsc --noEmit` and targeted ESLint clean. Nothing crosses a workspace boundary: the scoped response and downstream guards stay fail-closed either way. This is a meaningful correctness fix for multi-workspace browser automation, so I am tagging it **P2**.

openwong2kim:
Merged as-is — no changes needed on top.

This went through a three-way independent review (two external models plus my own pass), and none of them found a defect introduced by this change. The scoping-before-wait ordering, the external-backend skip, and the post-wait backend read all held up, and the mutation-checked tests were appreciated — it's rare that a review can confirm each test fails against the exact code it's meant to pin.

Two small follow-up ideas surfaced, both about the pre-existing grace mechanism rather than this PR: the 1.5s wait could race `waitForTarget` against the timeout instead of sleeping blind (an early registration would return immediately), and the 1500ms literal is now duplicated between code and tests and could live in a named constant. Neither blocks anything — noting them here in case you feel like a quick follow-up.

Thanks for the careful write-up; the "deliberate choices, cheap to overrule" section made this easy to review.

Issue #690, opened 2026-07-28
cliShim live-registry PATH tests: exit 10 (READ_FAILED) on Windows CI — unchecked seed(), not the #576 timeout


openwong2kim:
## Summary

`src/main/__tests__/cliShim.test.ts` intermittently fails on `windows-latest` with
**exit code 10**, not a timeout. This is a *different* failure mode from #576
(cold-start timeout), whose fix — `{ timeout: 15_000 }` on the suite, landed in
#581 — is present and working. That timeout does not cover this one.

Exit 10 is `PATH_EDIT_EXIT.READ_FAILED` (`src/main/cliShim.ts:91`): the PATH-edit
script could read the registry by neither strategy, so it failed closed and
refused to write. In other words, **the sandbox key did not exist when `run('add')`
executed** — the write under test never got the chance to be wrong. The assertion
that fires is just the messenger.

## Evidence

Failing job (blocked PR #673, which touches nothing near the CLI shim):
https://github.com/openwong2kim/wmux/actions/runs/30402512339/job/90432818079

```
FAIL src/main/__tests__/cliShim.test.ts > PATH edit (live registry, sandbox key)
     > FullLanguage: add appends without destroying existing entries
AssertionError: expected 10 to be +0 // Object.is equality
 ❯ src/main/__tests__/cliShim.test.ts:221:22
```

Test tally for that run: `2 failed | 8909 passed | 23 skipped (8934)` — the other
failure is the unrelated #689 flake. macOS and Ubuntu passed; the suite is
`describe.skipIf(process.platform !== 'win32')`, so Windows is the only job that
runs it. The job passed on rerun with no code change.

## Root cause

`seed()` (`cliShim.test.ts:191-196`) spawns its own `powershell.exe` to create the
sandbox key and set `Path`, and **its result is discarded**:

```ts
const seed = () => ps(
  `if (Test-Path 'HKCU:\\${SUB}') { Remove-Item 'HKCU:\\${SUB}' -Recurse -Force }\n` +
  `New-Item -Path 'HKCU:\\${SUB}' -Force | Out-Null\n` +
  `Set-ItemProperty -Path 'HKCU:\\${SUB}' -Name 'Path' -Value '${SEED}' -Type ExpandString`,
);
```

`ps()` swallows a non-zero exit into `{ code, out }` and no caller checks it, so a
seed that fails for any reason — runner contention, a `Remove-Item` racing the
[...]

Issue #689, opened 2026-07-28
test(diff.handler): Windows CI flake — 5s timeout + EBUSY on worktree temp cleanup


openwong2kim:
## Summary

`src/main/ipc/handlers/__tests__/diff.handler.test.ts` intermittently fails **only on `windows-latest`** in the `CI Cross-Platform Baseline` workflow. macOS and Ubuntu pass. It is a flake — the same test passes on `main` and other branches on rerun. It is unrelated to the PRs it blocks.

## Evidence

Failing run (blocked PR #673, which does not touch `diff.handler`): https://github.com/openwong2kim/wmux/actions/runs/30402512339/job/90420430118

Two errors from the same test (`diff.handler.test.ts:318` — *"alreadyApplied hunk 포함 선택은 probe 코드로 명시 거부"):

1. **Timeout**
   ```
   FAIL diff.handler.test.ts > ... > alreadyApplied hunk 포함 선택은 probe 코드로 명시 거부
   Error: Test timed out in 5000ms.
     ❯ src/main/ipc/handlers/__tests__/diff.handler.test.ts:318:3
   ```
2. **EBUSY on teardown** (after the timeout, `afterEach` cleanup runs)
   ```
   Error: EBUSY: resource busy or locked, rmdir 'C:\Users\RUNNER~1\AppData\Local\Temp\wmux-diffh-kP8c6C\wt'
     ❯ Object.cleanup src/main/ipc/handlers/__tests__/diff.handler.test.ts:66:20
   Serialized Error: { errno: -4082, code: 'EBUSY', syscall: 'rmdir', path: '...\wmux-diffh-kP8c6C\wt' }
   ```

Test tally for the run: **1 failed | 8910 passed | 23 skipped (8934)** — i.e. a single isolated flake.

## Root cause

`makeScenario` (`diff.handler.test.ts:40-68`) creates a real git repo + a linked worktree under `mkdtempSync(tmpdir(), 'wmux-diffh-')` and returns:
```ts
cleanup: () => rmSync(base, { recursive: true, force: true })
```

Two Windows-specific fragilities compound:

1. **Tight timeout.** The `it(...)` at `:318` chains several real git invocations against the worktree (read → apply → `git add -A` → `git commit` → re-read → re-apply). The default 5s `testTimeout` is tight for git-worktree I/O on GitHub's Windows runner, so under load the test blows past 5s even though it is correct.
[...]

Issue #576, opened 2026-07-24
cliShim live-registry PATH tests: first case flakes on the 5s default timeout (Windows CI) — harness, not a regression


snowyukitty:
## Summary

The `Baseline (windows-latest)` CI job intermittently fails on **one** case in the live-registry suite added by #574:

```
FAIL src/main/__tests__/cliShim.test.ts > PATH edit (live registry, sandbox key)
     > FullLanguage: add appends without destroying existing entries
Error: Test timed out in 5000ms.
  at src/main/__tests__/cliShim.test.ts:212:5

Test Files  1 failed | 548 passed | 1 skipped (550)
```

This is **not a regression in the #573 fix** — the assertion never runs. It is a marginal-timeout flake: the case sits just over vitest's 5000 ms default because it is the *first* one in the group to spawn PowerShell. Filing separately because it will intermittently redden **any** PR that runs the Windows baseline, not just the one where I hit it.

Observed on PR #575, job [89385430176](https://github.com/openwong2kim/wmux/actions/runs/30062032905/job/89385430176).

### It is non-deterministic, and unrelated to the PR that surfaced it

Two consecutive runs on the same PR isolate it:

| Run | Commit | Delta from previous run | `Baseline (windows-latest)` |
|---|---|---|---|
| [30061684395](https://github.com/openwong2kim/wmux/actions/runs/30061684395) | `24aa1f69` | — | **pass** (4m37s) |
| [30062032905](https://github.com/openwong2kim/wmux/actions/runs/30062032905) | `2b083579` | 13 lines of regex in `src/renderer/hooks/__tests__/useRpcBridge.browserClose.test.ts` | **fail** |

The only change between them is a tightened regex inside an unrelated renderer test. `cliShim.test.ts` neither imports nor is imported by anything in that diff. Same code path, one pass and one fail.

`validate` (`tsc --noEmit` + `npm test`), `bench`, and Baseline on `ubuntu-22.04` / `macos-14` were green on the failing run — the suite is `describe.skipIf(process.platform !== 'win32')`, so Windows is the only job that runs it.

### It is cold start, not workload

Timings for all ten cases in that group, from the failing run:

| Case | Time |
|---|---|
| **`FullLanguage: add appends without destroying existing entries`** | **5673 ms ← fails** |
| `FullLanguage: add is idempotent` | 2052 ms |
| `FullLanguage: remove strips only the bin entry` | 2302 ms |
[...]

PR #700, opened 2026-07-29
test: stop two Windows CI fixtures from reporting as product failures (#689, #690)

snowyukitty:
Two Windows-only CI fixtures that fail in a way that reads as a product regression. Test files only. Folded into one PR as suggested on #690.

## #690 — a failed fixture could wear the shim's own exit code

`cliShim.test.ts` drives the real generated PATH-edit script against a throwaway HKCU key. `seed()` spawned `powershell.exe` and its result was discarded, so a seed that never created the key let the case run anyway: the script then failed closed with `PATH_EDIT_EXIT.READ_FAILED`, and CI reported exit 10 — a real, meaningful code from production — on PRs touching nothing near the CLI shim. As the issue put it, the assertion that fired was only the messenger.

- Fixture spawns go through `psOk(label, script)`, which throws on a non-zero exit and carries the child's stderr. A broken harness now names itself.
- `seed()` verifies its own post-condition **inside the same spawn**: the key must actually hold the seeded raw value. Read with `DoNotExpandEnvironmentNames`, because the `%VAR%` entry surviving unexpanded is the property these cases exist to protect. This is the half that catches a seed which *ran* but left nothing behind — the reported failure mode.
- Key removal is guarded by `Test-Path` under `$ErrorActionPreference = 'Stop'` instead of `-ErrorAction SilentlyContinue`. That switch suppresses the message but still leaves `$?` false, so `powershell -Command` exits 1 and "nothing to delete" would have been the same signal as "could not delete" — the very conflation being removed here. I got this wrong first and the suite caught it.
- One case asserts `psOk` throws, so a future revert to swallowing it fails loudly.
- Spawn timeout 20s → 10s, the value the issue proposed, bringing it under the suite's 15s per-test budget. Above that budget a hung `powershell.exe` can only ever surface as a bare vitest timeout naming no spawn in particular.

**What this does not do:** it makes the flake *diagnosable*, not *gone*. If the registry write itself is what is intermittently failing, CI still goes red — it will say `fixture "seed sandbox key" failed with exit N` instead of exit 10. The issue offers a seed retry conditional on that turning out to be the cause; deliberately not added, because adding it now would delete the only signal that could establish it. Happy to follow up once a run tells us.

## #689 — teardown raced the Windows handle release

Each scenario builds a real repo, most with a linked worktree, and teardown deleted the tree with a bare `rmSync`. On Windows those handles stay open for a beat after the last git call returns — Defender may still be reading it too — so the delete threw `EBUSY` out of `afterEach`, failing the run on its own and masking whatever the test had actually reported.

`rmSync` already retries exactly the `EBUSY`/`EPERM`/`ENOTEMPTY` family, so this is a parameter rather than a hand-rolled backoff loop. Measured on Windows 11 with the handle held by a **separate process** — the real git/Defender shape, and the only one a synchronous retry loop could ever outlive:

```
bare rmSync                      THREW EBUSY after 1 ms
rmSync maxRetries:10 delay:100   OK after 625 ms
```

All four temp-tree teardowns route through one helper, including the two describes that build their trees inline rather than through `makeScenario` — one of those creates a linked worktree of its own, so leaving it bare would have mitigated the race everywhere except a place it demonstrably occurs.

The 5s default is also too tight for git-worktree I/O on the Windows runner, so a correct test failed for being slow. Raised once for the file rather than on each of ten describes, since every case in it drives real git. `vi.setConfig` is file-local — verified on Vitest 4.1.0 by running a file that sets 30s ahead of one that does not, in the same worker: the second still timed out at 5000ms. No other file inherits the larger budget.

## Deliberate, cheap to overrule

- **10s spawn cap.** A single spawn between 10s and 15s, in a case whose total stays under the suite budget, now fails where it previously passed. A spawn that slow is treated as pathological; the alternative — a cap above the suite budget — can never surface as its own error. Locally these spawns run 1.5–1.7s.
- **No seed retry**, per the note above.
- **No `git worktree remove` before the delete.** `rmSync`'s retry covers the observed errno without adding a spawn per teardown that can fail on its own.
- **No changelog fragment** — test-only, matching #581.

## Verification

Run on Windows 11, which is the platform that actually executes both suites (`cliShim`'s block is `describe.skipIf(process.platform !== 'win32')`).

- `npx vitest run src/main/__tests__/cliShim.test.ts --maxWorkers=1` → 26 passed, 8 skipped
- `npx vitest run src/main/ipc/handlers/__tests__/diff.handler.test.ts --maxWorkers=1` → 37 passed, 1 skipped
- Mutation-checked, both new guards: a seed forced to exit non-zero reports `fixture "seed sandbox key" failed with exit 9`; a seed that writes the wrong value reports `seed left Path as 'SABOTAGE'`. Before this change both surfaced as `expected 10 to be +0`.
- ESLint on both files: 66 warnings, 0 errors — byte-identical to the same two files on `main`, so nothing new.

Not run: the full suite and `npm run lint`, both already red on `main`.

Refs #690, #689.

## Summary by CodeRabbit

* **Tests**
  * Improved Windows test reliability using stricter PowerShell timeout handling and more consistent `{code, out, err}`-style reporting for clearer failure attribution.
  * Strengthened fixture and sandbox setup/teardown checks to better detect and explain non-zero exits and avoid false positives from leftover state.
  * Increased test timeouts for faster Windows repo I/O and made scenario directory cleanup more resilient with a shared retry-capable removal helper.

### DIFF
--- a/src/main/__tests__/cliShim.test.ts
+++ b/src/main/__tests__/cliShim.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, beforeAll } from 'vitest';
 import {
   buildShimCmd,
   buildPathEditScript,
@@ -179,19 +179,28 @@ describe.skipIf(process.platform !== 'win32')('PATH edit (live registry, sandbox
 
   // Must stay below the suite's per-test budget above: a spawn that outlives it
   // can only ever surface as a bare vitest timeout, which says nothing about
-  // which of the three or four spawns in a case actually hung.
+  // which of the three or four spawns in a case actually hung. Warm spawns
+  // measure ~1s, so this is a wide margin over anything legitimate.
   const SPAWN_TIMEOUT_MS = 10_000;
+  // The first powershell.exe in a process pays a one-time cold start — its own
+  // binaries and the .NET images are not in the OS file cache yet. #576
+  // measured 5.6s for it; a run on this branch exceeded 10s. It is absorbed
+  // once in a hook with a budget of its own rather than billed to whichever
+  // case happens to run first, which is how #576 flaked and what its 5s → 15s
+  // widening left unaddressed — a wider budget moves that threshold, it does
+  // not stop a one-time cost from landing inside a per-test one.
+  const WARMUP_TIMEOUT_MS = 60_000;
 
-  const ps = (script: string) => {
+  const ps = (script: string, timeoutMs: number = SPAWN_TIMEOUT_MS) => {
     try {
       return { code: 0, out: execFileSync(PS, ['-NoProfile', '-NonInteractive', '-Command', script], {
-        encoding: 'utf8', windowsHide: true, timeout: SPAWN_TIMEOUT_MS,
+        encoding: 'utf8', windowsHide: true, timeout: timeoutMs,
       }), err: '' };
     } catch (e) {
       const err = e as { status?: number | null; stdout?: string; stderr?: string; signal?: string };
       // A timeout kills the child with a signal and leaves status null. Say so —
       // "hung" and "exited non-zero" need different fixes.
-      const killed = err.signal ? `killed by ${err.signal} after ${SPAWN_TIMEOUT_MS} ms` : '';
+      const killed = err.signal ? `killed by ${err.signal} after ${timeoutMs} ms` : '';
       return {
         code: err.status ?? -1,
         out: err.stdout ?? '',
@@ -208,8 +217,8 @@ describe.skipIf(process.platform !== 'win32')('PATH edit (live registry, sandbox
    * is #690, where a seed that never created the key surfaced as READ_FAILED
    * and read as a genuine regression.
    */
-  const psOk = (label: string, script: string) => {
-    const r = ps(script);
+  const psOk = (label: string, script: string, timeoutMs?: number) => {
+    const r = ps(script, timeoutMs);
     if (r.code !== 0) {
       throw new Error(`fixture "${label}" failed with exit ${r.code}\n${r.err || r.out}`.trim());
     }
@@ -254,6 +263,13 @@ describe.skipIf(process.platform !== 'win32')('PATH edit (live registry, sandbox
     return ps(prefix + buildPathEditScript(BIN, op, SUB, BAK));
   };
 
+  // Pay the cold start here, where it has its own budget and its own name. If
+  // PowerShell is genuinely unusable this reports that once, up front, instead
+  // of disguising it as the first case's failure.
+  beforeAll(() => {
+    psOk('warm up powershell', 'exit 0', WARMUP_TIMEOUT_MS);
+  }, WARMUP_TIMEOUT_MS + 5_000);
+
   afterEach(() => {
     // Also checked: teardown that could not run leaves keys behind, and the
     // damage lands on the *next* case — the failure would be attributed to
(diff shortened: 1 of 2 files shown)


openwong2kim:

This PR is a test-only hardening pass that fixes two Windows-only CI fixtures so they stop reporting as product failures: the `cliShim` PATH-edit suite (#690) and the `diff.handler` worktree suite (#689). Both are the flakes that red-light unrelated PRs and force re-runs.

In `cliShim.test.ts`, the unchecked `seed()` spawn now goes through a `psOk()` helper that throws on a non-zero exit and verifies its own post-condition in the same spawn, so a broken harness names itself instead of surfacing as the shim's `READ_FAILED` (exit 10). The spawn timeout drops from 20s to 10s to stay under the suite's 15s budget, teardown is guarded the same way, and a regression test pins the fix. In `diff.handler.test.ts`, the file-level timeout rises to 30s for git-worktree I/O, and every temp-tree teardown routes through a `removeScenarioTree()` helper that uses `rmSync`'s built-in `maxRetries`/`retryDelay` to outlive the EBUSY/EPERM window Windows leaves on linked worktrees.

The diff is confined to the two test files, +91/−13, and matches the PR body claim-for-claim. A couple of deliberate non-changes are called out in the description (no seed retry, no `git worktree remove`, no changelog) — both reasonable for test-only work mirroring #581.

Priority: **P3**. Low severity, no runtime impact; value is unblocking clean CI signal on PRs that touch neither file.

snowyukitty:
CI caught my own guard case on Windows, and the failure is worth writing down rather than just fixing.

`powershell.exe` was killed by the new 10s spawn cap and the assertion — which pinned the exit value at `3` — reported `psOk` as broken when `psOk` had worked exactly as intended: it threw, named the fixture, and said the child was killed after 10000 ms. Two faults, both mine.

**The cold start does not belong inside a per-test budget.** The first `powershell.exe` in a process pays a one-time cost — its binaries and the .NET images are not in the OS file cache yet. #576 measured 5.6s; this run exceeded 10s. Whichever case runs first is billed for it, which is how #576 flaked in the first place, and raising the per-test budget from 5s to 15s in #581 only moved the threshold. `e98db772` absorbs it in `beforeAll` under a budget of its own. Every in-case spawn is then warm — measured ~1s — so the 10s cap has a wide margin over anything legitimate, and a PowerShell that genuinely cannot start is reported once, up front, by name.

**The guard now asserts the property, not the exit value.** A spawn killed on a slow runner is still `psOk` doing its job; an assertion that calls that a regression is the same mistake this PR set out to fix, one level up.

For the record: reviewing this branch before it was opened, two reviewers flagged the 10s cap as a false-failure risk on a loaded runner. I kept the value on the grounds that no failure had been substantiated. It substantiated itself on the first CI run.

Verified on Windows 11: 26 passed / 8 skipped; the guard still fails if `psOk` is made to swallow a non-zero exit; `tsc --noEmit` clean; ESLint unchanged from `main`.

openwong2kim:
Merged. Thanks for this one — splitting "the fixture broke" from "the shim broke" is the right cut, and the comments explaining why each spawn gets checked will save whoever hits this next a lot of guessing.

One thing that came up in review and we're leaving alone for now: `vi.setConfig` in diff.handler.test.ts raises `testTimeout` but not `hookTimeout`, and most of the heavy git I/O actually sits in `beforeEach` via `makeScenario()`, still on the 10s default. Nothing is failing on it today, so we'll add it if a hook ever starts timing out.

PR #701, opened 2026-07-29
fix(browser): scope CDP target selection to the caller's workspace (#695 main-side leg)

snowyukitty:
First of the two halves you sketched on #695: the main-side authority. The MCP client half is deliberately separate — see *Scope*, which also carries one correction to the sketch.

## The lookup

One `WebviewCdpManager` serves every workspace in the process. `getTarget(surfaceId?)` answered `this.sessions.values().next()` when given no surface — the first live session, which is whichever workspace opened a browser first — and returned any named surface outright when given one. `ensureAwake()` matched it and went further: with nothing live it remounts and reloads a discarded guest. That is not a read. It is an action taken inside a workspace the caller does not own.

`getTarget`, `ensureAwake` and `waitForTarget` now take the caller's workspace.

- Naming one restricts **every** branch to targets that provably belong to it, the explicit-`surfaceId` branch included — a surface id from another workspace is precisely what a scoped caller must not reach.
- Ownership is never inferred from being the only candidate: an untagged target is refused rather than assumed, the rule `browser.cdp.info` already applies to the target list (#591).
- Naming none is unscoped exactly as before. The pipe is same-trust for first-party CLI callers and single-workspace setups would break for no gain — the compatibility note in your sketch.

The wake path needs the owner to outlive the session, because a discarded guest has none, so `GuestState` mirrors the workspace reported at register time — recorded when known, never cleared, so a legacy re-registration that omits the id cannot downgrade a surface whose owner is already established. Ownership is re-checked *after* the wake as well as before it: the surface re-registers in between, commonly by a path that reports no workspace at all. The wake promise is shared between concurrent callers, so that check belongs to whoever receives the result rather than whoever started it.

## Two places the scope had to reach further than the lookup

**`browser.cdp.target` was an existence oracle.** Post-checking an unscoped wait cannot hide anything: a live foreign surface answers instantly, a surface that never existed costs the full timeout, and the latency alone says which is which. `waitForTarget` now carries the scope and stays parked on a registration the caller does not own, so the two cases are identical in message and in timing. `register()` no longer drops parked waiters, so one still owed its own registration can receive it.

**`browser.navigate` falls back to the renderer bridge**, and that bridge resolves a surface with no workspace check — it takes the caller's id, or picks its own default when there is none. Neither is safe to hand a scoped caller unless this side has already proven ownership. A scoped caller that resolves no target is now refused; one that resolved a target keeps the bridge, pinned to that exact surface, because leaving the id absent hands the choice back to the default pick this change exists to remove. Unscoped callers are untouched. This is the only renderer fallback that dropped the workspace — `browser.tabs` and `browser.close` already forward it (#575, #586).

## Scope

**Nothing changes behavior today.** No caller sends a `workspaceId` to any of these RPCs yet — `browser.cdp.info` and `browser.open` are the only two that carry one, and neither resolves a target through these lookups. This PR adds the authority; the follow-up makes the MCP tools exercise it.

That split is also where the sketch needs a correction: it says the MCP tools *"hold `requireWorkspaceId` already"*. They do not. `src/mcp/index.ts:836` injects the resolver into `registerNavigationTools` only — `registerInspectionTools`, `registerInteractionTools`, `registerStateTools` and `registerExtractionTools` take `(server)` and nothing else. So the client half is not a field added to a few `sendRpc` calls; it needs the resolver threaded into four more modules, or routed through the `PlaywrightEngine` singleton that already holds one. Worth deciding before that PR is written.

Deliberately left, each cheap to overrule:

- **Lease acquisition is not atomic with the ownership check.** `resolveTargetSurface` returns a surface id and the lease is taken after an await. For that id to become foreign the renderer would have to register the same surface id under a different workspace; and because each handler re-resolves under its own scope, the bounded worst case is briefly unthrottling a foreign guest, not driving it.
- **`register()`'s `workspaceId ?? prev?.workspaceId` inheritance** is pre-existing #554 lifecycle behavior. Rebinding ownership semantics is a separate change.
- **Renderer-side navigate scoping.** A `decideBrowserNavigate` mirroring #586's `decideBrowserClose` would let a scoped caller be *routed* instead of refused. That is renderer work, not main-side authority.
- **No changelog fragment**, since nothing user-visible changes until the client half lands. Happy to add one there, or here if you would rather it be recorded now.

## Verification

- `npx vitest run src/main/browser-session/__tests__ src/main/pipe/handlers/__tests__ --maxWorkers=1` → 640 passed
- Wider sweep including `src/mcp/playwright/__tests__` → one failure, `lazyPlaywright.test.ts > loads playwright-core on first call` at 11.4s against the 5s default. It passes alone on this branch **and** on unmodified `origin/main`, reddening only under batch load — the same one-time-cost-inside-a-per-test-budget shape as #689/#690, here from loading `playwright-core`. Not caused by this change.
- Mutation-checked against unmodified `origin/main`: thirteen of the added cases fail there. The four that pass are the compatibility assertions — they pin *historical* unscoped behavior, so passing on `main` is what they are for.
- Each security check was verified to have something holding it. Deleting the in-flight wake join check left all 639 tests green; that branch now has a case that fails without it.
- ESLint on the four changed files: 9 problems, 3 errors — byte-identical to the same files on `main`. (Pre-existing `no-empty-function` on `new Promise(() => {})`.)

The four commits are the review trail: the first is the change, the next two respond to findings from independent review rounds, the fourth closes a coverage gap. Squash as you see fit.

Refs #695.

## Summary by CodeRabbit

* **Security & Access Control**
  * Workspace-scoped browser operations now restrict target selection, navigation, automation, screenshots, input, and state access to owned surfaces.
  * Unauthorized or untagged targets are rejected for scoped requests, preventing cross-workspace access and existence leaks.
  * Wake-up and remount flows re-check ownership before returning a target.

* **Compatibility**
  * Unscoped browser requests retain their existing behavior.
  * Renderer-based navigation continues to work for verified owned surfaces.

### DIFF
--- a/src/main/browser-session/WebviewCdpManager.ts
+++ b/src/main/browser-session/WebviewCdpManager.ts
@@ -55,6 +55,20 @@ const DISCARD_AFTER_MS =
 // page has to fully reload before dom-ready re-registers it.
 const WAKE_TIMEOUT_MS = 15_000;
 
+/**
+ * Workspace ownership test for a target or guest (#695).
+ *
+ * A caller that named no workspace is unscoped and matches everything — that
+ * is the pre-#695 contract the pipe still depends on. A caller that named one
+ * matches only what it can be *proven* to own: an untagged target is refused
+ * rather than assumed, the same rule browser.cdp.info applies to the target
+ * list (#591). Ownership is never inferred from being the only candidate.
+ */
+function ownedBy(ownerWorkspaceId: string | undefined, callerWorkspaceId: string | undefined): boolean {
+  if (!callerWorkspaceId) return true;
+  return ownerWorkspaceId === callerWorkspaceId;
+}
+
 interface GuestState {
   /** Effective visibility reported by the renderer (workspace ∧ window ∧ ¬zoom ∧ selected). */
   visible: boolean;
@@ -75,6 +89,11 @@ interface GuestState {
   discarded: boolean;
   /** Dwell timer armed while the guest is throttle-eligible; fires a discard. */
   discardTimer: NodeJS.Timeout | null;
+  /** Owning workspace, mirrored from the last register() that reported one
+   *  (#695). It lives here rather than only on the session because a discarded
+   *  guest has no session, and ensureAwake() must still be able to tell whether
+   *  the surface it is about to wake belongs to the caller. */
+  workspaceId?: string;
 }
 
 export class WebviewCdpManager {
@@ -225,6 +244,12 @@ export class WebviewCdpManager {
     // re-register fires on every navigation, and re-arming grace there would
     // let a hidden page that reloads periodically stay unthrottled forever.
     const gs = this.ensureGuestState(surfaceId);
+    // Remember the owner so a later wake can be scoped (#695). Only ever
+    // recorded, never cleared: a legacy re-registration that reports no
+    // workspace must not silently downgrade a surface we already know the
+    // owner of. A surface belongs to one workspace for its whole life, so
+    // there is no case where this goes stale.
+    if (workspaceId) gs.workspaceId = workspaceId;
     // A registration IS the wake: the renderer remounted the webview (wake
     // signal, user click, or surface became visible again).
     gs.discarded = false;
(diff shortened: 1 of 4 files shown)


openwong2kim:

This is the main-side authority leg of #695, the gap that #679 deliberately left open: until now `WebviewCdpManager.getTarget()`/`ensureAwake()` answered `sessions.values().next()` (the first live guest in the process) whenever no `surfaceId` was supplied, so a caller whose workspace could not be proven could still read, drive, navigate, or wake another workspace's page. The PR adds an `ownedBy(owner, caller)` ownership test plus a `workspaceId` mirror on `GuestState` (recorded once at register time and never cleared, so even a discarded guest — which has no session — keeps its owner), and threads the caller's `workspaceId` through `getTarget`, `ensureAwake`, and `waitForTarget`, including the explicit-`surfaceId` branch (a named foreign surface is exactly what a scoped caller must not reach) and the renderer navigate fallback (now pinned to the proven surface, or refused outright when nothing was resolved). Wake handling gets extra care: the post-wake re-check closes the legacy-re-register-runs-untagged race, and a scoped caller joining an in-flight wake applies its own check to the shared result rather than inheriting whoever started it; `browser.rpc.ts` passes `scopeOf(params)` into every lookup site.

It is a security-hardening change against the cross-workspace read/drive class inside the same-user ceiling (the same class as #679/#554), gated behind an opt-in scope: no caller sends `workspaceId` today, so behavior is unchanged until the follow-up MCP-client half threads the resolver into the four tool modules that currently lack it (a correction to the original #695 sketch that the PR body flags). The 582 additions are overwhelmingly tests — ownership on every branch, the wake re-check, the navigate fallback in every combination, and the latency-indistinguishability property of the scoped wait — with mutation-checking against `origin/main` cited. Marking **P2** to match #695; the work is sound and well-scoped, and the one pre-existing lint (identical on `main`) is unrelated.

openwong2kim:
Merged. Thank you — this was a pleasure to review.

Two reviewers went over it independently. Codex came back with nothing. The other raised three, and reading the code myself I only kept one of them, and only as a wording nit: the test names and some comments read as unconditional security properties when the scope is cooperative — it holds for callers that name their workspace honestly, which is the #113 ceiling and not something this PR could have closed. Worth softening whenever you next touch these files, not worth a round trip now.

The other two I disagreed with, and I want to say why in case someone finds them later. The suggestion to make `gs.workspaceId` first-write-wins would have desynchronized it from the session's `workspaceId ?? prev?.workspaceId` — `getTarget` reads the session and `ensureAwake`'s discarded scan reads the guest state, so that would have left two disagreeing authorization sources where there is currently one. Mirroring the existing semantics was the right call. And the point about the `waitForTarget` mock not holding the timing property is true, but the property is held one layer down in the manager unit test; it's split, not missing.

What made this reviewable was the parts you did not have to write: the correction to the sketch about `requireWorkspaceId`, the note that the wake re-check had nothing holding it until the fourth commit, and the honest accounting of `lazyPlaywright.test.ts` failing under batch load on `main` too. Saying which of your own assertions pass on `main` and why is not a thing most PRs bother to do, and it is exactly what let me spend the time on the parts that mattered.

Squashed, since the four commits are review trail rather than separable changes. Refs #695 stays open for the client half — one thing to keep in mind there: `PlaywrightEngine.resolveSelectionContext` returns `key: surf:${id}` with no `workspaceId` when a surface is named explicitly, so that path will not exercise the authority you just added unless it is threaded through too.

PR #703, opened 2026-07-29
feat(deck): enforce the orchestrator's turn contract with hooks and skills

openwong2kim:
## What

The `claude-pty` orchestrator brain's execution contract used to live entirely in prompt text typed into the TUI — nothing enforced it. This PR moves the contract into the generated settings profile and the brain home:

- **Stop gate**: the orchestrator can no longer end a turn while delegated panes are still `running`/`awaiting_input`. The gate verdict rides the `hooks.signal` RPC response; the bridge's new `--gate` mode reads it and exits 2, so the model sees the refusal reason and keeps driving its workers. Fail-open backstops: 3 consecutive blocks, stale (>30s) or missing fleet snapshots, and a dead main process.
- **Self-explaining denials**: `AskUserQuestion` joins the deny list, and the PreToolUse deny command becomes a generated script (0600, unlinked on dispose) that tells the model which tool was blocked and to use `deck_ask_decision` instead — pwsh-quoting-safe by taking the tool name as an argv argument.
- **wmux-owned brain skills**: `delegate` and `approve` skills are installed into `~/.wmux/brains/<wsId>/.claude/skills/` on spawn. Ownership is marker-based: a SKILL.md the operator took over is never overwritten (read failures other than ENOENT count as owned).

## Review

3-model panel (Claude / Codex / GLM) ran on the diff; all confirmed findings are fixed in the last four commits (turn-scoped block counter, gate transport fail-closed split, snapshot freshness, gate-mode token priority, skill-ownership read failure). One finding was rejected by design: the 3-block cap is the deliberate deadlock escape, documented inline.

## Tests

`stopGate` 6 cases, `brainSkills` 7, adapter gate/counter cases, bridge gate-loss 6 + token 2 — 9,056-test parallel suite green, tsc 4-project build clean.

## Summary by CodeRabbit

- **New Features**
  - Added orchestration safeguards that keep turns active while worker panes still have outstanding work.
  - Added built-in delegation and approval skills for guided workflows.
  - Added durable decision cards for questions and approval requests.
  - Added gate mode that clearly reports blocked actions and safely fails closed when verdicts are unavailable.

- **Bug Fixes**
  - Blocked-tool errors now provide actionable explanations and alternatives.
  - Improved hook delivery and retry handling for timeouts, connection failures, and missing responses.


openwong2kim:

This PR moves the `claude-pty` orchestrator's execution contract out of preamble text and into enforced machinery: a Stop gate that refuses to end a turn while worker panes are still `running`/`awaiting_input` (with deliberate fail-open backstops — null/stale snapshots, a 3-block consecutive cap, and a dead main), generated `delegate`/`approve` skills installed into the brain home with marker-based operator-ownership protection, and self-explaining PreToolUse denials that tell the model which tool was blocked and what to reach for instead (`deck_ask_decision` replacing `AskUserQuestion`). The gate verdict rides the existing `hooks.signal` RPC response and is turned into exit 2 by a new `--gate` mode in the bridge, which fails closed only when a request provably reached main and then lost its answer (every connect failure fails open so a Claude Code session running outside wmux is never wedged).

The change matters because the orchestrator's worst habit — delegate, say "delegated", stop, and leave the fleet undriven until a wake event happens to fire — is now structurally prevented rather than urged in prompt text, and the new `isGateAnswerLost` / token-priority reasoning closes a real silent-allow bug (a host with both tokens on disk authenticated the gate with the wrong credential and downgraded every refusal to an allowed Stop). Coverage is thorough: 6 stopGate cases, 7 brainSkills cases, 5 bridge `isGateAnswerLost` cases plus 2 token-precedence cases, and the 9,056-test parallel suite is reported green with a clean 4-project `tsc`.

One thing worth confirming during owner dogfood: the 3-block cap is the documented deadlock escape, but it means a genuinely-stuck worker pane costs a nudge rather than a held turn — fine by design, just worth watching that the cap is not masking a status-reporting bug in the mirror.

PR #705, opened 2026-07-29
fix(channels): wake idle agents after daemon restart

### LINKED ISSUE #693: Wake worker still reads the registry's liveness field, so an idle agent stays unwakeable after a daemon restart
#692 moved the *mention* path off the principal registry's `liveness` field and onto the daemon's own session table. The wake worker was not moved, and it has the same problem.

## The gap

`PrincipalService.livePtyIdOf()` returns a pty only when the record says `live`:

https://github.com/openwong2kim/wmux/blob/main/src/daemon/principals/PrincipalService.ts#L104-L108

and that is what the wake worker resolves through:

https://github.com/openwong2kim/wmux/blob/main/src/daemon/channels/channelWakeWorker.ts#L396-L400
https://github.com/openwong2kim/wmux/blob/main/src/daemon/index.ts#L4579

The `liveness` field can only be set back to `live` by a renderer. The daemon backfills every pane-agent to `stale` when it restarts, and the two paths back are:

- the `daemon:connected` re-seed, which filters to panes the renderer does **not** already know (`!surfaceAgent[id]?.name`), so a known agent is skipped
- `principalRegisterPane` on an activity broadcast, which requires the pane to emit output

An **idle** agent hits neither. It has finished its work and is waiting, which is exactly the state you want to wake it from, and it stays marked `stale` indefinitely — so `livePtyIdOf` returns nothing and the worker declines to nudge it.

## Repro sketch

1. Start an agent in a pane, let it go idle.
2. Restart the daemon.
3. Address the agent through a path that goes via the wake worker.
4. It is never nudged; the record is `stale` with no way back until the pane prints something.

Not verified live yet — the reasoning is from the code paths above, same as the mention-path bug in #692, which had the same shape.

## Fix direction

`resolvePanePin` already does this for the mention path: ownership from the registry, liveness from `sessionManager.getSession(ptyId)?.meta.state`, live meaning `attached || detached`.

https://github.com/openwong2kim/wmux/blob/main/src/daemon/principals/resolvePanePin.ts

The session table is rebuilt from persisted metadata on startup with no renderer involved, which is what makes it answerable here. `livePtyIdOf` could take the same lookup, or the wake worker could resolve through `resolvePanePin` directly.

Worth checking whether anything else reads `liveness` for a decision rather than for display before changing the field's meaning.

snowyukitty:
## Summary

Restore channel wake targeting for idle agents after a daemon restart. The
principal registry supplies the PTY coordinate while the daemon-owned live
session snapshot remains the liveness authority; a stable principal that no
longer resolves now fails closed instead of waking a sibling agent.

## Changes

- Return a pane principal's last registered PTY coordinate without interpreting
  renderer-owned registry liveness.
- Accept that coordinate only when it appears in the daemon's attached/detached
  session snapshot; existing workspace, deferred-session, agent, and attached-
  Claude checks stay intact.
- Refuse missing, dead, mismatched, or agent-less exact principal targets rather
  than falling back to a heuristic that could inject into a sibling pane.
- Cover restart backfill through the real principal service and wake-worker
  production seam, plus fail-closed sibling routing.

### Fixed

- **Channel delivery wakes idle agents again after a daemon restart.** Restart
  recovery marks every pane principal stale, and an agent that has finished and
  is waiting produces no output to refresh that flag, so the wake worker could
  leave it unwakeable indefinitely. The registry now supplies only the pane's
  PTY coordinate; the daemon's rebuilt attached/detached session table decides
  whether that target is live. Exact principal targets that cannot be verified
  are skipped rather than redirected to a sibling agent. (#705)

## Stability tier impact

n/a — no RPC shape, event type, permission enforcement point, or Named Pipe
contract changed.

## Test plan

- Principal routing regression suites: 51 passed.
- Daemon channel and principal suites: 459 passed.
- `npx tsc -p tsconfig.daemon.json --noEmit`
- `npx tsc --noEmit`
- `npm run build:mcp && npm run probe:mcp`
- `npx tsc --noEmit -p relay/tsconfig.json`
- ESLint passed on all touched non-entrypoint files; `src/daemon/index.ts`
  retains only its pre-existing baseline findings outside this diff.
- `npm run test:runtime`: 90 passed, 9 skipped in the clean run. A later
  concurrent stress rerun hit unrelated five-second timeouts in the untouched
  device-store runtime file; 47/48 passed when that file was rerun alone.
- Full `npm test` reached 9,105 passed before unrelated Windows timeout/`EBUSY`
  failures under parallel load. Every failed file passed when rerun in
  isolation (105 passed, 8 skipped).

## Related issues / PRs

Closes #693.
Follow-up to #692.

## Reviewer checklist

- [x] Tests cover the new behavior and the regression case.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed channel delivery after daemon restarts so previously idle agents can be woken again.
  - Ensured messages route only to the exact verified agent session.
  - Prevented stale or unavailable targets from incorrectly waking sibling agents or bare shell panes.
- **Tests**
  - Added coverage for restart recovery, stale routing data, unavailable targets, and agent-only wake behavior.
- **Documentation**
  - Added a changelog entry describing the restart and routing fixes.

### DIFF
--- a/changelog.d/705.md
+++ b/changelog.d/705.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **Channel delivery wakes idle agents again after a daemon restart.** Restart
+  recovery marks every pane principal stale, and an agent that has finished and
+  is waiting produces no output to refresh that flag, so the wake worker could
+  leave it unwakeable indefinitely. The registry now supplies only the pane's
+  PTY coordinate; the daemon's rebuilt attached/detached session table decides
+  whether that target is live. Exact principal targets that cannot be verified
+  are skipped rather than redirected to a sibling agent. (#705)
--- a/src/daemon/channels/__tests__/channelWakeWorker.test.ts
+++ b/src/daemon/channels/__tests__/channelWakeWorker.test.ts
@@ -398,7 +398,7 @@ describe('pickTarget — never guess', () => {
 describe('pickTargetWithPrincipal — R2 registry direct targeting', () => {
   const PID = 'pane:ws-b/p1';
 
-  it('directly selects the LIVE principal\'s ptyId session without the heuristic (auto-name memberId)', () => {
+  it('directly selects the principal ptyId present in the live snapshot without the heuristic (auto-name memberId)', () => {
     // memberId "w2-1(codex)" never matches the slug heuristic — without the
     // principal path this member would never get nudged. With two same-slug
     // panes the heuristic would have returned null on ambiguity.
@@ -412,7 +412,7 @@ describe('pickTargetWithPrincipal — R2 registry direct targeting', () => {
     expect(target?.id).toBe('pty-y');
   });
 
-  it('a stale principal (undefined ptyId) falls back to the existing heuristic', () => {
+  it('a missing principal coordinate falls back to the existing heuristic', () => {
     const target = pickTargetWithPrincipal(
       [session({ id: 'only', lastDetectedAgent: 'codex' })],
       'ws-b',
@@ -501,7 +501,7 @@ describe('pickTargetWithPrincipal — R2 registry direct targeting', () => {
 
   it('the heuristic fallback carries the same agent-required discipline (no shell nudge)', () => {
     const PID2 = 'pane:ws-b/pX';
-    // Stale principal → heuristic fallback; the only pane is an agent-less
+    // Missing principal coordinate → heuristic fallback; the only pane is an agent-less
     // shell → null (never auto-submit into a bare shell).
     expect(
       pickTargetWithPrincipal(
--- a/src/daemon/channels/__tests__/channelWakeWorker.test.ts
+++ b/src/daemon/channels/__tests__/channelWakeWorker.test.ts
@@ -16,6 +16,14 @@ import {
   type WakeUnreadEntry,
   type WakeSessionView,
 } from '../channelWakeWorker';
+import {
+  PrincipalService,
+  type PrincipalWriterLike,
+} from '../../principals/PrincipalService';
+import {
+  panePrincipalId,
+  type PrincipalState,
+} from '../../../shared/principals';
 
 function entry(partial: Partial<WakeUnreadEntry>): WakeUnreadEntry {
   return {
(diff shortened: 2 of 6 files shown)


openwong2kim:

This PR fixes #693: after a daemon restart, every pane-agent principal is backfilled to `stale`, and the old `PrincipalService.livePtyIdOf()` refused to return a ptyId for any stale record — so the wake worker could never nudge an idle agent that had finished its work and was quietly waiting. The fix cleanly separates the two concerns: `ptyIdOf()` now returns the last-registered PTY coordinate without interpreting registry liveness, and the wake worker proves liveness itself against `listLiveSessions()` (the daemon-owned attached/detached session table, which is rebuilt from persisted metadata on startup with no renderer involved). This is the right authority split — the registry owns identity, the session table owns liveness.

A second, equally important change tightens `pickTargetWithPrincipal` to **fail closed**: when a stable principal coordinate is missing, dead, workspace-mismatched, or no longer hosts an agent, the worker now returns `null` instead of falling back to the slug heuristic. Previously a stale or dead direct hit could silently re-route a nudge to a sibling pane in the same workspace — a cross-pane injection that this PR eliminates. The test suite covers both the restart-recovery path (exact rebuilt session is selected despite a stale registry row) and the fail-closed sibling scenarios.

Labeling **P2**. The bug it fixes is real and leaves idle agents permanently unwakeable after any daemon restart, which is a common operational scenario. The diff is well-scoped, the test coverage is thorough (restart backfill through the real `PrincipalService`, fail-closed routing, agent-less shell guard), and the changelog fragment follows the new convention. No blocking concerns from review.

snowyukitty:
Addressed the stale-comment review note in e2a69588: session death still updates registry/UI liveness, while the comment now names listLiveSessions() as the wake-target safety authority.

openwong2kim:
Live-dogfooded this against a real daemon process (not unit tests) to confirm the fix works end-to-end.

Setup:
- Built this branch into a daemon bundle in an isolated worktree, spawned it against a temp `USERPROFILE`/pipe, drove it entirely over the production JSON-RPC wire.
- One workspace with two ambiguous eligible panes (both `claude`-detected), a pane-agent principal registered on only one of them (paneA), a channel with a message mentioning that principal.
- SIGKILLed the daemon and relaunched it against the same state dir — this triggers `PrincipalService`'s restart backfill, which flips every pane-agent principal's `liveness` to `stale`, reproducing the exact precondition of #693.

Result: the wake worker correctly nudged paneA after restart —

```
[wake] nudging pr705-agent-a for ch-...#agentA (1 unread)
[wmux] #pr705-dogfood: 1 unread (1 mention you) — run: wmux channel read ch-... --since
```

The unregistered, ambiguous sibling pane (paneB) was correctly NOT nudged, confirming routing goes to the exact registered principal rather than falling back to the ambiguous slug heuristic.

Merging.

Issue #691, opened 2026-07-28
Interactive OpenTUI apps (omp / oh-my-pi) render a fully blank/frozen pane — emulator readScreen stalls


spc1jh:
## Summary

Full-screen TUI apps built on **OpenTUI** (e.g. `omp` / oh-my-pi, an AI coding agent)
do not render at all inside a wmux pane. The pane stays **completely blank/frozen** for
the entire session — the app is running and emitting output, but nothing is drawn. The
only way out is pressing Ctrl+C many times, at which point the buffered UI briefly
flashes as the app tears down.

The same `omp` binary renders perfectly in a standalone Windows Terminal / PowerShell /
cmd window. Other agent TUIs (Claude Code, Codex) render fine inside wmux — so this is
specific to OpenTUI's rendering model.

This looks like a **severe case of the same repaint/redraw family** already tracked in
#678 (Kimi Code CLI jitter), which itself references #318, #333, and #525. Where #678 is
a mild jitter during input-box growth, OpenTUI's more aggressive full-frame renderer
turns the same underlying problem into a total render failure.

## Environment

- OS: Windows 11 Enterprise (10.0.22631)
- wmux: **3.35.0** (note: #678 reports the redraw family still present on 3.37.2)
- Pane shell: PowerShell 7.6.4 over ConPTY
- App: `omp` (oh-my-pi) **17.1.7**, TUI built on OpenTUI (native Zig frame renderer)

## Steps to reproduce

1. In a wmux PowerShell pane, run interactive `omp` (no args).
2. omp boots (MCP servers connect, session starts) but the pane shows nothing.
3. Wait — the screen never updates.
4. Press Ctrl+C repeatedly; omp finally exits and the drawn UI flashes on teardown.

## Expected

The OpenTUI full-screen UI renders and updates live inside the pane, as it does in a
standalone terminal.

## Actual

Pane is blank/frozen for the entire session while the app is alive the whole time.

## Evidence that wmux's emulator is the stall point

While omp runs, the wmux screen-snapshot path **times

- `terminal_read` (readScreen RPC) → `RPC timeout: inevery call.
- `pane.close` on that pane also frequently times out.
- Meanwhile `terminal_read_events` shows the app is a
  `totalBytesWritten` climbs to ~36,000–58,000 bytes within ~10s.

Since `readScreen` is the same mechanism used to paint the pane, its timeout lines up
exactly with the blank display. `omp -p "..."` (non-ien)
[...]

PR #708, opened 2026-07-30
fix(renderer): stop es2020 lowering from wedging panes that probe DECRQM

openwong2kim:
## The bug

A pane running `opencode` froze on startup — the process stayed alive and kept
updating its title, but the pane never painted again and every MCP read of it
failed with `RPC timeout: input.readScreen (5000ms)`. Panes running Claude Code,
Codex CLI and Kiro CLI in the same workspace were fine. Shrinking `tail_lines`
from 300 to 20 changed nothing, so this was never about read cost.

## Root cause

Two defects, layered.

### 1. The renderer build miscompiled xterm

`vite.renderer.config.ts` set no `build.target`, so Vite's default `modules`
(es2020) applied. Below es2021 esbuild lowers logical assignment (`a ||= b`) to
`a || (a = b)`, and for a binding it has proved dead it then drops the `let a;`
declaration while keeping the write.

`@xterm/xterm` 6's `InputHandler.requestMode` — the DECRQM (`CSI ? Ps $ p`)
handler — is exactly that shape. Upstream `lib/xterm.mjs` is correct:

```js
requestMode(e, i) { let r; (P => (...))(r ||= {}); ... }
```

The shipped bundle was not:

```js
requestMode(t, e) { (y => (...))(void 0 || (i = {})); ... }
//                                        ^ no declaration for `i`
```

Under ESM strict mode that write throws. Captured live from the running
renderer over CDP:

```
ReferenceError: i is not defined
  at kd.requestMode (vendor-xterm-Brb_LsZ_.js:24:103533)
  at wd.parse       (vendor-xterm-Brb_LsZ_.js:24:67927)
stateAtThrow: 5   // ParserState.CSI_INTERMEDIATE — the '$' of `$p`
```

Only opencode tripped it because only opencode probes terminal modes: its
startup capability burst carries six DECRQM requests
(`\e[?1016$p \e[?2027$p \e[?2031$p \e[?1004$p \e[?2004$p \e[?2026$p`).

The throw escapes xterm's `WriteBuffer._innerWrite`, which is what makes it
permanent: `_bufferOffset` never advances and nothing reschedules, so the
Terminal stops parsing for the rest of its life. Measured on the wedged pane:

| | healthy pane | wedged pane |
|---|---|---|
| `_writeBuffer.length` | 0 | 105 |
| `_bufferOffset` | 0 | 0 (nothing consumed) |
| `_pendingData` | 0 | 37912 |
| `parser.currentState` | 0 GROUND | 5 CSI_INTERMEDIATE |

39989 bytes had reached the session; 37912 were still unparsed.

### 2. The read path waited on that buffer forever

`useTerminal.ts` ended hydration with an unbounded parse barrier
(`terminal.write('', resolve)`). On a wedged buffer the callback never comes,
the renderer never replies, and main's 5s IPC deadline
(`_bridge.ts` `DEFAULT_TIMEOUT_MS`) fires. That is the timeout users saw, and
it is why `tail_lines` made no difference — the hang is upstream of the read.
`pane.search` goes through the same barrier and timed out identically.

Renderer-only: the daemon bundles `@xterm/headless` with `esbuild --bundle` and
no minify, so `terminal_read_events` kept working throughout — which is what
made it clear the session itself was healthy.

## Changes

- **`vite.renderer.config.ts`** — pin `target: 'es2022'`. The renderer only ever
  runs in this app's own Chromium (Electron 41 / Chrome 146), so the
  conservative default bought nothing and cost correctness.
- **`src/renderer/terminal/parseBarrier.ts`** (new) — `awaitParseBarrier()`,
  bounded at 1s. Resolves `true` when xterm reports the pane parsed, `false` on
  timeout; a synchronous `write` throw settles as a timeout too.
- **`src/renderer/hooks/useTerminal.ts`** — hydration uses the bounded barrier
  and warns when it expires. A read now returns a slightly stale buffer instead
  of nothing at all, whatever state xterm is in.

Bundle diff, before → after:

```
before: requestMode(t,e){ (y=>(...))(void 0||(i={})); ...
after:  requestMode(t,e){ let i;(b=>(...))(i||={}); ...
```

## Tests

- `src/renderer/terminal/__tests__/parseBarrier.test.ts` — 5 cases, including a
  wedged terminal resolving `false` rather than hanging, and a late callback
  arriving after the timeout.
- `scripts/__tests__/rendererBuildTarget.test.mjs` — 4 cases. Minifies xterm's
  `requestMode` shape at the configured target, imports the result as strict
  ESM and calls it, and asserts es2020 really does produce the broken form so
  the guard cannot quietly stop guarding.
- `useTerminal.hiddenRetention.test.ts` — existing source-level assertion
  updated for the new barrier.

## Verification

| | |
|---|---|
| `tsc --noEmit` | pass |
| `eslint` (changed files) | 0 new errors (3 pre-existing in `useTerminal.ts`, unchanged on this branch) |
| `npm run test:parallel` | 9202 passed, 636 files |
| `npm run test:runtime` | 90 passed |
| `npm run package` | pass; app ships the corrected chunk |

**End-to-end is not automated here.** A harness that boots an isolated instance
(temp HOME + `WMUX_DATA_SUFFIX`), runs `opencode` in the default pane and calls
`input.readScreen` failed three times before reaching the call: the pane's
surface never received a `ptyId` (`SessionManager saveAsync: 0 pty []`). That is
a harness/environment problem, not a signal about this change — nothing was
measured. Debugging it was dropped deliberately. Real-world confirmation will
come from dogfooding a fresh build: launch `opencode` in a pane and check that
`terminal_read` returns content instead of timing out.

## Review note

An automated review of this branch reported `parseBarrier.ts` as missing. It was
untracked at bundle time, so the reviewer saw the import but not the module. It
is included in this commit; `tsc` and the packaged build both resolve it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal screen-read reliability by adding a bounded parse completion wait during input hydration.
  * Prevented pane freezes by stabilizing terminal mode probing with the renderer build target.
  * Avoided indefinite reads by returning safely when parse completion is delayed or a terminal write callback never arrives.
* **New Features**
  * Introduced a parse barrier utility with a short timeout to improve MCP readScreen behavior.
* **Tests**
  * Added regression coverage for parse barrier timeouts and renderer esbuild target behavior.

### DIFF
--- a/changelog.d/708.md
+++ b/changelog.d/708.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- **A pane running a TUI that asks the terminal about its own modes no longer
+  freezes.** `opencode` probes six terminal modes the moment it starts, and each
+  probe hit a miscompiled xterm handler that threw. The throw killed that pane's
+  parser for good: the process kept running and its title kept updating, but the
+  screen never painted again and every agent read of it came back
+  `RPC timeout: input.readScreen (5000ms)` — with nothing to explain why. The
+  renderer bundle is now built at a target that stops mangling the handler.
+  (#708)
+- **Reading a pane can no longer hang.** The read path waited on the pane's
+  parser without a deadline, so a stalled pane took the whole read down with it
+  and the caller got nothing. Reads now wait a bounded moment and return what
+  the pane has, which is worth far more than a timeout. (#708)
--- a/scripts/__tests__/rendererBuildTarget.test.mjs
+++ b/scripts/__tests__/rendererBuildTarget.test.mjs
@@ -1,0 +1,79 @@
+import { describe, it, expect } from 'vitest';
+import { readFileSync } from 'node:fs';
+import { transformSync } from 'esbuild';
+
+/**
+ * Regression guard for the renderer's esbuild target.
+ *
+ * Below es2021 esbuild lowers logical assignment (`a ||= b`) to `a || (a = b)`,
+ * and when it has proved the binding dead it drops the `let a;` declaration
+ * while KEEPING the write — producing `void 0 || (x = {})` against an
+ * undeclared name, which throws under ESM strict mode.
+ *
+ * That miscompiled @xterm/xterm 6's DECRQM handler (`InputHandler.requestMode`,
+ * `CSI ? Ps $ p`). The throw escaped xterm's WriteBuffer drain loop, so the
+ * first TUI to probe terminal modes (opencode does, on startup) permanently
+ * wedged that pane's write buffer: the pane froze and every MCP read of it
+ * (input.readScreen / pane.search) timed out.
+ *
+ * The shape below is xterm's source verbatim in structure — a `let` with no
+ * initializer, fed to a TS-enum IIFE via `||=`, then never read again.
+ */
+const XTERM_REQUEST_MODE_SHAPE = `
+export function requestMode(params, ansi) {
+  let DecRqmTypes;
+  ((v) => (v[v.NOT_RECOGNIZED = 0] = 'NOT_RECOGNIZED', v[v.SET = 1] = 'SET'))(DecRqmTypes ||= {});
+  const modes = { origin: true };
+  const reply = (m, s) => String(m) + ';' + String(s);
+  return ansi ? reply(params, 0) : reply(params, modes.origin ? 1 : 2);
+}
+`;
+
+function rendererTarget() {
+  const src = readFileSync(new URL('../../vite.renderer.config.ts', import.meta.url), 'utf8');
+  const match = /^\s*target:\s*'([^']+)'/m.exec(src);
+  return match?.[1];
+}
+
+describe('renderer build target', () => {
+  it('is pinned in vite.renderer.config.ts', () => {
+    expect(rendererTarget()).toBeTruthy();
+  });
+
+  it('does not lower logical assignment (the es2020 miscompile)', () => {
+    const { code } = transformSync(XTERM_REQUEST_MODE_SHAPE, {
+      loader: 'js',
+      format: 'esm',
+      minify: true,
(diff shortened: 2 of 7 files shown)


openwong2kim:
@codex review

openwong2kim:

This PR fixes a pane-freeze bug where any TUI that probes terminal modes on startup (opencode sends six DECRQM requests) permanently wedged the pane: the process kept running but the screen never painted again, and every `input.readScreen` / `pane.search` against it timed out with `RPC timeout: input.readScreen (5000ms)`. The root cause is excellent forensics — Vite's default `modules` target (es2020) made esbuild lower `a ||= b` to `a || (a = b)` and drop the dead `let` declaration while keeping the write, so `@xterm/xterm` 6's `InputHandler.requestMode` shipped as `void 0 || (i = {})` against an undeclared name, throwing `ReferenceError` under ESM strict mode the first time a mode probe hit it.

The fix is minimal and correct: pin the renderer build to `es2022` (the renderer only runs in this app's own Chromium, so the conservative target bought nothing), and add a bounded parse barrier (`parseBarrier.ts`, 1s timeout) so a wedged xterm write buffer can never hang a read again — reads now return a slightly stale buffer instead of timing out and returning nothing. The regression-guard test (`rendererBuildTarget.test.mjs`) is particularly well done: it proves the es2020 default really does produce the broken form, so the guard cannot quietly stop guarding. This is the same render-stall family as #691 (OpenTUI apps rendering blank) and #678, and it likely closes or substantially advances those.

**Priority: P1** — a complete, well-tested fix for a severe user-facing freeze (terminal panes are wmux's core function), with ready-to-merge code and strong test coverage including the wedged-buffer path that was the actual failure mode.

PR #710, opened 2026-07-30
Default the orchestrator to the terminal brain

openwong2kim:
The Command Deck now starts on the terminal brain — it drives the user's own
`claude` binary, so it needs no API key and re-reads its per-workspace
`CLAUDE.md` from disk on every spawn. Nothing is removed: the SDK brain stays
selectable in Settings and remains the automatic fallback when daemon mode is
unavailable. Only the default moves.

## Making the default actually reach existing installs

The first pass looked correct and did nothing for anyone who already had wmux
installed. `AppLayout` has always serialized `deckBrainVendor` unconditionally,
so "absent" describes no session on disk — every profile predating this carries
a literal `'claude'`, whether its user deliberately picked the SDK brain or
never opened Settings. Reading that value as a choice pins the whole existing
base to the SDK brain and quietly narrows the change to new profiles.

The old schema simply never recorded intent, so a one-shot
`deckBrainVendorMigrated` marker now carries it. Before the marker, `'claude'`
is read as the old default and upgraded once (`'hermes'`/`'claude-pty'` were
only ever reachable by an explicit pick, so they survive). After it, every
recorded vendor is authoritative — which is what lets someone pick the SDK
brain back and keep it across loads.

The upgrade is non-destructive. Commander sessions are keyed per vendor and
nothing is cleared, so switching back in Settings resumes the exact SDK
conversation the migration stepped away from. The unavoidable cost: a user who
deliberately chose the SDK brain before this ships is migrated once too, because
no data distinguishes them.

## The requested vendor is not always the one that runs

The terminal brain's pty *is* a daemon session, so with no daemon
`createAdapter` hands back an SDK brain — while every derived value still said
`'claude-pty'`. That wrote SDK session ids under the `::claude-pty` key (leaving
the real terminal conversation unresumable), told a brain that has `Write` it
had no durable memory, forwarded full power to an SDK adapter the UI had just
disabled, and tagged the fallback manager `'claude-pty'` so the daemon coming
back could never dislodge it — the swap check was comparing two *requests*.
Resolving the effective vendor once, in `ensureManager`, fixes all four.

## Controls that were inert on the new default path

- **Model picker.** It rode only on the `DECK_SEND` payload, so it reached main
  exactly when a human typed into the deck composer — and the terminal brain's
  layout has no composer (its TUI is the input path). Automation-driven turns
  spawned brains on whatever the last typed turn left cached. It is now
  main-authoritative like full power and the vendor.
- **Full power.** It tunes `settingSources`/`canUseTool`, which an interactive
  session has no equivalent for, so its toggle is now inert with a reason under
  the other vendors instead of silently doing nothing when clicked.
- **Vendor coercion.** `DECK_BRAIN_VENDOR_SET` still failed closed to
  `'claude'`, splitting main and the renderer onto different vendors — and
  different session keys — for one bad input.

## Testing

19 new cases: the migration matrix (old default upgraded, explicit picks kept,
marker set and honoured, unknown ids), effective-vendor resolution (no-daemon
downgrade carries the key and memory policy with it; promotion once the daemon
returns), and the model authority (picker applies through the composer-less
wake path, stale-model retirement, ACP left alone, hostile strings sanitized).

Two existing tests were leaning on the old default implicitly and one was
locking in the old coercion; those now state their vendor explicitly.

`tsc` clean, 9251 unit + 90 runtime tests pass.

## Summary by CodeRabbit

* **New Features**
  * The terminal-based brain is now the default orchestrator, using the local Claude subscription (no API key required).
  * Added syncing for the orchestrator “model” setting from the UI to the orchestrator, including for scheduled/wake turns.
  * Orchestrator vendor/model selection now updates more reliably across app restarts via an automatic one-time migration (preserving history).

* **Bug Fixes**
  * Improved fallback behavior when daemon mode isn’t available, preventing terminal conversation loss for workspaces started without it.
  * Model/vendor selection and conversation clearing now target the correct persisted workspace/session.
  * Full-power mode UI now shows clearer availability and disables itself when unsupported.

### DIFF
--- a/changelog.d/710.md
+++ b/changelog.d/710.md
@@ -1,0 +1,29 @@
+### Changed
+
+- **The orchestrator now runs on the terminal brain by default.** The Command
+  Deck drives your own `claude` binary in an embedded terminal instead of the
+  Agent SDK, so it needs no API key and picks up its per-workspace `CLAUDE.md`
+  from disk every time it starts. The SDK brain has not gone anywhere: it stays
+  selectable under Settings › Orchestrator, and wmux still falls back to it
+  automatically when daemon mode is unavailable. Existing installs move over on
+  the next launch. The move is reversible and costs no history — each brain
+  keeps its own conversation, so switching back in Settings picks your SDK
+  conversation up exactly where it was. (#710)
+
+### Fixed
+
+- **The orchestrator model picker now applies to the terminal brain.** The
+  model only ever reached the orchestrator when you typed into the deck
+  composer, and the terminal brain has no composer — you type into its terminal.
+  Choosing a model therefore did nothing for it, and scheduled or event-driven
+  turns ran on whatever model your last typed message happened to leave behind.
+  The picker now applies on every path. (#710)
+- **A workspace that started without a running daemon no longer loses its
+  terminal conversation.** With no daemon the deck quietly serves the SDK brain
+  instead, but it filed that conversation under the terminal brain's name. The
+  terminal conversation could not be resumed afterwards, the substitute brain
+  was told it had no memory to write to, and the workspace stayed on the
+  substitute even after the daemon came back. (#710)
+- **Full-power mode no longer looks available when it cannot do anything.** It
+  only applies to the SDK brain, so under the terminal and Hermes brains the
+  toggle now explains that instead of silently ignoring the click. (#710)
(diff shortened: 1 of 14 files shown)


openwong2kim:
@codex review

PR #711, opened 2026-07-30
Add a New session control to the orchestrator bar

openwong2kim:
Replacing a workspace's orchestrator with a fresh one had no affordance — only
a complete main-process path that nothing called. `DECK_CONVERSATION_CLEAR`,
its preload binding and its store tests all shipped some time ago; a grep for
callers in `src/renderer` returned nothing. This is the button.

## Why it earns its place

None of the obvious substitutes actually work:

- **Restarting wmux doesn't.** The commander session id is persisted precisely
  so the next launch resumes the same conversation.
- **Deleting the workspace does** — and takes its entire pane tree, its
  identity across channels and A2A with it. That is a terrible exchange rate
  for replacing one brain.
- **Quitting the terminal brain's TUI doesn't either.** The next turn respawns
  with `--resume`, so you get a new *process* on the *same* conversation. It
  looks exactly like a reset and isn't, which is the specific confusion this
  control removes.

## Naming

"Reset" has no object here — the deck also has a loop, schedules, a mode, a
transcript and panes, all resettable in principle — and `restart` already means
a **pane** restart in wmux's vocabulary (`pane.restarted`, supervision
`armed`/`stopped`). Either name invites "did it restart my panes?". "New
session" says what it produces, matches what the code does (a fresh adapter
with no `resume`), and is the word Claude Code users already have.

## What survives

Everything except the conversation: panes, surfaces, worktrees, the
`#commander` transcript and report rail, durable memory, `deck-policy.md`, the
autonomy mode, loops, schedules and any pending decision. The tooltip says so —
"New session" is otherwise easy to read as "history gone".

## Clear is paired with a wake

Not left to the operator. The terminal brain is the default now, and clearing
kills its embedded pty, which retracts the terminal and drops the deck out of
the pty layout — where no composer exists. Waking immediately means the promise
the button makes is what actually appears: a fresh orchestrator that has
already re-read `pane_list`, rather than an empty dock. A *failed* clear is
deliberately not woken, since that would revive the conversation the operator
just asked to leave.

## Confirmation and placement

The repo's two-step armed chip, not a modal. The action destroys no artifact,
but it is one-way and can interrupt a live turn — exactly the weight that
pattern carries (`ChannelItem` documents it as the house style for one-way
actions; `window.confirm` is reserved for filesystem destruction like worktree
removal). Red only once armed, per DESIGN.md reserving solid red for the final
confirm; neutral at rest like the other AI-directed controls on the bar.

It sits last among the always-on controls in `renderControlBar`, after
Schedules and before the ephemeral recovery chip — so in the pty layout it
lands beside Wake and the two brain-lifecycle buttons sit together.

It stays **enabled while a turn streams**. A brain stuck in a bad turn is the
most common reason to want a fresh one, so disabling it exactly then would be
backwards; the wording changes to say the click interrupts.

## Testing

10 cases: first click arms and does not fire, second commits, mouse-leave
disarms, a workspace switch disarms (arming intent must not follow the operator
to a different orchestrator), clear→wake ordering, no wake after a failed
clear, a rejected wake still leaves the chip usable, and the busy behaviour
(enabled, wording swaps).

## Summary by CodeRabbit

* **New Features**
  * Added a **New Session** control to the workspace command deck.
  * Uses an “arm then confirm” flow to replace the current orchestrator with a fresh session.
  * Provides busy-state wording and keeps the control available during turns.
  * Starts the new session while preserving panes/worktrees and continuing loops/schedules.
  * Added English and Korean translations for the new control and tooltip.
* **Tests**
  * Added dynamic UI tests covering the two-step confirmation and clear-then-wake behavior.

### DIFF
--- a/changelog.d/711.md
+++ b/changelog.d/711.md
@@ -1,0 +1,14 @@
+### Added
+
+- **You can hand a workspace a fresh orchestrator without losing anything
+  else.** A **New session** control now sits with the other orchestrator
+  controls at the top of the deck. It retires the current orchestrator and
+  starts a new one that reads your project files again and takes a fresh look
+  at your panes, which is what you want once a long conversation has drifted or
+  a turn has wedged. There was no way to do this before: restarting wmux
+  deliberately resumes the same conversation, quitting the orchestrator's own
+  terminal just reopens it on that conversation, and deleting the workspace
+  takes every pane with it. Your panes, worktrees and conversation history all
+  stay, and loops and schedules keep running. It asks for a second click before
+  it commits, and stays available even mid-turn — a stuck turn is usually
+  exactly when you need it. (#711)
--- a/src/renderer/components/Deck/CommanderView.tsx
+++ b/src/renderer/components/Deck/CommanderView.tsx
@@ -66,6 +66,7 @@ import {
 import { buildQuickActions, type DeckQuickAction } from './deckQuickActions';
 import { renderBrainMarkdown } from './BrainMarkdown';
 import { DeckSchedulesPanel } from './DeckSchedulesPanel';
+import { NewSessionChipContainer } from './NewSessionChip';
 import { DeckLoopPanel } from './DeckLoopPanel';
 import { DeckDecisionCard } from './DeckDecisionCard';
 import BrainTerminalEmbed from './BrainTerminalEmbed';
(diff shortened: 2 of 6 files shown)


openwong2kim:

This PR adds a **"New session" control** to the workspace orchestrator bar — a two-step (arm then confirm) chip that retires the live orchestrator and starts a fresh one that re-reads the project with no stale conversation assumptions. It finally gives the deck the affordance the already-shipped DECK_CONVERSATION_CLEAR IPC path never had a caller for, which is the notable part: the main-process plumbing existed, nothing in the renderer reached it, so the gap was purely the missing button.

Scope is tightly contained — the diff is NewSessionChip.tsx (192 lines) wired into CommanderView.tsx next to Schedules, plus en/ko i18n strings, a changelog fragment, and 10 dynamic React tests covering arm/commit, disarm on mouse-leave, workspace-switch disarm, clear->wake ordering, no-wake-after-failed-clear, rejected-wake resilience, and the busy wording swap. Nothing outside the conversation is touched: panes, worktrees, the #commander transcript, durable memory, policy, mode, loops and schedules all survive, which the design comments and tooltip spell out explicitly. The decision to keep the chip enabled mid-turn (with swapped wording) is the right call — a wedged turn is exactly when you want it.

One thing worth flagging for dogfood: clear->wake is sequential and the wake is best-effort (a rejected/busy wake is swallowed, leaving a fresh but idle brain). That is the safe failure mode, but on a busy daemon it can leave the operator staring at an empty dock until they hit Wake themselves — worth confirming the deck surfaces that state clearly enough that it doesn't read as "the button did nothing." Otherwise this is clean, well-tested, and ready.

PR #712, opened 2026-07-31
Add explicit approval choices for phone clients

openwong2kim:
## Summary
- extract structured Claude `AskUserQuestion` choices while preserving each option's real TUI digit
- accept `choiceKey` on phone approval resolves, validate it fail-closed, re-verify the option on screen, and persist the selected key
- suppress ambiguous lock-screen affirmative actions when an approval requires an in-app choice
- document explicit-choice resolution and negotiated alternate-screen phone scrolling

## Safety and compatibility
- malformed, empty, or deny-side `choiceKey` is rejected before the registry is called
- unknown or stale keys remain pending and send no PTY bytes
- omitting `choiceKey` preserves legacy behavior exactly: approve sends `1`, deny sends `ESC`, neither sends CR
- unsupported agents remain terminal-only; no approval keys are guessed

## Validation
- `git diff --check origin/main...HEAD`
- approval/web/push focused suite: **319/319 passed**
- `npm run build:daemon` (including daemon web CSP gate)

## Summary by CodeRabbit

- **New Features**
  - Approval requests can now present structured, numbered choices for selection.
  - Selected choices are saved and shown in approval history.
  - Push notifications can indicate when selection must be completed in-app.
  - Phone terminal scrolling now supports local scrollback and alternate-screen paging or mouse-wheel behavior.

- **Bug Fixes**
  - Invalid, unavailable, or invisible choice selections are rejected with clear errors.
  - Approval choice data is validated and safely preserved across restarts.
  - Choice labels and keys are sanitized for reliable display and selection.

### DIFF
--- a/docs/phone-client-contract.md
+++ b/docs/phone-client-contract.md
@@ -252,6 +252,18 @@ Check `/api/config` and hide the keyboard rather than letting a user type into a
 403. `fetch` resolves on 401 and 403 — a lone `.catch()` sees neither, which is a
 mistake the browser client made and shipped.
 
+Phone scrolling has two ownership modes. A terminal's normal buffer is local
+scrollback and must remain local (ordinary shells and Kiro use this path).
+Alternate-screen TUIs have no terminal scrollback. With `--allow-input`, a
+vertical phone drag may send line-granular wheel events only while the remote
+TUI has negotiated mouse reporting, and only encoded with that negotiated
+protocol. If no mouse mode is active and no wheel event was sent, the completed
+swipe may fall back to standard `PgUp`/`PgDn` terminal navigation. Claude Code's
+fullscreen renderer documents both wheel scrolling and those keys for its
+app-owned conversation history. Without `--allow-input`, send neither; never
+forward generic taps or drags, guess a mouse protocol, or pretend the alternate
+buffer is locally scrollable.
+
 ### Creating and closing panes
 
 ```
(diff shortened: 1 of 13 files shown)


openwong2kim:
@codex review

openwong2kim:

This PR adds structured per-option choice resolution for phone/web approval of Claude Code `AskUserQuestion` prompts. Until now, an approve could only send the first-option digit (`1`), so a phone user picking "option 2 of 3" had no way to express that remotely. The PR extracts a `choices: Array<{key, label}>` alongside the existing `options` array, where each `key` is the 1-based digit that selects that option in the TUI — and critically, the key preserves the original index even when unlabeled entries are dropped, so the digit sent matches what Claude's TUI actually expects. Clients opt in by sending `choiceKey` on the resolve call; omitting it preserves byte-for-byte legacy behaviour.

The safety design is the strong part. Every malformed, empty, deny-side, or unknown `choiceKey` fails closed with `invalid-choice-key` (400/422) before any PTY bytes are written, and the request stays pending rather than being silently answered with the wrong option. After validating the key against the stored choices, the registry re-reads the screen and confirms the specific option row is still visible (`looksLikeChoiceOnScreen`) before typing its digit — so a prompt that has re-rendered with reordered options is refused rather than mis-answered. The selected key is persisted on the resolved history record, and the push payload gains `requiresInAppChoice` so the Notification Service Extension can suppress ambiguous lock-screen affirmative actions when a structured choice must be made in-app. The phone-client contract docs document the new field, the wheel/PgUp scroll negotiation for alternate-screen TUIs, and the per-agent support matrix.

The contract is additive end to end — types, store sanitiser, HTTP and pipe RPC routes, push envelope — and the test suite covers deep-copy isolation, valid/invalid/stale keys, the screen re-verify refusal path, persistence, and the dual HTTP/RPC entry validation. The "fail closed, leave pending" stance is the right default for a keystroke that can't be un-typed. Priority: **P2** — it removes a meaningful blind-first-option limitation on the remote approval surface that the codebase had explicitly flagged as shipped-deliberately-but-unfinished.

PR #713, opened 2026-07-31
fix(security): make CDP remote-debugging port closeable via config (#613)

### LINKED ISSUE #613: security: Electron CDP remote-debugging port on by default — lowers same-user attack bar
## Finding

`src/main/index.ts:130-137` enables Electron's CDP (Chrome DevTools Protocol) remote-debugging port by default on every boot:

```ts
let cdpPort = 0;
if (process.env.WMUX_DISABLE_CDP !== 'true') {
  const basePort = 18800;
  const range = 100;
  cdpPort = basePort + crypto.randomInt(range);
  app.commandLine.appendSwitch('remote-debugging-port', cdpPort.toString());
  console.log(`[WinMux] CDP enabled on port ${cdpPort}`);
}
```

The port is randomized in `[18800, 18900)`. Randomization is presented as a boundary; **it is not** — a same-user process can read `DevToolsActivePort` or scan 100 loopback ports in milliseconds.

## Why this is tracked separately from other same-user vectors

`docs/SECURITY.md §3` declares same-user attacks out of scope. Most same-user vectors accept that framing honestly — they require **filesystem access** (to read `~/.wmux-auth-token`) or **process control** (to attach a debugger). CDP is different: it only requires **loopback socket access**, which is a lower bar.

| Vector | Required access |
|---|---|
| Read `~/.wmux-auth-token` | Filesystem |
| Attach debugger to daemon/main | Process control |
| **CDP** | **Loopback socket** — no file or process access needed |

Once an attacker reaches CDP, they get full renderer control and can drive every privileged IPC handler the preload exposes — `pty:create`, `pty:write`, `fs:writeFile`, `shell:openPath`, `rpc:invoke`. That is every defense in one path: token ACL, shell allow-list, SSRF guard, sensitive-path blocklist all bypassed.

Partial-compromise scenarios where this matters:
- Sandboxed renderer escape with loopback fetch but no filesystem
- Malicious npm package / editor extension / other Electron app that can hit loopback
- Container with shared network namespace but separated filesystem

Severity is **MEDIUM** (no remote path, no privilege escalation beyond same-user). Not P1. Tracked as P2 because the fix is cheap (~1h) and it is the single largest distance-closer within the same-user ceiling.

## Recommendation (priority order)

| Option | Effort | Trade-off |
|---|---|---|
| **A. Switch to `--remote-debugging-pipe`** | ~1h | Eliminates the port entirely. CDP runs over a pipe fd. Playwright `connectOverCDP` supports pipe transport — update `src/mcp/playwright/PlaywrightEngine.ts:791` and `src/main/browser-session/WebviewCdpManager.ts:179` accordingly. Chromium's recommended posture. **Preferred.** |
| B. Default off, opt-in via `--enable-cdp` flag | ~30min | Users who do not use browser automation get protection by default. Smallest diff. |
| C. CDP handshake token | ~2h | Intercept `/json/list` to require a token. Keeps the port, adds auth. More surface to maintain. |

## Verification (post-fix)

- [ ] `npm run lint && npm run test:parallel` passes
- [ ] Browser automation still works (`src/mcp/playwright/PlaywrightEngine.ts:283` `connectOverCDP`)
- [ ] Add a regression test asserting no `--remote-debugging-port` switch is appended when the opt-out path is taken (or that pipe transport is used, under option A)

## References

- Full analysis: [`docs/security-audit-2026-07-25.md §2`](https://github.com/openwong2kim/wmux/blob/main/docs/security-audit-2026-07-25.md#2-p0--close-the-cdp-shortcut) (lands in #617)
- Threat model: [`docs/SECURITY.md §3`](https://github.com/openwong2kim/wmux/blob/main/docs/SECURITY.md#3-what-wmux-does-not-try-to-protect-against)
- Rest of the same audit: #615 (F2-F6)
- Related: #607 (W1 cleartext), #608 (W4 CSP) — earlier review batch

openwong2kim:
## Summary

Closes #613.

Electron's CDP remote-debugging port was enabled on every boot, with only a `WMUX_DISABLE_CDP` env var to turn it off. CDP is the single largest same-user attack surface: unlike other same-user vectors (filesystem, process control), it only needs a **loopback socket** to reach full renderer control and drive every privileged IPC handler the preload exposes.

This PR makes the port closeable from `~/.wmux/config.json` and documents the threat model in SECURITY.md.

## Changes

- **`browser.cdp.enabled` config option** (`~/.wmux/config.json`). Defaults to `true` (browser automation depends on CDP). Set `false` to close the port. Coerced per-field like `lanlink`/`channels` — absent or malformed degrades to the default, never silently disabling automation.
- **`WMUX_DISABLE_CDP=true`** env var still force-disables regardless of config (unchanged behaviour, now also checked).
- **`browser.cdp.info`** now throws an actionable error when CDP is off, instead of returning port 0 and confusing clients with connection-refused.
- **SECURITY.md §3** documents the CDP same-user vector and both opt-out paths.
- **6 regression tests** in `config.test.ts` (old-config backfill, garbage handling, explicit `false` honoured).

## Latent bug fixed alongside

The CDP decision in `index.ts` previously ran **before** the `WMUX_DATA_SUFFIX` block, so a dev build (`-dev`) would read the **production** config (`~/.wmux`) for the CDP gate. Moved the decision after the suffix block so each instance reads its own config.

## What this does NOT do

- Does **not** change the default (`true`). Flipping the default to `false` would break browser automation for every existing install and is a product decision, not a security patch — tracked separately in #613's option B discussion.
- Does **not** switch to `--remote-debugging-pipe` (option A). That eliminates the port entirely but requires reworking the HTTP-based CDP clients (`PlaywrightEngine.connectOverCDP`, `WebviewCdpManager` `/json` fetch) onto pipe transport — a much larger change.

## Verification

- `tsc --noEmit` clean (excluding pre-existing unrelated errors)
- `config.test.ts` 34, `WebviewCdpManager` 42, `browser.rpc` 44, `browser.rpc.backend` 30 — all green (150 total)

## Checklist

- [x] PR does not bump `package.json` version
- [x] PR does not edit `CHANGELOG.md` (fragment added after PR number is known)
- [x] No new Korean comments added (repo is English-only)


openwong2kim:

This PR closes #613 by making Electron's CDP remote-debugging port closeable from `~/.wmux/config.json` via a new `browser.cdp.enabled` field (defaulting to `true` to keep browser automation working), while preserving the existing `WMUX_DISABLE_CDP=true` env escape hatch. It adds a per-field backfill in `coerceBrowserCdp` that mirrors the `lanlink`/`channels` discipline — an absent or malformed slice degrades to the default rather than triggering a whole-file reset, and an explicit `false` is honoured verbatim. When the port is off, `browser.cdp.info` now throws an actionable error instead of returning port 0 and confusing CDP clients with an opaque connection-refused.

Alongside the security work it fixes a real latent ordering bug: the CDP decision in `index.ts` previously ran *before* the `WMUX_DATA_SUFFIX` block, so a `-dev` build consulted the production config (`~/.wmux`) for the CDP gate. Moving the decision after the suffix block ensures each instance reads its own config. The SECURITY.md §3 update documents the same-user CDP vector and both opt-out paths, and the six regression tests cover old-config backfill, garbage handling, array-shaped values, non-boolean `enabled`, and the explicit `false` opt-out path.

This is the single largest distance-closer within the same-user ceiling identified by the 2026-07-25 audit, and the change is deliberately scoped: it does not flip the default to `false` (which would break browser automation for every existing install) nor switch to `--remote-debugging-pipe`. The scope discipline and the config-coercion parity with existing slices make this safe to merge. Priority: **P2** — it resolves a tracked security-hardening item at low risk.

PR #715, opened 2026-07-31
feat(worktask): detach a dependent child task from its parent

openwong2kim:
## What

Adds **detach**: releasing a dependent child WorkTask (a sub-workspace) from its parent **without touching the child's workspace, worktree, branch, PTY, or running agent**. The child keeps running — just independently.

A **"Detach from parent"** action appears in the sidebar context menu of a dependent child workspace.

## How

Rather than a new event that an old daemon could silently ignore (which would risk resurrecting the parent–child dependency), detach rides the existing `task.close`:

- The close record gains **`workspace-detached` evidence** carrying a `detachedAt` timestamp.
- An **old daemon** still reads the task as `closed`, so it never re-attaches the child.
- A **new daemon** distinguishes detach from a plain harvest close via `detachedAt`.
- The mission channel is **archived** (records preserved), the same as an ordinary close.

Detach uses the mission RPC path (`task.mission.close` with `detach: true`), which by design never removes the worktree — so the "hands off the worktree/branch/PTY" contract holds structurally.

## Closed-state consumers made `detachedAt`-aware

Because detach injects a *live-but-closed* state, every consumer that keys on open/closed is updated so a live detached worktree is never misread:

- **Cleanup scanner** excludes a live detached worktree — never offered as an `orphan-dir` (deletion candidate).
- **Closed-projection GC** exempts detached tasks, so the signal stays durable and the scanner keeps protecting the worktree.
- **Re-detach / already-detached** is an idempotent success instead of a false failure (no misleading toast; replay receipts carry `detached`).

## Tests

Adds regression coverage for: detach stamps `detachedAt` + returns `detached:true`, replay restores it across restart, GC exempts detached (but still evicts a plain-closed task), re-detach idempotency, and the scanner protection (live detached worktree excluded; the non-detached counterpart still surfaces as `preserved`). Full affected suite green; `tsc` (base + daemon) clean.

## Notes

- Per repo policy: no version bump, no `CHANGELOG.md` edit (a `changelog.d/` fragment will be added once this PR has a number).
- Accepted trade-off: the mission channel is archived on detach (design decision), so a still-live child writing to that channel will fail — surfaced during review, kept intentionally.

### DIFF
--- a/changelog.d/715.md
+++ b/changelog.d/715.md
@@ -1,0 +1,13 @@
+### Added
+
+- **You can set a dependent child task loose without stopping it.** A **Detach
+  from parent** action now sits in the context menu of a child workspace that a
+  fan-out spun up. Detaching releases that task from its parent — the parent
+  stops tracking it and its mission channel folds away into Archived — while the
+  workspace itself, its worktree, branch, terminals and running agent all keep
+  going, now on their own. Before this, a child task was tied to its parent for
+  its whole life: the only way to end the relationship was to close it, which
+  tears down the worktree and the work along with it. Detach keeps everything
+  and simply cuts the leash. The cleanup scanner knows a detached worktree is
+  still live, so it is never offered up for deletion, and detaching something
+  twice is harmless. (#715)
--- a/src/daemon/index.ts
+++ b/src/daemon/index.ts
@@ -3250,6 +3250,10 @@ function registerRpcHandlers(
       taskId,
       verifiedWorkspaceId,
       ...(typeof p['idempotencyKey'] === 'string' ? { idempotencyKey: p['idempotencyKey'] } : {}),
+      // Non-destructive detach close (frees the child from its parent). This RPC
+      // path never touches worktree/branch/PTY in the first place, so detach only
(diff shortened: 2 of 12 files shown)


openwong2kim:

This PR adds a `detach` operation that releases a dependent child WorkTask (sub-workspace) from its parent without touching the child's workspace, worktree, branch, PTY, or running agent — the child keeps running independently. The implementation is thoughtfully conservative: detach rides the existing `task.close` with new `workspace-detached` evidence and a `detachedAt` timestamp, so an old daemon still reads the task as closed (never re-attaching the child) while a new daemon distinguishes detach from a plain harvest close, and the mission channel is archived. Every closed-state consumer is made `detachedAt`-aware so a live detached worktree is never offered as a deletion candidate or evicted by closed-projection GC, and re-detach is idempotent. The regression coverage (detach stamping, replay across restart, GC exemption, idempotency, scanner protection) is solid. P3: a clean new workflow feature with correct lifecycle handling; not higher because it is additive functionality rather than a fix, and the archived-mission-channel trade-off on a still-live child is explicitly accepted.

PR #716, opened 2026-07-31
Detect Kiro CLI sessions

openwong2kim:
Kiro CLI panes were invisible to wmux: no agent name, no status, no notifications.
Kiro ships no hook bridge, so its identity has to come from the PTY stream.

## What changed
`'kiro'` joins the `AgentSlug` union in all five authoritative places (shared
events, hook signal types, orchestrator stems, the renderer candidate seed, and
the daemon allowlist), and `AgentDetector` learns a **compound** gate.

A product-name mention is not enough — agents routinely print logs and docs that
name other agents. Activation requires BOTH from the same PTY:

- anchored chrome: the `Kiro CLI vX` banner, the trust-mode footer, or the v3
  `kiro.dev/docs/cli/` line
- the anchored composer placeholder (`ask a question or describe a task`)

Evidence may arrive in either order; when the composer lands first its `waiting`
state is replayed once so a genuinely idle pane is not stuck reporting `running`.
The gate loop now also skips regexes for already-active agents, which makes this
cheaper than the loop it replaces on full-screen repaint traffic.

## Tests
9 new cases in `AgentDetector.test.ts`: both-required, either order, docs-URL and
trust-footer variants, mention-only rejection, no-steal from an active Claude
pane, per-detector evidence isolation, and slug/display round-trip.

`npx vitest run src/main/pty/__tests__/AgentDetector.test.ts src/renderer/channels/__tests__/agentCandidateSeed.test.ts integrations/shared/__tests__/signal-types.test.ts` → 66 passed
`npm run build:daemon` → clean

## Notes
Detection only. No approval route is added for Kiro (it has no hook surface), so
Kiro panes get identity, status and notifications but not phone approvals.

### DIFF
--- a/src/daemon/index.ts
+++ b/src/daemon/index.ts
@@ -340,7 +340,7 @@ const recoveredResumeBindings = new Map<string, ResumeBinding>();
 // in sync with AgentSlug in src/shared/events.ts and ALLOWED_AGENT_SLUGS in
 // integrations/shared/signal-types.ts.
 const KNOWN_AGENT_SLUGS: ReadonlySet<string> = new Set([
-  'claude', 'codex', 'gemini', 'aider', 'opencode', 'copilot', 'openclaude',
+  'claude', 'codex', 'gemini', 'aider', 'opencode', 'copilot', 'openclaude', 'kiro',
 ]);
 
 // === Constants ===
--- a/src/main/pipe/handlers/hooks.rpc.ts
+++ b/src/main/pipe/handlers/hooks.rpc.ts
@@ -1274,5 +1274,6 @@ function agentDisplayName(slug: AgentSignal['agent']): string {
     case 'opencode': return 'OpenCode';
     case 'copilot': return 'GitHub Copilot CLI';
     case 'openclaude': return 'OpenClaude';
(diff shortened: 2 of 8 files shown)


openwong2kim:

openwong2kim:

This PR adds detection for Kiro CLI panes, which were previously invisible to wmux (no agent name, status, or notifications) because Kiro ships no hook bridge and its identity must come from the PTY stream. The design is careful: a single product-name mention is not enough, since agents routinely print logs and docs naming other agents — activation requires both an anchored chrome line (the `Kiro CLI vX` banner, the trust-mode footer, or the v3 docs URL) and the anchored composer placeholder from the same PTY, with evidence accepted in either order and the composer `waiting` state replayed once if it lands first. The gate loop also skips regexes for already-active agents, making this cheaper than before on full-screen repaint traffic. Nine new AgentDetector cases cover both-required, either order, mention-only rejection, no-steal from an active Claude pane, and per-detector evidence isolation. P3: detection-only, no approval route is added (Kiro has no hook surface), and the compound gate is well-motivated and tested; nice-to-have agent support rather than a fix for an active problem.

PR #717, opened 2026-07-31
Clean up transient PTY state on close

openwong2kim:
`surfacePorts` and `surfaceAgentStatus` were the last two transient per-ptyId maps
not cleaned up when a pane or surface closed. Every close left a dead entry
behind, and because ptyIds are reusable a new surface could inherit the previous
one's status — reading as blocked or running from birth.

Found in the fleet-activity adversarial review, which flagged `surfacePorts` as
unsafe to copy as a cleanup precedent because it leaked itself.

## What changed
Both maps are now evicted at the two real teardown sites, matching the existing
`surfaceAgent` / `surfaceActivity` / `surfacePendingQuestion` pattern:

- `paneSlice.closePane` — for every surface under the closed subtree
- `surfaceSlice.closeSurface` — for the closed surface

## Tests
`paneSlice.test.ts`: subtree teardown clears both maps and leaves surfaces outside
the subtree untouched.
`surfaceSlice.test.ts`: both maps cleared, siblings preserved, and a reused ptyId
starts with no inherited status.

Verified the tests fail against the unfixed slices (4 failures) and pass after:
`npx vitest run src/renderer/stores/slices/__tests__/` → 118 passed

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -339,10 +339,9 @@
 - **Depends on:** fleet-activity-line ship 후.
 - **Priority:** P2
 
-## (bug) transient per-ptyId 맵이 surface close 시 leak (P3, found in fleet-activity adversarial review)
-- **What:** `surfacePorts`(및 `surfaceAgentStatus`)가 `closePane`/`closeSurface`에서 정리 안 됨 → 죽은 ptyId 엔트리 잔존. fleet-activity PR이 `surfaceActivity`는 양 사이트에서 정리하지만 기존 두 맵은 미수정.
-- **Why:** 적대 리뷰가 "surfacePorts를 cleanup 선례로 쓰지 말라(그 자체로 leak)"며 발견. 장기 세션서 죽은 ptyId 누적.
-- **Pros:** store 위생, 미세 메모리.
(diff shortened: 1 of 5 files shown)


openwong2kim:

openwong2kim:

This PR fixes a transient-state leak found in the fleet-activity adversarial review: `surfacePorts` and `surfaceAgentStatus` were the last two per-ptyId maps not cleaned up when a pane or surface closed, so every close left a dead entry behind — and because ptyIds are reusable, a new surface could inherit the previous one's status and read as blocked or running from birth. The fix evicts both maps at the two real teardown sites (`paneSlice.closePane` for every surface under a closed subtree, and `surfaceSlice.closeSurface` for the closed surface), matching the existing `surfaceAgent`/`surfaceActivity`/`surfacePendingQuestion` pattern. The tests verify subtree teardown clears both maps while leaving outside surfaces untouched, both maps cleared with siblings preserved, and a reused ptyId starts with no inherited status (4 failures against the unfixed slices). P3: a correct store-hygiene fix with a real (if low-severity) inherited-status consequence; small and well-contained.

PR #719, opened 2026-07-31
Enforce workspace scoping on status and events

openwong2kim:
Two workspace-isolation holes from the scoping survey, plus the verdict on the
three items it left unverified.

## U8 — `meta.setStatus` / `meta.setProgress` (REAL)
Both RPCs trusted a caller-supplied `workspaceId`, so any agent or MCP client
could write status/progress into a workspace it has no relationship with.

Now the workspace is **server-resolved** from `senderPtyId` (the same pattern as
`fanout.rpc` / `events.rpc`) and the caller-supplied value is ignored for
external callers. First-party callers (the renderer bridge) keep supplying it —
the renderer legitimately chooses which workspace to target.

Fail-closed matters here: the renderer applies a workspace-less payload to
`activeWorkspaceId` (`payloadWsId ?? state.activeWorkspaceId`), so returning
`undefined` would still write into whatever workspace the human is looking at —
the original vulnerability. An external caller that cannot prove which pane it is
calling from is now **refused**.

## Unscoped plugin `events.poll` (REAL)
`PluginFrame` polled with no `workspaceId`. Server-side only the PRIVATE types
(`a2a.task`, `channel.*`) are dropped for an unscoped poll; every lifecycle type
fell through to an all-workspace firehose, so a plugin holding
`events.subscribe` observed other workspaces' `pane.*` / `process.*` events.

The poll now carries `activeWorkspaceId` on both the first and subsequent polls,
and re-subscribes when the active workspace changes.

## U4 / U3 — accepted, not bugs
- **U4** (`principal.remove` / `markStaleWorkspace`): transport-mitigated —
  renderer-only IPC + `wmux.internal` capability + same-user ceiling.
- **U3** (`operatorList` "not pipe-registered"): terminology confusion. It is
  absent from the external-facing `RpcRouter` and registered only on the daemon's
  internal pipe, which only the main process can reach. Comment corrected.

## Tests
12 cases in `meta.rpc.test.ts` (resolution, forged-workspaceId rejection,
clamping, all four fail-closed paths, first-party passthrough) and 4 wiring
guards in `pluginFrameEventScope.test.ts`.

Both fail against `origin/main` (8 and 4 failures) and pass after:
`npx vitest run src/main/pipe/handlers/__tests__/meta.rpc.test.ts src/main/pipe/handlers/__tests__/events.rpc.test.ts src/renderer/plugins/__tests__/` → 64 passed

## Summary by CodeRabbit

* **Security**
  * Improved workspace isolation for metadata operations and plugin event polling.
  * Requests without a resolvable workspace are now rejected.

* **Bug Fixes**
  * Plugin event polling now follows the active workspace and refreshes when it changes.
  * Metadata status and progress updates use the correct workspace scope.

* **Tests**
  * Added coverage for workspace validation, caller behavior, progress handling, and workspace changes.

* **Documentation**
  * Clarified security boundaries for operator-list communication.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -46,11 +46,11 @@
 - **Context:** 스코핑 조사 판이 `gh pr diff 679`로 확인하던 중 판이 종료돼 결론을 못 봤다. 시작점: `PlaywrightEngine.ts`의 `resolveSelectionContext` / `explicitSurfaceId` 취급.
 - **Priority:** P0 — #679 머지 전에 답이 나와야 한다.
 
-## 스코핑 조사 U4·U8·U3 미검증 (P1)
-- **What:** `plans/workspace-scoping-survey.md`의 164지점 중 미검증 3건. **U8**(`meta.setStatus/setProgress`가 호출자 스코프 없이 사람이 보는 활성 워크스페이스에 씀)이 U7과 같은 계열로 보인다. U4(`principal.remove`/`markStaleWorkspace`가 스탬프 존재만 확인, 대상 비교 없음), U3(`operatorList` 주석은 "파이프 미등록"인데 `daemon/index.ts:3050`에 등록됨).
-- **Why:** U7은 REAL이라 PR #679가 됐고, U1은 검증 결과 ACCEPTED였다. 나머지도 갈라야 한다 — 관찰만 남기면 다음 사람이 어느 쪽인지 모른다.
-- **Context:** 판정 기준·1차 오판 사유가 문서에 기록돼 있다. 낮은 effort로 관찰 → 높은 effort로 검증 2단계가 U1에서 실효를 증명했다.
-- **Priority:** P1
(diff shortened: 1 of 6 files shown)


openwong2kim:

This PR closes two workspace-isolation holes surfaced by the scoping survey: `meta.setStatus`/`setProgress` accepted a caller-supplied `workspaceId` (so any agent could write status into a workspace it has no relationship with), and `PluginFrame` polled `events.poll` with no workspace scope, leaking other workspaces `pane.*`/`process.*` lifecycle events to any plugin holding `events.subscribe`. Both are now fixed with server-side workspace resolution from `senderPtyId` and a fail-closed refusal when identity cannot be proven — the right call given the renderer would otherwise apply an unscoped payload to the active workspace. Twelve meta RPC cases plus four plugin wiring guards, all failing against `main` and passing after, back the change. Tagging P2: these are real cross-workspace write/leak surfaces (same-user ceiling, so not P1), and the fix is clean and well-scoped.

PR #720, opened 2026-07-31
Show unread count on the dock and tray

openwong2kim:
With the wmux window minimized or hidden, a single OS toast was the only unread
signal — nothing persistent.

## What changed
- macOS: `app.dock.setBadge(N)`, capped at `999+`, cleared with an empty string
  at zero.
- Windows/Linux: the tray tooltip becomes `[N] wmux` (no dock exists).
- `StatusBar` pushes the existing `computeUnreadCount` result to main whenever it
  changes; main normalizes it to a non-negative integer before any platform call.

Two hygiene fixes while wiring it: the tray helper is imported statically like
every other import in `registerHandlers.ts` (it was an inline `require`), and the
renderer-supplied count is floored and finite-checked so a hostile or buggy value
cannot reach `setBadge`.

## Tests
12 cases: both platform branches, clear-at-zero, the 999+ cap and its exact
boundary, no cross-talk between dock and tray, a no-op before the tray exists,
plus wiring guards for the static import, the normalization, and listener
teardown.

`npx vitest run src/main/__tests__/ src/renderer/components/StatusBar/__tests__/` → 91 passed

## Notes
Cross-platform QA on all three OSes is still outstanding; the unit tests pin the
per-platform logic, not the OS rendering.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -183,14 +183,11 @@
      list; `useNotificationPolicy` skips toast/sound/ring/flashFrame for muted
      workspaces while still recording the entry in the panel (policy A4). -->
 
-## (E5) Tray icon unread badge (cross-platform)
-- **What:** `src/main/tray.ts`에 unread count를 표시. macOS: `app.dock.setBadge(N)`, Windows: tray tooltip prefix `[N] wmux`, Linux: best-effort (NotificationServer 표준은 제한적).
-- **Why:** wmux 창이 minimize/hide 상태에서는 OS 토스트 한 번 외에는 알림 신호 부재. tray가 영구 신호 채널.
-- **Pros:** 창 안 열어도 unread 인지. macOS 사용자에게 자연스러운 UX.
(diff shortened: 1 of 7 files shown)


openwong2kim:

openwong2kim:

This PR adds a persistent unread signal so a minimized or hidden wmux window is not relying on a single transient OS toast: macOS gets `app.dock.setBadge(N)` capped at `999+` and cleared at zero, while Windows/Linux get a `[N] wmux` tray tooltip. The StatusBar pushes the existing `computeUnreadCount` result to main on change, and main normalizes it to a non-negative integer before any platform call. Two hygiene fixes ride along correctly: the tray helper is now a static import like every other import in `registerHandlers.ts` (was an inline `require`), and the renderer-supplied count is floored and finite-checked so a hostile or buggy value cannot reach `setBadge`. Twelve cases cover both platform branches, the clear-at-zero boundary, the 999+ cap, no cross-talk, and wiring guards. P3: a straightforward, well-tested visibility improvement; cross-platform OS-rendering QA is still outstanding but the unit tests pin the per-platform logic.

PR #721, opened 2026-07-31
Add a per-workspace agent roster to the sidebar

openwong2kim:
The sidebar showed one aggregated status per workspace, so a workspace running
several agents gave no way to see which one needed you without clicking through
panes and tabs.

## What changed
Each workspace row gains a collapsible agent roster: one line per detected agent,
with its status, its location (pane · tab · position), and its pending question or
current activity. Clicking a row focuses exactly that pane and surface. The list
auto-expands for the active workspace and collapses in the background.

The selector deliberately does **not** reuse `selectFleetPanes`. Fleet aggregates
the most urgent status in a leaf onto that leaf's *active* surface, which is right
for a pane card but would attribute a background tab's status to the wrong agent.
Roster identity and status stay keyed to the same PTY.

Status precedence, strongest first: a transcript-derived pending question → an
unseen attention state → the retained lifecycle state, with a fresh activity stamp
able to promote `idle` to `running`. Identity-only boot hydration seeds `running`
with no activity signal, so that synthetic value reads as `idle` until real output
or a hook proves otherwise — otherwise a quiet recovered pane would pulse forever
and disagree with the workspace aggregate.

Also here: an activity-only hook payload is ordered proof the agent resumed after
an older `complete`/`waiting`, so `useNotificationListener` reconciles the
lifecycle mirror. Only this listener knows event order — a selector cannot compare
an activity timestamp against an attention state that carries no timestamp. An
explicit state on the same payload still wins.

Status labels are shortened (`Agent running` → `Running`) because the roster shows
them beside the agent name, where the prefix was redundant.

## Tests
19 selector cases (per-PTY identity, background surfaces, nested branches, all
precedence rules, brain-pty exclusion, non-terminal/unspawned skips, focus
scoping, attention counting, memoized-reference stability) and 2 new listener
cases proving the reconciliation and that it does not fire on an empty activity
string.

Both new listener cases fail against `origin/main` and pass after:
`npx vitest run src/renderer/stores/selectors/ src/renderer/hooks/__tests__/useNotificationListener*` → 161 passed
`npx tsc --noEmit` → 0 errors


openwong2kim:

openwong2kim:

This PR adds a collapsible per-workspace agent roster to the sidebar: one line per detected agent with its status, location (pane, tab, position), and pending question or current activity, and clicking a row focuses exactly that pane and surface. The selector design is worth calling out — it deliberately does not reuse `selectFleetPanes`, which aggregates the most urgent status onto a leaf's active surface (correct for a pane card but wrong for attributing a background tab to the right agent), so roster identity and status stay keyed to the same PTY. Status precedence (transcript-derived pending question, then unseen attention state, then retained lifecycle, with activity able to promote idle to running) and the synthetic-running-as-idle boot hydration are well-reasoned. Nineteen selector cases plus two listener reconciliation cases, both new listener cases failing against `main`, back the change. P3: a useful visibility improvement for multi-agent workspaces with solid test coverage; additive UX rather than a correctness fix.

PR #722, opened 2026-07-31
Keep authoritative agent status across redraws

openwong2kim:
A pane that was actually blocked on a question kept reporting `running`. Byte
throughput and lifecycle edges compete inside the daemon, and throughput was
winning: an idle Claude repaint five seconds after a Stop hook resurrected stale
`running` metadata.

## What changed
`DaemonPTYBridge` now gives an explicit detector/hook terminal state priority over
passive bytes:

1. A terminal status (`waiting` / `complete` / `awaiting_input` / `error`)
   **settles** the turn. While settled, idle repaints are ignored entirely.
2. Only a real submitted-input boundary re-opens the gate — a CR/LF outside
   bracketed paste, or an explicit control that carries no Enter
   (`noteInput(data, true)` for an approval digit/ESC).
3. An explicit `running` edge also re-opens it, for autonomous work.
4. Within one chunk, activity is processed **first** so a terminal pattern found in
   that same chunk is forwarded last and remains authoritative.

`noteInput` tracks bracketed paste across writes, so newlines inside a pasted
draft stay inert and only the later CR counts as a submit.

`ActivityMonitor.beginTurn()` re-arms a cycle from that submitted boundary: after
a real Enter the next output byte is enough to prove the turn is running, so a
short reply that never crosses the 2 KB threshold no longer stays stuck in the
previous state.

Call sites: the approval registry (digit/ESC, force-submitted), pipe client input,
the phone HTTP write, the ChannelWakeWorker, and hook-driven agent events (which
settle the bridge before broadcasting).

## Tests
21 new cases: baseline burst behaviour, all four terminal states, the running
re-open, typing vs CR vs LF, bracketed-paste inertness including an unterminated
paste across writes, force-submitted vs bare digit, same-chunk ordering, and both
reset paths. Plus 9 `beginTurn` cases including the sub-threshold reply that is
the actual bug.

20 of 21 fail against `origin/main`; after the change:
`npx vitest run src/daemon/__tests__/ src/main/pty/__tests__/ src/daemon/web/__tests__/` → 1109 passed
`npm run build:daemon` → clean

### DIFF
--- a/src/daemon/DaemonPTYBridge.ts
+++ b/src/daemon/DaemonPTYBridge.ts
@@ -2,7 +2,7 @@ import { EventEmitter } from 'node:events';
 import type { IPty } from 'node-pty';
 import { OscParser } from '../main/pty/OscParser';
 import { TerminalNotificationParser } from '../main/pty/oscNotification';
-import { AgentDetector } from '../main/pty/AgentDetector';
+import { AgentDetector, type AgentEventStatus } from '../main/pty/AgentDetector';
 import { ActivityMonitor } from '../main/pty/ActivityMonitor';
 import { parseOsc7Cwd, detectPromptCwd } from '../main/pty/cwdDetect';
 import { sanitizeTitle } from '../main/pty/titleDetect';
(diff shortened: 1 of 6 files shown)


openwong2kim:

openwong2kim:

This PR fixes a real status-reporting bug: a pane genuinely blocked on a question kept reporting `running` because passive byte throughput outranked detector/hook lifecycle edges in the daemon — an idle Claude repaint five seconds after a Stop hook could resurrect stale `running` metadata. The fix gives explicit terminal states (waiting, complete, awaiting_input, error) priority over passive bytes, re-opens the gate only on a real submitted-input boundary (CR/LF outside bracketed paste, or a force-submitted approval digit/ESC) or an explicit running edge, and orders activity within a chunk so a terminal pattern found in the same chunk remains authoritative. The `noteInput` bracketed-paste tracking and the `beginTurn` re-arm for sub-threshold replies (the actual bug) are well-considered. 21 new daemon cases plus 9 beginTurn cases, 20 of 21 failing against `main`, give strong regression coverage. P2: a correctness fix to the status signal that drives notifications and the sidebar, with solid tests; not P1 because it is a metadata-staleness issue rather than a crash or data loss.

PR #723, opened 2026-07-31
Promote cap-skipped suspended sessions on demand

openwong2kim:
Boot recovery honours a session cap. A workspace beyond that cap came back with
its ptyId absent, so reconcile destructively cleared it — the pane lost both its
identity and its scrollback, even though the daemon still held a suspended session
for it.

## What changed
- `daemon.listSessions` accepts `{ includeSuspended }` and appends cap-skipped
  suspended sessions from the persisted state (no duplicates with the active set).
- New `daemon.promoteSession(id)` spawns exactly the one session the renderer
  needs, reusing the persisted geometry, env, agent and **id** so the ptyId stays
  stable and scrollback restores from the ring buffer.
- `pty:promote` IPC + `pty.promote()` preload binding.
- `AppLayout.reconcilePtys` attempts a promote **before** the clear path, and only
  for ptyIds it has confirmed are suspended.

Guards: already-active is an idempotent success; not-suspended is `NOT_FOUND`; a
missing id is `INVALID_PARAMS`; only the known transient ConPTY `error code: 87`
is retried, and with a bound. A failed promote (cap hit, spawn error) falls
through to the existing clear path, and an older main process without
`pty.promote` keeps the pre-change behaviour exactly.

## Tests
20 invariants across `pty.handler.promote.test.ts` and the extended
`appLayout.sessionSaveInvariants.test.ts` — promote-before-clear ordering,
suspended-only targeting, failure fallback, abort-signal handling, idempotency,
the retry bound, geometry/identity reuse, and the `includeSuspended` default.

All 20 fail against `origin/main`; after the change:
`npx vitest run src/main/ipc/__tests__/ src/renderer/components/Layout/__tests__/` → 113 passed
`npm run build:daemon` → clean, `npx tsc --noEmit` → 0 errors

## Not covered by tests
The actual PTY spawn, the scrollback dump replay, and the cap
`RESOURCE_EXHAUSTED` path need a live daemon. These are structural guards (the
house pattern for this file); the spawn path still wants dogfood.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -192,14 +192,11 @@
 - **Depends on:** 알림 파이프라인 복구 ship + 사용자가 minimize 사용 빈도 확인
 - **Priority:** P2
 
-## Fix B — Scrollback restore cap-aware suspended-session promote
-- **What:** `daemon.listSessions`에 `{includeSuspended}` 파라미터 추가 + `daemon.promoteSession(id)` 신규 RPC. `AppLayout.reconcilePtys` fallback 직전에 promote 시도. cap=40 초과로 자동 복구되지 못한 suspended session도 reconnect 가능하게.
-- **Why:** Fix 0 (mount gate + saved ptyId 보존)는 cap *안*의 graceful Quit session만 복원. cap *초과* session (50+ panes 사용자, 또는 force-kill 후 daemon snapshot은 살아있지만 cap에 밀린 case)은 여전히 fresh terminal. design doc §5 + Update에 spec.
-- **Pros:** scrollback restore feature 100% 완성 (cap 안 + 초과 모두). power user 사용성 결정적.
-- **Cons:** RPC 2개 추가 + dynamic test R2-R5 (cap-skipped promote 4 scenario). Effort ~3-4시간.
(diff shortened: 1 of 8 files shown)


openwong2kim:

openwong2kim:

This PR completes the scrollback-restore feature for sessions that exceed the boot-recovery session cap. Previously a workspace beyond the cap came back with its ptyId absent and reconcile destructively cleared it — losing identity and scrollback even though the daemon still held a suspended session. The fix adds `daemon.listSessions({includeSuspended})` and a new `daemon.promoteSession(id)` RPC so the renderer can spawn exactly the one session it needs (reusing persisted geometry, env, agent, and id for stable ptyId and scrollback restore), with `AppLayout.reconcilePtys` attempting a promote before the destructive clear path. Guards are thorough: already-active is idempotent, not-suspended is NOT_FOUND, only transient ConPTY error 87 is retried with a bound, a failed promote falls through to the existing clear path, and an older main process without `pty.promote` keeps pre-change behavior exactly. 20 invariants across the two test files, all failing against `main`, pin the ordering and fallback. P2: this finishes a power-user-facing recovery feature (50+ panes) with correct structural guards; live daemon dogfood of the spawn and RESOURCE_EXHAUSTED path is still outstanding, which keeps it off P1.

PR #724, opened 2026-07-31
Carry human requests through to verified done

openwong2kim:
The commander could report a delegated request as done the moment its workers went
quiet — a model turn ending was treated as completion. Nothing tracked that a
human had asked for something, so ownership evaporated across turns, pane stops,
A2A handoffs and restarts.

## What changed

**Durable ownership.** `deckWorkStore` records one active request per workspace.
Every direct human prompt creates or extends it — including prompts typed straight
into the embedded terminal brain, where `UserPromptSubmit` is the only place main
learns work was assigned. A second message is **appended as a follow-up**, never a
silent replacement, so workers already running keep their owner. The record
survives a restart.

**A finalization gate.** `deck_complete_work({summary, verification})` is the only
way to close it, and the server does not trust the prose: it re-checks the live
worker snapshot and **re-queries every A2A task** projected into the record before
deleting it. Not-canonically-completed, unavailable canonical state, or a failed
query all refuse. Deletion is compare-and-delete on the full revision — a human
follow-up keeps the same work id, so an id-only check would delete ownership of
instructions the brain never saw.

**Request-scoped autonomy.** While a request is open the coalescer may wake on
lifecycle and A2A edges and drive follow-up instructions even when the resting
mode is `off`, with a larger bounded wake budget. Two things deliberately do not
widen: `approvalPress` stays exactly what the standing mode says, and the global
auto-wake kill switch still wins.

**A2A receipts** (`completed` / `failed` / `input_required` / `canceled`) route
through the same coalescer as pointers, not proof — a completion claim reports its
verified-evidence count and is labelled `UNVERIFIED CLAIM` at zero.

**Stop gate.** A durable record blocks the turn with a finalize instruction even
when the pane snapshot is absent or stale, still bounded by the consecutive-block
cap so a broken store cannot trap the TUI.

## Tests
83 new/updated cases: 27 store (lifecycle, follow-up append/cap, all four
compare-and-delete races, A2A projection, corrupt/hostile file handling,
workspaceId re-derivation), 20 `deck.completeWork` (auth, text floors, outstanding
workers, all five A2A refusal paths, the mid-flight race), 18 coalescer, 6 stop
gate, 4 foreign-turn-start, plus the rendered brain block.

`npx vitest run src/main/deck/__tests__/ src/main/pipe/handlers/__tests__/ src/main/mcp/__tests__/ src/mcp/__tests__/` → 1507 passed
`npm run build:mcp` → clean


openwong2kim:

This is a significant behavioral change: the commander could previously report a delegated human request as done the moment its workers went quiet, because a model turn ending was treated as completion. This PR introduces durable per-workspace ownership via `deckWorkStore` (one active request, surviving restarts and A2A handoffs), a server-verified `deck_complete_work` finalization gate that re-checks live workers and re-queries every projected A2A task before deleting, request-scoped autonomy that can drive follow-up instructions without widening approval permissions, and A2A receipts routed through the coalescer as pointers rather than proof. The 83 new/updated tests cover store lifecycle, the compare-and-delete races, all five A2A refusal paths, the stop gate, and foreign-turn-start; approvalPress deliberately stays governed by the standing mode, which is the correct safety boundary. P2: a meaningful correctness improvement to the orchestration contract with thorough test coverage; not P1 because it adds a gate rather than fixing an active break, and live dogfood of the PTY spawn path is still outstanding.

openwong2kim:
Closing in favor of #730. The deck-active-work core is byte-identical across both branches (12 files: deckWorkStore, stopGate, CommanderEventCoalescer, deck.handler/rpc, firstParty, methodCapabilityMap, commanderSurface, rpc, ClaudePty/Sdk adapters, mcp/index); #730 is ~16 min newer and broader (adds the approvals subsystem + web/push).

Before #730 merges, port this PR's three UNIQUE test files so the shared deck core keeps coverage (they are absent from #730):
- `src/main/deck/__tests__/deckWorkStore.test.ts`
- `src/main/deck/__tests__/CommanderEventCoalescer.activeWork.test.ts`
- `src/main/pipe/handlers/__tests__/deck.completeWork.test.ts`

Heads-up: both #724 and #730 currently have failing CI (Baseline all platforms + validate). Branch left intact for salvage.

PR #725, opened 2026-07-31
Add tests for lifecycle routing and installer

openwong2kim:
Stacked on #714 — **base is `fix/cli-lifecycle-routing`, not `main`.** Merge #714
first; this then shows as a small test-only diff.

I built the same lifecycle change independently before noticing #714 already
covered it. Rather than ship a duplicate, this reduces to just the coverage that
was missing, against #714's source.

## What this adds
**50 test cases**, no behavioural change. The only source edit is `export` on nine
existing helpers in the OpenCode plugin so the path and target-walk logic can be
called from a test; the plugin runtime still loads only the default export.

### `integrations/opencode/__tests__/wmuxRouting.test.ts` (31)
- **Instance isolation** — every path (wmux home, main token, daemon token, daemon
  socket) scoped by `WMUX_DATA_SUFFIX`; a suffixed instance never reaches the
  production namespace; with no suffix the paths stay byte-identical to before.
- **Canonical socket** — the main socket resolves through `os.homedir()`, matching
  `src/shared/constants.ts getPipeName()`. Pinned specifically because a
  `USERPROFILE`-first lookup would send the bridge to a directory main never
  bound, while token/state paths deliberately keep honouring `USERPROFILE` for
  isolated dogfood.
- **Daemon-first walk** — ordering, per-endpoint skip on a missing token, the
  `WMUX_HOOKS_TO_MAIN` kill switch, the `WMUX_PIPE_NAME` single-target escape
  hatch (and that it does not fall back to the daemon), hint-over-derived
  preference, and per-namespace hint reads.
- **At-most-once delivery** — advances only when the request provably never
  reached a server; an outer-ok reply owns the signal; a post-write disconnect is
  ambiguous and stops the walk; a legacy-daemon transport refusal correctly falls
  through to main.
- win32 named-pipe naming for both endpoints.

### `src/shared/__tests__/lifecycleIntegrations.test.ts` (19)
missing / current / stale / foreign / source-missing classification, multi-marker
ownership, atomic install and refresh, **foreign files left byte-identical**, no
temp leftovers, source lookup through bundle and repo layouts, absolute
`XDG_CONFIG_HOME` honoured and a relative one ignored, live-checkout preferred
over a stale `dist`, and that `status` never writes.

### `src/shared/__tests__/mcpRegistration.test.ts` (+4)
A `notify` slot that is a plain string or contains non-string entries is a
conflict; an empty array names no program so wmux may claim it; a malformed config
reports distinctly from "no notify".

## Verification
`npx vitest run integrations/ src/cli/ src/shared/__tests__/lifecycleIntegrations.test.ts src/shared/__tests__/mcpRegistration.test.ts src/main/mcp/__tests__/` → **622 passed, 3 skipped**
`npx tsc --noEmit` → 0 errors

The routing suite runs against a throwaway `HOME`. These helpers read real token
and hint files, so a suite using the developer's home would read whatever a live
daemon happened to have written — that mistake produced two false failures while
writing this.

### DIFF
--- a/integrations/opencode/__tests__/wmuxRouting.test.ts
+++ b/integrations/opencode/__tests__/wmuxRouting.test.ts
@@ -1,0 +1,213 @@
+// Suffix isolation + daemon-first routing for the OpenCode lifecycle plugin.
+//
+// Two invariants this locks:
+//
+//  1. INSTANCE ISOLATION. WMUX_DATA_SUFFIX is an instance boundary. Every
+//     endpoint, credential and local-state path the plugin touches must stay
+//     inside the selected namespace, and a suffixed instance must never fall
+//     back to the production (unsuffixed) daemon token or hint. With no suffix
+//     every path stays byte-identical to the pre-isolation paths.
+//
+//  2. AT-MOST-ONCE DELIVERY. The plugin walks daemon → main. It may only
+//     advance when no server answered AND the request provably was not written;
+//     any id-matched reply (including a rejection) owns the signal, and a
+//     post-write disconnect is ambiguous, so both stop the walk.
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { homedir, userInfo } from 'node:os';
+import { join } from 'node:path';
+import { mkdtempSync, rmSync, writeFileSync, mkdirSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import {
+  dataSuffix,
+  getWmuxHomeDir,
+  getAuthTokenPath,
+  getPipeName,
+  getDaemonAuthTokenPath,
+  getDaemonPipeName,
+  resolveTargets,
+  shouldTryNextTarget,
+} from '../plugins/wmux.js';
+
+const SAVED = { ...process.env };
+
+// Every case runs against a throwaway home. These helpers read real files
+// (auth tokens, the daemon-pipe hint), so a test that used the developer's home
+// would both leak machine state into assertions and read whatever a live daemon
+// happens to have written.
+let home: string;
+
+beforeEach(() => {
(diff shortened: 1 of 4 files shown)


openwong2kim:

This PR fixes two real problems in how wmux installs and talks to agent lifecycle bridges. First, instance bleed: the Codex and OpenCode bridges resolved credentials, endpoints, and local state from unsuffixed paths, so a `-dev` instance's hooks were delivered to the production daemon while the dev deck waited forever — now every path is suffix-scoped with byte-identical behavior when no suffix is set. Second, clobbering: installation goes through one `lifecycleIntegrations` helper that only replaces destinations carrying a wmux ownership marker, reporting anything else as `foreign` and leaving it untouched, with temp+rename writes so a crash cannot leave a truncated bridge. It also brings daemon-first OpenCode routing with request-id correlation, official Codex `agent-turn-complete` payload handling that never logs prompt/assistant content, and safe `notify` slot handling. The 50 new test cases (31 routing, 19 installer) running against a throwaway HOME, plus clean tsc and build:cli, back the change. P2: this fixes a genuine cross-instance correctness bug and a file-clobbering risk with strong coverage.

openwong2kim:
Retargeted: this originally duplicated #714's source. Reduced to the missing test coverage and rebased onto `fix/cli-lifecycle-routing` so it stacks on #714 instead of competing with it.

openwong2kim:
Closing as duplicate — the lifecycle-integration test coverage this adds (`lifecycleIntegrations.test.ts`, `mcpRegistration.test.ts`) already landed on main via #714 + #727, and this branch's base (`fix/cli-lifecycle-routing`) is now fully merged. The test portions are therefore redundant.

Two genuinely-novel pieces remain here that are worth a fresh PR rebased on main:
- `integrations/opencode/__tests__/wmuxRouting.test.ts` (new file, not on main)
- the ~36-line delta in `integrations/opencode/plugins/wmux.js`

Branch left intact for salvage. Happy to extract those into a clean follow-up.

PR #726, opened 2026-07-31
chore(approvals): update stale per-option comment and drop unused export

openwong2kim:
## Summary

Follow-up cleanup from the #712 review. Two non-behavioural fixes:

- **`approvalKeystrokes.ts` header comment was stale.** It still deferred the per-option press to "M5/M6 scope, not here", but #712 implements exactly that via `choiceKey`. Rewrote the block to describe the default-mapping fallback and the `choiceKey` override, and pointed at `ApprovalRegistry.resolve` + the phone-client contract (`docs/phone-client-contract.md`).
- **`MAX_CHOICE_KEY_CHARS` was exported but never referenced.** All key validation uses the inline `/^\d{1,2}$/` regex. Removed the dead export.

## Safety and compatibility

- Comment-only change + one removed export. No behaviour change, no wire change.
- `MAX_CHOICE_KEY_CHARS` had zero references across the tree (confirmed via `rg`).

## Validation

- `vitest run src/daemon/approvals` — **127/127 passed**
- `vitest run src/daemon/web/__tests__/WebTerminalServer.test.ts` — **136/136 passed**
- `tsc -p tsconfig.daemon.json --noEmit` — clean for the changed files (only pre-existing unrelated `@anthropic-ai/claude-agent-sdk` module-miss in `src/main/deck/ClaudeSdkAdapter.ts`)

Stacked on #712 — merge after it.

### DIFF
--- a/src/daemon/approvals/approvalKeystrokes.ts
+++ b/src/daemon/approvals/approvalKeystrokes.ts
@@ -44,13 +44,20 @@
 //                 itself advertises, and it is the only mapping here that is
 //                 unambiguous for ANY question shape.
 //
-// KNOWN LIMITATION, deliberately shipped: for a yes/no question '1' is the
+// DEFAULT MAPPING + per-option override. For a yes/no question '1' is the
 // affirmative answer, but AskUserQuestion options are agent-authored, so on a
 // question whose first option is not the affirmative one, "approve" means
 // "pick the first option" rather than "say yes". Nothing on the wire tells us
-// which is which. The honest surface for a phone is therefore Deny (exact) +
-// Approve (first option); a richer per-option press belongs with the option
-// list in the phone UI, which is M5/M6 scope, not here.
+// which is which at the default-mapping layer.
+//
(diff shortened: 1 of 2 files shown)


openwong2kim:

This is a small cleanup follow-up from the #712 review: the `approvalKeystrokes.ts` header comment was stale (it still deferred per-option key press to M5/M6 scope, but #712 implements exactly that via `choiceKey`), so the block is rewritten to describe the default-mapping fallback and the `choiceKey` override, pointing at `ApprovalRegistry.resolve` and the phone-client contract. The second change removes `MAX_CHOICE_KEY_CHARS`, which was exported but never referenced — all key validation uses the inline `/^\d{1,2}$/` regex, confirmed via zero references across the tree. Safety is as stated: comment-only plus one removed export, no behavior or wire change. Validation (127/127 approvals tests, 136/136 web tests, clean daemon tsc for the changed files) is appropriate for the scope. P3: a correct, minimal hygiene PR with no behavioral risk.

PR #727, opened 2026-07-31
test(lifecycle): cover lifecycleIntegrations + notify-slot edge cases

openwong2kim:
Follow-up to #714 — adds the test coverage flagged as a gap in review. **Test-only; no behavior change, no version bump, no CHANGELOG edit** (per repo policy).

## What this adds

**`src/shared/__tests__/lifecycleIntegrations.test.ts`** (new, 24 tests) — covers the new `src/shared/lifecycleIntegrations` module that #714 introduced with no tests:
- `inspectLifecycleAsset` — every state: `current` (byte-equal), `missing`, `stale` (owned marker, different body), `foreign` (no marker), `source-missing`, `error` (unreadable dest), and that it never creates a dir / throws.
- `installLifecycleAsset` — installs missing (`action=installed`), refreshes stale (`action=refreshed`), no-op when current, **never overwrites a foreign destination**, no-op on source-missing, no leftover `.tmp` files.
- `findLifecycleAssetSourceFrom` — prefers a live `integrations/` checkout over a stale `dist/` build at the same level, walks up parents, returns null when absent (covers the opencode `wmux.js` ↔ bundle `wmux-opencode-plugin.js` rename too).
- `resolveLifecycleIntegrations` — codex/opencode dest paths, ownership markers, `XDG_CONFIG_HOME` honored + relative-XDG fallback.
- `statusLifecycleIntegrations` — codexNotify `stale` on path mismatch, `stale` when the configured script no longer exists, `wmux` when all current.
- `installLifecycleIntegrations` — `ok=false` on a fatal (source-missing), `codexNotify` gated to `null` until the bridge reaches `current`.

**`src/shared/__tests__/mcpRegistration.test.ts`** (+4 regression tests) — pins the notify-slot behavior change in #714:
- `notify = "string"` → `foreign`, byte-stable (regression: old `getNotify` returned null for non-arrays and **overwrote** it).
- `notify = ["node", 123]` (non-string element) → `foreign`, untouched.
- `notify = []` (empty) → claimed (a no-op slot).
- unparseable config → `readCodexNotifyStatus` reports `malformed` (was `none`).

## Validation
- `vitest run` on both files: **51 passed** (24 new + 27 existing).
- `tsc --noEmit`: only pre-existing env noise (missing optional `@anthropic-ai/claude-agent-sdk`, Vite build-time globals); nothing in `shared/` or the new test file.
- **Live dogfood** of the shipped `integrations/codex/bin/wmux-codex-notify.mjs` under `WMUX_DATA_SUFFIX=-dogfood` with an official `agent-turn-complete` payload carrying a sentinel secret in `input-messages`/`last-assistant-message`:
  - paths resolved to `~/.wmux-dogfood/` (suffix-scoped — never probed production `~/.wmux/`),
  - the sentinel appears in **zero** bridge artifacts (log + spool) — privacy holds,
  - bridge v0.3.0, exit 0.

### DIFF
--- a/src/shared/__tests__/lifecycleIntegrations.test.ts
+++ b/src/shared/__tests__/lifecycleIntegrations.test.ts
@@ -1,0 +1,296 @@
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import {
+  CODEX_NOTIFY_MANAGED_MARKER,
+  OPENCODE_PLUGIN_MANAGED_MARKER,
+  findLifecycleAssetSourceFrom,
+  inspectLifecycleAsset,
+  installLifecycleAsset,
+  installLifecycleIntegrations,
+  resolveLifecycleIntegrationPaths,
+  statusLifecycleIntegrations,
+  type LifecycleAssetSpec,
+  type LifecycleIntegrationPaths,
+} from '../lifecycleIntegrations';
+import { CODEX_NOTIFY_BASENAME } from '../configIO';
+import { getMcpTarget } from '../mcpTargets';
+
+let home = '';
+const codexTarget = getMcpTarget('codex')!;
+
+const SOURCE_TEXT = `${CODEX_NOTIFY_MANAGED_MARKER}\n// wmux-managed lifecycle bridge body\n`;
+const FOREIGN_TEXT = '// a user-owned script; no wmux marker anywhere\n';
+const MARKERS = [CODEX_NOTIFY_MANAGED_MARKER] as const;
+
+/** Build a spec whose source is a real temp file with the given contents. */
+function specWithSource(text: string, destinationPath: string): LifecycleAssetSpec {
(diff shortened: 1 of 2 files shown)


openwong2kim:

This is a test-only follow-up to #714 (now superseded by #725) that adds the coverage flagged as a gap in review: 24 new tests for the `lifecycleIntegrations` module (every `inspectLifecycleAsset` state, `installLifecycleAsset` install/refresh/no-op/foreign-preservation/no-temp-leftover behavior, `findLifecycleAssetSourceFrom` preference and parent walk, `resolveLifecycleIntegrations` paths and XDG handling, `statusLifecycleIntegrations` staleness, and `installLifecycleIntegrations` fatal handling) plus 4 regression tests pinning the notify-slot behavior change (string notify stays foreign and byte-stable, non-string element stays foreign, empty array is claimable, malformed config reports `malformed`). Validation is thorough: 51 passed across both files, tsc clean for the changed surface, and a live dogfood of the shipped codex notify bridge under a suffix-scoped data dir confirming the sentinel secret appears in zero bridge artifacts. P3: pure test coverage with no behavior change, no version bump, no CHANGELOG edit per repo policy — valuable hardening of the contract but low-risk by definition.

PR #728, opened 2026-07-31
Fix agent status truth and suspended-session recovery

openwong2kim:
Three fixes to what a pane's agent status actually reports, and to recovering a
pane the daemon still holds. They ship together because all three edit
`src/daemon/index.ts` — split apart they only conflict with each other.

Each is a separate commit, reviewable on its own.

---

## 1. `feat(agents): detect Kiro CLI sessions`
Kiro panes were invisible: no agent name, no status, no notifications. Kiro ships
no hook bridge, so identity has to come from the PTY stream.

`'kiro'` joins the `AgentSlug` union in all five authoritative places, and
`AgentDetector` learns a **compound** gate. A product-name mention is not enough —
agents routinely print logs that name other agents. Activation needs BOTH from the
same PTY: anchored chrome (the `Kiro CLI vX` banner, the trust-mode footer, or the
v3 `kiro.dev/docs/cli/` line) AND the anchored composer placeholder. Evidence may
arrive in either order; when the composer lands first its `waiting` state is
replayed once so an idle pane is not stuck at `running`.

Detection only — no approval route (Kiro has no hook surface).

## 2. `fix(daemon): keep authoritative agent status across redraws`
A pane blocked on a question kept reporting `running`: an idle Claude repaint five
seconds after a Stop hook resurrected stale metadata, because byte throughput was
outranking lifecycle edges.

Now a terminal status (`waiting` / `complete` / `awaiting_input` / `error`)
**settles** the turn and idle repaints are ignored. Only a real submitted-input
boundary re-opens the gate — a CR/LF outside bracketed paste, or an explicit
control that carries no Enter (an approval digit/ESC). An explicit `running` edge
also re-opens it for autonomous work. Within one chunk, activity is processed
first so a terminal pattern in that same chunk is forwarded last and wins.

`ActivityMonitor.beginTurn()` re-arms from that boundary, so a short reply that
never crosses the 2 KB threshold is no longer stuck in the previous state.

## 3. `feat(pty): promote suspended sessions on demand`
Boot recovery honours a session cap, so a workspace beyond it came back with its
ptyId absent and reconcile destructively cleared it — losing the pane's identity
and scrollback even though the daemon still held a suspended session.

`daemon.listSessions({includeSuspended})` plus a new `daemon.promoteSession(id)`
spawn exactly the one session needed, reusing the persisted geometry, env, agent
and **id** so the ptyId stays stable and scrollback restores from the ring buffer.
`AppLayout.reconcilePtys` tries the promote **before** the clear path, only for
ptyIds it confirmed are suspended. A failure (cap hit, spawn error) falls through
to the existing clear path, and an older main process without `pty.promote` keeps
the previous behaviour exactly.

---

## Tests
50 new cases across the three concerns: 9 Kiro compound-gate, 21 turn-priority
(bracketed-paste inertness, force-submitted, same-chunk ordering, both reset
paths), 9 `beginTurn`, and 20 promote invariants (ordering, suspended-only
targeting, failure fallback, idempotency, retry bound, identity reuse).

Verified against `origin/main` before the change: the Kiro suite fails 9, the
turn-priority suite fails 20 of 21, the promote suite fails 20.

Combined on this branch:
`npx vitest run src/daemon/__tests__/ src/main/pty/__tests__/ src/daemon/web/__tests__/ src/main/ipc/__tests__/ src/renderer/components/Layout/__tests__/ src/renderer/channels/__tests__/ integrations/shared/__tests__/` → **1279 passed**
`npm run build:daemon` → clean · `npx tsc --noEmit` → 0 errors

## Not covered by tests
The real PTY spawn, the scrollback dump replay, and the cap
`RESOURCE_EXHAUSTED` path need a live daemon. Those promote assertions are
structural guards (the house pattern for that file); the spawn path still wants
dogfood.


openwong2kim:

This PR fixes three correctness problems around what a pane's agent status actually reports and how a suspended session is recovered, bundled together because all three touch `src/daemon/index.ts`. First, it makes the daemon detect Kiro CLI sessions by adding a compound gate (anchored chrome banner AND the composer placeholder from the same PTY, either order with replay) so Kiro panes are no longer invisible; this is detection-only since Kiro has no hook surface. Second, it corrects a status-truth regression where an idle Claude full-screen repaint five seconds after a Stop hook resurrected stale `running` metadata: a terminal status (waiting/complete/awaiting_input/error) now settles the turn and suppresses later passive byte throughput until a real submitted-input boundary (CR/LF outside bracketed paste, or an explicit approval digit/ESC with `forceSubmitted`) or an explicit `running` edge re-opens the gate. Third, it adds cap-aware suspended-session promote so a workspace beyond the session cap comes back with its `ptyId` stable and scrollback restored from the ring buffer, via `daemon.listSessions({includeSuspended})` + `daemon.promoteSession(id)` attempted before the destructive clear path.

The status-priority change is the most subtle: within a single PTY data chunk, activity is processed first so a terminal pattern in that same chunk is forwarded last and wins, and `ActivityMonitor.beginTurn()` is re-armed from the submitted-input boundary so short replies are no longer stuck in the previous state. Tests are thorough (50 new cases across Kiro compound-gate, turn-priority including bracketed-paste inertness and same-chunk ordering, beginTurn, and promote invariants with failure fallback and idempotency), all verified to fail against `origin/main` first. Assigning P2: the status-truth fix addresses a real, visible misreport and the promote path completes scrollback recovery for power users; the Kiro detection is additive. Live-daemon dogfood of the spawn/scrollback/cap paths is still outstanding per the PR notes.

Files of note: `DaemonPTYBridge.ts` (the turn-priority core), `AgentDetector.ts` (Kiro compound gate), `ActivityMonitor.ts` (beginTurn re-arm), `pty.handler.ts` (promote RPC), and `AppLayout.tsx` (reconcile ordering).

PR #729, opened 2026-07-31
Surface which agent needs you in the sidebar

openwong2kim:
Three renderer changes to seeing which agent needs you. They ship together
because they share the per-surface state maps and the workspace i18n block.

Each is a separate commit, reviewable on its own.

---

## 1. `fix(renderer): clean transient PTY state on close`
`surfacePorts` and `surfaceAgentStatus` were the last two transient per-ptyId maps
not cleaned up on close. Every close left a dead entry, and because ptyIds are
reusable a new surface could inherit the previous one's status — reading as blocked
or running from birth. Found in the fleet-activity adversarial review, which
flagged `surfacePorts` as unsafe to copy as a cleanup precedent because it leaked
itself.

Both maps are now evicted at the two real teardown sites (`closePane` for the
whole subtree, `closeSurface` for the one surface), matching the existing
`surfaceAgent` / `surfaceActivity` pattern.

## 2. `feat(notifications): show unread app badges`
With the window minimized, one OS toast was the only unread signal.

macOS gets a dock badge (capped `999+`, cleared at zero); Windows/Linux get a
`[N] wmux` tray tooltip. `StatusBar` pushes the existing `computeUnreadCount`
result whenever it changes and main normalizes it to a non-negative integer before
any platform call. Two hygiene fixes while wiring it: the tray helper is imported
statically like every other import in `registerHandlers.ts` (it was an inline
`require`), and the renderer value is floored and finite-checked.

## 3. `feat(sidebar): add workspace agent roster`
The sidebar showed one aggregated status per workspace, so a workspace running
several agents gave no way to tell which one needed you without clicking through
panes and tabs.

Each workspace row gains a collapsible roster: one line per detected agent with
its status, location (pane · tab · position), and pending question or current
activity. Clicking focuses exactly that pane and surface.

The selector deliberately does **not** reuse `selectFleetPanes`: Fleet aggregates
the most urgent status in a leaf onto that leaf's *active* surface, which is right
for a pane card but would attribute a background tab's status to the wrong agent.
Roster identity and status stay keyed to the same PTY.

Precedence, strongest first: a transcript-derived pending question → an unseen
attention state → the retained lifecycle state, with a fresh activity stamp able
to promote `idle` to `running`. Identity-only boot hydration seeds `running` with
no activity signal, so that synthetic value reads as `idle` until real output
proves otherwise — otherwise a quiet recovered pane pulses forever and disagrees
with the workspace aggregate.

Also here: an activity-only hook payload is ordered proof the agent resumed after
an older `complete`/`waiting`, so `useNotificationListener` reconciles the
lifecycle mirror. Only this listener knows event order — a selector cannot compare
an activity timestamp against an attention state that carries no timestamp. An
explicit state on the same payload still wins.

Status labels are shortened (`Agent running` → `Running`) because the roster shows
them beside the agent name, where the prefix was redundant.

---

## Tests
Store leak regressions (subtree and single-surface teardown, sibling preservation,
reused-ptyId isolation), 12 badge cases (both platform branches, clear-at-zero,
the cap and its exact boundary, no dock/tray cross-talk, pre-tray no-op, plus
wiring guards), 19 roster selector cases (per-PTY identity, background surfaces,
nested branches, every precedence rule, brain-pty exclusion, focus scoping,
memoized-reference stability), and 2 listener cases for the reconciliation.

All verified to fail against `origin/main` first: 4 store failures, 4 plugin/badge
wiring failures, 2 listener failures.

Combined on this branch:
`npx vitest run src/renderer/ src/main/__tests__/ src/main/ipc/__tests__/` → **2955 passed**
`npx tsc --noEmit` → 0 errors

Combining surfaced one defect the standalone badge branch hid: that suite used
`import.meta.url`, which vitest accepts but `tsc` rejects under this project's
`module` setting. Fixed to the `__dirname` form the rest of the suite uses.

## Not covered by tests
Cross-platform badge QA on all three OSes is still outstanding; the unit tests
pin the per-platform logic, not the OS rendering.

## Summary by CodeRabbit

* **New Features**
  * Added workspace agent rosters showing agent status, activity, attention counts, and focused panes.
  * Added unread notification indicators: macOS dock badges and tray tooltip counts on Windows/Linux.
  * Added expand/collapse controls and localized labels in English and Korean.

* **Bug Fixes**
  * Improved activity updates so resumed agents correctly show as running.
  * Prevented stale agent status and connection data from persisting after panes or surfaces close.


openwong2kim:

This PR bundles three renderer-side improvements to seeing which agent needs you, shipped together because they share per-surface state maps and the workspace i18n block. First, it fixes a transient PTY-state leak where `surfacePorts` and `surfaceAgentStatus` were never cleaned up on close, leaving dead `ptyId` entries that a reused id could inherit as a stale status from birth. Second, it adds a cross-platform unread badge — a macOS dock badge (`999+` cap, cleared at zero) and a Windows/Linux tray tooltip (`[N] wmux`) — so a minimized window has a persistent unread signal beyond a single OS toast. Third, it adds a per-workspace agent roster in the sidebar with one line per detected agent carrying status, location (pane · tab · position), and pending question or current activity, with a correct precedence chain (transcript question → unseen attention → retained lifecycle, with activity able to promote idle to running).

Each change is a separate commit. The roster selector deliberately does not reuse `selectFleetPanes`, which would attribute a background tab's status to the wrong agent — a subtle correctness point worth a look during review. The store-leak fix and badge wiring come with regression tests (subtree/single-surface teardown, reused-ptyId isolation, 12 badge cases including platform branches and the cap boundary, 19 roster selector cases); cross-platform badge rendering on all three OSes is explicitly called out as not yet covered by unit tests. Assigning P3: these are meaningful UX hygiene and feature work but not urgent on their own.

Files of note: `WorkspaceAgentRoster.tsx`, `workspaceAgentRoster.ts` (selector), `tray.ts`/`registerHandlers.ts` (badge), and the `paneSlice.ts`/`surfaceSlice.ts` eviction additions.

PR #731, opened 2026-07-31
chore(approvals): update stale per-option comment and drop unused export

openwong2kim:
## Summary

Follow-up cleanup from the #712 review. Two non-behavioural fixes:

- **`approvalKeystrokes.ts` header comment was stale.** It still deferred the per-option press to "M5/M6 scope, not here", but #712 implements exactly that via `choiceKey`. Rewrote the block to describe the default-mapping fallback and the `choiceKey` override, and pointed at `ApprovalRegistry.resolve` + the phone-client contract (`docs/phone-client-contract.md`).
- **`MAX_CHOICE_KEY_CHARS` was exported but never referenced.** All key validation uses the inline `/^\d{1,2}$/` regex. Removed the dead export.

## Safety and compatibility

- Comment-only change + one removed export. No behaviour change, no wire change.
- `MAX_CHOICE_KEY_CHARS` had zero references across the tree (confirmed via `rg`).

## Validation

- `vitest run src/daemon/approvals` — **127/127 passed**
- `vitest run src/daemon/web/__tests__/WebTerminalServer.test.ts` — **136/136 passed**
- `tsc -p tsconfig.daemon.json --noEmit` — clean for the changed files

Rebased onto main after #712 merged.

## Summary by CodeRabbit

- **Documentation**
  - Clarified how approval prompts handle multiple options and structured choices.
  - Documented selecting a specific visible option by its digit after confirmation.
  - Clarified that unspecified selections preserve the default first-option behavior.
  - Explained that legacy option labels are display-only, while validated choices determine selection.

- **Maintenance**
  - Removed an unused public constant without changing choice validation or end-user behavior.

### DIFF
--- a/src/daemon/approvals/approvalKeystrokes.ts
+++ b/src/daemon/approvals/approvalKeystrokes.ts
@@ -44,13 +44,20 @@
 //                 itself advertises, and it is the only mapping here that is
 //                 unambiguous for ANY question shape.
 //
-// KNOWN LIMITATION, deliberately shipped: for a yes/no question '1' is the
+// DEFAULT MAPPING + per-option override. For a yes/no question '1' is the
 // affirmative answer, but AskUserQuestion options are agent-authored, so on a
 // question whose first option is not the affirmative one, "approve" means
 // "pick the first option" rather than "say yes". Nothing on the wire tells us
-// which is which. The honest surface for a phone is therefore Deny (exact) +
-// Approve (first option); a richer per-option press belongs with the option
-// list in the phone UI, which is M5/M6 scope, not here.
+// which is which at the default-mapping layer.
+//
+// Per-option press is now supported: the registry extracts structured
+// `choices` (key = the original 1-based digit, label) from the hook payload,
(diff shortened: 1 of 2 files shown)


openwong2kim:

This is a follow-up hygiene PR to #712 (explicit approval `choiceKey` selection). It rewrites the stale per-option header comment in `approvalKeystrokes.ts` that still deferred that feature to "M5/M6 scope" even though #712 shipped exactly that via the registry `choiceKey` override, and it drops the exported-but-unreferenced `MAX_CHOICE_KEY_CHARS` (validation uses the inline `/^\d{1,2}$/` regex). A second commit finishes the same cleanup in `askUserQuestion.ts`, where two more "M5/M6 per-option press" references no longer reflect reality.

The change is purely documentation plus one removed dead export: no behavioural, wire, or API change. It matters because comments that describe future work as upcoming when the work has already landed are actively misleading — the next reader will trust the header and re-derive the feature, or assume the code does less than it does. Low risk, mergeable as-is; worth landing so the approval code is self-describing again.

PR #732, opened 2026-07-31
test(opencode): white-box routing tests for the lifecycle plugin

openwong2kim:
Salvages the novel part of the closed #725. (The rest of #725 duplicated the \`lifecycleIntegrations\`/\`mcpRegistration\` coverage that already landed on main via #714 + #727, so it was closed.)

## What this adds
- Exports the OpenCode plugin's pure routing helpers (\`dataSuffix\`, \`getWmuxHomeDir\`, \`getAuthTokenPath\`, \`getPipeName\`, \`getDaemonAuthTokenPath\`, \`getDaemonPipeName\`, \`resolveTargets\`, \`shouldTryNextTarget\`) so they can be unit-tested directly. **Adding \`export\` is the only change to \`wmux.js\` — no behavioral change**; the default \`WmuxBridge\` export the OpenCode runtime loads is unchanged.
- \`integrations/opencode/__tests__/wmuxRouting.test.ts\` — locks two invariants:
  1. **Instance isolation** — \`WMUX_DATA_SUFFIX\` is an instance boundary; every endpoint/credential/state path stays in-namespace and a suffixed instance never falls back to the production daemon token/hint.
  2. **At-most-once delivery** — the daemon→main walk only advances when no server answered AND the write provably never happened.

## Cross-platform fix
The original #725 test was POSIX-only (asserted \`.sock\` paths) and would have failed the repo's **win32** Baseline CI. The socket assertions are now guarded: POSIX \`.sock\` checks run POSIX-only (\`it.skipIf\`), and the win32 named-pipe derivation is covered by the existing \`win32 naming\` suite (plus a win32 branch on the hint-namespacing test).

## Validation
- \`vitest run\` on the file: **18 passed / 2 skipped** on win32 (the 2 skipped are POSIX-only socket tests that run on Linux/macOS CI).
- \`node --check\` on the exported \`wmux.js\`: OK.
- Test-only + export-keyword additions → no version bump, no CHANGELOG fragment.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive coverage for OpenCode WMUX routing, including isolated configurations, home-directory and pipe overrides, daemon fallback behavior, retry decisions, and Windows pipe naming.
* **Refactor**
  * Exposed WMUX routing helpers for external access and testing without changing their behavior.

### DIFF
--- a/integrations/opencode/__tests__/wmuxRouting.test.ts
+++ b/integrations/opencode/__tests__/wmuxRouting.test.ts
@@ -1,0 +1,224 @@
+// Suffix isolation + daemon-first routing for the OpenCode lifecycle plugin.
+//
+// Two invariants this locks:
+//
+//  1. INSTANCE ISOLATION. WMUX_DATA_SUFFIX is an instance boundary. Every
+//     endpoint, credential and local-state path the plugin touches must stay
+//     inside the selected namespace, and a suffixed instance must never fall
+//     back to the production (unsuffixed) daemon token or hint. With no suffix
+//     every path stays byte-identical to the pre-isolation paths.
+//
+//  2. AT-MOST-ONCE DELIVERY. The plugin walks daemon → main. It may only
+//     advance when no server answered AND the request provably was not written;
+//     any id-matched reply (including a rejection) owns the signal, and a
+//     post-write disconnect is ambiguous, so both stop the walk.
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { homedir, userInfo } from 'node:os';
+import { join } from 'node:path';
+import { mkdtempSync, rmSync, writeFileSync, mkdirSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import {
+  dataSuffix,
+  getWmuxHomeDir,
+  getAuthTokenPath,
+  getPipeName,
+  getDaemonAuthTokenPath,
+  getDaemonPipeName,
+  resolveTargets,
+  shouldTryNextTarget,
+} from '../plugins/wmux.js';
+
+const SAVED = { ...process.env };
+
+// Every case runs against a throwaway home. These helpers read real files
+// (auth tokens, the daemon-pipe hint), so a test that used the developer's home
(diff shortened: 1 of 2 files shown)


openwong2kim:

**P3 — test hardening (no behavioral change).** This PR salvages the novel part of the closed #725 — white-box routing tests for the OpenCode lifecycle plugin — while dropping the `lifecycleIntegrations`/`mcpRegistration` coverage that already landed via #714/#727. The only change to production code is adding the `export` keyword to nine pure routing helpers (`dataSuffix`, `getWmuxHomeDir`, `getAuthTokenPath`, `getPipeName`, `getDaemonAuthTokenPath`, `getDaemonPipeName`, `resolveTargets`, `shouldTryNextTarget`, `getHomeDir`) inside `integrations/opencode/plugins/wmux.js`; the default `WmuxBridge` export the runtime loads is untouched.

The new `integrations/opencode/__tests__/wmuxRouting.test.ts` locks two invariants worth locking: (1) **instance isolation** — `WMUX_DATA_SUFFIX` is a hard boundary, and a suffixed instance never falls back to the production daemon token or pipe hint; (2) **at-most-once delivery** — the daemon→main walk only advances when no server answered *and* the write provably never happened, so an inner logical rejection (e.g. `no-workspace-match`) or an ambiguous post-write disconnect stops the walk rather than double-delivering. The socket-path assertions are also correctly guarded for cross-platform CI: POSIX `.sock` checks run POSIX-only (`it.skipIf`), and the win32 named-pipe derivation has its own suite — the original #725 test would have failed the repo's win32 Baseline. Verification in the PR body (`vitest run` 18 passed / 2 skipped on win32, `node --check` on `wmux.js`) looks consistent with the diff.

Low risk and a clear win for the test bed: it future-proofs two isolation/delivery contracts that are easy to silently regress and were previously only exercised indirectly. One thing worth confirming at merge time is that re-running the POSIX-only skips on a Linux/macOS job actually executes them (the two `it.skipIf(process.platform === 'win32')` cases), so the socket-form assertions do not become silently dead on every CI platform.

Issue #733, opened 2026-07-31
A finished pane can stay 'running' forever, and the orchestrator kills the shell trying to clear it


openwong2kim:
## What happens

Ask the terminal orchestrator to run a one-liner in a plain shell pane:

> Run `echo hello-from-orchestrator` in the PowerShell pane

The command runs, exits 0, and the orchestrator reads the output back correctly. Then it cannot finish. The pane still reports `running`, so the Stop gate refuses to end the turn. The orchestrator re-reads the screen, re-sends the command, and then escalates: it runs `exit` in the pane, and when that does not clear the status either, it starts sending Ctrl+D. It killed a live shell session that belonged to me. I stopped it with Escape.

Three minutes of that produced 277 `SessionManager.saveAsync` and 97 `pty.list` calls for a command that finished in milliseconds.

Dogfooded on `main` (dev build, daemon mode, Windows 11). The terminal brain is the default orchestrator since #710, so this is on the default path.

## Root cause 1 — a guard written for a deleted toast now swallows the status clear

`recentlySuppressed()` in `src/main/notification/idleSuppression.ts` exists, by its own docblock, to suppress the ActivityMonitor "Task may have finished" **notification** after a resize or a burst of typing. That toast was removed later — the comments in both `PTYBridge.onActiveToIdle` and `DaemonNotificationRouter.onIdle` say so explicitly ("keep the stale-'running' to 'idle' clear but DROP the byte-silence toast").

The guard was never narrowed to match. It still sits in front of the handler, so today it gates the only surviving job: the `running` to `idle` status clear.

Instrumented run, fresh app boot, no orchestrator involved:

```
[DBG-STUCK] main.onIdle pty=daemon-45c17ed3 sinceLastAgentEvent=neverms
[DBG-STUCK] recentlySuppressed pty=daemon-45c17ed3 byResize=true(22212ms) byWrite=true(1576ms)
[DBG-STUCK] main.onIdle pty=daemon-45c17ed3 -> SWALLOWED by recentlySuppressed (status stays 'running')
```

Exactly one `session:idle` arrived for that pane in the whole session, and it was dropped. The pane sat at an idle prompt reporting `running` for the rest of the session, and the top bar kept saying "1 running".

Two things make it permanent rather than transient:

[...]

PR #734, opened 2026-08-01
Stop a finished pane from reporting `running` forever, and from being killed to clear it

openwong2kim:
Part of #733. F1 and F3 from that issue; F2 (active-work expiry) follows separately — see "What is left" below.

## The pane that never went idle

Ask the orchestrator to run a one-liner in a plain shell pane. The command runs, exits 0, the orchestrator reads the output back correctly — and then cannot finish, because the pane still reports `running`. It re-reads, re-sends, and escalates: it runs `exit` in the pane, then starts sending Ctrl+D. It killed a live shell that belonged to me.

Two separate defects, and the second one is the reason the first one cost data.

## F1 — a guard written for a deleted toast was swallowing the status clear

`recentlySuppressed()` existed to suppress the ActivityMonitor "Task may have finished" **notification** after a resize redraw or a burst of typing. That toast was removed later; the comments in both handlers say so plainly. The guard was never narrowed to match, so it ended up in front of the only job the handler still had: clearing `running` back to `idle`.

Two things made a suppressed clear permanent rather than merely late:

- `ActivityMonitor` mutates its state *before* invoking callbacks, so the transition is consumed even when the callback is dropped. Nothing retries it, and a quiet pane never produces another burst.
- For a plain shell `AgentDetector` never matches, so no `session:agent` ever arrives. `session:idle` is the *only* path that can clear the status at all.

The argument that settled it: `onActive` was never gated. A resize redraw is several KB, so it raises a false `running` — and then the guard blocked the very clear that would have corrected it. It made the exact case it was named for worse. Either both edges are gated or neither is, and gating neither is the one that keeps the status honest.

So the call goes, which leaves the function unreferenced; `markUserWrite`/`lastUserWriteAt` exist only to feed it, so they go too. `markResize` stays — `recentlyResized()` (3 s, AgentDetector dedup) is a different concern with its own caller. Agent panes keep both protections they already had: `explicitTerminalStatus` in the daemon and `AGENT_EVENT_SUPPRESSION_MS` here.

The trigger is ordinary, which is worth stating: a `pty:resize` fires on every workspace switch and layout change, and the window was 30 seconds.

### Why a green suite missed it

Four test files stubbed the function that carries the bug:

```
recentlySuppressed: vi.fn().mockReturnValue(false),
```

Suppression never happened in the test world, so the branch that wedges a pane was never executed by any of the 9251 tests. That is the whole answer.

A regression test that mocked it would repeat that blindness, so the new one drives the real module: it calls `markResize()` the way `pty.handler` does, then asserts the clear still goes out. I checked it is not vacuous by reintroducing a suppression-shaped guard — three of its four cases fail. The four dead mock lines are removed with the function.

## F3 — a status number is not resolved by killing the thing it describes

The gate itself behaved as designed: it failed open after three refusals. The damage happened inside those three, and it will happen again the next time pane status is wrong about something, so this is worth fixing independently of F1.

The refusal string is the only thing the model reads about a block, and it was silent on the one action that loses data. It now says so.

Prose is not a control, though — the brain is a separate `claude` process and can ignore a sentence. `stopGate` already computes the exact set of panes it is blocking on, so `stopGateState` records that set and `input.rpc` refuses session-terminating input aimed at it. The scope is that intersection and nothing wider: an orchestrator that is not gate-held may close shells freely, and even a gate-held one may close panes it is not blocked on. `outstandingPtyIds` is exported from `stopGate` rather than reimplemented, so the panes the guard protects cannot drift from the panes the model was told about.

The record expires after 60 s on its own. The bug this PR fixes came from state that outlived its meaning and was never re-checked; the new state must not repeat that.

## Verification

Unit tests alone were not the gate here — this bug survived a full green suite for the reason above. The landing check was the live repro on a dev build.

Before, on a fresh boot with a reattached pane: `PowerShell running`, top bar `1 running`, at an idle prompt, indefinitely. After, same boot and same pane: `PowerShell idle`, `The agent is idle.` Running a command takes it to `running` and it settles back to `idle` on its own.

Suite: 9607 passed. Two failures in `mergeSession.test.ts` are `EBUSY: resource busy or locked` on Windows worktree cleanup, unrelated to these files — that file passes 18/18 in isolation.

## What is left

F2 from #733 (active-work expiry) is deliberately not here — it carries a policy question about what counts as stale, and holding F1 for it did not make sense. Until it lands, an old `deck-work.json` record can still drive auto-wakes; I watched that happen during verification, so it is not theoretical. `autoWakesUsed` lives in `CommanderEventCoalescer` memory and resets on construction, which is why a restart hands a wedged record a fresh 12/12 budget — keying the budget to the record means persisting it.

#733 stays open for that.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed panes that remained marked as running after commands completed, including after resizing.
  * Prevented protected panes from being closed or terminated while they block an active turn.
  * Normal pane termination remains available when protection does not apply.
* **Tests**
  * Added regression coverage for pane status clearing, protection behavior, workspace isolation, expiry, and termination commands.
* **Documentation**
  * Documented the pane-status and protected-termination fixes.

### DIFF
--- a/changelog.d/734.md
+++ b/changelog.d/734.md
@@ -1,0 +1,22 @@
+### Fixed
+
+- **A pane that finished its command no longer reports `running` forever.**
+  Clearing a pane back to `idle` sat behind a 30-second window built to suppress
+  a "Task may have finished" notification that had since been removed. A
+  workspace switch (any `pty:resize`) inside that window made the clear vanish,
+  and nothing retried it — the activity monitor consumes its state transition
+  before invoking callbacks, and a quiet pane never produces another burst. For
+  a plain shell, where no agent is detected, that clear is the only path there
+  is, so the pane sat at an idle prompt reporting `running` until it was closed.
+  The window also blocked the correction for the false `running` a resize
+  redraw raises in the first place. Agent panes are unaffected: their precise
+  hook and detector statuses still outrank byte silence. (#733)
+
+- **The orchestrator can no longer close a pane to clear its status.** When a
+  wedged pane held the Stop gate open, the terminal brain read "resolve these
+  panes" as "end these panes" and ran `exit`, then Ctrl+D, in a live shell the
+  user owned. The refusal text now names the prohibition, and session-terminating
+  input is refused outright when it targets a pane the caller's own turn is
+  currently blocked on. The scope is exactly that intersection — an orchestrator
+  that is not gate-held closes shells as before, and even a gate-held one may
+  close panes it is not blocked on. (#733)
--- a/plans/pane-status-stuck-running-2026-08-01.md
+++ b/plans/pane-status-stuck-running-2026-08-01.md
@@ -1,0 +1,243 @@
+# Pane status stuck at `running`, and the destructive escape from it (2026-08-01)
+
+- Status: **design, pre-review**. Tracking issue: #733.
+- Scope: `src/main/notification/idleSuppression.ts`, `DaemonNotificationRouter.onIdle`,
+  `PTYBridge.onActiveToIdle`, `src/main/deck/deckWorkStore.ts`, `stopGate.ts`,
+  and the turn-contract text shipped by #703.
+- Root cause confirmed empirically (instrumented dev build, daemon mode, Windows).
+  Evidence and repro are in #733; this document only decides what to change.
+
+## 0. What we are fixing
+
+Three defects that compound into one incident: a pane that finished its command
+keeps reporting `running`, the Stop gate therefore refuses to end the turn, and
+the brain escapes by killing the pane.
+
+They are separable and should be judged separately. F1 is the bug. F2 is what
+makes it recur after a restart. F3 is what turns a status bug into data loss.
+
+## 1. F1 — the status clear must not be gated by the notification suppressor
+
+### What is wrong
+
+`recentlySuppressed()` was written to suppress a **toast** ("Task may have
+finished") after a resize redraw or a burst of typing. That toast was later
+deleted; the comments in both handlers say so. The guard was never narrowed, so
+it now sits in front of the only surviving job: clearing `running` to `idle`.
+
+Two properties make a suppressed clear permanent rather than late:
+
+- `ActivityMonitor` mutates its state before invoking callbacks, so the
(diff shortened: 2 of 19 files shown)


openwong2kim:

**Priority: P2** — fixes the acute data-loss path from #733 (F1 + F3), where an orchestrator wedged by a stuck pane escalated to `exit`/Ctrl+D and killed a live user shell.

**What it touches.** Two coordinated fixes across 16 files:
- **F1 (idleSuppression.ts / PTYBridge / DaemonNotificationRouter):** removes `recentlySuppressed()` and `markUserWrite()` outright. The 30 s window was written for a notification toast that has since been deleted, but it kept gating the *only* surviving job — clearing `running` → `idle`. A workspace switch (`pty:resize`) inside the window dropped the clear, and since `ActivityMonitor` consumes its transition before invoking callbacks, a quiet plain-shell pane was wedged at `running` forever. The narrow `recentlyResized()` guard (3 s, AgentDetector dedup) is retained. Four test files that stubbed the function with `() => false` — the reason a 9251-test green suite never exercised the wedging branch — are cleaned up, and a new regression test drives the *real* module through the same `markResize()` entry point `pty.handler` uses.
- **F3 (stopGateState.ts + input.rpc.ts):** prose alone cannot control a separate `claude` process. A new short-lived (60 s TTL, self-expiring) record captures exactly the panes the Stop gate is blocking on — sourced from `outstandingPtyIds()` exported from `stopGate.ts` so the guard cannot drift from the refusal string — and `input.rpc` refuses session-terminating input (`exit`, `logout`, raw EOT) aimed at that intersection. Scope is deliberately narrow: a non-gate-held orchestrator, or a gate-held one closing an unrelated pane, is unaffected.

**Why it matters.** This is the default orchestrator path (#710), the trigger is ordinary (every workspace switch fires `pty:resize`), and the failure destroys user data. The enforcement half is independently valuable — pane status will misreport again, and when it does the brain must not reach for a kill.

**Scope note.** F2 (active-work expiry, `deck-work.json` surviving restart) is intentionally deferred to keep #733 open — worth tracking there, since the PR confirms auto-wakes still happen during verification.

PR #735, opened 2026-08-01
Park deck work that outlived the app, and make autonomy mode describe how Claude launches

openwong2kim:
### Active work outlived the app, and drove the fleet on the next launch

A durable work record used to authorize the orchestrator forever. Left over from a previous run it drove the fleet seconds after launch: the startup reconcile walked every surviving record and ordered "continue or repair incomplete work" with neither the autonomy mode nor a wake budget consulted. Captured live: an eight-hour-old objective replayed `claude --continue` into a live pane while the workspace sat at Mode: Off, and a human instruction sent at that moment was filed as a follow-up under the stale objective instead of being answered.

Permission now belongs to the app launch that was granted it. Records are stamped with the boot identity; one whose stamp is not the current one comes back **parked** — still owned, still holding the Stop gate, still rendered, but no longer driving anything by itself. The startup reconcile asks (`resume or drop`) instead of acting.

Two details that matter:
- The stamp is written **only** by a human turn, so nothing the fleet does can un-park a record.
- It is deliberately **not** `updatedAt` — an A2A transition bumps that, and a recovered pane's transition inside the startup window would have re-armed the very record this quiets. A UUID, not a timestamp, so clock skew cannot decide it.

### The Stop gate never told the model not to kill panes when no pane was named

The "Do NOT close or kill a pane" sentence added for #733 lived only in the pane branch. The three active-work-only exits shipped without it — and that is the branch a stale record lands on, naming no pane, which also puts it out of reach of `input.rpc` (that protects only the ptyIds a verdict names). Lifted to a shared constant.

### Autonomy mode now describes how Claude is launched (owner decision, 2026-08-01)

The modes named a wake policy and nothing else — nothing about them reached the process. `buildBrainLaunchCommand` only ever passed `--model`, and the pty adapter never read the autonomy store, so "Mode: Off" meant *fewer ambient wakes* while the brain still spawned and still ran turns.

| mode | launch posture |
|---|---|
| `off` | the brain does not run; turn entry refuses with `mode_off`, a running brain is disposed, the composer is disabled with the reason |
| `assist` | `--permission-mode acceptEdits` |
| `danger` | `--dangerously-skip-permissions` (this is `auto`, renamed for what it does) |

Wake policy becomes its own persisted axis instead of a shadow of the mode. Migration rides the seam that already exists for the four-mode era: `auto` joins `manual`/`orchestrate` in `LEGACY_MODE_MAP`, and an entry with no stored policy is seeded from its old mode — so **every existing workspace keeps the wake behaviour it had**.

The refusal is fail-closed: an autonomy store that throws refuses the turn rather than falling through to a launch, and it is checked at both turn entry points so the ambient drivers cannot route around it.

## Verification

`tsc --noEmit` clean. Deck/IPC/pipe suites green.

Tests write a real store to a temp dir rather than mocking it — #733 shipped because four suites mocked the culprit. The parking tests are verified **non-vacuous**: neutering the predicate fails six of them.

Live dogfood of the repro: seeded an eight-hour-old record carrying a previous boot id, restarted, and got zero `claude --continue` processes, zero panes driven, the record preserved, and a decision raised instead. The parked render drops "You OWN this request" / "Continue delegating" for "NOT authorization to act".

One crash was found **only** by running it, after the unit tests were already green: `MODE_SKIN[mode].btn` indexed a Record with a value that crosses an IPC boundary, so a main process answering with a retired mode name took the whole deck rail down through its ErrorBoundary. Now falls back to the `off` skin, with a test for the unknown-mode case.

## Not in scope, deliberately

- Objective supersede for a parked record (a human turn still appends a follow-up under the old objective).
- A2A cancellation on `clearActiveDeckWork`, which is still a blind delete.
- **#728** — recovery runs `claude --continue`, which revives the agent's conversation and with it its old task. Clearing `deck-work.json` does not prevent that, and it was the actual mechanism of harm here; parking removes the *autonomous* path into it, not the human-initiated one. Do not read this PR's dogfood as covering it.

Refs #733

## Summary by CodeRabbit

* **New Features**
  * Added workspace agent modes: Off, Assist, and Danger, with distinct Claude permission behavior.
  * Added separate wake-policy controls for automated activity.
  * Mode changes now update related interface surfaces automatically.
* **Bug Fixes**
  * Work in progress is safely parked across restarts and requires explicit resume or discard.
  * Off mode now prevents agent activity and disables message composition.
  * Improved safeguards and warnings when stopping work or closing panes.
* **Documentation**
  * Clarified lifecycle-event visibility and durable-request behavior.


openwong2kim:

**Priority: P2** — owner-authored fix for the root causes behind #733 (a finished-pane reporting `running`, driving the orchestrator to kill a live shell), and a substantive rework of how workspace autonomy modes reach the Claude process.

**What it touches.** Three coupled changes across the deck subsystem (`src/main/deck/`): (1) active work records that outlived the app are now **parked** on the next launch — still owned and still holding the Stop gate, but no longer driving wakes or spawns, with a `bootId` stamp that only a human turn can refresh so nothing the fleet does can self-un-park; (2) the "do not close or kill a pane to clear its status" instruction is lifted from the pane-scoped branch into a shared `stopGate` constant so the active-work-only refusals (the exact branch a stale record lands on) carry it too; and (3) autonomy `mode` is redefined to actually describe how Claude is launched — `off` refuses to spawn (and disposes a running brain), `assist` passes `--permission-mode acceptEdits`, and `danger` (the old `auto`, renamed for what it does) passes `--dangerously-skip-permissions`. Wake frequency is split out into its own persisted `wakePolicy` axis, seeded from the legacy mode during migration so every existing workspace keeps its current wake behaviour.

**Why it matters.** #733 was dogfooded on the default orchestrator path and the repro is ordinary — a resize within 30s of a short command is enough to wedge the pane permanently, after which the brain reached for `exit` and Ctrl+D to make the status change. Parking removes the durable-record mechanism that turned a millisecond command into 36 refusals across 12 wakes, the shared constant removes the one exit branch the pane-protection guard missed, and the mode rework closes a quieter but real gap: "Mode: Off" previously meant fewer wakes while the brain still spawned and ran turns. The changes are fail-closed by construction (an autonomy store that throws refuses the turn rather than falling through to a launch, and a resolver that throws drops the permission flag rather than the spawn), and the body honestly scopes what is *not* covered — #728 (recovery runs `claude --continue`, reviving the agent transcript) is explicitly called out as untouched. 36 files, +1150/−208, with new coverage including non-vacuous parking predicates (neutering the predicate fails six tests), platform-correct flag quoting on win32, and an ErrorBoundary fallback for an unknown retired mode name.

PR #737, opened 2026-08-01
fix(deck): stop the New session tooltip claiming the conversation survives

openwong2kim:
The `New session` chip's tooltip said:

> Panes, worktrees and **the conversation history stay**; loops and schedules keep running.

Discarding that conversation is the entire reason the chip exists.
`DECK_CONVERSATION_CLEAR` disposes the live brain and drops the persisted
session id so the next turn starts a fresh conversation instead of resuming.
So the tooltip reads as a promise that the brain still remembers — exactly
backwards for the one control an operator reaches for when the brain is stuck
on stale context.

What actually survives is the **#commander transcript** — the audit trail, as
the handler itself says ("The channel transcript stays — history is the audit
trail; only the brain's context resets"). The old line put panes, worktrees and
"conversation history" in one list, which made those two very different things
read as one claim.

Now: what the brain forgets and what stays as the record are one sentence, and
the untouched list (panes, worktrees, loops, schedules) is its own.

Strings only — `en.ts`, `ko.ts`, and the component's inline fallback. No logic
change.

## Tests

`npx tsc --noEmit` clean; `NewSessionChip` suite (4 tests) green. Found while
dogfooding #711 on a live build.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected the New Session button tooltip to clarify that the orchestrator conversation is forgotten while the `#commander` transcript remains.
  - Clarified that panes, worktrees, loops, and schedules are unaffected when starting a new session.
  - Updated the Korean translation to reflect the same behavior.

### DIFF
--- a/changelog.d/737.md
+++ b/changelog.d/737.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **The New session button no longer promises to keep the conversation it
+  discards.** Its tooltip listed the conversation history among the things that
+  stay, when starting a fresh session is precisely what clears it — the reverse
+  of what happens, on the one control you reach for when the orchestrator is
+  stuck on stale context. It now says the orchestrator forgets the conversation
+  and the #commander transcript remains as the record. (#737)
--- a/src/renderer/components/Deck/NewSessionChip.tsx
+++ b/src/renderer/components/Deck/NewSessionChip.tsx
@@ -146,7 +146,7 @@ export function NewSessionChip({
         armed
           ? confirmLabel
           : t?.('deck.newSessionTooltip') ||
-            'Replace this workspace’s orchestrator with a fresh session. Panes, worktrees and the conversation history stay; loops and schedules keep running.'
+            'Replace this workspace’s orchestrator with a fresh session. The brain forgets the conversation so far — the #commander transcript stays as the record. Panes, worktrees, loops and schedules are untouched.'
       }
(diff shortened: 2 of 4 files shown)


openwong2kim:

This is a strings-only fix for a tooltip that was actively misleading. The `New session` chip explained itself as "Panes, worktrees and the conversation history stay; loops and schedules keep running" — but `DECK_CONVERSATION_CLEAR` does precisely the opposite of "conversation history stay": it disposes the live brain and drops the persisted session id so the next turn starts fresh. So the one control an operator reaches for when the orchestrator is stuck on stale context was promising to keep the exact thing it discards, which is the reverse of helpful.

The new copy separates what the brain forgets (the conversation so far) from what survives as the record (the `#commander` transcript — the audit trail, per the handler itself), then lists the truly untouched surfaces (panes, worktrees, loops, schedules) on their own. The change touches `en.ts`, `ko.ts`, and the inline fallback in `NewSessionChip.tsx`; no logic changes. The `NewSessionChip` suite (4 tests) is green and `tsc --noEmit` is clean.

**Priority: P3.** It is a documentation-grade fix to a tooltip, not a functional regression — but the old wording was wrong in a way that misleads at exactly the wrong moment, so it is worth merging promptly while the #711 dogfooding context is live and the change is trivial to verify.

PR #738, opened 2026-08-01
fix(tray): stop the session nudge and the unread badge erasing each other

openwong2kim:
Found dogfooding #729 on a live build, then reproduced on a packaged one.

## The bug

Off macOS there is no dock, so the tray tooltip is the only unread surface —
and it had two independent authors, each writing the whole string:

- `updateUnreadBadge` → `[N] wmux` (base hardcoded)
- `updateTraySessionCount` → `wmux — N background sessions running`

Whichever spoke last erased the other. The losing case is exactly the one the
badge exists for: hide the window to the tray with unread mail, the hide-path
session refresh fires, and the count disappears from the only place still
showing it — while the unread count is still non-zero.

Reproduced on the packaged build by reading the real tooltip through UI
Automation:

```
after badge push : [5] wmux
after hide        : wmux — 7 background sessions running
```

Showing the window again clears the nudge to `null` and wipes the badge a
second time.

## The fix

Both signals write to module state; the tooltip is composed in one place from
both, so they coexist: `[5] wmux — 7 background sessions running`.

Two things fall out of holding the state:

- the session count is recorded **before** the `!tray` guard, and `createTray`
  applies whatever has been recorded — so a badge that arrives during boot is
  no longer dropped until the number happens to change again (the previous
  "pre-tray no-op" silently lost it);
- `destroyTray` clears both, so a stale count cannot compose itself onto the
  next tray.

The tooltip badge also picks up the `999+` cap the dock badge already had — same
count, and an unbounded one stretches the tooltip.

## Tests

The existing unit tests passed throughout: each pinned one writer in isolation
and the defect only exists where they meet. Added 8 cases — both orders, both
clears, singular/plural with a badge present, the cap, boot ordering, and the
destroy reset. **Six of the eight fail against `origin/main`** (the other two
end in the same terminal state under both versions, so they are guard rails,
not evidence).

```
npx vitest run src/main/__tests__/ src/main/ipc/__tests__/   → 258 passed, 8 skipped
npx tsc --noEmit                                             → 0 errors
```

Not covered: the OS rendering itself on macOS and Linux. The mac branch is
untouched (dock badge, early return) and the Windows path is what was verified
live.

### DIFF
--- a/changelog.d/738.md
+++ b/changelog.d/738.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **The unread count no longer vanishes the moment you minimise wmux.** On
+  Windows and Linux the tray tooltip is the only place an unread count shows
+  once the window is hidden — and hiding the window was itself what erased it,
+  because the "background sessions running" note and the unread badge each
+  overwrote the whole tooltip. They now share it, so a hidden wmux shows both.
+  A count that arrives while wmux is still starting up is no longer dropped
+  either. (#729, #738)
--- a/src/main/__tests__/trayUnreadBadge.test.ts
+++ b/src/main/__tests__/trayUnreadBadge.test.ts
@@ -19,6 +19,7 @@ vi.mock('electron', () => ({
     setToolTip = setToolTipMock;
     setContextMenu = vi.fn();
     on = vi.fn();
+    destroy = vi.fn();
   },
   Menu: { buildFromTemplate: vi.fn(() => ({})) },
   nativeImage: {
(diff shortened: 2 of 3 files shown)


openwong2kim:

This fixes a real UX defect on the Windows/Linux tray: two independent tooltip writers — `updateUnreadBadge` (`[N] wmux`) and `updateTraySessionCount` (`wmux — N background sessions running`) — each overwrote the whole string, so whichever fired last erased the other. The losing case is exactly the one the badge exists for: hiding the window with unread mail triggers the session refresh, which then wipes the unread count off the only surface still showing it.

The fix routes both signals through module state (`traySessionCount`, `trayUnreadCount`) and composes the tooltip in a single `applyTrayTooltip`, yielding `[5] wmux — 7 background sessions running`. Two consequences of holding the state are welcome: a badge arriving during boot is now recorded before the `!tray` guard and applied by `createTray` (previously dropped silently), and `destroyTray` clears both so a stale count cannot compose onto the next tray. The tooltip badge also picks up the dock-side `999+` cap. The test plan is solid — eight new cases (both author orders, both clears, singular/plural with badge, the cap, boot ordering, destroy reset), six of which fail against `origin/main`, plus a clean `tsc` and full vitest run. The mac path is untouched (dock badge, early return) and the Windows path is what was verified live, so the only unexercised branch is OS rendering on Linux.

**Priority: P2.** The symptom — unread count silently disappearing the moment you minimize — is confusing and strikes precisely when the badge is most needed, but it is a display-only defect with no data loss. Shipping this is worth doing while the tray-tooltip surface and the #729 nudge work it was dogfooded against are both fresh.

PR #741, opened 2026-08-01
Self-heal panes when the shared glyph atlas merges pages

openwong2kim:
## Problem

Since 3.38.2, Korean-heavy sessions started showing scattered wrong glyphs across several panes at once. A pane-border resize (full `refresh()`) did not repair it; opening a new split pane repaired every pane at once.

## Root cause (verified against @xterm/addon-webgl 0.19 internals)

The WebGL glyph texture atlas is shared across every same-config terminal (`acquireTextureAtlas` module cache). Its page pool caps at `min(32, MAX_TEXTURE_IMAGE_UNITS)` — 16 on Apple Silicon. When the pool is full, `_createNewPage()` merges four pages into one and deletes the originals, rewriting glyph `texturePage` indices and shifting the pages behind them. Sibling panes' renderers keep **diff-skipped vertex data** pointing at the old indices and sample the wrong page — which explains every observed symptom: multi-pane, refresh-immune (`_updateModel` diff-skips unchanged cells, so stale vertex entries survive a re-raster), and fixed only by a new pane's addon recreate.

CJK is the realistic trigger workload: Hangul alone has 11,172 distinct syllables, so Korean streams mint new glyphs continuously and exhaust 16 pages. The timing correlation with 3.38.2 is the DECRQM fix (#708) un-wedging TUI panes so more panes stream more content — the merge bug itself is pre-existing upstream.

## Fix — event-driven self-heal, public API only

The atlas forwards its page-removal event (fired **only** on a merge) through each owner's renderer to each owner's addon, so every sharing pane hears the merge on its own `WebglAddon.onRemoveTextureAtlasCanvas`. Each pane defers one macrotask (the event fires inside the atlas's own mutation mid-render; a merge emits one removal per deleted page, so the deferral also coalesces the burst) and calls its own `clearTextureAtlas()` — which clears the shared texture once (the atlas-level clear no-ops after the first pane), clears **that pane's glyph model**, and requests a redraw. That is the coherent all-owners rebuild upstream itself performs on font changes, so the #191 one-pane-clears hazard cannot occur.

No polling, no timers while idle, no internal-field reads. If upstream drops the event or the method, the heal degrades to a silent no-op.

## Review-driven redesign

The first iteration was a 2s poll over addon internals with a prevent/cure page-pressure gate. A three-model review panel (Claude, Codex, GLM) independently converged on the same faults: `clearTexture()+refresh()` never rebuilds the diff-skipped vertex model (the guard itself would corrupt panes — 2-model consensus, conf 10), the last-page-in-use anti-thrash gate mismatched the allocator's refill order (3-model consensus), and poll-window detection could be masked by regrowth (2-model). This event-driven design eliminates all three by construction.

## Tests

8 unit tests on the heal (deferral, burst coalescing, re-heal, dispose/cancel, degradation, throw-safety, injected scheduler); terminal + hooks suites 720 green; tsc + eslint clean. Heal disposal is wired into all four addon-teardown paths (pool evict, context loss, fonts.ready recreate, unmount).

### DIFF
--- a/changelog.d/741.md
+++ b/changelog.d/741.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Scattered wrong glyphs across several panes at once during CJK-heavy sessions. The shared WebGL glyph atlas's page-merge path corrupts sibling panes' cached glyph references when its page pool fills (16 pages on Apple Silicon; Hangul's 11,172 distinct syllables fill it fast); a new guard now rebuilds the atlas coherently for every sharing pane just before that path would run, and repairs within seconds if it ever does. (#741)
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -24,6 +24,7 @@ import { attachImeStormGuard } from '../terminal/imeStormGuard';
 import { webglContextPool } from '../terminal/webglContextPool';
 import { teardownWebglAddon } from '../terminal/webglTeardown';
 import { createGlyphRepaintScheduler, type GlyphRepaintScheduler } from '../terminal/glyphRepaint';
+import { atlasGuard } from '../terminal/atlasGuard';
 import { createDeadInputWatchdog } from '../terminal/deadInputWatchdog';
 import { awaitParseBarrier } from '../terminal/parseBarrier';
 import { STALE_REPLAY_INPUT_MODE_RESETS, STALE_REPLAY_DISPLAY_RESETS } from '../terminal/staleReplayModeReset';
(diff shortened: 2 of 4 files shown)


openwong2kim:

This PR fixes a multi-pane glyph-corruption rendering bug in the WebGL terminal renderer that specifically surfaces during CJK-heavy sessions (Hangul's 11,172 syllables exhaust the atlas's 16-page cap on Apple Silicon and trigger addon-webgl's page-merge path, which rewrites glyph `texturePage` indices under the feet of sibling panes). The fix adds a lightweight 2 s poll (`atlasGuard.ts`) that groups panes by their shared atlas, coherently clears and refreshes *every* sharing pane in the same tick just before the merge would fire (PREVENT), or immediately after if it slips through (CURE), with a "last page in use" gate that doubles as the anti-thrash condition and optional-chained addon-internals access that degrades to a silent no-op on upstream reshape.

The integration surface is small and well-scoped: a register/unregister pair wired into `useTerminal`'s main effect and cleanup, a changelog fragment, and 11 unit tests covering prevent/cure/anti-thrash/grouping/DOM-renderer-fallback/lifecycle plus the `maxAtlasPages`-unreadable fallback. The coherent all-owners rebuild deliberately mirrors what upstream itself does on font changes, sidestepping the #191 one-pane-clears hazard, which is the right call given the corruption only ever appeared across multiple panes at once.

Assigning **P2** — it is a genuine, reproducible, user-facing rendering failure for Korean (and broader CJK) workloads that no `refresh()` or resize can clear, and it has no real workaround beyond opening a throwaway pane; the fix is targeted, thoroughly tested, and low-risk, so it is worth merging soon while the addon-webgl 0.19 internals it depends on are still current.

PR #742, opened 2026-08-01
Let the macOS in-app update actually install

openwong2kim:
## The bug

Pressing **Check for updates** on macOS downloaded and SHA-256-verified the new version, then did nothing. No restart, no error, and nothing on any press after that.

`before-quit-for-update` belongs to Electron's `autoUpdater`, not to `app`, but the listener was registered on `app`. The `as unknown as NodeJS.EventEmitter` cast added to "work around the missing type" silenced the compiler error that would have said so, so `isQuitting` was never set.

`quitAndInstall()` closes every window and installs only once the window list empties. The hide-to-tray `close` intercept cancels that close while `isQuitting` is false. So the window list never emptied, the install never ran, and Squirrel's ShipIt helper waited on a process that was never going to exit. `isInstalling` then latched true for the process lifetime, so every later press hit a guard that returned without emitting anything — hence the silence on retry.

Verified on a live 3.38.1 install: ShipIt sat at `Detected this as an install request` for 45+ minutes with the staged 3.38.2 bundle fully unpacked and correctly signed, while the app kept running with its window open.

## The fix

Replace the dead event with two required constructor hooks the updater calls directly, so there is no event name left to get wrong. The prepare hook sets the quit flag, stops the broker, and flushes session state (the normal before-quit teardown is skipped entirely on this path). The abort hook reverses all of it.

The handoff is deadlined in two phases, because they fail differently:

```
setFeedURL -> checkForUpdates --- staging (Squirrel unpacks ~120 MB) --> update-downloaded
     |                                                                          |
     +-- staging deadline (10 min) ------------------------------ (cleared) ----+
                                    onBeforeInstallQuit -> quitAndInstall
                                         +-- handoff deadline (30 s) --> unwind or exit
```

Preparing the quit at the handoff rather than before staging keeps the app out of the closable, broker-stopped state for the whole download, and stops a tight deadline from aborting a healthy but slow update.

Every exit from an attempt detaches the Squirrel listeners. Without that a late `update-downloaded` could call `quitAndInstall()` after the abort had already restored the quit flag, re-creating the same hang.

Also in this change:

- **Restore the broker on abort.** `stop()` latches `stopped` and `start()` refuses to run while it is set, so one aborted install left MCP down for the rest of the session with nothing saying so. Adds `BrokerSupervisor.resume()`.
- **Wire recovery-created windows like the boot window.** Only boot attached the hide-to-tray `close` intercept, so a window from the `activate` or recovery path destroyed itself on close instead of hiding. Extracted to `adoptMainWindow`.
- **Report after the window is back.** The failure is reported once the restored window can receive IPC, not before it has listeners.
- **No more silent refusals.** Both early returns in `performInstall` now emit `UPDATE_ERROR`.
- **Sweep leaked installers.** An interrupted install left a verified ~120 MB artifact in temp; four had accumulated on the machine this was found on.

## Tests

The regression test asserts the prepare hook fires **before** `quitAndInstall`. Asserting that `quitAndInstall` was called is not enough — that assertion passed for the entire life of the bug. Verified by re-introducing the fault: the ordering tests fail, and pass again once restored.

New coverage: handoff stall unwinds and stays retryable; staging is given minutes but a stage that never finishes is still reported; a late Squirrel event after a failed attempt cannot re-trigger the install or double-report.

Full suite: 9585 passed, 0 failed.

## Scope

macOS Apple Silicon only. Windows reaches `app.quit()` and runs the normal `before-quit` pass, so it is unaffected and unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update installation reliability, including recovery when installation is interrupted or fails.
  * Added clearer error feedback when an update restart does not complete.
  * Prevented stale update events from triggering unintended reinstalls.
  * Cleaned up temporary files from interrupted updates.
  * Improved window recovery and hide-to-tray behavior during startup, activation, and updates.
  * Safely paused and restored background services and application sessions during installation.

### DIFF
--- a/changelog.d/742.md
+++ b/changelog.d/742.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **Check for updates now installs the update on macOS.** Pressing it downloaded
+  and verified the new version and then did nothing at all: no restart, no
+  error, and nothing on any press after that. Installing asks macOS to close
+  every window first and only goes ahead once they are all gone, but wmux keeps
+  its window alive so it can sit in the tray, so the close was cancelled and the
+  installer waited forever on a wmux that was never going to quit. The signal
+  that was meant to let those windows go was listening for something that is
+  never sent to it, so it never arrived. wmux now says so directly and there is
+  no message left to miss. If installing still fails to restart the app, wmux
+  stops waiting, brings the window back and tells you what happened instead of
+  going quiet, and the button works again on the next press rather than
+  silently refusing for the rest of the session. Updates that were interrupted
+  no longer leave their downloaded copy behind either, where they had been
+  piling up at around 120 MB each. (#742)
--- a/changelog.d/742.md
+++ b/changelog.d/742.md
@@ -1,6 +1,6 @@
 ### Fixed
 
-- **Check for updates now installs the update on macOS.** Pressing it downloaded
+- **Check for updates now installs the update on Apple Silicon Macs.** Pressing it downloaded
   and verified the new version and then did nothing at all: no restart, no
   error, and nothing on any press after that. Installing asks macOS to close
   every window first and only goes ahead once they are all gone, but wmux keeps
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -359,7 +359,13 @@ let mainWindow: BrowserWindow | null = null;
 // reordering hook/router boot earlier than the PTY layer.
 let hookSignalRouter: HookSignalRouter | null = null;
 const ptyBridge = new PTYBridge(ptyManager, () => mainWindow, () => hookSignalRouter);
-const autoUpdater = new AutoUpdater(() => mainWindow);
+// The hooks carry the two things the updater cannot reach: the quit flag that
+// lets windows actually close (without it quitAndInstall never installs — see
+// prepareInstallQuit) and the way back if that handoff stalls.
+const autoUpdater = new AutoUpdater(() => mainWindow, {
+  onBeforeInstallQuit: () => prepareInstallQuit(),
+  onInstallQuitAborted: () => abortInstallQuit(),
+});
 
 const rpcRouter = new RpcRouter();
 markBoot('pre-pipe-server-ctor');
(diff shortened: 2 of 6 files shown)


openwong2kim:

This PR fixes a silent failure in the macOS in-app updater: the `before-quit-for-update` listener was registered on `app` instead of Electron's `autoUpdater`, and the `as unknown as NodeJS.EventEmitter` cast suppressed the type error that would have caught it — so `isQuitting` never flipped, the hide-to-tray close intercept kept cancelling `quitAndInstall()`'s window close, ShipIt waited on a process that would never exit, and every later press hit the `isInstalling` latch and emitted nothing. The fix replaces the dead event with two required constructor hooks (`onBeforeInstallQuit` / `onInstallQuitAborted`) the updater calls directly, adds a two-phase deadline (10 min staging, 30 s handoff) so a stalled handoff unwinds instead of hanging forever, detaches Squirrel listeners on every exit to prevent late events re-triggering the install, and adds `BrokerSupervisor.resume()` plus `adoptMainWindow()` so an aborted install restores MCP and brings a window back rather than leaving the app unreachable. Scope is correctly bounded to Apple Silicon macOS (Windows reaches `app.quit()` and the normal before-quit pass, so it is unchanged), and the regression test deliberately asserts that the prepare hook fires *before* `quitAndInstall` — asserting the call alone passed for the entire life of the bug.

The change is well-structured and well-tested: +438/−54 across `src/main/index.ts`, `BrokerSupervisor.ts`, `AutoUpdater.ts`, and two test files (9585 passed, 0 failed), with new coverage for the handoff-stall unwind, staging timeout reporting, and late-event suppression. Two follow-ups stand out from review and are worth addressing before or shortly after merge, neither blocking: (1) `onBeforeInstallQuit()` is called without the try/catch that already guards its abort counterpart — a throw there escapes Squirrel's emitter and leaves the app in the prepared-quit state until the 30 s deadline; (2) `BrokerSupervisor.resume()` has a latent stale-child window where a pending `exit` from the killed broker can reach `scheduleRestart` after `resume()` has already spawned a new one, addressable with an identity guard on the exit handler. Both are low-probability under the current ~30 s abort timing but cheap to harden. Assigning **P2** — it resolves a complete, silent failure of a shipped update path on macOS, but users can still update by downloading the release manually and the blast radius is Apple-Silicon-only.

PR #744, opened 2026-08-01
chore(release): 3.38.3

openwong2kim:
Patch release. Folds the three pending changelog fragments, bumps the version, and regenerates the API reference header.

### Fixed in 3.38.3

- The Schedules panel's **Add schedule** button no longer gets clipped off the right edge of a narrow deck. (#740)
- Wrong glyphs no longer scatter across several panes at once during CJK-heavy sessions. The shared WebGL glyph atlas's page-merge path corrupted sibling panes' cached glyph references once its page pool filled. (#741)
- **Check for updates** now installs the update on Apple Silicon Macs, instead of downloading it and then doing nothing, silently, on that press and every press after it. (#742)

### Release mechanics

- `changelog.d/{740,741,742}.md` folded into `CHANGELOG.md` under `## [3.38.3] — 2026-08-02`
- `package.json` / `package-lock.json` 3.38.2 → 3.38.3 (lock: the two version fields only, dependency tree untouched)
- `docs/api/reference.md` regenerated so the baked version matches, per the CI drift guard

Full suite: 9585 passed, 0 failed.

After merge, push the `v3.38.3` tag onto the release commit to trigger the installer builds.


openwong2kim:

Patch-release PR that cuts **3.38.3** — version-only mechanics with no source code touched. It folds the three pending `changelog.d/` fragments (#740 Schedules-panel button clipping, #741 shared WebGL glyph-atlas page-merge corruption under CJK-heavy sessions, #742 broken *Check for updates* on Apple Silicon) into `CHANGELOG.md` under a new `## [3.38.3] — 2026-08-02` heading, bumps `package.json` / `package-lock.json` 3.38.2 → 3.38.3 (the two version fields only, dependency tree untouched), and regenerates `docs/api/reference.md` so its baked version string matches and the CI drift guard stays green. Seven files, +27/−26, single commit.

Nothing here can reach runtime behaviour, so risk is confined to the release mechanics themselves: the CHANGELOG section ordering, the fragment deletions matching what was folded, and the `reference.md` regeneration lining up with `gen-api-reference.mjs`. At the time of review CI was still running (macOS/Ubuntu baselines and CodeRabbit already green; `validate`, windows-latest baseline, and `bench` in flight) — those should confirm before the `v3.38.3` tag is pushed.

**Priority: P3.** Routine release chore — approve and merge once the remaining checks pass, then push the `v3.38.3` tag onto the release commit to trigger the installer builds.

Issue #747, opened 2026-08-02
A resize while a selection is live leaves xterm and the PTY stuck at the old cols/rows


openwong2kim:
### What happened

`shouldFitWhilePreservingSelection` skips `fitAddon.fit()` while the terminal has
an active selection, so a resize mid-drag cannot clear what the user is selecting.
The guard's contract, stated in `src/renderer/utils/fitGuard.ts`, is:

> Skipping `fit()` while the user has an active selection prevents the clear;
> **the next ResizeObserver tick (after the user releases) handles the deferred
> resize.**

There is no such tick. `ResizeObserver` fires on *size change*, not on selection
release. If the container resizes while a selection is live and nothing resizes
again afterward, the deferred `fit()` never runs: xterm keeps the old
`cols`/`rows`, and because `sendResize` is downstream of `fit()`, the daemon PTY
keeps them too.

The result is a terminal permanently mis-sized against its container until
something else happens to resize it. Output wraps at the wrong column, and
full-screen TUIs (Claude Code, vim, less) draw against stale dimensions.

### Reproduction

1. Open two panes so one can be resized independently.
2. In pane A, click-drag to select some output and **hold the selection**
   (release the mouse but do not click elsewhere — `hasSelection()` stays true).
3. While the selection is live, change pane A's size: drag the split separator,
   toggle the sidebar (`Ctrl+B`), or resize the window.
4. Click into pane A to clear the selection. Do not resize anything else.

Pane A's xterm and its PTY are still at the pre-resize `cols`/`rows`. `stty size`
inside that pane reports the old dimensions. Type a long line and it wraps at the
wrong column.

### Where

- `src/renderer/utils/fitGuard.ts` — the guard, and the docstring asserting a
  follow-up tick that does not exist.
- `src/renderer/hooks/useTerminal.ts:2009` — the `ResizeObserver` path that skips
  and returns without recording that a fit is owed.
- `src/renderer/hooks/useTerminal.ts:1183` — the `fonts.ready` path, same skip.
  Its comment already calls itself "mostly defensive", so it matters less.
- `src/renderer/hooks/useTerminal.ts:1209` — `onSelectionChange` exists but only
  drives auto-copy. It is the natural place to notice the release.

### Suggested fix

Record that a fit was skipped (a `pendingFit` flag beside the existing
`resizeDebounceTimer` / `lastSentCols` locals), and in the `onSelectionChange`
handler, when the selection becomes empty and `pendingFit` is set, run the same
[...]

Issue #645, opened 2026-07-27
[Feature]: Move Panes


altmank:
### Problem / motivation

Needs ability to move panes around to a new layout manually in UI

### Proposed solution

Add support for moving panes or dragging into new layouts. 

### Confirmation

- [x] I searched existing issues and Discussions and this is not a duplicate.

openwong2kim:
## [wmux-hermes] Triaged — P3

This is a feature request to allow users to manually reposition panes (drag/move) into a new layout from within the wmux UI, filed by `@altmank`. The current tmux-style pane model lets you split, close, and resize panes, but it lacks an affordance for re-parenting or re-flowing an existing pane into a different container/position, so the only way to get a desired arrangement today is to split in the right order from scratch. Adding move/drag would require work on the pane tree data structure in `src/main/pty` (or wherever the layout model lives), the renderer layout component, and probably an IPC handler to commit the re-parent operation atomically; it is a non-trivial change but purely UX, with no security or correctness implications. Marking **P3** since there are active security/design initiatives (#613, #615, #604/#605) that should land first, and a workaround (re-split) exists today.

**Suggested next step:** scope a pane re-parent IPC + a drag interaction on the layout container, and decide whether to align with tmux semantics (`break-pane` / `join-pane`) or a free-form grid model.

openwong2kim:
Noticed the fork — are you planning to take a crack at this yourself? If so, that'd be very welcome and I'm happy to point you at the relevant code before you start (the pane layout tree lives in a few specific places, and there are a couple of invariants worth knowing about upfront so the first attempt doesn't fight them).

If you'd rather leave the implementation to us, that's fine too — it's triaged P3, so it would land behind current work either way. Just let me know which way you'd like to go so we don't duplicate effort.

PR #748, opened 2026-08-02
Let the multiview grid be arranged vertically

### LINKED ISSUE #746: [Feature]: Let the multiview grid be arranged vertically (stacked) instead of always 2×2
### Problem / motivation

The workspace multiview grid picks its shape automatically and offers no way to
override it. The column count is hardcoded in `WorkspaceViewport.tsx`:

- 2 workspaces → 2 columns (side by side)
- 3–4 workspaces → 2 columns (so 4 becomes a 2×2 grid)
- 5+ → 3 columns

On a tall or portrait display — or when the workspaces are mostly agent
transcripts, which read better tall and narrow — a 2×2 grid is the wrong shape.
There is no setting, no keybinding, and no drag affordance to get a single
column of stacked tiles instead. `Ctrl+Shift+G` only clears multiview entirely.

### Proposed solution

Let the user pin the multiview arrangement instead of always inferring it. The
smallest useful version is a layout mode with three values:

- `auto` — today's behavior, stays the default
- `columns` — one row, tiles side by side
- `rows` — one column, tiles stacked vertically

Surfaced as a toggle on the multiview chrome (and/or a keybinding), persisted
with the rest of the UI state.

Two places must change together:

1. `src/renderer/components/Layout/WorkspaceViewport.tsx` — the
   `gridTemplateColumns` expression that currently hardcodes 2/3 columns.
2. `src/renderer/stores/slices/uiSlice.ts` — `focusMultiviewDirection`
   duplicates the column count (`const cols = ids.length <= 4 ? 2 : 3`) to walk
   the grid for directional focus. If only the CSS changes, arrow-key focus
   movement breaks under any non-default arrangement. The column count should be
   derived once and shared by both.

### Alternatives considered

- Deriving the shape from the viewport aspect ratio instead of an explicit
  setting. Less control, and it would still need the same shared column-count
  fix in `focusMultiviewDirection`.
- Folding this into #645 (Move Panes). Different surface — #645 is about
  rearranging panes *inside* a workspace tree; this is the workspace tile grid
  above it. They share no code, and pairing a ~30-line change with a
  drag-and-drop feature would hold the small one back.

### Additional context

Related but distinct: #645.

### Confirmation

openwong2kim:
Closes #746.

## The problem

The multiview grid inferred its shape and offered no override. Two workspaces
became two columns, three or four became a 2×2 grid, five or more became three
columns. On a tall display — or when the tiles are agent transcripts, which read
better narrow — 2×2 is the wrong shape, and there was no way to ask for a single
stacked column. `Ctrl+Shift+G` only cleared multiview entirely.

## What this adds

A persisted `multiviewArrangement` preference with three values, set from a
segmented control in Settings → Layout, next to Sidebar position:

- **Auto** — the previous behavior, still the default
- **Columns** — one row, tiles side by side
- **Rows** — one column, tiles stacked

## Notes for review

**The column count was duplicated.** `uiSlice.focusMultiviewDirection` had its own
copy (`ids.length <= 4 ? 2 : 3`) to walk the grid for `Ctrl+Shift+Arrow`. Changing
only the CSS would have left arrow navigation moving through a layout that is no
longer on screen. It now lives in `multiviewColumnCount()` and both callers read
it.

**The default is unchanged, and the tests pin that.** An earlier revision of this
branch applied the new `minmax()` floors to `auto` as well, which would have given
every existing user on a narrow window a scrollbar, hidden tiles, and a different
PTY size purely by upgrading. Floors and scrolling now apply to `columns`/`rows`
only, and `multiviewGridStyle()` is asserted on the exact track strings rather
than the column count — counting columns is what let the regression through in the
first place.

**Explicit arrangements get a size floor rather than unbounded shrink.** `columns`
with six tiles on a 1400px window would otherwise produce 229px tiles, about 27
terminal columns, where a TUI wants 80. The tiles stop at 320px and the grid
scrolls instead. An explicit arrangement never silently degrades back to `auto`;
it just stops producing unusable terminals. Arrow-focus onto a scrolled-out tile
brings it into view, and that effect re-runs on arrangement changes too, since
switching `auto`→`columns` can push the active tile out of view without the tile's
own active state ever changing.

**The session read-back is whitelisted**, not the bare truthiness check
`sidebarPosition` uses. That matches the rule stated a few lines below it in the
same file: a forward-version session file naming a fourth arrangement must not
park an unknown string in the store, where the settings control would render with
nothing selected.

## Testing

- `multiviewGrid.test.ts` (new, 8 cases) — `auto` reproduces the legacy 2/2/3
  progression *and* emits bare `1fr` tracks with no scroll container; `rows` and
  `columns` floor and scroll; an unrecognized arrangement degrades to `auto`.
- `uiSlice.test.ts` — arrow navigation walks correctly under `rows` and `columns`,
  including that the cross-axis arrows are no-ops.
- `workspaceSlice.loadSession.test.ts` — the preference round-trips, and an unknown
  value leaves the default intact.
- Full renderer suite: 2756 passing across 218 files. `tsc --noEmit` and `eslint`
  clean.

Manual: verified all three arrangements at 2/4/6 tiles, arrow navigation in each,
scroll-into-view on off-screen tiles, and that the choice survives a restart.

## Out of scope

- #645 (Move Panes) — that rearranges panes *inside* a workspace tree; this is the
  workspace tile grid above it. No shared code.
- #747 — a resize while a selection is live leaves the PTY at stale dimensions.
  Pre-existing in the terminal resize path; switching arrangements is one more way
  to trigger it. Filed separately.

## Summary by CodeRabbit

- **New Features**
  - Added Appearance → Layout controls for arranging multiview tiles automatically, by columns, or by rows.
  - Multiview navigation, grid sizing, scrolling, and active-tile visibility now adapt to the selected arrangement.
  - Selected layout preferences are restored when reopening a session.

- **Bug Fixes**
  - Invalid or unsupported saved layout values safely fall back to automatic arrangement.

- **Tests**
  - Added coverage for multiview layouts, navigation, sizing, overflow, and session restoration.


openwong2kim:

**Priority: P3** — implements the multiview arrangement preference from #746.

This PR closes #746 cleanly. It adds a persisted `multiviewArrangement` preference (`auto` | `columns` | `rows`) surfaced as a segmented control in Settings → Layout, and the default (`auto`) is kept byte-identical to the pre-existing grid so no existing user gets a layout or PTY-size change on upgrade. The strongest part is that it collapses the duplicated column count into a single `multiviewColumnCount()` helper that both `WorkspaceViewport.tsx` (the grid CSS) and `uiSlice.focusMultiviewDirection` (Ctrl+Shift+Arrow navigation) now read — this is exactly the shared-source fix the issue asked for, and it closes the real failure mode where changing only the CSS would have left arrow navigation walking through a layout no longer on screen.

The diff is disciplined: explicit arrangements get a tile size floor (320×200px) with scroll instead of unbounded shrink, the session read-back is whitelisted against a forward-version fourth mode rather than using a bare truthiness check, and the test suite pins the exact `1fr` track strings for `auto` plus directional nav under `rows`/`columns` (8 new cases + round-trip persistence). The scroll-into-view on arrangement change is a thoughtful touch, since switching `auto`→`columns` can strand the active tile off-screen without its `isActive` ever flipping. Ready to merge — no blockers from triage.

PR #749, opened 2026-08-02
fix(mcp): scope browser RPC fallbacks by workspace

### LINKED ISSUE #695: browser tools' RPC fallback still selects a webview target workspace-blind (the gap #679 left open)

snowyukitty:
## Summary

Scope MCP browser-tool RPC fallbacks and automation leases to the calling session's resolved workspace, closing the cross-workspace target-selection gap left after #679. Browser operations now fail closed when ownership cannot be proven while retaining compatibility with current and supported legacy main processes.

## Changes

- Resolve one immutable browser target scope per tool invocation and reuse it for initial/late automation lease acquisition, page discovery, navigation, auto-open, and every fallback RPC.
- Require scoped Playwright page lookup, validate legacy target ownership and CDP target IDs, and preserve only the narrowly-proven legacy auto-open path.
- Add typed workspace-scope refusals plus source-level and behavioral regression tests across all browser tool groups, including browser_fill's click/evaluate/type fallback sequence.

### Fixed

- **Scoped MCP browser fallbacks to their calling workspace.** Browser tools no longer let lease acquisition, CDP discovery, or RPC fallback select a webview from whichever workspace happens to be active when the caller's target is missing or ambiguous. (#749)

## Stability tier impact

- [ ] Additive change to a stable surface (new optional param, new field, new event type).
- [ ] Breaking change to a stable surface (requires major bump — block until release planning is in scope).
- [x] Change to an experimental surface (note in release notes).
- [ ] Change to an internal surface (no contract impact).

The browser/CDP RPC and MCP tool family is experimental per docs/api/inventory.md.

n/a — no RPC method, event type, capability, or public wire shape is added. Existing experimental browser RPCs now consistently receive the already-supported workspaceId.

## Test plan

- npx tsc --noEmit — passed.
- npm run build:mcp — passed.
- MCP workspace/browser regression suite — 14 files, 157 tests passed.
- Main-side browser ownership suite — 2 files, 94 tests passed.
- npm run test:parallel — 655 files / 9,627 tests passed; 2 files / 27 tests skipped.
- npm run test:runtime — 96 passed, 6 skipped, 2 unrelated PowerShell OSC 133 tests failed because local node-pty could not AttachConsole. Re-running that file reproduced the environment failure; this PR does not touch daemon or shell-integration code.
- Exact-digest independent review — Claude Opus, Grok 4.5, and Gemini 3.1 Pro High all approved.

## Related issues / PRs

Closes #695.

Main-side ownership enforcement is already present from #701.

## Reviewer checklist

- [ ] No accidental commit of secrets, tokens, .env, or unrelated large binaries.

## Summary by CodeRabbit

* **Bug Fixes**
  * Browser automation now remains scoped to the calling workspace across navigation, interaction, inspection, extraction, file, state, utility, and wait operations.
  * Prevented fallback discovery and RPC requests from selecting another workspace’s browser surface.
  * Ambiguous, missing, unverified, or unresolved browser targets now fail safely instead of being used.
* **Tests**
  * Expanded coverage for workspace routing, target ownership, automation leases, fallback behavior, and browser tool registration.
* **Documentation**
  * Added a changelog entry describing workspace-scoped browser fallbacks.

### DIFF
--- a/changelog.d/749.md
+++ b/changelog.d/749.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- **Scoped MCP browser fallbacks to their calling workspace.** Browser tools no
+  longer let automation leases, CDP discovery, or RPC fallbacks select another
+  workspace's webview when the intended target is missing or ambiguous. (#749)
--- a/src/mcp/__tests__/workspaceRouting.test.ts
+++ b/src/mcp/__tests__/workspaceRouting.test.ts
@@ -22,8 +22,9 @@ import path from 'node:path';
 //      browser.rpc.ts; the handler ignores workspaceId). It carries NO workspace
 //      identity, matching browser_session_stop/status/list — so it can never
 //      reintroduce the active-workspace fallback bug.
-//   4. browser_tabs injects requireWorkspaceId() at its production registration
-//      site, so its isolated unit tests cannot mask accidental weak wiring.
+//   4. Every Playwright browser tool injects requireWorkspaceId() at its
+//      production registration site, so isolated unit tests cannot mask
+//      accidental weak wiring or an unscoped fallback/lease path.
 describe('MCP workspace routing (source-level invariants)', () => {
   const indexPath = path.join(__dirname, '..', 'index.ts');
   const rawSrc = fs.readFileSync(indexPath, 'utf-8');
@@ -179,15 +180,27 @@ describe('MCP workspace routing (source-level invariants)', () => {
     expect(src).toMatch(/setWorkspaceIdResolver\(\s*requireWorkspaceId\s*\)/);
   });
 
-  it('browser_tabs production wiring injects requireWorkspaceId (#565)', () => {
-    // navigation.tabs.test.ts injects a resolver so it can exercise the tool in
-    // isolation. This source lock covers the complementary production seam: a
-    // future rewire to the weak resolver must fail even if those unit tests pass.
-    const registrations = src.match(/registerNavigationTools\([\s\S]*?\);/g) ?? [];
-    expect(registrations).toHaveLength(1);
-    expect(registrations[0]).toMatch(
-      /registerNavigationTools\(\s*server\s*,\s*\{\s*resolveWorkspaceId:\s*requireWorkspaceId\s*\}\s*\)/,
+  it('every Playwright browser tool receives the same strict workspace resolver (#695)', () => {
+    // Unit tests inject a resolver in isolation. This source lock covers the
+    // complementary production seam: every browser registration must share the
+    // strict requireWorkspaceId dependency, never the weak/UI-active resolver.
+    expect(src).toMatch(
+      /const browserToolDeps\s*=\s*\{\s*resolveWorkspaceId:\s*requireWorkspaceId\s*\}/,
     );
+    for (const name of [
+      'Navigation',
+      'Interaction',
+      'Inspection',
+      'State',
+      'Wait',
+      'File',
+      'Utility',
+      'Extraction',
+    ]) {
+      expect(src, `${name} tools must receive browserToolDeps`).toMatch(
+        new RegExp(`register${name}Tools\\(\\s*server\\s*,\\s*browserToolDeps`),
+      );
+    }
   });
 
   it('browser_session_start is GLOBAL — carries no workspace identity (no resolver calls)', () => {
@@ -231,7 +244,7 @@ describe('MCP workspace routing (source-level invariants)', () => {
 
   it('browser_wait is wired through the same immutable catalog profile', () => {
     expect(src).toMatch(
-      /registerWaitTools\(\s*server\s*,\s*MCP_CATALOG_OPTIONS\s*\)/,
+      /registerWaitTools\(\s*server\s*,\s*browserToolDeps\s*,\s*MCP_CATALOG_OPTIONS\s*\)/,
     );
   });
 });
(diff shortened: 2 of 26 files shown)


openwong2kim:

This PR closes #695 — the cross-workspace browser target-selection gap that #679 deliberately left open. After #679, a caller whose workspace identity could not be resolved was still handed a page on the RPC fallback path, because every browser tool swallowed the engine's `getPage()` refusal with `.catch(() => null)` and then selected a webview workspace-blind via `WebviewCdpManager.sessions.values().next()`. This change closes that hole by resolving one immutable `BrowserTargetScope` per tool invocation and threading it through lease acquisition, page discovery, navigation, auto-open, and every fallback RPC, so a refusal at the engine layer now propagates instead of degrading into a workspace-blind fallback.

The design is disciplined. A new `browserScope.ts` introduces `WorkspaceScopeUnresolvedError` (typed, with a stable code), `requireBrowserTargetScope()` (resolves identity once and freezes it), `sendScopedBrowserRpc()` (scope is authoritative over caller-supplied params, so a helper cannot silently drop or override `workspaceId`), and `allowScopedRpcFallback()` (replaces the blind `.catch(() => null)` — it re-throws scope refusals while still permitting ordinary page-discovery failures to fall back). The public engine entry narrows from `getPage(surfaceId?)` to `getPageForScope(scope)`, making a missing `workspaceId` a compile error at every tool call site; a source-level invariant test (`workspaceRouting.test.ts`) locks that every `register*Tools` receives the same strict `browserToolDeps`. Legacy-main compatibility is handled honestly: `selectRegisteredTarget()` inspects ownership tags and refuses foreign or untagged targets rather than trusting the requested id, while a newly-auto-opened surface is marked proven-by-construction so it survives a legacy main that does not tag its own response.

Coverage is thorough — 14 new/updated MCP test files (157 tests) and a main-side ownership suite (94 tests), plus the full parallel suite at 9,627 passing. One thing to watch on review: the `matchPinnedPage` change now requires a `Target.getTargetInfo` round-trip for every candidate (the single-URL fast path is gone), which is a correctness-for-cost trade that is the right call here but worth noting for any latency-sensitive caller. Assigning **P2** — it closes an important security-adjacent issue within the same-user ceiling, and the approach is the right shape for the codebase; CI checks were still running at review time, so confirm they are green before merge.

PR #750, opened 2026-08-02
fix(terminal): retry deferred fits after selection clears

### LINKED ISSUE #747: A resize while a selection is live leaves xterm and the PTY stuck at the old cols/rows

snowyukitty:
## Summary

Retry terminal fits that were skipped to preserve an active xterm selection. This prevents xterm and the backing PTY from remaining at stale dimensions after the selection clears.

## Changes

- Route ResizeObserver fitting through a debounced scheduler that records selection-blocked attempts and retries them from `onSelectionChange`.
- Reuse the existing scroll-preserving fit and deduplicated PTY resize path for deferred work, including fits skipped by font/theme and visibility guards.
- Add focused scheduler race/teardown tests and source-level wiring guards for the xterm-bound hook.

### Fixed

- **Terminal dimensions recover after protected selections.** Resizing a pane while text remained selected could leave xterm and the PTY at stale dimensions until another resize; clearing the selection now runs the deferred fit.

## Stability tier impact

n/a

## Test plan

- `npx vitest run src/renderer/utils/__tests__/deferredTerminalFit.test.ts src/renderer/utils/__tests__/fitGuard.test.ts src/renderer/hooks/__tests__/useTerminal.deferredFit.test.ts` — 14 passed.
- `npx vitest run src/renderer` — 2,754 passed across 219 files.
- `npm run test:parallel` — 9,621 passed, 27 skipped across 656 files.
- `npx tsc --noEmit` — passed.
- Targeted ESLint for the new scheduler/tests and fit guard — passed.
- `npm run test:runtime` — 96 passed, 6 skipped, 2 unrelated `shellIntegration.runtime.test.ts` cases timed out after the Windows `node-pty` helper reported `AttachConsole failed` on local Node 24.18.0.
- `npm run lint` — not clean on current `main` baseline (218 existing errors); no errors in the targeted new helper/test files.
- Manual Electron reproduction — not run locally; CI and maintainer smoke testing remain useful for the xterm interaction.

## Related issues / PRs

Closes #747.

## Reviewer checklist

### DIFF
--- a/changelog.d/750.md
+++ b/changelog.d/750.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- **Terminal dimensions recover after protected selections.** Resizing a pane
+  while text remained selected could leave xterm and the PTY at stale
+  dimensions until another resize; clearing the selection now runs the
+  deferred fit. (#750)
--- a/src/renderer/hooks/__tests__/useTerminal.deferredFit.test.ts
+++ b/src/renderer/hooks/__tests__/useTerminal.deferredFit.test.ts
@@ -1,0 +1,27 @@
+import fs from 'node:fs';
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+const source = fs.readFileSync(
+  path.resolve(process.cwd(), 'src/renderer/hooks/useTerminal.ts'),
+  'utf8',
+);
+
+describe('useTerminal deferred resize fit wiring', () => {
+  it('routes ResizeObserver work through the deferred fit scheduler', () => {
+    expect(source).toMatch(
+      /const resizeObserver = new ResizeObserver\(\(\) => \{\s*resizeFit\.requestFit\(\);\s*\}\);/,
+    );
+  });
+
+  it('retries pending resize work from the selection-change callback', () => {
+    expect(source).toMatch(
+      /terminal\.onSelectionChange\(\(\) => \{[\s\S]*?resizeFit\.onSelectionChange\(\);\s*\}\);/,
+    );
+  });
+
+  it('records every non-observer fit skipped for an active selection', () => {
+    const deferredCalls = source.match(/deferUntilSelectionClears\(\)/g) ?? [];
+    expect(deferredCalls).toHaveLength(3);
+  });
+});
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -15,6 +15,11 @@ import { openTerminalUrl } from '../utils/browserPaneActions';
 import { runCopyWithFeedback } from '../utils/copyWithFeedback';
 import { shouldFitWhilePreservingSelection } from '../utils/fitGuard';
 import { createAutoSelectionCopy } from '../utils/autoSelectionCopy';
+import {
+  createDeferredTerminalFit,
+  type DeferredTerminalFitHandle,
+  type DeferredTerminalFitOutcome,
+} from '../utils/deferredTerminalFit';
 import { decodeOsc52Write } from '../utils/osc52Clipboard';
 import { terminalFontFamilyCss } from '../utils/terminalFont';
 import { createPathLinkProvider } from '../terminal/pathLinkProvider';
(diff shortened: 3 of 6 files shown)


openwong2kim:

This PR fixes #747 — the bug where resizing a pane while an xterm selection was live left both xterm and the backing PTY stuck at stale `cols`/`rows`, because `ResizeObserver` never fires merely from releasing a selection. It introduces a new `createDeferredTerminalFit` scheduler (`src/renderer/utils/deferredTerminalFit.ts`) that takes over the resize-fit lifecycle: it debounces fit attempts (replacing the inline timer that previously lived in `useTerminal.ts`), records fits that `shouldFitWhilePreservingSelection` blocks, and retries them from `terminal.onSelectionChange` once the selection clears. The same debounced, scroll-preserving fit + `sendResize` path now serves every caller — the `ResizeObserver`, `fonts.ready`, font/theme switch, and workspace/tab visibility paths — and the three non-observer sites explicitly record a deferred fit via `deferUntilSelectionClears()` so they no longer drop the work on the floor.

The fix also corrects the misleading docstring in `fitGuard.ts` that used to assert a follow-up `ResizeObserver` tick existed. Test coverage is strong: 223 lines of unit tests for the scheduler itself (debounce coalescing, supersession, selection-deferred retry, re-selection mid-retry, skipped-fits-not-retried, teardown) plus source-level wiring guards in `useTerminal.deferredFit.test.ts`. The refactor is well-scoped to internal surfaces and the default behavior is unchanged — this is a clean, reviewable fix for a real correctness bug. Assigning **P2**, matching #747.

snowyukitty:
Superseded by #754, which fixes #747 and has now merged. Closing this alternative implementation to avoid duplicate work.

PR #753, opened 2026-08-02
Keep multiviewIds consistent when members go away

### LINKED ISSUE #751: Closing a multiview member from the sidebar leaves a stale id, so the grid can render a single tile
### What happens

`removeWorkspace` splices the workspace out of `state.workspaces` but never prunes
`state.multiviewIds` (`src/renderer/stores/slices/workspaceSlice.ts`, the
`removeWorkspace` action). A workspace that was a multiview member survives there
as an id with no workspace behind it.

The multiview render gate then disagrees with what it can actually draw:

```ts
// src/renderer/components/Layout/WorkspaceViewport.tsx
if (multiviewIds.length >= 2 && multiviewIds.includes(activeWorkspaceId)) {
  const tiles = multiviewIds
    .map((id) => workspaces.find((w) => w.id === id))
    .filter((ws): ws is Workspace => ws !== undefined);
```

The gate counts `multiviewIds` (which includes the ghost); `tiles` drops it. With
exactly one live member plus one ghost, the gate passes and the grid renders a
**single tile** — complete with the multiview tile header and its X button —
instead of falling back to the normal single-workspace view. It stays that way
until the user clears multiview with `Ctrl+Shift+G`.

### Reproduction

1. Put three workspaces in multiview (`Ctrl`/`Cmd`+click them in the sidebar).
2. Close one of them from the **sidebar** (the X on the sidebar row, or
   `Ctrl+Shift+W` while it is active) — not from the multiview tile's X.
3. Close a second one the same way.

One live member remains, but `multiviewIds` still holds three entries, so the app
renders a one-tile grid with multiview chrome rather than the plain single view.

### Why it did not show up before

Nothing consumed `multiviewIds` in a way that made the ghost visible.
`focusMultiviewDirection` used to walk the raw list too, so both halves were
wrong in the same direction. PR #748 made the grid's track count read the
filtered `tiles`, and CodeRabbit caught that the nav had to match — so nav and
tracks now agree, but the **gate** still counts the unfiltered list.

### Suggested fix

Prune on removal — in `removeWorkspace`, drop the id from `multiviewIds` inside
the same transaction that splices the workspace, and clear the list entirely if
fewer than two members remain (matching what `removeMultiviewWorkspace` already
does). Changing the gate to `tiles.length >= 2` would hide the symptom, but the
stale id would still be in the store for any future consumer to trip over.

Note the deliberate comment in `loadSession`'s neighborhood about not parking
unknown ids in the store — this is the same class of problem from the other end.

### Notes

Pre-existing, not a regression. Found while dogfooding PR #748. Not fixed there:
that PR defends the two paths it touches (grid tracks and arrow navigation) but
the pruning belongs in the workspace-removal path, not the layout code.

### LINKED ISSUE #752: Ctrl+click on the active workspace collapses the whole multiview grid
### What happens

`Ctrl`/`Cmd`+clicking a workspace in the sidebar toggles its multiview membership.
Doing it on the **currently active** workspace removes it from the grid — and
because the render gate requires the active workspace to be a member, the entire
grid collapses to a single view. Every other tile disappears at once.

From the user's side this reads as "I clicked one workspace and the whole
multiview reset."

### Reproduction

1. Four workspaces, Workspace 4 active.
2. `Cmd`+click Workspace 1, 2, 3 — a 2×2 grid appears.
3. `Cmd`+click **Workspace 4** (the active one, which the toggle implicitly added
   as the grid's seed).

Expected: Workspace 4 leaves the grid, the other three stay tiled.
Actual: the grid vanishes; the app shows Workspace 4 full-screen. Workspaces 1–3
are still in `multiviewIds` but nothing is rendered from them.

### Why this looks like an oversight rather than a design choice

The tile's own X button already handles exactly this case, with a comment saying
so — `WorkspaceViewport.handleRemoveTile`:

```ts
// Close a multiview tile. If closing the ACTIVE tile with others remaining,
// hand focus to a neighbor first — otherwise the grid gate
// (multiviewIds.includes(activeId)) fails the next render and the view
// collapses to the just-closed workspace (reads as "the window reset").
```

It hands focus to a neighbour before removing. The sidebar path does not — it
calls the toggle directly:

```ts
// src/renderer/components/Sidebar/Sidebar.tsx
const handleCtrlSelect = useCallback((wsId: string) => {
  toggleMultiviewWorkspace(wsId);
}, [toggleMultiviewWorkspace]);
```

Two entry points for the same intent, one of which knows about the failure mode.

### Suggested fix

Give the sidebar path the same handoff, or better, move it into the store so
both callers get it: when the toggle would remove the active workspace and two or
more members would remain, activate a neighbouring member first. `handleRemoveTile`
could then collapse to a plain call to the store action instead of duplicating the
logic in the view.

### Notes

Pre-existing, not a regression. Found while dogfooding PR #748 — the driver script
hit it by Cmd+clicking every sidebar row including the active one, and the grid
silently closed.

openwong2kim:
Closes #751. Closes #752.

Both were found by dogfooding #748 in a packaged build, not by the test suite.
Both are pre-existing: the multiview group could describe workspaces that were
not, or could not be, on screen.

## #751 — a closed member left a stale id, so the grid could render one tile

`removeWorkspace` spliced the workspace out of `workspaces` but never touched
`multiviewIds`. The render gate counts the raw list while the tiles are filtered
against live workspaces:

```ts
if (multiviewIds.length >= 2 && multiviewIds.includes(activeWorkspaceId)) {
  const tiles = multiviewIds
    .map((id) => workspaces.find((w) => w.id === id))
    .filter((ws): ws is Workspace => ws !== undefined);
```

One live member plus one stale id passes the gate and draws a **single tile**
with full multiview chrome — header, X button — instead of falling back to the
normal single-workspace view. It stays that way until `Ctrl+Shift+G`.

Reproduce: put three workspaces in multiview, then close two of them from the
sidebar (not from the tile's X).

The id is now dropped inside the same transaction that removes the workspace,
and the group clears when fewer than two members would remain.

This does not contradict the `multiviewIds is intentionally preserved` note in
`setActiveWorkspace` — that is about navigating *away* from a group, which the
user can undo by clicking a member. A removed workspace cannot come back, so
keeping its id serves nobody.

## #752 — Ctrl+click on the active workspace closed the whole grid

The gate needs the active workspace to be a member, so toggling it off took every
other tile with it. It reads as "I clicked one workspace and multiview reset."

The tile's own X button already handled this, with a comment naming the failure:

```ts
// Close a multiview tile. If closing the ACTIVE tile with others remaining,
// hand focus to a neighbor first — otherwise the grid gate
// (multiviewIds.includes(activeId)) fails the next render and the view
// collapses to the just-closed workspace (reads as "the window reset").
```

The sidebar path called `toggleMultiviewWorkspace` raw and had no equivalent —
two entry points for one intent, one of which knew about the trap.

The handoff moves into `toggleMultiviewWorkspace` and `removeMultiviewWorkspace`.
`WorkspaceViewport` drops its copy and wires the X straight to the store action,
so the two paths cannot drift apart again. No handoff happens when the group would
collapse anyway (two members), since reassigning focus there would yank the user
to a workspace they did not ask for.

## Testing

- `uiSlice.test.ts` — toggling off the active member activates a neighbour and the
  group survives; the same for the X-button action; falls back to the previous
  member when the active tile is last; no handoff when the group collapses anyway.
- `workspaceSlice.test.ts` — closing a member prunes it; closing down to one member
  clears the group; closing a non-member leaves it alone.
- Full renderer suite: 2764 passing across 218 files. `tsc --noEmit` and `eslint`
  clean.

The pruning is written as `state.multiviewIds?.indexOf(id) ?? -1` to survive test
stores that mount `workspaceSlice` without `uiSlice`, matching the existing guard
in `toggleMultiviewWorkspace`'s cold-park block.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved focus handling when removing or toggling the active multiview tile.
  * Automatically moves focus to an adjacent tile, with reliable fallback behavior.
  * Preserves multiview membership for remaining tiles during removal.
  * Removes closed workspaces from multiview groups.
  * Automatically clears multiview mode when fewer than two tiles remain.
  * Prevents closing the entire grid when removing the active tile via Ctrl+click.

### DIFF
--- a/changelog.d/753.md
+++ b/changelog.d/753.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **Closing a workspace no longer leaves multiview in a broken state.** Closing
+  members of a multiview group from the sidebar could leave the app showing a
+  single tile still wrapped in multiview chrome, with no way back except
+  `Ctrl+Shift+G`. The closed workspace is now dropped from the group, and the
+  group disbands once fewer than two members remain. (#753)
+- **Ctrl+click on the workspace you are looking at no longer closes the whole
+  grid.** Removing the active workspace from a multiview group took every other
+  tile with it. Focus now moves to a neighbouring tile first, the way the tile's
+  own ✕ button already behaved. (#753)
--- a/src/renderer/components/Layout/WorkspaceViewport.tsx
+++ b/src/renderer/components/Layout/WorkspaceViewport.tsx
@@ -68,12 +68,12 @@ const MultiviewWorkspaceSlot = memo(function MultiviewWorkspaceSlot({
 }: {
   workspace: Workspace;
   isActive: boolean;
-  /** Count of multiview members — the close handler needs it to hand off focus. */
+  /** Count of multiview members — a change here can reflow this tile out of view. */
   multiviewCount: number;
   /** Current grid arrangement — a layout change can move this tile out of view. */
   arrangement: MultiviewArrangement;
   onActivate: (id: string) => void;
-  onRemove: (id: string, isActive: boolean, multiviewCount: number) => void;
+  onRemove: (id: string) => void;
 }) {
   // An explicit arrangement lets the grid scroll (see multiviewGrid.ts), so a
   // tile reached by Ctrl+Shift+Arrow can be off-screen. `nearest` scrolls the
@@ -109,7 +109,7 @@ const MultiviewWorkspaceSlot = memo(function MultiviewWorkspaceSlot({
         <button
           onClick={(e) => {
             e.stopPropagation();
-            onRemove(workspace.id, isActive, multiviewCount);
+            onRemove(workspace.id);
           }}
           className="ml-auto opacity-60 hover:opacity-100"
           style={{
@@ -153,18 +153,11 @@ export function WorkspaceViewport() {
   const removeMultiviewWorkspace = useStore((s) => s.removeMultiviewWorkspace);
   const t = useT();
 
-  // Close a multiview tile. If closing the ACTIVE tile with others remaining,
-  // hand focus to a neighbor first — otherwise the grid gate
-  // (multiviewIds.includes(activeId)) fails the next render and the view
-  // collapses to the just-closed workspace (reads as "the window reset").
-  const handleRemoveTile = (id: string, isActive: boolean, count: number) => {
-    if (isActive && count > 2) {
-      const removedIdx = multiviewIds.indexOf(id);
-      const nextActive = multiviewIds[removedIdx + 1] ?? multiviewIds[removedIdx - 1];
-      if (nextActive) setActiveWorkspace(nextActive);
-    }
-    removeMultiviewWorkspace(id);
-  };
+  // The focus handoff that used to live here — activate a neighbour before
+  // removing the active tile, so the grid gate doesn't collapse the whole view —
+  // moved into removeMultiviewWorkspace (#752). The sidebar's Ctrl+click path
+  // had no equivalent and lost the grid; keeping one copy in the store is what
+  // stops the two from drifting apart again.
 
   // Fix 0 — while startup reconcile is in flight, show a placeholder. Chrome
   // stays mounted (it's AppLayout, not here) so the user sees wmux is alive.
@@ -204,7 +197,7 @@ export function WorkspaceViewport() {
             multiviewCount={multiviewIds.length}
             arrangement={multiviewArrangement}
             onActivate={setActiveWorkspace}
-            onRemove={handleRemoveTile}
+            onRemove={removeMultiviewWorkspace}
           />
         ))}
       </div>
--- a/src/renderer/stores/slices/__tests__/uiSlice.test.ts
+++ b/src/renderer/stores/slices/__tests__/uiSlice.test.ts
@@ -442,8 +442,15 @@ describe('UISlice — multiview', () => {
       // @ts-expect-error — cross-slice field overlaid for the test
       activeWorkspaceId: id,
     }));
-    // @ts-expect-error — cross-slice fields injected for the test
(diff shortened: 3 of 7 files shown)


openwong2kim:

This PR fixes two multiview consistency bugs found while dogfooding #748: the stale-id single-tile grid (#751) and the Ctrl+click-on-active grid collapse (#752). It closes both issues in one coherent change, touching five files across the renderer store and the layout viewport.

The strongest aspect of the approach is consolidation. The focus handoff — activate a neighbouring multiview member before removing the active one, so the render gate (`multiviewIds.includes(activeId)`) does not collapse every tile — moves from `WorkspaceViewport.handleRemoveTile` into `toggleMultiviewWorkspace` and `removeMultiviewWorkspace` in `uiSlice`. The tile X is then wired straight to the store action, so the sidebar Ctrl+click path and the tile X share one implementation and cannot drift apart again (which is exactly how #752 happened). The `removeWorkspace` pruning in `workspaceSlice` drops the id inside the same transaction that splices the workspace, with a `multiviewIds?.indexOf` guard for test stores that mount `workspaceSlice` without `uiSlice` — a thoughtful touch matching the existing cold-park guard. The test coverage is thorough: handoff to neighbour, last-member fallback, the no-handoff-when-group-collapses case, and the stale-id pruning including the single-member auto-clear.

Priority **P2** — two important UX bugs in a core layout feature, both pre-existing and user-visible (the grid silently vanishing or rendering a stranded single tile). The change is small (+157/−20), mergeable, and the CodeRabbit suggestion around the `[A, C, X, B]` / `[A, B, C]` ordering edge case is worth a quick look before merge to confirm the workspace-index fallback in `removeWorkspace` and the multiview-index handoff in `uiSlice` agree when workspace and multiview orderings diverge.

PR #754, opened 2026-08-02
Settle a fit deferred by the selection guard

### LINKED ISSUE #747: A resize while a selection is live leaves xterm and the PTY stuck at the old cols/rows

openwong2kim:
Closes #747.

## The bug

`shouldFitWhilePreservingSelection` skips `fit()` while a selection is live, so a
reflow can't wipe what the user is dragging out. Every skip site relied on the
same promise, written into fitGuard's own docstring:

> the next ResizeObserver tick (after the user releases) handles the deferred
> resize.

There is no such tick. `ResizeObserver` fires on *size change*, and releasing a
selection is not one. A resize that landed while a selection was live was dropped
outright: xterm kept the old `cols`/`rows`, and since `sendResize` is downstream
of `fit()`, so did the daemon PTY. Output wrapped at the wrong column and
full-screen TUIs drew against stale dimensions until something unrelated happened
to resize the container.

Reproduce: select some output in a pane and leave the selection up, drag the split
separator or toggle the sidebar, then click to clear the selection. `stty size`
inside that pane still reports the pre-resize dimensions.

## The fix

A skipped fit is a recorded debt, settled by the event the docstring assumed —
`onSelectionChange`, once the selection is actually gone. The retry runs the same
`runFit` the ResizeObserver uses, so it keeps scroll preservation and the
`sendResize` dedupe instead of drifting as a thinner copy.

`claimFit(term, pendingFitRef)` replaces direct guard calls. Checking the guard
and recording the debt were two steps a call site could get half-right, and
getting them half-right *is* the bug — so they are one call now. All four guarded
sites (fonts.ready, ResizeObserver, font/theme, visibility) go through it.

## Review notes

Panel review caught three things worth calling out, all fixed here:

- **The retry frame was neither tracked nor cancelled**, and `runFit` skipped the
  `terminalRef.current !== terminal` identity check that every other async site in
  this hook performs. A `ptyId` change re-runs the create effect, so a queued
  frame could fit the previous container and then send those dimensions to
  `ptyIdRef.current` — pointing at the new pty by then. One cancellable handle is
  now shared by both schedulers and cancelled beside the debounce timer in
  cleanup.
- **Debt lifecycle.** It is reset when a new terminal takes over the hook, and the
  shared `fit` callback settles it too — that path fits and resizes, so leaving
  the flag set made every later selection change re-run a fit already done.
- **The first version of the source lock was weak.** It matched a single-line
  guard shape, so a new site written across multiple lines, or via a helper call,
  would skip the debt and the test would still pass — a false negative in exactly
  the regression it existed to catch. `claimFit` makes the invariant structural:
  the lock is now "nothing calls the raw guard", and the decision itself is unit
  tested.

## Known limitation (deliberate)

While a selection is *held*, the fit stays deferred — that is the guard's whole
purpose, not an oversight. A user who selects text, resizes, and leaves the
selection up keeps the old dimensions until they clear it. Fitting sooner would
destroy the selection, which is the behaviour the guard was added to prevent.

One behaviour drift is accepted: the fonts.ready site used to run
`terminal.refresh(0, rows - 1)` after its fit, and the settle path does not.
That site's own comment calls itself "mostly defensive" (fonts resolve on mount,
before a selection can exist), and the glyph repaint scheduler covers staleness
on reveal.

## Testing

- `fitGuard.test.ts` — `claimFit` allows and records correctly, does not clear a
  debt owed by another site, and handles a null term. Real unit tests, not source
  matching.
- `useTerminal.deferredFit.test.ts` — source-level lock for the wiring that needs
  a live xterm to exercise: nothing calls the raw guard, the selection handler
  settles through `runFit`, `runFit` keeps scroll preservation + `sendResize` +
  the identity guard, the frame is cancellable and cancelled at teardown, and the
  observer delegates instead of duplicating the fit body. Same pattern as the
  `#191` atlas lock beside it.
- Both locks were mutation-tested: removing the debt record, and rewriting a site
  to call the raw guard across multiple lines, each fail with a message naming the
  fix.
- Full renderer suite: 2777 passing across 219 files. `tsc --noEmit` clean.
  `eslint` reports 3 pre-existing `no-empty-function` errors in `useTerminal.ts`
  that are present on `main` at the same call sites (line numbers shifted); no new
  ones.

### DIFF
--- a/changelog.d/754.md
+++ b/changelog.d/754.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **Resizing while text is selected no longer leaves the terminal at the wrong
+  size.** If you selected some output and then resized — dragging a split
+  separator, toggling the sidebar, resizing the window — the terminal and the
+  shell behind it stayed at the old dimensions even after you cleared the
+  selection. Output wrapped at the wrong column and full-screen programs drew
+  against stale dimensions until something else happened to resize the pane. The
+  deferred resize now runs as soon as the selection goes away. (#754)
--- a/src/renderer/hooks/__tests__/useTerminal.deferredFit.test.ts
+++ b/src/renderer/hooks/__tests__/useTerminal.deferredFit.test.ts
@@ -2,6 +2,12 @@ import { describe, it, expect } from 'vitest';
 import fs from 'node:fs';
 import path from 'node:path';
 
+// Source locks must not depend on line endings: `validate` runs on
+// windows-latest, where the checkout is CRLF, so any assertion carrying a
+// literal \n passes on macOS/Linux and fails only in CI. Normalise once here and
+// keep every pattern below whitespace-tolerant.
+const readSource = (p: string) => fs.readFileSync(p, 'utf-8').replace(/\r\n/g, '\n');
+
 // #747 regression lock (source-level).
 //
 // The selection guard skips fit() so a reflow can't wipe what the user is
@@ -17,7 +23,7 @@ import path from 'node:path';
 // #191 atlas lock and the Fix D / A6 invariants alongside it.
 describe('#747 — a deferred fit must be recorded and settled', () => {
   const hookPath = path.join(__dirname, '..', 'useTerminal.ts');
-  const src = fs.readFileSync(hookPath, 'utf-8');
+  const src = readSource(hookPath);
 
   it('no site calls the raw guard — every one goes through claimFit', () => {
     // This is the load-bearing assertion. Calling shouldFitWhilePreservingSelection
@@ -47,7 +53,11 @@ describe('#747 — a deferred fit must be recorded and settled', () => {
     // desync this fixes.
     const start = src.indexOf('const runFit = () => {');
     expect(start).toBeGreaterThan(-1);
-    const block = src.slice(start, src.indexOf('\n    };', start));
+    // Bound the block by the next declaration rather than an indentation
+    // pattern — a formatter change must not silently shrink or widen the slice.
+    const end = src.indexOf('const autoCopy = createAutoSelectionCopy', start);
+    expect(end, 'runFit is no longer followed by autoCopy — re-anchor this slice').toBeGreaterThan(start);
+    const block = src.slice(start, end);
     expect(block).toMatch(/fitAddon\.fit\(\)/);
     expect(block).toMatch(/sendResize\(/);
     expect(block).toMatch(/scrollToLine\(/);
@@ -64,11 +74,12 @@ describe('#747 — a deferred fit must be recorded and settled', () => {
     // Without a handle, several selection changes in one debt window each queue
     // their own fit, and a frame can still land after the terminal is disposed.
     expect(src).toMatch(/pendingFitRaf/);
-    const cleanupStart = src.indexOf('if (resizeDebounceTimer) clearTimeout(resizeDebounceTimer);\n      if (pendingFitRaf');
     expect(
-      cleanupStart,
+      src,
       'the queued fit frame is not cancelled next to the debounce timer in cleanup',
-    ).toBeGreaterThan(-1);
+    ).toMatch(
+      /clearTimeout\(resizeDebounceTimer\);\s*if \(pendingFitRaf !== null\) cancelAnimationFrame\(pendingFitRaf\);/,
+    );
   });
 
   it('the ResizeObserver delegates to runFit instead of duplicating it', () => {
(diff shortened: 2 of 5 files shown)


openwong2kim:

**P2** — Fixes #747, the selection-guard deferred-fit bug where resizing a pane while text was selected left xterm and the daemon PTY pinned to stale cols/rows until something else happened to resize the container.

This PR turns a skipped fit into a recorded debt (`pendingFitRef`) and settles it from `onSelectionChange` — the event the guard's docstring always assumed would fire but never does, since releasing a selection is not a container size change. The core design move is `claimFit(term, pendingRef)`, which merges the guard check and the debt record into a single call so a site can no longer get it half-right — which is the exact shape of the original bug. All four guarded sites (fonts.ready, ResizeObserver, font/theme, visibility) now route through it, and the raw guard is locked out at the source level in the test.

The retry deliberately runs the same `runFit` path the ResizeObserver uses (scroll preservation, `sendResize` dedupe, identity guard), rather than a thinner copy that would drift. Panel-review hardening shows up across the commits: the queued rAF handle is cancellable and cancelled at teardown, the debt is cleared on the path that actually fits (not on `claimFit`), and the source lock was made line-ending agnostic after a CRLF/CRLF failure on windows-latest CI — a nice catch that would have left the lock passing only off-CI. Test coverage is solid: real unit tests for `claimFit` (allow/refuse/debt/null-term) plus a source-level wiring lock for the xterm-bound hook, both mutation-tested. Notable that this overlaps with PR #750 (same issue, different author) — the two differ in structure (`claimFit` consolidation here vs. a separate debounced scheduler there), so a choice between them is a design call rather than a correctness one.

Touches `src/renderer/hooks/useTerminal.ts`, `src/renderer/utils/fitGuard.ts`, and their tests; no external or experimental surface changed, and `tsc --noEmit` is clean.

PR #755, opened 2026-08-02
fix(lanlink): coalesce lastSeenAt writes off the receive path

### LINKED ISSUE #665: lanlink: noteSeen() persists the whole peer store on every inbound record — a PowerShell spawn per message on win32
Found during review of #662. Pre-existing.

`onAeadRecord` calls `peers.noteSeen(peerUuid)` for every delivered record, purely to update a `lastSeenAt` timestamp. `noteSeen()` calls `persist()`, which does:

1. `atomicWriteJSONSync` of the **entire** peer store (rotation + backup rename + rename + fsync)
2. `reHardenTokenFileAcl` on the result — **on win32 a synchronous `whoami.exe` + `powershell.exe` shell-out**
3. `fsyncBestEffort`

`bumpHighWater()` adds a second full persist on the same path when the sequence advances.

The rate cap is `MAX_RECORDS_PER_SEC = 50`, so a single paired peer can drive up to 50 of these per second. This repo's own measurement (`src/shared/security.ts`) puts a *successful* ACL apply at 1.8–3.8s per process under AV — and it is `execFileSync`, so it blocks the daemon event loop, not just the lanlink path. A peer sending a burst of messages could stall the control pipe.

Note the interaction with #658: on a host where the ACL apply *fails*, this path throws on every inbound record, and each throw takes down the connection through `pump()`'s catch.

## Suggested direction

- `lastSeenAt` does not need durability per message. Update it in memory and flush on an interval, or on connection close.
- More broadly, the sync re-harden on a hot path is the wrong primitive. `security.ts` already has a deferred async variant for the cold-start case; something similar applies here.
- Better still: write the file owner-only in the first place so no post-hoc re-harden is needed at all. That would also remove one of the two candidate causes of #658.

Related: #658, #662, #663.

openwong2kim:
## Summary

`onAeadRecord` called `peers.noteSeen()` for every delivered record, purely to refresh a `lastSeenAt` timestamp. Each call persisted the **entire** peer store: a whole-store `atomicWriteJSONSync` plus a synchronous owner-DACL re-harden — on win32 a `whoami.exe` + `powershell.exe` `execFileSync`, measured in this repo at 1.8–3.8s per process under AV. Because it is `execFileSync` it blocks the daemon event loop, not just the lanlink path. Paired with `bumpHighWater`'s own persist, one peer streaming at `MAX_RECORDS_PER_SEC` (50) could drive 100 of those a second and stall the control pipe.

`lastSeenAt` does not need per-message durability — it only feeds LRU eviction (`pickEvictable`) and `lanlink.peers.list`, both of which read the in-memory map.

## Changes

- `noteSeen` refreshes `lastSeenAt` in memory and arms a coalescing flush (30s timer, unref'd). The flush also runs on connection teardown and on `dispose()`, which the daemon shutdown path already calls.
- Any security-critical persist (pairing, revoke, burn, high-water) settles the pending refresh on its way past — and clears the dirty flag only once the write actually succeeded, so a failed persist still retries.
- `flushSeen()` never throws. On the receive path a throw would reach the server's `pump()` catch and take down the connection over a timestamp write — the #658 interaction called out in the issue.
- Regression tests at both levels: the store no longer writes per `noteSeen`, flushes exactly once, survives a failing write, and settles on the timer; the server costs **one** store write per delivered record (was two — verified by reverting the fix, which turns the assertion into 10 vs 5).

## Deliberately unchanged

`recvHighWater` keeps its durable per-record persist. It is the only cross-connection replay gate (`server.ts` re-ACKs and drops anything at or below it), so deferring it would widen the replay window after a crash. The receive path therefore drops from two store writes per record to one, not to zero. Making the persist primitive itself cheap on win32 — the issue's third suggestion, which would also remove a candidate cause of #658 — is left to a follow-up, since it changes the fail-closed C12 semantics in `shared/security` and `atomicWrite`.

Closes #665

## Stability tier impact

- [ ] Additive change to a stable surface.
- [ ] Breaking change to a stable surface.
- [ ] Change to an experimental surface.
- [ ] Change to an internal surface.

## Summary by CodeRabbit

* **Performance**
  * Reduced repeated peer-store writes by batching “last seen” updates and saving them on a timer or when connections close.
  * Ensured pending updates are persisted during shutdown.

* **Reliability**
  * Added retry handling when saving updates fails, without interrupting normal operation.

* **Tests**
  * Added coverage for batched persistence, timer-based saves, connection cleanup, shutdown behavior, and write failures.

### DIFF
--- a/changelog.d/755.md
+++ b/changelog.d/755.md
@@ -1,9 +1,10 @@
 ### Fixed
 
-- **LanLink no longer rewrites its whole peer file for every message received.** Each
-  inbound record used to force two full peer-store saves, and on Windows every save
-  shelled out to `whoami.exe` and `powershell.exe` to re-apply the owner-only file
-  permissions — a step measured at seconds per call on machines running antivirus, and
-  one that blocked the whole daemon while it ran. A peer sending a burst of messages
-  could stall wmux. The "last seen" timestamp is now kept in memory and written on a
-  timer, when the connection ends, and on shutdown. (#755)
+- **LanLink halved the disk writes a received message costs.** Each inbound record used
+  to force two full rewrites of the peer file, and on Windows every rewrite shelled out
+  to `whoami.exe` and `powershell.exe` to re-apply the owner-only file permissions — a
+  step measured at seconds per call on machines running antivirus, and one that blocked
+  the whole daemon while it ran, so a peer sending a burst of messages could stall wmux.
+  The "last seen" timestamp now rides along with the write the record already needed,
+  instead of forcing one of its own. The remaining write is the replay guard, which has
+  to be durable; making that one cheap on Windows is still open. (#755)
--- a/changelog.d/755.md
+++ b/changelog.d/755.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **LanLink no longer rewrites its whole peer file for every message received.** Each
+  inbound record used to force two full peer-store saves, and on Windows every save
+  shelled out to `whoami.exe` and `powershell.exe` to re-apply the owner-only file
+  permissions — a step measured at seconds per call on machines running antivirus, and
+  one that blocked the whole daemon while it ran. A peer sending a burst of messages
+  could stall wmux. The "last seen" timestamp is now kept in memory and written on a
+  timer, when the connection ends, and on shutdown. (#755)
--- a/src/daemon/lanlink/__tests__/peers.test.ts
+++ b/src/daemon/lanlink/__tests__/peers.test.ts
@@ -284,9 +284,10 @@ describe('peers — per-peer store', () => {
 
     // A throw here would reach the server's pump() catch and kill the connection
     // over a lastSeenAt write — the #658 interaction called out in #665.
-    it('never throws when the write fails, and retries the refresh later', () => {
+    it('never throws when the write fails, and the recovered write carries the refresh', () => {
       const origPlatform = Object.getOwnPropertyDescriptor(process, 'platform');
       Object.defineProperty(process, 'platform', { value: 'win32', configurable: true });
+      vi.useFakeTimers();
       try {
         let hardenOk = true;
         const s = new PeerStore(dir, {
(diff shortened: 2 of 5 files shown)


openwong2kim:

This PR closes #665 by moving lanlink's `lastSeenAt` timestamp refresh off the per-record persist path. Today every inbound AEAD record calls `noteSeen()`, which does a full peer-store `atomicWriteJSONSync` plus a synchronous owner-DACL re-harden — on win32 that is a `whoami.exe` + `powershell.exe` `execFileSync` measured at 1.8–3.8 s under AV, and because it is `execFileSync` it blocks the daemon event loop, not just lanlink. A single paired peer streaming at `MAX_RECORDS_PER_SEC` (50) could drive 100 of those a second and stall the control pipe.

The fix is a coalesced in-memory refresh flushed on a 30 s timer (unref'd), on connection teardown, and on `dispose()`. The scoping is careful in the places that matter: `flushSeen()` deliberately never throws so a failed ACL write cannot reach `pump()`'s catch and kill the connection (#658 interaction), `seenDirty` is only cleared after a persist succeeds so a failing write retries on the next `noteSeen`, and `recvHighWater` keeps its durable per-record persist because it is the only cross-connection replay gate — so the receive path drops from two store writes per record to one rather than zero. The test coverage matches: store-level tests prove zero writes per `noteSeen`, exactly-once flush, in-memory-vs-disk ordering, resilience to a failing write, and dispose/teardown settlement, plus a server-level test asserting one write per delivered record (was two, verified by reverting). Five files, +239/−6, no external surface touched — the stability classification is correct.

Assigning **P2** to match the issue it resolves: the hot path is real and reachable under ordinary pairing load on Windows with AV, and this is a clean, well-tested fix that removes a genuine event-loop stall. One follow-up worth tracking is the win32 ACL primitive itself (the issue's third suggestion), which would also remove a candidate root cause of #658; that is correctly out of scope here because it changes the fail-closed C12 semantics in `shared/security` and `atomicWrite`.

PR #757, opened 2026-08-02
Say which command tripped the critical signal, and write down that it is notify-only

### LINKED ISSUE #605: session:critical is a pattern match, not an approvable request — rework needed before any remote approve/deny surface

openwong2kim:
Closes #605.

## The gap

`session:critical` emitted a pattern **label** and discarded the line it matched:

```ts
const value = clean.slice(0, 80);   // the matched line
if (this.lastEmittedFor.get(key) === value) return;
cb({ action: cp.label, riskLevel: cp.riskLevel });   // …thrown away
```

`cp.label` is one of eleven hardcoded strings, so `git push --force origin main`
and `git push -f scratch` produce byte-identical events. A phone showing one of
these can say something forceful happened somewhere, and nothing more.

## What changed

`CriticalEvent` gains `matchedLine` — the same 80-char ANSI-stripped slice the
detector already computed for its dedup key, so what a surface shows is exactly
what the dedup judged. Control bytes (C0/DEL/C1) are replaced and whitespace
collapsed, because this string lands in notification bodies and on a phone.

It rides the existing path untouched: `DaemonPTYBridge` → `session:critical` →
`/api/events`, where the payload is spread whole. No wire change beyond the new
field; `tier` stamping and dedup are unmodified.

The rest is writing down what this signal is, where a future reader will look —
`onCritical`'s doc, the type, and the phone contract. It fires on printed
**output**, not on a pending action: nothing is blocked, nothing is waiting,
there is no addressee for a reply, and repeats within a cycle are deduped away.
Notify-only, permanently. Anything answerable gates on the hook-sourced approval
registry (`src/daemon/approvals/`), which has real request identity.

The two remaining complaints in #605 were already closed by the approval work:
per-request identity landed as `ApprovalRequest.id`, and the actionable signal is
already hook-sourced (`HookIngest.driveApprovals` refuses detector-sourced
`awaiting_input`).

## Deliberately not changed

`tierFor` still maps a `critical`-risk signal to `act`. That was a considered
call in #631 — the failure that matters is a destructive action delivered
quietly — and "notify-only" is about whether a surface may **answer** the event,
not about how loudly it arrives.

## Tests

4 new (matched line distinguishes two hits of one label; ANSI + control-byte
stripping; the 80-char cap; the line reaching the SSE wire). Full suite: 9676
passing, `tsc` clean on both configs, lint clean on every touched file.

## Summary by CodeRabbit

* **New Features**
  * Critical-action notifications now include the matched command or terminal line.
  * Context is sanitized and limited to 80 characters for safer display.
  * Critical signals are clearly notify-only; approval controls remain available only for approval events.
* **Documentation**
  * Updated the phone client contract and changelog with critical-event payload details, compatibility guidance, and notification behavior.

### DIFF
--- a/changelog.d/757.md
+++ b/changelog.d/757.md
@@ -1,9 +1,11 @@
 ### Changed
 
-- Critical-action notifications now name the command. The `critical` attention
-  event carries `matchedLine` — the terminal line that tripped the pattern,
-  sanitized and capped — so a heads-up can say `git push --force origin main`
-  instead of just "git push --force".
+- Critical-action notifications now show the matched terminal line. The
+  `critical` attention event carries `matchedLine` — the printed terminal
+  output that tripped the pattern, sanitized and capped — so a heads-up can
+  say `git push --force origin main` instead of just "git push --force". (This
+  is the text the terminal displayed, which may quote a command rather than run
+  one; it is not proof the command executed.)
 - Documented that the critical signal is notify-only: it fires on printed
   output, not on a pending action, and is not something a remote surface can
   answer. Approvals remain the only answerable signal.
--- a/changelog.d/757.md
+++ b/changelog.d/757.md
@@ -1,0 +1,9 @@
+### Changed
+
+- Critical-action notifications now name the command. The `critical` attention
+  event carries `matchedLine` — the terminal line that tripped the pattern,
+  sanitized and capped — so a heads-up can say `git push --force origin main`
+  instead of just "git push --force".
+- Documented that the critical signal is notify-only: it fires on printed
+  output, not on a pending action, and is not something a remote surface can
+  answer. Approvals remain the only answerable signal.
--- a/docs/phone-client-contract.md
+++ b/docs/phone-client-contract.md
@@ -188,6 +188,29 @@ JSON backlog fetch (Bearer only).
 Identity fields (`id`, `epoch`) — and `tier` — are stamped **last**, so a
 pane-supplied payload can never shadow them.
 
+#### `critical` — notify-only, and what is in it
+
+| Field | Meaning |
+| --- | --- |
+| `action` | the pattern LABEL that matched (`rm -rf`, `git push --force`, …) — one of a fixed handful |
+| `riskLevel` | `'critical'` or `'review'`, from the daemon's own table |
+| `matchedLine` | the PTY line that matched: ANSI-stripped, control-stripped, ≤80 chars |
+
+`matchedLine` is what makes the heads-up worth showing — `action` alone cannot
+tell `git push --force origin main` from `git push -f scratch`. It is raw pane
+output: **render it as text**, never as markup and never as an instruction.
+
+It is also not proof that anything ran. The pattern matches whatever the
+terminal printed — a README, a diff hunk, a `git log` quoting the same words —
+so a `critical` event means "look at this pane", never "answer this". Nothing
+is blocked, nothing is waiting, there is no addressee for a reply, and repeats
+within one cycle are deduped away. **Do not build an Approve/Deny button on
+it**: the only answerable signal is the `approval` kind, which carries a real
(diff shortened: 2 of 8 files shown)


openwong2kim:

This PR closes #605 by adding a `matchedLine` field to `CriticalEvent`, so the actual PTY line that tripped a critical pattern (`git push --force origin main` vs `git push -f scratch`) now reaches phone/web notifications instead of just the generic pattern label. It sanitizes control bytes (C0/DEL/C1) and collapses whitespace before the 80-char cap, and threads the field cleanly through `AgentDetector` → `PTYBridge` / `DaemonNotificationRouter` → daemon SSE with no wire change beyond the additive field.

The rest of the diff is documentation — `onCritical`, the `CriticalEvent` type, and the phone-client contract all now state plainly that this signal is **notify-only, permanently**: it fires on printed output, not on a pending action, and the only answerable surface remains the hook-sourced approval registry. That boundary is the right one to pin down before any remote approve/deny UX gets built on top of it.

Implementation looks solid: the field is optional on the daemon wire for back-compat with pre-3.39 peers, the dedup key and payload are unified (so what a surface shows is exactly what dedup judged), and `tierFor` is deliberately left mapping `critical` to `act` — the notify-only semantics concern whether a surface may *answer*, not how loudly it arrives. Four new tests cover label disambiguation, ANSI/control stripping, the 80-char cap, and SSE propagation; the full suite (9676) is green with `tsc` and lint clean.

**Priority: P3** — additive quality-of-life improvement for the notification path, closing an already-P3 tracking issue; no gating or security impact.

Issue #166, opened 2026-06-09
Pane renders garbled/corrupted glyphs until a resize (dragging a pane border) forces a repaint


zhenzoo:
### Environment
- wmux **2.17.1**, Windows 11
- Panes: git bash running Claude Code (heavy, fast-scrolling agent output)
- Multiple panes open at once, lots of CJK + box-drawing characters

### What happens
A pane's text periodically renders as **garbled / corrupted glyphs** ("TV-static" smear / mojibake), even though the underlying content is fine. The instant I **drag a pane border** (which resizes the panes), the affected pane **repaints correctly** and the garbage disappears.

### Screenshots
**Garbled pane (before):**

<img width="2019" height="2033" alt="Image" src="https://github.com/user-attachments/assets/beec97cb-726d-4960-a90e-e29c34cdb7a3" />

**After dragging a pane border (fixed):**

<img width="1984" height="2047" alt="Image" src="https://github.com/user-attachments/assets/4de3fd11-d3cb-4b15-a27b-6a1c18abc4be" />

### Repro (non-deterministic but frequent)
- During / after a burst of fast output, and/or when several panes update at once.
- More likely with CJK text + box-drawing chars + many panes + long-running sessions.

### Likely cause
xterm.js repaints only changed regions; when the display model desyncs from the buffer (fast scroll / bulk write to an unfocused pane / CJK width), the stale area isn't fully repainted until a **resize forces a full `refresh()`** — the same class as other "corruption above never repaints until resize" reports. If the WebGL/Canvas renderer addon is in use, a corrupted **texture atlas** also produces exactly this "garbage glyphs until resize" symptom.

### Possible fixes
- Force `term.refresh(0, term.rows - 1)` on triggers: a pane regains focus, after a large write burst, or a low-frequency periodic full refresh.
- If using `@xterm/addon-webgl` / `@xterm/addon-canvas`: call `clearTextureAtlas()` on resize/focus and handle `onContextLoss`.

Happy to provide a screen recording if useful. Thanks for wmux! 🙏

PR #759, opened 2026-08-02
fix(renderer): restore the DOM renderer when WebGL teardown skips xterm's own restore

openwong2kim:
## Problem

Panes flicker and then settle **black** until relaunch. Production logs show the mechanism: `Uncaught TypeError: Cannot read properties of undefined (reading 'dimensions')` thrown from the xterm vendor bundle **2,990 times across one 85-minute storm**, coinciding with surface register/unregister churn. Meanwhile `Too many active WebGL` / `context lost` warnings: zero — this is not context-cap eviction.

## Root cause

`WebglAddon.dispose()` restores the DOM renderer via a disposable registered in its **own DisposableStore**. If any earlier disposable in that store throws mid-dispose (GL work against a dying context can), the store's loop aborts and the restore step never runs. `teardownWebglAddon` swallows that throw — correct for teardown robustness — but the pane was left with **no renderer at all**: xterm's `RenderService.dimensions` getter reads `this._renderer.value.dimensions` unguarded, so every subsequent render tick (cursor blink, scroll, fit) threw. The pane flickers as frames fail, then stays black.

The refresh()-based repaint band-aids (#166/#191/#318) never touched this layer — they repaint *through* the render service, which is exactly the thing that is broken here.

## Fix

After dispose, **verify** the render service still holds a renderer; if not, restore the DOM renderer ourselves — the same `setRenderer(_createRenderer())` + `handleResize` sequence the addon's skipped disposable would have run. All three teardown sites (pool evict, fonts.ready atlas rebuild, unmount — where `terminal.dispose()` can be deferred by an active drag, #582) pass the terminal so the invariant holds everywhere.

Every private-path access is optional-chained (internals change ⇒ degrade to old behaviour, no crash), and the exact paths are shape-locked by tests against the installed @xterm sources so a package bump fails loudly instead of silently reviving the black-pane bug.

## Tests

- `webglTeardown.test.ts`: +5 behavioural cases (restore fires when renderer missing; no-op on live renderer / disposed terminal / changed internals; `isRendererMissing` truth table) and +4 shape locks (`_core._renderService.setRenderer(_core._createRenderer())`, `handleResize`, `RenderService`'s unguarded `_renderer.value.dimensions` read, `_core._store._isDisposed`).
- Full renderer hook/terminal suites: 737 tests, 57 files — green. `tsc --noEmit` clean; lint errors in `useTerminal.ts` are preexisting on main.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where terminal panes could turn black after WebGL renderer teardown errors.
  * Terminals now automatically restore the standard renderer when needed, including during deferred cleanup.
  * Improved recovery handling while preserving active renderers and disposed terminals correctly.
* **Tests**
  * Added coverage for renderer recovery and teardown edge cases.

### DIFF
--- a/changelog.d/759.md
+++ b/changelog.d/759.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **Panes no longer flicker and settle black after a WebGL teardown.** When
+  xterm's WebGL addon was disposed (pane hidden, tab switch, font reload) and
+  an internal step threw mid-dispose, the addon's own "hand back to the DOM
+  renderer" step was silently skipped — leaving the pane with no renderer at
+  all. Every render tick after that (cursor blink, scroll, resize) threw, so
+  the pane flickered and then stayed black until relaunch. Teardown now
+  verifies a renderer survived and restores the DOM renderer itself when it
+  did not. (#759)
--- a/src/renderer/hooks/__tests__/useTerminal.webglTeardown.test.ts
+++ b/src/renderer/hooks/__tests__/useTerminal.webglTeardown.test.ts
@@ -28,8 +28,8 @@ describe('useTerminal WebGL teardown (source-level lock)', () => {
     );
   });
 
-  it('routes the pool-evict, fonts.ready and unmount paths through teardownWebglAddon', () => {
+  it('routes the pool-evict, fonts.ready, unmount and context-loss paths through teardownWebglAddon', () => {
     const calls = SRC.match(/teardownWebglAddon\(/g) ?? [];
-    expect(calls.length).toBeGreaterThanOrEqual(3);
+    expect(calls.length).toBeGreaterThanOrEqual(4);
   });
 });
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1103,7 +1103,7 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
     // just loses GPU acceleration until it is granted a context again.
     function disposeWebgl() {
       if (!webglAddonRef.current) return;
-      teardownWebglAddon(webglAddonRef.current);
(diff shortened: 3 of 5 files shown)


openwong2kim:

This PR addresses a high-impact renderer bug: panes flicker and then settle **black** until relaunch, traced through production logs to `WebglAddon.dispose()` silently skipping its own DOM-renderer restore step when an earlier disposable in its store throws mid-dispose. The root-cause analysis is precise — the pane is left with no renderer at all, so xterm's `RenderService.dimensions` getter throws on every subsequent render tick (`Cannot read properties of undefined (reading 'dimensions')`), observed 2,990× across one 85-minute storm. Crucially, the author identifies that the existing refresh()-based repaint band-aids (#166/#191/#318) never touched this layer because they repaint *through* the render service, which is the very thing that was broken.

The fix is well-scoped and defensive: after dispose, `teardownWebglAddon` now verifies a renderer survived and restores the DOM renderer itself via the same `setRenderer(_createRenderer())` + `handleResize` sequence the addon would have run. All three teardown sites (pool evict, fonts.ready atlas rebuild, unmount — including the deferred-dispose window from #582) plus the context-loss path now route through this and pass the terminal so the invariant holds everywhere. Private-path access is optional-chained so an @xterm internals change degrades to a no-op rather than crashing, and the exact paths are shape-locked against the installed package sources so a bump fails loudly instead of silently reviving the black-pane bug.

**Priority: P2.** This is a visible production failure affecting a core surface (the terminal pane itself), but it is scoped to the WebGL teardown path, the fix is contained to `webglTeardown.ts` and its call sites, and the existing refresh band-aids keep non-WebGL terminals unaffected. The 9 behavioural and shape-lock tests give strong confidence in the invariant.

**Labels:** P2

PR #767, opened 2026-08-02
feat(sidebar): compress workspace row density, one-line agents, full-width branch

openwong2kim:
## Summary

The branch label in each sidebar workspace row was hard-capped at
`max-w-[130px]`, so names like `feat/deck-new-session` were truncated even
when the row had plenty of room. Dropping the cap and giving the span
`min-w-0 truncate` lets the branch use the full width of its line: it keeps
its content width (so the git signal-light and PR badges stay right next to
it), and shrinks/elides only when a long name would otherwise overflow.

```
before:  ⎇ feat/deck-n…          ●7
after:   ⎇ feat/deck-new-session  ●7
```

One line changed in `WorkspaceContextLine` (`WorkspaceItem.tsx`). No behavior
or logic change — the git signal-light, PR badge, and worktree marker keep
their existing positions and the agent roster (separate line) is untouched.

## Review

Reviewed by a three-model panel (Claude, Codex, GLM). All three flagged the
same regression in the first attempt (`flex-1` pushed the badges to the far
right on short branch names); the fix adopted the unanimous recommendation of
`min-w-0 truncate` without `flex-1`, which keeps the badges adjacent while
still uncapping the width.

## Test plan

- [x] `tsc --noEmit` clean (0 errors)
- [x] `eslint` clean on the changed file (0 errors)
- CSS-only Tailwind class change; visual behavior verified against the
  intended before/after above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sidebar branch label display by using available row width.
  * Branch names now truncate only when necessary, while keeping Git status and pull request badges visible.

### DIFF
--- a/changelog.d/767.md
+++ b/changelog.d/767.md
@@ -1,3 +1,11 @@
+### Changed
+
+- **The workspace list fits more of your fleet on screen.** Rows are tighter,
+  and each running agent in a selected workspace now sits on a single line
+  (name, pane, and status together) instead of a stacked, boxed card. When an
+  agent is actually waiting on you, its question still opens on a second line
+  in red so you can see what it needs at a glance.
+
 ### Fixed
 
 - **Long branch names in the sidebar are no longer cut short.** Each
--- a/changelog.d/767.md
+++ b/changelog.d/767.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **Long branch names in the sidebar are no longer cut short.** Each
+  workspace row shows its git branch under the name, and that label used to be
+  capped at a fixed width, so a branch like `feat/deck-new-session` was
+  truncated to `feat/deck-n…` even when the row had room to spare. The label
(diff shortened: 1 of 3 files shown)


openwong2kim (commented):

This PR removes the hard `max-w-[130px]` cap on the sidebar branch label in `WorkspaceItem.tsx`, replacing it with `min-w-0 truncate` so long branch names like `feat/deck-new-session` use the full row width and only elide when they would genuinely overflow. The change is a single Tailwind class swap on one span inside `WorkspaceContextLine` — no logic, behavior, or layout structure changes — and the git signal-light and PR badges stay adjacent to the branch name as before. The included changelog entry and the three-model review notes (unanimous `min-w-0 truncate` without `flex-1`) make the intent and the rejected alternative clear.

Priority assigned: **P3**. It is a clean, well-scoped visual polish fix with no risk surface.

PR #769, opened 2026-08-02
Hand PTY size to the phone when the desk is not looking

openwong2kim:
## Problem

`POST /api/sessions/:id/resize` exists so a phone can reflow a pane to its own viewport, but on a live install it never fires: every session the Mac app holds open is `attached`, so the route answers `409 desk-owns-size` without exception. A 151×47 desk pane rendered width-fit on a 402×559 portrait viewport wastes nearly half the screen (#766).

Simply dropping the 409 would start the fight the route's own comment warns about: the desk re-derives its geometry on every layout pass, so "last writer wins" is a PTY thrashing between two shapes.

## Change — visibility-based ownership

Only one side re-asserts its shape at any moment, so there is no fight to referee:

- **Renderer** reports whether a pane is actually on screen — workspace shown, tab active (`isVisible`, already computed by `Terminal.tsx`), AND the window itself visible (`document.visibilityState`, which covers minimize and full occlusion). Fire-and-forget `pty:setViewerVisibility` IPC → new `daemon.setSessionViewerVisibility` RPC. A hidden pane's own fit was already silent (zero-size container guard), so a hidden desk cannot resize back.
- **Daemon** (`handleSessionResize`) answers `409 desk-owns-size` only for `attached && viewerVisible`. Attached-but-hidden takes the phone's numbers. The deferred-pane unmute handshake and the 250 ms rate limit are unchanged and still checked after ownership.
- **Reclaim** is the existing mount/reveal fit, which resizes unconditionally. One new case: restoring a minimized window fires no ResizeObserver tick (the container never changed size), so the visibility effect re-runs `fit()` on document reveal — the desk silently takes the size back the moment somebody looks.

### The one ordering decision worth reviewing

The flag resets to visible on **detach**, deliberately not attach. The renderer's mount-time visibility report can land before its attach RPC completes; an attach-time reset would overwrite a fresh "hidden" with the default and permanently mark every hidden retained pane as visible. Resetting on detach means the next attacher starts at the safe default (desk owns), and a current renderer re-reports on mount and on every flip.

### Compatibility

- Older renderer that never reports: flag stays `true` → pre-#766 behavior exactly (always 409 while attached).
- Older daemon: the RPC is unknown → error swallowed by the fire-and-forget relay; phone behavior unchanged.
- Contract doc (`docs/phone-client-contract.md`) resize section updated: visibility is not observable in `/api/sessions` — the probe is the request itself; desk and phone both watching stays 409 with the existing fit+scale fallback.

## Tests

- `DaemonSessionManager`: default-visible, report survives attach, detach resets, unknown-id report is a no-op.
- `WebTerminalServer`: attached-but-hidden pane hands the size to the phone (200 with applied geometry); attached-and-visible still 409.

## Verification

- `tsc --noEmit` (main + daemon tsconfigs) clean
- eslint on touched files: error count identical to baseline
- full `test:parallel` suite: 9744 passed

### DIFF
--- a/changelog.d/769.md
+++ b/changelog.d/769.md
@@ -1,0 +1,11 @@
+### Changed
+
+- **A phone can reshape a pane the desk holds but is not looking at.** `POST
+  /api/sessions/<id>/resize` used to answer `409 desk-owns-size` for every
+  attached pane, and every live pane is attached while the Mac app is open —
+  so the phone always rendered a desk-shaped (151×47-ish) pane letterboxed on
+  a portrait screen. Ownership is now visibility-based: the renderer reports
+  whether a pane is actually on screen (workspace + tab active, window not
+  minimized or occluded), and the 409 is reserved for panes somebody is
+  actually watching. The desk silently takes the size back on the next
+  attach or reveal. (#769)
--- a/docs/phone-client-contract.md
+++ b/docs/phone-client-contract.md
@@ -291,7 +291,7 @@ app-owned conversation history. Without `--allow-input`, send neither; never
 forward generic taps or drags, guess a mouse protocol, or pretend the alternate
 buffer is locally scrollable.
 
-### Resizing a pane — the desk owns the size whenever it is attached
+### Resizing a pane — the desk owns the size while it is actually showing it
 
 ```
 POST /api/sessions/<id>/resize   body: {cols, rows}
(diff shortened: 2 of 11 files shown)


openwong2kim:

This PR implements the visibility-based size ownership proposed in #766, letting the phone reshape a pane the desk holds but is not currently showing. It threads a new `pty:setViewerVisibility` IPC (`useTerminal.ts` → preload → `pty.handler` → `daemon.setSessionViewerVisibility` RPC) so the daemon can distinguish "attached and watched" from "attached but nobody looking," and `handleSessionResize` now reserves the `409 desk-owns-size` for `attached && viewerVisible`. The design is careful about the right edges: the flag defaults to visible and resets on detach (not attach) so older renderers and mount-time report races degrade to pre-existing behavior, the renderer re-fits on window-level reveal to reclaim the size without a ResizeObserver tick, and the contract doc is updated to match.

The change touches eleven files across daemon, main, preload, renderer, and shared constants — proportionate to the feature — and ships targeted tests on both sides (visibility lifecycle in `DaemonSessionManager`, attached-but-hidden handing the size to the phone in `WebTerminalServer`) plus a full `test:parallel` pass (9,744). It is a clean, well-scoped implementation of a real mobile UX problem (a 151×47 desk pane letterboxed on a portrait phone wastes nearly half the screen), so I am marking it **P2** to match the tracking issue. One worth verifying on a live fleet before merge: the ordering decision on detach-vs-attach reset, since the comment block is the only thing guarding it.

openwong2kim:
@codex review

openwong2kim:
Parking this by owner decision. The mechanism works and CI is green, but two known gaps make it premature to flip on: the handover has no hysteresis (a 20-second alt-tab while a phone holds the pane hard-wraps that window of scrollback at phone width, permanently), and the desk gets no indication that a phone took the size (Orca ships a banner + explicit take-back for the same feature). If this comes back, it restarts from this branch with those two on top — see the review discussion above and the prior-art notes.

PR #771, opened 2026-08-02
Project a pane's transcript as structured turn events (daemon half of #655)

openwong2kim:
## The daemon half of #655, extracted

Owner decision on #655 (Chat View): the desktop renderer surface is parked — its UX did not earn 15k lines in dogfood — but the daemon pipeline underneath it is UI-free, was the most heavily reviewed part of that PR (three-model review, byte budgets, path confinement, backpressure), and is the foundation any future conversation surface (notably the phone) builds on. Leaving it on the feature branch means it rots against main; landing it now costs nothing visible.

**No behavior changes. No consumers yet.** Every addition is additive and dormant until something calls the new RPCs.

### What lands

- **`src/daemon/transcript/`** — `TranscriptProjector` (bounded tail reads via `readTail`, `fs.watch` + polling fallback, page/entry byte budgets, code bodies held daemon-side and fetched on expand), `parseEntry` (one transcript JSONL line → 0..n normalized `TurnEvent`s; never throws, unknown entry classes yield `[]`), `TranscriptDiscovery` (finds `<agentSessionId>.jsonl` by name so a pane is projectable before its first turn ends).
- **`src/shared/transcript/turnEvents.ts`** — the `TurnEvent` wire contract.
- **`transcriptPathGuard`** — a hook-supplied transcript path is confined to the Claude projects root (realpath symlink check) and its basename must equal `<agentSessionId>.jsonl`. Refusal drops the path, never the binding. Enforced at both writers (`HookIngest` and the `daemon.setResumeBinding` RPC), turning a would-be arbitrary-file-read into a refused binding.
- **`DaemonPipeServer`** — per-client identity (`RpcClientContext`), unicast `sendTo` with a 4 MB per-subscriber backpressure cap (a stalled reader loses its subscription instead of growing daemon heap unbounded), `onClientClose` teardown, and a first-party classification (`daemon.client.identify`) so the transcript RPCs answer only the app's own socket — a pane's full conversation never reaches CLI/MCP pipe clients.
- **`daemon.transcript.{status,snapshot,subscribe,unsubscribe,codeBlock}`** — token-only string methods, deliberately not in the `RpcMethod` union (same class as `daemon.hooks.*`), so no gen-api-reference surface moves.
- **`HookIngest`** — vets `transcript_path` and nudges the projector on every resolved signal, wrapped so a projector failure can never turn into a fatal hook.
- **`activitySummary`** — a `Skill` tool call summarizes as `/skill-name` instead of a bare `Skill` (this also improves the Fleet View activity line today).

### Carried tests

Projector / parser / discovery / readTail suites with their fixtures, the subscriber-guard suite, the path-guard suite, and the updated `HookIngest` + x6 durability assertions — verbatim from the #655 branch. Full parallel suite: 9,911 passing locally.

### For the phone track (future, explicitly out of scope here)

The transcript RPCs are gated `firstPartyOnly` on purpose. Exposing projection to paired phones over `WebTerminalServer` reverses a deliberate trust decision (transcripts carry full history beyond the ring buffer) and needs its own design pass — tracked separately, not smuggled in here.

## Summary by CodeRabbit

* **New Features**
  * Added live transcript viewing with snapshots, incremental updates, subscriptions, paging, and code-block retrieval.
  * Automatically discovers Claude transcripts, including relocated project directories.
  * Added targeted session updates and improved client connection cleanup.
  * Improved activity summaries for Skill tool usage with slash-format commands.
* **Bug Fixes**
  * Prevented unsafe, stale, malformed, truncated, or rotated transcripts from producing incorrect updates.
  * Improved handling of disconnects, backpressure, transcript discovery failures, and session cleanup.
  * Added safer handling for malformed transcript content and unsupported entries.


openwong2kim:

This PR lands the daemon-side transcript projection pipeline extracted from #655 — the UI-free foundation that tails Claude Code transcript JSONL, normalizes it into structured `TurnEvent`s, and serves snapshot/subscribe/code-block RPCs, with no behavior changes and no consumers yet. The scope is large but well-bounded: `TranscriptProjector` with byte-budgeted `readTail` (256 KB tail, 128 KB serialized cap) and `fs.watch` plus polling fallback, `transcriptPathGuard` confining hook-supplied paths to the Claude projects root via realpath at both writers, per-client identity and unicast `sendTo` in `DaemonPipeServer` with a 4 MB backpressure cap that drops stalled subscribers rather than growing daemon heap, and a first-party classification so transcript content reaches only the app's own socket. The trust-boundary decisions are the right defaults — transcripts carry full conversation history beyond the ring buffer, so gating the RPCs `firstPartyOnly` and enforcing path confinement before projection closes the obvious arbitrary-file-read and cross-client-leak vectors, and the x6 durability assertion now verifies `setResumeBinding` vets `transcriptPath` on the wire path too. P2 is appropriate as foundational plumbing the phone conversation surface depends on, with the explicit and correct carve-out that phone-side exposure over `WebTerminalServer` needs its own design pass rather than being smuggled in here.

openwong2kim:
Review triage on the three CodeRabbit findings, all verified against source before deciding:

- **Surrogate-pair split in `capText` (Minor) — fixed** in 063b0561. The cut now backs off one code unit when it would land on a high surrogate, same reasoning `sliceToBytes` already documents for tool bodies.
- **FIFO TOCTOU in `readTail` (Major) — skipped.** The race is real syscall-wise, but the only writer who can swap `<agentSessionId>.jsonl` for a FIFO inside `~/.claude/projects` is the local user attacking their own daemon — explicitly outside this surface's threat model (`markFirstParty` doc: not 'a local attacker who already stole the token'). The suggested `O_NONBLOCK` fix is also unavailable on Windows, so there is no portable no-race open; the lstat-first refusal remains the best available bound.
- **Discovery poll fan-out (Major) — skipped, and the reasoning is now recorded in the code** (063b0561). The proposed root-mtime short-circuit misses the exact event the poll exists for: the transcript appears inside an existing project directory, which bumps that directory's mtime, not the root's — and `fs.watch(root)` is non-recursive, so the poll is the only trigger that sees it. A correct per-directory mtime probe costs one stat per project dir, which is the same syscall count as the scan's own candidate stats. Cost stays bounded by design (one readdir + ≤512 stats per 1 s tick, only while a pane is searching, ≤5 min).

openwong2kim:
@codex review

openwong2kim:
Codex review pass (codex-cli 0.146.0, reasoning=high) — 2 P1 + 2 P2, each verified against source before deciding:

- **[P2] Second subscriber never seeded — CONFIRMED, fixed** in 2ee15684. subscribe()'s own comment promised "subscribe alone gets the conversation", but a client joining an established watch got an empty delta from the shared EOF cursor. Now rewinds to a reset snapshot on join (the same forceReset+tailOffset=-1 pair every other reset site uses), with a test. Harmless for today's single consumer; exactly where the phone's multi-client future would have bitten.
- **[P1] macOS rename events carrying only the staging name — REFUTED as stated.** The claimed reproduction ("the ATOMIC RENAME runtime test reproduces this timeout") is false: all 32 runtime tests pass, including that one. The design already watches the DIRECTORY (not the file), treats null filenames as "maybe ours", and falls back to polling on watcher error. Kept as designed.
- **[P1] Budget halving can skip a record readable at the larger window — CONFIRMED as a theoretical edge, skipped as a blocker.** The skip is never silent: it emits an `oversizedSkipped` meta row and holds a true line boundary (D1 in readTail). It requires a single JSONL line ≥128 KB; the widest entry across four real transcripts measured for #655 was 6,066 bytes, and the per-entry inline budgets added in that PR keep normalized events from busting the page budget in the first place.
- **[P2] Same-inode in-place rewrite landing a newline exactly at the old cursor — CONFIRMED as theoretical, deferred.** Claude Code replaces transcripts by atomic rename (which the ino check catches); an in-place rewrite that preserves a line boundary at the exact cursor byte is not a writer that exists here. A cursor-fingerprint is the right shape if one ever does — noted for the phone-track design pass, where per-client cursors get revisited anyway.

Issue #756, opened 2026-08-02
browser_navigate reports an RPC timeout when the navigation guard's DNS lookup is slow


openwong2kim:
> **Correction #2 (author).** This issue has now had two wrong root causes from
> me. The first blamed `PlaywrightEngine.attemptAutoOpen`; the second blamed
> `ensureAwake`'s 15s wake budget sitting behind the 10s client deadline. Both
> are wrong. The measured cause is below. Reproduction data from the earlier
> versions was real, but I attributed it to the wrong mechanism.

## What happens

`browser_navigate` fails with

```
RPC timeout: browser.navigate (10000ms)
```

for a destination whose DNS is slow or does not resolve. The message names the
transport; the real failure is DNS, and it appears nowhere.

## Why

[The navigation guard](https://github.com/openwong2kim/wmux/blob/main/src/main/security/navigationPolicy.ts#L138)
resolves the hostname before any navigation happens, to check it against the
SSRF blocklist:

```ts
addresses = await lookup(hostname, { all: true, verbatim: true });
```

That await has no ceiling. `dns.lookup` inherits the OS resolver's retry
schedule, which can run well past ten seconds. Every client deadline in front of
it — [`TIMEOUT_MS = 10000`](https://github.com/openwong2kim/wmux/blob/main/src/mcp/wmux-client.ts#L8)
in the MCP client, [`TIMEOUT_MS = 5000`](https://github.com/openwong2kim/wmux/blob/main/src/cli/client.ts#L23)
in the CLI — therefore loses the race and reports its own timeout while the
resolver is still grinding.

## Reproduced

Same browser state, varying only the hostname:

| hostname | DNS latency (measured directly) | `browser_navigate` |
| --- | --- | --- |
| `example.org` | 7ms | **11ms**, succeeded |
| `example.com` | 11082ms, then `ENOTFOUND` | **10014ms**, `RPC timeout` |

(`example.com` failing to resolve is a quirk of the machine this was found on.
The mechanism is not: any hostname the resolver is slow to answer for produces
the same misleading error.)

## Why it is not the wake path

The previous version of this issue claimed `ensureAwake` blocks for
`WAKE_TIMEOUT_MS = 15_000` behind the client deadline. It does not fire here:
lightweight mode is default-off, so guests are not discarded, and
`waking discarded surface` / `wake timed out` appear **zero times** across every
`%APPDATA%\wmux\logs\main-*.log` on a machine that reproduces this bug.

## Fix

[...]

Issue #694, opened 2026-07-29
sendRpc's longPollMs has no caller — the server-side wait RPC it was built for does not exist yet


openwong2kim:
#675 added a per-call long-poll mode to the MCP RPC client. Nothing calls it. The PR says so plainly ("Plumbing only... the server-side wait RPC it exists for is separate work"), so this is a tracking issue, not a complaint — it exists so the code does not quietly rot into something nobody remembers the purpose of.

## What is there

https://github.com/openwong2kim/wmux/blob/main/src/mcp/wmux-client.ts#L205-L230

`sendRpc(method, params, { longPollMs })` raises that one call's timeout (capped at `LONG_POLL_MAX_MS`, 15 minutes) and switches its retries off. Both effects hang off the single option deliberately: "long timeout, keep retrying" would reopen the same wait several times over and must not be expressible.

The only consumers today are its own tests.

## Why it was added

An agent that delegates work has no way to wait for the answer. The MCP client's transport constants (10 s timeout, 3 retries) make a long wait impossible, and raising the global timeout would be strictly worse than the status quo — every ordinary call would hang three times as long on a dead daemon.

## What closes this

A daemon-side RPC that holds a request open until something happens (a task reaching a terminal state, an inbox gaining an entry) and an MCP tool that calls it with `longPollMs`. Related shape: the delegation contract documented in `docs/how-to/delegate-to-agents.md`, where "a channel post does not start a turn" leaves the sender with nothing to wait on.

If that work is not going to happen, the honest alternative is deleting the option rather than leaving a capped, retry-suppressed code path that has never run in production.

openwong2kim:
## [wmux-hermes] Triage Summary

This is a tracking issue for the per-call `longPollMs` option added to the MCP RPC client in #675, which raises a single call's timeout to a 15-minute cap and disables retries for that call — but which has no production caller today. The option exists because an agent that delegates work has no way to wait for the answer: the transport's default 10 s timeout / 3-retry constants make a long wait impossible, and raising the global timeout would make every ordinary call hang three times as long on a dead daemon. The design deliberately couples "long timeout" and "retries off" into one option so the dangerous "long timeout, keep retrying" combination is not expressible.

[...]

PR #776, opened 2026-08-03
Move panes around the layout (#645)

### LINKED ISSUE #645: [Feature]: Move Panes

openwong2kim:
Panes could only be created and closed. Getting the arrangement you wanted meant splitting in exactly the right order from the start, and if you got it wrong the only way back was to close panes and re-split — which is what #645 asked us to fix.

Now you can rearrange a layout you already have, three ways:

- **Drag** the grip in a pane's tab strip. Drop on another pane's edge to move it there; drop on its centre to swap the two. The pane you are hovering shows where it will land. `Esc` aborts.
- **Keyboard**: `Ctrl+B` then `H`/`J`/`K`/`L` walks the active pane left/down/up/right; `{` / `}` swap it with its neighbour, as in tmux.
- **Command palette**: the four directional moves.

Panes keep their names (ordinals are never reissued) and their scrollback wherever they land.

## Why there is no "keep the pane mounted" machinery

The plan for this started with a DOM-stable pane host: render each pane body once and move that node between layout slots, so a pane never unmounts and never loses anything. It was dropped after measuring, not after arguing.

A cross-parent move does recreate the xterm instance — but the daemon resync ladder restores the buffer faithfully. Measured on a live build: **500 → 500 lines, and at depth 6000 → 6000 lines including the oldest**. The only thing actually lost is the text selection. Browser surfaces are `<webview>` tags whose guest is recreated on any DOM reparent, so the host would not have saved them either. The complexity had nothing left to buy.

## Semantics

tmux, not a free-form grid. An insert always wraps the target in a fresh binary `[50,50]` branch, exactly as `splitPane` does, so a moved tree is indistinguishable from a split tree and every existing invariant keeps applying.

`movePane` and `swapPanes` both take an explicit `workspaceId` — a multiview gesture can start in a workspace that is not the active one, and the store cannot guess which. Focus is an input rather than a side effect.

## Pre-existing defects fixed along the way

- `closePane` spliced a child out and left `parent.sizes` untouched. With three or more children the survivors rendered at the wrong widths. Two-pane splits collapse the branch, which is why it stayed hidden. Now fixed in the shared `detachPane` primitive, with its own test.
- `PaneContainer`'s 200ms resize debounce had no unmount cleanup and no idea which children its sizes described, so a resize ending just before a restructure could overwrite the new layout.

## Verification

Unit and jsdom tests throughout (the drag test file is explicit about what jsdom cannot prove — the panels library never applies a layout without real measurement).

Live dogfood on a dev build, all passing: a terminal pane moved across the tree (PTY alive, scrollback intact), a browser pane moved (URL preserved), and move → restart → restore returning an identical tree with identical ordinals. The drag gesture itself was driven end to end: the indicator predicts the landing spot, `Esc` leaves both tree and focus untouched, and the tabs' existing text-export drag is unaffected.

Two bugs were caught only by driving the real app — a selector returning a fresh object each read (infinite render), and `Esc` cancelling the drag but still moving focus. Both are covered by tests now.

## Three-way review

Reviewed independently by Claude, Codex and GLM-5.2. No P1. Five P2s, all fixed with a regression test each: drag state stranded when a pane closes mid-drag, pointer capture never released, the grip dead in a background multiview tile, no focus after a centre swap, and `lostpointercapture` unhandled. Notably the three reviewers overlapped almost not at all — each found a different class of problem.

Closes #645

## Summary by CodeRabbit

* **New Features**
  * Rearrange panes by dragging and dropping, with support for swapping or moving panes to specific edges.
  * Move the active pane using keyboard shortcuts or command-palette commands.
  * Added pane movement and swapping actions to configurable key bindings.
  * Pane names, scrollback, focus, and layout sizing are preserved during rearrangement.

* **Bug Fixes**
  * Fixed incorrect pane sizing after closing a pane in layouts with three or more panes.

* **Tests**
  * Added comprehensive coverage for pane movement, dragging, swapping, sizing, and cleanup behavior.

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -437,7 +437,7 @@
 
 ## RPC timeout abandons live server work (P2, from #756 eng review outside voice 2026-08-02)
 - **What:** Give RPC a propagated deadline, cancellation, and a status/idempotency path so a timeout is recoverable instead of terminal.
-- **Why:** Restores the stated principle above ("timeout/예외 = unknown, 절대 dead 아님") instead of widening the window around the violation. Today slow work masquerades as failure and a retry can duplicate side effects (a navigation, a click).
+- **Why:** Restores the principle already stated further up this file — a timeout or an exception means `unknown`, never `dead` — instead of widening the window around the violation. Today slow work masquerades as failure and a retry can duplicate side effects (a navigation, a click).
 - **Context:** Surfaced by Codex outside voice reviewing the #756 plan, 2026-08-02. `wmux-client.ts:141` rejects and destroys the socket while the main-side handler keeps running uncancelled; `RpcRouter.ts:385` flattens errors to a bare string so no machine-readable cause survives the wire. #756 only bounds one handler wait (the DNS guard) and deliberately does NOT fix this. Touching the RPC envelope is a Substrate contract change (PROTOCOL.md, inventory.md, stability tier).
 - **Depends on:** Should follow #756 so the timeout path has a known-good case to test against. **Effort:** L. **Priority:** P2.
 
--- a/TODOS.md
+++ b/TODOS.md
@@ -428,3 +428,28 @@
 - **Depends on:** P0-0 measurement fixes. **Effort:** M → S-M (CC). **Priority:** P2-P3.
 
 - [ ] Update the release procedure in CLAUDE.md: main is a protected branch, so the chore(release) commit must land through a release PR (squash), with the v* tag pushed onto the release commit — direct `git push` to main is rejected. (2026-07-30, found while shipping v3.38.1)
+
+## Browser guest discard/wake path has never actually run (P3, from #756 eng review 2026-08-02)
+- **What:** Determine whether the #517 lightweight discard/wake path is reachable in practice, and either exercise it or retire the 15s wake budget.
+- **Why:** `ensureAwake`/`WAKE_TIMEOUT_MS = 15_000` (`WebviewCdpManager.ts`) is live code that appears never to execute: lightweight mode is default-off and `waking discarded surface` / `wake timed out` occur **zero times** across every `%APPDATA%\wmux\logs\main-*.log`. Untested, unexercised code that sits in the path of every browser RPC is a standing hazard — it was mistaken for the cause of #756 precisely because it looks load-bearing.
+- **Context:** Found 2026-08-02 while diagnosing #756. Earlier notes here claimed a discarded guest "fails to remount"; that observation came from `browser_close`, which *unregisters* the surface rather than discarding it, so no wake was ever attempted. Start by enabling lightweight mode and confirming a guest is discarded after `DISCARD_AFTER_MS`, then that `ensureAwake` wakes it.
+- **Depends on:** — **Effort:** M. **Priority:** P3.
+
+## RPC timeout abandons live server work (P2, from #756 eng review outside voice 2026-08-02)
+- **What:** Give RPC a propagated deadline, cancellation, and a status/idempotency path so a timeout is recoverable instead of terminal.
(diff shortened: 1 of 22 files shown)


openwong2kim:
Thanks — all four addressed in 0cfc8a6b. I checked each against eslint rather than taking the diagnosis as given, and one of them turned out to be right for a different reason than stated.

**`PaneContainer.tsx:59` — ref written during render.** Valid, fixed. `childIdKeyRef` moved into a `useLayoutEffect`. A discarded render (StrictMode, a concurrent re-render) would otherwise leave the ref describing children that were never committed, and the stale-write guard would then compare against a layout that does not exist.

**`PaneContainer.tsx:80` — the eslint suppression.** Removed, but the reason is not that the deps are complete. Running eslint on the file gives:

```
80:32  error  Definition for rule 'react-hooks/exhaustive-deps' was not found
```

This repo's eslint config has no `react-hooks` plugin, so the suppression names a rule that cannot be resolved and the comment *itself* is the error. It predates this PR (same error on `origin/main`'s copy at line 68) but it sits on a line this PR touches, so it goes now. `paneChildren` stays out of the dependency array on purpose — `childIdKey` already encodes the child set, and the array identity churns on unrelated store writes. That is now a comment instead of a suppression.

**`PaneDragGrip.test.tsx:42` — `no-empty-function`.** Valid, fixed with `vi.fn()`. Confirmed as real errors, not warnings.

**`TODOS.md:440` — Korean in a repo artifact.** Valid, fixed. Worth noting the line was quoting an older Korean entry further up the file (line 305); per this repo's rule new content is English but existing Korean is not mass-rewritten, so the quote is now an English rendering and the old line is untouched.

For transparency: that TODO entry, and the one above it, come from a different investigation (#756) and were swept into this branch by an over-broad `git add`. They are accurate notes, so I have kept them rather than dropping the content, but they are not part of the #645 change.

openwong2kim:

PR #776 implements pane rearrangement in the layout, closing the long-standing enhancement request #645. It adds three complementary inputs: a pointer-drag grip on each pane tab strip (edge drop to reparent, centre drop to swap, `Esc` to abort), `Ctrl+B` prefix shortcuts (`H`/`J`/`K`/`L` to walk the active pane, `{`/`}` to swap with a neighbour, mirroring tmux), and the four directional moves in the command palette. The store primitives (`movePane`, `swapPanes`, `moveActivePaneDirection`) follow the same `[50,50]` binary-branch shape `splitPane` already produces, so a moved tree is indistinguishable from a split one and existing invariants carry over unchanged.

It is renderer-only across 21 files (+1844/−80) and also lands two genuine defect fixes along the way: `closePane` left `parent.sizes` untrimmed so ≥3-child layouts rendered at wrong widths after a close, and `PaneContainer`'s 200 ms resize debounce could overwrite a freshly restructured branch (no unmount cleanup, no child-identity guard) — both now covered with regression tests. Pane names and scrollback are preserved across moves (the daemon resync ladder rebuilds the buffer), and the PR deliberately scopes out the RPC/MCP-facing `pane.move` and cross-workspace move, documenting them in `TODOS.md` as follow-ups. It has already passed a three-way review (Claude / Codex / GLM) with five P2s caught and fixed, and CodeRabbit is green; CI is still running at time of triage. Tagging P3 as a non-blocking, self-contained feature that is review-complete.

Skipped by the dedup filter this cycle: #770, #766, #694, #691, #690, #689, #666, #659, #645, #682 (all already triaged).

PR #777, opened 2026-08-03
Keep the mode menu and the composer note on screen

openwong2kim:
Two clipping fixes in the deck rail, both reported from a live session.

### The mode dropdown ran off the top of the window

The mode chip sits in a bar at the bottom of the deck rail, so its menu opens
upward. Three options with descriptions are taller than the room above that
chip in a short window, so the menu overflowed past the top of the screen and
the first option — `off` — was rendered but unreachable. Autonomy could be
raised and never lowered, which is the wrong direction for a control whose
whole job is to be the brake.

The menu now measures the room on each open, opens toward whichever side has
more of it, and caps its height there (scrolling when it has to). Two tests
cover both directions.

### The composer's mode-off note was cut off

The composer is a two-row box and it was showing the full two-sentence reason
as its placeholder, so the tail of the sentence was clipped. The placeholder
now carries a one-line form; the full sentence stays on the hover title, where
there is room for it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Workspace mode menus now automatically position above or below the control and remain scrollable within the available screen space.
  * Improved visibility of the orchestrator-disabled message in the deck composer with a concise placeholder and full details available on hover.
  * Added English and Korean translations for the updated composer message.
* **Tests**
  * Added regression coverage for menu positioning, height limits, and the disabled composer state.

### DIFF
--- a/changelog.d/777.md
+++ b/changelog.d/777.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- The workspace mode dropdown no longer runs off the top of the window. It
+  opens toward whichever side has room and caps its height there, so `Off`
+  stays reachable in a short window — autonomy can always be lowered again.
+- The orchestrator's "mode is off" note in the deck composer is no longer cut
+  off. The placeholder shows a one-line form; the full explanation stays on
+  hover.
--- a/src/renderer/components/Deck/AgentModeChip.tsx
+++ b/src/renderer/components/Deck/AgentModeChip.tsx
@@ -18,7 +18,7 @@
 // the container. Renders nothing when the preload is absent, so pure jsdom tests
 // of the parent view are unaffected.
 
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useEffect, useLayoutEffect, useRef, useState } from 'react';
 import { tokenAttrs } from '../../themes';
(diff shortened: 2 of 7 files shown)


openwong2kim:

This PR fixes two clipping defects in the deck rail that were surfaced from a live session. The workspace mode dropdown (rendered in `AgentModeChip.tsx`) opens upward from a chip anchored at the bottom of the deck column, so in a short window the three-item menu ran past the top of the screen and the `off` option became unreachable — meaning the operator could raise autonomy but never lower it again. The fix measures the room on each open (and on resize), flips the menu toward whichever side has more space, and caps its height there with scroll fallback, so every option stays reachable. The composer gets a parallel one-line form of the mode-off reason (`CommanderView.tsx`) because the full two-sentence note was being clipped inside the two-row composer box; the long form is preserved on the hover title.

Touches `AgentModeChip.tsx`, `CommanderView.tsx`, the en/ko locale strings, and ships regression tests for both menu directions plus an updated composer test. A changelog fragment is included. The work is small, self-contained, and addresses a genuine usability regression on the autonomy control, so I am tagging it P2. The branch is mergeable on `main`; I have no blocking notes — the measurement logic and the 120px floor read cleanly, and the tests pin the exact behavior described in the commit message.

PR #779, opened 2026-08-03
Give the Claude integrations a place to live

openwong2kim:
### The gap

wmux installs three things into Claude Code, and all three already had a
working IPC probe and a one-click install path:

| | |
|---|---|
| hook bridge | lifecycle signals — turn completion, approvals, subagents |
| MCP registration | the agent's access to wmux tools at all |
| usage statusline | model / context / rate limits under the input box |

The only UI that ever offered any of them was the first-run wizard. Skip it
once and the CLI (`wmux setup-hooks`, `wmux mcp register`,
`wmux setup-statusline`) was the entire re-entry path — an operator who could
not remember those commands had no way to discover the feature existed.

### Settings → Claude integration → Setup

A card at the top of the tab: one row per integration, state read from the
real probe rather than inferred, and an install button only where something is
missing. Required vs recommended is stated in words — the color grammar
reserves red for destructive, and a red "required" on a healthy install reads
as an error.

Two things it deliberately does not do:

- **No nagging.** No banner, no toast. The one existing nudge (the hook
  install prompt at launch and on a mode raise) is untouched. This is a place
  you go, not a thing that comes to you.
- **No writing without a click.** A probe that throws leaves its row `unknown`,
  never `missing` — an install button offered because IPC hiccuped would write
  to a config the user never asked us to touch. And `ok: true` with every
  target skipped is reported as a failure, not a success, because nothing was
  written.

### The wizard gets the hook bridge

It offered the statusline but never the hooks — the required half. Now it
offers both, and shows the hook row even when it is already installed: a
first-run checklist that silently omits its most important item teaches the
operator that item does not exist.

### The bug underneath

`statuslineBridge()` read `electronAPI.statuslineBridge` while the preload
exposes `electronAPI.deck.statuslineBridge`. The accessor's "older preload"
fallback turned the miss into `null`, and `null` means "hide the offer without
a word" — so the wizard's statusline block has never rendered in any real
build, and no test caught it because none referenced the bridge. Fixed, with a
contract test that pins both bridges to the preload path and fails if either
is read off the root again.

## Summary by CodeRabbit

- **New Features**
  - Added a Claude integration setup card in Settings for managing hooks, MCP tools, and usage statusline.
  - Added one-click installation and registration actions with clear status, error, and skipped-state feedback.
  - Updated the first-run wizard to offer Claude hook setup and show its current installation state.
  - Added English and Korean localization for the new setup experiences.

- **Bug Fixes**
  - Corrected statusline detection during first-run setup.

- **Tests**
  - Added coverage for integration states, installation flows, bridge handling, and wizard rendering.

### DIFF
--- a/changelog.d/779.md
+++ b/changelog.d/779.md
@@ -1,0 +1,14 @@
+### Added
+
+- Settings → Claude integration now opens with a **Setup** card listing the
+  hook bridge, the MCP registration, and the usage statusline — each with its
+  real install state and a one-click install for whatever is missing.
+  Previously the first-run wizard was the only place these were ever offered,
+  and skipping it left the CLI as the only way back.
+- The first-run wizard now offers the Claude Code hook bridge alongside the
+  statusline, and shows it as installed when it already is.
+
+### Fixed
+
+- The wizard's statusline offer never appeared: it looked for the preload
+  bridge at the wrong path and silently hid itself when it came back empty.
--- a/src/renderer/components/FirstRunWizard.tsx
+++ b/src/renderer/components/FirstRunWizard.tsx
@@ -54,8 +54,8 @@ function firstRunBridge(): FirstRunBridge {
   return api.firstRun;
 }
 
-// Statusline bridge (preload `statuslineBridge`). Optional: older preloads may
-// not expose it, so the accessor returns null instead of throwing and the
+// Statusline bridge (preload `deck.statuslineBridge`). Optional: older preloads
+// may not expose it, so the accessor returns null instead of throwing and the
 // wizard simply hides the statusline offer.
 interface StatuslineBridge {
   status: () => Promise<{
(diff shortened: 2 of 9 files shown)


openwong2kim:

This PR adds a dedicated **Setup** card under Settings → Claude integration that surfaces the real install state of all three integrations (hook bridge, MCP registration, usage statusline) with a one-click install for whatever is missing. Previously the first-run wizard was the only UI that ever offered them — skip it once and the CLI (`wmux setup-hooks`, `wmux mcp register`, `wmux setup-statusline`) was the entire re-entry path. The wizard itself also gains the hook bridge offer, which is the required half it never showed.

It also fixes a genuine bug underneath: `statuslineBridge()` read `electronAPI.statuslineBridge` while the preload exposes `electronAPI.deck.statuslineBridge`, so the "older preload" fallback turned the miss into `null` and the wizard's statusline block has **never rendered in any real build**. The new `IntegrationSetupSection` reads probe results rather than inferring state, leaves a row `unknown` rather than `missing` when a probe throws (so an IPC hiccup cannot trigger an unwanted write), and reports a per-target skip as a failure rather than a success.

**Priority: P2.** The dead statusline path means the offer has been silently broken in shipped builds, and surfacing the required integrations in a discoverable place materially improves onboarding. The card is additive and well-guarded against spurious writes, so risk is low.

PR #780, opened 2026-08-03
Add a skip-permissions checkbox to fan-out and remember the launch command

openwong2kim:
## What

Multi Task (fan-out) launched whatever sat in the **agent command** field, and that field reset to `claude` on every open. Running a fleet in bypass mode meant typing `--dangerously-skip-permissions` by hand every single launch, and a Codex user retyped their whole command each time.

The dialog now has:

- a **`--dangerously-skip-permissions` checkbox** next to the agent command,
- a **launch command preview** — the line the task pane will actually fire, including the appended prompt-file argument,
- **memory**: the command the last fan-out really launched is prefilled next time.

## How

The checkbox is a *projection* of the command string, not a second source of truth — typing the flag by hand ticks it, unchecking strips it. That is what keeps the preview honest.

It only appears for a launcher wmux knows the flag for (Claude Code), matching the rule the resume chip already follows (`agentSupportsPermissionFlag`): a Codex user is never shown a flag their CLI rejects. Instead they type their own bypass flag once and the memory keeps it. If that Claude-only flag is left behind after switching launchers, the dialog says so and offers a one-click **Remove**.

Parsing goes through the shared `tokenize`, so:

- a flag inside a quoted argument (`claude --append-system-prompt "never use --dangerously-skip-permissions"`) is neither counted nor destroyed on toggle,
- a quoted launcher path with a space (`"C:\Program Files\claude.cmd"`) still resolves to the `claude` stem.

The **restored** command is validated before it is prefilled: known agent launcher, no shell metacharacters, length-capped. main interpolates `agentCmd` unquoted into the PTY line, and the restore path is the one place a command reaches a shell without a human typing it — a typed command is trusted, a remembered one is checked.

`KNOWN_AGENT_STEMS` is exported from `orchestratorRole.ts` for that check (previously module-private).

## Tests

- `fanoutAgentCmd.test.ts` — 18 assertions across stem parsing, the claude-only gate, quote-aware detection/stripping, stale-flag detection, and restore validation (injection payloads rejected, Windows launcher path accepted).
- `FanOutDialog.skipPermissions.dynamic.test.tsx` — 6 cases driving the real dialog in jsdom: default state, check/uncheck round trip, hand-typed flag syncing the box, non-claude launcher hiding it, toggling from an empty field, and the stale-flag Remove button.

`tsc --noEmit` clean; full `test:parallel` shows only the two failures already present on this base (`genApiReference` drift guard, `deck.handler.loop`).

## Summary by CodeRabbit

* **New Features**
  * Fan-out remembers the most recently launched command.
  * Added a permission-bypass toggle for supported Claude commands.
  * Added command previews and warnings for unsupported or outdated flags.
  * Improved handling of prompt-file arguments and launch outcomes.
  * Added Korean localization for the new fan-out messaging.

* **Bug Fixes**
  * Safely removes stale permission flags and rejects unsafe command restoration.

* **Tests**
  * Added coverage for command parsing, flag handling, persistence, and dialog interactions.

### DIFF
--- a/changelog.d/780.md
+++ b/changelog.d/780.md
@@ -1,0 +1,3 @@
+### Added
+
+- **Fan-out can launch in bypass mode without retyping the flag.** The Multi Task dialog now offers `--dangerously-skip-permissions` as a checkbox, shows the launch command it will fire, and prefills the command the last fan-out actually used. The checkbox appears only for Claude Code, the launcher wmux knows that flag for; users of another CLI type their own bypass flag once and it is remembered. A Claude-only flag left behind after switching launchers is called out with a one-click removal.
--- a/src/renderer/components/AgentToolbar/FanOutDialog.tsx
+++ b/src/renderer/components/AgentToolbar/FanOutDialog.tsx
@@ -19,6 +19,16 @@ import { useT } from '../../hooks/useT';
 import { t } from '../../i18n';
 import Button from '../ui/Button';
 import Input from '../ui/Input';
+import {
+  SKIP_PERMISSIONS_FLAG,
+  applySkipPermissions,
+  hasSkipPermissions,
+  fanoutAgentStem,
+  hasStaleSkipPermissions,
+  loadLastAgentCmd,
+  saveLastAgentCmd,
+  supportsSkipPermissions,
+} from './fanoutAgentCmd';
 
 // 리뷰 발견(Codex+GLM+Claude 3/3 합의) — compete 모드에서 `mode === 'parallel' ?
 // taskPrompts : []`처럼 인라인 배열 리터럴을 useEffect 의존성에 넣으면 매 렌더마다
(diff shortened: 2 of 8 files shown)


openwong2kim:

This PR enhances the Multi Task (fan-out) dialog with a `--dangerously-skip-permissions` checkbox, a live launch-command preview, and memory of the last launched command via `localStorage`. The checkbox is modeled as a projection of the command string rather than independent state, which keeps the preview always WYSIWYG with what the task pane actually fires — typing the flag by hand ticks the box, unchecking strips it. The flag is only offered for Claude Code (matching `agentSupportsPermissionFlag`), and a stale Claude-only flag left on another launcher is surfaced with a one-click Remove.

The security design on the restore path is the part worth calling out specifically. `loadLastAgentCmd` validates the persisted value against shell metacharacters (`;&|``$(){}`), enforces a 512-char cap, and requires the launcher stem to be in `KNOWN_AGENT_STEMS` — because main interpolates `agentCmd` unquoted into the PTY line, and the restore path is the one place a command reaches a shell without a human typing it. Parsing goes through the shared `tokenize`, so a flag nested inside a quoted argument (e.g. `--append-system-prompt "never use --dangerously-skip-permissions"`) is correctly ignored rather than matched. Test coverage is solid: 18 unit assertions on the helpers plus 6 jsdom cases driving the real dialog through check/uncheck round trips, stale-flag removal, and empty-field toggling.

**Priority: P2** — touches the permission-bypass surface for a core workflow, and the injection-validation on the remembered command is the kind of thing worth reviewing carefully before merge. The overall structure and the decision to keep `KNOWN_AGENT_STEMS` as the restore allowlist are sound.

PR #786, opened 2026-08-03
feat(web): phone turn-view contract (delta, turns route, --allow-transcript, nudge)

openwong2kim:
Closes #782.

## What
The daemon half of the phone turn (conversation) view: the phone reads the same Claude Code session reflowed to phone width, **stateless** so it can never scramble the desktop Chat View sharing the pane.

## Changes
- `TranscriptProjector.delta()` — forward paging. Reset fires only on a shrink (size < cursor); rotation past the cursor is caught by `isLineBoundary`. `fitWithReceipt` surfaces `budgetDropped`.
- `GET /api/sessions/:id/turns` + `handleSessionTurns` — opaque cursor codec, 403 (with restart command) when `--allow-transcript` is off, 503 when the projector isn't wired, 200 `{available:false, reason}` (never 500) for no-hook / stale-session / not-claude / unsafe-transcript-path, 404 for unknown panes.
- `--allow-transcript` (default off) threaded web.ts → RPC → daemon → `WebPersistedState` → `/api/config` and `status()`. Field absent reads as false, so a pre-flag daemon falls back to the mirror with no 404 probe.
- Turn nudge is a NON-RECORDING `transcript.nudge` SSE frame: bypasses `publish()`/`attentionLog` so a busy pane cannot push an approval create out of the bounded log and blank the phone badge. Coalesced per pane to 1s, delivered only to principals watching that pane.
- `codeBlock` fetch checks `isLineBoundary(srcOffset)` before serving — the offset is an arbitrary-read primitive.
- `no-binding` splits into `no-hook` vs `stale-session`; `not-claude`/`unsafe-transcript-path` stay separate. `turnEvents.ts` status-reason set corrected to match the code.

## Verification
- `tsc --noEmit` 0; `test:parallel` 10038 passed; `test:runtime` 140 passed; `build:daemon` ok.
- Tests: readTail Korean multibyte + non-boundary fromOffset; Projector delta reset (shrink) + stateless (no subscriber disturbance) + huge tool result; WebTerminalServer turns 403/401/404/{available:false,reason} + the non-recording nudge leaving attentionLog untouched; korean.jsonl + huge-tool-result.jsonl fixtures.

## Review
3-model panel (Claude+Codex+GLM). 4 CRITICAL fixed: RPC/2nd-site/status wiring left the feature dark; delta reset on every append was dead code. 1 INFO fixed (watcher leak on session delete). 1 INFO deferred (lastDetectedAgent sticky mislabel — needs a live-process signal).

[...]

PR #788, opened 2026-08-03
fix(web): prevent transcript response caching

snowyukitty:
## Summary
- mark every response produced by `GET /api/sessions/:id/turns` as `Cache-Control: no-store`
- cover disabled, unavailable, snapshot, and forward-delta responses
- add the numbered changelog fragment for #788

## Why
The turn view can contain thinking blocks, full tool inputs, and contents from files the agent read. Without an explicit cache policy, a phone or intermediary can retain a replayable transcript response after the operator revokes transcript access.

This is a focused privacy/correctness follow-up to #786 and addresses the unhandled review finding at https://github.com/openwong2kim/wmux/pull/786#discussion_r3706538615.

## Verification
- `npx vitest run src/daemon/web/__tests__/WebTerminalServer.test.ts --maxWorkers=1` — 154 passed
- `npx eslint src/daemon/web/WebTerminalServer.ts src/daemon/web/__tests__/WebTerminalServer.test.ts` — 0 errors; 9 unchanged warnings outside the diff
- `npx tsc --noEmit` — passed
- `npx tsc -p tsconfig.daemon.json --noEmit` — passed
- `npm run build:daemon` — passed
- `npm run test:parallel` — 681 files / 10,041 tests passed; 2 files / 24 tests skipped
- `node scripts/collect-changelog.mjs --check` — passed with 3 fragments (2 Fixed, 1 Added)
- `git diff --check upstream/main...HEAD` — passed

### DIFF
--- a/changelog.d/788.md
+++ b/changelog.d/788.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Phone turn-view responses now use `Cache-Control: no-store`, preventing browsers and intermediaries from retaining cached transcript pages after transcript access is revoked. (#788)
--- a/src/daemon/web/WebTerminalServer.ts
+++ b/src/daemon/web/WebTerminalServer.ts
@@ -1735,6 +1735,10 @@ export class WebTerminalServer {
     sessionId: string,
     principal: WebPrincipal,
   ): void {
+    // Transcript pages can contain thinking blocks, full tool inputs, and file
+    // contents. Keep every response on this route out of client/intermediary
+    // caches so revoking transcript access does not leave a replayable copy.
+    res.setHeader('Cache-Control', 'no-store');
     if (this.opts?.allowTranscript !== true) {
       this.json(res, 403, {
         error: 'transcript-disabled: server started without --allow-transcript',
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR sets `Cache-Control: no-store` on every response from `GET /api/sessions/:id/turns`, covering all four paths (disabled 403, unavailable 200, snapshot 200, and forward-delta 200). The turn view can carry thinking blocks, full tool inputs, and file contents the agent read, so without an explicit cache directive a phone browser or an intermediary (CDN, proxy) can hold a replayable transcript copy even after transcript access is revoked server-side. It is a focused correctness/privacy follow-up to #786 that picks up a review finding left unhandled there.

The diff is tight — a single `res.setHeader` placed before the first response branch, one new test asserting the header across all response shapes, and a changelog fragment. All CI is green (validate, Windows/macOS/Ubuntu baselines, perf, CodeRabbit), it merges cleanly into `main`, and the added test (`★ turn-view responses are never cacheable`) exercises the unavailable, snapshot, and delta responses so the header cannot silently regress on any one path. Worth merging as-is.

Priority: **P3** — a hardening measure on working functionality rather than an active bug; the route already functioned correctly, this closes the cache-retention gap. P2 would be justified if there are known deployments where intermediaries actively cache this route.

PR #791, opened 2026-08-04
fix(transcript): preserve budget-drop receipt on reset

snowyukitty:
## Summary

- preserve the `budgetDropped` receipt when a forward transcript cursor resets to a fresh snapshot
- keep the existing reset, cursor, paging, and snapshot semantics unchanged
- add focused forward-path and reset-path regression coverage

This is a correctness follow-up to #786 and addresses [the unhandled review finding](https://github.com/openwong2kim/wmux/pull/786#discussion_r3706538581). It does not close a separate issue.

The reset regression was confirmed before the implementation change: the targeted file had 40 passing tests and one failure because the reset result returned `budgetDropped: undefined`; the forward-path receipt test already passed.

## Verification

- [x] `npx vitest run src/daemon/transcript/__tests__/TranscriptProjector.test.ts --maxWorkers=1` — 41 passed
- [x] `npx eslint src/daemon/transcript/TranscriptProjector.ts src/daemon/transcript/__tests__/TranscriptProjector.test.ts` — 0 errors; 10 pre-existing non-null-assertion warnings in the test file
- [x] `npx tsc --noEmit`
- [x] `npx tsc -p tsconfig.daemon.json --noEmit`
- [x] `npm run build:daemon`
- [x] `npm run test:parallel` — 10,048 passed, 24 skipped (681 files passed, 2 skipped)
- [x] `node scripts/collect-changelog.mjs --check`
- [x] `git diff --check upstream/main...HEAD`

Not run: the runtime test suite and full-repository lint. The selected verification uses the focused lint, both TypeScript projects, daemon build, and the full non-runtime suite.

## Summary by CodeRabbit

- **Bug Fixes**
  - Phone clients now receive a `budgetDropped` indicator when oversized transcript entries are omitted from reset snapshots.
  - Reset and forward transcript updates now consistently respect event size limits.
  - Transcript cursors advance correctly when oversized entries are skipped, preventing repeated processing.

- **Tests**
  - Added coverage for oversized transcript entries in forward and reset updates.

### DIFF
--- a/changelog.d/791.md
+++ b/changelog.d/791.md
@@ -1,0 +1,5 @@
+### Fixed
+
+- **Transcript reset pages retain omission receipts.** Phone clients now receive
+  `budgetDropped` when a reset snapshot skips an oversized transcript entry, so
+  they can render an omission seam instead of silently closing the gap. (#791)
--- a/src/daemon/transcript/TranscriptProjector.ts
+++ b/src/daemon/transcript/TranscriptProjector.ts
@@ -232,9 +232,16 @@ export class TranscriptProjector {
     // A shrunk file (stat.size < from) is readTranscriptDelta's own reset path.
 
     if (reset) {
-      const page = this.snapshot(sessionId);
+      const { result: page, budgetDropped } = this.fitWithReceipt((maxBytes) =>
+        readTranscriptPage(resolved.transcriptPath, { maxBytes }),
+      );
       if (!page) return null;
-      return { events: page.events, cursor: page.cursor, reset: true };
+      return {
+        events: page.events,
+        cursor: page.cursor,
+        reset: true,
+        ...(budgetDropped ? { budgetDropped: true } : {}),
+      };
     }
 
     const { result, budgetDropped } = this.fitWithReceipt((maxBytes) => {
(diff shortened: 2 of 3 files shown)


openwong2kim:

This is a small, well-scoped correctness follow-up to #786 that fixes an unhandled review finding in the transcript delta path. The reset branch of `TranscriptProjector.delta` was using `this.snapshot()` directly, which silently dropped the `budgetDropped` receipt — so when a cursor reset produced a fresh snapshot that skipped oversized entries, phone clients received no signal that events had been omitted for budget reasons and could not render the omission seam. The fix routes the reset path through the same `fitWithReceipt` helper already used by the forward path, preserving the receipt and keeping both code paths consistent. Verification is thorough (focused vitest run, both TypeScript projects, daemon build, full non-runtime suite at 10,048 passed), and the two new regression tests lock both the forward-path behaviour and the newly-fixed reset-path behaviour independently.

Priority: **P3** — this is a polish-level correctness fix to a data-contract field in an actively-shipping feature (the #782 phone turn-view), not a functional breakage; the client silently misses an omission seam on reset but nothing crashes or corrupts.

PR #792, opened 2026-08-04
fix(logSink): stop the EPIPE recursion storm and add rotation

openwong2kim:
## Problem

When the reader of an inherited stdio pipe goes away while the process keeps
running, the log sink's pass-through write fails with `EPIPE`. On macOS that
failure is delivered **asynchronously, as an `error` event on the stream**,
not as a synchronous throw — and nothing listened for it. Node promotes an
unhandled stream `error` to `uncaughtException`, whose handler reported it
with `console.error`, which routed straight back through the tee onto the same
broken pipe.

```
console.error(err)
  → tee → appendFileSync(dailyLog)        ← file grows, no byte cap
        → orig(chunk) → returns normally
            → async 'error' {code:'EPIPE'}
                → no listener → uncaughtException
                    → console.error(...)  ← loop closes
```

Each cycle is a complete, cleanly-exiting call, so this is a hot loop through
the event loop rather than stack recursion — it never overflows and never
self-limits. It ends only when the process dies.

Measured from the log samples on disk:

| Day | Size | EPIPE lines |
|---|---|---|
| 2026-07-20 | 5.17 GB | 7,211,714 |
| 2026-07-26 | 17.7 MB | 31,312 |
| 2026-07-27 | 30.8 MB | 66,422 |
| **2026-07-31** | **84.9 GiB (91 GB)** | ~3.35 M sampled |
| **2026-08-02** | **5.0 GiB** | ~99 M over 521 s |

Peak rate ~190,000 lines/s (~12,700 EPIPE cycles/s), flat from the first
second. The necessary precondition is a build whose stdio pipe reader has gone
away while the process survives; a renderer crash preceded 2 of 3 observed
storms but is not required.

The existing mitigations — a `writing` reentrancy flag and a `try/catch`
around the pass-through — were both already present and both structurally
unable to see this: the `try/catch` only catches synchronous throws, and the
flag is released before the async event ever fires. They are correct for
Windows, where pipe writes are synchronous, and blind to macOS.

## How this was established

A first implementation was landed from the stack traces alone. It was then
re-investigated independently, with three parallel agents working from
separate angles — a code trace of the write path, a forensic pass over the
retained log samples, and a fix design benchmarked against Node/pino/winston
practice. They reached the same root cause, which raised confidence in the
direction. The cross-check against the implementation also surfaced two
blockers the first pass had missed, both fixed here.

## Fix

- **Consume the asynchronous stream error.** `createResilientTee` listens for
  `error` and permanently retires that pass-through. No re-probe: the probe
  write is itself what regenerates the storm. Lines keep reaching the file, so
  nothing is lost. The synchronous-throw and write-callback paths disable it
  too — three distinct delivery channels for the same failure.
- **Bound the file.** Daily logs are capped at 16 MiB with three numbered
  archives (64 MiB/day). Previously there was no byte cap at all, only a
  14-day prune, which is what allowed 84.9 GiB. A legacy oversized file is
  tail-truncated before it can become an equally oversized archive.
- **Make rotation safe across processes.** An installed build and a dev build
  derive the log path from the same log directory and interleave in one file.
  Rotation now takes the size from `stat` after every write (`O_APPEND` writes
  are atomic per call, so that is the only size true for every writer) and runs
  under an exclusive `<file>.lock`, re-checking the size while holding it. A
  lock held by another process defers rotation instead of duplicating the
  rename chain and losing an archive; a lock abandoned by a crashed process is
  broken once stale.
- **Fix the daemon sink too.** It carried the identical pattern with the
  identical blind spot. The transport primitives moved to
  `src/shared/logTransport.ts` so the daemon can use them without importing
  `electron`, and both sinks now guard their `uncaughtException` /
  `unhandledRejection` handlers against reporting a broken pipe through
  `console`.
- **Explain a disabled stream.** Retiring a pass-through writes one line to
  the file naming the stream and the error code. Any stream error still
  disables forwarding — narrowing the set risks an unanticipated code
  restarting the storm — but silence with no trace made a dead console
  indistinguishable from a process that stopped logging.

A comment in the daemon sink claimed it was the more exposed of the two
because the launcher spawns it detached. That is inverted: `stdio: 'ignore'`
opens `/dev/null` for fds 0-2, and writes to `/dev/null` cannot fail with
EPIPE. Corrected — the real exposure is a foreground daemon.

## Verification

| Gate | Result |
|---|---|
| `npm run test:parallel` | 10,055 passed, 21 skipped |
| `npm run test:runtime` | 140 passed, 12 skipped |
| `tsc --noEmit` (base + daemon) | clean |
| `build:daemon` / `build:mcp` / `build:cli` | ok |
| `eslint` (new and changed files) | clean |

Six new tests, none of which existed before — the sinks had no coverage at
all. They pin: no recursion after an asynchronous EPIPE (one pass-through
write, then 100 file-only writes), the synchronous-throw path, rotation and
archive retention, oversized-write splitting, legacy-file bounding, shared-file
rotation without overshoot, lock deferral, stale-lock breaking, and the
disabled-stream notice.

The 20 pre-existing `eslint` errors in `src/daemon/index.ts` are untouched —
identical counts before and after.

## Notes

Touches only the log sink files plus two handler guards in
`src/daemon/index.ts`, so it does not overlap #790.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented broken standard-output and standard-error pipes from causing repeated errors or disrupting application shutdown.
  * Continued capturing logs to files when inherited output streams fail.
  * Added resilient handling for synchronous and asynchronous stream failures.

* **Improvements**
  * Limited daily log files to 16 MiB with up to three rotated archives.
  * Improved safe log rotation across concurrent processes.


openwong2kim:
@codex review

openwong2kim:

This PR fixes a severe log-sink defect where a broken inherited stdio pipe (EPIPE delivered asynchronously on macOS) drove an unbounded recursion through `uncaughtException` → `console.error` → the same broken pipe, growing daily log files to **84.9 GiB** in a single session and peaking near 190,000 lines/second. The root cause — an asynchronous `error` event that the existing synchronous `try/catch` and reentrancy flag were structurally blind to — is now eliminated: `createResilientTee` listens for the error event and permanently retires the pass-through, so file mirroring continues while console forwarding stops for good.

The change set is well-scoped across 9 files (+684 / −110). The transport primitives (`createResilientTee`, `BoundedLogWriter`, `isBrokenPipeError`) are factored into a new `src/shared/logTransport.ts` so both the main and daemon sinks share the same protection without the daemon importing Electron. Bounded rotation (16 MiB cap, 3 archives) is made multi-process safe by reading the real file size from `stat` after each write and running rotation under an exclusive `<file>.lock` with stale-lock breaking — this correctly handles installed and dev builds interleaving in one daily file. Six brand-new tests pin the async-EPIPE retirement path, synchronous-throw path, rotation/archive retention, oversized-write splitting, legacy-file bounding, shared-file rotation, lock deferral, and stale-lock breaking; verification reports 10,055 tests passing with clean `tsc`/`eslint`.

**Priority: P1.** This is not an annoyance — it is a disk-filling fault that already produced an 84.9 GiB file and was verified against three independent investigation angles. The fix is complete, thoroughly tested, and touches only the log-sink files plus two handler guards, with no overlap with #790. Ready to merge once CI is green.

PR #796, opened 2026-08-04
Deliver liveness, code-block fetch, and gate state to the phone

openwong2kim:
## Summary

Three additions to the daemon's phone surface, handed over from the iOS turn-view
work: each one was a contract gap that made a planned iOS feature unbuildable.

**Liveness header.** The turn view wants a persistent activity header — thinking,
tool running with elapsed time, waiting for you — and that header cannot come from
the turn snapshot: an agent that stalls mid-turn writes nothing, so a
snapshot-derived header shows a frozen pane as a working one. `agent.tool_started`
already existed for exactly this ("feeds the phone header" is in its comment) but
never left the main pipe. Each `agent.event` is now projected onto a header state
and fanned out over `/api/events` as `agent.liveness`, following the three rules
`transcript.nudge` set: non-recording, coalesced per pane, watchers only.

**Code block fetch.** Turn pages carry code blocks and large tool outputs as chips
and never the body — one 256KB tail re-encoded with bodies would blow past main's
control-buffer cap. The desktop expands a chip over the `daemon.transcript.codeBlock`
RPC; the phone had no equivalent, so every chip it rendered was display-only. Adds
`GET /api/sessions/:id/turns/block`, the same bounded line read behind the same
`--allow-transcript` grant.

**Gate state.** `/api/config` listed which tools are gated but never whether the
gate was armed, so a client's toggle opened on a guessed default and could only
learn the truth by flipping something. Adds `gateEnabled`, plus a `gate.state` push
so one phone disarming the gate does not leave every other device showing the old
position.

## Review

Two-model panel (Claude + GLM) on the finished diff. Six defects found and fixed in
`4c7e2224`, four of them verified against the code before fixing:

- `agent.subagent_stop` arrives as `status:'complete'`, so every finished subagent
  flipped the header to idle on a pane that was still working — and being a settled
  state it skipped coalescing.
- Nothing released the header from "waiting on you". The wide PostToolUse hook that
  would have is gone, so after a phone answered a gate (or the 120s self-defer fell
  back to the local prompt) the elapsed counter kept climbing.
- `/api/config` reported the runtime gate flag rather than the effective state. The
  daemon arms the gate only when the flag is clear AND the web server can resolve
  gates, so a read-only server let everything through while config said "on".
- The orchestrator brain's TUI was readable. It is a live session, `listSessions`
  excludes it, and both transcript routes checked existence alone. Both now answer
  404 — the same as a missing pane, so the id is not confirmed.
- The 1MB body cap was unreachable (the projector reads one line up to 512KB), so
  the documented `truncated` flag could never appear. Now 256KB, and the contract
  says what a 404 on an oversized line actually means.
- `tool_name` reached every watching device unvalidated. Bounded and stripped.

## Test plan

- [x] `npx tsc --noEmit` clean
- [x] Full suite: 10113 passed. Two failures are outside this diff and pre-existing
      on the base — `THIRD_PARTY_NOTICES` drift and a `bridgePermissionGate`
      subprocess timeout.
- [x] 10 new tests: liveness projection (including the subagent regression and tool
      name sanitizing), non-recording / watcher-only / immediate-flush delivery,
      block route stale-ref + param validation + cap, brain-pane 404, effective gate
      state and its SSE push.

## Follow-ups not taken here

- `at` is the daemon's clock and the phone renders elapsed from it, so device clock
  skew lands in the number. Switching to a relative value would drop the coalescing
  delay, so this is worth revisiting only if iOS actually sees bad values.
- Gate transitions leave no audit trail. Recording who disarmed it is worth doing,
  but it turns `gate.state` into a recorded event and re-opens the replay contract.

## Summary by CodeRabbit

- **New Features**
  - Added live activity indicators for agent panes, including working, waiting, permission, and completed states.
  - Added mobile access to code blocks and large tool outputs from transcripts.
  - Added synchronized permission-gate status across connected clients.

- **Bug Fixes**
  - Prevented phones from accessing the orchestrator’s hidden conversation through its session ID.
  - Improved status updates after permission approval so activity no longer appears stuck.

- **Documentation**
  - Documented live activity and permission events, reconnect behavior, transcript expansion, access rules, and size limits.

### DIFF
--- a/changelog.d/796.md
+++ b/changelog.d/796.md
@@ -1,0 +1,21 @@
+### Added
+
+- **The phone can show what a pane is doing right now.** The turn view carries a
+  live activity header — thinking, which tool is running and for how long, waiting
+  for you — instead of leaving you to guess from a conversation that has stopped
+  growing. It rides its own channel rather than being read off the transcript, so
+  an agent that stalls mid-turn looks different from one that is still working.
+- **Code blocks and large tool outputs open on the phone.** Tapping a code chip
+  fetches the body instead of showing a chip that does nothing. Oversized bodies
+  arrive as a head and say they were cut, rather than pretending the block ends
+  there.
+- **The permission gate's state is visible and stays in sync.** A phone's gate
+  toggle opens showing whether the gate is actually armed, and turning it off from
+  one device updates the others immediately instead of leaving them showing the old
+  position until something else makes them re-read.
+
+### Fixed
+
+- **The orchestrator's own conversation is no longer readable from a phone.** The
+  brain pane is hidden from the pane list, but the transcript routes only checked
+  that a session existed — a device holding its id could read the whole thing.
--- a/docs/phone-client-contract.md
+++ b/docs/phone-client-contract.md
@@ -186,6 +186,7 @@ JSON backlog fetch (Bearer only).
 | `approval` | `{sessionId, approvalId, phase, state, agent, createdAt, tier, risk?, ...}` |
 | `transcript.nudge` | `{sessionId}` — the turn view for that pane has new content; re-fetch |
 | `agent.liveness` | `{sessionId, state, tool?, agent, at}` — what the pane is doing right now |
+| `gate.state` | `{gateEnabled}` — the permission gate was armed or disarmed |
 
 `phase` is `create` / `resolve` / `expire` / `supersede`.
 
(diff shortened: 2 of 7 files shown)


openwong2kim:

This PR delivers three contract additions to the daemon's phone surface that were blocking the iOS turn-view work: a live agent-liveness header (`agent.liveness` SSE event), a code-block/body fetch route (`GET /api/sessions/:id/turns/block`), and exposed permission-gate state (`gateEnabled` on `/api/config` plus a `gate.state` push). It touches `WebTerminalServer.ts` (the new routes, per-pane coalescing, and watcher-only SSE fan-out), a new `agentLiveness.ts` projection module, the daemon wiring in `index.ts`, the phone-client contract docs, and adds 10 targeted tests covering the subagent-stop regression, the effective-gate-state read, the brain-pane 404, and the mid-character truncation seam.

The bundled review-pass commit meaningfully raises the quality bar, and the most consequential fix in it is a security one: the orchestrator brain pane was readable through both transcript routes by any device holding its (guessable, approval-payload-carried) session id, and both routes now answer 404 identically to a missing pane so the id is not confirmed. The liveness projection also correctly treats `agent.subagent_stop` as non-terminal (it arrives as `status:'complete'` while the pane's own turn keeps going), reports the *effective* gate state rather than the runtime flag alone, and sanitizes the untrusted `tool_name` off the local pipe before it reaches every watching device.

Priority is P2: this is substantial, well-tested feature work with a real security fix folded in, already through a two-model panel review and mergeable. One open CodeRabbit nit remains — an empty `srcOffset=` currently parses as `0` rather than rejecting as a bad ref — which is worth addressing before merge but does not block the core functionality.

PR #798, opened 2026-08-04
Drop each pane's render model when the atlas guard wipes the pool

openwong2kim:
## What

`atlasGuard`'s rebuild emptied the shared WebGL glyph atlas and then called `terminal.refresh()` on each owner pane. That is only half of what upstream does, and the missing half is what makes the repair work.

`refresh()` re-runs `WebglRenderer._updateModel` over every row, but the loop skips any cell whose content is unchanged:

```ts
// WebglRenderer.ts:507 — @xterm/addon-webgl 0.19.0
// Nothing has changed, no updates needed
if (this._model.cells[i] === code && ...bg && ...fg && ...ext) {
  continue;
}
```

The skip is keyed on `code/bg/fg/ext` alone and knows nothing about atlas state. So every unchanged cell kept the vertex UVs written when it was last rendered — coordinates into a pool that had just been emptied and was re-packing from scratch. The pane sampled the new atlas at old positions: scattered wrong glyphs, worst on CJK (11,172 Hangul syllables churn the pool while ASCII sits undisturbed in the first page).

Upstream never separates the two operations — `WebglRenderer.clearTextureAtlas()` (`WebglRenderer.ts:307`) is exactly `_charAtlas.clearTexture(); _clearModel(true); _requestRedrawViewport()`.

## Why it appeared now

Latent until #790. While `clearTexture()` was a permanent no-op the rebuild changed nothing, so it could not desync anything. Fixing the clear armed it.

Field log, v3.38.7 on 2026-08-05 — nine wipes in the 90s before a corrupted-Hangul screenshot, and **zero** merge cures, so the merge path never ran:

```
20:40:58Z prevent — pages=16/16 used, cap=16, panes=3, clear=cleared
20:41:02Z prevent — pages=13/16 used, cap=16, panes=2, clear=cleared
20:41:38Z prevent — pages=16/16 used, cap=16, panes=3, clear=cleared
20:41:40Z prevent — pages=13/16 used, cap=16, panes=4, clear=cleared
```

## Changes

- `rebuildGroup` wipes the shared pool once, then calls each owner addon's public `clearTextureAtlas()` (its own `clearTexture` short-circuits by then, so the pool is still emptied exactly once). Order matters: a model dropped before the wipe would repopulate from the doomed atlas.
- Warn when a pane cannot be reached, instead of half-repairing in silence.
- Rate-limit PREVENT (`GUARD_PREVENT_COOLDOWN_MS`, 30s). A rebuild now re-rasters every pane, and a saturated pool re-arms the gate on the next poll. At saturation there is nothing left to prevent, and a merge that does land is repaired coherently by CURE — which is never rate limited.

## Tests

7 new cases in `atlasGuard.test.ts`, and the fake addon now models `clearTextureAtlas` faithfully:

- every rebuild drops each pane's render model, not just the shared texture
- model dropped on CURE and `recoverNow`, not only PREVENT
- pool wiped **before** any model is dropped (ordering)
- warns when a pane cannot drop its model
- `clearRenderModel` degrades to `false` on a reshaped/dead addon
- PREVENT rate-limited while the pool stays saturated
- the cooldown never delays CURE

`31 passed` in `atlasGuard.test.ts`, `213 passed` across `src/renderer/terminal` + `useTerminal.atlasClear`, `tsc --noEmit` and `eslint` clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved rendering reliability for CJK characters during heavy multi-pane output.
  * Panes now refresh correctly after shared glyph resources are cleared.
  * Reduced unnecessary recovery operations through rate limiting.
  * Added clearer diagnostics when pane rendering data cannot be refreshed.

### DIFF
--- a/changelog.d/798.md
+++ b/changelog.d/798.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- Terminal panes no longer render scattered wrong glyphs (worst on Korean and
+  other CJK text) during heavy multi-pane output. The shared-atlas guard emptied
+  the glyph pool and repainted, but a repaint skips every cell whose text has not
+  changed — so those cells kept pointing at glyph positions that no longer
+  existed. Each pane now rebuilds its render model alongside the wipe, the way
+  the renderer does it internally. The guard's speculative wipes are also rate
+  limited, so a saturated glyph pool no longer means a full re-raster of every
+  pane every two seconds.
--- a/src/renderer/terminal/__tests__/atlasGuard.test.ts
+++ b/src/renderer/terminal/__tests__/atlasGuard.test.ts
@@ -3,6 +3,7 @@ import {
   clearAtlasTexture,
   createAtlasGuard,
   detectMerge,
+  clearRenderModel,
   extractAtlas,
   GUARD_POLL_MS,
   GUARD_MARGIN_PAGES,
@@ -29,8 +30,20 @@ class FakeAtlas {
    *  effect (upstream early-returns unless _pages[0] is idle), so PREVENT
    *  re-fires on every tick instead of being gated by the anti-thrash rule. */
   clearIsNoop = false;
+  /** clearTexture calls made by the guard's own pool wipe. Kept separate from
+   *  `paneClears` so a test can still say "the guard rebuilt N times" — the
+   *  per-pane model clear reaches the same upstream method. */
   clearTexture(): void {
     this.clearCalls++;
+    this.applyClear();
+  }
+  /** clearTexture reached through a pane's `addon.clearTextureAtlas()`. */
+  paneClears = 0;
+  clearTextureFromPane(): void {
+    this.paneClears++;
+    this.applyClear();
+  }
+  private applyClear(): void {
     if (this.clearIsNoop) return;
     // Faithful to upstream: it decides "already clean, nothing to do" by
     // probing ONLY page 0, then empties every page IN PLACE (count unchanged).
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR fixes a regression in the WebGL glyph atlas guard where rebuilding the shared texture atlas corrupted rendering instead of repairing it — scattered wrong glyphs, worst on CJK (the field report is corrupted Hangul on v3.38.7). The root cause is precise: `rebuildGroup` emptied the shared pool and called `terminal.refresh()`, but `WebglRenderer._updateModel` skips any cell whose `code/bg/fg/ext` is unchanged, so those cells kept vertex UVs pointing into a pool that had just been re-packed from scratch. The fix pairs the wipe with a per-pane `clearTextureAtlas()` (upstream's own clear+model-clear+redraw sequence), in the correct order — pool wipe first, model drop after — and adds a 30s cooldown on the speculative PREVENT path since the rebuild is now a full re-raster of every pane. CURE is correctly exempt from the cooldown so real merges are still repaired immediately.

The change is well-contained to `atlasGuard.ts` and its tests, the seven new cases cover the failure modes that matter (per-pane model drop, ordering, partial-repair warning, PREVENT rate-limiting vs CURE exemption), and the reported test run is green across the renderer suite. One thing worth a careful look on review: `clearRenderModel` swallows all exceptions from `clearTextureAtlas()` as a soft `false`, which is the right call for a disposing pane but means a genuinely broken addon fails silently into the warn path rather than surfacing the underlying error — acceptable, but the warn message could name the pane so an operator can tell which one is stale.

Priority: P1 — this is a visible rendering regression on the core terminal surface that the prior atlas fix (#790) directly armed, and it hits CJK text hardest, which is the primary script for this user base.

openwong2kim:
Rejected the date finding: 2026-08-05 is correct. The cited log lines are UTC (`2026-08-04T20:40:16Z` … `20:41:40Z`), which is 2026-08-05 05:40-05:41 KST — the same minute as the corruption screenshot. The timezone mismatch was genuinely confusing to read, so both comments now state it.

Issue #799, opened 2026-08-04
Workspace vanished + a confirmed-closed workspace resurrected during CLI create/close churn (v3.38.6, Windows 11, no daemon restart)


krm0007:
### What happened?

While scripting the CLI (`new-workspace` / `close-workspace` / `send` / `read-screen`), a pre-existing workspace that was never passed to any `close-workspace` call disappeared from `list-workspaces`, and a workspace the CLI had confirmed closing (`Closed workspace: ws-…`) reappeared and is now the only workspace listed. The daemon `bootId` was unchanged throughout, so this was not an app/daemon restart.

**Environment:** wmux 3.38.6 (Squirrel install), Windows 11 Pro 10.0.26200. App + daemon running throughout (daemon uptime continuous, `bootId b3da5dad…` unchanged before/after). Claude Code hooks installed via `wmux setup-hooks`.

**Expected:**
- A workspace never named in any `close-workspace` call survives other workspaces' create/close cycles.
- A workspace acknowledged as closed stays closed.

**Hypothesis:** smells like workspace-state persistence racing with rapid create/close (state file overwritten with a stale snapshot?), since the daemon itself never restarted.

### Steps to reproduce

All in one daemon session (bootId unchanged):

1. Baseline: `list-workspaces` shows `ws-20a438a4… "Workspace 1"` with 6 ptyIds (6 Claude Code panes), plus nothing else.
2. A script created 4 scratch workspaces over ~20 minutes (`new-workspace --name …`), started `claude` in their panes via `send`/`send-key`, read screens with `read-screen --pane <ptyId>`.
3. Cleanup ran `close-workspace <id>` ONLY on those 4 script-created ids. One of them, `ws-14993e40… "wmux-delegate-probe"`, was acknowledged: `Closed workspace: ws-14993e40…`.
4. Afterwards: `Workspace 1` (`ws-20a438a4…`) is gone from `list-workspaces`; `read-screen --pane daemon-44eb4770` (one of its panes) exits 255; only 2 of its 6 `claude` cli.js processes survive.
5. `ws-14993e40…` — the one explicitly confirmed closed — is back in `list-workspaces` and is now the only workspace (agentStatus `complete`).

### Expected behavior

- A workspace never named in any `close-workspace` call survives other workspaces' create/close cycles.
- A workspace acknowledged as closed (`Closed workspace: ws-…`) stays closed.

### wmux version

3.38.6

### How did you install wmux?

Setup.exe (direct download)

[...]

PR #801, opened 2026-08-05
Fix two CLI lifecycle defects: meta.* pane identity and workspace.close false receipts

### LINKED ISSUE #800: CLI: set-status / set-progress always fail with "cannot resolve the calling pane's workspace — send a verified senderPtyId" (v3.38.6)
### What happened?

`wmux set-status <text>` and `wmux set-progress <n>` fail unconditionally with:

```
Error: meta.setStatus: cannot resolve the calling pane's workspace — send a verified senderPtyId
```

They fail when invoked from a shell running inside a wmux pane (a context where `wmux send` and `wmux read-screen` self-resolve the caller fine), from descendant processes of that shell, and also when an explicit `--pane <ptyId>` is passed (the flag appears to be ignored for `meta.*` commands).

**Why it matters:** agent orchestration wants to publish a run's live status/progress into the workspace chrome (status line / progress bar) — currently there is no working write path to those fields from automation at all.

### Steps to reproduce

1. Open any workspace, with a Windows PowerShell shell running in a pane.
2. In that shell run: `wmux set-status "hello"` → error above, exit 1.
3. `wmux set-status "hello" --pane <that pane's own ptyId>` → same error, exit 1.
4. For contrast, `wmux send "echo hi" --pane <same ptyId> --submit` and `wmux read-screen --pane <same ptyId>` work from the same shell — so caller resolution works for `input.*`, just not `meta.*`.

### Expected behavior

`meta.setStatus` / `meta.setProgress` resolve the caller the same way `input.send` does (verified PID-map walk), or accept an explicit `--workspace <id>` / `--pane <ptyId>` target.

### wmux version

3.38.6

### How did you install wmux?

Setup.exe (direct download)

Windows PowerShell 5.1

### Windows version

Windows 11 Pro (build 26200)

None / not relevant

```shell

```

### Confirmation

openwong2kim:
Fixes #800. Addresses the confirmed half of #799 (details below — #799 stays open for the other half).

Two independent CLI-facing defects from the same scripting session, one PR per the owner's call.

## 1. `set-status` / `set-progress` always failed (#800)

`meta.setStatus` / `meta.setProgress` resolve the caller's workspace
server-side from `senderPtyId` and refuse a caller they cannot place (the U8
scoping change — a workspace-less metadata write lands on whichever workspace
the human happens to be looking at). `src/cli/commands/system.ts` sent only
`{ text }` / `{ value }`, so there was nothing to resolve. `wmux channel`
already had the resolution ladder; `system.ts` was never wired to it, and it
did not parse `--pane` at all.

Both commands now walk the same ladder `wmux channel` uses:

1. `--pane <ptyId>` — name the pane whose workspace to write (also stripped
   from the payload, so a leading `--pane` no longer gets read as the status
   text);
2. the verified PID-map walk (`resolveSelfContext`);
3. the `WMUX_PTY_ID` pane env, which survives a walk miss from a descendant
   process and a down main pipe.

With none of them resolvable the command exits 1 **before** the RPC with an
actionable message instead of surfacing the daemon's internal resolver text.
Server-side scoping is untouched: the CLI still never claims a workspace, it
only supplies the pane the daemon resolves from.

## 2. `workspace.close` handed out false receipts (#799, second symptom)

The handler returned `{ ok: true }` unconditionally, but `removeWorkspace` is a
silent no-op in two cases: an id that matches nothing, and the store's
last-workspace guard (wmux always keeps one workspace open). So
`wmux close-workspace <id>` printed `Closed workspace: ws-…` for a workspace
that was still open.

That is the reporter's "confirmed-closed workspace reappeared" — it did not
reappear. It was the last one, the removal was refused, and only the CLI's
receipt lied. Now:

- unknown id → `{ error }`;
- last-workspace refusal → `{ error }` explaining why, instead of the previous
  behaviour where the guard only skipped the PTY dispose and still reported
  success;
- the `{ ok: true }` receipt is issued only after re-reading the store to
  confirm the workspace is actually gone.

The post-check is an assertion, **not** a race fix — panel review correctly
pointed out that nothing awaits between the guards and the mutation, so two
concurrent closes cannot interleave and the check cannot fail today. It is kept
because the guards duplicate conditions that live in `removeWorkspace`, and the
entire bug here is that those two drifted apart without anything noticing.

`getResultError()` in the CLI already exits 1 on a payload-level `{error}`
(the same false-receipt class fixed for `surface.close`), so no CLI change is
needed for this half.

### What this does not fix

The #799 first symptom — a pre-existing workspace vanishing from
`list-workspaces` without ever being named in a `close-workspace` call — is not
explained by this path. `removeWorkspace` only ever removes the id it is given.
Daemon/main logs are requested on the issue (2026-08-04 00:50–01:15 +02:00);
#799 stays open for that half. The two symptoms are consistent in order:
something removed `Workspace 1` first, which left the scratch workspace as the
last one, which is what made its close get refused and produce the false
receipt.

## Verification

- New `src/cli/commands/__tests__/system.test.ts` — 7 tests: walk hit, `--pane`
  override (and that it skips the walk), leading `--pane` not read as payload,
  env fallback, throwing walk, fail-closed exit before any RPC.
- New `src/renderer/hooks/__tests__/useRpcBridge.workspaceClose.test.ts` — 5
  source-structural tests (`handleRpcMethod` is not exported and pulls in the
  store/window, same constraint as the existing `useRpcBridge.browserClose` /
  `useRpcBridge.focus` tests). Confirmed they fail 3/5 against the pre-fix
  handler. Source is CRLF-normalized before matching so a Windows checkout
  cannot break the \n-anchored regexes (this was the one CI failure on the
  split-out predecessor PR).
- `npx vitest run src/cli src/renderer/hooks` — all green locally
  (269 + 556); `tsc --noEmit` on `tsconfig.json` + `tsconfig.cli.json`, eslint
  clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved `set-status` and `set-progress` pane targeting with optional `--pane` selection.
  - Added clearer errors for invalid targets or commands run outside recognized panes.
  - Improved handling of multi-word status messages and progress values.
  - `close-workspace` now reports clear errors for unknown workspaces and attempts to close the last remaining workspace.
  - Success is reported only after workspace closure is confirmed.

- **Documentation**
  - Updated CLI help to describe pane targeting options.

### DIFF
--- a/changelog.d/801.md
+++ b/changelog.d/801.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **`wmux set-status` and `wmux set-progress` work again.** Both commands failed
+  every time with an internal message about an unresolvable calling pane, so
+  there was no way for automation to publish a run's live status or progress
+  into the workspace chrome. They now identify the calling pane the same way
+  `wmux send` does, and `--pane <ptyId>` targets another pane's workspace.
+  Outside a wmux pane they exit with a message that says what to do instead of
+  the daemon's internal resolver text. (#801)
--- a/changelog.d/801.md
+++ b/changelog.d/801.md
@@ -7,3 +7,9 @@
   `wmux send` does, and `--pane <ptyId>` targets another pane's workspace.
   Outside a wmux pane they exit with a message that says what to do instead of
   the daemon's internal resolver text. (#801)
+- **`wmux close-workspace` no longer reports closing a workspace it did not
+  close.** Closing an unknown workspace, or the last remaining one (wmux always
+  keeps one open), printed `Closed workspace: ws-…` and exited 0 while the
+  workspace stayed open — so a cleanup script had no way to tell a real close
+  from a refused one. Both cases now fail with an explanation, and the success
+  receipt is only issued after the workspace is confirmed gone. (#801)
--- a/src/cli/commands/__tests__/system.test.ts
+++ b/src/cli/commands/__tests__/system.test.ts
@@ -111,6 +111,52 @@ describe('set-status / set-progress carry a senderPtyId (#800)', () => {
     });
   });
 
+  it('accepts the --pane=<v> form instead of publishing it as the status', async () => {
+    // The two-parser version posted the literal "--pane=pty-other" to the
+    // caller's own workspace and exited 0.
+    await handleSystem('set-status', ['--pane=pty-other', 'hello'], false);
+    expect(rpc).toHaveBeenCalledWith('meta.setStatus', {
+      text: 'hello',
+      senderPtyId: 'pty-other',
+    });
+  });
+
+  it('joins the whole payload instead of dropping every word after the first', async () => {
+    await handleSystem('set-status', ['Build', 'failed'], false);
+    expect(rpc).toHaveBeenCalledWith('meta.setStatus', {
+      text: 'Build failed',
+      senderPtyId: 'pty-self',
+    });
+  });
+
+  it('rejects --pane with no value rather than eating the payload', async () => {
+    await expect(handleSystem('set-status', ['--pane'], false)).rejects.toThrow(ExitCalled);
+    await expect(handleSystem('set-status', ['--pane', '-x', 'hi'], false)).rejects.toThrow(
+      ExitCalled,
+    );
+    expect(rpc).not.toHaveBeenCalled();
+  });
+
+  it('rejects an unknown flag rather than publishing it as text', async () => {
+    await expect(handleSystem('set-status', ['--panne', 'pty-x'], false)).rejects.toThrow(
+      ExitCalled,
+    );
+    expect(rpc).not.toHaveBeenCalled();
+  });
+
+  it('passes a dash-leading payload through after --', async () => {
+    await handleSystem('set-status', ['--', '--not-a-flag'], false);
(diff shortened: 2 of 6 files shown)


openwong2kim:

This fixes #800: `wmux set-status` / `wmux set-progress` failed every time with the daemon's internal "cannot resolve the calling pane's workspace" error, because `src/cli/commands/system.ts` never attached a `senderPtyId` to the `meta.setStatus` / `meta.setProgress` RPC — even though `input.*` commands (`wmux send`) already had the resolution ladder wired up. The diff closes that asymmetry by reusing the same ladder `wmux channel` uses (`--pane` override → verified PID-map walk → `WMUX_PTY_ID` env), strips `--pane` out of the payload so a leading flag is no longer swallowed as the status text, and fail-closes with an actionable message before any RPC when no identity resolves. New `system.test.ts` covers the walk hit, the `--pane` override, the env fallback, a throwing walk, and the pre-RPC exit; `vitest run src/cli` passes 264 and `tsc --noEmit` is clean. This matters because it restores the only write path automation has to the workspace status line / progress bar — P2, matching the underlying issue. Merging this also closes #800.

PR #802, opened 2026-08-05
Stop workspace.close acknowledging a refused removal

openwong2kim:
Addresses half of #799 — the confirmed half. See "What this does not fix" below.

## Cause

`workspace.close` in `useRpcBridge.ts` returned `{ ok: true }` unconditionally.
`removeWorkspace` is a silent no-op in two cases: an id that matches nothing,
and the store's last-workspace guard (wmux always keeps one workspace open). So
`wmux close-workspace <id>` printed `Closed workspace: ws-…` for a workspace
that was still open.

That is the reporter's second symptom. The workspace the CLI confirmed closing
did not reappear — it was the last one, the removal was refused, and only the
CLI's receipt lied. A scripted cleanup then had no way to tell the difference.

Same false-receipt class `getResultError()` (`src/cli/utils.ts`) was introduced
for on `surface.close`, so the CLI already exits 1 on a payload-level `{error}`
— returning one is all it takes for the command to report honestly.

## Change

- unknown id → `{ error }` instead of a receipt;
- last-workspace refusal → `{ error }` explaining why, instead of the previous
  behaviour where the guard only skipped the PTY dispose and still reported
  success;
- the `{ ok: true }` receipt is issued only after re-reading the store to
  confirm the workspace is actually gone. The post-check is what makes
  concurrent closes honest as well: several `close-workspace` calls firing at
  once can each pass the entry-time length guard and still have the slice
  refuse the last one.

PTY disposal ordering is unchanged.

## What this does not fix

The report's first symptom — a pre-existing workspace vanishing from
`list-workspaces` without ever being named in a `close-workspace` call, with 4
of its 6 agent processes dead — is not explained by this path. `removeWorkspace`
only ever removes the id it is given, and nothing else in the RPC touches
another workspace. Chasing it needs the daemon/main logs the reporter offered
(2026-08-04 00:50–01:15 +02:00); I have asked for them on the issue, so #799
stays open.

Worth noting the two symptoms are consistent in order: something removed
`Workspace 1` first, which left the scratch workspace as the last one, which is
what made its close get refused and produce the false receipt.

## Verification

- New `src/renderer/hooks/__tests__/useRpcBridge.workspaceClose.test.ts` — 5
  source-structural tests (`handleRpcMethod` is not exported and pulls in the
  store/window, same constraint as the existing `useRpcBridge.browserClose` /
  `useRpcBridge.focus` tests). Confirmed they fail 3/5 against the pre-fix
  handler.
- `npx vitest run src/renderer/hooks` — 556 passed.
- `tsc --noEmit` clean, eslint no new warnings.

### DIFF
--- a/changelog.d/802.md
+++ b/changelog.d/802.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **`wmux close-workspace` no longer reports closing a workspace it did not
+  close.** Closing an unknown workspace, or the last remaining one (wmux always
+  keeps one open), printed `Closed workspace: ws-…` and exited 0 while the
+  workspace stayed open — so a cleanup script had no way to tell a real close
+  from a refused one. Both cases now fail with an explanation, and the success
+  receipt is only issued after the workspace is confirmed gone. (#802)
--- a/src/renderer/hooks/__tests__/useRpcBridge.workspaceClose.test.ts
+++ b/src/renderer/hooks/__tests__/useRpcBridge.workspaceClose.test.ts
@@ -1,0 +1,83 @@
+import { describe, it, expect } from 'vitest';
+import * as fs from 'node:fs';
+import * as path from 'node:path';
+
+/**
+ * `workspace.close` must not hand out a false receipt (issue #799).
+ *
+ * The handler returned `{ok:true}` unconditionally, but `removeWorkspace` is a
+ * silent no-op in two cases: an id that matches nothing, and the store's
+ * last-workspace guard (wmux always keeps one workspace open). A scripted
+ * cleanup therefore saw `Closed workspace: ws-…` for a workspace that was still
+ * open — and the reporter's "confirmed-closed workspace reappeared and is now
+ * the only workspace" was never a resurrection: it was the last one, the removal
+ * was refused, and only the CLI lied about it.
+ *
+ * Same false-receipt class `getResultError()` (cli/utils.ts) was introduced for
+ * on surface.close — the CLI already exits 1 on a payload-level `{error}`, so
+ * returning one is all it takes for `close-workspace` to report honestly.
+ *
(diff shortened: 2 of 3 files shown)


openwong2kim:

This addresses the second symptom of #799 — `wmux close-workspace` printing `Closed workspace: ws-…` for a workspace it never actually removed. The handler in `useRpcBridge.ts` returned `{ ok: true }` unconditionally, while `removeWorkspace` is a silent no-op for both an unknown id and the last-workspace guard (wmux always keeps one open), so a scripted cleanup had no way to distinguish a real close from a refused one. The diff makes both no-op branches return an `{ error }` (the CLI already exits 1 on a payload-level error via `getResultError`), preserves the existing PTY-dispose ordering, and — importantly — re-reads the store after the mutation to confirm the id is gone, which is what keeps the receipt honest under concurrent `close-workspace` calls racing past the entry-time length check. New `useRpcBridge.workspaceClose.test.ts` is a source-structural guard (the handler is not exported) asserting the error branches, the post-check, the dispose ordering, and a single `ok` receipt; `vitest run src/renderer/hooks` passes 556. Note this does **not** close #799 — the report's first symptom (a pre-existing workspace vanishing without any `close-workspace` call, with 4 of its 6 agent processes dead) is not on this path; `removeWorkspace` only ever touches the id it is given, and the owner has asked for the daemon/main logs (2026-08-04 00:50–01:15 +02:00) to chase that separately. P2, tracking the issue until the vanishing half is explained.

openwong2kim:
Consolidated into #801 per the owner's call — same two fixes, one PR. The Windows CI failure here (CRLF breaking the \n-anchored structural-test regexes) is fixed in the consolidated branch by normalizing line endings before matching.

PR #803, opened 2026-08-05
docs(relay): correct which builds the sandbox worker serves

openwong2kim:
The sandbox worker's comment in `relay/wrangler.toml` claimed it was "for TestFlight/debug builds". Only the second half is right.

A TestFlight build is signed with `aps-environment: production` and its device tokens resolve only on Apple's production host. Pointing a TestFlight install at the sandbox worker therefore leaves `APNS_ENV` as a deployment default that is wrong for it. The per-request `apnsEnvironment` in `src/index.ts` still routes such a push correctly when the app reports its own stage, but that is a fallback covering for a misconfiguration rather than the intended shape.

Comment only — no behavior change.

## How this surfaced

While switching this Mac's relay LaunchAgent to the production worker ahead of the first TestFlight build, this comment is what made the sandbox worker look like the correct target for it.

### DIFF
--- a/relay/wrangler.toml
+++ b/relay/wrangler.toml
@@ -23,7 +23,16 @@ enabled = true
 #                        serve rather than becoming an open proxy to Apple.
 
 # `wrangler deploy --env sandbox` publishes a second worker that talks to
-# Apple's sandbox host, for TestFlight/debug builds. Its secrets are separate.
+# Apple's sandbox host, for builds installed straight from Xcode — the ones
+# whose `aps-environment` entitlement reads `development`.
+#
+# TestFlight is NOT one of them. A TestFlight build is signed `production` and
+# its device tokens only resolve on Apple's production host, so pointing a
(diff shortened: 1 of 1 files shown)


openwong2kim:

This is a docs-only change to `relay/wrangler.toml` (+10/-1, one file, no behavior change) that corrects the comment above the `[env.sandbox]` worker definition. The old comment claimed the sandbox worker was "for TestFlight/debug builds" — the new comment clarifies it serves builds installed straight from Xcode (whose `aps-environment` entitlement reads `development`), and explicitly calls out that TestFlight is *not* one of them because TestFlight builds are signed `production` and their tokens only resolve on Apple's production host. It is accurate and worth landing: the per-request `apnsEnvironment` fallback in `src/index.ts` currently masks the misconfiguration, so this prevents anyone else from pointing a TestFlight install at the sandbox worker based on the comment alone.

Marking **P3** — comment-only with no functional impact, low risk, safe to merge once the remaining CI check (windows-latest) finishes.

PR #806, opened 2026-08-05
feat(terminal): floating scroll-to-bottom button when scrolled up

rayss868:
## Summary

When a terminal's scrollback runs far ahead of the current viewport, there
was no one-click way to get back to the latest output — you had to mouse it
all the way down or use the scrollbar. This PR adds a **scroll-to-bottom
button**: a small chip floating at the bottom-right of the terminal that
appears only while you are scrolled up, and jumps to the latest output on
click.

Per DESIGN.md's "no dead gauges" rule, it is invisible at rest — the
affordance exists exactly when scrollback has run ahead, and disappears the
moment you are back at the bottom.

## Design

- **Auto-hide, not always-on.** The button is rendered only while the user is
  scrolled up. This avoids any permanent chrome over the terminal hero.
- **Bare arrow icon.** A monochrome chevron; per the 5±2 amber-points grammar
  it does not show a distance count. Warm accent appears on hover only.
- **Scroll semantics reused from the existing code.** The "scrolled up"
  state mirrors the resize-preservation logic in `useTerminal`:
  `viewportY < baseY` means scrolled up; `viewportY === baseY` means at the
  bottom (following output). So the button's visibility judgement is provably
  consistent with the rest of the terminal.
- **`scrollToBottom()`.** Clicking calls xterm's built-in `scrollToBottom()`.

## Changes

- **New `ScrollToBottomButton.tsx`** — floating button component. Subscribes
  to `onScroll`, `onWriteParsed` (new output / baseY growth), and `onResize`,
  refreshing the scrolled-up boolean each time; React bails out on unchanged
  values, so an output flood does not cause re-render churn. Disposables are
  cleaned up on unmount. Uses the app's established gpui raised-button recipe
  (`bg-surface` fill + hairline + inset highlight), `FOCUS_RING`, and the
  `useT` i18n hook.
- **`Terminal.tsx`** — renders the button as an overlay next to
  `BookmarkIndicator`, passing `terminalRef.current`.
- **`en.ts`** — adds the `terminal.scrollToBottom` i18n key (other locales
  fall back to English).

## Test plan

- [ ] `npm start`, open a pane, generate a long scrollback (e.g. `seq 1 5000`)
- [ ] Scroll up → button appears bottom-right; scroll back to bottom → it
      disappears
- [ ] Click the button while scrolled up → jumps to the latest output
- [ ] While following output at the bottom, new output arrives → button stays
      hidden and terminal keeps following
- [ ] Resize the pane while scrolled up → button state stays correct

## Summary by CodeRabbit

- **New Features**
  - Added a terminal “Scroll to bottom” button that appears when viewing earlier output.
  - Clicking the button returns the terminal viewport to the latest output.
  - Added accessible English labeling for the new control.

### DIFF
--- a/changelog.d/806.md
+++ b/changelog.d/806.md
@@ -1,0 +1,7 @@
+### Added
+
+- **Scroll-to-bottom button on the terminal.** When a terminal's scrollback
+  runs far ahead of the viewport, a small button now appears at the
+  bottom-right; one click returns to the latest output. It shows only while
+  you are scrolled up and hides again the moment you are back at the bottom.
+  (#806)
--- a/src/renderer/components/Terminal/ScrollToBottomButton.tsx
+++ b/src/renderer/components/Terminal/ScrollToBottomButton.tsx
@@ -1,0 +1,71 @@
+import { useCallback, useEffect, useState } from 'react';
+import type { Terminal } from '@xterm/xterm';
+import { useT } from '../../hooks/useT';
+import { FOCUS_RING } from '../focusRing';
+
+/**
+ * Floating "scroll to bottom" button. Appears at the bottom-right of the
+ * terminal ONLY while the user is scrolled up (not following output) and
+ * disappears the moment they are back at the bottom. Clicking jumps to the
+ * latest output.
+ *
+ * Per DESIGN.md's "no dead gauges" rule: never a permanent overlay over the
+ * terminal hero — the affordance exists exactly when scrollback has run far
+ * ahead and is otherwise invisible.
+ */
+export default function ScrollToBottomButton({ terminal }: { terminal: Terminal | null }) {
+  const t = useT();
+  const [scrolledUp, setScrolledUp] = useState(false);
+
+  const update = useCallback(() => {
+    if (!terminal) {
+      setScrolledUp(false);
+      return;
+    }
+    // xterm: at the bottom means viewportY equals baseY (following output);
+    // scrolled up means viewportY < baseY. Same relation the resize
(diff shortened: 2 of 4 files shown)


openwong2kim:

This PR adds a floating "scroll to bottom" button to the terminal that appears only while a pane is scrolled up in its scrollback and vanishes the instant the viewport returns to the latest output. The implementation is a single new component, `ScrollToBottomButton.tsx`, mounted as an overlay from `Terminal.tsx` alongside the existing `BookmarkIndicator`, and it correctly reuses xterm's own bottom-of-buffer relation (`viewportY < baseY`) rather than reinventing one — the same check the resize-preservation logic in `useTerminal` already relies on, so the affordance stays provably consistent with following-output behaviour.

The scope is small and clean: one new self-contained React component with properly disposed event subscriptions (`onScroll`/`onWriteParsed`/`onResize`), a single i18n key in `en.ts` (with the documented English fallback for other locales), a changelog fragment, and three lines wired into `Terminal.tsx`. There are no new dependencies, no changes to PTY/daemon/IPC code, and the per-event re-render concern is addressed by setting an identical boolean (React bails out), so an output flood cannot cause churn. The auto-hide design also honours DESIGN.md's "no dead gauges" rule, which is the right call for a terminal hero surface.

I am tagging this **P3** — a polished, genuinely useful UX nicety that is correctly scoped and low-risk. Nice detail: the visibility judgement mirrors the existing follow-output logic rather than introducing a second, drift-prone definition of "at the bottom."

PR #807, opened 2026-08-05
Disarm replayed mouse modes on app-only restart

openwong2kim:
## The bug

Quit wmux, relaunch it, and the shell prompt has junk already typed into it:

```
wong2kim@wong2kimui-MacBookAir wmux % 35;9;12M
```

That string is the tail of an SGR mouse report, `ESC[<35;9;12M`. zsh's ZLE
consumes the undefined `ESC[<` prefix and self-inserts the rest. `35` is
`32 (motion) + 3 (no button)`; `9;12` is the cell the pointer crossed.

## Why it happens

1. A TUI agent running in the pane arms mouse tracking (`?1003h`, `?1006h`).
2. Those bytes sit in the session's ring buffer.
3. On reattach the buffer is replayed verbatim into a fresh xterm, re-executing
   the arming sequences. (The snapshot path is no better here: the headless
   terminal's computed mode state is "tracking on", so `buildModesTail` faithfully
   re-emits it.)
4. The pane now holds a plain shell with mouse reporting enabled that it never
   negotiated. The next pointer move writes a report into its stdin.

`STALE_REPLAY_INPUT_MODE_RESETS` was built for exactly this (2026-07-02 RCA),
but the gate was too narrow:

```ts
if (sessions.find((s) => s.id === ptyId)?.resumeAgent) { ... }
```

`resumeAgent` is populated in `daemon/index.ts` only for ids in `recoveredIds`,
i.e. sessions the daemon recovered during its **own** boot. Quitting the app
while the daemon keeps running produces a replay with no recovery, so the gate
was false and the disarm never ran.

## The fix

Extract the gate into `shouldResetStaleReplayModes()` and add
`commandRunning === false` alongside `resumeAgent`.

`commandRunning` comes from the pane's OSC 133 prompt log and is already
surfaced unconditionally by `pty.list`. `false` means the shell is sitting at
its prompt with no foreground command — so no live process can legitimately own
mouse, focus, or bracketed-paste reporting, and anything the replay armed is
stale by construction. It also covers non-agent TUIs (vim, htop) that
`resumeAgent` never described.

Both declining states are deliberate:

| `commandRunning` | meaning | action |
|---|---|---|
| `true` | a live TUI owns the PTY | skip — never clobber its modes |
| `undefined` | no shell integration, no evidence | skip |
| `false` | shell at prompt | reset |

No daemon changes. Renderer-only, terminal-side only — the PTY never sees these
bytes.

## Known gap

A pane with shell integration off **and** no daemon-boot recovery still has no
signal, so it stays unfixed. Widening further would mean guessing, and a wrong
guess disables a live agent's mouse support — a worse failure than the junk
string.

## Tests

`shouldResetStaleReplayModes` is unit-tested across all four daemon states,
including the app-restart case this PR fixes. The source-level wiring lock now
rejects an inline `?.resumeAgent` check; it fails against the pre-fix source and
passes after.

```
Test Files  59 passed (59)
     Tests  771 passed (771)
```

`tsc --noEmit` clean. ESLint clean on all changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal recovery after stale session replays by clearing outdated mouse-tracking modes when a pane is known to be at a shell prompt.
  * Preserved mouse tracking for running terminal applications and bracketed paste during app-only restarts.
  * Applied full input-mode cleanup when the shell has exited.
  * Panes without shell integration still require restarting the daemon for recovery.


openwong2kim:

This PR fixes a real annoyance where quitting and relaunching wmux (while the daemon keeps running) left junk like `35;9;12M` typed into the shell prompt — the tail of an SGR mouse report from a previously-running TUI agent whose `?1003h`/`?1006h` arming survived in the replay buffer. The existing `STALE_REPLAY_INPUT_MODE_RESETS` disarm was gated too narrowly on `resumeAgent`, a field the daemon only populates during its own boot recovery, so an app-only restart with a live daemon slipped through entirely. The fix extracts a `shouldResetStaleReplayModes()` predicate that also fires on `commandRunning === false` (OSC 133 reporting the shell at its prompt), which is the precise condition under which no foreground process can legitimately own mouse/focus/paste modes — and it correctly covers non-agent TUIs (vim, htop) that `resumeAgent` never described, while leaving a genuinely running TUI's modes untouched.

The change is renderer-only and terminal-side only (the PTY never sees these bytes), which keeps the blast radius small: four files, +97/-14, touching `staleReplayModeReset.ts`, the `useTerminal` wiring site, its unit tests, and a changelog fragment. The four-case gate test plus the source-level wiring lock that now rejects an inline `?.resumeAgent` check give good regression protection, and the reported 771/771 suite pass with clean `tsc` and lint is exactly the bar this kind of surgical fix should clear.

I am labelling this **P2** — it resolves a visible, reproducible defect in the attach/resume path that users hit on every app restart, the fix is low-risk and well-scoped, and it strengthens an invariant (never leak a dead process's input modes into a fresh shell) rather than adding surface area. Nothing blocks a merge from my side; worth landing in the next patch release alongside the broader redraw/replay work tracked in #678.

openwong2kim:
## Panel review round (Claude / Codex / GLM, independent contexts)

Three findings confirmed and fixed in `437f035`; one was checked and rejected.

### Fixed

**Bracketed paste desync — the widened gate promoted a paste-corruption path to
the everyday case.** wmux decides whether to wrap a paste in `ESC[200~` by
reading xterm's own `modes.bracketedPasteMode` (`clipboardChunk.ts:159`,
`Terminal.tsx:289`, `useTerminal.ts:1647`). A shell at its prompt has `?2004`
armed and expects wrapped pastes. Writing `?2004l` terminal-side desyncs the
two, and a multi-line paste then lands as N commands the shell executes
immediately. Safe while the gate only fired on known-dead panes; not safe on
every attach.

Fixed by splitting the constant: `STALE_REPLAY_MOUSE_MODE_RESETS` (mouse
reporting only) for alive-shell panes, `STALE_REPLAY_INPUT_MODE_RESETS`
(unchanged, includes `?1004`/`?2004`) reserved for known-dead panes.

**Precedence inversion.** `resumeAgent` outranked `commandRunning === true`, so
a recovered pane where the user had since started a fresh TUI got that TUI's
mouse modes cleared. `commandRunning === true` now declines outright.
`shouldResetStaleReplayModes` became `staleReplayResetLevel` returning
`'full' | 'mouse' | 'none'`.

**Changelog overstated the scope** — it read as though panes without shell
integration were covered. They are not; scoped accordingly.

### Rejected after verification

One reviewer flagged nested interactive shells emitting OSC 133 on the same PTY,
which would pin `commandRunning` to `false` while an outer TUI stayed alive. The
mechanism does not exist here. zsh restores `ZDOTDIR` to the user value at
`shell-integration.ts:281-286` *before* the hooks are installed, so a nested zsh
never gets them; bash assigns `PROMPT_COMMAND`/`PS0` without `export` and the
functions without `export -f`, so a nested bash inherits nothing.

The narrower form the other two reviewers raised at the same location — a remote
shell with its own OSC 133 integration emitting markers over the same PTY — is
real, and is covered by the deferred item below.

### Deferred (not in this PR)

- **TOCTOU:** the gate is evaluated after a `pty.list` round-trip, so a TUI
  started during that window can have its modes cleared. Pre-existing in kind;
  this change raises its frequency by moving the gate onto the everyday path.
  Wants a sequence-counter guard.
- **Process truth:** crossing the gate with `agentProcessAlive` would close both
  the remote-shell false positive and the no-shell-integration miss. Needs the
  daemon to surface that field unconditionally (today it rides along only with a
  resume binding, `daemon/index.ts:2213`).

### Gates

`tsc --noEmit` clean. 775 tests pass across 59 files (4 new). ESLint clean on
changed files.

PR #809, opened 2026-08-05
Record what the explicit surfaceId trace actually established

openwong2kim:
The top P0 in `TODOS.md` was an open question rather than a defect: #679 made browser page selection fail closed when the caller's workspace identity cannot be resolved, but the trace of the *other* route — a caller that names a `surfaceId` explicitly — was never finished. Left as-is, the next reader starts that trace over.

It is finished now. The tool layer refuses: through `getPageForScope` there is no route to another workspace's surface, and what closed it was the #695 work that landed after #679, not #679 itself.

- **Entry** — every browser tool goes through `getPageForScope(scope)`, and `requireBrowserTargetScope()` refuses to build a scope without a workspace. `getPage()` is private with no other caller. This is a convention rather than a structural guarantee, though: the explicit-surface branch builds a context regardless of workspace, and `selectRegisteredTarget` short-circuits when none is named. It holds because every caller goes through the scoped entry point, which the entry now states as the load-bearing fact it is.
- **Selection** — `selectRegisteredTarget()` matches the named surface against the caller's workspace. The two branches differ in who does the checking: on a main honoring `targetsScoped` the server-side filter in `browser.cdp.info` is the only check, and the client-side comparison exists solely on the legacy branch.
- **Fallback and ownership** — the RPC fallback that #679's changelog flagged as untouched is covered: `sendScopedBrowserRpc` injects the workspace and overrides any surface smuggled through params, every handler passes `scopeOf(params)`, and `ownedBy()` guards the explicit-surface branch of `getTarget()`.

Regression coverage already exists at both ends — `PlaywrightEngine.test.ts` rejects a foreign explicit target from a legacy main, and `WebviewCdpManager.test.ts` asserts `getTarget('b-surface', 'ws-A')` is null. The four related suites pass (83 tests).

The heading is deliberately "툴 계층은 닫혀 있다" and not "구멍 없음", because two paths outside that layer stay open and an entry that reads as a clean bill of health would get cited as one. A caller that omits `workspaceId` entirely is unscoped, and that set is wider than the first-party CLI the comments assume: `browser.evaluate` and `browser.read` are declarable capabilities and `PermissionEnforcer` has no workspace dimension, so an approved third-party plugin reaches it too. Separately, `browser.cdp.info` returns `cdpPort` even on a scoped response, so `browser.read` alone suffices to drive every page over CDP directly. Both are recorded in the entry and tracked in #810.

Docs only — no code changes, no changelog fragment (nothing user-facing shipped).

### DIFF
--- a/TODOS.md
+++ b/TODOS.md
@@ -42,9 +42,9 @@
 
 ## ✅ #679 명시적 `surfaceId` 경로 검증 — 구멍 없음 (2026-08-05)
 - **원래 질문:** PR #679는 *신원 해석 실패* 시 브라우저 페이지 선택이 fail-open 하던 걸 막았다. 호출자가 `surfaceId`를 **명시적으로 넘기는** 경로도 소유권을 검증하는가?
-- **답:** 검증한다. 다만 그걸 닫은 건 #679가 아니라 그 뒤의 #695 작업이다. 3층 전부 확인:
+- **답:** 검증한다. 다만 그걸 닫은 건 #679가 아니라 그 뒤의 #695 작업이다. 세 층을 따라간 결과:
   1. **진입** — 브라우저 툴 전 호출 지점이 `getPageForScope(scope)` 하나만 쓴다. `scope`는 `requireBrowserTargetScope()` 산출물이고 workspaceId가 비면 fail-closed(`browserScope.ts`). `getPage()`는 private, 다른 진입점 없음 → 명시 surfaceId도 workspaceId 없이는 도달 불가.
-  2. **엔진 선택** — `selectRegisteredTarget()`이 명시 surfaceId를 대조한다: 최신 main은 `targetsScoped`로 서버가 필터, 레거시 main은 클라이언트가 `target.workspaceId !== workspaceId`를 `WORKSPACE_SCOPE_UNRESOLVED`로 거부(태그 없는 타깃도 거부).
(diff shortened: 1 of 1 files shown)


openwong2kim:

This is a docs-only change (+8/-5 on `TODOS.md`) that resolves the open P0 question left after #679: whether a caller passing an explicit `surfaceId` to the browser page-selection path is subject to the same ownership checks. The PR traces the route through three layers — entry (`getPageForScope`/`requireBrowserTargetScope`, with `getPage` private and unreachable without a workspace), engine selection (`selectRegisteredTarget` filtering server-side via `targetsScoped` for modern mains and rejecting foreign/untagged targets as `WORKSPACE_SCOPE_UNRESOLVED` for legacy mains), and the RPC fallback plus ownership guard (`sendScopedBrowserRpc` injecting the workspace, every handler passing `scopeOf(params)`, and `WebviewCdpManager.getTarget()`'s `ownedBy()` covering the explicit-surface branch) — and correctly attributes the actual closure to the later #695 work rather than #679 itself.

The reasoning is sound, the regression coverage already cited (`PlaywrightEngine.test.ts` and `WebviewCdpManager.test.ts`) matches the claims, and the one remaining door (a caller sending no `workspaceId` at all) is correctly framed as the existing same-user trust ceiling from #113 rather than a new defect. No code, changelog, or user-facing behavior changes, so the bar here is accuracy and completeness of the trace — both of which hold up. Safe to merge; labeling P3 as documentation-only.

PR #812, opened 2026-08-06
feat(remote): attach and mirror workspaces from another machine's wmux web server

openwong2kim:
## Summary

Attach workspaces from another machine's `wmux web` server into the local app and mirror their panes (view + type) — no separate remote-desktop program needed.

- **Daemon**: new `GET /api/workspaces` — live workspace list derived from session env (same auth tier as `/api/sessions`, brain-pty filtered, addressable ids).
- **Main**: `RemoteHostsStore` (bearer tokens persisted via `secureWriteTokenFile`, owner-only perms), `RemoteHostClient` (Bearer-header SSE with comment/unknown-event filtering, jittered capped reconnect, serialized input queue), `remote.handler` (host-add `/api/config` probe with distinct failure messages, sender-targeted pushes, idempotent attach, exit/reload/quit cleanup).
- **Renderer**: separate `remoteWorkspacesSlice` (remote entries are structurally unreachable from `removeWorkspace`/pane-tree code), "Attach remote workspace…" entry in the PresetPicker, attach modal (masked URL input, per-host remove), sidebar section with host badges, mirror grid with hidden-but-alive mounting (max 6 concurrent panes, read-only banner, disconnected/exited overlays).
- **Docs**: `docs/how-to/remote-workspaces.md` + README FAQ line + changelog fragment.

Deliberately out of scope: remote workspace create/rename/close (attach/detach only — detach never destroys anything remote), resize negotiation (mirror letterboxes at remote geometry), attach persistence across app restarts.

## Setup

On the remote machine: `wmux web --tailscale --allow-input`, then paste the printed URL into the attach modal locally.

## Test plan

- `tsc --noEmit` clean; full suite green: test:parallel 10266 passed / test:runtime 138 passed, 0 failures.
- New coverage: daemon route grouping/401/brain-pty exclusion; store roundtrip + owner-only file perms; SSE parse (comment/unknown frames, split chunks), reconnect cap, write coalescing/ordering; handler probe mappings, idempotent attach, exit/reload cleanup, wrapHandler URL-token redaction; slice invariants incl. every `activeWorkspaceId` assignment clearing remote selection; modal + mirror component tests.
- Manual smoke suggested before release: attach against a second instance (`WMUX_DATA_SUFFIX=-demo` + `wmux web --allow-input`).

## Summary by CodeRabbit

* **New Features**
  * Attach remote `wmux web` workspaces to the desktop sidebar.
  * View live mirrored panes, scrollback, shell details, and working directories.
  * Optionally forward keyboard input, with read-only mode when unavailable.
  * Manage remote hosts, select workspaces, and detach connections.
  * Recover automatically from temporary connection interruptions.

* **Documentation**
  * Added setup, usage, limitations, troubleshooting, and security guidance.
  * Updated the FAQ and documentation index.

* **Bug Fixes**
  * Sensitive access tokens are redacted from diagnostic logging.


openwong2kim:

This PR introduces remote workspace attach — the ability to register another machine's `wmux web` server in the local desktop app and mirror its live panes in the sidebar (view plus, when the remote allows it, keystroke forwarding). It is a substantial feature spanning 34 files / ~3700 additions across four layers: a new daemon `GET /api/workspaces` route, a main-process surface (`RemoteHostsStore` for token persistence, `RemoteHostClient` SSE bridge with capped reconnect, `remote.handler` IPC wiring with per-sender cleanup and idempotent attach), a renderer slice + UI (attach modal, sidebar section, mirror grid with read-only/disconnected overlays), and docs (`docs/how-to/remote-workspaces.md` + README FAQ + changelog fragment).

The security posture is the part worth calling out specifically. The pasted `wmux web` URL carries a bearer token that grants scrollback read and, under `--allow-input`, command injection on the remote machine — so the credential handling is what this PR lives or dies on, and it is handled carefully. Tokens are persisted through the same `secureWriteTokenFile` / owner-only ACL path used for the daemon auth token and lanlink peers, and re-hardened on every load. The diff includes a deliberate two-layer redaction for the token: `wrapHandler` now strips a `token=` query param from a bare-string URL argument before it can reach `args_summary` logs (defense in depth, with its own tests), and `hostsAdd` wraps `store.add()` so a fail-closed write error never escapes with the raw token URL still in scope. The `--tls-cert`/`--tls-key` path is explicitly unsupported for attach and the docs say why — Node `fetch` rejects self-signed certs, so attach is steered onto `--tailscale`.

On the test and scoping side this lands clean: `tsc --noEmit` passes, the full parallel/runtime suites are reported green (10266 / 138), and the new coverage is substantial and targeted — daemon route grouping/401/brain-pty exclusion, store roundtrip plus owner-only file perms, SSE parse with comment and unknown-frame and split-chunk handling, reconnect cap, write coalescing/ordering, probe-failure discrimination, idempotent attach, reload/exit cleanup, and the slice invariant that every `activeWorkspaceId` assignment clears a stale remote selection. The structural decision to keep remote entries in a separate `remoteWorkspacesSlice` array so `removeWorkspace`/pane-tree/persistence code cannot reach them by construction is the right one — it is a type-level guarantee rather than a runtime check, which is exactly how you want a feature that touches the pane tree to be isolated. Out-of-scope items (remote create/rename/close, live resize negotiation, attach persistence across app restarts) are enumerated and reasonable for a first cut.

**Priority: P3.** This is a net-new optional capability (no existing functionality regresses if it ships later), the remote machine must opt in explicitly with `wmux web`, and the surface is additive rather than touching shared critical paths. It is well-built and can land at a normal cadence; the one thing worth eyeballing before merge is the manual smoke against a second `WMUX_DATA_SUFFIX=-demo` instance the test plan suggests, since the SSE reconnect and reload-cleanup paths are the kind of thing unit tests only approximate.

openwong2kim:
@codex review

PR #814, opened 2026-08-06
feat(ssh): drive remote-box agents over SSH (P0a)

openwong2kim:
## What

Run an agent pane over SSH to a remote box the operator already owns, by
declaring `kind: "ssh"` in `wmux.json`. The remote pane reuses the daemon's
existing data plane (scrollback, OSC 133 / agent detection, notifications,
loops, Fleet View, channels) unchanged — only the transport is new.

This is the **P0a slice** of closing the remote-execution gap (see the eng
review). P0b (supervised reconnect, phone approval delivery for remote)
follows after dogfood.

## Why

The hardest competitor gap to ignore (Orca federation/cloud, Conductor Cloud,
herdr/cmux remote) is "agents on a machine that isn't the one you're looking
at." The cheapest real answer that respects wmux's "your box, your keys, your
browser" posture is SSH to a box the operator controls — not wmux hosting
compute. It turns the already-shipped unattended runtime + approval gate into
one coherent product: a fleet on any machine you own, supervised, with
approvals that surface where you already work.

## How

- **`SessionProcess` interface** (the clean seam). `node-pty`'s `IPty` satisfies
  it structurally, so widening the stored type + the bridge parameter is the
  whole data-plane refactor. New transports implement it.
- **`SshChannelSession`** — an `ssh2` channel as a `SessionProcess`. Synthetic
  negative pid (no OS pid on the operator's box); liveness is event-driven via
  `onExit` (channel close/error), not pid-polled. `resize` → `setWindow`,
  `kill` → channel close + `client.end()`.
- **`createSession` branches on `kind: "ssh"`**; rejects supervision on ssh
  (reconnect is P0b). `DaemonSession.kind` persists the transport. SSH sessions
  skip the pid-based `ProcessMonitor` watch and are skipped by recovery replay.
- **`wmux.json` leaf** gains `kind: "ssh"` + `ssh { host, port, username,
  privateKeyPath, agent, remoteCommand }`, mutually exclusive with
  `command`/`cwd`/`url`/`role`/`restart`. Host charset is locked (no shell
  metacharacters); `privateKeyPath` must be absolute / `~` / drive so a
  checked-in wmux.json can't point a key at the project root.
- **`system.capabilities`** advertises `transports: ["local","ssh"]`.

## Security

- **Host-key TOFU** against `~/.wmux/ssh-known_hosts`: first use records the
  SHA256 fingerprint; a later mismatch fails closed. Without this, ssh2 accepts
  any host key and an attacker answering for the name collects the operator's
  key/agent/password and proxies every byte.
- **Key material never persists.** The config carries only a path; the key is
  read at connect and discarded. `~` is expanded to `$HOME`.
- **`onExit` late-subscriber fix.** The sync fail path (unreadable key /
  connect throw) can `fireExit` before the bridge attaches `onExit`; the exit is
  now stored and replayed to a late subscriber so a failed connect can never
  strand the session as a deathless zombie (SSH skips the pid reaper).

## Review

`/plan-eng-review` (complexity gate → P0a/P0b slice), then `/review-team`
(GLM-5.2 + Claude; Codex CLI skipped). Three P1s found and fixed before merge:
host-key MITM, `~` expansion, onExit zombie. Shared P2 (recovery gate) fixed.

## Verification

- `tsc` clean; `npm run build:daemon` succeeds (ssh2 + cpu-features
  externalized in the esbuild bundle, same pattern as node-pty).
- 3131 daemon + shared tests green; 73 SSH tests (11 schema, 7 SshChannelSession
  via mocked ssh2, plus the existing suite).
- **Live dogfood against a real ssh2 server: 9/9 GREEN** — handshake, banner,
  keystroke round-trip, resize window-change, TOFU recording, host-key mismatch
  rejection (no banner leak), bad key path clean exit, synthetic negative pid.
  (Transport layer. The full Electron app path — wmux.json leaf → renderer
  funnel → desktop pane — is type-checked and slice-tested but not yet
  live-run against a real remote box; that's the remaining owner dogfood.)

## Known limits (P0b)

- **No reconnect.** A dropped channel or a daemon restart ends the remote pane;
  recovery skips SSH sessions on purpose. Re-open from the layout.
- **Approvals are desktop-only** for remote agents (the approval pipeline is
  local-PTY-shaped; remote delivery is P0b).
- **No supervision / restart** on ssh leaves (rejected by the schema).
- **Remote OSC bytes are parsed with local privileges** (a malicious remote
  could spoof OSC 7 cwd / OSC 9 notifications). Acceptable for an
  operator-owned box; documented as a trust boundary.

## Commits

1. `9514ccb3` — SessionProcess seam, SshChannelSession, schema, capability
2. `1e34a9ef` — end-to-end plumbing (wmux.json leaf → remote pane)
3. `6e64453a` — review fixes (host-key TOFU, tilde, onExit zombie, recovery gate)
4. `d33fe8cb` — daemon build externalizes ssh2 + cpu-features

## Summary by CodeRabbit

- **New Features**
  - Added SSH-backed panes configured through `wmux.json`.
  - Supports remote shells or commands with key-file, password, or SSH-agent authentication.
  - Added SSH host-key verification and connection settings.
  - Project layouts can now seed SSH panes with validated hosts, ports, users, and commands.
  - Capabilities now advertise local and SSH pane transports.

- **Bug Fixes**
  - Failed SSH connections no longer leave orphaned panes.
  - Added terminal resizing and reliable handling of remote session exits.


openwong2kim:

This PR adds SSH-backed remote panes — the P0a slice of closing the remote-execution gap — by introducing a transport-agnostic `SessionProcess` interface that lets `ssh2` channels back the same pane surface as local `node-pty` children. It touches 20 files across the daemon (`SshChannelSession`, the factory branch in `DaemonSessionManager`, recovery/process-monitor skips), the shared schema (`normalizeSshLeaf` with host-charset locking, port-range checks, and a project-root-escape guard on `privateKeyPath`), and the renderer→preload→main plumbing needed to thread a `kind: 'ssh'` leaf end-to-end. The data plane (RingBuffer, OscParser, AgentDetector, bridge) is reused unchanged because only the transport differs.

The security posture is the standout: three P1s from the `/review-team` pass were fixed before opening — host-key TOFU against `~/.wmux/ssh-known_hosts` (without it ssh2 accepts any host key, making a checked-in `wmux.json` host an MITM collector), `~` expansion on `privateKeyPath`, and an `onExit` late-subscriber fix that prevents a failed connect from stranding the pane as a zombie since SSH skips the pid reaper. The known limits (no reconnect, desktop-only approvals, no supervision, remote OSC bytes parsed with local privileges) are documented as P0b and acceptable for the operator-owned-box model. Given that it closes the hardest competitor gap (agents on a machine that isn't the one you're looking at) while preserving the "your box, your keys" posture, I'm tagging this **P1**.

One thing to note before merge: `mergeable` is currently `CONFLICTING`, so a rebase onto `main` is needed. The remaining owner dogfood — running the full Electron app path (wmux.json leaf → renderer funnel → desktop pane) against a real remote box — is called out as not yet live-run, so that should gate the merge even after conflicts resolve.

openwong2kim:
Moving this to draft pending demand evidence, not because of anything wrong with
the implementation.

## Why

The "Why" section here argues from a competitor gap (Orca federation, Conductor
Cloud, herdr/cmux). That is a market observation, not evidence that this
project's users need it. Checked against actual usage: nobody has yet tried to
run an agent on a remote box and been blocked. The feature was built before the
need was felt.

Every bug fixed in the 3.39.0 cycle came out of actually using the thing
(#815 pairing feedback, #817 detach, #818 accelerator collision). This one has
not been through that.

## Not a duplicate of #815

Worth recording so nobody re-litigates it. #815 (remote workspace attach over
`wmux web`) mirrors panes owned by a remote wmux daemon. This runs the pane on
the LOCAL daemon over an ssh channel, so it needs only sshd on the far end and
the pane is a first-class session (agent detection, notifications, loops, Fleet
View). Different problems. Neither replaces the other.

## The real comparison is the status quo

Today you can open a pane and type `ssh box`. What this adds on top is the
`wmux.json` declaration and first-class session treatment. That is real, but it
is the increment that has to justify the cost — not the whole capability.

## Cost of merging as-is

The build cost is sunk; these are the ongoing ones:

- `ssh2` enters the dependency tree permanently, whether or not any config
  declares `kind: "ssh"` — install size, supply chain, CVE surface.
- Host-key TOFU is implemented here rather than delegated. Trust management for
  SSH is a sharp area, and review already found a MITM hole in this code once.
- Three capability holes ship with it: supervision refused, `ProcessMonitor`
  skipped, recovery replay skipped. Every future session-layer feature then has
  to answer "does this hold over ssh?"

Runtime risk is close to zero while unused — the branch is opt-in. Dependency
and maintenance cost is not.

## What unblocks this

Run the status quo for real: ssh to a box, run an agent through a whole task,
and write down what actually hurt.

- Nothing hurt → close this.
- Agent detection / notifications / loops were the pain → merge it, with the
  reason recorded.
- The pain was losing the session when the connection dropped → the scope is
  wrong. That is supervised reconnect, explicitly deferred to P0b here, so this
  slice would not have fixed the thing that hurt.

Draft rather than closed so the implementation and its review history stay
attached and this is cheap to revive.

Note: the branch is currently DIRTY against main (3.39.0 landed since), so CI
is not running on it. It needs a rebase before any merge attempt regardless.

PR #815, opened 2026-08-06
Per-device input grants and revocation, and an attach modal that shows pairing worked

openwong2kim:
Two related gaps found while dogfooding remote attach between two Macs.

## 1. Pairing succeeded and the screen looked untouched

`handlePairHost` / `handleAddHost` discarded the `host` the IPC call already
returns and only refreshed the list, so `selectedHostId` stayed `null` and the
right pane kept rendering its nothing-selected branch. That branch was wired to
`remote.pasteUrlHint` — the placeholder for the *paste-URL input* — so after
burning a single-use pairing code the modal told the operator to go paste a URL
they had just avoided needing. Success and no-op were indistinguishable.

- A successful pair or add now selects that host and lists its workspaces.
- The shared empty-state string is split into a no-hosts hint and a
  select-a-host hint, and a host whose panes are all closed says so instead of
  rendering a blank pane.
- Enter submits both forms. They were mouse-only; typing a host address and a
  code and pressing Enter did nothing, which was the other half of "I entered
  it and nothing happened".

## 2. Per-device revocation existed but nothing could reach it

`daemon.web.deviceRevoke` has been in the daemon since per-device credentials
shipped. No operator surface ever called it: the only revocation reachable from
the UI was `wmux web --stop`, which cuts every device at once. Device
credentials also have no expiry by design, so a phone or laptop you wanted to
retire could not be retired without cutting the ones you still wanted.

Adds **Paired devices…** to the Web popover: the roster by pair-time name and
last-seen, with a two-step revoke per device (red tint at rest, solid red on
the confirm, per DESIGN.md — revocation is permanent and the rows look alike).

The roster reads the device store rather than the running server, so it opens
while `wmux web` is stopped. That is exactly when someone who has just stopped
sharing wants to know what still holds a credential, and stopping the server
does not retire the credentials it handed out.

## Honesty on the failure paths

A panel review of the first draft found that every failure collapsed into one
verdict whose copy asserts "its live connections were cut". A daemon returning
that really had cut them. An RPC that timed out, hit a cut pipe, or reached an
older daemon may never have run the revoke at all — and said so in the same
words. That is a safety claim about a credential that is still fully live.

- Reasons are split: `unknown` for a daemon that did not answer,
  `persist-failed` for one that did.
- The daemon's `closed` count rides through. `{ok:false, closed:0}` and
  `{ok:false, closed:2}` are different facts, and only the second earns the
  "connections were cut" sentence.
- Reads that fail say so rather than rendering an empty roster and "0 active" —
  on a credential surface "we could not ask" must not display as "nobody has
  access".
- A revoke in flight cannot be dismissed by Escape or a backdrop click, so the
  verdict of a call the operator confirmed is always shown.
- Failures render in the device's own row; a shared line under a re-sorting
  list named no device.
- List records are validated element-wise — `WebDeviceSummary` mirrors the
  daemon's type rather than importing it, so a `revokedAt: null` would have
  painted a live device as a tombstone and removed its revoke button.
- Failed revokes are logged. The daemon audits only successful ones.
- Focus moves into the dialog and Tab is trapped there.

## Typing is now per device too

`--allow-input` used to be one process-wide switch over five routes that have
always been one grant: typing, spawning a pane, closing a pane, toggling the
permission gate, and approving a tool permission. On a surface whose whole
point is per-device control, the one capability that matters — arbitrary
command execution on the machine — had no per-device setting.

It does now, with the server flag kept as the **ceiling**. A server started
without `--allow-input` grants nothing to anyone, so "I started it read-only"
is still a complete answer and the CLI banner still means what it says. Within
that, each device brings its own grant; the operator token does not, because it
IS the operator.

Two defaults, deliberately different, because they answer different questions:
a record already on disk with no grant belongs to a device that has been typing
under the server flag all along, so it stays granted — upgrades do not mute
anyone. A pairing where nobody stated a grant falls back to the server's own
flag, which is what keeps headless hosts working: `wmux web --allow-input` in a
terminal has no popover to tick and no roster to grant from later. An
explicitly stated grant always wins, so the GUI's unticked checkbox still means
read-only.

`/api/config` answers with the caller's own grant instead of the server flag —
a read-only phone was being told `true` and rendering a composer that rejected
every keystroke. Stream tickets carry no grant at all: a ticket exists because
EventSource cannot set headers, lives two minutes, and is never revalidated, so
baking one in would be an input permission that outlives the operator taking it
away.

Three holes found while wiring it, each its own commit-level note: `coerceDevice`
rebuilt records field-by-field on load and silently dropped the new grant every
restart; "New code" re-minted a pairing code without the grant, so a ticked
checkbox registered a read-only device; and a `setInput(false)` whose write
failed left the device's streams up.

## Still not addressed

- `/api/upload` has its own `--allow-upload` grant and is not folded into this
  axis, so a read-only device can still upload if uploads are on.
- `POST /api/sessions/:id/resize` remains ungated, as before.
- A downgrade to a build that predates grants, followed by a re-upgrade, loses
  an explicit read-only setting (the roster is still schema version 1).
- There is no CLI to change a grant after pairing; that stays a GUI action.

## Verification

- ~40 new tests across the roster modal, the IPC handlers, the attach modal,
  the device store, and the server's auth gates. The six attach-modal
  regressions fail against the previous behaviour; the grant tests cover all
  five input routes refusing a read-only device, the ceiling overriding a
  granted one, the operator token not being narrowed, `/api/config` reporting
  per caller, tickets carrying no grant, the grandfather round-trip, and
  headless pairing still typing.
- `tsc --noEmit` clean; eslint clean on all changed files (the 20 pre-existing
  errors in `src/daemon/index.ts` are unchanged by this branch).
- Full suite: 9420 passed, 37 skipped, 0 failed (640 files).
- Dogfooded between a MacBook Air and a Mac mini over Tailscale.
- Two independent review rounds (Claude + GLM) on the roster and the auth
  model. Every confirmed finding is fixed in-branch; GLM's second-round verdict
  found no privilege-escalation path.

## Summary by CodeRabbit

- **New Features**
  - Added a paired-device roster in the Web popover with names, status, and last-seen times.
  - Added permanent per-device revocation with confirmation and immediate connection invalidation.
  - Added per-device input permissions; new pairings default to read-only unless enabled.
  - Added clearer loading, empty, and error states for device management.
  - Newly paired or added remote hosts are automatically selected.

- **Bug Fixes**
  - Improved Enter-key submission safeguards for pairing and URL forms.
  - Clarified remote workspace messages when hosts or workspaces are unavailable.

- **Documentation**
  - Documented device revocation, re-pairing, persistence, and input permissions.


openwong2kim:

This PR fixes two gaps surfaced while dogfooding remote attach between Macs, and they land together because the roster modal and the attach modal share the Web popover. The first is a UX dead-end: `handlePairHost`/`handleAddHost` discarded the host the IPC call already returned and only refreshed the list, so `selectedHostId` stayed null and the right pane kept rendering a nothing-selected branch wired to the paste-URL hint — a successful pair was visually indistinguishable from a no-op and told the operator to paste a URL they'd just avoided needing. It now auto-selects the returned host, splits the shared empty-state string into distinct no-hosts / select-a-host / no-workspaces messages, and wires Enter on both forms (previously mouse-only). The second and larger piece surfaces `daemon.web.deviceRevoke`, which has existed in the daemon since M3 but was never reachable from any operator surface: the only revocation the UI exposed was `wmux web --stop`, which cuts every device at once. A new **Paired devices…** entry in the Web popover lists every device holding a credential by pair-time name and last-seen, with a two-step per-device revoke (red tint at rest, solid red on confirm per DESIGN.md), and it reads the device store rather than the running server so it opens while `wmux web` is stopped.

What stands out is the care taken on the failure paths of a credential surface, which a panel review (Claude + GLM) correctly zeroed in on. Every failure used to collapse into `persist-failed`, whose copy asserts "its live connections were cut" — accurate when the daemon actually ran the revoke, but a thrown RPC, a cut pipe, or an older daemon with no such method may mean the revoke never executed, and reporting it the same way would tell an operator a credential was severed when the request never left the machine. Reasons are now split (`unknown` vs `persist-failed`), the daemon's `closed` count rides through so `{ok:false, closed:0}` and `{ok:false, closed:2}` render different sentences, reads that fail say so rather than rendering an empty roster that reads as "nobody has access," a revoke in flight cannot be dismissed by Escape or backdrop, records are validated element-wise (a stray `revokedAt: null` would have painted a live device as a tombstone and hidden its revoke button), and failed revokes are logged since the daemon audits only successful ones. 21 new tests cover the roster modal, IPC handlers, and attach-modal regressions; `tsc` clean, eslint clean, full suite 7148 passed / 0 failed; dogfooded over Tailscale. The one documented limitation worth tracking is that `--allow-input` remains a server-wide flag with no per-device read-only mode — revoking is the per-device control, the input grant is all-or-nothing, which is an auth-model change left out of scope here.

Priority: **P2**. The attach-modal fix resolves an active UX bug that wastes single-use pairing codes and misleads operators at the moment of success, and the roster fills a real operational/security gap by making an existing per-device revocation capability actually reachable — neither is a P1 outage, but together they materially improve the remote-sharing trust story.

PR #817, opened 2026-08-06
Make Detach on a remote workspace row reachable

openwong2kim:
Reported from dogfooding 3.39.0: right-clicking an attached remote workspace
opens its menu, and clicking **Detach** does nothing.

## Root cause

The handler was never reached. `RemoteWorkspaceItem` dismisses its menu from a
document-level `mousedown` listener, and `mousedown` fires before `click` — so
pressing Detach unmounted the button under the pointer, and the button's own
`onClick` never ran. The menu container was stopping propagation, but on
`click`, one event too late.

```
mousedown on Detach
  → document listener → setMenuPos(null) → button unmounts
  → click never fires  → onDetach never called
```

It stops `mousedown` now. `AttachRemoteModal` already uses that shape for its
backdrop, so this brings the two menus into line rather than inventing a
pattern.

## Tests

Four, in a new `RemoteWorkspaceItem.test.tsx`:

- the press as it actually happens — `mousedown`, assert the menu survived,
  then `click`, assert `onDetach` fired. **Fails against the previous code.**
- right-click opens the menu
- a `mousedown` outside still dismisses it, without detaching
- Escape still closes it, without detaching

The last three pass either way. They are here so the fix cannot be "solved"
later by breaking dismissal.

## Not in this PR

The second bug from the same session — Cmd+Shift+R making the remote row
disappear — is not an issue with this component. wmux never calls
`Menu.setApplicationMenu()`, so Electron's default menu owns the accelerator
table and Cmd+Shift+R is its **Force Reload**; on macOS that is an NSMenu key
equivalent the renderer cannot `preventDefault()`. The reload wipes the
renderer store, and attached remote workspaces are per-app-run state.

That same cause already produced the paste-corruption race worked around in
`useTerminal.ts:962`, and it also means the documented "Ctrl+Shift+R = rename
workspace" shortcut is dead on macOS. The root fix is for wmux to own its
application menu, which is a separate change with a much wider blast radius
(three platforms, thirteen existing Cmd+Shift bindings, and the existing
workaround's assumptions) and is tracked on its own.

## Verification

`tsc --noEmit` clean, eslint clean on both files, Sidebar suite 42/42.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the **Detach** action for attached remote workspaces so it works reliably from the context menu.
  - Context menus now remain open long enough for menu actions to complete while still dismissing correctly when clicking outside or pressing Escape.

- **Documentation**
  - Added a changelog entry describing the remote workspace detachment fix.


openwong2kim:

This PR fixes a UI defect where right-clicking an attached remote workspace in the sidebar and choosing **Detach** silently did nothing — the action was entirely unreachable, not merely flaky. The root cause is a DOM event-ordering race: the menu's document-level `mousedown` dismiss listener fires before `click`, so the button unmounts under the pointer before its `onClick` can ever run.

The fix is a one-line behavioural change that moves `stopPropagation` from the container's `onClick` to its `onMouseDown`, matching the pattern `AttachRemoteModal` already uses for its backdrop. It lands with a focused new test suite (`RemoteWorkspaceItem.test.tsx`) whose key case reproduces the real press sequence — `mousedown` then `click` — and is explicitly noted to fail against the previous code, plus three guards ensuring outside-click dismissal and Escape still work. `tsc` and eslint are clean and the Sidebar suite stays 42/42, so this is a low-risk, well-scoped fix that can go in. The separate Cmd+Shift+R / application-menu issue called out in the PR body is correctly left for its own change given its wider blast radius.

PR #819, opened 2026-08-06
Mirrored remote panes: app font and theme, shared width model, and clipping

openwong2kim:
Reported from dogfooding: "is the text in this pane oddly bold and slightly
bigger, or am I imagining it?"

Not imagining it.

## Root cause

`RemoteMirrorTerminal` was the only terminal in the app constructed with no
visual settings:

```ts
const term = new Terminal({ convertEol: false, scrollback: 2000, disableStdin: false });
```

So it ran on xterm's defaults. Side by side with what a local pane gets from
`useTerminal.ts`:

|                        | local pane            | mirrored pane            |
|------------------------|-----------------------|--------------------------|
| `fontSize`             | 14 (setting)          | **15** (xterm default)   |
| `fontFamily`           | Cascadia Code         | **`monospace`**          |
| `theme`                | selected theme        | **xterm default**        |
| `minimumContrastRatio` | applied               | **absent**               |

One pixel larger and a different typeface is precisely what "slightly bigger
and bolder" looks like.

The theme was the part that would not have been named: DESIGN.md gives terminal
content the theme's ANSI palette, and this pane was rendering remote output in
xterm's own — so remote ANSI colours and the pane background were both off, on
the one surface where a wrong colour reads as the remote host's output rather
than as a local bug.

## Fix

Read the same four settings a local pane reads, pass them at construction so
the first paint is already correct, and keep them in sync with a runtime effect.

The settings deliberately do **not** join the mount effect's dependency list.
Re-creating the terminal on a font change would drop the mirrored scrollback,
and the remote only repaints on a fresh attach — there is nothing to rebuild
from. Same reason `useTerminal.ts` mutates `options` in place.

## Tests

Two, both failing against the previous code:

- construction carries the app's font size, family, theme and contrast floor —
  asserted against xterm's specific defaults (`15`, `'monospace'`), since those
  defaults are the regression
- a font-size change mutates the existing instance rather than replacing it
  (instance count unchanged, not disposed, new value visible)

The existing suite's fake Terminal grew an options bag to make this observable;
its other 11 tests are unchanged and still pass.

## Verification

`tsc --noEmit` clean, eslint clean, Remote + Sidebar suites 55/55.

## Summary by CodeRabbit

* **Bug Fixes**
  * Mirrored remote terminals now inherit local font, theme, background, and contrast settings.
  * Wide Unicode characters render correctly in remote terminal panes.
  * Terminal content is clipped to prevent visual overflow.
  * Font-size changes now apply immediately without recreating the terminal.

### DIFF
--- a/changelog.d/819.md
+++ b/changelog.d/819.md
@@ -1,22 +1,31 @@
 ### Fixed
 
-- **Mirrored remote panes now use your terminal font and theme.** An attached
-  remote workspace built its terminal without any of the app's visual settings,
-  so it fell back to xterm's own defaults — `monospace` at 15px on black,
-  outside the theme's ANSI palette. Next to a local pane it read as slightly
-  bolder and slightly larger, because it was. Font size, font family, theme and
-  the contrast floor now match a local pane, and changing them updates the
-  mirror without dropping what the remote has already sent.
+- **Mirrored remote panes use your terminal font, theme and contrast floor.**
+  An attached remote workspace built its terminal without any of the app's
+  visual settings, so it fell back to xterm's own defaults — `monospace` at
+  15px on black, outside the theme's ANSI palette. Next to a local pane it read
+  as slightly bolder and slightly larger, because it was. Changing the settings
+  now updates the mirror without dropping what the remote has already sent.
 
-- **Mirrored remote panes no longer tear on Korean and other wide text.** The
-  mirror re-renders a grid the remote daemon computed, but it was the one
-  terminal in the app that skipped the shared Unicode width model — so every
-  double-width character put the two grids one cell further apart, and rows
-  arrived interleaved and torn. It now measures the way every other wmux
-  terminal does.
+  One difference remains: local panes render through WebGL and mirrors use
+  xterm's DOM renderer, whose text antialiasing differs. Matching size, family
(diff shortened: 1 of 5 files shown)


openwong2kim:

This PR fixes a visual inconsistency in `RemoteMirrorTerminal`, which was the only terminal in the app constructed without the app's visual settings — so an attached remote workspace fell back to xterm's own defaults (`monospace` at 15px, xterm's stock theme, no contrast floor), reading as "slightly bolder and bigger" next to a local pane. The change reads the same four settings a local pane reads (`terminalFontSize`, `terminalFontFamily`, `theme`, `minimumContrastRatio`), passes them at construction so the first paint is already correct, and keeps them in sync via a runtime effect that mutates `term.options` in place rather than recreating the terminal (which would wipe the mirrored scrollback the remote only repaints on re-attach). The two added tests assert against xterm's specific defaults (`15`, `'monospace'`) as the regression target and verify a font-size change updates the existing instance rather than replacing it — both fail against the old code, the other 11 in the suite are untouched. It's a small, well-scoped fix with the right test coverage; the discipline of keeping the mount effect `[]`-keyed (mirroring `useTerminal.ts`) is the one thing worth confirming on review, since that invariant is what protects the scrollback.

PR #820, opened 2026-08-07
Replace token files instead of overwriting them in place

openwong2kim:
CI found this, not a person: `RemoteHostsStore.test.ts > removes a host by id`
started timing out at 5s on the Windows `validate` job. It is a synchronous
test doing two small local writes, so the timeout was never about the test.

## What was actually wrong

`secureWriteTokenFile` writes **in place**, and an in-place overwrite is what
forces the expensive branch. `writeFileSync` preserves an existing file's ACL,
so a broad EXPLICIT ACE already on it survives and only the PowerShell DACL
rebuild removes one — 1.8-3.8s under antivirus, per the measurement already
recorded in `DeviceStore`.

The remote-host registry is rewritten on **every** add and **every** remove, so
all but the first write paid it. `removes a host by id` is the only test in that
file that persists twice, which is why it is the only one that failed. On a real
Windows machine that stall lands on the × button in the attach modal.

## The fix

`secureReplaceTokenFile`: create a new inode, harden it, write into it, rename
it over the target. A freshly created file carries only INHERITED ACEs, which is
exactly the case the fresh-file primitive is documented as security-equivalent
for, and the DACL travels with the file through the rename.

Deliberately **not** the deferred re-harden `DeviceStore` uses. Its own comment
says why that would be wrong here: *"The deferred window is acceptable here in a
way it would not be for a token file: this file holds no secret."* This registry
is a token file — its records grant scrollback read and keystroke injection on
other machines.

## What review changed

The first draft removed a stall and introduced two new problems. Both reviewers
caught the same two; one found three more.

| finding | why it mattered |
|---|---|
| hardening ran *after* the write | the in-place path had **no** exposure window; a fresh inode has only inherited ACEs, so the tokens sat readable for two process spawns |
| `pid.timestamp` temp name | enumerable, and `writeFileSync` follows symlinks — a pre-planted name takes delivery of the token |
| no rename retry | antivirus holds the handle of a file just written (#658); icacls runs on that temp immediately before, so a bare rename throws on exactly the machines this targets — and `REMOTE_HOSTS_REMOVE` has no try/catch, trading a freeze for an error |
| no fsync | "a crash cannot leave a half-written registry" was not a claim the code could make; `load()` folds a parse failure into an empty list, so a zero-length file reads as *every remote host gone* |
| swallowed unlink, no sweep | a crash between write and rename leaves a full copy of the credentials under a temp name that nothing ever looks at |

The rename policy mirrors `atomicWrite/core.ts` rather than importing it:
`shared/` must not depend on `daemon/`, and that module's writer unconditionally
rotates the previous file to `.bak` — which for a token file means a second,
unhardened copy of live credentials on disk.

## Scope

One caller. `peers.ts`, `webStateStore.ts` and the other `secureWriteTokenFile`
users may have the same latent stall, but whether a file is rewritten often
enough to care is a per-file judgement, not a mechanical substitution.

## Verification

- 8 new tests on the replace path (exclusive creation with an unpredictable
  name, harden-before-write ordering, no PowerShell, fsync-before-rename, retry
  on EBUSY, no retry on ENOSPC, nothing written when hardening fails) plus 2 on
  the sweep
- `tsc --noEmit` clean, eslint clean, full suite 9711 passed / 0 failed
- Unblocks #819, whose `validate` job fails on this same timeout

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote host credential updates to prevent Windows UI freezes.
  * Prevented partially written or unhardened credential files from being published.
  * Added cleanup for stale temporary credential files.
  * Improved resilience when transient file replacement errors occur.

* **Security**
  * Strengthened temporary credential files before writing sensitive data.
  * Added atomic replacement and cleanup safeguards to protect existing credentials.


openwong2kim:

This PR fixes a Windows UI freeze and tightens credential-file handling on the remote-hosts registry. `RemoteHostsStore` rewrites its token file on every add/remove, and `secureWriteTokenFile` wrote in place — which under antivirus forces a PowerShell DACL rebuild measured at 1.8–3.8s, landing directly on the × button in the attach modal (CI caught it as a 5s test timeout). The fix introduces `secureReplaceTokenFile`: create a fresh inode with an unpredictable name (`O_EXCL`), harden it *before* the bytes go in, fsync, then rename over the target with retry on EBUSY/EPERM/EACCES. Because a freshly-created file carries only inherited ACEs, it skips the expensive rebuild entirely, makes the write atomic, and closes an exposure window the first draft introduced.

Scope is deliberately one caller (`RemoteHostsStore`); `peers.ts` and `webStateStore.ts` keep `secureWriteTokenFile` until per-file judgement says otherwise. Review (Claude + GLM) caught five real issues on the first draft and all are addressed, the suite is 9711/0 green, and it unblocks #819 whose `validate` job fails on this same timeout. This is a P2: the freeze is user-facing and the file holds bearer tokens granting scrollback read and keystroke injection on other machines, but it is not a crash or data-loss path.

**Priority: P2**

openwong2kim:
Moving to draft. The problem is real; this approach does not solve it, and CI
plus a two-model panel agree on why.

## The load-bearing claim does not hold

This PR reorders to "harden the empty temp, then write the credential into it,"
and argues that closes the exposure window an in-place overwrite never had.

On Windows it does not. Access is evaluated when a **handle is opened**, not on
every read. An attacker who opens a read handle in the gap between
`openSync` creating the file (with only the parent's inherited ACEs) and icacls
finishing — one to two process spawns, longer under antivirus — keeps reading
through that handle after the DACL is tightened, including the token bytes
written afterwards. Reordering moved the window; it did not close it.

Closing it properly means the temp must never exist under broad ACEs at all:
a dedicated owner-only subdirectory hardened once, with temps created inside it.
That is a different design from what is here.

## And the stall it removes comes back on some machines

`#90` is documented in this same module: icacls parses argv in the console's
legacy OEM codepage, so a non-ASCII profile mangles. That reasoning was applied
to the *principal* and not to the *temp path*. On `C:\Users\홍길동\.wmux\…` the
icacls fast path fails and falls through to the PowerShell rebuild — so the
1.8-3.8s stall this PR exists to remove happens on **every** add and remove for
those users, where before it happened on all but the first write. Not obviously
an improvement.

## Other findings

| | |
|---|---|
| **2-MODEL** | `Atomics.wait` blocks the main thread up to 310ms on the retry path. Far better than 3.8s, same kind of freeze — and it fires precisely in the AV environments this targets |
| **2-MODEL** | path/fd split: icacls hardens the PATH while the write goes to the FD. A directory-writable attacker who swaps the temp after `open` gets their file hardened and published, while the token goes to an orphaned inode |
| Claude (8) | the first-write branch delegates to `secureWriteTokenFile`, bypassing every guarantee this function adds — O_EXCL, atomicity, fsync, harden-before-write — at the moment the token first touches disk |
| Claude (8) | `sweepStaleTokenTemps` only runs when the real file exists, so a crash before the first rename leaves a full credential copy that is never collected |
| Claude (5) | the sweep unlinks temps another process may be mid-write on, and the resulting `ENOENT` is not in the retry set |

## CI

`validate` passes — but so does it on #819, which does **not** have this change.
The timeout that started this was marginal, not deterministic, so the urgency I
attached to it was wrong.

`Baseline (windows-latest)` now fails on
`RemoteHostsStore.test.ts > tolerates a corrupt file on construct`, a *different*
5s timeout, because `load()` now runs `sweepStaleTokenTemps` and that readdir
costs real time on Windows. The change traded one source of slowness for
another.

## What stays true

`secureWriteTokenFile` writes in place, an in-place overwrite forces the
PowerShell DACL rebuild, and this registry is rewritten on every add and remove.
That cost is measured and documented in `DeviceStore`, and it lands on the ×
button. Worth fixing — with a design that survives the above.

Draft rather than closed so the findings stay attached to the attempt.

Issue #822, opened 2026-08-07
Allow configuring the sidebar / UI font size — sidebars hard-coded at 8–12px


cnxiekun:
### Problem / motivation

The terminal font size is configurable (Settings → Appearance → Terminal → Font size, 12–24px), but the **sidebar and UI font sizes are hard-coded** and cannot be adjusted. On high-DPI displays, large monitors, or for accessibility needs, the small sidebar text is genuinely hard to read.

### Where it's hard-coded (`src/renderer`)
- `components/Sidebar/Sidebar.tsx` — `text-[11px]`, `text-xs`
- `components/Sidebar/MiniSidebar.tsx` — `text-[10px]`, `text-[8px]`
- `components/FleetView/FleetView.tsx` — `text-xs`, `text-[10px]`, `text-[11px]`, `text-sm`
- `components/FleetView/FleetCard.tsx`, `ApprovalInboxList.tsx` — `text-[10px]`, `text-[11px]`

The theme system (`src/renderer/themes.ts`) only exposes **color** variables (`--accent-*`, `--bg-*`, `--text-main`…), there is no user-CSS injection, and no UI-zoom setting — so none of these sizes can be overridden.

### Proposed solution

1. Add a **UI / sidebar font size** control in Settings → Appearance (mirroring the terminal one), and/or
2. Route the sidebar text through the existing text-size tokens (`--text-body-size` / `--text-caption-size` in `globals.css`) or new tokens, so custom themes or a future "UI scale" setting can scale them.

An MVP could be a single "UI scale" multiplier over those tokens.

### Alternatives considered

- Changing the terminal font size only affects the terminal panes, not the sidebar.
- Custom themes only accept color variables — no font-size tokens.
- No user-CSS injection hook exists today.
- tmux-style multiplexers and VS Code both expose an overall UI/editor font size; a similar control here would match user expectations.

### Additional context

- wmux 3.38.9 on Windows 11
- The terminal font-size setting correctly affects only the terminal panes, confirming the sidebar sizes are independent and currently fixed.

### Confirmation

- [x] I searched existing issues and Discussions and this is not a duplicate.

openwong2kim:
## [wmux-hermes] Triage Summary

[...]

PR #824, opened 2026-08-07
Re-place native window controls under a renderer zoom

openwong2kim:
Groundwork for the configurable UI scale asked for in #822. This lands the
mechanism only — there is no user-facing setting yet.

## Why not the sweep

The obvious route to a scalable UI is to rewrite every hard-coded font size.
That is 481 `text-[Npx]` sites across 61 files plus 23 inline styles, and it
mints a permanent `text-ui-N` vocabulary that every future contributor has to
learn — while `globals.css` already declares a semantic 16/14/13/11 ramp it
intends to migrate to.

The cheaper route is `webFrame.setZoomFactor`, which is what VS Code's
`window.zoomLevel` does. The reason to reject it was that the native window
controls are positioned in OS coordinates and do not follow a renderer zoom, so
zooming desyncs them from the custom 36px titlebar that DESIGN.md's chrome
module is built around.

That desync is real. It is also fixable, which is what this PR shows.

## Measured

A harness with the same window options as `createWindow` (`titleBarStyle:
'hidden'`, `trafficLightPosition: {x:12, y:11}`) drawing a 36px bar with a
hairline at its bottom edge, captured and measured at the pixel level:

| Condition | Bar height | Light centre | Misalignment |
|---|---|---|---|
| zoom 1.0 (today) | 35.5pt | 17.75pt | **0.00pt** |
| zoom 1.4, no re-place | 50.0pt | 17.75pt | **-7.25pt** |
| zoom 1.4, re-placed | 50.0pt | 24.75pt | **-0.25pt** |

0.25pt is rounding — under one physical pixel on a retina display.

## The arithmetic

The traffic lights render at a fixed 14pt at every zoom level. So the offset
that centres them has to be recomputed against the scaled bar:

```
y = (36 * zoom - 14) / 2
```

Scaling the shipped offset instead (`11 * 1.4 = 15.4`) lands 3pt high and
reintroduces the desync. `uiZoom.test.ts` asserts against that wrong answer by
name so the mistake cannot come back.

## What is in

- `uiZoom.ts` — clamp, the re-centring arithmetic, and a guarded apply that
  calls `setWindowButtonPosition` on macOS and rides the existing
  theme-following `setTitleBarOverlay` path on Windows.
- `uiZoom.test.ts` — 6 cases covering the clamp, the zoom-1 identity, the
  re-centring, and the degenerate small-bar case.
- `createWindow.ts` — `WMUX_UI_ZOOM` applies a factor at launch so this can be
  dogfooded before the Settings surface is designed.

Unset, the window is untouched and every computed value is identical to today.

## Not verified

- **Windows.** The `titleBarOverlay` height path is written and typechecked but
  not run — the measurements above are macOS. It needs a Windows dogfood pass
  before any setting ships on top of this.
- **The real renderer.** The numbers come from the harness, not from wmux's own
  titlebar. Same window options, but not the same app.
- **Terminal double-scaling.** xterm reads `fontSize` from JS, so at a non-1
  zoom the terminal scales twice — once by the user's own font setting, once by
  the window zoom. Dividing by the zoom factor in `useTerminal` is the likely
  answer, but it is not in this PR.

Refs #822

## Summary by CodeRabbit

* **Improvements**
  * Improved window scaling support across macOS and Windows.
  * Native macOS traffic-light controls and Windows title-bar overlays now remain correctly positioned and sized when the interface zoom changes.
  * Zoom values are safely constrained to supported limits, preventing invalid sizing or negative positioning.
* **Reliability**
  * Window display updates are handled safely across supported configurations, including windows that are closing or unavailable.

### DIFF
--- a/changelog.d/824.md
+++ b/changelog.d/824.md
@@ -1,0 +1,8 @@
+### Added
+
+- **The native window controls now follow a renderer zoom.** Scaling the UI used
+  to leave the macOS traffic lights and the Windows window buttons behind — they
+  are drawn by the OS and do not move with the renderer, so the custom titlebar
+  grew while the controls stayed put. They are now re-placed whenever the zoom
+  changes, which keeps them centred in the titlebar at any scale. Nothing in the
+  app zooms yet; this is the groundwork for a configurable UI size.
--- a/src/main/window/__tests__/uiZoom.test.ts
+++ b/src/main/window/__tests__/uiZoom.test.ts
@@ -1,0 +1,60 @@
+import { describe, it, expect } from 'vitest';
+import {
+  clampUiZoom,
+  macTrafficLightPosition,
+  winOverlayHeight,
+  UI_ZOOM_MIN,
+  UI_ZOOM_MAX,
+} from '../uiZoom';
+
+/**
+ * #822 — the arithmetic that decides whether UI zoom can keep the native
+ * window controls aligned with DESIGN.md's 36px chrome module.
+ *
+ * The plan's rejected-alternative claim was that the native controls "do not
+ * follow a renderer zoom", so zoom desyncs the titlebar. These cases pin the
+ * correction: the controls are re-placeable, and the offset must be
+ * RE-CENTERED against the scaled bar, not itself scaled. Scaling the y-offset
+ * (11 → 15.4 at 1.4) is what produces the desync the plan feared.
+ */
+
+describe('clampUiZoom', () => {
+  it('clamps below the minimum and above the maximum', () => {
+    expect(clampUiZoom(0.1)).toBe(UI_ZOOM_MIN);
+    expect(clampUiZoom(9)).toBe(UI_ZOOM_MAX);
+  });
+
+  it('passes valid values through and falls back to 1 on garbage', () => {
+    expect(clampUiZoom(1.4)).toBe(1.4);
+    expect(clampUiZoom(Number.NaN)).toBe(1);
+    expect(clampUiZoom('1.4')).toBe(1);
+    expect(clampUiZoom(undefined)).toBe(1);
+  });
+});
+
+describe('macTrafficLightPosition', () => {
+  it('reproduces the shipped position at zoom 1', () => {
+    // createWindow.ts uses { x: 12, y: 11 } — (36 - 14) / 2 = 11.
+    expect(macTrafficLightPosition(1)).toEqual({ x: 12, y: 11 });
+  });
+
+  it('re-centers rather than scaling the y-offset', () => {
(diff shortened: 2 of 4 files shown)


openwong2kim:

This PR lands the mechanism for a configurable UI scale — the groundwork for #822 — by adopting `webFrame.setZoomFactor` and, critically, fixing the native-window-control desync that was the reason to reject that approach. It adds `uiZoom.ts` (clamp, the re-centring arithmetic, and a guarded cross-platform apply), six unit cases pinning the arithmetic against the scaled-vs-scaled-offset mistake, and a `WMUX_UI_ZOOM` launch hook in `createWindow.ts` so it can be dogfooded before the Settings surface exists; a changelog fragment is included. The measured pixel data (0.25pt misalignment at 1.4× on macOS) and the test that asserts `macTrafficLightPosition(1.4).y` is *not* `round(11 * 1.4)` make the central claim verifiable without booting Electron.

Two things I want tracked before this is merged. First, the Windows `titleBarOverlay` height path is written and typechecked but never run — the measurements are macOS-only, so it needs a real Windows dogfood pass (this is wmux's primary platform). Second, xterm reads `fontSize` from JS, so at any non-1 zoom the terminal scales twice; dividing by the zoom factor in `useTerminal` is called out as the likely fix and should land with, or before, any user-facing setting built on top of this. Priority P2: solid, tested infrastructure for a genuine usability issue, but it is plumbing with no user-facing surface yet and known unverified paths.

Refs #822.

openwong2kim:

This PR is groundwork for the configurable UI scale requested in #822 — it lands the zoom mechanism only, with no user-facing Settings surface yet. The key insight is rejecting the `webFrame.setZoomFactor` route that VS Code uses *unless* the native-window-control desync it causes is also fixed, and this PR shows that fix: it recomputes the macOS traffic-light offset against the scaled titlebar using `y = (36 * zoom - 14) / 2` (the shipped `11 * zoom` answer is provably wrong and `uiZoom.test.ts` asserts against it by name), and rides the existing `setTitleBarOverlay` path on Windows. The diff adds `uiZoom.ts` (clamp + re-centring arithmetic + guarded apply), 6 tests, and a `WMUX_UI_ZOOM` launch hook for dogfooding.

It matters because the naive approach to UI scaling would have either desynced the native controls or forced a 481-site font-size rewrite, and this gives a third option that neither does. The PR is appropriately scoped as mechanism-only: it explicitly calls out three things not verified (Windows `titleBarOverlay` path is typechecked but unrun, the harness is not wmux's real titlebar, and xterm would double-scale because it reads `fontSize` from JS). Priority **P2** — it is the unblocking prerequisite for the #822 setting, but it is not itself user-visible and the Windows path still needs a dogfood pass before anything ships on top of it.

PR #825, opened 2026-08-07
docs(readme): lead with the atomic adoption guarantee

openwong2kim:
## What

- The hero paragraph now leads with what is actually unique about the harvest step: selected hunks land as one all-or-nothing `git apply` — the target tree takes the whole selection or stays untouched. "Review hunk by hunk" undersold this; per-hunk checkboxes exist elsewhere, the transactional guarantee does not.
- Adds a founding-date line (building in the open since March 2026) to the intro blockquote.

Docs only — no code, no changelog fragment.

## Summary by CodeRabbit

* **Documentation**
  * Updated the README to explain all-or-nothing adoption of selected changes from isolated worktrees.
  * Added information about wmux being developed openly since March 2026.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 ### The workspace multiplexer for AI agents.
 
-Run **fleets of Claude Code, Codex & Gemini in parallel** — each agent in its own pane, or fan one prompt out into **N isolated git worktrees** you review **hunk by hunk**. Native on **Windows & macOS**, with approval gates, agent-to-agent channels, and an integrated browser your agents can drive. Walk away — after a crash or **full OS reboot**, they come back mid-conversation.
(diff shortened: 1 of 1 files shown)


openwong2kim:

This PR is a copy-only change to the README hero paragraph: it re-leads with the atomic adoption guarantee (selected hunks land as one all-or-nothing `git apply`, so the target tree takes the whole selection or stays untouched) rather than the previous "review hunk by hunk" framing, and adds a founding-date line to the intro blockquote. It touches no code, no config, and carries no changelog fragment — the diff is two sentences.

The reword matters because the transactional guarantee is the feature that actually distinguishes the harvest step from what per-hunk checkboxes elsewhere already offer; underselling it in the first paragraph is a missed first impression. Low-risk, safe to merge as-is.

Priority: **P3** — documentation polish, no functional impact.

PR #826, opened 2026-08-07
feat(fanout): per-task port assignment and trust-gated worktree setup hook

openwong2kim:
## What

Two quality-of-life gaps in the fan-out journey:

1. **Port range** — `wmux.json` can declare `fanout.portRange` (e.g. `"3000-3010"`). Each spawned task gets a distinct free port injected as `WMUX_TASK_PORT` in its pane env, so N tasks starting dev servers stop colliding. Ports are pre-assigned before any spawn (re-scanning per task would double-assign ports a dev server has not bound yet). The port value never reaches a shell — env overlay only — so it is not trust-gated.
2. **Setup hook** — `fanout.setup` declares a shell command run in each fresh worktree before the agent starts (install deps, copy `.env`). It rides the existing `wmux.json` trust gate: shown verbatim in the trust dialog, runs only on trusted bytes, and anything else (untrusted / stale / denied) fail-closes — a stale config skips the hook and reports `setupSkipped` rather than running old bytes. Hook failure fails that task closed: mission compensated, worktree preserved, no pane.

## Tests

- `fanoutEnvironment.test.ts` (new, 6): distinct ports, busy-port skip, exhausted range, parse rejections, trust-gate matrix.
- `FanOutService.test.ts` (+2): distinct `WMUX_TASK_PORT` reaches spawn env; stale-config hook does not execute.

## Notes

- Changelog fragment: `changelog.d/t2-fanout-ports.md`; how-to at `docs/how-to/fan-out-task-environment.md`.
- Result fields `port` / `setupFailed` / `setupSkipped` are not yet surfaced in the fan-out report UI — follow-up.

## Summary by CodeRabbit

* **New Features**
  * Added fan-out task port allocation through `wmux.json`, with assigned ports exposed as `WMUX_TASK_PORT`.
  * Added optional trusted worktree setup commands that run before task startup.
  * Added reporting for skipped setup hooks and unavailable ports.
* **Bug Fixes**
  * Failed setup commands prevent task startup while preserving the prepared worktree.
  * Invalid, privileged, busy, or exhausted port ranges are handled safely.
  * Setup timeouts terminate related processes without blocking other tasks.
* **Documentation**
  * Added configuration and behavior guidance for fan-out ports and setup hooks.

## Prior art

Concept-level borrowing only, no code imported: per-task port ranges appear in uzi (devflowinc/uzi); post-create worktree setup hooks appear in Orca (stablyai/orca) repository hooks. The implementation here is built on wmux's own PortAllocator pattern and wmux.json trust gate.

### DIFF
--- a/changelog.d/t2-fanout-ports.md
+++ b/changelog.d/t2-fanout-ports.md
@@ -5,8 +5,11 @@
   spawns is assigned one free port from that window, exported to its pane as
   `WMUX_TASK_PORT`. Before this, eight tasks that all ran `npm run dev` fought
   over one port and seven of them died on startup. Ports are probed and assigned
-  before any task spawns, so no two tasks in a fan-out collide; if the window
-  runs out, the remaining tasks simply start without the variable.
+  before any task spawns, so no two tasks in a fan-out collide — and a port just
+  handed out stays claimed for ten minutes, so a second fan-out started while
+  the first one's servers are still booting doesn't reuse it either. Windows are
+  capped at 512 ports; if one runs out, the remaining tasks simply start without
+  the variable.
 
 - **Worktree setup hook for fan-out.** `fanout.setup` in `wmux.json` is a shell
   command run inside each freshly created worktree *before* its agent starts —
@@ -17,5 +20,9 @@
   review it. An edit demotes the file to stale and the hook stops running. When
   a declared hook is skipped for lack of trust the fan-out says so instead of
   quietly doing nothing, and a hook that fails leaves its task unspawned with
-  the worktree preserved rather than starting an agent in a half-prepared tree.
+  the worktree preserved rather than starting an agent in a half-prepared tree —
+  only that task; the rest of the fan-out continues. Output is never capped (a
+  chatty `npm ci` won't be killed), and a hook that hits its five-minute ceiling
+  has its whole process tree killed, so a background install can't outlive it
+  and keep writing into the preserved worktree.
   See `docs/how-to/fan-out-task-environment.md`.
--- a/changelog.d/t2-fanout-ports.md
+++ b/changelog.d/t2-fanout-ports.md
@@ -1,0 +1,21 @@
+### Added
+
+- **Fan-out tasks can each get their own port.** Declare `fanout.portRange`
+  (e.g. `"3000-3010"`) in your repo's `wmux.json` and every task a fan-out
+  spawns is assigned one free port from that window, exported to its pane as
+  `WMUX_TASK_PORT`. Before this, eight tasks that all ran `npm run dev` fought
+  over one port and seven of them died on startup. Ports are probed and assigned
+  before any task spawns, so no two tasks in a fan-out collide; if the window
+  runs out, the remaining tasks simply start without the variable.
+
+- **Worktree setup hook for fan-out.** `fanout.setup` in `wmux.json` is a shell
+  command run inside each freshly created worktree *before* its agent starts —
+  the `cp ../../.env .` and `npm ci` you used to type into every agent's first
+  turn. It is trust-gated exactly like supervised panes: the command is shown
+  verbatim in the trust dialog and runs only against `wmux.json` bytes you have
+  explicitly approved, so a hook arriving via a pull request is inert until you
+  review it. An edit demotes the file to stale and the hook stops running. When
+  a declared hook is skipped for lack of trust the fan-out says so instead of
+  quietly doing nothing, and a hook that fails leaves its task unspawned with
+  the worktree preserved rather than starting an agent in a half-prepared tree.
+  See `docs/how-to/fan-out-task-environment.md`.
--- a/docs/how-to/fan-out-task-environment.md
+++ b/docs/how-to/fan-out-task-environment.md
@@ -18,7 +18,9 @@ Both are declared once, in your repo's `wmux.json`:
 
 ## `fanout.portRange`
 
-Written as `"min-max"`, inclusive, within 1024–65535. Each spawned task is
+Written as `"min-max"`, inclusive, within 1024–65535 and at most 512 ports wide
+(the window is probed by binding, so a whole-ephemeral-space range would mean
+tens of thousands of probes before the first task starts). Each spawned task is
 assigned one free port from the window and gets it as **`WMUX_TASK_PORT`** in
 its pane's environment, so eight parallel `npm run dev` runs don't fight:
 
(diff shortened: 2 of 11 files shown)


openwong2kim:

This PR adds two quality-of-life features to the fan-out journey, both declared once in `wmux.json`. First, `fanout.portRange` (e.g. `"3000-3010"`) pre-assigns each spawned task a distinct free port exposed as `WMUX_TASK_PORT`, so N parallel `npm run dev` invocations stop colliding — ports are probed and reserved *before* any task spawns to avoid double-assigning a port a dev server hasn't bound yet. Second, `fanout.setup` declares a trust-gated shell command (e.g. `npm ci`, `cp ../../.env .`) that runs inside each fresh worktree *before* its agent starts, riding the existing `wmux.json` trust gate exactly like supervised panes; a stale or untrusted config skips the hook and reports `setupSkipped` rather than running old bytes, and a failed hook leaves the task unspawned with the worktree preserved.

The trust-gating discipline here is the important part and it is done right: `portRange` is not trust-gated because the value only ever reaches `WMUX_TASK_PORT` env (never a shell), while `setup` is gated because it is arbitrary shell. Tests cover distinct-port assignment, busy-port skip, exhausted range, the trust-gate matrix (6 new in `fanoutEnvironment.test.ts`), and verification that a stale-config hook does not execute. Priority **P2** — directly improves the flagship fan-out workflow at scale, well-tested, but it changes the spawn path and should get real dogfooding before release.

openwong2kim:
Held this back from tonight's batch (#824/#825/#827/#828/#829/#830/#831 all went in) because I couldn't get a live run of it.

Fan-out derives the repo from the calling pane's directory, and on this box the pane's tracked cwd never follows a `cd` — filed that separately as #833. So every attempt ended at:

```
```

and there's no way around it from the wire surface (`pane.split` takes no cwd, `workspace.new` only forwards `name`). Nothing here is implicated — the cwd path isn't in this diff.

The other half is that the setup hook only runs on trusted wmux.json bytes, and trusting them is a dialog I'm not going to click through on someone's behalf. So the trusted-execution path needs a human at the keyboard anyway.

Plan: once #833 is sorted, open the app on a repo, run the fan-out, and check both halves — distinct `WMUX_TASK_PORT` per task, and the hook running on trusted bytes / reporting `setupSkipped` on stale ones.

openwong2kim:
Tried again with a proper trace. Still blocked, and now I know exactly where.

The cwd signal does reach the renderer — daemon broadcasts `cwd.changed`, and the workspace metadata updates from it. What does not update is the **surface** cwd, which is the one `resolveSenderSurfaceCwd()` reads. Both writes happen in the same callback; only the workspace one lands. Details and the pipe trace are in #833.

The last workaround I had in mind is gone too: `surface.new` ignores its `cwd` param, so I cannot even create a pane already sitting in the repo. Between that, `pane.split` taking no cwd, and `workspace.new` forwarding only `name`, there is no way from any surface to hand fan-out a pane whose cwd is a git repository.

So this stays parked on #833 rather than on anything in this PR. Once that is fixed the check is quick — distinct `WMUX_TASK_PORT` per task, and the hook running on trusted bytes vs reporting `setupSkipped` on stale ones (the trusted half still needs a human on the trust dialog).

openwong2kim:
Dogfooded this live on Windows (approved the prompt, real worktrees). The trust half works exactly as described, but **the port never reaches the pane**.

## What worked

```
tasks[0] probe-a  port 3200  worktree …\probe-a-kb6u97tf  branch wtask/probe-a-kb6u97tf
tasks[1] probe-b  port 3201  worktree …\probe-b-g81m12do  branch wtask/probe-b-g81m12do
setupSkipped: "untrusted"
```

Distinct ports assigned from `"3200-3210"`, two worktrees, two branches. And the fail-closed half is right: with the wmux.json untrusted, `setup-ran.txt` is **absent in both worktrees** — the hook did not run, and the result says why.

## What doesn't

`WMUX_TASK_PORT` is not in either task pane's environment. Read back off the live sessions:

```
daemon-24a57ee1  WMUX_TASK_PORT=(absent)  cwd=probe-a-kb6u97tf
daemon-e40b5325  WMUX_TASK_PORT=(absent)  cwd=probe-b-g81m12do
```

Not a truncated view — the env has 109 keys. The only WMUX_* keys present are `WORKSPACE_ID`, `WORKSPACE_NAME`, `MEMBER_ID`, `PTY_ID`, `DATA_SUFFIX`, `SHELL_INTEGRATION`, i.e. exactly the ones `resolveSpawnEnv` force-injects after its strip.

The cause is in `resolveSpawnEnv` (`src/main/pty/resolveSpawnEnv.ts`), which the renderer's `pty.create` goes through:

```js
// baseline strip
for (const key of Object.keys(env)) {
  if (key.toUpperCase().startsWith('WMUX_')) delete env[key];
}
…
applyOverlay(env, profileEnv);   // profileEnv is the fan-out taskEnv
```

and `applyOverlay` skips the reserved namespace by design:

```js
if (key.toUpperCase().startsWith('WMUX_')) continue;
```

That skip is deliberate — it is what makes "a profile can never spoof pane identity" unconditional rather than profile-only. So any `WMUX_*` key handed in as caller env is dropped on the floor, and `WMUX_TASK_PORT` is one.

The renderer side is fine (`env: taskEnv` is attached to `pty.create`, and `withWorkspaceProfile` preserves it) — it dies one layer below.

## Why the tests pass

`FanOutService.test.ts` asserts the port "reaches spawn env", but that is the env FanOutService *hands over*. Nothing exercises the `resolveSpawnEnv` hop where the reserved-prefix filter lives, so the gap sits precisely between the two.

## Options

1. Force it alongside identity in `pty.handler`'s `identity` map (applied after the overlay). Keeps the documented `WMUX_TASK_PORT` name, and is honest about what it is: something the app forces, not something a caller supplies.
2. Rename it outside the reserved prefix (e.g. `TASK_PORT` / `PORT`). Cheaper, but the name is the contract with the docs and the how-to, and `PORT` collides with what dev servers already read.

I'd go with (1). Either way it wants a test at the `resolveSpawnEnv` level, since that is the layer that ate it.

Everything else here I'd happily merge — holding only on this.

openwong2kim:
Fixed and re-verified live. Both halves now hold.

**Port actually reaches the pane** — read off the live sessions after the fix:

```
daemon-c59b0225  WMUX_TASK_PORT=3200  cwd=…\worktrees\…\port-a-mo8wsd54
daemon-2c13e139  WMUX_TASK_PORT=3201  cwd=…\worktrees\…\port-b-4tk7lco8
other panes carrying WMUX_TASK_PORT: 0
```

Distinct, from the declared `3200-3210`, and confined to the fan-out — nothing leaked into the rest of the fleet.

**Setup hook, trusted path** — trusted the wmux.json, ran a third fan-out:

```
hook-a  port=3202  worktree=hook-a-xbxrxepv
  setup-ran.txt: PRESENT
  contents: SETUP-HOOK-RAN
  setupSkipped = undefined
```

**Setup hook, untrusted path** (earlier run, same repo before trusting):

```
setupSkipped: "untrusted"
setup-ran.txt: absent in both worktrees
```

So the hook runs on trusted bytes and fail-closes otherwise, exactly as the description says, and the port allocator hands out a fresh one per task across runs (3200/3201, then 3202).

The fix forces the port alongside the other identity stamps rather than letting it ride the caller env overlay, since `resolveSpawnEnv` strips the reserved `WMUX_*` namespace and `applyOverlay` skips those keys by design. Digits-only, so the exception can't carry an arbitrary value. Three tests added at the `resolveSpawnEnv` layer — the hop nothing covered before.

PR #827, opened 2026-08-07
feat(daemon): outbound webhook/ntfy notification sink

openwong2kim:
## What

An outbound-only notification path for users without the paired phone app: the daemon POSTs a minimal JSON payload to user-configured sinks (`{type: "webhook"|"ntfy", url, events?}`, max 8) when an approval is raised or an agent turn ends. Off unless configured; `WMUX_NOTIFY_SINKS=0` kill switch. No new listening surface — the loopback-only default is untouched.

## Payload policy (the important part)

Hard allowlist (`NOTIFY_PAYLOAD_FIELDS`): event kind, static title, agent name, 8-char pane/workspace prefixes, re-derived `risk` boolean, timestamp, fresh random id. **No agent-authored text travels** — not the approval question, not tool names, not paths. The APNs path may quote them because its envelope is sealed to the device; a webhook body lands in third-party logs (a public ntfy topic being the common case). `risk` is re-derived via `hasElevatedRisk` so this path can never disagree with the push path about what is dangerous.

Delivery is fire-and-forget: queue cap 32, drop-oldest, 5s timeout, no retry, manual redirect, http(s) only. Sink URLs are never logged (an ntfy topic is itself the secret). Turn-end taps fire on `agent.stop` only — not `subagent_stop`, which would ping a dozen times per turn.

## Tests

12 added: payload allowlist (negative assertions that question text, commands, paths, full ids never serialize), queue cap/drop-oldest behavior, config coercion rejections.

## Notes

- Changelog fragment: `changelog.d/t3-webhook-sink.md`.
- No CLI surface yet (config.json editing only); attention pings carry no workspace id (would need new plumbing) — follow-ups.

## Prior art

Concept-level borrowing only, no code imported: outbound ntfy/webhook notification for blocked agents follows the pattern of community Claude Code ntfy hooks. Payload policy and delivery semantics are wmux's own (modeled on this repo's PushSender).

### DIFF
--- a/changelog.d/t3-webhook-sink.md
+++ b/changelog.d/t3-webhook-sink.md
@@ -1,0 +1,24 @@
+### Added
+
+- **Webhook and ntfy notifications, no phone app required.** Point
+  `notifySinks` in `~/.wmux/config.json` at a webhook URL or an ntfy topic and
+  the daemon pings it when an agent asks for an approval or finishes a turn.
+  Until now the only way to hear about a blocked agent from away from the
+  keyboard was the iOS app and its push relay; this is the plain-HTTP path for
+  everyone else. Off unless you configure it, outbound only — the daemon opens
+  no new port — and `WMUX_NOTIFY_SINKS=0` turns it off without editing config.
+
+  ```json
+  "notifySinks": [
+    { "type": "ntfy", "url": "https://ntfy.sh/my-topic", "events": ["approval"] },
+    { "type": "webhook", "url": "https://hooks.example/wmux" }
+  ]
+  ```
+
+  The ping is deliberately thin: the event kind, a fixed title, the agent name,
+  short pane and workspace id prefixes, a derived risk tier and a timestamp. It
+  never carries the agent's question, tool input, terminal output, file paths or
+  any id that can address a pane — the destination is a server you chose but the
+  body travels in the clear, and a shared ntfy topic is not the place for your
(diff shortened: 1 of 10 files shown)


openwong2kim:

This PR adds an outbound-only notification path for users without the paired phone app: the daemon POSTs a minimal JSON payload to user-configured webhook or ntfy sinks when an approval is raised or an agent turn ends. It is off unless configured (`notifySinks` array, max 8) with a `WMUX_NOTIFY_SINKS=0` kill switch, opens no new listening surface, and keeps the loopback-only default intact. Delivery is fire-and-forget — queue cap 32 with drop-oldest, 5s timeout, no retry, manual redirects — matching `PushSender`'s shape.

The payload policy is the part worth flagging because a webhook body lands in third-party logs (a shared ntfy topic being the common case): a hard allowlist (`NOTIFY_PAYLOAD_FIELDS`) carries only the event kind, a static title, agent name, 8-char prefixes, a re-derived `risk` boolean, and a timestamp. No agent-authored text travels — not the approval question, not tool names, not paths, not full ids — and `risk` is re-derived via `hasElevatedRisk` so the webhook path can never disagree with the APNs path about what is dangerous. Sink URLs are never logged, since an ntfy topic is itself the secret. The 12 tests include negative assertions that question text and commands never serialize. Priority **P2** — closes a real competitor gap (phone-less alerting) with a careful security posture, but it is a new outbound-network surface and deserves dogfooding on the delivery/queue-drop behaviour before release.

PR #828, opened 2026-08-07
feat(push): suppress phone push while the desktop app is present

openwong2kim:
## What

Approval pushes currently fire even when the user is sitting in front of the desktop app with the inbox open. This adds presence-based suppression: when the desktop client is connected AND a window was focused within the last 90s (configurable), the APNs push for an approval is skipped. The event still reaches the in-app inbox — only the phone ping is suppressed.

## Semantics

- **Signal**: transition-based, no heartbeat. Electron main reports `browser-window-focus`/`blur` over a new tiny RPC (`daemon.presence.desktop`, `{focused: boolean}`), plus a one-shot sync when a daemon connection is established (a fresh daemon has heard no transitions).
- **Present** = connected && focused && report age within `staleAfterMs` (default 90s).
- **Fail-open toward notifying**: unreported, blurred, stale, clock-skewed, garbage config, or headless daemon → the push is sent. Suppression requires a positively fresh signal. Pipe close reclaims only that client's report (a late close from a replaced connection is ignored).
- **Critical carve-out**: `risk === 'critical'` payloads always push, present or not (`CRITICAL_RISK_IGNORES_PRESENCE` named constant).
- Decision lives in one pure module (`src/daemon/push/presence.ts`); `PushSender` internals untouched, so other outbound sinks can adopt the same predicate.

## Config

`pushPresenceSuppression: { enabled: true, staleAfterMs: 90000 }` with per-field backfill; garbage values restore defaults without touching sibling fields.

## Tests

17 added: predicate boundary cases (exactly 90s vs +1ms, stale, critical, disabled, headless, future-clock, NaN window), tracker ownership/blur reclaim, main-side reporter transitions and no-op paths.

## Notes

- Changelog fragment: `changelog.d/t4-presence-suppression.md`.
- `attention` events have no APNs call site today; when one appears it should consult the same predicate.

## Prior art

Concept-level borrowing only, no code imported: presence-based push suppression is the pattern Claude Code remote control uses (skip push while the desktop client is active). Signal design (transition-based focus RPC, fail-open, critical carve-out) is wmux's own.

## Summary by CodeRabbit

* **New Features**
  * Push notifications for non-critical approvals are now suppressed while the desktop app is recently focused.
  * Suppressed notifications remain available in the in-app inbox and are delivered when presence ends or becomes stale.
  * Critical-risk notifications and uncertain presence states continue to trigger pushes immediately.
  * Presence detection responds to focus, blur, lock, suspend, unlock, and resume events.
  * Added configurable suppression with a default 90-second freshness window.

* **Documentation**
  * Documented desktop presence-based push-notification suppression behavior.

### DIFF
--- a/changelog.d/t4-presence-suppression.md
+++ b/changelog.d/t4-presence-suppression.md
@@ -1,0 +1,16 @@
+### Added
+
+- **Your phone stops buzzing for approvals you are already looking at.** When
+  the desktop app is connected and its window was focused within the last 90
+  seconds, the push notification for a new approval is skipped — the request
+  still lands in the in-app inbox, which is the thing you were reading anyway.
+  Before this, sitting at the desk answering prompts also meant a phone
+  lighting up for every one of them.
+
+  Every uncertainty still sends: a daemon with no desktop attached, a version
+  of the app too old to report focus, or a focus report that has gone stale all
+  fall back to notifying. A `critical`-risk approval is pushed even when you are
+  present — a destructive action is worth the second channel.
+
+  Configurable via `pushPresenceSuppression` in `~/.wmux/config.json`
+  (`enabled`, default `true`; `staleAfterMs`, default `90000`).
--- a/changelog.d/t4-presence-suppression.md
+++ b/changelog.d/t4-presence-suppression.md
@@ -2,15 +2,26 @@
 
 - **Your phone stops buzzing for approvals you are already looking at.** When
   the desktop app is connected and its window was focused within the last 90
-  seconds, the push notification for a new approval is skipped — the request
+  seconds, the push notification for a new approval is held back — the request
   still lands in the in-app inbox, which is the thing you were reading anyway.
   Before this, sitting at the desk answering prompts also meant a phone
   lighting up for every one of them.
 
+  Held, not dropped. The moment you leave — the window blurs, you lock the
+  screen, the machine sleeps, the app quits, or the focus report simply ages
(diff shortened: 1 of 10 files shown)


openwong2kim:

This PR suppresses the APNs push for an approval when the user is already sitting in front of the desktop app: the push is skipped when the desktop is connected AND its window was focused within the last 90 seconds (configurable), while the approval still reaches the in-app inbox — only the phone ping is suppressed. The signal is transition-based (no heartbeat): Electron main reports `browser-window-focus`/`blur` over a new `daemon.presence.desktop` RPC with a one-shot sync on daemon connect, and presence fails open toward notifying on every uncertainty (unreported, blurred, stale report, clock skew, garbage config, headless daemon all send the push). A critical-risk approval always pushes regardless of presence.

The architecture here is the notable part. The `shouldSuppressPush` predicate lives in a standalone `presence.ts` module rather than inside `PushSender`, precisely because the question "is the human present?" is independent of transport — so #827's webhook sink and any future sink can ask the same predicate. The `CRITICAL_RISK_IGNORES_PRESENCE` carve-out is a deliberate exception, not an oversight: suppression trades a buzz for quiet, and that trade is worth making only for routine prompts. The 17 tests cover predicate boundary cases (exactly 90s vs +1ms, stale, critical, disabled, headless, future-clock, NaN window), tracker ownership/blur reclaim, and main-side reporter transitions. Priority **P2** — quality-of-life improvement that reduces notification noise without touching PushSender internals, well-scoped and fail-safe by construction.

PR #829, opened 2026-08-07
fix(gate): let a bypassPermissions session through the permission gate

openwong2kim:
## The bug

The #783 permission gate opens a remote approval record for every gated tool —
`Bash`, `Write`, `Edit`, `MultiEdit`, `NotebookEdit`, `Task`, `KillShell` — and
holds the hook's RPC open until a phone answers or the deadline fires.

It did that **regardless of the session's own permission mode**. A session
launched with `--dangerously-skip-permissions` has explicitly opted out of being
asked, yet it still stalled on every `Bash` call and then fell back to exactly
the local prompt the user opted out of. With no phone attached, that is a
deadline-length pause on *every* gated tool call.

From the user's seat this reads as **"I launched in bypass mode and it asks for
approval on every tool call."** `gatedTools` defaults to the full list, and
`wmux setup-hooks` installs the wide PreToolUse matcher unconditionally, so any
user who ran `setup-hooks` and launched in bypass mode hits this.

## The fix

Claude Code stamps the live mode on every PreToolUse payload as
`permission_mode` (verified against a real payload: `permission_mode:
"bypassPermissions"`), and the bridge already forwards the payload verbatim. So
the daemon can honour it with **no bridge change and no config migration**.

A `bypassPermissions` signal now skips the gate and broadcasts
`agent.tool_started` like any non-gated tool, so the phone's liveness header is
unchanged.

Only that one mode passes through. `acceptEdits` still prompts for `Bash`, so
its gate stays meaningful, as do `plan`, `default`, and an absent mode.

## Tests

Two cases added to `HookIngest.test.ts`, in a block that injects a gate config
that actually gates `Bash` (the shared fixture leaves `gateConfig` unset, which
would make the assertions vacuous):

- a `bypassPermissions` signal opens no gate record, returns no `gateId`, and
  still emits `agent.tool_started`
- `acceptEdits`, `default`, `plan`, and an absent mode each still open exactly
  one gate

Verified these fail without the fix: reverting the `!bypassing &&` guard fails
exactly one test, restoring it passes all 62 in the file.

`tsc -p tsconfig.daemon.json` clean. Full `test:parallel` run: the only failures
are in `notificationSlice.test.ts`, which passes 23/23 on its own — a
parallel-run flake unrelated to this change (nothing here touches the renderer).

### DIFF
--- a/changelog.d/829.md
+++ b/changelog.d/829.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- **A bypass-mode session no longer stops at the permission gate.** An agent
+  launched with `--dangerously-skip-permissions` has already declared that it
+  should never be asked, but the remote permission gate opened an approval
+  record for it anyway — on every `Bash`, `Write`, `Edit`, `MultiEdit`,
+  `NotebookEdit`, `Task`, and `KillShell` call. With no phone attached, each of
+  those calls paused for the full gate deadline and then fell back to exactly
+  the local prompt the user had opted out of, which read as "bypass mode asks
+  for approval on every tool call". The gate now reads the session's live
+  permission mode and lets a bypass session straight through, while the pane's
+  liveness header keeps updating as before. Sessions that do still prompt —
+  `acceptEdits`, `plan`, and the default mode — keep their gates unchanged.
+  (#829)
--- a/src/daemon/hooks/HookIngest.ts
+++ b/src/daemon/hooks/HookIngest.ts
@@ -511,10 +511,22 @@ export class HookIngest {
       ? signal.payload.tool_name
       : null;
     const gatedTools = this.deps.gateConfig?.().gatedTools ?? [];
-    const isGated = toolName !== null && gatedTools.includes(toolName);
+    // A `bypassPermissions` session has already declared "never ask me" — the
+    // user launched it with `--dangerously-skip-permissions` (or set
+    // `permissions.defaultMode`). Gating it anyway re-asks the exact question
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR fixes a real footgun in the #783 permission gate: a session launched with `--dangerously-skip-permissions` was still being stopped on every gated tool call (`Bash`, `Write`, `Edit`, …) because the gate ignored the session's own permission mode and opened an approval record unconditionally. With no phone attached that meant a deadline-length stall on *every* gated call followed by a fallback to exactly the local prompt the user had opted out of — which reads to the user as "bypass mode asks for approval on every tool." The fix is a one-line guard in `HookIngest.handlePermissionGate` that reads the `permission_mode` Claude Code already stamps on each PreToolUse payload and lets only `bypassPermissions` pass through (no bridge change, no config migration), while `acceptEdits`/`plan`/`default`/absent keep their gates.

The change is appropriately narrow and well-tested: the two added cases in `HookIngest.test.ts` inject a gate config that actually gates `Bash` (the shared fixture would otherwise make the assertions vacuous) and the author verified they fail without the `!bypassing &&` guard. Scope is correct too — it touches only the gate decision path, still broadcasts `agent.tool_started` so the phone liveness header is unchanged, and explicitly preserves the gate for modes that still prompt.

Priority is **P2**: it is a correctness defect in a shipped feature (bypass mode is documented and the wide `setup-hooks` matcher means any user who ran it hits this), but it is not data loss or a crash. Worth merging ahead of the surrounding remote work since it stands alone.

PR #830, opened 2026-08-07
fix(remote): make /api/stream snapshots mode-safe, resize-aware, and query-silent

openwong2kim:
Attaching a remote workspace to a long-running fullscreen TUI painted a garbled, interleaved screen: rows landed on top of scrollback, fragments of earlier frames stayed on the left edge, and the app never took over the screen.

Three independent defects on the `/api/stream` path, all of which also affect the phone web frontend — this is not mirror-specific.

## 1. The snapshot window was byte-safe but not mode-safe

`capSnapshot` sends the last 256 KB of the ring and cuts only at UTF-8 and escape-sequence boundaries. Terminal MODE state is not a boundary. A fullscreen app sends `ESC[?1049h` exactly once, at startup, then paints with absolute cursor positioning forever, so hours later that switch sits far outside the window. The client replayed absolute-positioned frames while still on the normal buffer, which is precisely the interleaved-over-scrollback symptom.

The local GUI never hits this because `HeadlessSnapshot` refuses alt-screen and `SessionPipe` falls back to a full raw replay. The SSE path cannot afford that — the window exists to keep a reconnecting phone from pulling megabytes — so it now reconstructs the mode state instead.

`OutputModeTracker` is fed every chunk written to the ring (O(chunk) at write time) and answers in O(1) at stream open, preserving `snapshotWindow.ts`'s invariant that per-call work is independent of buffer size. Its `preamble()` is prepended to the capped window, riding the existing `snapshot` event rather than a new event name, so a cached frontend replays it as ordinary bytes — which is what it is.

Accepted limitation, documented at the source: an alt-screen app idle since before the window contributes no repaint of its own, so only the part of its frame the window happens to contain shows until it next redraws. Mostly-right beats garbled.

## 2. Geometry changes never reached an open stream

`meta` was sent once per connection. After a remote-side resize every subsequent absolute-positioned frame was computed for a grid the client no longer had. Applied resizes now emit from `noteResize()` — one event covering both the phone route and the desk RPC — and each stream answers with a fresh `meta` + `snapshot` PAIR, trailing-debounced 150 ms so a drag-resize costs one repaint. The pair matters: the Electron mirror client holds a `meta` until a `snapshot` follows and never dispatches a lone one.

## 3. The mirror answered device queries on the remote's stdin

xterm replies to DA1, DSR/CPR and DECRPM itself and delivers those replies through the same `onData` the mirror forwards to `paneWrite`. Replaying a snapshot that contained such queries made the mirror type phantom responses into the remote pane's shell. The remote's own GUI terminal is the authoritative responder — the same reason `HeadlessSnapshot` wires no `onData` at all. `paneWrite` is now gated while a repaint is being parsed.

## Tests

215 passing across the touched suites.

- `outputModeTracker.test.ts` (new) — mode toggles, `h`/`l` pairs, last-writer-wins, sequences split across chunk boundaries, mouse encoding pairs, RIS.
- `WebTerminalServer.test.ts` — a ring holding `ESC[?1049h` followed by more than 256 KB of output yields a snapshot that opens with the alt-screen preamble; a normal-buffer session gets no preamble at all (byte-identical to before); an applied resize produces a debounced `meta`+`snapshot` pair on an open stream.
- `RemoteMirrorTerminal.test.tsx` — parser output during a gated repaint does not reach `paneWrite`.

### DIFF
--- a/changelog.d/830.md
+++ b/changelog.d/830.md
@@ -1,0 +1,24 @@
+### Fixed
+
+- **Mirrored remote panes running a fullscreen app no longer paint a garbled
+  screen.** Attaching to a pane that had been running vim, htop or an agent CLI
+  for a while showed rows interleaved over old scrollback, with fragments of
+  earlier frames stuck along the left edge, and the app never took over the
+  screen. The initial paint is capped to a window of recent output, and a
+  fullscreen app announces itself exactly once, at startup — so hours later that
+  announcement was outside the window and the terminal was still on the ordinary
+  buffer while the bytes assumed the fullscreen one. The daemon now tracks which
+  modes a pane's output has switched on and restores them ahead of the paint.
+  This affects phones and tablets opening the web terminal too, not only
+  mirrored workspaces.
+
+- **Resizing a pane no longer breaks every viewer watching it.** Geometry was
+  sent once when a viewer connected, so after a resize on the machine that owns
+  the pane, everything drawn afterwards was positioned for a grid the viewer no
+  longer had. Viewers are now repainted at the new size once the resize settles.
+
+- **A mirrored pane no longer types into the remote shell by itself.** Terminals
+  answer certain queries automatically, and a mirror was sending its answers on
+  to the remote pane's input — so replaying scrollback that contained such a
+  query injected stray characters into a live shell on the other machine. The
+  machine that owns the pane is the only one that answers now.
--- a/changelog.d/830.md
+++ b/changelog.d/830.md
@@ -8,17 +8,21 @@
   fullscreen app announces itself exactly once, at startup — so hours later that
   announcement was outside the window and the terminal was still on the ordinary
   buffer while the bytes assumed the fullscreen one. The daemon now tracks which
-  modes a pane's output has switched on and restores them ahead of the paint.
-  This affects phones and tablets opening the web terminal too, not only
-  mirrored workspaces.
+  modes a pane's output has switched on and restores them ahead of the paint,
+  including for panes restored after a restart. This affects phones and tablets
+  opening the web terminal too, not only mirrored workspaces.
 
 - **Resizing a pane no longer breaks every viewer watching it.** Geometry was
   sent once when a viewer connected, so after a resize on the machine that owns
   the pane, everything drawn afterwards was positioned for a grid the viewer no
-  longer had. Viewers are now repainted at the new size once the resize settles.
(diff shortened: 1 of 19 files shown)


openwong2kim:

This PR fixes three independent defects on the `/api/stream` SSE path, all of which surfaced as a garbled/interleaved screen when attaching a remote mirror to a pane running a long-lived fullscreen TUI — and all of which also affect the phone/tablet web frontend, since they share this path. The headline one is that `capSnapshot` is byte-safe but not *mode-safe*: it sends the last 256 KB of the ring cut at UTF-8/escape boundaries, but a fullscreen app emits `ESC[?1049h` exactly once at startup, so hours later that switch is far outside the window and the client replays absolute-positioned frames onto the normal buffer (interleaved with scrollback). The fix introduces `OutputModeTracker` — fed every chunk written to the ring (O(chunk) at write time, O(1) at stream open, preserving snapshotWindow's size-independent invariant) — whose `preamble()` is prepended to the capped window. The local GUI never hit this because `HeadlessSnapshot` refuses alt-screen and falls back to a full replay, which the SSE path cannot afford.

The other two fixes are solid and well-scoped: (2) geometry changes now propagate to open streams — `noteResize()` emits a `resize` event answered with a trailing-debounced (150 ms) `meta`+`snapshot` *pair* (the pair matters: the Electron mirror client never delivers a lone `meta`), so a drag-resize costs one repaint; (3) the mirror no longer answers DA1/DSR/CPR device queries on the remote's stdin, gated by a `repaintingRef` that suppresses `paneWrite` for the lifetime of a `term.write` callback (the same reason `HeadlessSnapshot` wires no `onData`). Tests are thorough — 130-line `outputModeTracker.test.ts` covering chunk-boundary splits and multi-mode params, `WebTerminalServer.test.ts` asserting the alt-screen preamble opens the payload while a plain session is byte-identical to before, and `RemoteMirrorTerminal.test.tsx` proving a parser reply during a gated repaint never reaches `paneWrite`.

Priority is **P2**. It is a correctness defect in a shipped, user-facing path (remote mirror + phone web terminal), with a clean, well-reasoned fix and no migration cost. The accepted limitation (an alt-screen app idle since before the window contributes no repaint until it next redraws) is the right call — mostly-right beats garbled. Not P1 because it is a display-corruption issue, not data loss or a crash, and the local GUI is unaffected.

PR #831, opened 2026-08-07
fix(remote): restore attachments across reload and keep mirrored panes live

openwong2kim:
Two lifecycle defects in attached remote workspaces, both client-side only. No daemon change, so this keeps working against older remote builds — version skew is real for this feature.

## 1. Cmd+R and app restart silently dropped every attachment

`remoteWorkspacesSlice` is memory-only by design, so a reload recreated it empty and the Remote section vanished from the sidebar. The SSE teardown in `installSenderCleanup` is correct and untouched — stale streams must die on reload. What was missing is that the attach DESCRIPTORS died with them.

Descriptors now live in the main process, in `~/.wmux/remote-attachments.json`, a sibling of `RemoteHostsStore`. Deliberately a separate file: a descriptor holds no credential (the bearer token stays in `remote-hosts.json`), so it uses the ordinary atomic-JSON path rather than `secureWriteTokenFile`, and keeping them apart makes that difference structural instead of a convention to remember. Only `{key, hostId, hostLabel, workspaceId, name}` is stored, and `toDescriptor()` strips structurally, so a caller handing over a full `AttachedRemoteWorkspace` cannot get its pane list onto disk. Removing a host cascades.

`useRemoteAttachmentsLifecycle` replays the descriptors on boot, fetching each host's CURRENT workspaces — one call per host, not per workspace — so panes always come from the live host and never from disk. A host that cannot be reached, or a workspace that no longer exists there, is restored in a `stale` state rather than dropped: a sleeping laptop must not delete the user's attachment. Restore adds without selecting, so it never steals the view on startup.

## 2. Panes opened or closed on the remote never showed up

The pane list was a snapshot taken at attach time. A pane closed on the remote left a dead tile forever; a pane opened there was invisible until a manual re-attach.

The wire carries no topology events, but a close does surface indirectly — the PTY dies, which produces an SSE `exit` — so that is used as a refresh trigger, debounced 400 ms. An open produces nothing at all, which is what the 10 s poll is for; it is armed only while something is attached, so an app with no mirrors makes no periodic requests. Both paths run the same refetch-and-diff and are serialised, so an exit burst plus a poll tick collapse into one round. `mergePaneSets` keeps survivors in their slot and appends new panes, so the grid never reshuffles, and an identical fetch keeps the old array identity so the poll does not re-render mirrors forever.

## Tests

186 passing across the touched suites.

- `RemoteAttachmentsStore.test.ts` (new) — roundtrip across a fresh construct (the restart path), never persists a pane list even when handed one, key dedup, remove, host cascade, corrupt file, foreign shape, missing file.
- `remote.handler.test.ts` — add/list roundtrip, pane-list stripping, refusal for an unregistered host, write-failure returns false rather than rejecting, host-removal cascade, and a renderer reload tears down live SSE attaches but keeps the descriptors.
- `remoteWorkspacesSlice.test.ts` — `mergePaneSets` removal, addition, reorder stability, new-pane placement, field freshness; no-churn identity; stale cleared on success and never dropping a row; restore does not select and dedups.
- `useRemoteAttachmentsLifecycle.dynamic.test.tsx` (new) — restore with fresh panes, stale on unreachable host, stale on vanished workspace, no request when nothing is persisted; exit drops the closed pane, an exit burst is one fetch, unsubscribe on unmount; no poll while unattached, the poll picks up a remotely-opened pane, the poll stops after detach, a failed poll marks stale without dropping.

## Summary by CodeRabbit

* **New Features**
  * Remote workspace attachments now persist across reloads and application restarts.
  * Mirrored workspaces and panes synchronize automatically with remote hosts.
  * Disconnected or unavailable workspaces remain visible and reconnect when recovered.
  * Removing a remote host also removes its associated workspace mirrors.

* **Bug Fixes**
  * Improved handling of malformed or unavailable remote workspace data.
  * Prevented duplicate panes and improved attachment and detachment reliability.
  * Added clearer disconnected-state indicators for stale workspaces.

### DIFF
--- a/changelog.d/831.md
+++ b/changelog.d/831.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **Attached remote workspaces survive a reload and a restart.** Pressing Cmd+R,
+  or quitting and reopening wmux, silently emptied the Remote section of the
+  sidebar and every mirror with it — re-attaching meant walking back through the
+  picker each time. Attachments are now remembered, and are restored on launch
+  with a fresh pane list fetched from the host. A host that is asleep or
+  unreachable keeps its row, marked as stale, rather than being deleted;
+  detaching is still yours to decide.
+
+- **Mirrors follow panes opened and closed on the remote machine.** The pane
+  layout was frozen at the moment you attached: a pane closed on the other
+  machine left a dead tile in the grid forever, and a pane opened there was
+  invisible until you detached and attached again. The grid now follows the
+  remote workspace, with surviving panes staying in place instead of shuffling
+  when one comes or goes.
--- a/changelog.d/831.md
+++ b/changelog.d/831.md
@@ -4,9 +4,11 @@
   or quitting and reopening wmux, silently emptied the Remote section of the
   sidebar and every mirror with it — re-attaching meant walking back through the
   picker each time. Attachments are now remembered, and are restored on launch
-  with a fresh pane list fetched from the host. A host that is asleep or
-  unreachable keeps its row, marked as stale, rather than being deleted;
-  detaching is still yours to decide.
+  with a fresh pane list fetched from the host. Every row appears immediately,
+  even when a host takes its time answering. A host that is asleep or
+  unreachable keeps its row, marked as stale, rather than being deleted, and
+  its mirrors reconnect on their own once it comes back; detaching is still
+  yours to decide.
 
 - **Mirrors follow panes opened and closed on the remote machine.** The pane
   layout was frozen at the moment you attached: a pane closed on the other
@@ -14,3 +16,8 @@
   invisible until you detached and attached again. The grid now follows the
   remote workspace, with surviving panes staying in place instead of shuffling
   when one comes or goes.
+
+- **Removing a remote host clears its mirrors.** The host disappeared from the
+  picker but its mirrored workspaces stayed in the sidebar for the rest of the
+  session, quietly failing to refresh against a machine wmux no longer knew
+  about. They now go with it.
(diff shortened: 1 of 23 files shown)


openwong2kim:

This PR closes two lifecycle gaps in attached remote workspaces, both client-side and requiring no daemon change — so it keeps working against older remote builds, which matters for this feature. The first is that `remoteWorkspacesSlice` was memory-only by design, so a Cmd+R reload or an app restart silently emptied the Remote section of the sidebar and wiped every mirror with it; the fix persists only the attach *descriptors* (`{key, hostId, hostLabel, workspaceId, name}` — deliberately a separate file from `remote-hosts.json` and stripped structurally so no credential or pane list can land on disk) in a new `RemoteAttachmentsStore`, and `useRemoteAttachmentsLifecycle` replays them on boot by re-fetching each host's *current* panes (one call per host, not per workspace). A host that is asleep or unreachable, or a workspace that has vanished there, restores in a `stale` state rather than being dropped — a sleeping laptop must not delete the user's attachment — and restore adds without selecting, so it never steals the view on startup.

The second fix addresses pane membership: the pane list was frozen at attach time, so a pane closed on the remote left a dead tile forever and a pane opened there was invisible until a manual re-attach. Since the wire carries no topology events, the close is caught indirectly (PTY death → SSE exit → debounced 400 ms refresh) and the open — which produces nothing at all — is caught by a 10 s safety-net poll armed *only while something is attached*, so an app with no mirrors makes no periodic requests. `mergePaneSets` keeps survivors in their slot and appends new panes, so the grid never reshuffles, and an identical fetch keeps the old array identity so the poll does not re-render mirrors forever.

Priority is **P2**. It fixes a genuine data-loss-on-reload UX defect in a shipped feature (attachments silently disappearing is the kind of thing that erodes trust in the remote-mirror product), the design decisions are well-reasoned and defensively scoped, and the test coverage is strong: 186 passing across the touched suites, including the new `RemoteAttachmentsStore.test.ts` (roundtrip, pane-list stripping, host cascade, corrupt/foreign/missing file), the new `useRemoteAttachmentsLifecycle.dynamic.test.tsx` (stale-on-unreachable, exit-burst collapse, poll lifecycle), and `remoteWorkspacesSlice.test.ts` additions for `mergePaneSets` stability and restore semantics.

PR #832, opened 2026-08-07
fix(deck): stop-gate decision release, cap-out hysteresis, and active-work injection diet

openwong2kim:
## Problem

A 90-minute supervision loop (transcript, 2026-08-07) exposed three compounding token sinks in the Command Deck orchestrator:

1. **Gate vs. waiting-on-human**: the Stop gate refused to end a turn while active work was open, but the brain was waiting on a human answer — its only remaining move was re-printing the same question, which it did 4x per cycle (3 refusals + cap).
2. **Per-turn cap, per-minute loop**: the consecutive-block counter resets every turn, so a wake loop over *unchanged* fleet state re-bought the same three refusals every cycle.
3. **Full re-injection**: the complete [active-work] block (objective + every follow-up + A2A rows + imperatives) was re-typed onto the visible TUI on every turn, and the follow-up dedupe was last-item-only, so re-sent instructions accumulated quadratically.

## Changes

- **stopGate rule 4** — a PENDING deck decision releases the gate. Mirror image of the coalescer's existing wake suppression under a pending decision; the resolve path already kicks an explicit resume turn.
- **stopGate rule 5** — cap-out hysteresis. The capping allow carries a fingerprint of what was held on (`work.id@updatedAt | sorted pane:status`); the handler records it **only when that exact state was actually refused**, and the gate stays quiet on identical state until it changes, a human acts (DECK_SEND / TUI prompt / Wake / decision resolve), or a 15-min TTL expires.
- **Refusal strings** — every refusal now forbids re-sending the previous message and names deck_ask_decision as the legitimate way to wait (it releases the gate).
- **followUps dedupe** — whitespace-normalized compare against the objective and *every* retained follow-up, append-time only (stored records are never rewritten; the completeWork CAS is untouched).
- **[active-work] injection diet** (terminal brain only) — a byte-identical block collapses to a one-line reminder that keeps the id and the two imperatives (still ACTIVE, deck_complete_work). Settled like the ambient blocks (only a clean turn marks it shown); PARKED records always render in full; a TUI conversation reset (foreign session id change) forgets the memory so the full block re-sends.

## Verification

- `tsc --noEmit` clean; `npx vitest run src/main/deck src/main/ipc/handlers`: **954 passed**; full `test:parallel`: 10,387 passed (1 pre-existing environment-dependent failure in `bridgePermissionGate.test.ts`, fails identically at HEAD).
- New/extended tests: stopGate (rule 4/5, fingerprint, refusal strings), stopGateState (cap-out TTL/clear), deckWorkStore (non-adjacent + whitespace-variant dedupe, reminder line), and `deck.handler.activeWorkDiet.test.ts` (full → reminder → full-on-change → reminder re-arm).
- An adversarial multi-agent review of the diff confirmed 3 defects, all fixed in this PR: decision-resolve/Wake not re-arming a capped-out gate; the diet memory surviving a TUI conversation reset; a cap-out suppressing a state no refusal ever named.

## Summary by CodeRabbit

* **Bug Fixes**
  * Pending decisions now correctly release the stop gate.
  * Repeated refusal messages are suppressed until work or state changes.
  * Follow-up instructions are deduplicated, including whitespace variations and repeats of the original objective.
  * Stop-gate suppression expires safely and resets after relevant actions or state changes.

* **Improvements**
  * Active work is shown compactly after its initial display, expanding again when details change or conversations reset.
  * Parked work continues to display complete details.

### DIFF
--- a/changelog.d/832.md
+++ b/changelog.d/832.md
@@ -1,0 +1,22 @@
+### Fixed
+
+- **The orchestrator no longer re-prints the same message while waiting on you.** A
+  pending decision (deck_ask_decision) now releases the Stop gate — previously the
+  gate kept refusing the turn while the decision block forbade acting, leaving the
+  brain nothing to do but restate its question, up to four times per wake cycle. (#832)
+- **A wake loop over unchanged fleet state no longer re-buys the same refusals every
+  cycle.** When the Stop gate caps out, it remembers exactly what it was holding on
+  and stays quiet until that state actually changes, you act (a message, the Wake
+  button, answering a decision), or a 15-minute reminder interval passes. Only a
+  state the gate actually refused can be suppressed. (#832)
+- **Re-sent instructions no longer pile up in the active-work record.** Follow-up
+  dedupe now checks the objective and every retained follow-up with
+  whitespace-insensitive comparison, instead of only the most recent entry. (#832)
+
+### Changed
+
+- **The terminal brain's [active-work] block goes on a diet.** An unchanged block
+  collapses to a one-line reminder (id + the finish-via-deck_complete_work contract)
+  instead of re-typing the full objective, follow-ups, and task list onto the
+  terminal every turn. Any change to the record — or a conversation reset in the
+  TUI — re-sends the block in full. (#832)
--- a/src/main/deck/__tests__/deckWorkStore.test.ts
+++ b/src/main/deck/__tests__/deckWorkStore.test.ts
@@ -21,6 +21,7 @@ import {
   loadActiveDeckWorks,
   hasPendingDeckWorkA2aTasks,
   renderActiveDeckWorkBlock,
+  renderActiveDeckWorkReminderLine,
   renderStrandedDeckWorkBlock,
   getDeckWorkPath,
   setDeckWorkBootId,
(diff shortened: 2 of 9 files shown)


openwong2kim:

This PR fixes three compounding token sinks in the Command Deck orchestrator, all surfaced by a 90-minute supervision loop on 2026-08-07. It touches `stopGate.ts` (two new fail-open rules), `stopGateState.ts` (cap-out hysteresis record), `deckWorkStore.ts` (whitespace-insensitive follow-up dedupe and a one-line reminder renderer), and `deck.handler.ts` (the wiring that feeds `pendingDecision`/`suppressedFingerprint` into the gate and the changed-only injection contract for the active-work block).

Rule 4 releases the gate when a deck decision is pending — the mirror image of the coalescer's existing wake suppression — so a brain blocked from acting while waiting on a human stops re-printing its question four times per cycle. Rule 5 records a fingerprint of what the gate capped out on and stays quiet on identical state until it changes, a human acts, or a 15-minute TTL fires; the handler only records a suppression for a state that was actually refused at least once, so a worker dispatched mid-run cannot be silenced. The active-work diet collapses a byte-identical block to a reminder line carrying the id and the two ownership imperatives, re-sending in full on any change or TUI conversation reset, and PARKED records are deliberately exempt because the reminder speaks ownership language. The diff is clean: the pure modules are well tested, the wiring tests in `deck.handler.activeWorkDiet.test.ts` exercise the full→reminder→full-on-change→re-arm cycle, `tsc` is clean, and 10,387 tests pass with the one pre-existing parallel flake noted.

Priority **P2**: this is a meaningful behavior and efficiency fix to the orchestrator loop, but it is not a regression or a data-loss path, so it does not need the P1 panel.

openwong2kim:
Not in tonight's batch — the rest of the stack (#824/#825/#827/#828/#829/#830/#831) merged, this one still needs a live look.

The gate only evaluates inside `ClaudePtyBrainAdapter`, so exercising rules 4 and 5 for real means running an actual Command Deck brain against a fleet, not just poking RPCs. That's a real session with real token spend, and the failure this fixes (a wake loop re-buying the same three refusals every cycle) takes a while to show itself — a two-minute run proves nothing either way.

The unit side is already thorough, so this is about watching the loop behave over time rather than doubting the logic. Will pick it up in a supervision session where a brain is running anyway and report what the refusal/injection volume actually looks like.

openwong2kim:
Dogfooded this live on Windows before merging — a real claude-pty brain against a real fleet, with the gate callback instrumented so each verdict was visible.

Rule 4 is exactly what you describe. Two consecutive evaluations, same work, same fleet:

```
{"pendingDecision":false,"activeWork":true,"block":true,
 "reason":"Do not end this turn yet: 1 worker pane still need you — daemon-587eb539 (running)…",
 "fingerprint":"work-4a286dc8…@1786123451716|daemon-587eb539:running"}

{"pendingDecision":true, "activeWork":true,"block":false}
```

and the brain got there on its own — it read the refusal, called `deck_ask_decision`, and the gate released. That is the loop the refusal strings were rewritten to produce, working end to end without anyone steering it.

The fingerprint tracks state the way it should: once the worker stopped being `running`, the same work id produced `work-4a286dc8…@1786123451716|` with an empty pane list, so a changed fleet cannot be mistaken for the state that was refused.

Rule 5 never fired, which I read as the fix working rather than a gap — the brain kept finding a legitimate exit (ask, or finish), so no refusal run ever reached the cap. On the brain's own screen the whole thing terminates instead of spinning:

```
work-c2dea1b8… is still ACTIVE. Do not claim completion or end yet…
deck_complete_work succeeded for work-c2dea1b8…. Ready for the next instruction.
```

No repeated full [active-work] block in the transcript either. The instrumentation was local to my dogfood tree and is not in this branch.

PR #834, opened 2026-08-07
fix(cwd): resolve the host platform from the preload bridge

### LINKED ISSUE #833: Surface/pane cwd never follows a cd while the workspace's does — blocks fan-out
## What happens

`cd` in a pane updates the **workspace's** tracked cwd but never the
**surface/pane's**. Same shell, same OSC 7, two different places to land — one
of them lands.

Measured right after a `cd "D:\"` in a pane:

```
workspace.list  → ws "Workspace 1"  metadata.cwd = D:\          <- correct
surface.list    → ptyId daemon-d0361347  cwd = C:\Users\rizz    <- stale
pane.list       → pane b48547c0          cwd = C:\Users\rizz    <- stale
```

The surface/pane value never catches up — polled for 16s and across several
`cd`s. It appears to stay at whatever the pane started with.

Shell integration is loaded and emitting:

```
PS> echo "SI=$($function:prompt -match 'osc7')"
SI=True
```

so the signal is being produced, and the workspace row proves it is being
received and parsed. Only the surface-level consumer misses it.

## Why it matters

`task.fanout.start` derives the repo from the **surface** cwd, not the
workspace one — `resolveSenderSurfaceCwd()` matches the caller's `ptyId` in the
surface list and reads `row['cwd']` (`src/main/pipe/handlers/fanout.rpc.ts`).
So it reads the stale value and refuses:

```
```

even when the pane is sitting inside a repo and the workspace row already says
so. There is no way to work around it from the wire surface either —
`pane.split` takes no cwd and `workspace.new` only forwards `name` — so a pane
cannot be created in a chosen directory to sidestep it.

Anything else reading pane/surface cwd (MCP callers, agents) sees the same stale
value. The UI reads the workspace row, which is why this is invisible in normal
use.

## Repro

1. Open a pane; note `surface.list` cwd and `workspace.list` metadata.cwd — both
   are the start directory.
2. `cd` into any other directory.
3. Read both again: the workspace one follows, the surface/pane one does not.

## Notes

Found while dogfooding the #824/#827/#828/#829/#830/#831 batch; not caused by
it — the cwd path (`notification:cwd-changed` → `updateSurfaceCwd`) isn't in
that diff. Windows 11 / pwsh 7 / dev build. Not bisected, so I don't know how
long it has been this way or whether other platforms share it.

(Earlier revision of this issue claimed OSC 7 wasn't being followed at all —
that was wrong, the workspace row does follow it. The split between the two
consumers is the actual bug.)

snowyukitty:
## What

`isPlausibleCwd()` resolves the host platform from the preload bridge before
falling back to `process.platform`. That single default was the whole bug.

Fixes #833.

## Why the surface cwd froze but the workspace cwd did not

`updateSurfaceCwd` guards its write with the shape check and passes **no**
platform:

```ts
// src/renderer/stores/slices/surfaceSlice.ts
if (!isPlausibleCwd(cwd)) return;
```

The default read `process.platform`. The renderer runs under
`contextIsolation: true` / `nodeIntegration: false`, so it has no `process`
global at all — the guard fell through to its `'linux'` fallback, where
`isWinShape` is fatal:

```ts
if (platform === 'win32') return true;
return !isWinShape;   // every C:\… is "implausible"
```

So on a Windows host **every** `cd` was dropped on the floor at the write.

The workspace-level cwd survived because it has a second feeder with no shape
guard at all — `metadata.onUpdate` → `applyToWorkspace`, alongside the
unguarded write in the same coalescer callback in `useNotificationListener`.
One field, two writers, one of them guarded: that asymmetry is what made this
look like "OSC 7 isn't being followed" when the signal was arriving correctly
the entire time.

The triage note that `gitBranch` follows the `cd` is the confirming detail, and
it also proves the event really does reach the renderer: the daemon's
`GitContextWatcher` is driven off `session:cwd`, the same event that is
broadcast as `cwd.changed` → `IPC.CWD_CHANGED`. Every hop between them is
unconditional. Nothing upstream of the renderer was broken.

## Consequences this fixes

- `surface.list` and `pane.list` both serve `surface.cwd || workspaceCwd`, so
  both reported the spawn directory forever.
- `task.fanout.start` derives its repo from that value
  (`resolveSenderSurfaceCwd` in `fanout.rpc.ts`), so fan-out refused to start
  from any pane whose app had not opened directly on the repository — the
  `FAILED_PRECONDITION` in the report.

## Why fix the default rather than the call site

Two other renderer call sites — `GitTab.tsx:60` and `ReviewTab.tsx:82` — had
already hit this and worked around it locally by passing
`window.electronAPI.platform`, with a comment naming the exact trap. Those are
both *read* paths; the *write* path was left on the broken default, which is
why the stored value was wrong while the views that re-checked it looked fine.
Patching the third call site would have left the same trap armed for the
fourth. Fixing the default fixes every present and future caller, and call
sites that pass a platform explicitly are unaffected.

Per-call-site verdicts for `isPlausibleCwd`:

| Call site | Verdict |
|---|---|
| `main/pty/cwdDetect.ts:88` | passes platform explicitly; main has a real `process`. No change. |
| `renderer/components/Deck/GitTab.tsx:60` | already passes the bridge platform. No change (and now correct even if it stops). |
| `renderer/components/Deck/ReviewTab.tsx:82` | same. No change. |
| `renderer/stores/slices/surfaceSlice.ts:315` | **the bug** — omitted platform on the write path. |

## Adjacent instance, deliberately not bundled

`createPathLinkProvider` (`renderer/terminal/pathLinkProvider.ts:166`) has the
same `process.platform`-or-`'linux'` default and its only caller
(`useTerminal.ts:907`) omits the argument, so UNC paths are not linkified on
Windows. Its docstring already documents that degradation as understood, and it
is cosmetic rather than a correctness bug, so it is out of scope here. Happy to
send it separately if you want it.

## Verification

- `src/shared/__tests__/cwdShape.test.ts` — new `default platform resolution`
  block. Every pre-existing case passes a platform explicitly, so the default
  was never exercised; that is why the suite could not have caught this.
- `src/renderer/stores/slices/__tests__/surfaceSlice.test.ts` — reproduces the
  reported symptom at the store level, plus a case proving the 2026-07-20
  prompt-scrape guard still rejects a `C:\…` on a POSIX host.
- Both stub the host platform through the preload bridge, the same channel the
  renderer really reads, so they are meaningful on a POSIX CI runner.
- I also fixed a test that passed for the wrong reason: the "unknown ptyId"
  no-op case asserted with `D:\nope`, which a POSIX runner rejects on shape
  before the ptyId check does any work. It uses a POSIX path now.
- Confirmed load-bearing: with `process.platform` stubbed to `linux` to emulate
  a POSIX runner, the new test fails without the change
  (`expected false to be true`) and passes with it.
- `npx tsc --noEmit` clean; `eslint` clean on the touched files.

Verified by static trace plus the regression tests above. I did not do a live
run of a packaged build against a real daemon, so the integration hop is argued
from the code rather than observed.

## Not addressed

The report also notes there is no wire-level workaround because `pane.split`
takes no `cwd` and `workspace.new` forwards only `name`. That is a separate
enhancement and is left alone here; it stops being a workaround-blocker once
the tracked cwd is correct.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows support for pane working directories, including updates after changing directories.
  * Directory listings now report the current path correctly.
  * Repository detection for task fan-out now recognizes pane directories.
  * Improved validation of Windows and macOS path formats across platforms.
  * Added a reliable fallback for detecting the host platform.

### DIFF
--- a/changelog.d/834.md
+++ b/changelog.d/834.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **A pane's working directory follows `cd` again on Windows.** The per-surface
+  cwd was frozen at whatever directory the pane started in, so `surface.list`
+  and `pane.list` reported a stale path no matter how many times you changed
+  directory — and `task.fanout.start`, which derives the repository from that
+  value, refused to start with "the calling terminal's directory is not inside a
+  git repository" even when the pane was sitting inside one. The shape guard on
+  the cwd write resolved the host platform from `process.platform`, which does
+  not exist in the renderer under context isolation; it therefore assumed a
+  POSIX host and discarded every `C:\…` as impossible. It now reads the platform
+  from the preload bridge. The workspace-level directory was never affected,
+  which is why only agents and MCP callers saw the stale value. (#834)
--- a/src/renderer/stores/slices/__tests__/surfaceSlice.test.ts
+++ b/src/renderer/stores/slices/__tests__/surfaceSlice.test.ts
@@ -182,12 +182,62 @@ describe('surfaceSlice.updateSurfaceCwd', () => {
     const paneId = state.workspaces[0].rootPane.id;
     slice.addSurface(paneId, 'pty-1', 'pwsh', 'C:\\a');
 
-    slice.updateSurfaceCwd('', 'D:\\nope');
-    slice.updateSurfaceCwd('ghost', 'D:\\nope');
+    // POSIX shapes on purpose: a Windows path would also be dropped by the
+    // cwdShape guard on a POSIX runner, so this would pass without the ptyId
+    // check doing any work at all.
+    slice.updateSurfaceCwd('', '/nope');
+    slice.updateSurfaceCwd('ghost', '/nope');
 
     const pane = state.workspaces[0].rootPane;
     if (pane.type !== 'leaf') throw new Error('expected leaf pane');
     expect(pane.surfaces[0].cwd).toBe('C:\\a');
+  });
+
+  // Issue #833 — on a Windows host the per-surface cwd never followed a `cd`.
+  // updateSurfaceCwd guards with isPlausibleCwd() and passed no platform; the
+  // renderer has no `process` (contextIsolation), so the default resolved to
+  // 'linux' and rejected every `C:\…`. The workspace-level cwd has a second,
+  // unguarded feeder (the metadata route), which is why only the surface value
+  // looked frozen — and why `surface.list`/`pane.list` reported the spawn
+  // directory forever, breaking task.fanout.start's repo precondition.
+  //
+  // The suite runs on POSIX in CI, so the host platform is stubbed through the
+  // preload bridge — the same channel the renderer really reads.
+  it('follows a cd on a Windows host (#833)', () => {
+    const host = globalThis as { electronAPI?: { platform?: string } };
+    host.electronAPI = { platform: 'win32' };
+    try {
+      const { state, slice } = createHarness();
+      const paneId = state.workspaces[0].rootPane.id;
+      slice.addSurface(paneId, 'pty-1', 'pwsh', 'C:\\Users\\rizz');
+
+      slice.updateSurfaceCwd('pty-1', 'D:\\AI_Projects\\wmux-fork');
+
+      const pane = state.workspaces[0].rootPane;
+      if (pane.type !== 'leaf') throw new Error('expected leaf pane');
+      expect(pane.surfaces[0].cwd).toBe('D:\\AI_Projects\\wmux-fork');
+    } finally {
+      delete host.electronAPI;
+    }
+  });
+
+  it('still rejects an implausible shape for the host (prompt-scrape guard)', () => {
+    const host = globalThis as { electronAPI?: { platform?: string } };
+    host.electronAPI = { platform: 'darwin' };
+    try {
+      const { state, slice } = createHarness();
+      const paneId = state.workspaces[0].rootPane.id;
+      slice.addSurface(paneId, 'zsh-1', 'zsh', '/Users/me');
+
+      // The 2026-07-20 incident: "PS C:\…>" scraped off screen text on a mac.
+      slice.updateSurfaceCwd('zsh-1', 'C:\\Users\\me');
+
+      const pane = state.workspaces[0].rootPane;
+      if (pane.type !== 'leaf') throw new Error('expected leaf pane');
+      expect(pane.surfaces[0].cwd).toBe('/Users/me');
+    } finally {
+      delete host.electronAPI;
+    }
   });
 });
 
--- a/src/shared/__tests__/cwdShape.test.ts
+++ b/src/shared/__tests__/cwdShape.test.ts
@@ -1,5 +1,14 @@
-import { describe, it, expect } from 'vitest';
+import { describe, it, expect, afterEach } from 'vitest';
 import { isPlausibleCwd } from '../cwdShape';
+
+type BridgeHost = { electronAPI?: { platform?: string } };
+
+/** Stand in for the preload bridge — the renderer's only host-platform source. */
+function setBridgedPlatform(platform: string | undefined): void {
+  const host = globalThis as BridgeHost;
+  if (platform === undefined) delete host.electronAPI;
+  else host.electronAPI = { platform };
+}
 
 // Regression (2026-07-21): a pane's cwd was stored as the literal string "path"
 // — a prompt-scrape false positive that the old win32 rule ("any non-empty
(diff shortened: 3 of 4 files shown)


openwong2kim:

This is a focused, correct fix for #833 — the per-surface cwd freezing on Windows. The root cause is sharp: `isPlausibleCwd()` defaulted its platform argument to `process.platform`, but the renderer runs under context isolation with no `process` global, so the default resolved to `'linux'` and discarded every `C:\…` path on the write path (`updateSurfaceCwd`). The workspace-level cwd kept following `cd` because it has a second, unguarded feeder, which is exactly the asymmetry that made the bug look like "OSC 7 isn't being followed" rather than a guard dropping the update. The new `hostPlatform()` resolver reads the preload bridge first, then `process`, then falls back to `'linux'`, and fixing the default (rather than patching the one broken call site) closes the same trap at `GitTab`, `ReviewTab`, and any future caller in one go.

The scope is tight — one function in `cwdShape.ts` plus regression tests in both `cwdShape.test.ts` and `surfaceSlice.test.ts` that stub the platform through the same preload-bridge channel the renderer really reads, so they are meaningful on a POSIX CI runner. I like that the existing "unknown ptyId" no-op test was corrected to use a POSIX path so it no longer passes for the wrong reason on POSIX runners, and that the adjacent `pathLinkProvider` UNC case is explicitly called out as out of scope (cosmetic, separate change).

Priorities line up: #833 is P2 and this unblocks `task.fanout.start`, whose `resolveSenderSurfaceCwd` reads exactly the surface cwd this fixes, so I am labelling this PR **P2**. The reasoning is argued from a static trace plus the new tests rather than a live packaged run against a real daemon, which is worth keeping in mind during review but does not block merge given the test coverage and the narrowness of the change.

openwong2kim:
Nice catch, and thanks for going after the default rather than the call site — I'd started on the narrower version (#835, now closed) and yours is the right shape. Patching `surfaceSlice` alone would have left the same trap for whoever adds the fourth caller.

I'd already been chasing this on Windows for a different reason, so here is live confirmation on real hardware rather than a stubbed bridge.

Instrumented `updateSurfaceCwd` in a dev build and drove a `cd`:

```
[probe833] updateSurfaceCwd called daemon-6d3a8eea C:\Windows\System32 plat= win32 guard= true
[probe833] surface FOUND, writing C:\Windows\System32
surface cwd after cd: C:\Windows\System32
```

Evaluated the guard inside the renderer, which pins the diagnosis exactly:

```
{"hostPlatform":"win32","win32":true,"withHost":true,"noArg":false}
```

`noArg: false` is the bug in one value — same path, same process, only the omitted platform differs.

And the consequence you mention in the body is real. Before: `task.fanout.start` refused a pane that had `cd`'d into a repo. After:

```
{"status":"accepted","repoPath":"...\scratchpad\fanrepo","taskCount":2}
```

So this unblocks fan-out (#826), which had no wire-level workaround — `pane.split` takes no cwd, `workspace.new` forwards only `name`, and `surface.new` ignores the `cwd` it is handed. That last one still looks broken to me and is worth a follow-up; it is a separate write path and this PR doesn't need to grow to cover it.

One small thing, entirely your call: `isPlausibleCwd` now reads `globalThis.electronAPI` and this module is also imported by the daemon and main. Both fall through to `process.platform` since nothing defines that global there, so it is correct as written — but a one-line note saying the bridge is renderer-only and the fallback is the non-renderer path would save the next reader the trace I just did.

openwong2kim:
Merged. Thanks again — and it turned out to be worth more than the tooltip it looks like from the outside.

I picked this up because fan-out (#826) was dead in the water: `task.fanout.start` reads the surface cwd, so it refused every pane with

```
```

even sitting inside a repo, and there was no way around it from any surface — `pane.split` takes no cwd, `workspace.new` forwards only `name`, `surface.new` ignores the cwd it's handed. With your fix in, the same call gets it right:

```
{"status":"accepted","repoPath":"…\scratchpad\fanrepo","taskCount":2}
```

That let me actually dogfood #826 for the first time, which immediately surfaced a second bug that had been hiding behind this one: the per-task `WMUX_TASK_PORT` never reached the pane either, because `resolveSpawnEnv` strips the reserved `WMUX_*` namespace and `applyOverlay` skips those keys. Two live-only defects, and yours was the gate on finding the other.

Also appreciated that you fixed the existing no-op test on the way through — the one asserting Windows paths on a POSIX runner. That assertion was passing for the wrong reason and is a good part of why this survived as long as it did. Easy to leave alone; you didn't.

PR #835, opened 2026-08-07
fix(renderer): read the host platform when validating a surface cwd

### LINKED ISSUE #833: Surface/pane cwd never follows a cd while the workspace's does — blocks fan-out

openwong2kim:
## The bug

`isPlausibleCwd(cwd, platform)` defaults `platform` to `process.platform`. The
renderer has no `process` — confirmed live over CDP:

```
{"hasProcess":"undefined","platform":null,"electronAPIplatform":"win32"}
```

so the default resolved to `'linux'`, and the guard's last line

```js
if (platform === 'win32') return true;
return !isWinShape;
```

rejected every Windows path. `updateSurfaceCwd` returned early on each OSC 7
report, so a surface's `cwd` stayed at whatever the pane started with — for the
whole session.

## What made it invisible

Nothing upstream is broken. Sitting on the daemon pipe while driving a `cd`:

```
daemon broadcasts:
    cwd.changed/361347 = C:\Windows        <- emitted, received, parsed

~1.5s later:
    workspace metadata cwd : C:\Windows     <- applied
    surface cwd            : C:\Users\rizz  <- dropped
```

Both writes happen in the **same callback** (`cwdCoalescer`, `useNotificationListener.ts`):
`updateSurfaceCwd` first, then `updateWorkspaceMetadata`. The second one has no
plausibility guard, so it landed. The sidebar reads the workspace row — which is
why the UI looked correct while `surface.list` / `pane.list` served a stale path.

## Why it matters beyond a tooltip

`task.fanout.start` derives the repo from the **surface** cwd
(`resolveSenderSurfaceCwd`). With the surface frozen at home, fan-out refused a
pane that was sitting inside a repo:

```
```

and there was no way around it — `pane.split` takes no cwd, `workspace.new`
forwards only `name`, and `surface.new` ignores the `cwd` it is given. That last
one is likely the same root and is left for a follow-up; this PR fixes the live
update, which is what unblocks fan-out.

## The fix

Pass the host platform explicitly, from the preload bridge. `GitTab` and
`ReviewTab` already do this — with a comment naming this exact failure — so this
call site was simply missed. `hostPlatform()` gives that source a name instead of
a fourth inline cast, and returns `undefined` when there is no `window` so the
callee keeps its own default under plain-node unit tests.

## Tests

Two cases that cover each other across runners, which matters because the
symptom is platform-dependent:

- host `win32` → a Windows path is applied. Fails without the fix on a POSIX
  runner (default `'linux'` rejects it).
- host `darwin` → a Windows path is still rejected. Fails without the fix on a
  Windows runner (default `'win32'` wrongly accepts it) — verified locally:
  reverting the one-line change fails exactly this case, 1 failed / 32 passed.

The pre-existing tests here used POSIX paths deliberately ("the runner is POSIX,
so a Windows path would be rejected") — which is precisely why they never caught
this.

`tsc --noEmit` clean; `src/renderer/stores` + `cwdShape`: 738 passed.

Closes #833

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows pane working directories not updating correctly after using `cd`.
  * Corrected displayed paths and directory-dependent operations, including fan-out.
  * Improved platform-specific path validation to preserve the correct working directory across Windows and macOS.


openwong2kim:
Closing in favour of #834, which landed first and is the better fix.

I patched the one call site; #834 fixes the default itself, so the trap isn't left armed for the fourth caller. It also repairs the existing no-op test that used Windows paths on a POSIX runner — those assertions were passing for the wrong reason, which is part of why this bug survived so long.

For what it's worth, I verified the behaviour live on Windows before seeing #834. Instrumented `updateSurfaceCwd` in a dev build:

```
[probe833] updateSurfaceCwd called daemon-6d3a8eea C:\Windows\System32 plat= win32 guard= true
[probe833] surface FOUND, writing C:\Windows\System32
surface cwd after cd: C:\Windows\System32
```

and with the fix in place `task.fanout.start` gets the right repo from a pane that had `cd`'d into it — `{"status":"accepted","repoPath":"...\fanrepo"}` — where before it refused with FAILED_PRECONDITION. So #834 unblocks fan-out too. I'll say the same on that PR.

PR #838, opened 2026-08-08
Stop transplanting a workspace-level 'running' onto the active pane

### LINKED ISSUE #837: A hand-opened shell pane counts as an outstanding worker and blocks deck_complete_work
## What happened

Dogfooding the terminal brain (claude-pty) on 2026-08-08, the orchestrator could not close its own work item. It ended up asking me this, verbatim, through `deck_ask_decision`:

> Pane daemon-587eb539 is idle by every signal I can read, but `deck_complete_work` keeps rejecting with `workers_outstanding` (status "running") — how should I clear it?
>
> Options: Override/clear the stale worker status on daemon-587eb539 yourself, then I retry `deck_complete_work` · You restart or re-attach that pane's shell (I must not close/kill it), then I retry · Leave work-4a286dc8 open indefinitely and stop retrying

The stop gate agreed with the rejection. With the gate callback instrumented:

```
{"consecutiveBlocks":0,"pendingDecision":false,"activeWork":true,"block":true,
 "reason":"Do not end this turn yet: 1 worker pane still need you — daemon-587eb539 (running)…",
 "fingerprint":"work-4a286dc8-…@1786123451716|daemon-587eb539:running"}
```

Two things about that pane make this odd:

1. **It was never a worker.** `daemon-587eb539` is a plain pwsh pane I had opened by hand — no agent, no delegation, nothing spawned into it. It got counted as an outstanding worker anyway.
2. **The daemon did not think anything was running in it.** At the same time:

```
id=daemon-587eb539  state=attached  commandRunning=false  lastDetectedAgent=undefined
```

So the renderer-side fleet snapshot said `running` while the daemon's own view of the session said no command was running and no agent was ever detected.

## Where it resolves

`fleetPanes` (`src/renderer/stores/selectors/fleet.ts`) trusts the workspace-level status for the ACTIVE pane:

```js
const metaStatus = isActivePane ? wsMeta?.agentStatus : undefined;
const status = attention
  ?? (metaStatus && metaStatus !== 'idle' ? metaStatus : undefined)
  ?? (hookRunning ? 'running' : undefined)
  ?? 'idle';
```

That pane was its workspace's active pane, so whatever `ws.metadata.agentStatus` held won outright. I had been driving it with `input.send` (cd / echo) while testing, which is presumably what set it — but it stayed `running` past the point where the commands had finished.

Not permanent: everything reads `idle` now. So this is a stale window, not the forever-stuck case.

## Why it still matters

It blocks `deck_complete_work` for as long as it lasts, and the brain has no way out on its own — the pane isn't a worker it can address, and it is explicitly forbidden from closing a human's shell. Burning a human decision on "please clear a status" is the cost.

Worth noting the failure mode has improved: this is the #733 family, where the orchestrator used to `exit` the user's shell trying to clear exactly this. It no longer does — it asked instead, which is #832's refusal-string/gate-release work behaving correctly. The remaining gap is that it had to ask at all.

## Leads

- Should a pane with no agent ever have been counted as an outstanding worker? Filtering the gate's worker set by "has an agent" would sidestep the status question entirely.
- What clears `ws.metadata.agentStatus` after byte/detector-driven `running` on a plain shell, and is there a path where the clear is missed? #733 / #734 fixed one such path.
- The daemon already knows better (`commandRunning=false`). Nothing consults it here.

Windows 11, pwsh 7, dev build off `main` at 555d982c.

openwong2kim:
Fixes #837.

## What was happening

`ws.metadata.agentStatus` is **one slot per workspace**, written by whichever PTY last broadcast — `restrictContext` in `useNotificationListener.ts` strips `cwd`/`git`/`pr` but lets `agentStatus` through. Meanwhile `surfaceAgentStatus` retains only the ATTENTION statuses (`paneSlice.ts` `ATTENTION_STATUSES` deletes `running`/`idle`/`error`).

So a background worker's `running` survives nowhere but that shared slot, and `selectFleetPanes` attributed it to whichever leaf is `ws.activePaneId`.

That is how a hand-opened pwsh pane that never ran an agent read `running`: the status was transplanted from the real worker, which — not being the active pane — read `idle` itself. The pane then landed in the completion gate's outstanding set (`deck.rpc.ts:139-141`, `stopGate.ts:134`), `deck_complete_work` rejected with `workers_outstanding`, and the orchestrator had to spend a human decision clearing a status the daemon already knew was wrong.

## The fix

**The workspace tier never needed to carry `running` at all.**

Both emitters of that broadcast (`PTYBridge.ts:495`, `DaemonNotificationRouter.ts:915`) send a `ptyId`, so every one of them stamps `markSurfaceRunning(ptyId)` — the per-pty clock tier 3 already reads. Dropping `running` from tier 2 costs no coverage and does not shorten it: the shared slot is cleared back to `idle` after ~5s of silence (`PTYBridge.ts:98`, `DaemonNotificationRouter.ts:946`), while the tier 3 stamp holds for `HOOK_RUNNING_TTL_MS`.

One line in `fleet.ts`. `error` and every other non-running workspace state are untouched.

## What this replaced, and why

The first version of this PR vetoed the borrowed status when OSC 133 said the shell was at a prompt. A review panel took it apart, and it deserved it:

- **The veto could invert the bug.** `commandRunningByPtyId` refreshes only on a 15s `pty.list` poll, while `ws.metadata.agentStatus` updates on every METADATA_UPDATE. A stale `false` suppressed a real `running`, which would let the completion gate pass *while a worker was live* — worse than the bug being fixed. A stale `true` delayed the release in the other direction.
- **It swallowed `error`.** `error` is not an ATTENTION status, so the workspace slot is its only carrier for the active pane. An agent that died and dropped back to a shell prompt would have had its `error` silently rewritten to `idle`.
- **It judged a multi-tab pane by its active surface alone.** Tab 1 running `claude`, tab 2 a shell, switch to tab 2 → the live agent reads `idle`.
- **It did nothing for shells without OSC 133** (cmd.exe, no integration snippet), where the original bug survived untouched.

The current approach has none of these: no polled input, no veto, `error` untouched, and no shell-integration dependency.

## Tests

`fleet.test.ts` — four cases on the reported topology (active hand-opened shell + background worker, workspace slot at `running`):

- the active pane that produced no running of its own → `idle`
- the pane whose own pty stamped it → still `running`
- `error` on the active pane → comes through, not vetoed
- a pane's own attention status → never affected

Two existing fixtures pinned the old behaviour (a pane inheriting the workspace-level `running`); they now stamp the pane's own pty, which is what the live path always does.

`tsc --noEmit` clean, `eslint` clean, full `test:parallel` 10612 passed. The one failure in that run, `integrations/claude/bridgePermissionGate.test.ts`, is a 15s timeout waiting on a daemon that is not running locally — it fails the same way without this branch, and this diff touches only the renderer selector and its tests.

## What this does not close

A hand-opened pane running a *genuine* foreground command still blocks the gate — it produces real byte activity, so it stamps a real `running`. Whether an agent-less pane should count as an outstanding worker at all is the issue's lead 1, and it wants its own discriminator: `agentName` can be `''` before the detector gate matches, so a naive "has an agent" filter buys the opposite and worse failure — the brain ending its turn on a live worker. Left open on the issue.

### DIFF
--- a/changelog.d/838.md
+++ b/changelog.d/838.md
@@ -1,9 +1,9 @@
 ### Fixed
 
 - **A pane you opened by hand no longer counts as an agent that is still
-  working.** Agent status was stored once per workspace, so a busy agent in a
+  working.** Agent status is stored once per workspace, so a busy agent in a
   background pane made whichever pane you had focused look busy too — including
   a plain shell that had never run an agent. The orchestrator then refused to
   close out its work, and the only way past it was to answer a question about
-  clearing a status that was never true. wmux now checks whether that specific
-  pane's shell is sitting at a prompt before believing it. (#838)
+  clearing a status that was never true. A pane now shows as running only when
+  its own terminal produced the activity. (#838)
--- a/changelog.d/838.md
+++ b/changelog.d/838.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **A pane you opened by hand no longer counts as an agent that is still
+  working.** Agent status was stored once per workspace, so a busy agent in a
+  background pane made whichever pane you had focused look busy too — including
+  a plain shell that had never run an agent. The orchestrator then refused to
+  close out its work, and the only way past it was to answer a question about
+  clearing a status that was never true. wmux now checks whether that specific
+  pane's shell is sitting at a prompt before believing it. (#838)
--- a/src/renderer/components/Deck/DeckFleet.tsx
+++ b/src/renderer/components/Deck/DeckFleet.tsx
@@ -70,13 +70,12 @@ export default function DeckFleet({
   const activeWorkspaceId = useStore((s) => s.activeWorkspaceId);
   const surfaceAgentStatus = useStore((s) => s.surfaceAgentStatus);
   const surfaceActivity = useStore((s) => s.surfaceActivity);
-  const commandRunningByPtyId = useStore((s) => s.commandRunningByPtyId);
   const paneLabel = useStore((s) => s.paneLabel);
   const paneRole = useStore((s) => s.paneRole);
   const roleBindings = useStore((s) => s.orchestratorRoleBindings);
 
   const panes = useMemo(() => {
-    const all = selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, commandRunningByPtyId });
+    const all = selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel });
     // Roster = live terminal panes of the ACTIVE workspace only (M1.5: the
     // deck is this workspace's orchestrator, so its roster is this
     // workspace's agents — the fleet-wide view lives in the titlebar vitals).
@@ -88,7 +87,7 @@ export default function DeckFleet({
       ),
       'attention',
     );
-  }, [workspaces, activeWorkspaceId, surfaceAgentStatus, surfaceActivity, paneLabel, commandRunningByPtyId]);
+  }, [workspaces, activeWorkspaceId, surfaceAgentStatus, surfaceActivity, paneLabel]);
 
   if (panes.length === 0) return null;
 
--- a/src/renderer/components/Deck/DeckFleet.tsx
+++ b/src/renderer/components/Deck/DeckFleet.tsx
@@ -70,12 +70,13 @@ export default function DeckFleet({
   const activeWorkspaceId = useStore((s) => s.activeWorkspaceId);
   const surfaceAgentStatus = useStore((s) => s.surfaceAgentStatus);
   const surfaceActivity = useStore((s) => s.surfaceActivity);
+  const commandRunningByPtyId = useStore((s) => s.commandRunningByPtyId);
   const paneLabel = useStore((s) => s.paneLabel);
   const paneRole = useStore((s) => s.paneRole);
   const roleBindings = useStore((s) => s.orchestratorRoleBindings);
 
   const panes = useMemo(() => {
-    const all = selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel });
+    const all = selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, commandRunningByPtyId });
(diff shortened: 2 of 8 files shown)


openwong2kim:

This is a tightly scoped renderer fix for #837, where a hand-opened shell pane inherited a workspace-level `running` status from a busy background agent and then blocked `deck_complete_work` with a spurious `workers_outstanding`. The root cause is that `ws.metadata.agentStatus` is a single slot per workspace written by whichever PTY last broadcast, while `surfaceAgentStatus` only retains ATTENTION statuses — so a background worker's `running` survived only in that shared slot and was painted onto whichever leaf happened to be `ws.activePaneId`.

The fix makes `selectFleetPanes` consult the daemon's per-pty OSC 133 signal (`commandRunningByPtyId`), which was already in the store and already ranked tier-1 authoritative by `isPaneAgentBusy` twenty lines above but was simply not read on this path. When OSC 133 reports `false` (shell back at a prompt), the borrowed workspace-level `running` is vetoed; `undefined` (no shell integration) preserves legacy behaviour exactly, and an attention status the pane holds in its own right is never overridden. The change also wires that map into `useWorkspaceMirrorPush` (15s `pty.list` poll) so the veto propagates inside the 30s window the completion gate reads, and threads the same signal through `DeckFleet`, `FleetView`, and `StatusBar` so the UI and the gate agree.

Test coverage is appropriate: four new cases in `fleet.test.ts` encode the reported topology (active hand-opened shell + background worker, workspace slot at `running`) and assert the idle-on-prompt, keep-running-on-foreground, legacy-no-integration, and never-veto-own-attention behaviours. I'd note the two residuals the PR description calls out are genuine and correctly left open — without OSC 133 on the active pane the transplant can still occur, and a foreground command in a hand-opened pane still blocks the gate; both are the "should an agent-less pane count as a worker at all" question and want their own discriminator. CI is still in progress at review time; the local `tsc`/eslint/9711-test claims in the description are the relevant signal once green.

Priority: **P2** — directly unblocks the orchestrator completion path for the #837 failure mode, which was burning human decisions on a status the daemon already knew was stale, but the scope is narrow and the residual cases do not affect correctness for the addressed path.

PR #840, opened 2026-08-08
Fit a mirrored remote pane to its cell instead of cropping it

openwong2kim:
## The symptom

An attached remote workspace's mirrored pane lost the right-hand half of every
long line, and the input box of the terminal app running inside it was not
visible at all — typing made a one-line sliver appear, still cut off.

## Why

A mirror renders the grid the **remote** daemon owns: `RemoteMirrorTerminal`
only ever calls `term.resize()` with cols/rows the remote sent, because
geometry has a single owner and a viewer must not resize someone else's pane.
The pixels, though, are local — it draws with this app's font settings. Nothing
related the two, so the rendered element was `remoteCols × localCellWidth` wide
regardless of the box it sat in.

#819 added `overflow-hidden` so that oversized element stopped painting over the
composer and the sidebar. That turned overflow into a **top-left crop**, and a
terminal app keeps its input box on the last rows — so the crop removed exactly
the prompt the user was typing into. Typing grows that box upward by a line,
which is the sliver.

## The fix

Shrink the mirror's own font until the remote's grid fits.

Deliberately **not** a CSS transform. xterm derives every mouse coordinate from
`getBoundingClientRect()` divided by an unscaled cell width, so a scaled mirror
maps clicks to the wrong cell — and under `scale(s)` the reachable area
compresses into the top-left `s` fraction while `getCoords` clamps to
cols/rows, making the right and bottom of the grid unreachable. Those clicks are
not local decoration: SGR mouse reports (`ESC[<0;c;rM`) leave through `onData` →
`paneWrite` into a live remote shell, and they do not match the `isDeviceReply`
gate, which only knows the DA/DSR shapes. Changing the font size keeps xterm's
metrics and the rendered size in agreement, so coordinates stay exact.

`fontSize` now has exactly one writer. The settings effect no longer assigns it;
the user's setting is the fit's upper bound instead. Two writers is how the fit
would be undone, since that effect re-runs on any settings change.

## How it settles

The fit measures `.xterm-screen`'s layout box, which carries the grid's natural
size and cannot feed back into itself the way `getBoundingClientRect` would.
Cell metrics are a staircase in font size rather than a line (xterm rounds
through `ceil`/`floor` and the device pixel ratio), so a pass re-measures and
runs again; passes for an unchanged key may only shrink, which is what makes the
loop settle rather than oscillate. The key covers box size, the remote grid, the
user's font size **and family** — anything left out is an input whose change the
fit would silently ignore, so there is a test per input.

A mirror in a non-active workspace sits inside `display:none`, where every
measurement is zero. The fit declines to decide there rather than assigning NaN,
and the ResizeObserver's zero-to-real transition is what makes the deferred fit
happen when the workspace is selected. Box-driven fits are debounced so a
divider drag does not restyle six terminals per frame, but a discrete request
(a remote resize, a settings change) is never postponed behind that debounce.

The theme background moves to the clipping box: once the grid is smaller than
the cell, the letterbox margin would otherwise show `--bg-base` beside the
terminal's own background.

## Not in scope

The window header still shows the **local** workspace while a remote one is
active — `selectActiveWorkspaceSummary` reads only `activeWorkspaceId` and knows
nothing about `activeRemoteKey`. Separate and smaller.

## Verification

- `src/renderer/components/Remote/__tests__/mirrorFit.test.ts` — new. The
  arithmetic is a pure module because jsdom reports every layout as 0×0, so the
  numbers cannot be checked through the component: fit-to-width, fit-to-height,
  tighter-axis, the minimum-size floor, quantise-down, the hidden/unmeasurable
  and non-finite cases, a nonsense restored ceiling, and both halves of the
  termination argument.
- `RemoteMirrorTerminal.test.tsx` — the font-size test now asserts the single
  writer contract (the settings effect must NOT write `fontSize`), with the
  no-recreate check moved onto font family.
- `tsc --noEmit` clean, `eslint` no errors, full suite 10,633 passing.
- Not yet confirmed by eye against a live remote workspace.


openwong2kim:

This PR fixes a real rendering defect in the remote mirror pane: because the mirror renders the remote daemon's grid with local font metrics but must not resize someone else's pane, a remote grid larger than its cell was cropped from the top-left by the `overflow-hidden` added in #819 — losing the right half of long lines and, more visibly, the TUI input box that lives on the bottom rows. The fix shrinks the mirror's own font until the remote grid fits, deliberately avoiding a CSS transform so xterm's mouse-coordinate derivation stays exact (clicks leave as SGR reports into a live remote shell). The arithmetic is isolated in a pure `mirrorFit.ts` module (since jsdom reports every layout as 0×0), with a shrink-only-on-later-passes rule that makes the measure→apply loop terminate and a single-writer contract on `fontSize` that prevents the settings effect from undoing the fit. Tests are thorough — `mirrorFit.test.ts` covers fit-to-width, fit-to-height, the floor, the termination argument, and degenerate inputs, and the component test now asserts the single-writer invariant.

**Priority: P2** — this is a genuine usability regression on the remote-mirror path (the input row the user is typing into is invisible), but it is scoped to remote mirror panes and is not a crash or data-loss issue.

One note for review: the PR itself flags that it has not yet been confirmed by eye against a live remote workspace — that is the remaining verification gate before merge.

PR #841, opened 2026-08-08
refactor(detector): one table for agent identity, declarative compound gates, fork inheritance

openwong2kim:
Adding an agent to wmux meant hand-editing the same eight strings in seven
places, plus two more files carrying the slug↔display maps, each with a comment
asking the next person to keep the others in lock-step. That had already failed
quietly, and the detector was also carrying one agent's identity rules as a
hardcoded branch.

This restructures the detector's data so those three problems stop being
possible. No new feature, no config surface, no new dependency.

## One table for agent identity

`src/shared/agentIdentity.ts` now owns a single array of `{ slug, display }`.
The union type, the runtime set, both lookups, the hook-envelope allowlist, the
resume-binding validator and the channel-candidate seeder all derive from it.
Adding an agent is one row. Existing import paths keep working through
re-exports, so no call site changed.

This fixes a live bug: `agentCandidateSeed.ts` listed seven of the eight slugs.
`] satisfies AgentSlug[]` rejects *extra* members but never *omissions*, so the
gap compiled and no test caught it. A slug-shaped `openclaude` therefore missed
that set, fell through to the display→slug map (which only knows `OpenClaude`),
and came back `undefined` — reintroducing the auto-name suffix loss that set was
added to fix, for one agent.

The stated reason for all the duplication was that the daemon bundle cannot
import outside `src/shared`. That is only half true: `DaemonPTYBridge.ts`
already imports `AgentDetector` straight out of `src/main/pty/`, so `rootDir`
was never the constraint — only `integrations/` is genuinely out of reach. The
stale comment is gone.

## Compound gates are a shape, not a special case

A product-name mention is not proof an agent is running; agents routinely print
logs and docs naming other agents. Kiro handled that by requiring two
independent pieces of chrome from the same PTY — but as an `ap.slug === 'kiro'`
branch plus three fields on the detector, so no other profile could use it.

A gate is now either a regex or `{ allOf: EvidenceClause[] }`. A clause carries
cheap literal hints, a set of matchers (`line` / `contains` / `normalized`), and
an optional `emitOnGateOpen` that replays with the text the clause captured.
Kiro is expressed entirely in that shape and the branch is gone; no `slug ===`
comparison remains in the engine.

Its eight existing tests pass **unmodified**, which was the acceptance bar: the
shape is only worth having if it can hold the discipline the branch held.

This also revives a matcher that could never fire. The whitespace-stripped path
— the one that exists for cursor-drawn composers whose spaces are gone before
the bytes reach us — opened its search window with `lastIndexOf('ask')`. The
needle ends in `...a task`, so the window always latched onto that trailing copy
and began past the text it was looking for. It matched nothing, ever.

## Forks inherit instead of restating

OpenClaude is Claude Code's TUI, so it showed Claude's approval prompts and the
profile restated all four: 200+ character patterns, box-glyph classes included,
byte-identical. Review rounds 2 through 5 each tightened them in both places by
hand, and the cost of missing one is on record — a worker pane sat on an
unmatched approval prompt for 100 minutes.

A profile can now declare `extends` plus an `omit` list. OpenClaude keeps its
own gate and its own bare `>` idle prompt, inherits the four approval patterns,
and omits the two Claude waiting patterns that do not apply to it. Inheritance
is flattened once at module load, so the matching hot path is unchanged.

## Two things deliberately not done

**`processLine`'s dedup `return`s were left alone.** They read like a swallowed
detection, and turning them into `continue` looks like an obvious fix. It is a
regression: Claude and OpenClaude carry byte-identical approval patterns, so
with both gates open that lets turn 2 of a repeated prompt re-emit under the
other agent's name. Verified — flipping one `return` makes the new test fail
with a second `awaiting_input` for OpenClaude. Critical and agent patterns also
cannot co-match (unanchored shell-command shapes vs whole-line-anchored TUI
chrome), so falling through buys nothing. There is now a flow diagram over
`processLine` and two tests that fail if someone changes it anyway.

**No literal prefilter on gate regexes.** Gate regexes run twice per completed
line on any pane with no identified agent, forever, which invites screening each
line with `String.includes` first. Benchmarked over 20k build-log lines before
adding one:

| profiles | plain regex sweep | with literal prefilter | |
|---|---|---|---|
| 7 (today) | 9.6ms | 11.3ms | 0.84× — slower |
| 16 | 26.5ms | 20.9ms | 1.27× |
| 48 | 75.0ms | 52.7ms | 1.42× |

One `toLowerCase()` per line costs more than the seven regex tests it skips. The
threshold is recorded in `TODOS.md` next to the work that would make it worth
adding, and in a note over `checkGates`.

## Verification

Unit: full suite green (10775 passing). The Kiro tests pass unmodified;
`extends`, `omit`, the `indexOf` fix and the dedup invariant are each
mutation-checked — reverting any one of them fails exactly the test that guards
it, and nothing else.

Live: ran the bundled daemon under an isolated `WMUX_DATA_SUFFIX` with real
PTYs, then the same eleven cases against the pre-change bundle. The old bundle
fails exactly one — the revived matcher — and passes the other ten, which is the
evidence that the restructure changed no behaviour.

GUI: launched an isolated instance alongside the running one and typed the
chrome into real panes. Pane tabs read `w1-1(kiro)` and `w1-2(openclaude)`; the
sidebar roster shows `Kiro CLI — Waiting` and `OpenClaude — Needs input`. That
OpenClaude status comes from an approval pattern its profile no longer contains,
which is the inheritance working end to end.

One honest gap: the `agentCandidateSeed` omission was not reproduced on screen.
Its failure needs slug-shaped input, and the pane tab label goes through the
display-name path, which was fine before. Unit tests and the mutation check
cover it; the screenshots do not.


openwong2kim:

This PR is a substantial internal refactor of the agent identity and detection engine (P2): it consolidates eight scattered slug/display tables into a single source-of-truth array in `src/shared/agentIdentity.ts`, makes Kiro's compound identity gate a reusable declarative `allOf` shape instead of a hardcoded `slug === 'kiro'` branch, and lets forked profiles (OpenClaude) inherit their parent's patterns via `extends`/`omit` instead of restating them byte-for-byte. It also fixes two real defects along the way — the `agentCandidateSeed` omission that dropped the `(openclaude)` name suffix, and a Kiro cursor-composer matcher whose `lastIndexOf('ask')` window could never latch onto its target — while deliberately leaving two tempting changes untouched (the `processLine` dedup `return` is load-bearing against cross-agent re-emission, and a gate literal prefilter benchmarks slower at the current profile count).

The verification is thorough: full unit suite green (10775), Kiro's eight existing tests pass unmodified against the new gate shape (the acceptance bar for the refactor), live daemon comparison against the pre-change bundle, and mutation checks on every invariant the new fields introduce. Net change is +821/−209 across `AgentDetector.ts`, `agentIdentity.ts`, `signal-types.ts`, `ptyRecovery.ts`, and `agentCandidateSeed.ts`, MERGEABLE against `main`. Worth merging soon — the duplication this removes had already failed once quietly, and the fork-inheritance path removes the documented 100-minute-unmatched-prompt failure mode from recurring on the next review.

Issue #810, opened 2026-08-05
Browser scoping is opt-out: a caller that omits workspaceId reaches any workspace's surface


openwong2kim:
## What

Two related paths let a caller reach a browser surface outside its own workspace. Neither is a regression — both predate the current scoping work — but the set of callers that can reach them is wider than the code comments assume, so recording them together.

### 1. A caller that omits `workspaceId` is unscoped, and approved third-party plugins are in that set

`ownedBy(ownerWorkspaceId, callerWorkspaceId)` in `src/main/browser-session/WebviewCdpManager.ts` returns `true` whenever the caller side is undefined, so `getTarget(surfaceId, undefined)` resolves any registered session — including one owned by a different workspace. On the pipe, that caller side comes from `scopeOf(params)` in `src/main/pipe/handlers/browser.rpc.ts`, which reads `workspaceId` straight out of the request body. Sending nothing is therefore sufficient to opt out of scoping.

The existing comments frame this as the same-user ceiling from #113 — the first-party CLI and a hand-rolled socket client are same-trust anyway. That framing is incomplete:

- `browser.evaluate`, `browser.read`, and `browser.screenshot` are ordinary declarable capabilities in `methodCapabilityMap.ts`. `RESERVED_PREFIXES` in `permissionGrammar.ts` is `['wmux.']`, so nothing stops a third-party plugin from declaring them.
- `PermissionEnforcer.check()` has no workspace dimension at all. It maps a method to a capability and checks the grant; there is no notion of which workspace the calling plugin belongs to, and `RpcContext` carries `origin` / `firstParty` / `clientName` but no owning workspace.

So a plugin the user approved for "browser evaluate" can omit `workspaceId` and run arbitrary JS in, or screenshot, whichever workspace's guest happens to be first in the registration map. The approval the user granted was for a capability, not for a scope, and nothing in the approval text promises workspace confinement.

### 2. `cdpPort` is returned even on a scoped response

`browser.cdp.info` filters `targets` server-side when a caller names its workspace, but `cdpPort` and `shellUrl` are workspace-agnostic and always returned in full — the handler comment says so plainly ("NOT a hard seal"). A caller holding only `browser.read` can take the port and enumerate or drive every page directly over `connectOverCDP`, bypassing the tool layer entirely.

## Why it matters

[...]

PR #845, opened 2026-08-09
fix(cli): scope browser navigation to caller workspace

snowyukitty:
## Summary

Scope `wmux browser navigate` to the verified workspace of the calling wmux pane. Inside wmux, navigation now follows the same self-context contract as `browser open` and `browser close`; outside wmux, the existing active-target fallback remains unchanged.

This prevents an in-session CLI invocation from navigating the first globally registered browser when another workspace happens to register first.

## Changes

- Resolve verified CLI self-context before dispatching `browser.navigate`.
- Send `workspaceId` only when a non-empty verified workspace identity is available.
- Preserve the existing outside-wmux active-browser fallback.
- Add focused CLI coverage and correct the internal browser-tabs contract documentation.

### Fixed

- **Scoped CLI browser navigation to its calling workspace.** `wmux browser navigate` now uses verified pane identity inside wmux instead of potentially navigating the first globally registered browser; invocations outside wmux retain the existing active-target behavior. (#845)

## Stability tier impact

- [x] Change to an `experimental` surface (note in release notes).

n/a — this uses the existing optional `workspaceId` parameter on `browser.navigate`; no RPC, parameter, or wire shape is added. The internal contract documentation is corrected to match the established verified self-context behavior.

## Test plan

- Focused navigate/self-context suite: 2 files, 10 tests passed.
- Full CLI suite: 15 files, 274 tests passed.
- `npx tsc --noEmit`, production-source ESLint, and `npm run build:cli` passed.
- `npm run test:parallel`: 719 files / 10,685 tests passed; 2 files / 24 tests skipped.
- Isolated packaged-app integration: [issue #810 dogfood harness](https://github.com/snowyukitty/wmux/blob/test/810-integration/scripts/issue-810-dogfood.mjs) passed 24/24. With workspace B deliberately registered first, a packaged CLI process launched from workspace A's pane ancestry navigated only A.
- Known unrelated local runtime baseline: 140 passed, 2 failed, 10 skipped. Both failures are in `shellIntegration.runtime.test.ts` on Windows (`node-pty` `AttachConsole failed` / PowerShell timeout); this branch does not touch daemon or shell integration paths.

## Related issues / PRs

Tracks #810. Independent second step of the accepted three-step series.

## Reviewer checklist

### DIFF
--- a/changelog.d/845.md
+++ b/changelog.d/845.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Scoped CLI browser navigation to its calling workspace.** `wmux browser navigate` now uses verified pane identity inside wmux instead of potentially navigating the first globally registered browser; invocations outside wmux retain the existing active-target behavior. (#845)
--- a/docs/internal/browser-tabs-workspace-contract.md
+++ b/docs/internal/browser-tabs-workspace-contract.md
@@ -395,13 +395,20 @@ only the intended surface through the full sequence.
 
 ## 8. Adjacent findings and explicit non-goals
 
-### 8.1 Omitted-`surfaceId` navigation remains a separate audit
+### 8.1 Omitted-`surfaceId` navigation is caller-workspace scoped
 
-Static inspection found that `browser_navigate` and `browser_navigate_back` send `browser.navigate` / `browser.goBack` without a workspace ID when their optional `surfaceId` is omitted. Main then calls `WebviewCdpManager.getTarget(undefined)`, whose documented fallback is the first globally registered target.
+Bundled MCP `browser_navigate` and `browser_navigate_back` already resolve the
+connection's workspace and route `browser.navigate` / `browser.goBack` through
+`sendScopedBrowserRpc`, even when their optional `surfaceId` is omitted. The
+remaining gap was the standalone `wmux browser navigate` command: it sent only
+the URL, so main could fall back to the first globally registered target.
 
-That path is not fixed merely by making `browser_tabs select` workspace-scoped. Until a separate audit resolves the default-target contract, callers should pass the `surfaceId` returned by `browser_tabs` to subsequent browser operations.
-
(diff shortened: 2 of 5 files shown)


openwong2kim:

This PR is the independent second step of the #810 series. It scopes `wmux browser navigate` to the verified workspace of the calling wmux pane, so an in-session CLI invocation no longer falls back to the first globally registered browser target when another workspace happens to register first. The CLI now resolves its parent-process self-context the same way `wmux open` and `wmux browser close` already do, sends `workspaceId` only when a non-empty verified identity is available, and preserves the existing active-target fallback for invocations outside wmux.

The change is small and well-contained: the handler grows a `resolveSelfContext` call before dispatch, the help text is corrected to describe the scoped behavior, the browser-tabs workspace contract section 8.1 is rewritten to reflect that the omitted-`surfaceId` navigation gap is now closed at the CLI layer rather than remaining an open audit item, and a new focused test file pins both the caller-workspace routing and the outside-wmux active-target fallback. It closes a real cross-workspace footgun (a background workspace navigating whichever browser registered first) without changing any RPC shape or main-process behavior.

Priority P3 — this is a focused fix to an experimental CLI surface with no contract impact, clear test coverage, and a narrow blast radius. It is correct and worth merging as part of the #810 stack, but the impact is smaller than #844 since it only affects the standalone CLI path that bundled MCP tools already scoped correctly.

openwong2kim:
Reproduced the bug and verified the fix on Windows.

Setup: workspaces A and B, each with a browser, B's registered first globally. Then `wmux browser navigate` from a pane in A.

- main: A's browser stayed put, B's navigated, and the CLI still printed `Navigated to: ...` — a silent cross-workspace hijack.
- this branch: A's navigated, B untouched.

Also checked the case where the calling workspace has no browser surface but another one does. It fails closed with `BROWSER_NOT_OWNED: ... Do not retry — address a surface from this workspace instead.` and leaves the other workspace alone.

PR #853, opened 2026-08-10
Let the renderer frame a wmux-plugin: page

### LINKED ISSUE #848: [Bug]: Panel enters restore loop when plugin is enabled
### What happened?

When I install and enable the `plugins/hello-panel` folder, the panel gets stuck in a restore loop.
I confirmed that this issue is resolved by adding the `wmux-plugin` schema to `cspPolicy` on line 162 of `src\main\window\createWindow.ts`.

### Steps to reproduce

1. Launch wmux
2. Click `Hello Panel` in the sidebar
3. Click `approve to enable`
4. Click the `Approve` button
5. The panel restoration loops

### Expected behavior

No looping.

### wmux version

3.40.1

### How did you install wmux?

Setup.exe (direct download)

Command Prompt (cmd)

### Windows version

Windows 11 25H2 (build 26200.8973)

None / not relevant

```shell

```

### Confirmation

openwong2kim:
Enabling any UI plugin puts its panel into an endless restore loop in production
builds. `PluginFrame.tsx` points its sandboxed iframe at
`wmux-plugin://<name>/<entry>`, but the renderer's production CSP listed only
`frame-src 'self' https: http:` — so the frame is blocked outright and the panel
retries forever.

Dev builds skip the CSP entirely (Vite's HMR needs `unsafe-eval`), which is why
`scripts/b1-plugin-host-dogfood.mjs` passes against a dev instance and this still
shipped. Moving the policy into its own module makes the directives assertable
without a packaged build.

## Changes

- `src/main/window/csp.ts` — the production policy, with `wmux-plugin:` added to
  `frame-src`.
- `ownsResponseCsp()` — plugin responses keep their own, stricter
  `PLUGIN_PAGE_CSP`. Overwriting it with the renderer policy would both break the
  page (its bundled scripts are not `'self'` to an opaque origin) and relax the
  sandbox, since `PLUGIN_PAGE_CSP` has no `connect-src` at all.
- `src/main/window/__tests__/csp.test.ts` — pins the directives, including the
  negative: the plugin scheme belongs in `frame-src` and nowhere else. Adding it
  to `script-src`/`style-src` would let the renderer document pull code out of a
  plugin bundle.

## Verification

Packaged builds, A/B, with hello-panel installed and an iframe mounted at the
same URL and sandbox `PluginFrame.tsx` uses. Policies read off the wire rather
than from source.

Before — renderer header `frame-src 'self' https: http:`:

```
Framing 'wmux-plugin://hello-panel/' violates the following Content Security
Policy directive: "frame-src 'self' https: http:". The request has been blocked.
```

After — renderer header `frame-src 'self' https: http: wmux-plugin:`, no
violations, and the plugin document loads carrying its own policy
(`default-src 'none'; ...; frame-src 'none'`), which confirms the response guard
holds.

Not covered: the approval-dialog path from unconfirmed plugin to rendered panel
was not driven, since that is a human approval gate. The CSP layer is where both
the fault and the fix are decided.

Closes #848

### DIFF
--- a/changelog.d/853.md
+++ b/changelog.d/853.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Enabling a UI plugin no longer puts its panel into an endless restore loop. The renderer's production Content-Security-Policy did not allow framing the `wmux-plugin:` scheme its sandboxed iframe loads from, so the panel was blocked and retried forever. Development builds set no CSP at all, so this only affected installed releases.
--- a/src/main/window/__tests__/csp.test.ts
+++ b/src/main/window/__tests__/csp.test.ts
@@ -1,0 +1,53 @@
+import { describe, it, expect } from 'vitest';
+import { buildProductionCsp, ownsResponseCsp } from '../csp';
+import { PLUGIN_PROTOCOL_SCHEME } from '../../../shared/pluginHost';
+
+/**
+ * #848 — enabling a UI plugin put its panel into an endless restore loop in
+ * production builds only, because the renderer's `frame-src` did not name the
+ * `wmux-plugin:` scheme that PluginFrame.tsx points its iframe at. Dev builds
+ * skip the CSP entirely, so nothing caught it before release.
+ */
+
+function directive(csp: string, name: string): string {
+  const found = csp.split(';').map((d) => d.trim()).find((d) => d.startsWith(`${name} `));
+  if (!found) throw new Error(`no ${name} directive in: ${csp}`);
+  return found;
+}
+
+describe('buildProductionCsp', () => {
+  it('lets the renderer frame a plugin page', () => {
+    expect(directive(buildProductionCsp(), 'frame-src')).toContain(`${PLUGIN_PROTOCOL_SCHEME}:`);
+  });
+
+  it('does not let the renderer document itself load plugin code', () => {
+    // The plugin scheme belongs in frame-src and nowhere else: a plugin
+    // bundle is a framed document, never a source for the host document.
+    const csp = buildProductionCsp();
+    for (const name of ['script-src', 'style-src', 'connect-src', 'img-src', 'font-src']) {
+      expect(directive(csp, name)).not.toContain(PLUGIN_PROTOCOL_SCHEME);
+    }
+  });
+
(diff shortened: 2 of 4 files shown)


openwong2kim:

This PR fixes #848, where enabling any UI plugin put its panel into an endless restore loop in production builds. The root cause is that `PluginFrame.tsx` points its sandboxed iframe at `wmux-plugin://<name>/<entry>`, but the renderer's production CSP allowed only `'self'`, `https:`, and `http:` in `frame-src` — so Chromium blocked the approved plugin frame outright and the load-recovery path retried forever. Development builds skip the CSP entirely (Vite's HMR needs `unsafe-eval`), which is why this shipped unnoticed.

The fix cleanly moves the policy into `src/main/window/csp.ts` as `buildProductionCsp()` with `wmux-plugin:` added to `frame-src` only, and adds `ownsResponseCsp()` to prevent the renderer policy from being stamped onto plugin bundle responses (which carry their own stricter `PLUGIN_PAGE_CSP` with no `connect-src`). The directive-level test in `csp.test.ts` pins both the positive (`frame-src` includes the scheme) and the negative (the scheme does not leak into `script-src`, `style-src`, `connect-src`, `img-src`, or `font-src`), which is the right guard since adding it anywhere else would let the renderer document pull code out of a plugin bundle.

Priority is P2 — the bug blocks every UI plugin in production but is recoverable and not data-destructive. Worth noting that this overlaps with #852 (snowyukitty), which addresses the same #848 root cause with the same `frame-src` change; the two PRs should be reconciled before merge so the winner carries both the `PLUGIN_PROTOCOL_SCHEME` constant reuse and the `ownsResponseCsp` response guard.

openwong2kim:
Closing this — #852 was already open for the same issue and got there first, I just didn't spot it before I started. Same call on the fix in both (scheme in `frame-src` only, negative test to keep it out of the other directives), so there's nothing here worth keeping over that one.

I've moved the packaged A/B results over to #852. The one thing my version had on top was a guard to stop the session-wide `onHeadersReceived` from overwriting the plugin response's own CSP — measuring it showed the handler never fires for a `protocol.handle` scheme, so that guard was dead code. Good thing it didn't ship.

PR #854, opened 2026-08-10
Own the application menu instead of inheriting Electron's

### LINKED ISSUE #818: wmux does not own its application menu, so Electron's defaults own the accelerator table
`Menu.setApplicationMenu()` is never called, so Electron installs its default
menu. That menu's roles own their accelerators, and on macOS they are NSMenu
key equivalents — dispatched before the web contents see the key, so a
renderer `preventDefault()` cannot take them back.

wmux is a terminal multiplexer that defines its own keymap (thirteen
`Cmd+Shift` bindings alone: `W D Q ] [ A U R L X H O G`). It is currently
building that keymap on top of someone else's accelerator table.

## Three symptoms, one cause

**1. `Cmd+Shift+R` destroys attached remote workspaces.** It is the default
menu's **Force Reload**, not wmux's rename shortcut. The reload wipes the
renderer store, and `remoteWorkspaces[]` is per-app-run state
(`remoteWorkspacesSlice.ts`), so every attached remote row disappears from the
sidebar and has to be re-attached from scratch. Reported from dogfooding
3.39.0.

**2. The rename-workspace shortcut is dead on macOS.** `useKeyboard.ts:679`
dispatches `wmux:rename-workspace` on `Cmd+Shift+R`, and `SettingsPanel`
advertises it in the shortcut list. The menu takes the key first, so the
documented shortcut has never worked on macOS.

**3. Paste corruption — already hit, already worked around.** `Cmd+V` leaks to
the default menu's paste role, so xterm's native paste path and wmux's async
IPC path write to the same pty and race. See the comment at
`useTerminal.ts:962`, which records a two-pass investigation and installs a
capture-phase guard. That guard treats the symptom at one call site; the cause
is this issue.

`Cmd+R` (plain Reload) is also live and equally destructive, just less likely
to be hit by accident.

## Fix

Define the application menu and own the accelerators. Reload and Force Reload
should be dropped or confined to development builds.

## What makes this bigger than it looks

- Three platforms. macOS needs the app menu, Window, and Services structure;
  Windows/Linux dispatch accelerators renderer-first, so their behaviour
  changes differently.
- The thirteen existing `Cmd+Shift` bindings need auditing against whatever
  menu is defined, plus the non-shift ones.
- `useTerminal.ts:962`'s guard is written against the assumption that the
  default menu is present. If this lands, that assumption changes and the
  guard should be re-evaluated (its comment explains why it is macOS-only and
  why enabling it on Linux would misfire) — but removing it is a separate,
  separately verified change, not part of this one.

Per the repo's review policy this is an infrastructure change and takes a
three-model panel before implementation.

## Not this issue

Persisting attached remote workspaces so they survive a reload is a different
fix for a different cause: `mainWindow.reload()` is also called by the renderer
crash and unresponsive recovery paths (`main/index.ts:596`, `:610`, `:1689`),
which no menu change affects. Worth doing on its own merits; it does not
address anything above.

openwong2kim:
`Menu.setApplicationMenu()` was never called, so Electron's default menu owned
the accelerator table. wmux is a multiplexer with its own keymap, and it was
building that keymap on top of someone else's accelerators.

On macOS the default menu's roles are NSMenu key equivalents: they dispatch
before the web contents see the key, so a renderer `preventDefault()` cannot take
them back. `Cmd+Shift+R` was Force Reload rather than rename-workspace, and the
reload wipes `remoteWorkspaces[]` — per-app-run state, so every attached remote
workspace vanished. On Windows/Linux the renderer wins the combos it binds, so
the exposure there is the ones it does not: `Ctrl+R` reloads whenever focus sits
outside a terminal, with the same data loss.

## Changes

- `src/main/menu/appMenu.ts` — wmux's own menu, installed before the first window
  so no window is ever briefly governed by the default table. The rule: declare
  every accelerator we keep, and none that `WMUX_KEYMAP` reserves. Entries we
  want for discoverability but not for their key (Close Window, Minimize) use a
  custom `click` rather than a `role`, because a role silently re-installs its
  default key equivalent.
- `src/shared/keymap.ts` — the combo list was hand-copied into three places and
  nothing kept them in sync. `SettingsPanel`'s copy had already drifted:
  `Ctrl+Shift+A`, `Ctrl+Shift+G`, `Ctrl+M`, `Ctrl+Tab` and the zoom keys were all
  bound but missing, so a custom keybinding on one of them got no conflict
  warning and then silently never fired. `useKeyboard.ts` still expresses its
  bindings as an if-chain; converting that is deliberately left out of this
  change.
- `src/main/__tests__/appMenu.template.test.ts` — fails CI if a menu item ever
  declares a reserved combo again. It resolves roles through a table of the
  accelerators Electron attaches implicitly, because a walk that only reads
  declared accelerators sees `undefined` on every role and passes vacuously —
  precisely the blind spot that let the default menu own these keys.
- `scripts/probe-menu-role-accelerators.js` — that table is read out of a running
  Electron rather than hand-copied, and the test fails when Electron's major
  drifts from the version it was probed against.

## Verification

Packaged builds, A/B, dumping `Menu.getApplicationMenu()` from the main process.

Before — Electron's default menu, holding every combo the issue named:

```
View   ▸ Reload             accel=CmdOrCtrl+R
       ▸ Force Reload       accel=Shift+CmdOrCtrl+R
       ▸ Actual Size/In/Out accel=CommandOrControl+0 / +Plus / +-
Window ▸ Minimize           accel=CommandOrControl+M
       ▸ Close              accel=CommandOrControl+W
```

After, all six are gone: View carries only Toggle Full Screen, File only Exit,
and Minimize declares no accelerator. `packaged: true` shows no Developer
submenu, so the dev-only Reload does not reach release builds. The probed role
table matched the live dump, including `quit` having no accelerator on Windows.

Not covered: macOS was not exercised — the `Cmd+Shift+R` data-loss path is
macOS-specific and this was verified on Windows. The dump does confirm its cause
directly, since the default menu demonstrably owns that accelerator.

The `Cmd+V` paste-corruption race guarded at `useTerminal.ts:962` is untouched:
Edit keeps `role: 'editMenu'`, so that guard's premise still holds. Re-evaluating
it is a separate change.

Closes #818

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed application-menu shortcuts being intercepted by Electron defaults across platforms.
  - Prevented conflicts between built-in menu actions and WMUX workspace, surface, reload, and terminal zoom shortcuts.
  - Improved detection of custom shortcut conflicts in Settings.

- **Improvements**
  - Standardized shortcut labels and macOS formatting.
  - Expanded the displayed shortcut list and aligned it with the active platform’s built-in bindings.
  - Ensured shortcut handling is active during application startup.

### DIFF
--- a/changelog.d/854.md
+++ b/changelog.d/854.md
@@ -1,0 +1,4 @@
+### Fixed
+
+- wmux now defines its own application menu instead of inheriting Electron's default one, which had been quietly owning shortcuts wmux binds. On macOS `Cmd+Shift+R` triggered Force Reload rather than renaming a workspace, and the reload dropped every attached remote workspace; `Cmd+W` closed the window instead of the active surface, and the zoom keys resized the UI instead of the terminal font. On Windows and Linux `Ctrl+R` no longer reloads when focus sits outside a terminal.
+- Settings → Shortcuts lists the built-in key bindings from one shared table, so a custom keybinding that collides with `Ctrl+Shift+A`, `Ctrl+Shift+G`, `Ctrl+M`, `Ctrl+Tab` or the zoom keys is now flagged as a conflict instead of being accepted and silently never firing.
--- a/scripts/probe-menu-role-accelerators.js
+++ b/scripts/probe-menu-role-accelerators.js
@@ -1,0 +1,57 @@
+/**
+ * Read each Electron menu role's default accelerator out of the Electron build
+ * that is actually installed, and print it as JSON.
+ *
+ * `src/main/menu/appMenu.ts` keeps a copy of this table (#818): a template walk
+ * cannot see a role's accelerator — the item's own `accelerator` field is
+ * undefined — so the collision test that keeps the application menu off wmux's
+ * keymap has to know what each role installs. A hand-copied table goes stale
+ * silently on an Electron upgrade, hence this probe plus the
+ * PROBED_ELECTRON_MAJOR tripwire in appMenu.template.test.ts.
+ *
+ * Usage (from the repo root):
+ *
+ *   npx electron scripts/probe-menu-role-accelerators.js
+ *
+ * Then reconcile the output with ROLE_DEFAULT_ACCELERATORS and bump
+ * PROBED_ELECTRON_MAJOR. Note that Electron reports platform-agnostic strings
+ * (`CommandOrControl+W`), and that a few roles genuinely differ per platform —
+ * run the probe on each OS you care about, or trust the per-platform keys
+ * already recorded.
+ */
+const { app, Menu } = require('electron');
+
+const ROLES = [
+  'about', 'hide', 'hideOthers', 'unhide', 'quit', 'close', 'minimize', 'zoom', 'front',
+  'undo', 'redo', 'cut', 'copy', 'paste', 'pasteAndMatchStyle', 'delete', 'selectAll',
+  'reload', 'forceReload', 'toggleDevTools', 'resetZoom', 'zoomIn', 'zoomOut',
+  'togglefullscreen',
+];
+
+const COMPOSITES = ['appMenu', 'fileMenu', 'editMenu', 'viewMenu', 'windowMenu'];
+
+app.whenReady().then(() => {
+  const out = { platform: process.platform, electron: process.versions.electron, roles: {}, composites: {} };
+
+  for (const role of ROLES) {
+    try {
+      const menu = Menu.buildFromTemplate([{ role }]);
+      out.roles[role] = menu.items[0] ? menu.items[0].accelerator || null : 'ROLE_REJECTED';
+    } catch (err) {
+      out.roles[role] = `ERROR: ${err.message}`;
+    }
+  }
+
+  for (const composite of COMPOSITES) {
+    try {
+      const menu = Menu.buildFromTemplate([{ role: composite }]);
+      const sub = menu.items[0] && menu.items[0].submenu ? menu.items[0].submenu.items : [];
+      out.composites[composite] = sub.map((i) => `${i.role || i.label || '(separator)'}=${i.accelerator || '-'}`);
+    } catch (err) {
+      out.composites[composite] = `ERROR: ${err.message}`;
+    }
+  }
+
+  console.log(JSON.stringify(out, null, 2));
+  app.exit(0);
+});
--- a/src/main/__tests__/appMenu.template.test.ts
+++ b/src/main/__tests__/appMenu.template.test.ts
@@ -1,0 +1,260 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+
+/**
+ * #818 — wmux owns its accelerator table.
+ *
+ * These lock the two things that can silently regress:
+ *
+ *   1. The template declares no accelerator that WMUX_KEYMAP reserves. The
+ *      assertion runs through `keymapCollisions`, which resolves a `role` to
+ *      the accelerator Electron attaches to it. Walking only the declared
+ *      `accelerator` fields would pass vacuously — every role item has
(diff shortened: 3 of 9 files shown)


openwong2kim:

This PR resolves #818 by having wmux define and install its own application menu via `Menu.setApplicationMenu()` before the first window is created, instead of inheriting Electron's default menu. On macOS that default menu's NSMenu key equivalents were firing before the renderer — `Cmd+Shift+R` force-reloaded (wiping every attached remote workspace) instead of renaming the workspace, and `Cmd+W` closed the window rather than the surface — while on Windows/Linux `Ctrl+R` reloaded whenever focus was outside a terminal with the same data-loss effect.

The approach is sound and goes beyond a minimal fix: a new canonical `WMUX_KEYMAP` table in `src/shared/keymap.ts` replaces the hand-copied key lists that had already drifted in `SettingsPanel` (missing `Ctrl+Shift+A`, `Ctrl+Shift+G`, `Ctrl+M`, `Ctrl+Tab`, and the zoom keys), and the regression test in `appMenu.template.test.ts` resolves role accelerators through a probed table rather than reading only declared `accelerator` fields — which is exactly the blind spot that let the default menu own these keys undetected. The `PROBED_ELECTRON_MAJOR` tripwire and `probe-menu-role-accelerators.js` script make an Electron-upgrade drift loud instead of silent, and the edit-menu role is deliberately preserved so the existing `useTerminal.ts:962` paste guard's premise stays intact.

Priority is P1 because the primary symptom is unrecoverable data loss (remote workspaces are per-app-run state and must be re-attached manually), and the fix also resurrects the rename-workspace shortcut that has been dead on macOS. The one gap to flag for review: macOS was not exercised in verification (only Windows), so the `Cmd+Shift+R` data-loss path is confirmed by inspection of the dumped menu rather than by a live repro on the affected platform.

openwong2kim:
@codex review

openwong2kim:
All four are real, thanks — pushed fixes for each.

**Menu installed after an await.** This was the one that mattered: the call sat after `await loader.loadAll()`, and `app.on('activate')` is registered at module scope, so once `ready` fires macOS can dispatch an activate while the callback is still parked. That handler builds a window under the default menu — the exact path the PR exists to close, left open during startup. Moved above the first await.

**Directional keys.** Right, and it is the same bug the table was meant to kill, one layer down. `e.key` is `ArrowUp`; rows said `Up`, so the exact `Set.has` never matched a single arrow binding. Rows now use the `Arrow*` spelling. The accelerator side did not need touching — `normalizeAcceleratorKey` already folds `arrowup` onto Electron's `Up`, and there is now a test pinning that so the two spellings cannot drift apart again.

**Zoom aliases.** Confirmed: the in/out handlers take `=`/`+`/`Equal`/`NumpadAdd` and the minus family without checking shift, so `Ctrl+Shift++` and friends never reach a custom binding. Added the shifted and `+`/`_` forms. Reset does require `!shift`, so it keeps no shifted alias.

**Platform-resolved modifiers.** Also right, and it is the inverse failure — a false warning rather than a missing one. Custom bindings match literal `e.ctrlKey` everywhere while a non-`literalCtrl` row fires on Command under macOS, so a custom `Ctrl+Shift+A` never meets the built-in there. Replaced the flat set with `builtinCombosFor(platform)`, which keeps the `literalCtrl` rows and the Ctrl-less ones (`Alt+ArrowUp`) on macOS and drops the rest.

New tests in `src/shared/__tests__/keymap.test.ts` cover all three storage-form cases plus the platform split, and assert the round trip against a copy of `formatKeyCombo` so a future rename on either side breaks the build.

Re-verified the packaged build after the move: menu dump is unchanged, so relocating the call did not disturb what gets installed. The activate-during-startup window itself is macOS-only and was not reproduced here.

PR #855, opened 2026-08-10
Read the registry PATH as UTF-16 so non-ASCII entries survive

### LINKED ISSUE #849: Chinese paths in Windows PATH get mangled to U+FFFD, breaking command resolution in wmux terminals
### What happened?

When a directory in the Windows PATH contains non-ASCII (Chinese) characters — e.g. `D:\软件\Python312` — wmux terminals fail to resolve executables from that directory. `which python` returns the Microsoft Store stub (`...\WindowsApps\python`) instead of the real Python, even though the real path is present in PATH and works fine in Windows Terminal / mintty.

In the wmux terminal's `$PATH`, the Chinese path appears as U+FFFD replacement characters:
```
16:/d/����/Python312        # mangled (does not exist on disk)
29:...\WindowsApps           # next resolvable python -> MS Store stub
33:/d/软件/Python312        # the real path, but too late in PATH
```

The registry stores the path correctly (`软件` = U+8F6F U+4EF6). But wmux mangles it to U+FFFD when initializing bash's PATH. Since the mangled path doesn't exist on disk, bash skips it and resolves to a later `WindowsApps` entry — a broken Microsoft Store stub that exits 49.

### Steps to reproduce

1. Have a directory with a Chinese name in Windows PATH (e.g. `D:\软件\Python312`)
2. Open a terminal in wmux (Git Bash)
3. Run `which python` → returns `...\WindowsApps\python` (wrong)
4. Same command in Windows Terminal's Git Bash → returns the correct path (works)

### Expected behavior

`which python` should return the real Python path (`/d/软件/Python312/python`) in wmux terminals, matching Windows Terminal behavior. Chinese paths in PATH should be handled correctly so command resolution works the same everywhere.

### wmux version

3.38.9

### How did you install wmux?

winget

PowerShell 7+ (pwsh)

### Windows version

Windows 11 Pro (build 26200)

None / not relevant

```shell

```

### Confirmation

openwong2kim:
A PATH entry like `D:\软件\Python312` reached the shell as `D:\?<FFFD><FFFD>\Python312` — a directory that exists nowhere. bash skipped it and resolved `python` to a later entry, the Microsoft Store stub, which exits 49. The same PATH works in Windows Terminal and mintty, because the mangling was ours.

## Root cause

`readRegistryEnvPath` piped `reg.exe`'s text output and decoded it as UTF-8. `reg.exe` encodes a pipe in the console/ANSI code page. Measured on a CP949 host with a sandbox key:

```
wrote:      "D:\软件\Python312"
raw bytes:  44 3a 5c 3f cb ec 5c ...
            D  :  \  ?  <件>    \
read utf8:  "D:\?<FFFD><FFFD>\Python312"
```

Two losses, in order: `软` is absent from that code page and is already `?` (0x3F) when Node receives it, and the surviving `件` (0xCB 0xEC) is not valid UTF-8, hence U+FFFD. The first loss happens inside reg.exe, so **no choice of decoding recovers it**.

Not CJK-specific — the same probe turned `héllo` into `hello`.

It also explains the duplicate the reporter saw in `$PATH`: `mergeFreshPathWithBase` leads with the fresh registry PATH and appends the process's own entries, and the mangled and intact spellings differ, so the case-insensitive dedup kept both.

## Why CI was green

Every existing test injects `deps.readRegistryPath`. The real reader — the function holding the bug — was never executed by the suite.

```
        withFreshWindowsPath(base, deps)
                 │
        deps.readRegistryPath ?  ──── yes ──▶ fake reader   ← every test
                 │ no
                 ▼
        readRegistryEnvPath()  ← the bug lived here, untested
```

## Change

`reg export` writes UTF-16LE, so no code page is involved. Measured against the alternatives on a REG_EXPAND_SZ value (the real type of `HKCU\Environment\Path`):

| candidate | round-trips? | cost |
|---|---|---|
| current: `reg query` piped as utf8 | FAIL | 29ms |
| **`reg export` → UTF-16LE file** | **PASS** | 33ms |
| PowerShell `Get-ItemProperty` | PASS | slow, ConstrainedLanguage risk |

`reg.exe` is kept rather than switching to a cmdlet, so the ConstrainedLanguage immunity this code was written for still holds. The 4ms sits behind the existing 5s cache.

- `parseRegExportValue` is pure and unit-tested. It reads the **root section only** — `reg export` includes subkeys, and a subkey's `Path` is not this value. It returns `null` for anything it cannot decode with certainty, because a malformed export must fail open to the caller's PATH rather than yield a partial one that silently drops entries.
- `.reg` escaping is decoded as `.reg`, not as JSON: only `\` and `\"`. A literal backslash-n in a path stays two characters.
- The UTF-16LE BOM is verified rather than assumed. Decoding text as the wrong encoding is the exact failure this PR fixes.
- Whole-key export removes the implicit size bound `query /v Path` had, and this is synchronous I/O on the pty path, so there is a size ceiling. Temp file gets a `crypto.randomBytes` name and `/y` (so a leftover file can never turn this into an overwrite prompt that blocks to the timeout), and its cleanup is swallowed so a failed unlink can neither mask a good result nor convert a caught failure into a throw.
- `readRegistryEnvPath` is exported so a test can drive the real reader.

## Verification

Both halves of the chain, measured:

- **registry → string**: a live test writes a non-ASCII `REG_EXPAND_SZ` value to a scratch key under `HKCU\Software` (never `HKCU\Environment`), calls the real reader, and asserts a byte-exact round trip.
- **env → node-pty → shell**: a real directory named `软件`, an executable inside it, passed through PATH — `which` returned `/tmp/.../软件/Python312/wmux849tool` intact and the tool ran. No loss in that half, so the registry read was the only break.

20 tests pass (8 parser failure cases: subkey-only `Path`, malformed hex, odd-length hex, truncated continuation, `hex(7)`, unterminated string, wrong shape, missing value). `tsc --noEmit` clean.

Reviewed by plan-eng-review plus a three-way independent pass (Claude / Codex / GLM-5.2). No P1 or P2 against this change; the two P3s that were actionable are fixed in the second commit.

## Not in scope

The daemon replays a persisted create-time env verbatim on reboot, so a session created before this fix keeps the mangled PATH until a pane is recreated. New panes are correct immediately. Touching daemon persistence carries its own regression risk and wants its own verification.

Closes #849

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Windows terminal environment handling for non-ASCII `PATH` directories, including CJK characters and accented Latin characters.
  - Improved compatibility with paths affected by code-page encoding and replacement characters.
  - Added safer handling for malformed or unavailable registry data.

### DIFF
--- a/changelog.d/855.md
+++ b/changelog.d/855.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Directories with non-ASCII names in `PATH` now resolve in wmux terminals. A path such as `D:\软件\Python312` was being read from the registry through a code-page-encoded pipe, which replaced characters the code page could not represent and left the rest as `U+FFFD`, so the directory matched nothing on disk and commands from it resolved to a later `PATH` entry instead — for Python, the Microsoft Store stub. Accented Latin characters were affected the same way.
--- a/src/shared/__tests__/windowsPathEnv.test.ts
+++ b/src/shared/__tests__/windowsPathEnv.test.ts
@@ -1,13 +1,145 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest';
+import { execFileSync } from 'child_process';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
 import {
   expandPercentRefs,
   composeRegistryPath,
   mergeFreshPathWithBase,
   withFreshWindowsPath,
   resetFreshPathCacheForTests,
+  parseRegExportValue,
+  readRegistryEnvPath,
 } from '../windowsPathEnv';
 
 beforeEach(() => resetFreshPathCacheForTests());
+
+/** Build a `.reg` export body around one root section. */
+function regFile(section: string, extra = ''): string {
+  return [
+    'Windows Registry Editor Version 5.00',
+    '',
+    '[HKEY_CURRENT_USER\\Environment]',
+    section,
+    extra,
+    '',
+  ].join('\r\n');
+}
+
+/** UTF-16LE bytes of `s` (NUL-terminated) as a `.reg` hex(2) token list. */
+function hexBytes(s: string): string {
+  const buf = Buffer.from(`${s}\0`, 'ucs2');
+  return [...buf].map((b) => b.toString(16).padStart(2, '0')).join(',');
+}
+
+/**
+ * #849 — the registry read used to pipe reg.exe text and decode it as UTF-8,
+ * but reg.exe encodes a pipe in the console code page: `D:\软件\Python312` came
+ * back as `D:\?<FFFD><FFFD>\Python312`, a path that exists nowhere, so command
+ * resolution fell through to a later PATH entry. Reading a UTF-16LE `.reg`
+ * export instead keeps every byte. These cases pin the parser's contract:
+ * decode exactly what it can prove, return null for everything else.
+ */
+describe('parseRegExportValue', () => {
+  it('round-trips non-ASCII through hex(2), which is what a user PATH is', () => {
+    const value = 'D:\\软件\\Python312;D:\\héllo;%SystemRoot%\\System32';
+    expect(parseRegExportValue(regFile(`"Path"=hex(2):${hexBytes(value)}`), 'Path')).toBe(value);
+  });
+
+  it('joins backslash-continued hex lines', () => {
+    const value = 'C:\\a;C:\\b';
+    const tokens = hexBytes(value).split(',');
+    const split = `"Path"=hex(2):${tokens.slice(0, 4).join(',')},\\\r\n  ${tokens.slice(4).join(',')}`;
+    expect(parseRegExportValue(regFile(split), 'Path')).toBe(value);
+  });
+
+  it('decodes REG_SZ escaping, which is not JSON escaping', () => {
+    // A literal backslash-n in a path must stay two characters.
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR fixes #849, where PATH entries containing non-ASCII characters (CJK, accented Latin) were mangled to U+FFFD because `readRegistryEnvPath` piped `reg.exe` text output — encoded in the console/ANSI code page — and decoded it as UTF-8, turning e.g. `D:\软件\Python312` into a path that exists nowhere on disk and leaving command resolution to fall through to the Microsoft Store `python` stub. The fix replaces `reg query` to a pipe with `reg export` to a UTF-16LE file, eliminating the code page entirely, and adds a pure, well-tested `parseRegExportValue` parser that reads only the root section, decodes `.reg` escaping correctly (not JSON), and returns `null` — failing open to the caller's existing PATH — for anything malformed rather than yielding a partial result that silently drops entries.

The change touches `src/shared/windowsPathEnv.ts` (the reader, now exported) plus 20 new tests in `windowsPathEnv.test.ts`, including a live `reg.exe` round-trip against a scratch `HKCU\Software` key and a node-pty end-to-end check confirming a real `软件` directory resolves. I'm labelling this **P1**: it is a correctness fix for a data-loss path on the pty-create path, but the design is conservative — it keeps `reg.exe` (preserving ConstrainedLanguage immunity), sits behind the existing 5s cache, and a three-model review found no P1/P2 issues. Worth merging once a final reviewer confirms the temp-file handling and `MAX_EXPORT_BYTES` ceiling are acceptable on the synchronous pty-create path.

PR #856, opened 2026-08-10
fix(daemon): make pushed control-pipe events opt-in

### LINKED ISSUE #659: control pipe multiplexes broadcast events with RPC responses, with no subscription and no warning — naive clients report phantom failures with empty error messages

openwong2kim:
## Summary

The daemon control pipe delivered broadcast events to every socket from the moment it connected, with no subscription step. A client that writes a request and reads one line back — the obvious implementation — therefore read an event instead of its reply whenever one happened to be pushed in between. An event frame carries no `ok` and no `error`, so such a client reported a failure with an empty error message and discarded the real reply when it arrived.

The asymmetry is what made it expensive: the mistake can fabricate failures but never successes, so it does not look like a client bug. It looks like the daemon intermittently failing whatever was asked of it, with no error text to go on. Two teams published root-cause analyses off the back of it and had to retract them.

`#662` documented the `id`-correlation rule, which was the cheap half of the fix. This is the design half: events are now opt-in, so the default is safe for a client that never reads the documentation.

## What changed

**`DaemonPipeServer`** — `broadcast()` now writes to an `eventSubscribers` set instead of every connected socket. `subscribeEvents` / `unsubscribeEvents` / `isEventSubscriber` manage it. The subscription is per-connection and dies with the socket (close, `stop()`, and `rotateToken()` all clear it), matching the lifetime rule `firstPartyClients` already uses, so a reconnecting client must subscribe again.

**RPC** — `daemon.events.subscribe` and `daemon.events.unsubscribe`, registered alongside the existing `daemon.transcript.subscribe` pair they are modelled on.

**`DaemonClient`** — the app is the one consumer that wants events, and it subscribes as the first line it writes on a fresh control pipe. That placement is deliberate: the daemon processes lines in order, so every event generated after it reads that line is delivered. The call is not awaited — waiting for the reply would only widen the gap.

**`docs/PROTOCOL.md` §2.9** — rewritten. The `id`-correlation rule still stands for subscribers; the section now leads with the opt-in and carries a note for anyone debugging against a daemon that predates it, where every connection is still implicitly subscribed.

## Compatibility

This changes the default for control-pipe clients: one that relied on events arriving unsolicited must now send `daemon.events.subscribe`. In-tree, nothing else does — the CLI (`src/cli/client.ts`) and the MCP server (`src/mcp/wmux-client.ts`) both correlate on `id` and read no events, so both are unaffected and both gain protection from the failure above. An older daemon answers `Unknown method` and keeps pushing to everyone as it always did, so a new app against an old daemon is unchanged.

## Test coverage

`src/daemon/__tests__/DaemonPipeServer.eventOptIn.test.ts`, six cases. The one that pins the fix is the first: a connected-but-unsubscribed client sends a request after a broadcast fires and receives exactly one line, its own reply. The rest cover delivery after subscribe, unsubscribe, isolation between a subscriber and an RPC-only client on the same server, subscription death on socket close, and refusing to subscribe a client that is already gone.

## Test plan

- [x] `tsc --noEmit` clean
- [x] `src/daemon` + `src/main`: 5854 passed, 0 failed (post-merge with `main`)
- [x] New opt-in suite: 6 passed

Closes #659

## Summary by CodeRabbit

* **New Features**
  * Added opt-in daemon event delivery for control connections.
  * Clients can subscribe or unsubscribe from pushed events at any time.
  * Event subscriptions are isolated per connection and cleared when connections close or reconnect.
  * The client automatically enables event delivery when supported.
* **Documentation**
  * Clarified event subscription, response correlation, pushed-event routing, and compatibility with older daemon versions.
* **Tests**
  * Added coverage for subscription, unsubscription, isolation, disconnect handling, and event broadcasting.

## Review round

A three-model review of this diff (Claude, Codex, GLM) found one real breakage the original change missed, plus two blind spots it created. All three are fixed in `4651f088`.

**Three in-repo dogfood gates consumed broadcasts without subscribing** — `scripts/hooks-daemon-dynamic.mjs`, `scripts/bridge-dogfood.mjs`, and `scripts/d1-osc-notification-dynamic.mjs` each open a dedicated connection to collect event frames. After the switch they collected nothing: wait asserts would burn their timeouts and "no event should arrive" asserts would pass vacuously. These are manual gates, not part of any vitest suite, which is why CI stayed green. Fixed by subscribing on each listener.

**`subscribeToEvents` discarded its own result** — the RPC envelope's `ok` only reports that the call was dispatched; the handler's answer rides in `result`, so a refusal read as a success. Worse, any transient failure was swallowed: the subscribe shares the daemon's RPC rate limit with every other client, so a boot-time burst could bounce it once and leave the app with no events for the life of the connection, nothing logged and nothing to retry. Now checks the handler's answer and retries with backoff, honoring `Unknown method` as terminal.

**A daemon with zero subscribers cannot tell correct clients from an old app** — an RPC-only CLI and an app too old to subscribe look identical from inside. The daemon now logs when a client identifies as `main` and still has not subscribed shortly after. Only the app identifies as `main`, so the CLI and MCP server cannot trip it.

Deferred to #858, none of them regressions from this PR: the accept-to-subscribe window (flagged by all three models), the missing backpressure guard on `broadcast` that `sendTo` already has, and whether `daemon.events.subscribe` should carry a first-party gate.

Also verified: `npm run test:parallel` excludes `*.runtime.test.ts`, so the runtime suite was run separately — 138 passed.

### DIFF
--- a/changelog.d/856.md
+++ b/changelog.d/856.md
@@ -1,0 +1,12 @@
+### Changed
+
+- **The daemon control pipe no longer pushes events at clients that never asked for
+  them.** Events used to reach every connection from the moment it opened, so a tool
+  that wrote a request and read one line back could read an event instead of its
+  reply — and since an event frame carries no error text, report a failure with an
+  empty message while quietly dropping the real answer. Events are now opt-in:
+  a connection gets replies and nothing else until it calls
+  `daemon.events.subscribe`. Tools that only make requests, including the wmux CLI
+  and the MCP server, need no change and can no longer hit that failure. A tool that
+  did rely on unsolicited events must now subscribe. See `docs/PROTOCOL.md` §2.9.
+  (#856)
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -252,21 +252,25 @@ Wire shape: `channelId`, `seq` (per-channel monotonic), `senderWorkspaceId`, `re
 
 ### 2.9 The control connection is multiplexed — correlate on `id`
 
-A control connection carries **two kinds of frame in the same newline-delimited JSON stream**: replies to requests you sent, and unsolicited events the daemon pushes to every connected client. There is no subscription step — a client receives pushed events from the moment it connects, whether or not it ever asked for any. See `DaemonPipeServer.broadcast` in `src/daemon/DaemonPipeServer.ts`.
+A control connection can carry **two kinds of frame in the same newline-delimited JSON stream**: replies to requests you sent, and events the daemon pushes. See `DaemonPipeServer.broadcast` in `src/daemon/DaemonPipeServer.ts`.
 
-The two are told apart by `id`:
+**Events are opt-in.** A connection receives replies and nothing else until it calls `daemon.events.subscribe`; `daemon.events.unsubscribe` turns them off again. The subscription is per-connection and dies with the socket, so a client that reconnects MUST subscribe again. A client that only makes requests — the CLI and the MCP server are both in this category — never needs to touch either method, and its stream is then a plain request/reply stream.
+
+Once subscribed, the two frame kinds are told apart by `id`:
 
 - **Replies always echo the `id` of the request they answer**, and always carry `ok`. This holds on every path, including `unauthorized`, `rate limited`, and `Invalid RPC request`. A reply to a request that could not be parsed at all carries `"id": null`.
 - **Pushed events never carry an `id`.** They carry a `type` (e.g. `title.changed`, `lanlink.remote.received`) and their own payload fields.
 
-Therefore a client **MUST**:
+Therefore a **subscribed** client **MUST**:
 
 1. Read frames until it sees one whose `id` matches the `id` it sent — that frame, and only that frame, is its response.
 2. Ignore (or route to an event handler) any frame with no `id`. Never treat one as a response.
 
-**Writing one request and reading exactly one line back is incorrect**, even though it appears to work: it succeeds until an event happens to be pushed between the request and its reply, at which point the client reads the event instead. An event frame has no `ok` and no `error`, so a client that assumes otherwise typically reports a failure with an empty error message and then discards the real reply when it arrives.
+**Once subscribed, writing one request and reading exactly one line back is incorrect**, even though it appears to work: it succeeds until an event happens to be pushed between the request and its reply, at which point the client reads the event instead. An event frame has no `ok` and no `error`, so a client that assumes otherwise typically reports a failure with an empty error message and then discards the real reply when it arrives.
 
 Note the asymmetry when debugging: this mistake can **fabricate failures but never successes**. A phantom failure with an empty error message, on an operation that appears not to have happened, is a symptom of reading the wrong frame — not of the daemon failing the request.
+
+Historical note, because it changes how you read an old bug report: before events were opt-in, every connection was pushed events from the moment it connected, so the failure above could hit a client that had never asked for an event at all. Two teams lost pairing windows to it and published root-cause analyses they had to retract (#659). If you are debugging against a daemon older than this section, assume every connection is subscribed.
 
 ## 3. Snapshot envelope (`pane.list`)
 
--- a/scripts/bridge-dogfood.mjs
+++ b/scripts/bridge-dogfood.mjs
@@ -97,7 +97,12 @@ function broadcastListener() {
   const frames = [];
   const s = net.createConnection(readPipeName());
   let buf = '';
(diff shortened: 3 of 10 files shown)


openwong2kim:

This PR closes #659 by making daemon control-pipe events opt-in rather than broadcasting them to every connected socket on connect. The core change in `DaemonPipeServer.broadcast()` now writes to an `eventSubscribers` set instead of `connectedSockets`, gated behind two new RPC methods (`daemon.events.subscribe` / `daemon.events.unsubscribe`) whose subscriptions are per-connection and die with the socket. `DaemonClient` subscribes as the very first line on a fresh control pipe so the in-tree consumer keeps receiving events, while the CLI and MCP server — which never read events — gain protection from the phantom-failure-with-empty-error-message bug for free.

The protocol documentation in §2.9 is rewritten to lead with the opt-in semantics, preserving the `id`-correlation rule for subscribers and adding a historical note so old bug reports stay interpretable. The six new tests in `DaemonPipeServer.eventOptIn.test.ts` pin the fix where it matters: the default-safe case (an unsubscribed client that sends a request after a broadcast receives exactly one line — its own reply), plus delivery, unsubscribe, isolation between a subscriber and an RPC-only client, socket-close cleanup, and refusal to subscribe a dead client.

This matters because the bug it fixes is asymmetric in a nasty way: it can fabricate failures but never successes, so it presents as the daemon intermittently failing with no error text rather than a client bug — which is precisely the misdiagnosis that cost two pairing teams real debugging time. CI is fully green across Windows/macOS/Linux baselines and the diff is tight (277 additions across six files), so this is safe to merge. One compatibility note worth keeping in mind: any external client that relied on unsolicited events must now call `daemon.events.subscribe`, and an older daemon will answer `Unknown method` to that call while continuing the legacy broadcast-all behavior.

Priority: **P2** — aligns with the #659 issue label; it resolves a real, reproduced protocol-level defect with a well-scoped, well-tested fix.

PR #857, opened 2026-08-10
fix(fleet): prevent non-agent panes from inheriting agent identity (#850)

### LINKED ISSUE #850: Non-agent pane (btop process monitor) misidentified as Claude Code agent in sidebar
### What happened?

After splitting a pane (Ctrl+D), the sidebar / fleet view misidentifies a non-agent pane as a Claude Code agent. Concretely:

- Left pane: Claude Code conversation (legit agent)
- Right pane: plain Git Bash running `btop` (process monitor, NOT an agent)

The sidebar shows TWO "Claude Code" entries — the real one and one for the btop pane, labeled as an agent needing attention ("Agent 6 · Needs you").

Process check confirms only ONE `claude.exe` is actually running (PID 3304), and btop runs as an independent `btop4win.exe`. So the right pane is definitively not a Claude agent, yet wmux labels it as one.

### Steps to reproduce

1. Open wmux with Git Bash as shell
2. Run `claude` in one pane (left)
3. Ctrl+D to split right
4. In the right pane run `btop`
5. Look at the sidebar / fleet view
6. Observe: right pane is shown as a second "Claude Code" agent

### Expected behavior

The right pane (btop) should not be detected as a Claude Code agent. The sidebar should only show the one real Claude Code conversation as an agent.

### wmux version

3.40.2

### How did you install wmux?

winget

PowerShell 7+ (pwsh)

### Windows version

Windows 11 Pro (build 26200)

None / not relevant

```shell

```

### Confirmation

openwong2kim:
## Summary

Fixes #850: non-agent panes (btop, vim, plain shell) misidentified as "Claude Code" agents in the sidebar/Fleet View.

Two independent root causes, two fixes:

- **Fleet selector metadata guard**: `selectFleetPanes` now requires `surfaceAgent[ptyId]` confirmation + `agentName` match before inheriting workspace-level `agentName`/`agentStatus` for the active pane. A non-agent active pane (btop after Ctrl+D split) can no longer borrow the real agent's "Claude Code · Needs you" badge. All three consumer call sites (FleetView, DeckFleet, StatusBar) updated to pass `surfaceAgent`.

- **AgentDetector compound gate**: Claude Code detection now uses a Kiro-style compound gate — requires both a banner signal AND a Claude-specific prompt pattern (`bypass permissions on`, `shift+tab to cycle`, or any approval prompt) before opening the gate. A process monitor showing "claude" in its process list cannot falsely activate detection. Prompt evidence replay preserves the correct status (waiting vs awaiting_input).

## Test plan

- [x] TypeScript type check passes
- [x] 121 unit tests pass (67 fleet + 50 AgentDetector + 4 new)
- [x] New test: non-agent active pane does NOT inherit workspace metadata
- [x] New tests: Claude compound gate — banner alone / prompt alone / both / order-independent / process monitor false positive / per-detector isolation / approval-as-evidence

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented non-agent panes from being incorrectly identified as AI agents in the sidebar, Fleet View, and status displays.
  - Improved detection accuracy by requiring stronger confirmation signals before labeling a pane as Claude or Kiro.
  - Ensured agent names and statuses are attributed only to the correct pane.
  - Improved handling of ANSI formatting, wrapped prompts, approval dialogs, and incomplete terminal output.
  - Prevented agent metadata from appearing on unrelated panes or being inherited from other active sessions.

### DIFF
--- a/changelog.d/850.md
+++ b/changelog.d/850.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Non-agent panes (btop, vim, plain shell) no longer misidentified as AI agents in the sidebar and Fleet View.** Two independent fixes: (1) the Fleet View selector now requires per-PTY agent identity (`surfaceAgent`) before inheriting workspace-level agent name and status for the active pane — a non-agent pane in a split can no longer borrow the real agent's "Claude Code · Needs you" badge; (2) the Claude Code agent detector now uses a compound gate (banner **and** prompt evidence, matching the Kiro CLI pattern) instead of a single banner regex, so a process monitor displaying "claude" in its process list cannot falsely open the detection gate. (#850)
--- a/src/main/pty/AgentDetector.ts
+++ b/src/main/pty/AgentDetector.ts
@@ -144,10 +144,10 @@ const KIRO_PROMPT_LINE = /^[▸>❯]?\s*ask\s*a\s*question\s*or\s*describe\s*a\s
 // Signal A (banner): the startup banner or OSC window-title mention.
 // Signal B (prompt): a Claude-specific prompt fragment proving the TUI is live.
 const CLAUDE_BANNER_RE = /(?<!Open)(?<!Open\s)Claude\s*Code|claude-code|╭.*(?<!Open)(?<!Open\s)Claude/;
-// Any Claude-specific TUI text that a process monitor would never display.
-// Includes waiting prompts AND approval keywords (a pane might show an
-// approval before any waiting footer if the first turn requests file edits).
-const CLAUDE_PROMPT_RE = /bypass permissions on|shift\+tab to cycle|Do you want to proceed|Allow tool use for|Do\s*you\s*want\s*to\s*(?:create|overwrite|make\s*this\s*edit\s*to)/;
+// Claude-specific TUI footer text that a process monitor would never display.
+// Only the idle-prompt fragments — approval keywords are excluded because they
+// appear in conversational text and would false-positive the gate (#850 CI).
+const CLAUDE_PROMPT_RE = /bypass permissions on|shift\+tab to cycle/;
 
 const AGENT_PATTERNS: AgentPattern[] = [
   // ── Claude Code ────────────────────────────────────────────────────────────
@@ -547,14 +547,13 @@ export class AgentDetector {
         const m = clean.match(CLAUDE_PROMPT_RE);
         if (m) {
           this.claudePromptSeen = true;
-          // Preserve the status that matches the evidence pattern so the
-          // replay emits the correct lifecycle event — waiting prompts
-          // replay as 'waiting', approval prompts replay as 'awaiting_input'.
-          const isApproval = /Do you want to proceed|Allow tool use for|Do\s*you\s*want\s*to\s*(?:create|overwrite|make\s*this\s*edit\s*to)/.test(m[0]);
+          // CLAUDE_PROMPT_RE only matches waiting-prompt fragments, so the
+          // replay is always 'waiting'. Approval patterns are excluded from
+          // the gate evidence to avoid conversational false positives.
           this.claudePromptEvidence = {
             text: m[0],
-            status: isApproval ? 'awaiting_input' : 'waiting',
-            message: isApproval ? 'Approval requested' : 'Ready for input',
+            status: 'waiting',
+            message: 'Ready for input',
           };
         }
       }
--- a/src/main/pty/AgentDetector.ts
+++ b/src/main/pty/AgentDetector.ts
@@ -61,143 +61,39 @@ type CriticalEventCallback = (event: CriticalEvent) => void;
 // SLUG-form agent identifier. Lowercase, no whitespace. Used as the
 // canonical key shared with hook-based signals (integrations/<agent>/).
 // HookSignalRouter dedup matches AgentDetector emissions against bridge
-// signals on this slug.
-//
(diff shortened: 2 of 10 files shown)


openwong2kim:

This PR addresses #850 with two complementary fixes: a metadata guard in the Fleet View selector and a compound detection gate for Claude Code. The selector change threads `surfaceAgent` identity through all three consumer call sites (FleetView, DeckFleet, StatusBar), so a non-agent active pane can no longer inherit a workspace's agent name and status — the direct cause of the false "Claude Code · Needs you" card on a btop pane. The detector change adds a banner-and-prompt compound gate for Claude, mirroring the Kiro pattern, which prevents a process monitor that merely displays "claude" in its process list from opening the detection gate. The new test coverage — compound-gate ordering, per-detector isolation, and the non-agent-inheritance case — directly pins the regression being fixed.

One observation worth noting: the diff also reverts the generic compound-gate infrastructure (`EvidenceClause`, the `extends`/`omit` inheritance system, `matchEvidence`) in favor of per-slug special-casing, and OpenClaude's approval patterns are now duplicated rather than inherited from Claude. The inheritance system existed specifically to keep those patterns in sync — its removal means the OpenClaude approval regexes will need to be maintained by hand going forward, and the dedicated inheritance test block has been removed. If the intent is to move to a simpler special-cased model, that tradeoff is worth confirming explicitly so future pattern changes do not silently desync between Claude and its fork.

openwong2kim:
Dogfooded this on a packaged build and pushed one more commit.

**The A/B, against `main`.** First pane made to look like an agent (banner + idle footer), then Ctrl+D so a fresh empty pane is the active one, then Fleet View read out of the DOM.

Before, the badge landed on the wrong pane entirely:

```
row 1: "Claude Code · Waiting"   ← the fresh empty pane
row 2: PowerShell · Idle          ← the pane that actually printed the banner
```

After: neither row names Claude, and a pane that really is an agent still gets its badge — worth checking explicitly, since a compound gate that never opens would "fix" the false positive by killing detection outright. Both hold.

**One thing the branch dropped: #837's running veto.**

`agentStatus` is one slot per workspace. The identity gate replaced `agentStatus !== 'running'` with a name match, on the reasoning that a matching `agentName` proves the status belongs to this pane. It does not — the shape #837 was built to stop is two panes running the *same* agent, and an orchestrator plus its worker are both "Claude Code", which is the normal arrangement here rather than an exotic one. The worker's `running` lands in the shared slot and the name match paints it onto the active orchestrator pane: a pane that ran nothing reads as an outstanding worker, which is what `deck_complete_work` counts.

The existing #837 cases kept passing only because their fixtures carry no `surfaceAgent` entries, so the new gate short-circuited before the veto could matter. `946e326e` restores the veto and adds the case that actually exercises it — same agent name on both panes, `running` in the workspace slot, active pane confirmed as an agent. Verified it fails without the veto and passes with it, so it is not a vacuous test.

`error` is still inherited: it is not an ATTENTION status, so the workspace slot is its only carrier for the active pane.

For the record, I had started down a different path — reverting the AgentDetector half, because the earlier revision on this branch had replaced the file with a shorter older one and lost the compound-gate machinery, `extends`/`omit`, and the `agentIdentity` split. `400f023b` fixes that properly and keeps both halves of #850, so I dropped my revert and kept only the fleet change above.

openwong2kim:
All three addressed in 6d0bfd84. Each one reproduced first — they were all real.

**The lifecycle test could not fail for its stated purpose.** Confirmed exactly as described: opening the gate emits a `running` notification on its own, so `toHaveBeenCalled()` passed even with `awaiting_input` vetoed. I proved it by removing the `status !== 'awaiting_input'` exemption in PTYBridge — the old assertion still passed, the new one fails. It now asserts `category: 'approval'`, which is what PTYBridge tags an awaiting_input notification with. Duplicate emit line removed.

**Claude gate probe now bounded like Kiro's.** Measured on btop-shaped frames, since the concern is a full-screen TUI on a pane where the gate never opens:

| frame | current | bounded+hinted | ratio |
|---|---|---|---|
| 10 KB | 0.014 ms | 0.001 ms | 10x |
| 24 KB | 0.030 ms | 0.001 ms | 31x |
| 40 KB | 0.057 ms | 0.001 ms | 57x |

One correction to the finding: the backtracking is linear, not catastrophic — V8 handles the `.*` plus lookbehind better than the description implies. The problem is not a single slow call, it is that a non-Claude pane never opens the gate, so this runs on every chunk for the pane's whole lifetime. Which is precisely the pane #850 is about, so the fix stands regardless of the constant.

**Replay/pattern-pass mismatch is real and reproducible.** `CLAUDE_PROMPT_RE` is an alternation, so it matches at the earliest POSITION; the pattern pass tries the waiting patterns in ARRAY order. They disagree whenever a footer carries both fragments with `shift+tab to cycle` first:

    "… shift+tab to cycle | bypass permissions on"
      gate -> "shift+tab to cycle"     pass -> "bypass permissions on"

Since the dedup key is the match text, that emits `waiting` twice. The replay now resolves through the same list the pass uses, derived from `AGENT_PATTERNS` rather than a second hand-written copy — a duplicated list is how these drift apart again. This is the requirement the Kiro replay already documents; Claude just wasn't following it. Exact-count tests added for the both-fragments footer and for `bypass permissions on` alone.

tsc clean, 2087 tests passing across the pty and renderer suites.

PR #861, opened 2026-08-11
Drop powershell.exe from Windows token-file ACL hardening

openwong2kim:
On a machine running PowerShell in Constrained Language Mode — the
AppLocker/WDAC default on managed corporate fleets — wmux has been quietly
failing to lock down its own token files.

The owner-only DACL rebuild shelled out to
`powershell.exe -NoProfile -NonInteractive -ExecutionPolicy Bypass -EncodedCommand`
running a .NET `FileInfo.SetAccessControl`. Under CLM that script dies on its
*first* .NET call — `[Console]::In.ReadToEnd()` — with
`MethodInvocationNotSupportedInConstrainedLanguage`. I measured 22 out of 22
failures on a real corporate box. Every one of those falls back to the plain
`icacls` strip, which only removes four well-known broad SIDs, so an explicit
ACE for any other principal survives. That installed base has been running
with weaker token ACLs than `SECURITY.md` §1.2 promises, and the only trace is
a `console.warn` that never even reaches the log file in the main process.

Same argument shape is what Norton Behavioral Protection flags as
`IDP.HELU.PSE85` (GHSA-8fj2-47w9-jxq3, reported by @TarJae). Two unrelated
mechanisms — a policy and a heuristic — converging on the same silent
downgrade is the actual argument for getting off PowerShell here, rather than
reshaping the command line.

There's a performance side to it too: the first `powershell.exe` spawn inside
an Electron process measured 1.8–2.3 s, and `peer-hmac-key` was the one call
site #213 missed, still synchronous on the daemon's launcher-blocked boot path.
That was 75–78 % of daemon cold start.

## What replaces it

Hardening now writes the content through a fresh inode instead of editing the
live file's ACL. `writeFileSync` preserves an existing file's DACL — that is
exactly why an in-place write can never drop a pre-existing explicit ACE — but
a file that did not exist before we created it carries only inherited ACEs, so
`/inheritance:r` on it is sufficient and the whole #124 case becomes
unreachable by construction.

The staging area is a per-operation directory, hardened with an inheritable
`(OI)(CI)` owner-only grant *while it is still empty*. The payload file is
therefore born owner-only; it never exists for an instant on an inode another
process could have opened a surviving handle to (Windows ACL changes do not
revoke open handles). Then an atomic rename over the target.

Two alternatives were tried against real on-disk ACLs and rejected:
`icacls /restore` with a hand-built DACL-only SDDL fails with *"Not all
privileges or groups referenced are assigned to the caller"*, processing 0
files — the same privilege wall that rules out `Set-Acl`; a two-pass
`icacls /save` + targeted `/remove:g` works but needs SDDL parsing for deny
ACEs, conditional ACEs and abbreviated principal aliases.

## The part that took three review rounds

Introducing a second rename introduced failure modes the old in-place edit
never had, and each round of review found the next one:

- A stale hardening snapshot could be renamed over a concurrent write —
  `PipeServer.rotateToken()` would leave the *old* token on disk while the
  server used the new one. The commit is now a synchronous compare+rename
  block; if the content changed, the harden stands down.
- `reHardenTokenFileAcl` returned a boolean that conflated "could not harden"
  with "nothing changed and nothing got weaker". `PeerStore` wires that signal
  straight to destructive branches: `persist()` unlinks the peer store,
  `loadOrCreateMachineKey` regenerates the machine HMAC key — which
  invalidates `lanlink-peers.json`'s MAC and drops every pairing on the
  `load()` right after it. It now returns a typed `HardenOutcome`, and the
  safe value is *verified* (`icacls /save`, parsed as SDDL), never assumed —
  on an upgrade boot the pre-existing ACL is precisely the weak one.
- The write path used to delete the previous token on failure. It no longer
  touches a token it never wrote, unless that token's own DACL fails to verify.

Then live dogfooding found the one no reviewer could: the collision does not
surface at the rename. A file opened with `FILE_SHARE_NONE` — what several AV
and backup products do while scanning — cannot be *read* at all, so the
snapshot read at the top threw `EBUSY` and the harden reported `failed`,
re-arming the exact self-DoS the collision handling exists to prevent. Reading
a security descriptor is not subject to share modes and an ACL edit needs only
`WRITE_DAC`, so the fix is to skip staging entirely and repair the DACL in
place, then read it back. Mocked unit tests could not have shown this; the
real daemon did.

## Verification

Against a file genuinely held with `FileShare.None`:

- broad file (7 ACEs incl. `Everyone`) → `hardened`, ends owner-only (1 ACE,
  protected), content byte-identical
- custom explicit ACE the strip cannot remove → still `failed`, fail-closed
  stays armed
- unheld control → normal swap path

Daemon dogfood on an isolated instance suffix, 3 cold starts: **0**
`powershell.exe` descendants (sampled every 60 ms through boot), hardens
96–129 ms (was 2155–2309 ms), daemon boot 278–497 ms (was 2748–3053 ms), both
secret files owner-only on disk, no staging leftovers.

`scripts/issue-124-acl-dynamic.mjs` 29/29 against real DACLs, including a new
case (c2) that first *proves* the plain icacls strip leaves a custom explicit
ACE behind and then that the rewrite removes it, and a case (e) pinning the
SDDL parser against live `icacls /save` output so a format drift cannot
silently turn transient collisions destructive.

54 unit tests, affected suites 1482 passing, `tsc` clean.

## Notes for review

- `SECURITY.md` §1.2 and `PROTOCOL.md` §5 are updated in the same PR — they
  described the .NET primitive as the contract, and `security.ts` names them
  as the thing to keep in sync.
- Contract change worth flagging: the re-harden now rewrites through a fresh
  inode, so the file's inode identity changes although its bytes do not. No
  current consumer cares (token readers are open-read-close; the only
  `fs.watch` users target `.git/HEAD` and transcript dirs), but a future
  watcher on a token file would see a rename.
- Follow-up, deliberately not in scope: at ~100 ms the sync/async split
  (`scheduleTokenFileReHarden` and its async twins) may no longer earn its
  complexity. Noted in a TODO comment rather than ripped out here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows token-file protection without requiring PowerShell.
  - Prevented rename collisions and failed security updates from deleting phone pairings or machine keys.
  - Added safer recovery for locked files, concurrent writes, and custom permissions.
  - Reduced Windows daemon startup time.
  - Avoided triggering Norton startup detection.

- **Documentation**
  - Updated security and protocol documentation to reflect stronger Windows protection and fail-closed handling.


openwong2kim:

This PR removes `powershell.exe` from the Windows token-file ACL hardening path, replacing the .NET `FileInfo.SetAccessControl` shell-out with a fresh-inode rewrite staged inside a per-operation directory that is hardened while still empty. The motivation is concrete and measured: on any machine running PowerShell in Constrained Language Mode (the AppLocker/WDAC default on managed corporate fleets), the old call failed 22/22 times on a real box, silently degrading to a plain `icacls` strip that leaves custom explicit ACEs behind — so those machines have been running with weaker token ACLs than `SECURITY.md` §1.2 documents. It also clears the Norton `IDP.HELU.PSE85` detection (GHSA-8fj2-47w9-jxq3) and removes 1.8–2.3 s (75–78 %) of daemon cold start.

The change touches `src/shared/security.ts` (the core rewrite), `src/daemon/lanlink/peers.ts` (where the new `HardenOutcome` type replaces a boolean that previously conflated "could not harden" with "nothing changed" — a conflation that could trigger destructive fail-closed branches like peer-store unlink and machine-key regeneration), and updates the dynamic ACL harness with a new case proving the custom-ACE gap. The PR went through three three-reviewer passes (Claude/Codex/GLM) closing five P1s around swap-window races, staging pre-create/hardlink defense, and verified `unchanged` via SDDL read-back, plus a live dogfood fix for `FILE_SHARE_NONE` collisions that mocks could not have caught.

This is a genuine security hardening fix on a documented guarantee — the kind of change that should land ahead of feature work. Assigning P1.

openwong2kim:
@codex review

openwong2kim:
Second review pass — all three addressed in 85c50281, each confirmed against the code first.

**[P1] rename-onto-existing.** Confirmed: `daemon/util/atomicWrite/rotation.ts` (lines 25-28, 84-89) already documents and works around this exact quirk, and the token write path had no equivalent. Not fixed by copying that pattern though — unlinking a *token* first opens a window where it does not exist, and a crash there loses it. `swapIntoPlaceAllowingUnlink` tries the atomic rename first and only falls back to unlink-then-rename on the specific rejection codes. The re-harden path deliberately does **not** get the fallback: a failed swap there already falls through to the in-place repair, which reaches the same owner-only DACL with no rename and no window.

**[Major] null SID blocking verification.** Confirmed and fixed. `icacls /save` renders principals as raw SIDs, so on the ASCII-username fallback path nothing could be compared and every collision deleted a previous token that may have been correctly hardened. Verification now resolves a SID for the comparison only — the ACL that gets *applied* still refuses a non-ASCII name, so the #90 codepage rule is untouched, and a failed resolution still means unverifiable and stays fail-closed.

**[P2] pr140 harness.** Confirmed — I updated the issue-124 harness for the new return type and missed this one. Fixed. Running it then surfaced a pre-existing defect in that harness: it never stripped `PSModulePath` for its PowerShell 5.1 children, so from a pwsh 7 parent every `Get-Acl`/`Set-Acl` threw and it died on its own instrumentation. Both harnesses now carry that fix. It passes 16/16 against the real `%USERPROFILE%\.wmux` descriptor — which is the coverage it exists for, since a fresh mkdtemp directory has a different ACL constitution and that difference is what caused the 2026-05-16 revert recorded in SECURITY.md.

openwong2kim:
Right on both counts, fixed in 38a49ee6.

The readability probe was not a proof and I should not have shipped it as one. `rename` needs DELETE access on the source to remove its directory entry, so a scanner holding the staging file with `FILE_SHARE_READ` passes the probe while the rename still fails — at which point the destination is already unlinked and the live token is gone. It narrowed the window without closing it, which is the worst shape for this kind of bug: rarer, and therefore harder to ever catch again.

There is no reliable way to disambiguate `EPERM`/`EACCES` on Windows, so those codes no longer get the fallback at all. Only `EEXIST` does, which is specific to the destination. Everything else falls through to the retry loop, and on the re-harden path to the in-place repair — neither of which can destroy anything.

The trade that buys: a volume reporting a rename-onto-existing refusal as `EPERM` rather than `EEXIST` now fails the write loudly instead of recovering. A failed token rotation is recoverable; a deleted live token is not.

Tests are as requested — `EPERM` and `EACCES` each assert the live token survives and `unlinkSync(tokenPath)` is never called, with `EEXIST` still asserting the unlink-then-rename recovery completes. The probe helper and the fs mocks it needed are removed rather than left dangling.

The other open thread (pr140 harness asserting `=== true`) was fixed earlier in 85c50281, along with a pre-existing defect it exposed: that harness never stripped `PSModulePath` for its PowerShell 5.1 children, so from a pwsh 7 parent it died on its own instrumentation. It now passes 16/16 against the real `%USERPROFILE%\.wmux` descriptor.

Issue #858, opened 2026-08-10
Control-pipe event subscription: closing the gaps opt-in left open (#659 follow-up)


openwong2kim:
Deferred from #856, which made pushed control-pipe events opt-in. Three findings from that PR's review were real but out of its scope. None of them is a regression from #856 — each is either pre-existing or newly reachable in a way worth closing deliberately.

## 1. Events generated between accept and subscribe are dropped

The daemon processes lines in order, so everything it emits after reading a client's `daemon.events.subscribe` is delivered. Events emitted in the window before that — one round-trip on a local pipe — are not, and there is no replay or cursor, so a one-shot state transition (`session.died`, `session.restarted`, `channel.message`) that lands in the window is gone for good.

Before #856 this window was zero: a socket was eligible for broadcasts the instant it was accepted. It is small, and the app has usually not finished its handler swap that early, but "small" is not "none" and the recovery path is missing rather than short.

Two shapes worth weighing:
- Resync after subscribe — the client re-reads state (`pane.list` snapshot) once its subscription is confirmed, so a missed transition is corrected rather than replayed.
- Seq-based replay — the subscribe reply carries a cursor and the daemon replays a bounded tail, which is stronger but needs the daemon to retain events it currently does not.

The daemon already has `asOfSeq` / `bootId` in the `pane.list` envelope, so the first shape may cost very little.

## 2. `broadcast` has no backpressure guard, but `sendTo` does

`sendTo` refuses to write once a client's `writableLength` passes `SUBSCRIBER_BACKPRESSURE_BYTES`, because Node's writable buffer has no ceiling and a subscriber that stops draining would otherwise grow the daemon's heap one payload at a time. `broadcast` has no such check — it writes to every subscriber unconditionally.

This predates #856, but the subscriber set is now the deliberate delivery path rather than an accident of being connected, which makes the missing bound worth fixing on purpose. The fix is the same check `sendTo` already makes, plus dropping the subscription and logging when a client exceeds it.

## 3. `daemon.events.subscribe` has no first-party gate

The transcript RPCs are gated with `firstPartyOnly` because a pane's conversation should not be readable by an MCP tool wired into another agent. The event stream carries workspace-scoped material too — `channel.message`, `a2a.task` — and any token holder can now turn it on through a documented API.

This is not a privilege escalation: before #856 every connected client received these events with no API call at all, so the gate would tighten today's behavior rather than restore yesterday's. It is worth deciding explicitly now that subscribing is a named operation.

[...]

PR #862, opened 2026-08-11
fix(daemon): bound broadcast subscriber backpressure

snowyukitty:
## Summary

- apply the existing per-subscriber backpressure cap to broadcast event delivery
- drop only a stalled client's event subscription while keeping its RPC socket connected
- log the dropped subscription and cover stalled/healthy subscriber isolation

This addresses gap 2 from #858 only. Event resync/replay and the first-party subscription gate remain intentionally out of scope.

## Testing

- `node_modules/.bin/tsc --noEmit`
- `node_modules/.bin/eslint src/daemon/__tests__/DaemonPipeServer.subscriberGuards.test.ts`
- `npm exec vitest run -- src/daemon/__tests__/DaemonPipeServer.subscriberGuards.test.ts src/daemon/__tests__/DaemonPipeServer.eventOptIn.test.ts` (13 passed)
- `npm run test:parallel` (723 files passed, 10,782 tests passed, 24 skipped)
- `npm run test:runtime` (10 files / 149 tests passed; 2 pre-existing `shellIntegration.runtime.test.ts` cases timed out because this host's node-pty helper reports `AttachConsole failed`)

Refs #858

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved event delivery under backpressure by removing only stalled event subscriptions when their output buffers exceed the safety limit.
  - RPC connections remain open, and healthy subscribers continue receiving events.
  - Added warning logs when a stalled subscription is dropped.

- **Documentation**
  - Clarified how broadcast behavior handles stalled subscribers.

### DIFF
--- a/changelog.d/862.md
+++ b/changelog.d/862.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **A stalled event subscriber can no longer grow the daemon's write buffer
+  without bound.** Broadcast delivery now drops only that client's event
+  subscription after its buffered output crosses the existing safety cap. The
+  RPC connection stays open, and healthy subscribers continue receiving events.
+  (#862)
--- a/src/daemon/DaemonPipeServer.ts
+++ b/src/daemon/DaemonPipeServer.ts
@@ -14,8 +14,8 @@ const MAX_LINE_BUFFER = 1024 * 1024; // 1 MB — prevent OOM from malicious clie
  * and the writable buffer has no ceiling. A subscriber that stops reading (a
  * hung renderer, a paused process, a phone on a dead link) would therefore grow
  * the daemon's heap by one full payload per append, indefinitely. Above this cap
- * per-subscriber events are REFUSED — `sendTo` returns false, and the caller's
- * contract is to drop that client's subscription rather than keep buffering.
+ * per-subscriber events are REFUSED — `sendTo` returns false, and `broadcast`
+ * drops the stalled event subscription rather than keep buffering.
  */
 export const SUBSCRIBER_BACKPRESSURE_BYTES = 4 * 1024 * 1024;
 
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR closes gap 2 from #858 by extending the existing per-subscriber backpressure cap (`SUBSCRIBER_BACKPRESSURE_BYTES`) to the `broadcast` event-delivery path in `DaemonPipeServer`. Until now only `sendTo` enforced that bound, so a subscriber that stopped draining (a hung renderer, a paused process, a phone on a dead link) could grow the daemon's write buffer — and therefore the always-on process's heap — without limit, one broadcast payload at a time. The fix is small and surgical: when a subscriber's `writableLength` crosses the cap, its event subscription is dropped and logged, while the underlying RPC socket stays connected and healthy subscribers keep receiving.

The change touches only `DaemonPipeServer.ts` (+11/−2), a changelog fragment, and a new test case that verifies stalled/healthy subscriber isolation — the stalled client is unsubscribed, the draining client still receives both the pre- and post-drop broadcasts, the connection count is unchanged, and the warning fires exactly once. It is deliberately scoped: event resync/replay (gap 1) and the first-party subscription gate (gap 3) from #858 remain out of scope, which keeps this PR easy to review and back out if needed.

This is defensive hardening of a pre-existing condition rather than a regression, and the parent tracking issue is P3, so I am labelling it **P3** — worth merging on its own merits but not blocking. The test coverage and the decision to keep the RPC connection alive (only the event feed is severed) both look right; no concerns from triage.

snowyukitty:
Addressed the review nitpick in `5155e2b8`: the backpressure test now directly asserts that the stalled socket is not destroyed, while retaining the connection-count assertion.

Targeted ESLint and both related test files pass (13/13). Fresh CI for this head is also green on validate, perf, and Windows/macOS/Linux baselines.

PR #860, opened 2026-08-11
feat(settings): add one-shot chrome presets

snowyukitty:
## Summary

Adds two one-shot Settings actions for interface visibility: **Minimal** hides five existing chrome surfaces together, and **Standard** restores the shipped defaults without storing a mode or derived Custom state.

## Changes

- centralize the exact Minimal and Standard preference maps, then apply all five values in one Zustand transaction
- preserve the Channels route invariant, expose neutral Settings actions, and announce completion with an accessible toast
- localize the controls across all 23 shipped locales and add store, UI, and locale-contract coverage

### Added

- **One-shot interface visibility presets.** Apply Minimal to hide optional chrome at once, or restore the shipped Standard visibility without entering a persistent mode.

## Stability tier impact

- [ ] No external surface touched (internal refactor / docs / tests / CI).
- [ ] Additive change to a `stable` surface (new optional param, new field, new event type).
- [ ] Breaking change to a `stable` surface (requires major bump — block until release planning is in scope).
- [ ] Change to an `experimental` surface (note in release notes).
- [x] Change to an `internal` surface (no contract impact).

## Substrate contract impact (Substrate 3.0)

n/a

## Test plan

- `npx vitest run src/renderer/stores/slices/__tests__/chromePresetSlice.test.ts src/renderer/components/Settings/__tests__/SettingsPanel.chromePreset.test.tsx src/renderer/i18n/__tests__/chromePresetLocale.test.ts` — 30 passed
- `npx tsc --noEmit` — passed
- changed-file ESLint — 0 errors; two pre-existing `SettingsPanel.tsx` warnings remain
- `npm run package` — passed, including production renderer and Electron packaging
- `npm test` — full parallel phase reached 10,800 passed / 9 unrelated Windows timeout failures; all 9 failing files passed when rerun with one worker (129 passed, 1 skipped). Runtime reached 149 passed / 1 skipped / 2 existing PowerShell shell-integration failures caused by `node-pty` `AttachConsole failed` on this host.
- automated visual QA was unavailable because no in-app or extension browser instance was connected; button interaction and accessibility semantics are covered in jsdom

## Related issues / PRs

Closes #859
Follow-up to #851

## Reviewer checklist

- [x] CHANGELOG entry above is filled in or explicitly marked n/a.
- [x] Stability tier impact is correctly classified.
- [x] Substrate contract docs (PROTOCOL.md, inventory.md, stability.md) updated if the surface changed.
- [x] Tests cover the new behavior and the regression case (if a bug fix).
[...]

PR #864, opened 2026-08-11
test(cli-shim): budget the live-registry spawns by count, and let a timeout say so

openwong2kim:
## Summary

The `PATH edit (live registry, sandbox key)` suite failed twice on #860, whose diff touched only Settings, locales and the store — nothing near the CLI shim, the registry, or PowerShell. Both failures were spawn timeouts on a loaded Windows runner, and neither reported itself as one. Re-running the same commit unchanged went green, which is what a flake looks like, but it took a log dig to establish that.

Two things made a slow runner look like a product regression.

## The per-test budget was not derived from what a case does

The comment on `SPAWN_TIMEOUT_MS` promised that a spawn outliving its own ceiling surfaces as a *named* spawn failure rather than a bare vitest timeout:

> Must stay below the suite's per-test budget above: a spawn that outlives it can only ever surface as a bare vitest timeout, which says nothing about which of the three or four spawns in a case actually hung.

The arithmetic only delivered that when exactly one spawn was slow. The most spawn-heavy case (`add is idempotent`) runs four — `seed` + `add` + `add` + `readRaw` — at 10s each, against a 15s case budget. Under general runner slowness, which is the failure mode that actually occurs, vitest killed the case first and the promise in the comment silently did not hold.

So the case budget is now derived from the spawn budget and the spawn count rather than hand-tuned alongside it:

```ts
const SPAWN_TIMEOUT_MS = 20_000;
const MAX_SPAWNS_PER_CASE = 4;   // seed + add + add + readRaw
const CASE_TIMEOUT_MS = SPAWN_TIMEOUT_MS * MAX_SPAWNS_PER_CASE + 10_000;
```

These are ceilings, not sleeps — the suite still finishes in ~24s locally. The spawning `afterEach` hook gets a budget too, since vitest's default 10s `hookTimeout` would otherwise fire before the spawn's own and trade a named failure for a bare one in teardown.

## The assertions threw the diagnosis away

`ps()` already builds `killed by ${signal} after ${ms} ms` on a timeout. Then `expect(r.code).toBe(0)` renders `expected -1 to be +0` and drops `r.err` on the floor. That message is the whole difference between "a spawn hung" and "the shim exited non-zero", and losing it is why #860's failure read as a real regression:

```
AssertionError: expected -1 to be +0 // Object.is equality
 ❯ src/main/__tests__/cliShim.test.ts:297:22
```

`-1` is not an exit code the shim can produce — it is `err.status ?? -1`, i.e. status was `null` because the child was signalled. The exit-code assertions now go through `expectExit`, which carries `r.err` into the failure message.

This is the same class of conflation as #690 (a broken fixture impersonating the shim), one layer further out: there, a fixture's exit code was swallowed; here, a spawn's cause of death was.

## Test plan

Run on Windows, against the real registry and real `powershell.exe`:

- `npx vitest run src/main/__tests__/cliShim.test.ts` → **26 passed, 8 skipped**, 24.2s
- With `run()` forced to a 1 ms timeout, the failure now names itself:
  ```
  AssertionError: killed by SIGTERM after 1 ms: expected -1 to be +0
  ```
  (was: `AssertionError: expected -1 to be +0`)
- With `SPAWN_TIMEOUT_MS` forced to 1 ms, the fixture path still reports first and by name — `fixture "seed sandbox key" failed with exit -1 / killed by SIGTERM after 1 ms` — so the #690 guarantee is intact.
- `npx tsc --noEmit` clean.

No product code changed; the assertions are the same assertions.

## Stability tier impact

n/a

## CHANGELOG entry

None — tests only, no user-facing change, so no `changelog.d/` fragment.

## Summary by CodeRabbit

* **Tests**
  * Improved reliability of Windows command-line PATH configuration checks.
  * Added clearer diagnostics when command execution does not complete as expected.
  * Increased timing tolerance for slower environments, reducing intermittent test failures.

### DIFF
--- a/src/main/__tests__/cliShim.test.ts
+++ b/src/main/__tests__/cliShim.test.ts
@@ -169,7 +169,23 @@ import { execFileSync } from 'child_process';
 // vitest's 5s default (observed 5.6s in CI while its siblings ran 1–2s), so it
 // flakes intermittently. Raise the per-test budget for the whole suite — these
 // are the slowest tests here anyway, and a genuinely hung reg.exe still fails.
-describe.skipIf(process.platform !== 'win32')('PATH edit (live registry, sandbox key)', { timeout: 15_000 }, () => {
+
+// A spawn budget, not a performance target: warm spawns measure ~1s, and this
+// only has to be wide enough that anything hitting it is genuinely stuck. #860
+// saw one exceed 10s on a loaded runner with nothing wrong with the shim.
+const SPAWN_TIMEOUT_MS = 20_000;
+// The most spawn-heavy case ("add is idempotent") runs seed + add + add +
+// readRaw. The per-test budget has to clear all of them at their own ceiling,
+// or vitest kills the case first and the failure arrives as a bare "timed out
+// in Nms" — naming neither the spawn nor the reason, which is precisely what
+// SPAWN_TIMEOUT_MS exists to report. Derived rather than hand-tuned because
+// the previous pairing (10s spawn, 15s case) claimed that guarantee in a
+// comment while the arithmetic only delivered it when exactly ONE spawn was
+// slow. General runner slowness — the actual failure mode — broke it.
+const MAX_SPAWNS_PER_CASE = 4;
+const CASE_TIMEOUT_MS = SPAWN_TIMEOUT_MS * MAX_SPAWNS_PER_CASE + 10_000;
+
+describe.skipIf(process.platform !== 'win32')('PATH edit (live registry, sandbox key)', { timeout: CASE_TIMEOUT_MS }, () => {
   const SUB = 'Software\\wmux-cliShim-test';
   const BAK = 'Software\\wmux-cliShim-test-bak';
   const BIN = 'C:\\Users\\u\\AppData\\Local\\wmux\\bin';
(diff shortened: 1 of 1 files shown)


openwong2kim:

This is a tests-only change to the Windows `cliShim` PATH-edit suite (`src/main/__tests__/cliShim.test.ts`) that addresses a CI flake observed on #860, where the most spawn-heavy case was killed by vitest's case timeout before any individual spawn budget could surface a named failure. The per-test budget is now derived from `SPAWN_TIMEOUT_MS * MAX_SPAWNS_PER_CASE + headroom` instead of hand-tuned alongside it, and the `afterEach` teardown hook gets its own spawn-aware budget so a hung cleanup can no longer arrive as a bare default timeout.

The second change introduces an `expectExit` helper that threads `r.err` (which carries `killed by SIGTERM after N ms` from `ps()`) into the assertion message. Previously `expect(r.code).toBe(0)` rendered `expected -1 to be +0` and discarded the one fact distinguishing a hung spawn from a genuine non-zero exit — which is exactly why #860's CI failure read as a product regression rather than a loaded runner.

No product code is touched, the assertions themselves are unchanged, and the local run is reported at 24s with 26 passed / 8 skipped. This is solid test-hygiene work that makes future CI flakes self-diagnosing; I'm labelling it P3 since it is internal-only with no user-facing or contract impact.

PR #865, opened 2026-08-11
fix(installer): repair dead AUMID shortcuts and stop depending on iconUrl for shortcut icons

### LINKED ISSUE #863: [Bug]: wmux 自动更新后任务栏及快捷方式图标丢失
### What happened?

wmux 通过 Squirrel 自动更新后，Windows 任务栏或开始菜单中的 wmux 图标会变为空白。

### Steps to reproduce

1. 安装并运行 wmux。                                                                                            
  2. 将 wmux 固定到任务栏。                                                                                       
  3. 等待或手动触发自动更新。                                                                                     
  4. 更新完成后重新启动 wmux。

### Expected behavior

 自动更新后，任务栏、桌面和开始菜单中的 wmux 图标保持正常，已有固定项仍然可以正确关联应用。 

### wmux version

3.40.2

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

Window11

_No response_

```shell

```

### Confirmation

openwong2kim:
## Summary

Windows resolves a taskbar button's icon through the shortcut that carries the window's AppUserModelID (`com.squirrel.wmux.wmux`) — not through the window or exe icon. One dead shortcut carrying that AUMID is enough to blank the running app's taskbar icon while the exe, the window icon, and every other shortcut stay healthy.

Root-caused with an A/B/A experiment on a live install: a Start Menu link left behind by a <=3.3.x install (the old top-level `Programs\wmux.lnk` layout) still pointed at `app-3.3.0\`, a directory deleted many updates ago. Removing it and restarting fixed the icon; restoring it broke the icon again. That legacy link had survived three installs untouched — `--createShortcut` only writes the Desktop and `Programs\<author>\` links, and Squirrel's own `fixPinnedExecutables` did not repair it either.

The same mechanism covers the reported repro (pin while running → update → blank pin): pinning a running window records the live `app-X.Y.Z` exe path into the pin's `.lnk`, and the next update deletes that directory.

## Fix

The `--squirrel-install` / `--squirrel-updated` hooks now run a shortcut-hygiene pass (best-effort, never blocks the install):

- **Repair, don't delete.** Links whose target lies under our install root but is not the version-stable root stub (`<root>\wmux.exe`) are retargeted to the stub with the icon pinned to `<root>\app.ico`. Pins cannot be re-created programmatically (Windows blocks pin automation), so a repaired pin keeps working where a deleted one is unrecoverable. Scope: the legacy top-level Start Menu link + the `User Pinned` folders; Desktop/publisher links stay owned by `Update.exe --createShortcut`.
- **One deletion:** the legacy top-level link, when dead and duplicated by the publisher-folder link — retargeting it would leave two identical Start Menu entries forever.
- **Kill the icon's network dependency.** Squirrel materializes `<root>\app.ico` by downloading the nuspec `iconUrl` (raw.githubusercontent.com) at install time — unreachable from some networks, including the reporter's region, leaving no icon source at all. The byte-identical `.ico` already ships in the package (`resources\icon.ico`), so the hooks stage it to `<root>\app.ico` and pass it to `--createShortcut --icon`.
- Foreign shortcuts (target outside our root) are never touched; a burned root (missing stub) disables the pass entirely.

WScript.Shell `Save()` was measured to preserve the link's property store, including the AUMID — the repair does not break pin grouping.

## Testing

- `vitest run src/main/__tests__/shortcutHygiene.test.ts` — 15 passed (script construction, PS-literal injection guard, output parsing)
- `vitest run --config vitest.runtime.config.ts src/main/__tests__/shortcutHygiene.runtime.test.ts` — 8 passed, end-to-end against **real `.lnk` files** via real PowerShell/COM: dead-pin retarget, live-but-versioned retarget, legacy dedupe/retarget both ways, foreign-shortcut isolation, burned-root bail-out, and AUMID preservation across `Save()`
- `vitest run src/main/__tests__/squirrel*.test.ts` — 37 passed (hook wiring semantics untouched)
- `tsc --noEmit` clean; eslint adds no new findings on touched files

Not covered here (tracked separately): the in-app updater launches `Setup.exe` before `app.quit()` completes, which can race Squirrel's install-root wipe against live file locks — reproduced and logged during this investigation, filing as its own issue.

Closes #863

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale Windows shortcuts that could cause blank taskbar icons after updates.
  * Existing shortcuts are now repaired or removed when necessary while preserving healthy shortcuts.
  * Installer updates now use the packaged application icon instead of downloading one externally.
* **Chores**
  * Added Windows-specific validation for shortcut repair, icon handling, and installation scenarios.

### DIFF
--- a/changelog.d/865.md
+++ b/changelog.d/865.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- **A dead shortcut can no longer blank the taskbar icon after an update.**
+  Windows resolves the taskbar button's icon through the shortcut carrying the
+  app's AppUserModelID, so one stale link — a pre-3.4 top-level Start Menu
+  entry, or a taskbar pin recorded against a deleted `app-X.Y.Z` directory —
+  blanked the icon of a perfectly healthy install. The installer hooks now
+  repair such links (retargeting them to the version-stable root launcher)
+  instead of leaving them behind. (#863, #865)
+- **Shortcut icons no longer depend on a download at install time.** The
+  installer previously fetched the shortcut icon from raw.githubusercontent.com,
+  which fails on networks where that host is unreachable and left shortcuts
+  with no icon at all. The icon that already ships inside the package is used
+  instead. (#863, #865)
--- a/src/main/__tests__/shortcutHygiene.runtime.test.ts
+++ b/src/main/__tests__/shortcutHygiene.runtime.test.ts
@@ -58,7 +58,10 @@ describe.skipIf(!onWindows)('shortcutHygiene end-to-end (real .lnk, real PowerSh
   let loc: RepairLocations;
 
   beforeEach(() => {
-    sandbox = fs.mkdtempSync(path.join(os.tmpdir(), 'wmux-hygiene-'));
+    // realpathSync.native expands 8.3 short components: CI runners hand back
+    // `C:\Users\RUNNER~1\...` from os.tmpdir() while the shell/COM layer
+    // resolves the long form, so raw mkdtemp paths fail every string compare.
+    sandbox = fs.realpathSync.native(fs.mkdtempSync(path.join(os.tmpdir(), 'wmux-hygiene-')));
     // Fake install root: <root>\wmux.exe stub + <root>\app-1.0.0\wmux.exe (the
     // "current version" the hook process runs from) + resources\icon.ico.
     root = path.join(sandbox, 'wmux');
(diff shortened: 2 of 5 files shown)


openwong2kim:
@codex review

openwong2kim:

This PR closes #863 by adding a shortcut-hygiene pass to the Squirrel `--squirrel-install` / `--squirrel-updated` hooks, fixing the blank taskbar icon that appears after auto-update on Windows. The root cause — a dead `.lnk` carrying the app's AUMID pointing at a version directory the next update deletes — is addressed by retargeting stale pins and legacy Start Menu links to the version-stable root stub (`<root>\wmux.exe`) rather than deleting them (pins cannot be re-created programmatically), and it additionally kills the installer's dependency on downloading `app.ico` from `iconUrl` by staging the byte-identical icon that already ships in `resources\icon.ico`. The change is well-scoped (5 files, ~700 additions) and the testing is thorough: 15 unit tests plus 8 end-to-end tests that run against real `.lnk` files through real PowerShell/COM, covering dead-pin retarget, AUMID preservation across `Save()`, foreign-shortcut isolation, and the burned-root bail-out; a three-model review (Codex, GLM, Claude) produced five findings that were all folded in, including a meaningful one where the path guard was rejecting apostrophes/dollar signs and silently skipping repair for profiles like `C:\Users\O'Connor`. CI is still running at triage time; the main watch-item for follow-up is the separate race the author flagged where the in-app updater launches `Setup.exe` before `app.quit()` completes, which should be filed as its own issue. Assigning P2 — it is a visible, recurring defect on the primary platform with a complete and reviewed fix worth prioritizing for merge.

Issue #502, opened 2026-07-19
[Bug]: Installer/Squirrel updater crashes when run while wmux is already running


p-poppe:
### What happened?

Running the new-version `Setup.exe` to update **while wmux is already running** (window + tray icon) crashes. After fully quitting wmux (kill from tray / Task Manager) and re-running the installer, the update completes fine.

Note: the crash does not appear to be captured in the Squirrel updater logs — those only record the eventual *successful* run after the app is killed. Happy to provide logs / a Windows Event Viewer entry if it helps and I can reproduce on request.

### Steps to reproduce

1. Have wmux running (window + tray icon).
2. Download the new-version `Setup.exe`.
3. Run it to update **while wmux is still running**.
4. → crash.

### Expected behavior

The installer / Squirrel flow should detect the running instance and gracefully close it, hand off the update, or prompt the user to close it first — instead of crashing. (Most Electron/Squirrel auto-updaters relaunch the app after updating.)

Workaround today: fully quit wmux first, then run the installer.

### wmux version

3.27.0

### How did you install wmux?

Setup.exe (direct download)

### Shell in the affected pane

N/A — this is an update/installer-time crash, not tied to a specific pane or shell.

openwong2kim:
Ok I will check thank U very much

PR #869, opened 2026-08-12
fix(updater): start the Windows installer against a dead tree

### LINKED ISSUE #866: In-app update can destroy the install: Setup.exe starts wiping the root before the app has quit
## What happens

`AutoUpdater.performInstall()` launches the downloaded `Setup.exe` and only then asks the app to quit:

```ts
// src/main/updater/AutoUpdater.ts:547-563
const openErr = await shell.openPath(tempPath);   // 1. installer starts
...
app.quit();                                        // 2. quit begins
```

Nothing waits for the process to actually die. That matters because Squirrel's `Setup.exe` is not an updater — it is a full reinstall, and its **first** action is a recursive delete of the entire install root. From `%LOCALAPPDATA%\SquirrelTemp\Squirrel-Install.log` on a run I did on 2026-08-11:

```
[11-08-26 21:23:29] Starting Squirrel Updater: --install .
[11-08-26 21:23:29] Install path C:\Users\...\AppData\Local\wmux already exists, burning it to the ground
[11-08-26 21:23:29] fatal: System.UnauthorizedAccessException:
                    'C:\Users\...\AppData\Local\wmux\app-3.40.1\ffmpeg.dll' access denied
                      at Squirrel.Utility.DeleteDirectory
                      at Squirrel.Update.Program.Install
```

The delete starts in the same second the installer launches. If any file under the root is still mapped by a live wmux process, `DeleteDirectory` throws, the exception escapes `Program.Install`, and the install aborts **with the root already half-deleted**.

## Why it is worse than a failed update

The end state I measured: the root retained only a 2-file `app-3.40.1\`. `Update.exe`, the stub `wmux.exe` and `app.ico` were gone. Every shortcut still pointed at `<root>\wmux.exe`, which no longer existed — so all icons went blank and there was no launcher left to start the app or retry the update from. Recovery needed re-running Setup.exe by hand.

## Two lock holders, not one

1. **The GUI process.** `app.quit()` is asynchronous: before-quit handlers, window teardown, the session-save round trip. The comment at AutoUpdater.ts:556-561 already states the right intent ("quit NOW ... so Squirrel installs against a dead instance") — the code just does it in the wrong order and never waits.

2. **The daemon.** It runs as `wmux.exe <app-X.Y.Z>\resources\daemon-bundle\index.js`, and a normal quit only DETACHES (index.ts:337-344), so it keeps executing out of the directory the installer is about to delete. `squirrelTeardown.ts` spares the daemon deliberately (lines 22-24, the persistence promise), and it runs inside the *new* version's `--squirrel-install` hook, which is after the burn. Neither mechanism can protect this phase.

The #502 fix addressed a later phase (post-extract shortcut work), so it does not cover this.

## Reproduction

1. Install version N, let it run (the daemon comes up with it).
2. Run version N+1's `Setup.exe` while the app is alive.
3. Watch `%LOCALAPPDATA%\SquirrelTemp\Squirrel-Install.log` — the fatal lands within a second, and `%LOCALAPPDATA%\wmux` is left gutted.

Step 2 is a stronger version of what the in-app path does (there the app is at least *trying* to quit), so this reproduces the failure mode, not the exact race timing. Whether a normal in-app update loses the race in practice is the open question — the ordering bug is unambiguous either way.

## Direction

Do not launch the installer from the live process. Spawn a detached waiter that blocks until our PID (and the daemon's) is gone, then starts `Setup.exe`, and quit. That is the only ordering that makes "installs against a dead instance" true. The daemon has to come down with it — a root-nuking install replaces the daemon's own binaries, so "sessions survive an update" is not achievable on this path regardless, and the current promise costs the user their whole installation when it fails. The update UI copy would need to change with it.

Worth adding either way: if the app starts and finds the root missing `Update.exe` or the stub, we are in the post-abort state and should say so and offer a repair, instead of leaving someone with blank icons and no way back.

Found while root-causing #863. Separate bug, higher severity — #863 is a cosmetic icon, this destroys the installation.

openwong2kim:
## What this is, and what it is not

Squirrel's `Setup.exe` is a first-install program, not an updater: it deletes the whole install root as its first action. `performInstall` launched it and only then called `app.quit()`, so the delete began while our own processes still held the tree open.

**It is not the widespread breakage I first claimed.** I measured the real path end to end — installed 3.40.2, left the daemon and four MCP servers running, pressed Check for updates — and it worked. The delete failed and Squirrel simply carried on:

```
warn:  Install path ...\wmux already exists, burning it to the ground
error: DeleteDirectory: could not delete - app-3.40.2: IOException (in use by another process)
error: DeleteDirectory: could not delete - wmux: IOException (directory not empty)
info:  Writing files to app directory: ...\app-3.41.0
info:  Finished Squirrel Updater
```

The exception TYPE decides it, not the lock:

| scenario | exception | outcome |
|---|---|---|
| Setup.exe run against a fully live app | `UnauthorizedAccessException` on a FILE (`ffmpeg.dll`) | escapes `Program.Install` — fatal, root half-deleted, no launcher left |
| in-app update (app is quitting) | `IOException` on a DIRECTORY in use | caught, logged, install continues |

So this PR closes two narrower things:

1. **The manual-launch destruction.** Running the installer while wmux is fully up produces the file-level access-denied that Squirrel does not survive. That is the #502 scenario, and it is still reachable today.
2. **A window where the app has no launcher.** Watching the successful update, the root stub was missing for 4 seconds (10:15:08 → 10:15:11) before Squirrel re-rigged it. An install that dies in there strands the user with blank icons and nothing to run.

Nobody's routine update has been at risk. Issue lowered to P2 accordingly.

## Approach

The installer is started by a **detached waiter** that proceeds only once every process under the install root is confirmed gone AND the root is confirmed unlocked. Both checks are needed: the pid list cannot close the window in which an agent's MCP host spawns a fresh server into the directory we are about to delete.

A root that never clears **aborts and leaves a marker** rather than launching. Refusing costs the user a retry; proceeding is what costs them the installation.

```
performInstall (win32, packaged only)
  ├─ free-space preflight, budgeted per volume ── short ──► refuse, nothing written
  ├─ spawn waiter FIRST (captures a handle per pid)
  │     └─ null ──► refuse, nothing killed, no quit state armed
  ├─ ask for the full-shutdown branch (daemon flushes scrollback, then exits)
  ├─ force-kill what nothing else owns (MCP servers belong to agent hosts)
  └─ quit
        waiter: wait handles (bounded) → probe locks → launch, or abort with a marker
```

Also here:

- **Packaged-only guard.** The teardown derives the install root from `process.execPath`, which unpackaged is the runtime binary. Without the guard the enumerate-and-kill targets unrelated processes under it — the unit suite lost its own worker to exactly that before it existed.
- **The daemon is spared the force-kill** and gets the graceful shutdown so it flushes scrollback, via a new `onInstallRequiresFullShutdown` hook that reuses the tray's existing full-teardown branch.
- **A refusal is reported on the next boot** instead of looking like the update button did nothing.
- **The update widget now says installing ends running sessions**, in all 23 locales. The previous copy promised the opposite ("sessions persist in the daemon"), which is false on this path.

## Three-way review

Reviewed independently by Codex, GLM-5.2 and this session. Seven findings, all reproduced before being fixed:

- **BOM-less UTF-8 script** (GLM). Windows PowerShell 5.1 decodes a BOM-less file as the system ANSI code page. Measured with a Korean profile path: the waiter exits 0 while writing to a mangled path — silent success, the worst shape this failure can take, on the maintainer's own locale.
- **`.exe`-only lock probe** (all three). The delete that destroyed a machine died on `ffmpeg.dll`. Measured on a live install: 1 `.exe` locked vs 9 binaries. Probing every file is the complete answer at 10.8 s per pass against 0.7 s for binaries — far too slow for a loop that re-checks twice a second — so the probe covers `exe/dll/node/asar`.
- **Unbounded `WaitForExit`** (Codex). `taskkill` is best-effort; a survivor left the waiter blocked forever with wmux already gone, so the update stalled silently.
- **Full-shutdown latch set before the handoff was known to exist** (Codex). A failed handoff left the one-way flag armed, so a later ordinary Quit would tear the daemon down and close every session — for an install that never happened.
- **`Start-Process` failure swallowed** (Codex + GLM), reported as exit 0.
- **Marker cleared on read** (GLM) — a crash inside the notice delay lost the refusal permanently.
- **`lockBudgetMs` unvalidated** (GLM) — a zero budget turns the lock gate into a no-op.

## Testing

- `vitest run src/main/updater/__tests__/ src/renderer/i18n/__tests__/updateSessionsLocale.test.ts` — **118 passed**
- `vitest run --config vitest.runtime.config.ts src/main/updater/__tests__/installTeardown.runtime.test.ts` — **8 passed**, against real processes, real file locks and the real waiter script: blocks while a tracked process lives, aborts when an untracked process holds the root (the TOCTOU case), aborts when only a DLL is locked (the file the field failure died on), gives up rather than waiting forever, and keeps the script outside the root Setup.exe deletes.
- `tsc --noEmit` clean; eslint adds no errors.

Not covered: a packaged N→N+1 run driven from the fixed build's own update button. The fix cannot be exercised that way until it ships in a release, since the code that runs during an upgrade belongs to the version being upgraded *from*.

Refs #866

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows update installation by safely closing active sessions, waiting for files and processes to become available, and verifying installation readiness.
  * Updates no longer launch when preparation fails, insufficient disk space is detected, or the installation directory remains in use.
  * Failed installation handoffs now display a notice on the next launch.

* **New Features**
  * Added localized warnings explaining that installing an update closes active sessions, available in all supported languages.

Closes #866

### DIFF
--- a/changelog.d/869.md
+++ b/changelog.d/869.md
@@ -1,0 +1,24 @@
+### Fixed
+
+- **Running the Windows installer while wmux is open can no longer destroy the
+  installation.** Squirrel's `Setup.exe` deletes the whole install folder as its
+  first step, and wmux used to start it without waiting to actually exit. With
+  the app fully up its loaded libraries are locked, the delete fails outright,
+  and the install stops half-finished — leaving blank shortcut icons and no
+  launcher to retry from. The installer is now started by a helper that waits
+  until nothing is using the folder any more, and refuses to start it at all if
+  something still is, so a failed update leaves your working version alone
+  instead of a broken one. The normal in-app update was not affected by this;
+  the failure needed the app to be fully running, as it is when the installer is
+  launched by hand. (#866)
+
+- **A failed update now says so.** If the installer could not be started safely,
+  wmux reports it the next time it opens instead of leaving you on the old
+  version with no explanation. (#866)
+
+### Changed
+
+- **Installing an update on Windows now ends running sessions, and says so
+  before you press the button.** The daemon holds the install folder open, so it
+  has to go down for the installer to run. The update panel previously implied
+  the opposite. (#866)
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -91,6 +91,7 @@ import { setConfiguredFirstPartyClients } from './mcp/firstParty';
 import { readConfiguredFirstPartyClients } from './mcp/firstPartyConfig';
 import { ClaudeWorker } from './a2a/ClaudeWorker';
 import { AutoUpdater } from './updater/AutoUpdater';
+import { readDaemonPid } from './updater/installTeardown';
 import { McpRegistrar } from './mcp/McpRegistrar';
 import { BrokerSupervisor, isMcpBrokerEnabled } from './mcp/BrokerSupervisor';
 import { WebviewCdpManager } from './browser-session/WebviewCdpManager';
@@ -404,6 +405,11 @@ const ptyBridge = new PTYBridge(ptyManager, () => mainWindow, () => hookSignalRo
 const autoUpdater = new AutoUpdater(() => mainWindow, {
   onBeforeInstallQuit: () => prepareInstallQuit(),
   onInstallQuitAborted: () => abortInstallQuit(),
+  // #866: the Windows installer deletes the install root, and the daemon runs
+  // out of it. Flip the same flag the tray's "Shut down wmux" item uses so
+  // before-quit takes the daemon.shutdown branch instead of detaching.
+  onInstallRequiresFullShutdown: () => { fullShutdownRequested = true; },
+  getDaemonPid: () => readDaemonPid(getWmuxDir()),
 });
 
 const rpcRouter = new RpcRouter();
--- a/src/main/updater/AutoUpdater.ts
+++ b/src/main/updater/AutoUpdater.ts
@@ -65,10 +65,6 @@ const INSTALL_STAGING_HEADROOM_BYTES = 200 * 1024 * 1024;
 // can hold files for a while, and waiting is free while refusing costs the user
 // their update.
 const INSTALL_LOCK_BUDGET_MS = 60_000;
-// The refused-install notice waits for the renderer to attach its update
-// listeners. Sending it at boot would drop it into a window that is not
-// listening yet, which is indistinguishable from never reporting at all.
-const REFUSED_INSTALL_NOTICE_DELAY_MS = 5_000;
 // Squirrel downloading and unpacking a ~120 MB bundle before it reports
 // 'update-downloaded'. Generous on purpose: a slow disk or an antivirus scan
 // makes this minutes, and aborting a healthy update is worse than waiting.
@@ -177,6 +173,14 @@ export class AutoUpdater {
   constructor(getWindow: () => BrowserWindow | null, hooks: AutoUpdaterHooks) {
     this.getWindow = getWindow;
     this.hooks = hooks;
+    // #866. Registered HERE, not in start(): start() runs at the end of the
+    // ready sequence, which waits on the daemon bootstrap — measured at 39s on
+    // a cold boot. The renderer mounts and asks for a refused install at
+    // ~0.7s, so a handler registered in start() is not there yet and the
+    // invoke rejects with "no handler registered", which is how this notice
+    // silently went missing in live dogfood. The constructor runs during
+    // module evaluation, before the window exists at all.
+    ipcMain.handle(IPC.UPDATE_TAKE_REFUSED_INSTALL, () => this.takeRefusedInstall());
   }
 
   start(): void {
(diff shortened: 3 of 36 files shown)


openwong2kim:
@codex review

openwong2kim:
@codex review

openwong2kim:

This PR fixes #866 — the Windows update path that could gut the installation — by replacing the launch-then-quit ordering with a detached PowerShell waiter that only starts `Setup.exe` once every process under the install root is confirmed gone **and** the root is confirmed unlocked. The new `installTeardown.ts` is the heart of the change: it enumerates processes by executable path (catching MCP servers the daemon-sparing `squirrelTeardown` misses), force-kills what nothing else owns, hands the daemon a graceful full-shutdown so it flushes scrollback, and refuses to launch rather than running against a locked tree — writing an abort marker the next boot surfaces as a real error instead of a silent no-op.

The scope is disciplined and the failure-mode reasoning is load-bearing throughout: a per-volume free-space preflight, a BOM on the waiter script (Windows PowerShell 5.1 decodes BOM-less UTF-8 as ANSI, which silently mangles Korean profile paths), handle-based waits instead of pid polls (pids recycle), binary-lock probes covering `.exe/.dll/.node/.asar` (the field failure died on `ffmpeg.dll`, not an exe), a packaged-only guard, and a one-way-shutdown latch armed only after the waiter is known to exist. Update-widget copy is corrected across all 23 locales to state that installing ends running sessions, since the daemon can no longer be spared on this path.

Testing is strong: 118 unit tests plus an 8-case runtime suite that exercises the waiter against real processes, real file locks, and the TOCTOU case where an untracked process re-locks the root. The one uncovered path — a packaged N→N+1 upgrade driven from the fixed build's own button — is correctly noted as unverifiable until this ships, since the code that runs during an upgrade belongs to the version being upgraded *from*.

**Priority: P2** — the issue was correctly lowered from P1 once measurement showed routine in-app updates were never at risk (Squirrel catches the directory-level `IOException` and continues); the destruction requires a manual `Setup.exe` launch against a fully-live app. Worth merging ahead of the next release so the repair-on-boot and honest update copy land together.

PR #870, opened 2026-08-12
perf(main): resolve PTY ownership and role bindings mirror-first

openwong2kim:
Every external MCP call used to pay one to three main→renderer round-trips
through `input.findOwnerWorkspace`: the ownership assert, the caller-identity
resolution, and the D2 role→model lookup. The renderer answers those late
exactly when the app is busy — a large-buffer flush storm is precisely when
the latency is felt — so the hot path depended on the slowest thing in the
process.

Main already keeps a renderer-pushed `WorkspaceMirror` (structural changes push
immediately, status churn debounced 300ms, plus a 30s refresh). This makes it
the first answer and keeps the round-trip as the fallback.

## What changed

A shared resolver, `src/main/workspace/ptyOwnership.ts`, with two postures:

- **assert** (an expected workspace is known — `input.send`, `deck.resolvePaneRoute`):
  the mirror may only short-circuit an **allow**, and only when it agrees with
  the expectation. Any miss, staleness, or disagreement falls through to the
  round-trip, which stays the sole deny authority — a stale mirror can never
  produce a false reject.
- **resolve** (caller identity from a verified `senderPtyId` — `events.poll`,
  `fanout`, `meta`, `a2a`, `a2a.channel`): a fresh mirror hit answers; misses
  still round-trip and the fail-closed `''` contract at each call site is
  preserved.

Freshness is bounded by the existing `STALE_TRUST_MS` (10s). Worth stating
plainly for review: this trades "always ask the freshest source" for a
bounded-staleness allow. Structural changes push immediately, so the realistic
exposure is IPC delivery latency, and the `senderPtyId` anchor is advisory
attribution under the #113 same-user ceiling by design (see the note in
`events.rpc.ts` — the unforgeable fix is peer-PID, deferred). The mirror does
not change that ceiling.

Role bindings now ride the mirror push payload as a complete `ptyId → binding`
map built by the same resolution the round-trip performs, so a bound pane's
model enforcement no longer costs a second round-trip on
`terminal_send submit:true`. A renderer that predates the field pushes nothing,
which the resolver treats as *unknown* and round-trips — never as "unbound".

## Review findings folded in

Three-way review (Claude / Codex / GLM-5.2) caught a silent one: the IPC parser
rebuilt `{ts, entries, fleets}` and **dropped** the new `roleBindings` field, so
every production push looked like an old renderer and the fast path was dead on
arrival. The unit tests missed it because they called `setSnapshot` directly,
bypassing the parser. There is now an end-to-end test that walks
parse → setSnapshot → resolve, plus parser cases for present/absent/empty/junk.

Also from review: `MirrorSnapshotState` indexes its role maps off `StoreState`
so a store rename breaks the build instead of silently producing an empty map
main would trust as "nothing bound"; and `peekRoleBinding(ptyId)` replaced a
whole-map copy per lookup on the `input.send` path.

## Verification

Unit: 12 new resolver tests (driving the real `WorkspaceMirror` singleton, not a
mock) plus the parser suite; full suite green apart from two pre-existing local
failures unrelated to this branch.

Live dogfood on a dev instance:
- cross-workspace `input.send` still denied (`not owned by workspace …`), so the
  mirror short-circuit never weakened the gate
- under renderer load (20k lines streaming), `events.poll` (mirror-resolved)
  median 4ms / max 5ms vs `input.readScreen` (round-trip by design) median 8ms /
  max 12ms

### DIFF
--- a/changelog.d/870.md
+++ b/changelog.d/870.md
@@ -1,0 +1,9 @@
+### Changed
+
+- **Terminal ownership checks stop waiting on the UI.** Every MCP call an agent
+  makes — sending to a pane, polling events, resolving its own identity — used
+  to ask the renderer who owns a terminal, once per check. That question was
+  slowest to answer exactly when the app was busiest. Main now answers from the
+  workspace snapshot the renderer already pushes it, and falls back to asking
+  whenever the snapshot is stale, missing, or disagrees. Cross-workspace access
+  is still refused on the renderer's word, never on a cached guess. (#870)
--- a/src/main/ipc/handlers/__tests__/workspaceMirror.handler.test.ts
+++ b/src/main/ipc/handlers/__tests__/workspaceMirror.handler.test.ts
@@ -91,3 +91,36 @@ describe('parseWorkspaceMirrorPayload — defensive renderer-trust validation', 
     expect(pane?.cwd).toBeUndefined();
   });
 });
+
+describe('parseWorkspaceMirrorPayload — roleBindings passthrough (3-way review: Codex)', () => {
+  // The original fast-path change validated the mirror by calling setSnapshot
+  // directly, bypassing this parser — which silently DROPPED roleBindings, so
+  // the resolver fell back to the round-trip on every call. These tests pin
+  // the production path: whatever reaches setSnapshot must carry the field.
+  const base = { ts: 1, entries: [], fleets: [] };
+
+  it('forwards a present roleBindings map (values opaque, keys validated)', () => {
+    const parsed = parseWorkspaceMirrorPayload({
+      ...base,
+      roleBindings: { 'pty-1': { agent: 'claude', model: 'haiku' }, '': { dropped: true } },
+    });
+    expect(parsed?.roleBindings).toEqual({ 'pty-1': { agent: 'claude', model: 'haiku' } });
+  });
+
+  it('an ABSENT field stays absent (old renderer ⇒ resolver must round-trip)', () => {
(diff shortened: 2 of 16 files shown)


openwong2kim:

This PR removes the main→renderer round-trip that every external MCP call used to pay for PTY ownership and caller-identity resolution, by answering from the `WorkspaceMirror` the renderer already pushes to main and keeping the round-trip as the fallback. The new shared resolver in `workspace/ptyOwnership.ts` exposes two clearly separated postures — `assert` (mirror may only short-circuit an allow, never a deny) and `resolve` (fresh mirror hit answers, misses still round-trip, fail-closed `''` preserved at each call site) — which keeps a stale mirror from producing a false reject. Role bindings now ride the mirror push payload as a complete `ptyId → binding` map, so model enforcement on `terminal_send submit:true` no longer costs a second round-trip, and an old renderer that omits the field is treated as unknown rather than "unbound". The parser-level `roleBindings` passthrough bug caught in three-way review (the rebuilt `{ts, entries, fleets}` silently dropped the field, leaving the fast path dead on arrival) is now pinned by end-to-end parse → setSnapshot → resolve tests, which is the right place to guard it.

**Priority: P2** — meaningful latency win on the hot path every agent RPC traverses, security-relevant IPC paths handled with fail-safe fallbacks, and thoroughly tested; worth a careful read on the `STALE_TRUST_MS` allow-vs-deny asymmetry at merge time.

openwong2kim:
@codex review

PR #871, opened 2026-08-12
perf: daemon log buffering, renderer hot-path fixes, pane.search tail bounding

openwong2kim:
Four independent hot-path costs found in a renderer/main profiling pass. They
are unrelated to each other; the common thread is that each one taxes a path
that runs constantly, and on Windows every synchronous fs call also pays an
EDR scan hook.

## Daemon log buffering

`log()` did one `appendFileSync` per line — a write the calling thread waits
on, on a process whose hot paths are session churn, recovery, and RPC bursts.
Info/debug lines now coalesce for 250ms or 64KB in `src/daemon/logWriter.ts`.

`warn`/`error` deliberately do **not** buffer: they first drain any pending
info lines, so the file stays in true chronological order (it exists for
post-hoc causal reconstruction), then append synchronously so a dying daemon's
last words are already on disk. A clean exit flushes the tail; a hard kill can
lose at most the last 250ms of info lines, which is documented in the module.
The existing postmortem sinks (`daemon/util/logSink.ts`, `main/util/logSink.ts`)
are untouched — their all-sync behavior is their contract.

## Renderer one-liners

- `notePtyOutput` ran a regex replace on **every** output chunk to strip
  DSR/CPR cursor sequences. Chunks without ESC — the overwhelming majority —
  now take an `indexOf` fast path. The empty-chunk rule is preserved (an empty
  chunk is still not activity) and still tested.
- `markPtyReady` fired the whole store-update machinery on every repeat call;
  it now returns before `set()` once the pty is ready.
- `SurfaceTabs` subscribed to three whole store maps, so **any** pane's status,
  label, or agent change re-rendered **every** tab strip in the app. Narrowed
  to this pane's label and active-surface slug (primitives), with a per-tab
  status-dot child that subscribes to its own `surfaceAgentStatus` entry.

## pane.search

Two problems. It held the renderer main thread for the entire sweep, and it
scanned up to 20k rows **oldest-first** per buffer — so on a large scrollback
the newest output, which is what a cross-pane search almost always wants, was
the part that got cut.

It now yields between panes and scans the newest `searchTailLines` rows
(default 5000, clamped to the 20k cap), reporting `truncated: true` whenever
older rows went unscanned, on both the live and parked-pane paths.

The yield uses `MessageChannel`, not `setTimeout(0)`: timers in a backgrounded
window are throttled to ≥1s each, which on a 20-pane workspace would add ~20s
and blow the 10s MCP deadline the surrounding code already guards against.

`wmux_search_panes` gains an optional `searchTailLines` parameter. The
generated API reference is unchanged (it derives from `rpc.ts`/`events.ts`, not
the MCP zod schemas) — regenerated and confirmed byte-identical.

## Review findings folded in

Three-way review (Claude / Codex / GLM-5.2):

- **Truncation lied above the cap.** With `searchTailLines=50000` and a 30k-row
  buffer the engine scanned only the newest 20k while the check compared
  against the unclamped request and reported `truncated: false` — a false
  completeness signal in exactly the search-deeper case the parameter exists
  for. All consumers now go through `normalizeSearchTailLines` before any scan
  or truncation check.
- **`lineIdx` went window-relative** under a tail window, so a match at row
  9000 of a 10k buffer displayed as "matched line ~4000" in the search UI. It
  is buffer-absolute again.
- **The in-app search bar inherited the agent-facing default.** It rides the
  same handler, so a human searching their own scrollback silently skipped the
  older half of a default 10k buffer, with no way to widen it. The human entry
  point now passes the user's configured scrollback depth.

## Verification

Unit: new tests for the log writer (ordering, force-flush, rotation, exit
drain), the clamp, absolute `lineIdx`, and the ESC fast path. Full suite green
apart from two pre-existing local failures unrelated to this branch.

Live dogfood on a dev instance, with a marker buried under 7000 lines:
- agent search (default tail) misses it and reports `truncated: true`; the
  human search bar finds it — the regression fix, demonstrated as an A/B on the
  same marker
- newest marker reports `lineIdx: 7004` (it would have been ~2000 with the
  window-relative bug)
- `searchTailLines: 20000` finds the buried marker; `50000` is clamped and
  still reports honestly; `0` is rejected
- `daemon.log`: 1104 lines, zero out-of-order timestamps, including a `warn`
  line — the drain-then-write-through path holds

## Summary by CodeRabbit

* **New Features**
  * Pane searches now prioritize recent output and support configurable search depth.
  * Search results indicate when older output may not have been included.
* **Improvements**
  * Searches remain responsive when scanning multiple panes.
  * Daemon logs are batched for efficiency while warnings and errors appear immediately and in order.
  * Log rotation and shutdown flushing are handled more reliably.
* **Bug Fixes**
  * Terminal tab updates are limited to the affected pane.
  * Reduced unnecessary processing during terminal output handling and readiness updates.

### DIFF
--- a/changelog.d/871.md
+++ b/changelog.d/871.md
@@ -1,0 +1,22 @@
+### Changed
+
+- **Cross-pane search reads the newest output first.** `wmux_search_panes` used
+  to scan up to 20,000 lines per pane starting from the oldest, so on a long
+  scrollback the recent output you were actually looking for could be the part
+  that got cut. It now searches the newest 5,000 lines by default, says
+  `truncated: true` when older lines went unread, and takes a new optional
+  `searchTailLines` to go deeper. It also yields between panes instead of
+  freezing the window for the whole sweep. The in-app search bar is unaffected
+  and still covers your full configured scrollback. (#871)
+
+### Fixed
+
+- **The daemon no longer stalls on its own log file.** Each log line was written
+  and flushed to disk synchronously, which on Windows also meant a virus-scanner
+  hook per line. Routine lines are now batched; warnings and errors still write
+  through immediately, after the pending lines, so `daemon.log` stays in order
+  and a crash still records what happened. (#871)
+- **Terminal tab strips stopped re-rendering on unrelated panes.** Any pane's
+  status, label, or agent change re-rendered every tab strip in the app; each
+  strip now watches only its own pane. Terminal output also skips a
+  per-chunk regular-expression scan it did not need. (#871)
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 80000,
-      "wireResultSha256": "6b51db8f0b0b906e35903f48ed371ab1622b1297d0cad613106547d7baf59ee3",
+      "wireResultSha256": "43a8ff42f7d8c85aa201d30bbada14c3280888e4830afc60bd3b7ebfba4261c3",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
@@ -97,7 +97,7 @@
     },
     "commander": {
       "maxListBytes": 47000,
-      "wireResultSha256": "68529d69c040a8b11d2da12b0ac82ecd4b560dd42a90a1acd5e7e15ca8ab7db3",
+      "wireResultSha256": "21d87db3c04f214f98c2bee54dda1a622970204c88233812df5a8c5634f8d13f",
       "instructionSha256": "5a97f851c892326eba7cbcf46292f6925834d5794172461d1ced997d90ac1deb",
       "toolNames": [
         "terminal_read",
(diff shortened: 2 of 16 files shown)


openwong2kim:

This PR lands four independent hot-path performance fixes surfaced in a renderer/main profiling pass, each targeting a path that runs continuously: daemon log buffering, a redundant per-chunk regex in `notePtyOutput`, repeated full store updates in `markPtyReady`, and an over-broad `SurfaceTabs` subscription that re-rendered every tab strip on any pane change. The most user-visible change is `pane.search`: it now yields between panes via `MessageChannel` (not `setTimeout(0)`, which background-tab throttling would inflate past the 10s MCP deadline) and scans the newest `searchTailLines` rows first (default 5000, clamped to the 20k cap) with `truncated: true` reporting, so cross-pane search stops blocking the renderer and stops cutting the newest output — the part actually wanted. The daemon log writer buffers info/debug for 250ms / 64KB while `warn`/`error` drain pending lines first then write synchronously, preserving chronological order and crash forensics; the 250ms tail-loss window under hard kill is an explicit, documented trade-off, and the all-sync postmortem sinks stay untouched.

**Priority: P2** — four well-scoped perf wins on constantly-running paths (log writes, terminal output, tab rendering, cross-pane search), with EDR-amplified Windows fs cost called out correctly and behavior pinned by new/updated tests.

openwong2kim:
@codex review

PR #872, opened 2026-08-12
chore(release): 3.41.1

openwong2kim:
Folds the three pending changelog fragments, bumps the version in `package.json`
and both `package-lock.json` sites, and regenerates `docs/api/reference.md` so
its baked-in version matches (the CI drift guard checks this).

## What ships

**The Windows installer no longer runs against a live install root.** This is
what closes the known-issue notice 3.41.0 shipped with: `Setup.exe` deletes the
whole install directory as its first step, and the updater used to start it
without waiting for wmux to actually exit. The installer is now started by a
helper that waits for the directory to be free and refuses to start at all if it
is not, so a failed update leaves the working version alone.

Note the caveat that applies to this upgrade too: the step at fault belongs to
the version you are upgrading *from*, so 3.41.0 → 3.41.1 still carries the old
behavior. The fix protects the upgrades after it.

**A refused install now reports itself on the next launch.** Found by dogfooding
#869 after it was otherwise complete: the notice was pushed to a listener that
only exists while the Settings panel is open, so with Settings closed nothing
received it and the marker was cleared anyway. The renderer now pulls it from an
always-mounted surface, and taking it is what clears the marker.

**Two performance changes.** Main resolves PTY ownership and role bindings from
the workspace snapshot the renderer already pushes it, instead of asking the UI
on every MCP call — the round-trip stays as the fallback and remains the sole
deny authority. The daemon batches its log writes (warnings and errors still
write through, in order), and cross-pane search reads the newest output first
instead of the oldest.

## Release checklist

- [x] `node scripts/collect-changelog.mjs` — 3 fragments folded, files removed
- [x] `[Unreleased]` renamed to `[3.41.1] — 2026-08-13`
- [x] version bumped in `package.json` + both `package-lock.json` sites
- [x] `node scripts/gen-api-reference.mjs` re-run (header now reads v3.41.1)
- [ ] `v3.41.1` tag pushed after merge — installer builds hang off the tag

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Windows update startup, shutdown messaging, and failure reporting.
  * Strengthened terminal ownership checks when workspace snapshot data is stale, missing, or inconsistent.
  * Prevented cross-workspace terminal access when authoritative pane information disagrees.

* **Improvements**
  * Cross-pane search now prioritizes recent output, reports truncation, and remains responsive across panes.
  * Improved daemon log delivery, with warnings and errors written immediately.
  * Reduced unnecessary terminal tab-strip and output re-rendering for smoother performance.

* **Documentation**
  * Published release notes and updated the API reference for version 3.41.1.


openwong2kim:

This is the release PR for wmux 3.41.1, a patch release that folds three pending changelog fragments (#869, #870, #871) into `CHANGELOG.md`, bumps the version in `package.json` and both `package-lock.json` sites, and regenerates `docs/api/reference.md` so its baked-in header matches the new version (the CI drift guard checks exactly that). The diff is purely mechanical: the three `changelog.d/*.md` fragments are removed verbatim and reappear, lightly reorganized under `### Changed` and `### Fixed`, in the new `[3.41.1] — 2026-08-13` section; the version strings move from `3.41.0` to `3.41.1`.

What ships under those notes matters: the Windows installer can no longer destroy the installation when launched while wmux is open (#866), terminal ownership checks stop round-tripping through the renderer on every MCP call (#870), cross-pane search reads the newest output first instead of the oldest (#871), and the daemon stops syncing every log line to disk and stalling under antivirus (#871). Each of those was merged in its own PR; this one is just the release cut.

The remaining checklist item is tagging `v3.41.1` after merge, since the installer builds hang off the tag. The installer-fix caveat from the PR body is worth keeping in mind: the protection applies to upgrades *after* 3.41.1, so a 3.41.0 → 3.41.1 step still carries the old `Setup.exe` behavior. Safe to merge once the tag build path is ready.

PR #877, opened 2026-08-13
fix(daemon): preserve events across subscribe handshake

snowyukitty:
## Summary

- retain daemon push events generated between socket accept and a first-RPC subscription
- flush the bounded backlog in order before the subscribe reply while keeping ordinary RPC clients reply-only
- disconnect on backlog overflow instead of continuing after silent partial loss
- document the first-RPC subscription contract and cover replay ordering plus overflow

Addresses gap 1 in #858. Gap 3 (the first-party subscription gate) remains deliberately out of scope.

## Verification

- `npm run build:daemon`
- `npx eslint src/daemon/DaemonPipeServer.ts src/daemon/__tests__/DaemonPipeServer.eventOptIn.test.ts --quiet`
- `npx vitest run src/daemon/__tests__/DaemonPipeServer.eventOptIn.test.ts src/daemon/__tests__/DaemonPipeServer.subscriberGuards.test.ts` (15 passed)
- `npm run test:parallel` (10,947 passed, 24 skipped)
- `npm run test:runtime` (165 passed, 5 failed in unrelated Windows runtime fixtures)
  - `bootMarker.runtime.test.ts` passed on isolated retry
  - two `webRestart.runtime.test.ts` mtime assertions fail identically on clean `upstream/main`
  - two `shellIntegration.runtime.test.ts` cases fail because node-pty reports `AttachConsole failed`

## Summary by CodeRabbit

* **New Features**
  * Preserves daemon events generated during the app’s subscription handshake and delivers them in order.
  * Supports one-time session and channel transition events without losing updates.
  * Reconnects clients when the pre-subscription event backlog exceeds its limit, allowing the app to rehydrate safely.

* **Documentation**
  * Documented first-request subscription behavior, event buffering, backlog handling, and reply correlation requirements.

### DIFF
--- a/changelog.d/877.md
+++ b/changelog.d/877.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- **Daemon events no longer disappear during the subscription handshake.** The
+  daemon now preserves events emitted after accepting the app's control socket
+  but before its subscription request arrives, so one-shot session and channel
+  transitions reach the app instead of falling into that window. (#877)
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -256,6 +256,8 @@ A control connection can carry **two kinds of frame in the same newline-delimite
 
 **Events are opt-in.** A connection receives replies and nothing else until it calls `daemon.events.subscribe`; `daemon.events.unsubscribe` turns them off again. The subscription is per-connection and dies with the socket, so a client that reconnects MUST subscribe again. A client that only makes requests — the CLI and the MCP server are both in this category — never needs to touch either method, and its stream is then a plain request/reply stream.
 
+The first-party app sends `daemon.events.subscribe` as its first authenticated RPC. The daemon retains a bounded per-connection backlog between accepting that socket and dispatching its first authenticated request. A first-request subscribe flushes those frames in order before the subscribe reply, closing the accept-to-subscribe loss window; any other first request discards the backlog and leaves the connection reply-only. If the bounded backlog overflows, the daemon closes the socket so the app reconnects and rehydrates rather than continuing after silent partial loss. A client that intends to receive pushed events SHOULD therefore subscribe first and MUST correlate replies by `id`, including while the subscribe request itself is pending.
+
 Once subscribed, the two frame kinds are told apart by `id`:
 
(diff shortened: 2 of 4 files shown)


openwong2kim:

This PR closes gap 1 of #858 — daemon push events generated in the short window between socket accept and a client's first authenticated `daemon.events.subscribe` were silently dropped, taking one-shot transitions like `session.died` or `channel.message` with them. The fix introduces a bounded per-connection backlog (1 MB) that queues broadcast frames until the first authenticated RPC decides intent: a `daemon.events.subscribe` flushes the backlog in order before the subscribe reply, while any other first RPC discards it and keeps the connection reply-only as the opt-in contract already promises. Overflow does not truncate silently — the daemon closes the socket so the app reconnects and rehydrates rather than continuing with a partial view, which is the right failure mode.

The change is small and well-targeted: 74 net lines in `DaemonPipeServer.ts` plus matching tests for replay ordering and the overflow disconnect path, a two-line PROTOCOL.md update documenting the first-request contract, and a changelog fragment. The structural choice to read `actual` styles back and keep the backlog per-socket (cleaned up on `close` and `stop`) avoids leaking state, and ordinary RPC-only clients (CLI, MCP server) never pay the cost because the backlog is only populated while at least one fresh socket is pending. Verification in the PR description is thorough — 10,947 parallel tests green and the new cases passing.

Priority P3: this is a meaningful reliability improvement to a narrow window, but the window itself is a single local round-trip and the parent tracking issue #858 is already P3, so it does not need to jump the queue.

openwong2kim:
Thanks for this — the accept-to-subscribe window was a real hole, and one-shot transitions like `session.died` were exactly the events it swallowed. Making the first authenticated RPC the arbiter is a clean way to close it without weakening the opt-in default, and I appreciate that you wrote the contract into `docs/PROTOCOL.md` rather than leaving it implicit in the code. Disconnecting on overflow instead of truncating is the right call too.

Two things I noticed while reading, neither blocking, both fine as follow-ups:

- An ordinary RPC client can be disconnected by an overflow it never opted into, since the backlog accumulates on every fresh socket until its first request. The window is tiny so I doubt it bites in practice, but the warning also calls that client an "event subscriber" when it may not be one.
- If the flush hits backpressure, `subscribeEvents` returns false after the backlog has already been dropped. A client that just retries subscribe then succeeds against an empty backlog, which is the quiet partial loss the overflow path deliberately avoids.

Merging as-is. Thanks again.
@

PR #880, opened 2026-08-13
fix(i18n): translate the toolbar and Git labels left in English in zh

openwong2kim:
## Summary

Three `zh.ts` values were carried over verbatim from `en.ts` when #876 completed the locale. Key parity does not catch this — the key is present, the string is just still English — so it only surfaces with the app actually running in Chinese, which is how it was found.

- `toolbar.broadcast`: `Broadcast` → `广播`
- `toolbar.fanOut`: `Multi Task` → `多任务` (`ko.ts` already translates this one)
- `git.pullRequests`: `Pull Requests` → `拉取请求`

The first two sit on the agent toolbar, which is on screen the whole time, so they read as English right next to 附加 / 文件 / 片段 / 富文本输入.

Also brings `toolbar.broadcastTooltip` and `toolbar.broadcastPrompt` into step with the button they describe. Both opened with a literal `Broadcast — ` while `toolbar.broadcastTitle` had already settled on 广播, so translating the button alone would have left the control disagreeing with its own tooltip.

## What is deliberately left alone

Five other `zh` values are identical to `en` and should stay that way: `Finder`, `LanLink`, `main` (the branch name), `NAME` (an env-var placeholder), and `shell`.

## Verification

- Found by running the app in Chinese and reading the toolbar, not by a test — see the note below.
- `npx eslint src/renderer/i18n/locales/zh.ts --quiet`
- Key parity against `en.ts` re-checked after the change: 1203/1203, 0 missing, 0 extra, 0 `{var}` placeholder mismatches.

## Follow-up worth having

There is no locale contract test in this repo — `en.ts` is the SSOT and every other locale is a `Partial` with an English fallback, so both missing keys and untranslated values degrade silently. A check that flags "value identical to `en` and contains no script from the target locale", with a small allowlist for proper nouns, would have caught this before it shipped.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Simplified Chinese translations for the agent toolbar’s Broadcast and Multi Task buttons.
  - Updated the Git panel’s Pull Requests heading to display in Simplified Chinese.
- **Documentation**
  - Added a changelog entry documenting the localization improvements.

### DIFF
--- a/changelog.d/880.md
+++ b/changelog.d/880.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- **Three labels stayed in English when wmux ran in Simplified Chinese.** The
+  agent toolbar's Broadcast and Multi Task buttons and the Git panel's Pull
+  Requests heading were carried over verbatim from the English locale, so they
+  read as English in an otherwise translated UI. (#880)
--- a/src/renderer/i18n/locales/zh.ts
+++ b/src/renderer/i18n/locales/zh.ts
@@ -370,10 +370,10 @@ export const zh = {
   'settings.inspect.proxyLabel': '编辑 {token}（{role}）',
 
   // ─── 智能体工具栏: fan-out / broadcast 按钮 ───
-  'toolbar.fanOut': 'Multi Task',
-  'toolbar.broadcast': 'Broadcast',
-  'toolbar.broadcastTooltip': 'Broadcast — 向当前工作区的每个终端面板发送相同文本（无隔离）',
-  'toolbar.broadcastPrompt': 'Broadcast — 发送到当前工作区的每个终端面板',
+  'toolbar.fanOut': '多任务',
+  'toolbar.broadcast': '广播',
+  'toolbar.broadcastTooltip': '广播 — 向当前工作区的每个终端面板发送相同文本（无隔离）',
+  'toolbar.broadcastPrompt': '广播 — 发送到当前工作区的每个终端面板',
 
   // ─── Fan-out 对话框（1 个提示 → N 个隔离任务）───
   'fanout.title': 'Fan-out — 1 个提示 → N 个隔离任务',
@@ -1065,7 +1065,7 @@ export const zh = {
   'git.create': '创建',
   'git.createFailed': '创建失败',
   'git.newBranchPlaceholder': '新分支名…',
-  'git.pullRequests': 'Pull Requests',
+  'git.pullRequests': '拉取请求',
   'git.noPrs': '没有打开的 pull request。',
   'git.noComments': '没有评论。',
   'git.ghMissing': '未安装 GitHub CLI（gh）— 安装后即可在此查看 pull request。',


openwong2kim:
@cnxiekun heads-up — this is a small follow-up to your locale work, not a correction of it. I dogfooded the app in Chinese after merging #876 and spotted three values that came across from `en.ts` untranslated. Key parity passes on all of them, which is exactly why neither CI nor my own pre-merge check saw it.

You are the native speaker here, so please push back on any of these if they read wrong:

- `广播` for Broadcast — chosen because `toolbar.broadcastTitle` in your PR already used 广播到终端, so the button now matches its own dialog.
- `多任务` for Multi Task. Note your `fanout.title` keeps `Fan-out` as a term of art, which I left alone; if you would rather the button keep the English term too, say so and I will drop that hunk.
- `拉取请求` for Pull Requests.

I also updated the two broadcast tooltip strings, which opened with a literal `Broadcast — `, so the control does not disagree with its own tooltip.

Your instinct on all three beats mine — happy to change any of them.

openwong2kim:

This PR fixes three Simplified Chinese (`zh.ts`) locale values that were carried over verbatim from `en.ts` during #876, leaving them in English within an otherwise translated UI: the agent toolbar's `Broadcast` (→ 广播) and `Multi Task` (→ 多任务) buttons, and the Git panel's `Pull Requests` (→ 拉取请求) heading. The two broadcast tooltip strings are also brought into step with the button they describe, so the control no longer disagrees with its own tooltip. The change is low-risk — five value swaps in a single locale file plus a changelog entry — and key parity (1203/1203) was re-checked after the edit, which is the exact gap that let these slip through in the first place. Worth noting the follow-up the author flagged: the repo has no locale contract test, so untranslated values degrade silently under the English fallback, and a check that flags values identical to `en` would have caught this pre-merge. I'd treat that as a separate, optional improvement rather than blocking here.

**Priority: P3** — a self-contained cosmetic i18n fix with no runtime impact.

openwong2kim:
Merging this so the labels are not sitting in English through the next release. @cnxiekun the question about the wording still stands — you are the native speaker and these are your strings. If `多任务` reads wrong next to the `Fan-out` you kept in `fanout.title`, or if you would pick different characters for any of the three, say so and I will change them in a follow-up. Merging first is about timing, not about the wording being settled.

cnxiekun:
Thanks — all three read correctly to me as a native speaker:
  - 广播 (Broadcast) ✓ — matches 广播到终端 already used in the dialog.
  - 拉取请求 (Pull Requests) ✓ — this is GitHub's official Chinese term too.
  - 多任务 (Multi Task) ✓ — reads fine. I'm comfortable keeping Fan-out in the fan-out dialog title as a product term; the mixed form doesn't feel wrong. 

Keeping Finder / LanLink / main / NAME / shell in English looks right.
No changes needed from my side.

PR #881, opened 2026-08-13
Repaint every visible pane when the wmux window regains focus

### LINKED ISSUE #879: [Bug]: Black screen repaint
## Summary

Switching away to another application and back to wmux leaves the active pane fully black except for the cursor. The pane is not actually empty - forcing a repaint (e.g. selecting all text in the pane) reveals it was displaying stale, corrupted content the whole time.
Tried with both of these toggled on and off - same result.
Image

Independent of shell type.
| Starting view | After switching back from alt-tab | Selecting everything after alt-tabbing back|
| ---------------|-------------------------------------| ------------------ |
| Image | Image | Image |

## Environment

- wmux version: 
- Install method: winget (`openwong2kim.wmux`)
- OS: Windows

WindowsProductName WindowsVersion OsBuildNumber OsArchitecture
------------------ -------------- ------------- --------------
Windows 10 Home    2009           19045         64-bit

## Actual

The pane stays black until a manual redraw is forced (e.g. via text selection). Once forced, the revealed content is not the correct prior output — it's illegible, fragment-level glyph corruption (partial glyph pieces, mixed colors, no readable structure), distinct from the clean-but-wrong mojibake seen elsewhere (see separate mojibake issue) and from the single-character swaps seen in intermittent corruption (see separate issue).

## Notes

This looks like a stale or uncleared frame being displayed after the window was occluded — consistent with Chromium/Electron paint-throttling on hidden windows, where a full repaint isn't forced when the window becomes visible again. Because select-all is what triggers the reveal (not just refocusing), the underlying buffer isn't being cleared/redrawn automatically on focus regain.

The severity of the corruption revealed likely correlates with how much PTY output backed up while the window was occluded — a larger backlog getting parsed/painted at once may make related parsing issues (see intermittent corruption issue) more likely to surface.

## Screenshots

(attach black-screen.png and revealed-garbled-content.png)

### Steps to reproduce

1. Run a command that produces multi-line, colored output (e.g. `wmux doctor`).
2. Alt-tab to a different application.
3. Alt-tab back to wmux.
4. Observe the pane is solid black except for a blinking cursor.
5. Select all text in the pane (click-drag or Ctrl+A) to force a redraw.

### Expected behavior

The pane repaints its actual content immediately on regaining focus/visibility.

### wmux version

3.38.6

### How did you install wmux?

winget

PowerShell 7+ (pwsh)

### Windows version

Windows 10 22H2 (build 19045.6466)

Claude Code

Here I only recording the wmux app and you can hear me clicking alt tab switching to different windows and when I click back onto the wmux, I get this black screen

https://github.com/user-attachments/assets/8916b49b-f436-4641-b790-4e9465e5ea2b

### Confirmation

openwong2kim:
Alt-tab away from wmux on Windows and come back, and the pane can be sitting there empty. It stays empty until you force a redraw by hand — select-all, or drag a pane border.

## What is actually wrong

The reporter's own screenshots settle it: their forced redraw brings the text back **correct**, with only the bottom row showing glyph fragments. The buffer was never damaged. The pane is not rendering corrupt content, it is rendering nothing, and nothing ever tells it to draw again.

Nothing can, on Windows. Measured against a bare Electron 41 build:

| window state | `document.visibilityState` | rAF |
|---|---|---|
| foreground | visible | 60/s |
| fully covered by another app | visible | 60/s |
| **minimized** | **visible** | 60/s |

`visibilitychange` fires exactly once per window, at teardown, and toggling Chromium's `CalculateNativeWinOcclusion` changes nothing — this is not a tunable.

Every repaint-on-wake path in the renderer keys off that signal, so on Windows they are all dead code: `atlasWakeRecovery`'s `visibility` trigger (leaving `system-resumed` as its only live trigger there) and `useTerminal`'s `docRevealed` reclaim. The one repaint that does run on alt-tab belongs to the textarea focus handler of the single pane holding keyboard focus, so in a split every other visible pane has no trigger at all.

Two more things measured against a real xterm 6 + `addon-webgl` 0.19 canvas, with the drawing buffer wiped behind xterm's back (which is what a dropped buffer looks like to xterm — its model still says everything is drawn):

- an idle pane **never** self-heals;
- a single `terminal.refresh(0, rows - 1)` restores it exactly. The heavier `clearTextureAtlas` is not needed.

## What this changes

Window refocus becomes its own wake boundary. One app-wide coordinator (`terminal/windowWakeRepaint.ts`, following the `atlasGuard` singleton + `AppLayout` init pattern) owns a single blur/focus listener pair. Panes register a visibility probe and their existing `glyphRepaint` scheduler under a new `window-wake` reason, so there is no second timer lifecycle around the same `terminal.refresh`.

- Repaints on the next frame, and once more 250 ms later. The pixel loss can land on either side of re-activation; the frame pass covers a loss that happened while the window was away, the timer covers one that lands at re-activation. If the reporter confirms the frame pass alone is enough, the timer goes.
- Re-checks window focus and pane visibility at execution time, so a pass that outlives the focus does nothing, and a pane in a hidden workspace is skipped (it already repaints via `glyphRepaint`'s `visible` reason when its tab comes back).
- Seeds its state from `document.hasFocus()` rather than from "we have seen a blur", so a pane that mounts while the app is already in the background still wakes.
- A blur cancels pending passes, so focus churn cannot stack them.
- Never clears the shared glyph atlas. Wiping it on plain refocus re-arms xterm's page-merge race, which is exactly why `atlasWakeRecovery` excludes focus from its own triggers.
- Logs at `warn`, not `debug`: Verbose is dropped when DevTools is closed, and a reporter needs to be able to tell "the repaint fired and did not help" from "it never fired".

Windows only. macOS and Linux report visibility correctly and are already covered, so they take no extra GPU work.

## Honest limits

The gap this closes is proven and machine-independent. The GPU-side reason the pixels go missing is **not** — it does not reproduce here (20 s fully occluded, canvas intact), and the reporter sees two other rendering oddities on the same machine, so the drop looks driver-specific.

That is why this says `Refs #879` rather than `Closes #879`: the same call as #319/#318, where a rendering fix that could not be verified locally stayed open until the reporter confirmed. Follow-up questions are posted on the issue.

`atlasWakeRecovery.ts` also gains a comment recording the Windows measurement, so the next reader does not assume its `visibility` trigger covers Windows.

## Tests

47 new tests. The scheduler is tested against injected window/document/rAF/clock: no repaint without a preceding blur, both passes per wake, every visible pane and only visible panes, mount-while-blurred, blur cancels pending passes, a pass that runs unfocused repaints nothing, one pane throwing does not stop the others, unregister and teardown, and `enabled: false` attaching nothing. Source-level locks pin the wiring a scheduler test cannot reach: that `useTerminal` actually registers and unregisters, and that `AppLayout` gates on win32 — a coordinator with zero registered panes is a silent no-op, which is the failure mode of this bug.

`tsc --noEmit` clean. Full suite: no new failures.

Refs #879

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed blank terminal panes on Windows after switching applications and returning.
  * Visible panes now repaint automatically when the window regains focus.
  * Repainting is skipped when panes are hidden or the window loses focus again.

### DIFF
--- a/changelog.d/881.md
+++ b/changelog.d/881.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **A pane no longer comes back blank after you switch to another app and back
+  (Windows).** Every repaint-on-wake path in the app was keyed to the browser's
+  page-visibility signal, and on Windows that signal never changes: the window
+  reports itself visible while it is fully covered by another app, and even
+  while it is minimized. The only repaint that fired on alt-tab belonged to the
+  single pane owning keyboard focus, so a pane whose canvas pixels were dropped
+  by the GPU had nothing to redraw it and stayed empty until you forced a redraw
+  by hand (selecting text, resizing). Window refocus is now its own wake
+  boundary: every visible pane repaints, on the next frame and once more shortly
+  after, without touching the shared glyph atlas. Windows only — macOS and Linux
+  report visibility correctly and were already covered. (#879)
--- a/src/renderer/components/Layout/AppLayout.tsx
+++ b/src/renderer/components/Layout/AppLayout.tsx
@@ -25,6 +25,7 @@ import ApprovalDialog from '../Company/ApprovalDialog';
 import ExecuteApprovalDialog from '../A2a/ExecuteApprovalDialog';
 import PermissionApprovalDialogContainer from '../Approval/PermissionApprovalDialogContainer';
 import { initAtlasWakeRecovery } from '../../terminal/atlasWakeRecovery';
+import { windowWakeRepaint } from '../../terminal/windowWakeRepaint';
 import CompanyView from '../Company/CompanyView';
 import MessageFeedPanel from '../Company/MessageFeedPanel';
 import OnboardingOverlay from '../Onboarding/OnboardingOverlay';
@@ -516,6 +517,18 @@ export default function AppLayout() {
         typeof onResumed === 'function' ? onResumed : () => () => {},
     });
   }, []);
+
+  // #879 — window-level wake repaint, Windows only. On Windows
+  // `document.visibilityState` never reports hidden (measured: not when the
+  // window is covered, not even when it is minimized), so the wake recovery
+  // above and useTerminal's `docRevealed` reclaim can never fire there, and a
+  // pane that lost its canvas pixels while the window was away has nothing to
+  // repaint it. macOS and Linux do flip visibility, so they are already
+  // covered and get no extra GPU work. Rationale and measurements live in
+  // terminal/windowWakeRepaint.ts.
+  useEffect(() => windowWakeRepaint.init({
+    enabled: window.electronAPI?.platform === 'win32',
+  }), []);
 
   // #517 slice C — discard mode mirrors the same way. Effective only while
   // lightweight mode is also on (belt-and-braces: main enforces this too).
--- a/src/renderer/hooks/__tests__/useTerminal.windowWake.test.ts
+++ b/src/renderer/hooks/__tests__/useTerminal.windowWake.test.ts
@@ -1,0 +1,62 @@
+import { describe, it, expect } from 'vitest';
+import fs from 'node:fs';
+import path from 'node:path';
+
+// #879 wiring lock (source-level).
+//
+// windowWakeRepaint.test.ts proves the coordinator's behaviour against injected
+// deps. What it cannot prove is that anything is actually plugged into it: a
+// coordinator with zero registered panes is a silent no-op, and that is exactly
+// the failure mode of this bug (a repair path that exists but never fires).
+// Mounting useTerminal for real needs a WebGL context and the whole preload
+// bridge, so the wiring is pinned at the source level — the same approach
+// useTerminal.atlasClear.test.ts takes for the #191 invariant.
+
+const hookSrc = fs.readFileSync(path.join(__dirname, '..', 'useTerminal.ts'), 'utf-8');
+const layoutSrc = fs.readFileSync(
+  path.join(__dirname, '..', '..', 'components', 'Layout', 'AppLayout.tsx'),
+  'utf-8',
+);
+
+describe('#879 — every pane is registered with the window-wake coordinator', () => {
+  const start = hookSrc.indexOf('windowWakeRepaint.register({');
+  const registerBlock = hookSrc.slice(start, hookSrc.indexOf('});', start));
+
+  it('useTerminal registers each terminal', () => {
+    expect(start).toBeGreaterThan(-1);
+  });
+
+  it('reports pane visibility from isVisibleRef so hidden panes are skipped', () => {
+    // A hidden pane already repaints through glyphRepaint's `visible` reason
+    // when its workspace/tab comes back; waking it here would be a full-range
+    // refresh nobody can see.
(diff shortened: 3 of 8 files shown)


openwong2kim:

This PR closes the proven half of #879: on Windows `document.visibilityState` is a constant (visible even when the window is fully covered or minimized, verified against bare Electron 41 with and without `CalculateNativeWinOcclusion`), so every wake-driven repaint path that keys off it was dead code, and a pane that lost its canvas pixels during alt-tab stayed blank until a manual redraw was forced. The diff introduces a single app-wide coordinator (`windowWakeRepaint.ts`, following the `atlasGuard` singleton + `AppLayout` init pattern) that listens for blur/focus and repaints every *visible* pane on the next frame and once more 250 ms later, without ever clearing the shared glyph atlas (which would re-arm xterm's page-merge race from #191). Registration is unconditional in `useTerminal`; the coordinator itself is the win32 gate, so non-Windows builds do no extra GPU work.

The design is disciplined about its own limits — the two-pass structure is explicitly a hedge against unproven Chromium surface behavior, the focus re-check at execution time prevents painting a window nobody is looking at, and the PR correctly says `Refs #879` rather than `Closes #879` because the GPU-side reason the pixels go missing is driver-specific and not reproducible on every machine. The 47 new tests cover the scheduler against injected window/document/rAF/clock, and the source-level wiring locks in `useTerminal.windowWake.test.ts` are exactly the right response to the bug's failure mode: a coordinator with zero registered panes is a silent no-op, and the locks pin the wiring a behavioral test can't reach.

Priority **P1** — this is a direct, well-scoped fix for a user-visible rendering failure that shipped in the winget release (#879 reproduces on 3.38.6), with rigorous testing and an honest `Refs` rather than premature `Closes`. Worth merging once the owner is satisfied; the follow-up questions already posted on #879 are the right place to determine whether the second 250 ms pass can be dropped.

openwong2kim:
Closing this unmerged. The reporter closed #879 himself: the black pane is gone on 3.41.1, and winget had been holding him on 3.38.6. He also never filed the other two rendering issues he mentioned, so all three glyph symptoms went away with the same upgrade.

Nothing between v3.38.6 and v3.41.1 bumped Electron or xterm, and the only renderer commits in that range on this path are the atlas guard chain: #790 taught CURE to detect a merge by page identity instead of a count drop, and #798 made the guard drop each pane's render model when it wipes the pool. On 3.38.6 `clearTexture` was a permanent no-op (upstream's page-0 probe), so the guard could see corruption and do nothing about it. Three glyph symptoms dying at once on upgrade is what fixing that looks like.

So the cause was ours, it is already fixed, and this PR is a repaint for a symptom nobody has. The gap it closes is real — measured, on Windows nothing fires when the window is re-activated, and a pane whose canvas pixels are dropped never self-heals — but I never proved the mechanism that drops them, said so in the PR, and the review called it a workaround built on an analogy. That analogy has now run out.

Two more reasons not to carry it. This repo has already been here: #166 → #191 → #318 were repaint band-aids that hid the real defect for three releases, and adding one immediately after finally fixing the real defect would mask a recurrence instead of letting it be seen. And the cost is permanent — two full-range refreshes per visible pane per alt-tab, on every Windows user, forever.

What the work actually bought is in #883: `document.visibilityState` on Windows never reports hidden, not when the window is covered and not even when it is minimized. That is what exposed #882, where the same measurement fixes a bug that does reproduce.

PR #883, opened 2026-08-13
Report the pane hidden when the window is minimized, hidden to tray, or locked

### LINKED ISSUE #882: Windows: viewer visibility is never reported hidden, so a phone can never take PTY size ownership
Found while investigating #879, same root cause, different feature.

#766 gave the daemon a visibility-based size-ownership rule: while the desktop renderer reports a pane visible, a phone asking to reshape the PTY gets `409 desk-owns-size`; while it reports hidden, the phone is allowed to reshape it. The renderer computes that report in `useTerminal.ts` as `isVisible && docVisible`, where `docVisible` tracks `document.visibilityState`.

On Windows that second term is a constant. Measured against a bare Electron 41 build:

| window state | `document.visibilityState` | rAF |
|---|---|---|
| foreground | visible | 60/s |
| fully covered by another app | visible | 60/s |
| **minimized** | **visible** | 60/s |

`visibilitychange` fires exactly once per window, at teardown, and toggling Chromium's `CalculateNativeWinOcclusion` changes nothing.

So on Windows the renderer never reports a pane hidden on window state alone. Minimize wmux, pick the session up on your phone, and the phone still gets `409 desk-owns-size` — the handoff #766 exists for does not happen. Switching the pane's workspace or tab still flips `isVisible`, so the feature is not entirely dead, it just never responds to the window itself.

The same measurement is what #879 turned out to be, and #881 fixes that one by treating window focus as the wake signal instead of visibility. This one needs its own answer: whether a blurred-or-minimized window should count as "not viewing", and whether the signal comes from window focus in the renderer or from `BrowserWindow` state in main (`isMinimized()` / `browser-window-blur`), which is the more honest source for "is anyone looking at this window".

Not urgent, and not a regression — it has never worked on Windows.

openwong2kim:
## The bug

#766 made PTY geometry ownership visibility-based: while the desk says it is showing a pane, a phone resize gets `409 desk-owns-size`; while the desk says the pane is hidden, the phone's numbers are applied. The renderer built the window half of that answer out of `document.visibilityState`.

On Windows that value is a constant. Measured on Electron 41:

| window state | `document.visibilityState` | rAF |
|---|---|---|
| foreground | visible | 60/s |
| fully covered by another app | visible | 60/s |
| **minimized** | **visible** | 60/s |

`visibilitychange` fires once per window, at teardown; disabling Chromium's `CalculateNativeWinOcclusion` changes nothing.

So the window term contributed nothing, and minimising wmux never handed the size over — the handoff #766 exists for did not happen. Workspace and tab switches still flipped the pane term, so the feature was not dead; it just never responded to the window. Never worked on Windows, so not a regression.

## What this changes

Main is the only side that can see the window, so main owns the signal.

```
BrowserWindow  minimize | restore | hide | show
powerMonitor   lock-screen | suspend        → nobody is at the machine
               unlock-screen | resume       → re-read, never assume
        │
        ├─ isWindowDisplayed(win) = !destroyed && isVisible() && !isMinimized()
        ▼   dedup
  WINDOW_DISPLAYED_CHANGED (push) ──► renderer store ──► useTerminal
  WINDOW_IS_DISPLAYED      (pull)          │              │
                                           │              ▼
                                           │   decideViewerVisibility()
                                           │      ├─ report → daemon
                                           ▼      └─ refit on reveal
```

- **Lock screen counts.** Locking neither hides nor minimizes the window, so a window-events-only reporter would keep claiming the desk is in use while the user is away from the machine — which is exactly when the phone is the only screen left. `desktopPresence.ts` learned this for notifications; this mirrors it.
- **Plain blur does not count.** A blurred window is still on screen. Letting a phone reshape the PTY behind someone reading that pane on a second monitor is the geometry fight #766 set out to prevent, and on Windows there would be no reveal event to take it back either.
- **Pull at mount, not just push.** `did-finish-load` does not mean the renderer installed its listener, and this window reloads after a renderer crash. Same push + pull shape as the existing fullscreen state. The store subscribes before pulling and lets a push win, so an in-flight invoke cannot resurrect a stale answer.
- **The reveal is half the fix.** A restored window fires no ResizeObserver tick (the container never changed size), so without an explicit refit the desk would release the size on minimize and never take it back. Report and reveal come from one pure function, keyed on the window terms only so a workspace switch does not fit twice.
- **A daemon reattach replays the report.** The daemon starts sessions at `viewerVisible: true` and resets to true on detach, and a hidden pane's visibility does not change just because the daemon respawned.
- **One subscription for the app**, not one per pane, and `shouldPollMetadata` now shares the window predicate instead of repeating it — keeping its own `isLoading` term, because a loading renderer is a reason not to poll it for cosmetics, not a reason to tell the daemon nobody can see the window.

Occlusion — covered but not minimized — stays unhandled, because Chromium on Windows will not report it. Documented hole, not an oversight. Daemon side is unchanged.

The last commit is documentation only: `atlasWakeRecovery.ts` now records the same measurement, because its own `visibility` trigger is dead on Windows for exactly this reason and nothing in the file said so.

## Tests

24 new. The decision (report + reveal) is a pure function with its own behavioural tests, including the two cases that define the bug: a minimized window hides the pane even though `docVisible` stays true, and a restore that only the new term can see still refits. The main-side reporter is tested against a stub window: each of the four window events, lock/suspend/unlock/resume, no push when the answer did not change, never sending to a destroyed window or destroyed webContents, and `current()` answering the pull. The renderer store covers pull-at-init, push-wins-over-in-flight-pull, failed pull leaving the safe default, and teardown. Source locks pin the call sites the pure tests cannot reach — a pure function nobody calls is how #766 shipped a correct-looking expression whose window term was dead.

Two `a2a` rpc tests replaced `shared/constants` wholesale rather than spreading it, so they broke the moment anything in their import graph reached for another export. They spread it now.

`tsc --noEmit` clean. Full suite: no new failures.

Closes #882

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal session resizing when the desktop is minimized, hidden in the system tray, screen-locked, or restored.
  * Corrected terminal visibility tracking across window, document, and pane states.
  * Automatically refits active terminals when the application window becomes visible again.
* **Reliability**
  * Improved recovery after reconnecting to terminal sessions and system resume events.
* **Tests**
  * Added comprehensive coverage for window visibility, terminal refitting, event handling, and recovery scenarios.

### DIFF
--- a/changelog.d/883.md
+++ b/changelog.d/883.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **Minimise wmux on Windows and your phone can finally resize the session.**
+  Handing PTY geometry to a phone depends on the desktop admitting nobody is
+  looking at the pane, and the desktop worked that out from the browser's
+  page-visibility signal — which on Windows never changes, not when the window
+  is covered and not even when it is minimised. So the phone was told the desk
+  owns the size forever, and the terminal it showed stayed wrapped for a window
+  nobody could see. The desktop now takes that answer from the window itself:
+  minimised, hidden to the tray, or screen locked all count as nobody looking,
+  and restoring the window takes the geometry back. Locking your screen releases
+  the size too, which is exactly when the phone is the only screen left.
+  (#882, follow-up to #766)
--- a/src/main/ipc/handlers/metadata.handler.ts
+++ b/src/main/ipc/handlers/metadata.handler.ts
@@ -2,6 +2,7 @@ import { ipcMain, BrowserWindow } from 'electron';
 import fs from 'node:fs';
 import { IPC } from '../../../shared/constants';
 import type { MetadataUpdatePayload } from '../../../shared/types';
+import { isWindowDisplayed } from '../../window/windowDisplayed';
 import { MetadataCollector } from '../../metadata/MetadataCollector';
 import { prStatusCache } from '../../metadata/PrStatusCache';
 import { gitSyncStatusCache } from '../../metadata/GitSyncStatusCache';
@@ -61,8 +62,11 @@ export function broadcastMetadataUpdate(
 export function shouldPollMetadata(win: BrowserWindow): boolean {
   if (win.isDestroyed()) return false;
   if (win.webContents.isLoading()) return false;
-  if (!win.isVisible() || win.isMinimized()) return false;
-  return true;
+  // Shared with the #882 viewer-visibility report so the two definitions of
+  // "the window is on screen" cannot drift apart. The loading check above stays
+  // local: a loading renderer is a reason not to poll it for cosmetics, but not
+  // a reason to tell the daemon nobody can see the window.
+  return isWindowDisplayed(win);
 }
 
 const collector = new MetadataCollector();
--- a/src/main/ipc/registerHandlers.ts
+++ b/src/main/ipc/registerHandlers.ts
@@ -40,6 +40,7 @@ import { updateUnreadBadge } from '../tray';
 import { eventBus } from '../events/EventBus';
 import { WMUX_EVENT_TYPES, type WmuxEventType } from '../../shared/events';
 import { VALID_TRANSITIONS, type TaskState } from '../../shared/types';
+import { windowDisplayedReporter } from '../window/windowDisplayed';
 
 const EVENT_TYPE_SET = new Set<WmuxEventType>(WMUX_EVENT_TYPES);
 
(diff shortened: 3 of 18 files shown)


openwong2kim:

This PR fixes #882 by moving the "is the desktop window actually showing" signal out of the renderer (`document.visibilityState`, which is a constant `visible` on Windows even when minimized) and into the main process, which is the only side that can see real window state. It introduces `windowDisplayed.ts` in main, fed by `BrowserWindow` minimize/restore/hide/show events plus `powerMonitor` lock-screen/suspend/unlock/resume, and a renderer-side `windowDisplayedStore` (pull-at-init then push) that every pane folds into the existing #766 viewer-visibility report through a new pure decision function, `viewerVisibility.ts`.

The design is careful about the cases that made the original bug invisible: the decision is extracted as a pure function with behavioural tests for the two defining cases (a minimized window hides the pane even though `docVisible` stays true; a restore only the new term can see still refits); the reporter is tested against a stub window for every event including the no-op dedup and the destroyed-window guards; and source-level wiring locks pin that `useTerminal` actually consumes the store and `AppLayout` initializes it once — the failure mode of the original bug was a correct-looking expression nobody fed. Lock screen is deliberately included (locking neither hides nor minimizes the window, yet it is exactly when the phone is the only screen), while plain blur is deliberately excluded (a blurred window is still on screen, and letting the phone reshape behind it is the geometry fight #766 was written to prevent).

Priority is P2: it closes a genuine functional gap that breaks the #766 phone-resize handoff on Windows, but it is not a regression (the feature has never worked there), is non-urgent, and the companion repaint PR #881 is the higher-impact fix for users.

PR #885, opened 2026-08-13
chore(release): 3.42.0

openwong2kim:
Minor, not patch: this release carries a new locale and a behavior change, not only fixes.

## Contents

**Added** — Simplified Chinese is now a complete locale: all 1,203 keys, up from about a third (#876). The toolbar and Git labels that were still English inside it are translated too (#880).

**Changed** — Browser automation refuses a call whose workspace it cannot determine, instead of driving whichever page happened to register first. Callers that already send their workspace are unaffected; `"mcp": { "mode": "shadow" }` restores the old behavior (#873).

**Fixed**
- The CJK IME candidate window follows the painted cursor instead of drifting by the scrolled-up row count (#875).
- Daemon push events are limited to the desktop app without losing startup events (#878).
- On Windows a pane is finally reported hidden when the window is minimized, hidden to the tray, or the screen is locked, so a phone can take PTY size ownership the way #766 intended. `document.visibilityState` never reports any of those there, so that half of the check had never worked on Windows (#882).
- The real-Git reboot-survival demos moved to the serial test lane, so they stop timing out against 700 parallel files on a loaded Windows runner (#884).
- Plus #877.

## Release mechanics

The four things a release commit has to get right, all done:

- `node scripts/collect-changelog.mjs` folded the eight fragments and removed them; `[Unreleased]` was promoted to `[3.42.0] — 2026-08-14`.
- `package.json` and **both** `package-lock.json` sites bumped (root and `packages[""]`) — 1 + 2 replacements, verified by count.
- `node scripts/gen-api-reference.mjs` re-run, so `docs/api/reference.md`'s baked-in version header matches. The CI drift guard enforces this.

Twelve files, same shape as 3.41.1 (#872). No source changes.

After merge: push the `v3.42.0` tag to trigger the installer build, Chocolatey and WinGet.


openwong2kim:

This is the release commit for v3.42.0, folding eight changelog fragments into the top-level `CHANGELOG.md` (promoting `[Unreleased]` → `[3.42.0] — 2026-08-14`) and removing them, bumping the version in `package.json` and **both** `package-lock.json` sites (root and `packages[""]`), and regenerating `docs/api/reference.md` so its baked-in version header matches. Twelve files, +69/−69, no source changes — the same shape as the 3.41.1 release (#872).

The contents justify the minor bump: a completed Simplified Chinese locale (#876), a behavior change where browser automation now refuses workspace-less calls instead of driving an arbitrary page (#873/#810), and a run of Windows-specific fixes (IME candidate-window anchoring #875, daemon event delivery #877/#878, minimized-window visibility for phone PTY sizing #882, reboot-survival demo lane #884).

**Priority: P3** — routine release mechanics. The checklist items (both lockfile sites bumped and counted, CI drift guard satisfied, no source changes) are all in order; this is ready to merge and tag `v3.42.0` to kick off the installer, Chocolatey, and WinGet builds.

PR #887, opened 2026-08-14
fix(a2a): report reply delivery outcome instead of silently suppressing nudges

openwong2kim:
## Problem

The reply path of `a2a.task.send` withholds its PTY nudge behind four safety guards (`pin_lost` / `same_ws_no_anchor` / `self_loop` / `unverified_sender`) but still returned a bare success. A sending agent could not tell a delivered reply from a suppressed one, and the receiver got no signal of any kind (the bus tee only ever covered the create path).

Two 2026-08-13 dogfood sessions (3 agents in one workspace) hit exactly this: every nudge was withheld invisibly and a human ended up hand-relaying an agent-to-agent debate, message by message.

## Changes

- **`decideReplyDelivery`** (new, `a2aAddressing.ts`): the four reply suppression guards extracted verbatim into a pure, unit-tested function — same semantics, same precedence, but the reason is a value instead of a silent skip.
- **`delivery` response field** on both branches of `a2a.task.send`: `{stored, notified, mode? | reason?, hint?}`. Every suppressed path now reports why and what the sender can do about it (per-reason recovery hints).
- **Bus pointer on suppressed replies**: emits the `a2a.task` `'updated'` event so a receiver polling `wmux_events_poll` still learns the thread moved. Fires only on the suppressed branch (LLM-turn paced — no ring flood).
- **Round cap**: replies past 5 completed round trips (`countRoundTrips` = min of per-side message counts) are refused with a `cap_reached` error. Deliberately NOT a status transition: the reply path never consults task status, and `input-required→working` is agent-reachable, so a transition would neither stop a runaway two-agent loop nor stay stopped.
- **`a2a_discover` `elapsedMs`**: stamped at the MCP tool entry. A dogfood report attributed a ~2865s stall to this call; the server span is bounded by the 5s bridge timeout and the handler is a pure in-memory map, so the stall accrued client-side — now the numbers can prove that.

## Explicitly not in this PR

- Loosening any of the four guards (they are correct — the pull-model nudge is the prompt-injection defense line; what was broken was their silence).
- Anchor stamping at delivery time — review showed the reply path has no daemon mirror-update RPC (create-only), and stamping would trade active-pane fallback for fail-closed on pane death in cross-ws. Revisit with observability data.
- Per-message EventBus emission on the happy path (flood-guard decision stands).

## Tests

- `a2aAddressing.test.ts`: +14 cases — all four suppress reasons, deliver shapes (same-ws/cross-ws), hint actionability, round-trip arithmetic incl. the 4→5 cap boundary and double-post non-inflation.
- `useRpcBridge.a2aPaneIdentity.test.ts`: source invariant updated to pin the single decision point, the surfaced `delivery.reason`/hint, the suppressed-reply bus emit, and the pre-store cap check.
- Full A2A suites green (79 tests), `tsc --noEmit` clean, `gen-api-reference` no drift.

## Summary by CodeRabbit

- **New Features**
  - Successful JSON discovery responses now include elapsed time.
  - A2A messages report delivery outcomes, including stored, notified, suppressed, and missing-target states.
  - Added recovery hints when delivery is suppressed.
  - Added protection against self-replies, unverified senders, and missing workspace targets.
  - Limited A2A conversations to five completed reply round trips and per-side message limits.
  - Task-pointer updates are emitted when nudging is withheld.

- **Bug Fixes**
  - Improved cross-workspace reply routing and duplicate-message handling.
  - Task sends now report when notifications are nudged, suppressed, or unavailable.

### DIFF
--- a/changelog.d/887.md
+++ b/changelog.d/887.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- A2A replies no longer fail silently: `a2a_task_send` now returns a `delivery`
+  field (`stored` / `notified` / `reason` / `hint`) that says whether the other
+  party was actually nudged and, if not, why and what to do about it. A reply
+  whose nudge is withheld also emits the task pointer event, so a polling
+  receiver still learns the thread moved.
+
+### Added
+
+- A2A thread round cap: replies past 5 completed round trips are refused with a
+  `cap_reached` error so two agents cannot ping-pong unattended forever.
+- `a2a_discover` responses include `elapsedMs` (measured at the MCP tool entry)
+  to separate server latency from client-side stalls.
--- a/changelog.d/887.md
+++ b/changelog.d/887.md
@@ -8,7 +8,8 @@
 
 ### Added
 
-- A2A thread round cap: replies past 5 completed round trips are refused with a
+- A2A thread round cap: replies past 5 completed round trips — or past a per-side
+  message ceiling that catches one-sided monologues — are refused with a
   `cap_reached` error so two agents cannot ping-pong unattended forever.
 - `a2a_discover` responses include `elapsedMs` (measured at the MCP tool entry)
   to separate server latency from client-side stalls.
--- a/src/mcp/index.ts
+++ b/src/mcp/index.ts
@@ -1180,7 +1180,35 @@ server.tool(
   'a2a_discover',
   'List all available workspaces/agents and their names. ALWAYS call this first when the user references a workspace by number or name (e.g. "3번", "Workspace 1") so you know valid targets.',
   {},
-  async () => callRpc('a2a.discover'),
+  async () => {
+    // elapsedMs: measured at the MCP tool entry, i.e. the caller-visible round
+    // trip through pipe + main + renderer. A dogfood report blamed a ~2865s
+    // stall on this call; the server side is bounded by a 5s bridge timeout
+    // (_bridge.ts) and the handler is a pure in-memory map, so any large number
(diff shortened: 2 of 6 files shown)


openwong2kim:

This PR fixes a genuine observability gap in the A2A reply path that the 2026-08-13 dogfood sessions hit directly: `a2a.task.send` returned a bare success whether or not the PTY nudge was actually delivered, so two agents could believe they were conversing while every nudge was silently suppressed by the four safety guards (`pin_lost` / `same_ws_no_anchor` / `self_loop` / `unverified_sender`) — leaving a human to relay the entire debate by hand.

The extraction of `decideReplyDelivery` into a pure, unit-tested function is the right move: the guard semantics are preserved verbatim (so the same-ws self-paste defense line is intact), but the suppression reason becomes a value surfaced in the new `delivery` field (`stored` / `notified` / `reason` / `hint`) instead of a silent skip. The bus-pointer emission on the suppressed branch closes the last blind spot — a polling receiver now learns the thread moved even when the nudge is withheld. The round cap (5 completed round trips → `cap_reached` error) is a sensible guardrail against unattended two-agent ping-pong, and the `elapsedMs` instrumentation on `a2a_discover` turns the next stall report into something diagnosable. Test coverage is thorough (14 new cases covering all four suppress reasons, delivery shapes, hint actionability, and the round-trip arithmetic including the 4→5 boundary).

Assigned **P2** — this is an important correctness/UX fix for core A2A communication, validated by real dogfood impact and backed by careful design.

PR #888, opened 2026-08-14
fix(terminal): anchor IME compositions to the cursor's resting cell

### LINKED ISSUE #874: [IME] Chinese IME candidate window not following cursor in wmux terminal
## Problem

When using a Chinese IME (WeChat 微信输入法) in wmux on Windows, the input experience is severely broken. This affects all non-English IME users:

1. **While Claude Code is working** (output streaming), the cursor jumps/flickers wildly while typing Chinese, and the IME candidate window follows the flickering cursor, flashing all over the screen. English typing is fine because no candidate window appears.
2. **Even when idle** (no agent activity), the cursor does not stay put — it frequently jumps to other regions of the terminal, and the candidate window shows up misaligned (several rows away from the actual cursor position).

The candidate window does NOT follow the cursor's current line — it floats in a different area, disconnected from the editing context (see screenshot).

## Expected behavior

- The IME candidate window should stay anchored to the actual cursor position (like in Windows Terminal / VS Code).
- The cursor should not jump around while the agent is streaming output.

## Environment

- OS: Windows 11 Pro 10.0.26200
- wmux: latest (v3.41.0)
- IME: 微信输入法 (WeChat IME)
- Reproduced: both while Claude Code is running and idle

## Screenshot

The candidate window (red box) is vertically misaligned from the actual cursor position by ~3-4 lines, floating in a separate area instead of following the cursor.

## Impact

This makes wmux nearly unusable for Chinese IME users during active work. Please consider IME support for non-English input methods — thank you!

Image

### Proposed solution

Implement proper IME candidate-window positioning in the terminal renderer:

1. Keep the cursor position stable while the agent streams output — do not repaint/relocate the cursor on every chunk, only when the user actually types.
2. Report the real cursor cell position to the OS IME so the candidate window anchors to the actual editing line (Windows IME uses the console's cursor position via the IME-aware console APIs / the cursor style report).
3. Ensure cursor coordinates stay in sync with the visible terminal cell during redraws (avoid off-by-N-lines drift that currently places the candidate window 3-4 rows below the real cursor).

This is the same behavior Windows Terminal, ConPTY-based terminals, and VS Code already provide for CJK IME users.

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

openwong2kim:
Follow-up to #875 for #874, driven by cnxiekun's v3.42.0 field verification (https://github.com/openwong2kim/wmux/pull/875#issuecomment — log with 20 `[wmux:ime-anchor]` records + 6 screenshots).

## The third cause

The buffer cursor sampled at `compositionstart` is not always the input caret. Claude Code (Ink) repaints its lower region every frame; while a frame's escape stream is parsed, the cursor rides along with the writes and is only parked back on the caret at the end of the frame. Frames arrive split across PTY chunks, so a composition can start in a JS turn where the cursor sits mid-frame — and the #875 freeze then holds that transient stable for the whole composition.

The reporter's records show it plainly, with `ybase == ydisp` and `correction=(0,0)` in every one:

| Records | Reading |
|---|---|
| `(6,42) … (32,42)` | x advancing along row 42 — the real caret; these anchored correctly |
| `(53,49) (113,49) (113,44)` | cursor parked where the last write ended: status-line bottom-right when idle, the streaming row while working — matching the three misplacement screenshots one-to-one |

`correction=(0,0)` is consistent: the #875 correction fixes *where a cell is painted*, not *which cell is the caret*.

## The fix

Freeze the anchor to the cursor's **resting cell**. A tracker fed by `onCursorMove` records the last cell the cursor held for ≥ 32 ms (inter-chunk gaps inside one repaint are sub-ms to a few ms; inter-frame rest is ≥ ~50 ms — ~10x margin both ways). A composition starting mid-burst anchors to that resting cell; a cursor at rest is trusted as-is, keeping the #875 behavior identical in the common case.

- Tracker is seeded from the live cursor at attach (no pre-first-move hole) and resets on resize and alt-screen switches, where reflow changes what an absolute row means.
- The frozen anchor stays a pixel point exactly as in #875 — only the source-cell selection changed.
- The viewport/col clamp now lives in one place (`pointFromCell`; `paintedCursorPosition` delegates).

**Known limit, accepted deliberately:** dwell time cannot distinguish the caret from any cell the cursor parks on for ≥ 32 ms. The compositionstart diagnostic (retagged `[wmux:ime-anchor2]`) now reports `src=instant|resting` plus `held`/`restAge`/`sel`, so a field log discriminates that residual case before any further complexity is added. If misplacement persists with `src=instant` and a large `held`, the next iteration is revisit-frequency tracking — gated on that evidence, not shipped speculatively.

## Tests

36 in the module suite (all green; terminal + hooks suites 858/858, typecheck clean):

- Pure tracker state machine: burst never promotes, ≥ 32 ms promotes on leave, same-cell no-ops keep the dwell clock, reset drops the resting cell, selection semantics incl. exact `held`/`restAge`.
- Wiring with an injected clock: mid-burst composition anchors to the resting caret, not the parked cursor; at-rest composition is a byte-identical #875 regression lock (no transform pinned-at-bottom); resize and alt-screen resets; dispose unsubscribes the two new listeners.
- Existing #875 suite and the upstream tripwire untouched.

Not `Closes #874` — that stays open until the reporter confirms on a release build, same as the #875 round.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved CJK IME candidate-window placement by anchoring it to the cursor’s stable cell, preventing misplacement on status or streaming rows.
  - Improved behavior during cursor movement, screen resizing, buffer changes, alternate-screen transitions, and off-screen cursor positions.

- **Tests**
  - Added comprehensive coverage for IME anchoring, viewport scrolling, reset behavior, event handling, and resource cleanup.

- **Documentation**
  - Added a changelog entry describing the IME placement fix.

### DIFF
--- a/changelog.d/888.md
+++ b/changelog.d/888.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **The IME candidate window now anchors to where you are typing, not where the
+  agent's redraw parked the cursor.** The v3.42.0 fix pinned the candidate list
+  in place for a whole composition, but it pinned it to wherever the terminal
+  cursor happened to sit at the instant the composition started — and while a
+  TUI like Claude Code repaints, the cursor transiently rides along with the
+  redraw. Start typing Chinese, Japanese, or Korean in that instant and the
+  candidate list landed on the status line or the agent's streaming row instead
+  of the input box. wmux now tracks the cell the cursor actually rests on
+  between repaints and anchors compositions there, so the candidate list stays
+  on your caret even when a composition starts mid-redraw. (#874)
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1061,14 +1061,20 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
     // Remove the whole diagnostic once #874 is confirmed fixed in the field.
     let imeAnchorLogsLeft = 20;
     const imeAnchor = attachImeAnchor(terminal, {
-      onCompositionDiagnostic: ({ baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy }) => {
+      onCompositionDiagnostic: ({ baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, src, held, restAge, selY, selX }) => {
         if (imeAnchorLogsLeft <= 0) return;
         imeAnchorLogsLeft -= 1;
         // Mirrored into the main-side log file by src/main/index.ts's
-        // console-message listener, so the user can share it.
+        // console-message listener, so the user can share it. The "2" in the
+        // tag marks the resting-cell build so a shared log is unambiguous
+        // about which release produced it. src/held/restAge/sel are the
+        // cause-3 discriminator: src=resting means the composition started
+        // mid-repaint and the anchor used the last resting cell instead of
+        // the instantaneous cursor.
         console.info(
-          `[wmux:ime-anchor] pty=${ptyIdRef.current} compositionstart ybase=${baseY} ydisp=${viewportY} ` +
-          `cursor=(${cursorX},${cursorY}) cellHeight=${cellHeight.toFixed(2)} correction=(${dx.toFixed(1)},${dy.toFixed(1)})` +
+          `[wmux:ime-anchor2] pty=${ptyIdRef.current} compositionstart ybase=${baseY} ydisp=${viewportY} ` +
+          `cursor=(${cursorX},${cursorY}) sel=(${selX},${selY}) src=${src} held=${held.toFixed(0)}ms ` +
+          `restAge=${restAge.toFixed(0)}ms cellHeight=${cellHeight.toFixed(2)} correction=(${dx.toFixed(1)},${dy.toFixed(1)})` +
           (imeAnchorLogsLeft === 0 ? ' (last one, diagnostic capped)' : ''),
         );
       },
--- a/src/renderer/terminal/__tests__/imeAnchor.test.ts
+++ b/src/renderer/terminal/__tests__/imeAnchor.test.ts
@@ -482,6 +482,33 @@ describe('#874 resting-cell tracker (cause 3, pure)', () => {
     // Mid-burst right after the reset: no resting cell -> instantaneous.
     expect(selectFreezeCell(t, 12, 0, 1106).src).toBe('instant');
   });
+
+  it('rejects a resting cell that has scrolled off the viewport', () => {
+    // Scrolling output (a build log, not an in-place TUI repaint) leaves the
+    // resting cell above ydisp within a frame or two. pointFromCell would clamp
+    // it to row 0 and park the candidate window at the top of the terminal —
+    // worse than the live cursor, which is always on screen.
+    const t = createRestingTracker(1000, 8, 1000);
+    noteCursorMove(t, 3000, 0, 1100); // 1000,8 promoted; buffer has scrolled on
+    const viewport = { top: 2960, rows: 40 }; // rows 2960..2999 are visible
+    const sel = selectFreezeCell(t, 3000, 0, 1105, viewport);
+    expect(sel).toMatchObject({ absRow: 3000, col: 0, src: 'scrolled_out' });
+  });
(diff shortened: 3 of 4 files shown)


openwong2kim:

This PR is the third iteration on the IME candidate-window anchor bug (#874), addressing a root cause that survived the v3.42.0 fix in #875: when a TUI like Claude Code repaints its lower region every frame, the buffer cursor transiently sits mid-frame during parsing, and a composition that starts in that JS turn would freeze the candidate window to wherever the cursor was parked mid-redraw rather than on the caret. The fix introduces a resting-cell tracker (`RESTING_MS = 32`) that records the last cell the cursor held long enough to be the real caret, and `selectFreezeCell` picks that resting cell when a composition starts inside a move burst, while keeping the instantaneous-cursor behavior byte-identical to #875 when the cursor is genuinely at rest.

The implementation is clean and well-scoped: `pointFromCell` is extracted as the single viewport/column-clamping primitive with `paintedCursorPosition` delegating to it, the tracker is a small allocation-free state machine with `noteCursorMove` / `resetRestingTracker`, and resize and alt-screen switches correctly invalidate the resting cell since both reflow events change what an absolute row means. The diagnostic logging is retagged `[wmux:ime-anchor2]` and now emits `src=instant|resting` plus `held`/`restAge`/`sel`, so a field log can discriminate the accepted residual case (dwell time cannot distinguish the caret from any cell parked ≥32 ms) from a genuine miss — the honest, evidence-gated framing noted in the PR body.

Test coverage is thorough: 36 tests across the pure tracker state machine (burst-no-promote, ≥32 ms promotes on leave, same-cell no-op dwell-clock behavior, reset semantics, exact `held`/`restAge` arithmetic) and the wiring layer (mid-burst anchors to resting caret, at-rest regression lock against #875, resize and alt-screen resets, dispose unsubscribes the two new listeners), with the full terminal + hooks suites at 858/858 and `tsc` clean. Deliberately not `Closes #874` pending reporter confirmation on a release build, matching the #875 round — a sensible caution. P2: this is a real quality-of-input fix for CJK users with strong test discipline, not a release blocker.

**Priority: P2**

openwong2kim:
Dogfooded on a dev build (CDP-driven synthetic `compositionstart`, since the renderer does not answer `Input.imeSetComposition`) and found one regression this branch introduces, now fixed in 020aee72.

**What works as designed.** An in-place TUI repaint — cursor save → jump to the lower region → write → restore, no newline, so `ybase` never moves — is exactly the cause-3 shape, and the resting cell wins it:

```
cursor=(6,38) sel=(18,44) src=resting held=10ms restAge=80ms correction=(72.0,91.2)
cursor=(7,38) sel=(0,44)  src=resting held=1ms  restAge=31ms correction=(-50.4,91.2)
```

The cursor is on the streaming row the repaint is writing; the anchor goes back to the caret. That is the fix doing its job.

**What broke.** A resting cell is by definition a *past* cell, so output that *scrolls* the buffer (a build log, `1..20000 | %{ … }`) leaves it above `ydisp` within a frame or two. `pointFromCell` then clamps that row into view and parks the candidate window at the very top of the terminal:

```
ybase=2560 cursor=(35,44) sel=(0,-1017) src=resting correction=(-244.8,-668.8)
ybase=3665 cursor=(0,44)  sel=(0,-2122) src=resting correction=(0.0,-668.8)
```

10 of 12 compositions anchored off-screen. #875 could not produce this — it only ever read the live cursor, which is always on screen — so it is a regression rather than a residual.

**The fix.** `selectFreezeCell` now takes the live viewport and rejects a resting cell that has left it, falling back to the instantaneous cursor. The rejection is its own diagnostic source (`src=scrolled_out`) rather than folded into `instant`, so a field log still separates "the cursor was at rest" from "the resting cell was thrown away". Re-run on the patched build: off-screen anchors 0, every anchor on the caret row; the in-place repaint case above is unchanged. +4 tests (40 in the module suite), typecheck and eslint clean.

Also verified in the same session: the diagnostic renders with all the new fields, and the two new subscriptions (`onCursorMove`, `buffer.onBufferChange`) attach and dispose without error against the real xterm build.

PR #889, opened 2026-08-14
feat(deck): icon strip for the deck tabs, glyph rail when collapsed

openwong2kim:
## What

The deck header read `Agent (Default) ▾ | Git | Channels` — three text labels eating the whole header row of a 248–320px column. Each tab is a 36px glyph now, orca-style. The tab name (and the orchestrator's current model) moved into the tooltip and the accessible name.

Collapsing the deck used to leave no way back except a labeled row at the foot of the workspace sidebar: the opposite edge from the panel it opened, and gone entirely once the sidebar itself was collapsed — `MiniSidebar` never carried those rows. A 36px glyph rail now stands in for the collapsed deck **on the deck's own edge**, and pressing a glyph reopens the deck straight onto that tab.

```
open                                    collapsed
┌────────────┬──────────────┐          ┌───────────────────┬───┐
│ terminal   │ 🤖 ⑂ # ⊕   » │          │ terminal (wider)  │🤖 │
│            │ ─────────────│          │                   │⑂  │
│            │ Worktrees    │          │                   │#  │
│            │ …            │          │                   │⊕  │
└────────────┴──────────────┘          └───────────────────┴─«─┘
```

- The workspace sidebar drops the Agent / Git / Channels / web rows — **144px back to the workspace list**, and the entry points now live next to what they open.
- `WebToggle` is a single icon-only rendition shared by both strips; its sidebar variant and upward-opening popover anchor went with the rows.

## Review panel

Reviewed by a 3-model panel (Claude + GLM; Codex skipped — out of credits). Everything below was found by review and fixed in the second commit:

- **Channels entry point vanished** — the glyph was gated on `channelsTabVisible`, which ships `false` in both chrome presets, so an open deck could not reach channels at all once the sidebar row was gone. Pressing the glyph is the opt-in now, exactly as that row was.
- **a11y fallout from dropping the labels** — unread/dirty counts were badge-only (invisible to a screen reader) and now ride in the accessible name; `role="tablist"` no longer owns non-tab buttons; the web button's `aria-pressed` (which contradicted its own `haspopup`/`expanded`) is gone.
- **Popover could run off a short window** — it clamps downward and caps its height, so `Stop` stays reachable.
- **Expanding no longer enables the frozen Channels UI** behind the user's back when a stale persisted `activeDeckTab: 'channels'` meets `channelsTabVisible: false`.
- The rail marks the tab it will return to; its `ErrorBoundary` is width-capped; dead `toggleChannelDock` and six orphaned `sidebar.*` i18n keys removed.

One finding was **rejected**: "the 36px rail breaks the `minimal` preset's zero-side-chrome intent." That preset already keeps `MiniSidebar` (48px) with `sidebarVisible: false` — the rail is the same pattern, so there was no zero-chrome contract to break.

## Verification

- `tsc --noEmit` clean; eslint clean on every touched file
- **3212 renderer tests pass**, including 8 new ones: tab/rail switching, the Channels opt-in, the expand fallback, tablist containment, and count announcement
- Driven in a running dev instance over CDP — deck open (`commander:true`, glyphs + web present) and collapsed (rail renders, glyphs reopen the right tab)

DESIGN.md's layout contract and decision log record the move.

## Summary by CodeRabbit

* **New Features**
  * Added a collapsible icon-based deck rail for Agent, Git, Channels, and web access.
  * Added tooltips, accessibility labels, active-state indicators, and unread badges.
  * Added an expand control that restores the previously selected deck tab.

* **Bug Fixes**
  * Improved web popover positioning and scrolling on smaller windows.
  * Restored screen-reader announcements for unread and dirty-item counts.

* **Style**
  * Replaced labeled sidebar entries with compact icon-only controls.

* **Documentation**
  * Updated the layout documentation to reflect the new navigation structure.


openwong2kim:

This PR converts the deck header from three text labels (`Agent (Default) ▾ | Git | Channels`) into 36px icon glyphs and introduces a new `DeckMiniRail` component — a vertical glyph rail that lives on the deck's own edge when collapsed. Previously the only reopen affordance was a labeled row at the foot of the workspace sidebar (the opposite edge from the deck, and entirely gone once the sidebar itself collapsed), so this closes a real navigational dead-end. The second commit is substantive review-driven work: it restores the Channels entry point (the glyph had been hidden behind `channelsTabVisible`, which ships false, and the sidebar row that covered it was deleted in the same change), fixes `role="tablist"` containment so the web glyph and header controls are no longer announced as tabs, and moves unread/dirty counts into accessible names now that the badge is icon-decoration. The PR is clean — `tsc --noEmit`, eslint, and 3212 renderer tests pass including 8 new ones — and the DESIGN.md decision log is updated. A nice-to-have layout refinement with no correctness or security implications, so tagging it P3.

PR #891, opened 2026-08-14
Let the orchestrator brain fan out, and route each task by role

openwong2kim:
## What

The deck's orchestrator brain could not create isolated work. `fanout_start`
was absent from `COMMANDER_TOOL_SURFACE`, so it never registered in commander
mode and the brain's `tools/list` simply did not contain it; `task.fanout.start`
was likewise outside `COMMANDER_RPC_METHODS`. With `Bash` denied as well, there
was no path at all from an orchestrator to a worktree — its only worker was a
pane split into the shared checkout, where two workers editing at once
overwrite each other and share one branch.

This adds fan-out to the commander surface, gives every fan-out task an
optional role that decides which agent CLI and model it launches on, and tells
the brain when to reach for it.

## Why this does not widen the role gate

Fan-out is create-only, never teardown, and every dangerous input is already
server-derived rather than caller-stated — repository, owning workspace, agent
command — behind a human approval prompt that is never auto-approved. Adding it
widens what the brain may *ask for*, not what it can do unattended.

The blocker was identity, not policy. A brain is a subprocess with no pane
ancestry, so the PID-map walk always misses and it has no `senderPtyId` — the
very check that keeps the pipe surface honest refused it forever. It now proves
itself with the commander token `RpcRouter` already validated into
`ctx.commanderWorkspace`, and anchors repo derivation on its workspace's active
pane so it runs the same derivation every other caller does. A validated
commander binding outranks a stated `senderPtyId`, or a brain could aim a
fan-out at a workspace it is not bound to. The pty path stays fail-closed for
everyone else.

What this deliberately does *not* claim: that the brain cannot influence which
repository it gets. `pane_focus` is on the commander surface, so it can move the
active pane first and pick any repo its workspace already has a pane in. Nothing
escapes the workspace — the token is bound to exactly one — and inside it the
approval prompt, which prints the resolved path, is what holds the line.

## Per-task roles

`roles[]` is index-aligned with `titles` and closed to the `ORCH_ROLES`
vocabulary. The renderer resolves each role against the operator's own role
bindings, so one fan-out can put review work on a different agent — or a cheaper
model — than its build work, without the wire ever carrying a command string
(`agentCmd` is interpolated unquoted into a shell line downstream, which is why
the handler rejects it outright). An unknown role is rejected rather than
silently downgraded, the resolved agent/model rides into the approval preview,
and the role is stamped on the pane.

The GUI dialog gains the same per-task control.

## Review

Reviewed by a two-model panel (Codex was unavailable — out of credits). It found
that the central claim was false as first written: `applyRoleBinding` refuses to
swap launchers on a stem mismatch, so a `Reviewer→codex` binding changed nothing
and never reached model injection either. That refusal is right for the line a
human types at a live pane; a fan-out command is assembled by wmux, so the swap
is a separate function used only on that path. Also fixed from the same review:
role/title index misalignment that delivered another task's agent, a stale-closure
that dropped the dialog's role selection, an approval preview that named roles
without what they resolve to, a re-fire that replayed the pre-binding command,
an unchecked role on the IPC path, silent truncation of extra roles, and a
shared idempotency key space between brains and pane agents.

## Verification

- `npm run test:parallel` — 10,650 passing. The one failure,
  `bridgePermissionGate` `ETIMEDOUT`, is a known local-daemon-dependent flake
  and reproduces on `main`.
- `tsc` across all four project configs, clean.
- `npm run probe:mcp` — the golden wire probe passes; commander profile goes
  38 → 39 tools with `fanout_start` present, at 46,835 / 47,000 payload bytes.
  Descriptions were trimmed to earn that budget back rather than raising it.
- New tests: the launcher-swap transform, role pairing across dropped titles,
  re-fire material, commander identity and key-space isolation, roles-are-not-
  executables, and the renderer wiring guards.

## Summary by CodeRabbit

- **New Features**
  - Added a persistent UI scale setting from 80% to 160%, applied live across the interface without restarting.
  - Added selectable roles for fan-out tasks, with approval previews showing resolved agents, models, and launch details.
  - Fan-out tasks now support isolated parallel workspaces, role-aware execution, active-pane anchoring, and reliable command replay.
  - Added localized UI-scale and fan-out role labels across supported languages.
- **Tests**
  - Expanded coverage for scaling, role validation, workspace handling, approvals, and task replay.


openwong2kim:

Feature PR that puts fan-out within reach of the orchestrator brain and adds per-task role routing across ~1,267 added lines in 55 files (commander surface, fan-out RPC handler, FanOutService, brain skills, renderer dialog, and all locale files). Before this, `fanout_start` was absent from `COMMANDER_TOOL_SURFACE` and `task.fanout.start` from `COMMANDER_RPC_METHODS`, and with `Bash` denied the brain had no path at all to an isolated worktree — its only worker was a pane split into the shared checkout, where two workers overwrite each other's edits and share one branch. The trust argument is the part worth reading closely: fan-out is create-only (never teardown), the dangerous inputs (repository, owning workspace, agent command) are server-derived rather than caller-stated, and every spawn still passes a human approval prompt that is never auto-approved. The identity problem — a brain is a subprocess with no pane ancestry, so the PID-map walk and `senderPtyId` check always miss — is solved by having the brain prove itself with the commander token already validated into `ctx.commanderWorkspace`, with a validated binding outranking a stated `senderPtyId`, and the pty path staying fail-closed for everyone else. The PR is honest about its own limits (the brain can still influence which repo via `pane_focus`, but nothing escapes the workspace binding and the approval prompt prints the resolved path), the two-model panel review visibly changed the code (launcher-swap as a separate function from `applyRoleBinding`, role/title index alignment, idempotency key-space isolation), and verification covers `tsc` across four configs, 10,650 passing tests with one known local-daemon flake, and the golden MCP wire probe going 38 → 39 commander tools within payload budget. Priority: P2 — a capability unlock rather than a defect, but it widens the commander surface and deserves careful reviewer attention on the identity derivation in `fanout.rpc.ts` before merge.

PR #892, opened 2026-08-14
docs(deck): English decision row, symmetric locale keys, precise changelog

openwong2kim:
Follow-ups from the review on #889 — three findings, all confirmed against the code.

- **DESIGN.md's new decision row was Korean.** `CLAUDE.md` is explicit: repo artifacts are English, always. Rewritten in English; the existing Korean rows are left alone per the same rule.
- **Locale keys landed asymmetric.** `ko` was missing `web.running`, and `zh` was missing `deck.expandDock` while already carrying `deck.collapseDock`. Both icon-only controls (the web glyph, the rail's expand chevron) get their accessible names localized in all three locales — with the text labels gone, that name is the only thing a screen reader has.
- **Changelog wording was imprecise.** It said "right-hand deck", but the deck renders on the edge opposite the workspace sidebar and follows `sidebarPosition`; it also read as though all four glyphs select a deck tab, when `web` opens a popover.

No behavior change. `tsc --noEmit` clean, i18n + Deck suites pass (266 tests).

## Summary by CodeRabbit

- **New Features**
  - Deck tabs now use icons with tooltips for clearer navigation.
  - Agent, Git, Channels, and web controls are accessible from the deck strip.
  - Collapsed decks retain an icon rail for reopening specific tabs.
- **Localization**
  - Added Korean text for web activity status.
  - Added Chinese text for expanding the dock.
- **Documentation**
  - Updated design documentation to reflect the revised navigation layout and accessible labeling.


openwong2kim:

Documentation and localization cleanup following the deck-icon work in #889, with no behavior change. It rewrites the one new Korean decision row in DESIGN.md into English (repo artifacts stay English per CLAUDE.md, while the pre-existing Korean rows are deliberately left alone), restores symmetry to the locale keys — `ko` was missing `web.running` and `zh` was missing `deck.expandDock` while already carrying `deck.collapseDock` — and tightens the changelog so it no longer says "right-hand deck" (the deck renders opposite the workspace sidebar and follows `sidebarPosition`) and no longer reads as though all four glyphs select a deck tab when `web` opens a popover. The accessibility angle is the part that matters most: the web glyph and the rail's expand chevron are icon-only controls, so the localized accessible name is the only thing a screen reader has, and this change makes sure all three locales carry it. Verification notes `tsc --noEmit` clean with the i18n and Deck suites green (266 tests). Priority: P3 — correct and worth merging alongside the feature it documents, but no runtime impact.

PR #893, opened 2026-08-14
feat(web): scroll the browser terminal by touch

### LINKED ISSUE #890: [mobile] Better touch scrolling for the wmux web terminal (suggestion)
### Problem / motivation

First off — **huge praise for the wmux web feature!** I just started using it today to remotely control my work PC's terminal from my phone (via a virtual LAN), and it's honestly amazing. Being able to see and type into my Claude Code sessions from a phone browser, cross-network, is a killer feature. Really well done.

---

This is a UX suggestion from a mobile user, and if the current behavior is intentional, that's completely fine.

## Suggestion / Possible improvement

First, thank you for the wmux web feature — it's genuinely great for remote access. This is a UX suggestion from a mobile user, and if the current behavior is intentional, that's completely fine.

When I use wmux web from my **phone** to review long Claude Code session history, scrolling through terminal scrollback is difficult:

1. **Touch swipe doesn't scroll** — swiping up/down over the terminal area doesn't scroll the content the way a normal webpage does.
2. **The scrollbar is thin and hard to grab** — the custom thin scrollbar is hard to hit accurately on a touchscreen.
3. **Can't drag continuously** — the scrollbar only moves in discrete jumps; continuous smooth dragging isn't possible.

It would be really nice if:
- Swiping on the terminal scrolled the xterm viewport naturally (like a normal mobile browser page).
- The scrollbar were more touch-friendly (wider hit target), or native touch scroll worked.
- Continuous smooth scrolling were possible.

This would make reviewing long outputs (e.g. Claude Code transcripts) on mobile much smoother.

## Environment
- Host: Windows 11 Pro 10.0.26200
- Client: Android phone browser, via `wmux web` over a virtual LAN
- wmux: 3.42.0
- Terminal uses xterm.js with `scrollback: 5000`

Happy to help test or provide more detail if helpful. Thanks for considering!

---

P.S. Just to set expectations: I mainly use wmux web from the phone as a quick emergency fallback when I'm away from the desk — mobile usage is occasional, not my primary workflow. So if the current touch behavior is a deliberate design choice and not worth changing, that's totally fine too — please don't feel obligated. Just thought I'd share the feedback in case it's an easy win.

### Proposed solution

Ideally, the wmux web terminal on mobile would support natural touch scrolling of the xterm viewport (swipe up/down = scroll, like a normal mobile page). If that's straightforward, making the scrollbar hit-target larger or allowing continuous drag would help too.

But again — this is a nice-to-have, not a must. If it conflicts with the intended design (e.g. gestures are reserved for terminal input), feel free to keep the current behavior. Mobile usage is only my occasional emergency fallback. Thanks for considering!

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

openwong2kim:
Reported in #890 by @cnxiekun, who framed it as mobile polish and said that if the current behaviour were deliberate, that would be fine. Measuring it says otherwise: **there is no working way to reach scrollback from a phone at all.**

## What was measured

Chrome at a 390×780 phone viewport with touch enabled, driving real CDP touch input against a live `wmux web` server, 300 lines of scrollback in a plain PowerShell pane:

| Affordance | Before |
|---|---|
| Touch swipe over the terminal | **no movement at all** (view stayed on the same line for `dy=340` and `dy=200`) |
| Touch-drag the vertical scrollbar slider | **no movement** |
| Vertical scrollbar geometry | bar 14px wide, slider 20px tall |
| Real mouse wheel | works, but **1 line per −120 notch** |
| `Shift+PageUp` / `PageDown` | works correctly |

A phone has no wheel and no `Shift+PageUp` on its soft keyboard, so the only history it could ever see was whatever the opening snapshot painted. The wheel being 1 line/notch is a desktop-browser problem too — 300 lines of history is 300 notches.

## What this does

**Touch scrolling** — `attachTouchScroll(term, host)` in its own frontend file (inlined by the build like `attentionFormat.js` / `pairQuery.js`, so its arithmetic is unit-tested against the exact bytes the phone runs). Called from BOTH terminal-creation sites; a source invariant asserts they name the same function, because a second copy of this wiring is exactly the drift that guard exists for.

- Sub-cell deltas accumulate with the remainder carried. Every touchmove on a high-DPR phone is a few pixels, so dropping it would mean a slow drag scrolls nothing, forever.
- The gesture locks to an axis, stickily, and ties go horizontal — `#stage`'s pan is an affordance that works today and stealing it is the worse failure.
- `preventDefault()` only on paths actually handled, and the handler stands down if something closer to the target already claimed the gesture.
- On the alternate screen there is no scrollback, so the swipe becomes arrow keys and the TUI scrolls itself (what xterm.js#1007 asks for). With input off that corner can do neither and says so once, rather than looking like a hang — without offering to enable input, which is the operator's posture to hold.
- A swipe the buffer cannot answer (already at the top or bottom) is not claimed at all, so a tall pane can still be panned in zoom mode.

**Wheel** — `scrollSensitivity` and `smoothScrollDuration` on the Terminal. xterm's own options, not a hand-rolled `wheel` listener that would double-scroll the day xterm changes.

**Scrollbar CSS** — the rules at `styles.css:200-208` targeted `.xterm-viewport::-webkit-scrollbar`, dead since xterm 6 replaced native viewport scrolling with a VS Code overlay element. Retargeted onto the element that actually draws. Geometry untouched.

## Why a downstream handler

xterm exposes **no touch API**: zero `touch` matches in `xterm.d.ts`, and the VS Code `Gesture` helper xterm 6 vendors sits behind an internal `Gesture.addTarget()`. Upstream agrees the gap is real (xterm.js#5377, #1007). Momentum is deliberately absent — xterm cannot do it structurally (xterm.js#594, "the viewport is actually underneath the row divs"), so it would mean hand-rolling velocity tracking and a deceleration loop, which is the most bug-prone part of any touch handler.

## Verification

Live, on a phone viewport with CDP touch input — every row below was a failing baseline before this branch:

| Leg | Result |
|---|---|
| Swipe into history and back | moves, and returns to exactly the starting line |
| Wheel −120 | 1 line → **4 lines** |
| Horizontal swipe | not claimed — `#stage` pan survives |
| At the bottom, swipe up | **not claimed** (0/14 touchmoves) — no dead zone |
| At the top, swipe down | **not claimed** (0/14) |
| `less` (alt screen) | scrolls |
| `vim` (alt screen) | scrolls |

56 unit tests (pure delta/axis arithmetic, jsdom wiring, source invariants), `tsc --noEmit` clean, eslint clean, and the CSP gate in `scripts/build-daemon-web.mjs` passes with the new inlined block.

Two bugs were caught by verification rather than by review, which is worth recording:

- **Arrow-key encoding.** The first alt-screen implementation sent raw CSI (`\x1b[A`). `less` claimed every touchmove and still did not move, because a TUI that turns on keypad-transmit (DECCKM) expects the SS3 form. The encoding now follows `term.modes.applicationCursorKeysMode`, read live at send time — an app can flip DECCKM without the handler being re-attached.
- **Clamped-edge dead zone.** The guard originally asked whether scrollback *exists* (`baseY > 0`) rather than whether it can move *in this direction*. At either end the gesture was claimed and nothing happened — reproducing, at the edges, the exact silent-no-op failure this change exists to remove. It now reads `viewportY` per direction and declines when it cannot answer.

## Not in scope

- **Momentum** — structurally unsupported upstream (above).
- **Widening the scrollbar hit target** — once a swipe works, the 14px bar stops being the only way in, and widening costs readable width on a 390px screen. Recorded in `TODOS.md` with the measurements and the condition to revisit.
- **Unicode addon on the web terminal** (`TODOS.md`) — a real mobile-web complaint, but a rendering-width bug with no shared code path with scrolling.

Not `Closes #890` — that stays open until the reporter confirms on a release build, same as the #874 rounds.


openwong2kim:

**Priority: P2** — important, should ship soon.

This PR makes the browser terminal's scrollback reachable again: it adds a touch-scroll handler for the wmux web terminal so a phone swipe scrolls the xterm viewport, translates swipes into arrow keys on the alternate screen (DECCKM-aware, so `less` and `vim` in keypad-transmit mode actually move), and repairs two regressions from the xterm 6 upgrade — the wheel moving exactly one line per notch in every desktop browser, and the scrollbar theming that had been silently dead since scrolling moved off `.xterm-viewport`. The implementation is a new inlined `touchScroll.js` (sub-cell remainder accumulation, sticky axis lock that preserves the `#stage` pan, direction-aware clamped-edge handling so zoom-mode panning survives), wired at both terminal-creation sites in `app.js`, with the CSP gate script count raised 4 → 5 in `build-daemon-web.mjs` and the dead CSS retargeted onto `.xterm-scrollable-element > .scrollbar.vertical`. The verification is thorough — 56 unit tests evaluating the exact bytes the phone runs, source invariants guarding against wiring drift between the 1-up terminal and split tiles, and live CDP touch-input runs that caught two real bugs (the SS3/CSI arrow encoding and the edge dead zone) — and the scope cuts are the right calls, with momentum structurally unsupported upstream and scrollbar widening deferred to `TODOS.md` behind explicit re-evaluation criteria. The measurement in the description also corrects the framing of #890: this was not difficult-but-possible mobile polish but a complete inability to reach scrollback from a phone, which is what justifies P2 here against the P3 the original report carried. Keeping #890 open until the reporter confirms on a release build is the right closing discipline.

cnxiekun:
Thanks! I'll verify on #890 once it ships in a release.

openwong2kim:
Ok thank u

PR #899, opened 2026-08-15
Stop the permission gate from re-prompting when wmux has no opinion

### LINKED ISSUE #898: [Bug]: wmux plugin breaks claude's flow
### What happened?

After official wmux plugin for claude is installed, from then on claude asks permissions for every single basic operation, even for "Read", even in the "bypass permissions on" mode which is really annoying and completely breaks the flow.
Looks like the issue might be living in the PreHook, but that's a guess, I don't really know.
Issue persists even after relaunch of the app.
Switching modes during the run itself doesn't help.

Claude said that it is deliberate choice for mobile integration or smth (sorry convo got wiped) and it got fixed by setting some approval flag to false in settings.json, but it just doesn't look like an expected behaviour.

### Steps to reproduce

1. Install official plugin
2. Restart the app
3. Turn on "bypass permission on" mode
4. Ask claude to do basically anything
5. Now you have to approve every atomic step, just as Read, Write, etc

### Expected behavior

1. Install official plugin
2. Restart the app
3. Turn on "bypass permission on" mode
4. Work as expected without having to manually approve each step

### wmux version

3.41.1

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

Windows 10 19045.6466

Claude Code

```shell

```

### Confirmation

openwong2kim:
An installed plugin made Claude Code ask for permission on every tool call — a plain `Read` included — in a session launched with `bypassPermissions`. Restarting did not clear it, changing the permission mode did not clear it, and `WMUX_GATE=0` did not clear it either, which is what made it look unfixable from outside.

## What was wrong

The gate answered `permissionDecision: 'ask'` on every path where wmux had nothing to say: gate switched off, headless, running outside a wmux pane, daemon unreachable, tool not gated at all, broker self-deferred after 120s. Each of those means "no opinion, use the normal flow", and `ask` was believed to encode exactly that. The comment said so in two places.

It does not. `ask` is an instruction to raise a prompt, and it overrides the session's permission mode.

Measured on Claude Code 2.1.233, one probe hook on the same wide matcher, under `--permission-mode bypassPermissions`:

| hook writes | `permission_denials` |
|---|---|
| `permissionDecision: "ask"` | `[Read, Bash]` |
| nothing (exit 0) | `[]` |

Same hook, same settings, same mode; the only difference was the JSON. Re-run against the real pre-fix `wmux-bridge.mjs` rather than a probe, it denied `Read`, `PowerShell`, `Bash` and `Glob` and hit max-turns without ever reading the file. With this branch, the same run returns the file contents and denies nothing.

The hook contract states the neutral form directly: exit code 0 with no output means the hook has no decision to report, so the call continues through the normal permission flow.

## Two things worth calling out

The escape hatch did not escape. `WMUX_GATE=0` also emitted `ask`, so turning the gate off left the prompting exactly where it was — the reporter's "changing the mode does not help".

`ask` was not only coming from the bridge's fast exits. The daemon answered `ask` for any tool the gate does not cover, which is the path a plain `Read` takes, and `HookIngest` already had a deliberate carve-out that refuses to gate a `bypassPermissions` session — the response then said `ask` anyway and prompted it regardless. That carve-out now actually takes effect.

## The fix

The bridge writes a decision only when there is a real one, and the daemon omits the field rather than filling it with `ask`.

The reason the gate never answers `allow` is unchanged and still right: the matcher sees every tool, so a blanket approval would override Claude Code's own prompts and the user's `settings.json` deny rules. Only a real remote approval says `allow`.

## Tests

The existing tests asserted `permissionDecision === 'ask'` on the fall-through paths, so they were holding the bug in place. They now assert that stdout is empty, which is the property that actually matters.

Closes #898

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored Claude Code’s native permission prompts when wmux cannot provide a decision.
  * Explicit approvals and denials continue to be honored without unnecessary prompts.

* **New Features**
  * Added detection for outdated or broken Claude Code plugin hooks.
  * Displays a persistent warning with a one-click command-copy action to update affected plugins.

* **Documentation**
  * Documented plugin update steps and automatic refresh behavior for hook installations.


openwong2kim:

This PR fixes the permission gate so it only ever speaks when it has a real verdict: `integrations/claude/bin/wmux-bridge.mjs` now writes nothing to stdout unless the daemon returned `allow` or `deny`, and the daemon (`src/daemon/index.ts`) omits `permissionDecision` instead of filling it with `ask` on every no-opinion path — gate disabled, headless, outside a wmux pane, daemon unreachable, non-gated tool, or a broker that self-deferred after 120s. The root cause is that `ask` was treated as "I have no opinion" when it is actually an active instruction that raises a prompt and overrides the session's permission mode, which is why an installed plugin re-asked permission for even a `Read` inside a `bypassPermissions` session (#898), and why neither restarting, switching modes, nor `WMUX_GATE=0` could stop it — the escape hatch emitted `ask` too. The measurement in the description (same probe hook, only the JSON changed: `ask` → denials on Read/Bash, empty output → none) is exactly the evidence that pins the semantics, and the rewritten tests asserting on empty stdout rather than a decision value are what prevents the old assertions from holding the bug in place again. Marking P2 rather than P1 — it badly breaks the plugin UX for affected users and closes a reported bug, but the fix is a behavioral correction on the fall-through paths, not a security or data-loss issue.

PR #902, opened 2026-08-15
fix(terminal): keep shared WebGL glyph atlas coherent

openwong2kim:
## Summary

Same-font terminals share one WebGL glyph atlas. Once that atlas has to reclaim space it sets a "rebuild your render model" flag that nothing ever clears, so from then on every pane sharing it rebuilds its whole model on every frame — a long CJK or TUI session leaves its panes permanently slower for the rest of the run. Two defects in the reclaim itself go with it: `clearTexture` gives up on page 0's cursor alone, so an atlas whose first page is idle while the rest are full clears nothing at all; and `_createNewPage` can append past the shader's sampler budget. This PR fixes the atlas so a wipe is total, a page is never allocated past that budget, and each renderer rebuilds once per invalidation instead of once per frame. The existing `atlasGuard` poll stays as a backstop and should almost never have to fire.

> **Corrected premise.** An earlier version of this description said sibling panes keep stale texture coordinates and paint wrong syllables or a black pane. That mechanism is not reachable on 0.19.0: `_requestClearModel` has four references in the whole addon and none of them reset it, so `beginFrame` only reads a flag that stays true and no sharing renderer can miss the signal. The cost of atlas saturation is the permanent per-frame rebuild above, not corruption. See the review thread for the source walk.

## Changes

- Patch `@xterm/addon-webgl@0.19.0` via `patch-package` (postinstall still runs `fix-node-pty.js` first).
- Local invariant model + tests for total wipe, page-budget, and per-observer generation.
- `atlasGuard` skips speculative PREVENT rebuilds when the atlas already exposes `clearModelGeneration`; broken-addon nudge path remains for unpatched addons.

### Fixed

- **A long CJK or TUI session no longer leaves its panes permanently slower.** Terminals sharing a font share one WebGL glyph atlas. Once that atlas had to reclaim space it set a "rebuild your render model" flag that nothing ever cleared, so from then on every pane sharing it rebuilt its entire model on every frame, for the rest of the run. Each pane now rebuilds once per reclaim instead of once per frame.

  Two defects in the reclaim itself go with it: clearing the atlas gave up as soon as its first page looked untouched, so an atlas whose first page happened to be idle while the rest were full cleared nothing at all; and it could hand the shader more pages than the shader is able to sample. The safety net that has been covering for this still runs, and should now almost never have to.

## Stability tier impact

n/a

## Test plan

- Unit: `atlasCoherence.test.ts`, `atlasGuard.test.ts` (including a 60s coherent-atlas no-storm case), plus existing `atlasWakeRecovery` / `glyphRepaint` / `webglContextPool` / `webglTeardown` / `useTerminal.atlasClear` / `useTerminal.webglTeardown` — 8 files, 86 passing in this worktree.
- Full `npm test` in this worktree: 11176 passed. One pre-existing failure also fails on `main` (`integrations/claude/__tests__/bridgePermissionGate.test.ts` — `spawnSync` ETIMEDOUT on the interactive-pane daemon path). Not introduced here.
- Manual (not run in this environment — no interactive Electron dogfood): two panes, same theme/font, flood Hangul or open two dense TUIs until the atlas would have hit 16 pages; switch focus — neither pane may show static/wrong syllables; alt-tab away and back should stay clean via wake recovery plus a total wipe.

The user-facing claim no longer rests on that dogfood — it is the per-frame rebuild regression, which the tests cover. The two-pane Hangul run is still worth doing, but it is not the merge gate it would have been for a paint claim.

## Related issues / PRs

Related: #166, #191, #318, #879

## Reviewer checklist

## New concepts

**`patch-package` for a third-party runtime fix.** This repo did not previously patch installed npm packages. The glyph-atlas defects live inside `@xterm/addon-webgl@0.19.0`; bumping xterm would pull IME/serialize changes we do not want in this PR. `patch-package` records a deterministic patch next to the exact installed version and reapplies it on `npm ci`.

Do not use it to vendor a rewritten xterm, or to carry a patch that cannot apply to the locked version. If the addon version moves, the patch must be regenerated or the upgrade belongs in its own PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved terminal WebGL rendering stability after glyph-atlas reclamation.
  * Prevented incomplete atlas clearing and excessive page allocation.
  * Reduced unnecessary per-frame pane model rebuilds and redundant recovery actions.
  * Preserved compatibility with both patched and legacy rendering components.

* **Chores**
  * Added an automatic installation step to apply required rendering fixes.
  * Locked the WebGL terminal add-on to a stable version.

* **Tests**
  * Added coverage for atlas clearing, allocation limits, cache resets, and generation tracking.


openwong2kim:

**Priority: P2** — rendering-integrity fix for the long-standing glyph-corruption family (#166, #191, #318, #879).

This PR fixes the shared WebGL glyph atlas at the source by patching `@xterm/addon-webgl@0.19.0` via `patch-package` (a first for this repo) instead of working around it from `atlasGuard`. The patch enforces four invariants: `clearTexture()` becomes total (probes every page, collapses to constructor shape, empties both glyph caches — the page-0-only probe was the 0.19 bug), page allocation never exceeds the shader's sampler budget (evict-all rather than a page the shader cannot sample), each `GlyphRenderer` observes wipes/merges independently via a monotonic `clearModelGeneration` instead of the shared one-shot boolean, and the renderer clamps texture binding to the allocated units. On our side, `atlasGuard.ts` correctly demotes itself to a backstop: it duck-types a coherent atlas via `clearModelGeneration` and skips the speculative PREVENT rebuilds that would otherwise re-arm every cooldown and fight a self-evicting atlas — the new 60s no-storm test pins that behavior down. The invariant model lives in wmux-owned test code (`atlasCoherenceModel.ts`), which is the right place to state the algorithm before patching a dependency.

Why it matters: Korean-heavy and long TUI sessions mint glyphs continuously and are the realistic way to exhaust the 16-page budget, and wrong syllables or a black pane reads to the user as data corruption even though it is "only" paint. Two things before merge: the manual dogfood the PR itself calls for (two same-font panes flooded to the 16-page cap, focus switches, alt-tab wake) has not been run in this environment, so please do not merge on the user-facing paint claim until that has been done; and when `@xterm/addon-webgl` eventually moves off 0.19.0, this patch must be regenerated or dropped in its own PR.

src/renderer/terminal/atlasGuard.ts
```diff
@@ -508,7 +518,11 @@ export function createAtlasGuard(options: AtlasGuardOptions = {}): AtlasGuard {
       const removed = latch?.fired ?? false;
       if (latch) latch.fired = false;
       const mergeSignal: MergeSignal | null = removed ? 'page-removed' : detectMerge(prev, tags);
-      const nearTrigger = len >= preventAt && used >= preventAt;
+      // Patched atlas (I2): self-evicts at the sampler budget. Duck-typed by
+      // optional clearModelGeneration. PREVENT on that shape is a storm, not
+      // a repair. Unpatched addons omit the field and still get the backstop.
+      const coherent = typeof atlas.clearModelGeneration === 'number';
+      const nearTrigger = !coherent && len >= preventAt && used >= preventAt;
```
openwong2kim, line 536: This gates PREVENT only. The patched atlas's own eviction path still reaches CURE below, and CURE has no cooldown — `_wipeToConstructorShape` collapsing 16 pages to 1 reads as `count-drop`/`page-identity`, and `_tryMergeSameSizePages` -> `_mergePages` fires the `page-removed` latch. So a coherent atlas gets a full group rebuild every time it does the right thing.

A generation check would separate the two: keep the last `clearModelGeneration` per atlas, and when it has advanced since the previous poll, treat it as "the atlas invalidated itself" — consume the latch, re-baseline the tags, skip the rebuild. The addon has already bumped the generation, so GlyphRenderer has re-rastered without our help.

src/renderer/terminal/__tests__/atlasGuard.test.ts
```diff
@@ -685,4 +745,35 @@ describe('atlasGuard', () => {
...
+      const guard = createAtlasGuard();
+      const pane = makePane(atlas);
+      guard.register(pane.entry);
+
+      vi.advanceTimersByTime(60_000);
+
+      const prevents = warn.mock.calls
+        .map((c) => String(c[0]))
+        .filter((l) => /\[wmux:atlas-guard] prevent/.test(l));
+      expect(prevents.length).toBeLessThanOrEqual(1);
```
openwong2kim: This is the only coherent-atlas case, and it sits at `maxPages - 1` for 60s, so `applyClear()` never runs — the self-eviction path is never entered and the CURE interaction above stays invisible. `CoherentFakeAtlas.growBy` is defined but never called anywhere in this file.

A case that grows past the cap and asserts zero `cure` lines would pin the invariant. While here, `<= 1` on all three is loose for a path that should never fire on a coherent atlas — `toBe(0)` would catch a single-firing regression.

package.json
```diff
@@ -9,7 +9,7 @@
     "wmux": "dist/cli-bundle/index.js"
   },
   "scripts": {
-    "postinstall": "node scripts/fix-node-pty.js",
+    "postinstall": "node scripts/fix-node-pty.js && patch-package",
```
openwong2kim: The patch filename pins 0.19.0 but the dependency is `"@xterm/addon-webgl": "^0.19.0"`. The lock holds it at 0.19.0 today, but any 0.19.x bump makes patch-package log a version mismatch and exit 0 — the fix vanishes, `typeof atlas.clearModelGeneration === 'number'` goes false, and the guard silently reverts to the old backstop with CI still green. Pinning the exact version, or a post-install check that the installed bundle exports `clearModelGeneration`, would make that loud.

Separately: `patch-package` is in devDependencies while `postinstall` always runs, so `npm ci --omit=dev` fails on a missing binary.

openwong2kim (commented):
The atlas work looks right to me, but the guard half doesn't hold up, and I think it undoes what this PR is for.

**The storm moves, it doesn't go away.** `coherent` only zeroes `nearTrigger`, so it gates PREVENT and nothing else. But a patched atlas doing exactly what I1/I2 tell it to do produces a *merge* signal: `_wipeToConstructorShape()` does `_pages.length = 0` then pushes one page, so the pool collapses 16 -> 1 (count-drop, and page 0 is a new object, so page-identity too), and `_tryMergeSameSizePages()` calls `_mergePages`, which this file already documents as the only emitter of `onRemoveTextureAtlasCanvas` — the `page-removed` latch. All three land on `mergeSignal`, which skips the `if (!mergeSignal)` cooldown and runs a full `rebuildGroup` plus a `console.warn`. So every self-eviction gets a whole-group model drop and re-raster stapled onto it, with no rate limit, in precisely the CJK-flood scenario this PR targets. That's the v3.38.7 field log again, just filed under `cure` instead of `prevent`, and now without the 30s brake.

The comment above the cooldown says CURE is safe to leave unlimited because "it only fires on a real merge". The patch is what makes that premise false, so the two changes have to be reconciled in the same PR.

**CI can't install.** All five jobs die in ~25s on `npm ci`: package.json and package-lock.json are out of sync, `Missing: esbuild@0.27.7` plus its platform packages. The lock diff drops 27 `node_modules/...` entries and adds 13 — it looks regenerated somewhere that didn't write the optional platform set, which cross-platform CI can't survive. Worth a clean `npm install` and a re-commit of the lock.

**A premise I'd like reconciled before this lands.** When we dug into this in July (against the sourcemapped originals, not the bundle), the conclusion was that the sibling-holds-stale-coordinates mechanism doesn't survive contact with the code: page merge and delete both set `_requestClearModel = true`, that flag is sticky with no reset-on-read, so no sharing renderer can miss it and each self-heals on its next `renderRows`. The consequence of atlas saturation came out as a per-frame full-rebuild *perf* regression, not visual corruption — and "resize fixes it but refresh doesn't" read as a buffer-level or firing-gate signal rather than atlas evidence. This PR's description assumes the opposite. One of the two is wrong and I can't tell which from here; if the field symptom is real then the July analysis missed something, and it would be good to say what, because that's what tells us this patch is aimed at the right layer.

Smaller things in the inline comments. Also: nothing here exercises the actual patched addon — `atlasCoherenceModel.ts` is our own restatement of the invariants and the guard tests run against a fake, so a regenerated or drifting patch has nothing to fail against.

Not merging on my side until the guard interaction and the lock are sorted.

changelog.d/902.md
```diff
@@ -0,0 +1,9 @@
+### Fixed
+
+- **Long-running CJK and TUI panes no longer paint the wrong glyphs
+  (or go black) until you drag a split.** Same-font terminals share one
+  WebGL glyph atlas; when that atlas filled, sibling panes kept texture
+  coordinates into pages that had been merged or failed to clear. The
+  atlas now wipes completely, never allocates a page the shader cannot
+  sample, and every pane rebuilds after a wipe. The existing safety net
+  still runs; it should almost never have to fire.
```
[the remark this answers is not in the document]
openwong2kim: > Rename this file to `changelog.d/902.md`.

Done in cb89068d.

src/renderer/terminal/__tests__/atlasCoherence.test.ts
```diff
@@ -0,0 +1,46 @@
...
+    const max = 4;
+    const atlas = createAtlasModel(max);
+    for (let i = 0; i < 200; i++) atlas.rasterize(`g${i}`, 64, 64);
+    expect(atlas.pageCount).toBeLessThanOrEqual(max);
+  });
+
+  it('I3: two observers both see a wipe', () => {
+    const atlas = createAtlasModel(4);
+    let a = atlas.generation;
+    let b = atlas.generation;
```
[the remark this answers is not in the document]
openwong2kim: > Use `const` for both observer baselines.

Done in cb89068d.

src/renderer/terminal/__tests__/atlasGuard.test.ts
```diff
@@ -685,4 +745,35 @@ describe('atlasGuard', () => {
...
+      expect(atlas.clearCalls).toBeLessThanOrEqual(1);
+      expect(pane.refreshes()).toBeLessThanOrEqual(1);
```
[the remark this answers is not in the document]
openwong2kim: > Require zero PREVENT repairs.

Assertions are now `toHaveLength(0)` / `toBe(0)` in cb89068d.

src/renderer/terminal/atlasGuard.ts
openwong2kim: > This gates PREVENT only. The patched atlas's own eviction path still reaches CURE below, and CURE has no cooldown

Agreed — that was the hole. Tick now stores last `clearModelGeneration` per atlas; if it advanced since the previous poll, we consume the latch, keep the new page-tag baseline, and skip the rebuild. GlyphRenderer already rebuilt via I3. Unpatched addons still have no field, so CURE is unchanged for them. cb89068d.

src/renderer/terminal/__tests__/atlasGuard.test.ts
openwong2kim: > A case that grows past the cap and asserts zero `cure` lines would pin the invariant.

Added. `growBy(1)` at the cap evicts (count-drop + new page objects) and the poll must emit zero `cure` and zero `prevent`. The under-cap 60s case is also `toBe(0)` now. cb89068d.

package.json
openwong2kim: > The patch filename pins 0.19.0 but the dependency is `"@xterm/addon-webgl": "^0.19.0"`.

Pinned to `0.19.0`. `patch-package` moved to `dependencies` so `npm ci --omit=dev` still has the binary. Lockfile regenerated with npm 10 so the optional platform set from main is back. A smoke test now fails if the installed 0.19.0 tree is missing `clearModelGeneration` / the I1 comment (6a824b90).

openwong2kim:
> The storm moves, it doesn't go away. `coherent` only zeroes `nearTrigger`...

Fixed in cb89068d. CURE now no-ops when `clearModelGeneration` advanced since the last poll — that is the atlas invalidating itself, and I3 already rebuilt every owner. Unpatched addons still get unlimited CURE. A grow-past-cap test asserts zero `cure` lines.

> CI can't install... package.json and package-lock.json are out of sync, `Missing: esbuild@0.27.7`

The previous lock was rewritten by npm 11 and dropped the optional platform set. Regenerated from `origin/main` with npm 10. esbuild mention count is back to 157, matching main.

> When we dug into this in July... `_requestClearModel` is sticky... The consequence of atlas saturation came out as a per-frame full-rebuild *perf* regression, not visual corruption... This PR's description assumes the opposite.

I1 and I2 don't depend on that reading: `clearTexture` was a page-0-only no-op, and `_createNewPage` could still append past the sampler budget. Those stay.

I3 is the conflict. The 0.19 sources this patch was written against consume `_requestClearModel` in `TextureAtlas.beginFrame` (first caller wins). That's why the patch deleted the boolean and gave each `GlyphRenderer` its own generation. If the July sourcemap walk showed the flag never reset, then I3 is an API cleanup rather than the visual-corruption root, and the user-facing claim rests on I1/I2 plus the existing model-drop in `atlasGuard`. I can't re-run the July sourcemap walk from here. If you still have that note, that's the artifact that decides whether the PR description should drop the "sibling stale coordinates" sentence.

> nothing here exercises the actual patched addon

6a824b90: the test file now reads the installed 0.19.0 `TextureAtlas.ts` and fails if the I1 comment or `clearModelGeneration` is missing.

Manual Hangul 2-pane dogfood is still not run.

openwong2kim:
Everything I raised is addressed — the generation check is in the right place (tags are re-baselined at the top of the tick, before the `continue`, so the collapse doesn't just defer a CURE by one poll), the grow-past-cap test pins it, the pin and the smoke test close the silent-skip hole, and CI is green.

On the one open question, I went and looked, because it's decidable from the installed tree rather than from memory. The July reading holds, and your "first caller wins" does not.

`_requestClearModel` has exactly four references in the whole pristine 0.19.0 addon:

```
src/TextureAtlas.ts:133:  private _requestClearModel = false;
src/TextureAtlas.ts:135:    return this._requestClearModel;
src/TextureAtlas.ts:195:      this._requestClearModel = true;
src/TextureAtlas.ts:798:          this._requestClearModel = true;
```

and `beginFrame` is the whole of line 134-136:

```ts
public beginFrame(): boolean {
  return this._requestClearModel;
}
```

It reads. It does not clear. There is no `= false` anywhere outside the field initializer, in `TextureAtlas` or anywhere else in the addon — `GlyphRenderer.beginFrame` just forwards, and `WebglRenderer` only branches on the return. So no sharing renderer can miss the signal: once a merge or a delete sets it, every `GlyphRenderer` reads `true` on its next frame and rebuilds. The sibling-keeps-stale-texture-coordinates mechanism isn't reachable on this version, and I'd take that sentence out of the description and out of the changelog entry.

**I'd keep I3 anyway, for the opposite reason.** A flag that is set and never reset doesn't cause a missed rebuild — it causes a permanent one. After the first merge, every renderer sharing that atlas full-rebuilds its model *every frame, forever*. That is what atlas saturation actually costs you, and it matches the symptom people report better than corruption does: a pane that gets progressively heavier during a long CJK session, on a machine where the same pane was fine an hour earlier. Replacing the sticky boolean with a per-renderer generation turns "every frame from now on" into "once per invalidation", which is exactly the shape this patch has.

I1 and I2 stand on their own and don't touch any of this. I1 in particular is plainly a bug in the same file — `clearTexture` early-returns on page 0's cursor alone (`src/TextureAtlas.ts:138-141`), so an atlas whose page 0 happens to be idle while fifteen other pages are full clears nothing at all.

So: keep the patch, change the story. The user-facing entry should promise the re-raster treadmill going away rather than corrected glyphs. If the two-pane Hangul dogfood *does* reproduce wrong syllables or a black pane, that's worth knowing on its own — it would mean something other than this atlas path is producing it, and we'd be shipping a real fix and a wrong explanation at the same time.

PR #903, opened 2026-08-15
Let a poll wait for a pane instead of spinning on terminal_read

openwong2kim:
An orchestrator that wants to know when another pane blocks on a question has had one option: read that pane's terminal on a timer and guess from the text. That is the loop behind most false "this agent is stuck" signals — a printed question reads exactly like a pending one.

Most of what is needed was already here. `pane.list` carries per-pane `agentStatus` and `pendingQuestion`, and `agent.lifecycle` already fires per pane with the kind that separates "turn ended" from "paused on a prompt". What was missing was a way to WAIT for that event without polling for it.

## What this adds

`wmux_events_poll` takes three optional params:

| | |
|---|---|
| `blockMs` | wait this long for a match instead of returning an empty page |
| `ptyId` | only events about this pane |
| `kinds` | narrow `agent.lifecycle` by kind |

Every default is the previous behavior. A caller that passes none of them is byte-for-byte unchanged, including the 44 pre-existing tests, which pass untouched.

## The waiting has to not lose events

That is most of the design:

- a match already in the ring returns immediately, so an event that landed before you asked is never missed;
- a wake is "something was emitted", not "your thing was emitted" (the bus hook is global), so unrelated traffic re-collects and keeps waiting rather than returning an empty page;
- an attempt that matches nothing never advances the cursor;
- scope resolution happens once per call, not per wake — it is a property of the caller, and a mid-park identity change must not be able to flip a private scope open.

`blockMs` is capped at 10 minutes, deliberately under the MCP host's 30-minute stdio idle window, so a single parked poll can never be the thing that trips it. Parked polls are capped at a quarter of the pipe server's connection budget: each one holds a connection, so that cap is the back-pressure. Over it we answer immediately rather than queueing.

The client-side RPC deadline is now per call. It was already a per-call timer on a per-call socket, so this is a parameter rather than a new mechanism, and only the blocking poll asks for more than the default. The default stays 10s — it applies to every other tool, and raising it would turn a genuinely wedged call into a long hang across all of them. A regression test holds that constant down.

## Reviewed three ways, twice

Claude + GLM-5.3 + Grok, then again on the fixes. Everything below only misbehaves once several polls park at the same time — which is the intended use, so none of it would have shown up dogfooding a single waiter.

- **Parked polls could exhaust the pipe server.** The cap was picked (32) without reference to `MAX_CONNECTIONS` (50). Park a fleet, add ordinary traffic, and the server stops accepting — every other MCP tool then fails with what the client reports as "wmux is not running". A self-inflicted outage that looks like a crash and does not point back here. The cap is derived from that budget now, and the constant is exported so the coupling is visible from both ends.
- **One waking poll could steal another's wake.** EventBus fans out with `for (const sub of this.subscribers)`, so a subscriber that removes itself mid-fanout shifts the array and the iterator skips the next one. Measured: `[A,B,C]` with A self-removing fires A and C, never B. The unsubscribe is deferred to a microtask.
- **Nothing cancelled a park when the client hung up.** `RpcContext` carries an optional `AbortSignal`; `PipeServer` ties one to each connection.
- **`resync` kept waiting.** It means the cursor fell out of the ring, so the events are already gone — it returns immediately now.
- Plus: a cheap per-event predicate instead of waking every parked poll for a full ring re-scan; `kinds: []` normalized to "no filter" (an empty Set is truthy, so it meant the opposite of `types: []` on the same call); a monotonic deadline; entitlements re-checked mid-park, narrow-only.

One finding was fixed and then reverted, which is recorded in the code because the fix looks obviously right: `nextCursor` advances past events your own filter skipped. Rewinding to the last delivered event does not help — non-matches interleaved before it are stepped over anyway (a test proved it), and rewinding to the first dropped event lets other panes' traffic pin the cursor forever. A scalar cursor cannot mean two filters at once, so the contract is stated instead of faked: **one cursor chain per filter combination**, said in the tool description, the test, and the code.

Three reported findings were verified as not real and are listed in the commit that closes them out.

## Verified on a running app

Not just unit tests — driven over the real pipe against a dev instance, which is the same path the MCP tool takes:

```
[0] plain poll           3ms    parked absent            unchanged for existing callers
[1] park + timeout    6013ms    parked=true              asked for 6000
[2] park + wake        734ms    [agent.awaiting_input]   woken by a real hooks.signal
[2b] subagent noise   4003ms    0 events                 sat through it
[3] already in ring      2ms    parked absent            immediate
```

`[2b]` is the one worth pointing at: an `agent.subagent_stop` on the same pane, same event type, did NOT wake a wait filtered to stop/awaiting_input. That is the #733 family — a nested subagent returning is not the pane becoming free — confirmed on the real hook path, not just in a unit test.

## Not in scope

The hook-coverage half of this work (more agents reporting state, and the A/B harness that proves an observation hook is harmless) is separate and comes next. This PR is the primitive that half will feed.

## Summary by CodeRabbit

* **New Features**
  * Added blocking event polling with configurable wait times and filters by pane and event type.
  * Added cancellation support when connections close or requests are interrupted.
  * Added extended timeouts for blocking event polls while preserving existing defaults for other requests.
  * Added connection and waiting-capacity safeguards.

* **Bug Fixes**
  * Improved event-wait handling for resynchronization, timeouts, disconnects, and unrelated events.
  * Improved wakeups for relevant recipient-only events and concurrent waiters.
  * Improved pane detection for pending questions and nested activity.


openwong2kim:

This PR turns `wmux_events_poll` from a spin into a wait: `blockMs` parks the call until a matching event lands, and `ptyId` / `kinds` narrow the wait to one pane and one lifecycle signal, which retires the terminal-read-and-guess loop behind most false "this agent is stuck" reports — a printed question and a pending one finally stop looking identical. It touches the pipe server (per-connection `AbortSignal`, `MAX_PIPE_CONNECTIONS` exported so the parked-poll cap is derived from the shared budget rather than picked), the RPC router (cancellation threaded orthogonal to the trust lane), the events handler, and the MCP client, where the 10s default deadline is preserved and only a blocking poll raises its own. The review work recorded in the diff is what I would weigh at merge: the deferred EventBus unsubscribe that stops one parked poll from stealing another's wake, the connection-budget coupling, narrow-only mid-park entitlement re-checks, and the honestly stated one-cursor-chain-per-filter contract instead of a fake losslessness. Verified live over the real pipe including the subagent-noise case, with the 44 pre-existing tests passing untouched — filed as P2, the enabling primitive for the orchestrator follow-up rather than an urgent fix.

src/main/pipe/handlers/__tests__/events.rpc.test.ts
```diff
@@ -994,3 +994,368 @@ describe('events.rpc — agent transport identity scoping (audit B3)', () => {
...
+  });
+
+  it('caps parked polls well under the pipe server connection budget', async () => {
+    // A parked poll holds a connection for its whole wait. Sized independently
+    // of MAX_CONNECTIONS it can exhaust the server, and every OTHER MCP tool
+    // then fails with what the client reports as "wmux is not running".
+    const { MAX_PIPE_CONNECTIONS } = await import('../../PipeServer');
+    // The cap must leave the majority of the budget for ordinary
+    // connect -> send -> close traffic.
+    expect(Math.floor(MAX_PIPE_CONNECTIONS / 4)).toBeLessThan(MAX_PIPE_CONNECTIONS / 2);
```
openwong2kim, line 1329: This asserts `12 < 25`. It reads `MAX_PIPE_CONNECTIONS`, does arithmetic on it, and compares the result to more arithmetic on the same constant — `MAX_PARKED_POLLS` never appears, and no poll is ever dispatched. Remove the cap enforcement from `events.rpc.ts` and this test still passes.

The property worth pinning is the one the comment describes: park `MAX_PARKED_POLLS` polls, dispatch one more, and assert that the last returns immediately with `parkedCapReached: true` while the others stay parked.

src/main/pipe/handlers/events.rpc.ts
```diff
@@ -284,40 +457,151 @@ export function registerEventsRpc(
...
+    };
+
+    // Cheap pre-filter for the parked path (see waitForEmit). Mirrors the
+    // narrowing filters only — never the scope filters, which is why it can be
+    // over-permissive without being wrong: a false positive costs one extra
+    // collect(), a false NEGATIVE would be a missed wake, so anything uncertain
+    // must pass.
+    const interesting = (event: WmuxEvent): boolean => {
+      if (types && !types.includes(event.type)) return false;
+      if (ptyId && (event as { ptyId?: unknown }).ptyId !== ptyId) return false;
```
openwong2kim, line 531: This mirrors the narrowing filters but not the scope ones, and the comment justifies that as "a false positive costs one extra collect()". For a poll that passed `ptyId` or `types` that is true. For a `blockMs`-only poll — which the tool description explicitly allows — nothing here returns false, so it wakes on *every* event on the bus, re-scans the whole 1024-entry ring through the full scope chain, gets an empty page because the events belong to other workspaces, and parks again. At fleet scale that is O(parked × ring) of synchronous main-process work per event, on a bus that ticks once per shell command under OSC 133.

`clientSet` / `clientScoped` are already resolved above, so putting them in here is free:

```ts
if (clientScoped && event.type !== 'a2a.task'
  && !clientSet.has(event.workspaceId)) return false;
```

The `a2a.task` exemption matters — it is dual-party, so its `workspaceId` is the sender and a receiver-side match would be filtered out wrongly. `collect()` stays the authority either way; this only skips wakes that provably cannot match.

src/main/pipe/handlers/events.rpc.ts
```diff
@@ -284,40 +457,151 @@ export function registerEventsRpc(
...
+    // needs to reconcile via pane_list NOW, and every extra minute is more
+    // history sliding out of the window.
+    const first = collect();
+    if (blockMs <= 0 || first.events.length > 0 || first.resync) return first;
+
+    // ── Parked path ───────────────────────────────────────────────────────────
+    // Over the cap we degrade to the immediate answer rather than queueing:
+    // parking is a convenience, and a caller that cannot park is strictly better
+    // off polling than waiting behind others for a connection slot.
+    if (parkedPolls >= MAX_PARKED_POLLS) {
```
openwong2kim, line 576: The cap check sits after `resolveCallerWorkspace` (line 372, a renderer round-trip) and `refreshNotificationsGate` (line 400, a trust-DB read). So a caller that is over the cap pays the most expensive part of the call and then gets told it cannot wait — and being over the cap means a fleet is parked, i.e. exactly when the main process is least able to spare a renderer hop per rejected poll.

Hoisting the check above the scope resolution fixes it: the answer in that branch is the pre-block behaviour, which does not depend on anything resolved after it.

openwong2kim (commented):
The scanner fix is right, and right for the right reason — pointing it at the export keeps the die-loudly behaviour instead of teaching it to shrug, and the table now says where the number actually lives. CI is green on all six.

The design holds up too. I went looking for a lost-wakeup and did not find one: `collect()` and `subscribe()` have no await between them, so nothing can land in the gap; the deferred unsubscribe is the correct fix for EventBus's `for (const sub of this.subscribers)` self-removal skip; the abort is per connection rather than per request, which is what keeps listeners off a shared socket; and a client-side timeout is not retryable (`!isRetryable && !isPerm → throw`), so a 600s poll cannot re-park itself through the retry loop or the TCP fallback.

Three things I would still fix before this lands, none of them a correctness bug in the happy path — all three only bite once several polls park at the same time, which is the case this feature exists for.

**The cap test proves nothing.** `expect(Math.floor(MAX_PIPE_CONNECTIONS / 4)).toBeLessThan(MAX_PIPE_CONNECTIONS / 2)` is arithmetic on an imported constant — 12 < 25. It never mentions `MAX_PARKED_POLLS` and never dispatches a poll, so deleting the enforcement entirely leaves it green. The comment above it describes the outage it prevents (server stops accepting, every other MCP tool reports "wmux is not running"), and that is exactly the property with no coverage.

**The pre-filter skips the scope filters, and a `blockMs`-only poll is the shape that pays for it.** Details inline.

**The cap check runs after the expensive part.** Also inline.

Smaller, not worth blocking on: `Math.min(blockMs, blockMs - elapsed)` can only ever pick the second operand — `performance.now()` is monotonic, so the clamp is unreachable; and the immediate path returns without a `parked` field at all while the capped path returns `parked: false`, so a caller has to treat `undefined` and `false` as the same thing.

openwong2kim:
Pushed the first two in 8161e832, and I need to withdraw the third.

**The cap-check ordering finding was wrong.** I said a capped caller pays the renderer hop for nothing. It does not: the branch returns `{...first, ...}`, and `first` is a real polled page whose emptiness is *what makes the code get there at all*. Deciding "there is nothing to return, so parking is what's left" requires the scope filters — an unscoped ring read would have found other workspaces' events and returned at the line above. So the hop is load-bearing for both the park decision and the answer, and hoisting the check would mean answering without knowing whether there was anything to answer with. Left alone.

**The pre-filter fix is in, but not the shape that was suggested.** Exempting only `a2a.task` would have shipped a missed wake. `clientSet` is consulted for the LIFECYCLE types only; the private types are scoped per recipient — `channel.message` carries the SENDER in `workspaceId` and the membership in `recipientWorkspaceIds`, and `channel.catalog` can carry a `'*'` broadcast. So the exemption is `PRIVATE_EVENT_TYPES`, and there is a test that parks a poll scoped to a workspace which is only a *recipient* and requires it to wake. Dropping the exemption fails it in 5s — the poll burns its whole budget, which is what a missed wake looks like from outside: the event is already in the ring, so nothing re-announces it.

**The cap test now tests the cap.** Fills every slot, shows the next caller is turned away with `parkedCapReached` rather than queueing, shows the parked ones were undisturbed and still wake on their own panes, and shows the slots come back. Confirmed by mutation — disabling the enforcement fails it.

`tsc`, `eslint`, and `src/main/pipe` + `src/mcp` (1065 passing) are clean here.

PR #904, opened 2026-08-15
Pin the atlas invariants the guard actually leans on

openwong2kim:
## Summary

Follow-up to #902. Three gaps between what the tests check and what the patched addon does. None is a runtime defect — the patch and the guard are both correct as merged, verified against the installed 0.19.0 sources. All three mean a regression in them would ship green.

## What was unpinned

**A merge never reached the guard's coherence check.** The guard skips its repair when `clearModelGeneration` advanced, on the grounds that the atlas already rebuilt its owners. A reducing merge is the case that matters: it fires the removal event, which outranks both polled signals (`removed` short-circuits `detectMerge`). Nothing exercised it — `CoherentFakeAtlas` had no removal event at all, so the latch was never even armed on that path. The only coherent case was an evict, which produces count-drop and page-identity but no event.

**The model diverged from the patch in two places**, so its green said nothing about either:

| | model, before | the patch |
|---|---|---|
| page occupancy | one `used` pixel counter | `glyphs.length`, both `currentRow` axes, `fixedRows` |
| `clearTexture` no-op | required an empty cache; left caches alone | keys on SHAPE alone; empties caches regardless |

The occupancy one matters most: a single counter cannot express a page that holds glyphs while its packing cursor sits at the origin. That is what a merged page looks like, and it is exactly the state upstream's cursor-only probe called empty — the bug I1 exists to fix. The model could not have caught it.

The cache one describes the opposite algorithm. The patch empties the caches on the no-op path because a sentinel (empty-glyph) entry occupies no page, so a shape probe cannot see it and it would otherwise outlive a clear.

**The I3 test asserted the same thing twice.** `const a = atlas.generation; const b = atlas.generation;` reads one counter into two variables — there are no two observers in it.

## Changes

- `CoherentFakeAtlas` gains the removal event and a reducing `mergePages()` that bumps the generation, matching `_tryMergeSameSizePages`.
- Two guard cases: a merge that bumps must not cure; a collapse **without** a bump still must. The second is what keeps the skip honest — it is conditioned on the generation, not on the atlas merely looking coherent, and this fails if that ever inverts.
- Model: multi-axis occupancy, a `mergePages()` that produces the glyphs-without-cursor page, and a `clearTexture` no-op branch that matches the patch. Sentinel glyphs are modelled, since they are the reason that branch clears caches.
- I3 becomes two independent observers, each holding its own last-seen generation — the property that replaced the shared one-shot boolean. A seeing the invalidation must not stop B from seeing it.

## CHANGELOG entry

n/a — tests and test-only model. No user-facing change.

## Stability tier impact

n/a

## Test plan

- `src/renderer/terminal` — 18 files, 280 passing in this worktree, including the 42 across `atlasGuard` and `atlasCoherence`.
- Both new guard cases were confirmed to be real assertions rather than tautologies: the no-bump case fails if the generation check is widened to skip on `coherent` alone.
- `tsc --noEmit` and `eslint` clean on the touched files.

## Related issues / PRs

Follows #902.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved texture-atlas maintenance during page merging and clearing.
  * Preserved cached zero-area glyphs without consuming unnecessary space.
  * Improved invalidation handling to prevent redundant updates while ensuring stale data is refreshed.

* **Tests**
  * Added comprehensive coverage for page occupancy, cache clearing, merging, and independent observer invalidation behavior.

### DIFF
--- a/src/renderer/terminal/__tests__/atlasCoherence.test.ts
+++ b/src/renderer/terminal/__tests__/atlasCoherence.test.ts
@@ -35,14 +35,88 @@ describe('atlas coherence model', () => {
     expect(atlas.pageCount).toBeLessThanOrEqual(max);
   });
 
-  it('I3: two observers both see a wipe', () => {
+  it('I1: a merged page counts as occupied even with its cursor at the origin', () => {
+    // The page a merge produces holds every glyph it absorbed while its
+    // packing cursor is untouched. Upstream probed the cursor alone, so it
+    // read exactly this page as empty and skipped the clear. Occupancy has to
+    // be multi-axis or the clear silently does nothing.
+    const atlas = createAtlasModel(8);
+    for (let i = 0; i < 5; i++) atlas.rasterize(`g${i}`, 64, 64);
+    expect(atlas.pageCount).toBeGreaterThan(4);
+
+    expect(atlas.mergePages(4)).toBe(true);
+    const afterMerge = atlas.pageCount;
+
+    atlas.clearTexture();
+    expect(atlas.pageCount).toBeLessThan(afterMerge);
+    expect(atlas.pageCount).toBe(1);
+    expect(atlas.cacheSize).toBe(0);
+  });
+
+  it('I1: the no-op branch still empties the caches, and does not bump', () => {
+    // A sentinel (empty-glyph) entry occupies no page, so a shape probe cannot
+    // see it — it would otherwise survive a clear. Nothing an observer holds
+    // has moved though, so this must NOT invalidate anyone.
     const atlas = createAtlasModel(4);
-    const a = atlas.generation;
-    const b = atlas.generation;
     atlas.rasterize('x', 8, 8);
     atlas.clearTexture();
-    expect(atlas.generation).not.toBe(a);
-    expect(atlas.generation).not.toBe(b);
+    const generationAfterWipe = atlas.generation;
+    expect(atlas.cacheSize).toBe(0);
+
+    // Constructor shape now. Put a cache entry back without occupying a page.
+    const sentinel = atlas.rasterize('sentinel', 0, 0);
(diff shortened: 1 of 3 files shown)


openwong2kim:
**Priority: P3** — test-only hardening, no runtime surface touched.

This PR closes three real gaps between what the tests asserted and what the patched `@xterm/addon-webgl` 0.19.0 addon actually does, so a regression in any of them would previously have shipped green. The two that matter most: `CoherentFakeAtlas` had no removal event at all, meaning the guard's strongest cure signal — a reducing merge, where `removed` short-circuits `detectMerge` — never reached the generation-bump skip logic the guard relies on; and the coherence model diverged from the patch on page occupancy (a single used-pixel counter cannot express a merged page that holds glyphs while its packing cursor sits at the origin, which is exactly the state upstream's cursor-only probe misread as empty, i.e. the bug I1 exists to fix) and on `clearTexture`'s no-op branch (the patch empties the caches keyed on shape alone so sentinel empty-glyph entries cannot outlive a clear — the old model required an empty cache, describing the opposite algorithm). The old I3 test read one counter into two variables and asserted nothing about independence; the rewrite runs two observers each holding their own last-seen generation, mirroring `GlyphRenderer.beginFrame`, so A consuming the invalidation provably cannot consume B's.

The new guard pair is well-aimed: a self-merging coherent atlas must not be cured, while a page collapse without a generation bump must still trigger the backstop — the second case is what keeps the skip conditioned on the generation itself rather than on the atlas merely looking coherent, and it fails if that ever inverts. All three files are tests or test-only model code, so this carries no user-facing risk and the 280-test run plus the tautology check on the new assertions gives good confidence the pins bite. Worth merging as regression armor for the #902 line of work; P3 because nothing here blocks or degrades the shipped product.

PR #906, opened 2026-08-15
fix: persist Anthropic usage meter opt-in

### LINKED ISSUE #896: [Bug]: "Show 5h / 7d utilization in status bar" is reset every time you restart the app
### What happened?

Every time I relaunch the app, "Show 5h / 7d utilization in status bar" is toggled off.
Initially thought it was because I didn't have an account to log in to, but after log in it still has that problem.

### Steps to reproduce

1. Toggle the feature on
2. Quit the app completely
3. Relaunch the app
4. Toggle is reset to off

### Expected behavior

1. Toggle the feature on
2. Quit the app completely
3. Relaunch the app
4. Toggle is still on

### wmux version

3.41.1

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

Windows 10 19045.6466

Claude Code

```shell

```

### Confirmation

snowyukitty:
## Summary

Persist the Anthropic 5h / 7d usage-meter opt-in so enabling it survives an app restart and resumes main-process polling.

## Changes

- Add the optional `anthropicUsageEnabled` preference to `SessionData` and save explicit `false` values.
- Restore only valid boolean values and synchronize the main-process usage poller during session hydration.
- Add regression coverage for enabled, disabled, missing, malformed, and save-side behavior.

### Fixed

- Persist the "Show 5h / 7d utilization in status bar" setting across app restarts.

## Stability tier impact

n/a — no protocol, RPC, event, permission, or named-pipe surface changes.

## Test plan

- `npx vitest run src/renderer/stores/slices/__tests__/workspaceSlice.loadSession.test.ts src/renderer/components/Layout/__tests__/appLayout.sessionSaveInvariants.test.ts` — 50 passed.
- `npx tsc --noEmit` — passed.
- `npm run test:parallel` — 11,235 passed, 24 skipped.
- `npm run test:runtime` — 168 passed, 1 skipped, 4 unrelated Windows/Node 24 runtime failures: two `node-pty` AttachConsole timeouts in shell integration and two existing web-restart mtime assertions.

## Related issues / PRs

Closes #896

## Reviewer checklist

## Summary by CodeRabbit

* **New Features**
  * Anthropic utilization preferences now persist across app restarts.
  * Usage polling resumes automatically when a session is restored.
  * Older or malformed session data safely defaults utilization tracking to disabled.

* **Bug Fixes**
  * Restored settings now correctly synchronize with utilization tracking.

### DIFF
--- a/changelog.d/906.md
+++ b/changelog.d/906.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **The Anthropic 5h / 7d utilization setting now survives app restarts.**
+  The toggle used to update only the running app, so every new launch silently
+  turned the status-bar meter off again. wmux now remembers the explicit opt-in
+  and resumes usage polling after restoring the session. Older or malformed
+  session files still default safely to off. (#906)
--- a/src/renderer/components/Layout/AppLayout.tsx
+++ b/src/renderer/components/Layout/AppLayout.tsx
@@ -266,6 +266,7 @@ function buildSessionData(dumped: Map<string, boolean>): SessionData {
     notificationSoundEnabled: state.notificationSoundEnabled,
     toastEnabled: state.toastEnabled,
     notificationRingEnabled: state.notificationRingEnabled,
+    anthropicUsageEnabled: state.anthropicUsageEnabled,
     mutedNotificationCategories: state.mutedNotificationCategories,
     customKeybindings: state.customKeybindings,
     autoUpdateEnabled: state.autoUpdateEnabled,
--- a/src/renderer/components/Layout/__tests__/appLayout.sessionSaveInvariants.test.ts
+++ b/src/renderer/components/Layout/__tests__/appLayout.sessionSaveInvariants.test.ts
@@ -80,6 +80,10 @@ describe('AppLayout — axis A session-save invariants', () => {
     expect(source).not.toMatch(/addEventListener\('beforeunload', saveSession\)/);
   });
 
+  it('persists the Anthropic usage opt-in explicitly, including false', () => {
+    expect(source).toMatch(/anthropicUsageEnabled:\s*state\.anthropicUsageEnabled/);
+  });
+
   // Fix B — cap-skipped suspended promote. Boot recovery honours a session cap,
   // so a workspace beyond the cap came back with its ptyId absent and reconcile
   // destructively cleared it (losing the pane's scrollback and identity). The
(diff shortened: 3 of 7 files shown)


openwong2kim:

This PR fixes #896, where the "Show 5h / 7d utilization in status bar" toggle silently reset to off on every app restart. The change adds an optional `anthropicUsageEnabled` field to `SessionData` (`src/shared/types.ts`), persists it explicitly (including `false`) from `AppLayout.buildSessionData`, and restores it in `workspaceSlice.loadSession` with strict `typeof === 'boolean'` validation plus a synchronization call into the main-process usage poller during hydration. The `useNotificationListener` re-mirror is kept as an idempotent fallback, and regression coverage spans enabled, disabled, missing, malformed, and save-side behavior. This is a small, well-scoped fix that closes a real user-facing annoyance; the only thing to watch in review is that `loadSession` now reaches `window.electronAPI.usage.setEnabled` directly, so the ordering between hydration and listener mount (the fallback path) stays covered by the new tests.

openwong2kim:
Merged — thanks. Verified it end to end on Windows rather than just reading it: toggled the switch through Settings, confirmed `anthropicUsageEnabled: true` reached `session.json`, restarted the app, and the 5h / 7d meter came back up in the status bar with live numbers. So the restore path reaches main's poller, not just the store. That is #896's exact repro, gone.

The shape is right too — putting it in `buildSessionData` makes it behave like every sibling setting rather than inventing a second persistence path, and the `typeof === 'boolean'` guard means an older session file (or a malformed one) falls through to the default instead of restoring `undefined`.

Two things I noticed and did not hold the merge for. Both are yours to take or leave:

1. The rewritten comment in `useNotificationListener.ts` says "UsagePoller.setEnabled is idempotent", but `UsagePoller` has no `setEnabled` — the path is `IPC.USAGE_TOGGLE` → `usagePoller.start()/stop()` (main/index.ts:833), and it is `start`/`stop` that are individually idempotent. Worth naming the real symbols so the next person debugging this does not go looking for a method that isn't there.

2. The save-side test matches the source text (`/anthropicUsageEnabled:\s*state\.anthropicUsageEnabled/`). It would also pass on a conditional form that skips `false` — which is the exact bug class #896 was. The load side has real behavioural coverage; the save side is the half that broke, so it is the half worth executing `buildSessionData` and asserting `=== false` on.

Happy to push either of those myself if you would rather not — say the word.

PR #907, opened 2026-08-15
feat(alarm): gate the completion alarm on an actually-ended turn

openwong2kim:
## Problem

The "work finished" alarm fires on raw stop-shaped signals — the vendor `Stop` hook, the detector's `waiting` heuristic, `SubagentStop` — the instant they arrive. A background build still running, a subagent finishing, a TUI repaint: all of them told the user work was done when it wasn't.

## Change

A verdict machine (`src/shared/hooks/CompletionAlarm.ts`) now sits in front of every alarm-firing path. Each cue passes through:

```
hook / detector signal
  → normalize (working | attention | stop | session | answered)
  → turn gate (did the agent work this turn? already announced?)
  → provisional window (~1.5s, discarded when a rebuttal cue arrives)
  → confirmed alarm | status only
```

- **Leftover background work**: the hook bridge mines the transcript tail at Stop time for `run_in_background` starts not yet settled by a task-notification, and stamps `wmux_leftover_work` on the envelope (>0 → the turn has not ended). Stateless, bounded tail, fail-open.
- **Rebuttal cues**: tool activity, byte activity, prompt arrival cancel a pending window. The lastDetectedAgent fallback keeps text-only turns on ungoverned panes covered.
- **Ledger discipline**: gated cue classes write the dedup ledger at CONFIRMATION, never at arrival — a held-then-rebutted stop leaves no ghost emit, and a confirmed broadcast can never carry a stale dedup.
- **Symmetric attention hold**: approval prompts hold their own window; `answered` cancels it.
- Applied on both the daemon path (`HookIngest`) and the local main-pipe fallback (`hooks.rpc`, `PTYBridge`) — the fallback is the most false-positive-prone surface (no hook permissions).
- Status dots and fleet badges stay loose on purpose — only the alarm is strict. `agent.lifecycle` events gain a `decision: 'internal'` trace value for rejected candidates (documented in `wmux_events_poll`; the deck coalescer filters it).

## Behavior changes

1. The lifecycle tee and `onClaudeTurnEnd` fire only on a CONFIRMED turn end — a rebutted stop never triggers the usage probe.
2. Subagent stops no longer toast.

## Review

Multi-model review panel before the PR: two 2-model consensus findings fixed (resize-redraw repaints no longer rebut the completion window; bridge task-notification settlement is fail-open), plus the deck coalescer internal-trace filter. Remaining solo findings were verified against source and rejected (type mismatch impossible, design-intended).

## Test plan

- [x] Truth-table unit tests for the verdict machine (rebuttal set, mutual cancels, gate rejections, session reset, dropPty timer cleanup)
- [x] HookIngest end-to-end: hold→confirm, hold→rebut, subagent internal, detector pending, detect-then-hook ledger ordering
- [x] Bridge leftover mining: in-flight → 1, settled via queue-operation/attachment → 0, rejected attempt → 0, fail-open
- [x] Wiring guards: pending skips only the broadcast; session:active repaint flag precedes the alarm feed
- [x] Full suite: 11,233 parallel + 140 runtime pass

## Summary by CodeRabbit

* **Bug Fixes**
  * Completion and attention alerts now wait for confirmed turn completion before appearing.
  * Alerts account for active background tasks and allow brief follow-up activity to cancel false alarms.
  * Resize-related screen redraws no longer interrupt genuine completion detection.
  * Subagent completion events no longer trigger user-facing notifications.
  * Rejected or duplicate completion signals remain available as internal status traces without triggering notifications or lifecycle actions.

* **Documentation**
  * Updated event documentation to describe internal completion decisions.


openwong2kim:

This PR fixes the completion alarm firing on raw stop-shaped signals (vendor `Stop` hooks, the detector's `waiting` heuristic, `SubagentStop`) the instant they arrive — a background build still running, a subagent finishing, or a TUI repaint could all tell the user work was done when it wasn't. It introduces a verdict machine (`src/shared/hooks/CompletionAlarm.ts`: normalize → turn gate → ~1.5s provisional window → confirmed alarm or status-only) and applies it on both the daemon path (`HookIngest`) and the local main-pipe fallback (`hooks.rpc`, `PTYBridge`), the latter being the most false-positive-prone surface. The hook bridge also mines the transcript tail at Stop time for `run_in_background` starts not yet settled by a task-notification and stamps `wmux_leftover_work` on the envelope, so leftover background work holds the alarm; the mining is stateless, bounded (1MB tail), and fail-open. This matters because the alarm is the product's core "your agent needs you" signal — a false positive erodes trust in every other notification, and the truth-table coverage plus the review-driven hardening (repaint rebuttals, ledger-at-confirmation discipline, fail-open settlement matching) address exactly the failure modes that would silently suppress alarms. One callout for review: the signal is now a full state machine on two ingest paths, so the daemon and local fallback wiring (ledger ordering, `decision: 'internal'` trace filtering in the deck coalescer) deserves the closest look during dogfood.

PR #908, opened 2026-08-15
Tell the user an update is ready, instead of waiting to be asked

### LINKED ISSUE #897: [Bug]: "Check for updates" feature doesn't work as expected
### What happened?

I try to run "Check for updates" to auto update the app and it is seemingly downloading the update but doesn't actually update the version. Attaching video.

### Steps to reproduce

1. Launch the app
2. Go to "General" settings tab
3. Click "Check for updates"
4. It seemingly installs the update
5. App is restarted
6. Version is still the same as before

### Expected behavior

1. Launch the app
2. Go to "General" settings tab
3. Click "Check for updates"
4. It seemingly installs the update
5. App is restarted
6. Latest version is displayed in the settings

### wmux version

3.41.1

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

Windows 10 19045.6466

None / not relevant

```shell

```

### Confirmation

openwong2kim:
Three people described the same thing on #897: wmux downloads an update and then nothing happens. Pressing "Check for updates" by hand is the only thing that ever installs it — on macOS too, so this was never a Windows or a Squirrel problem. One reporter found the verified installer sitting in temp, quit through the tray, ran it manually, and it worked fine.

## The install path is not broken

A background poll calls `check()` with the default `oneShot=false`, and `performInstall` is reached only when that one-shot intent is set. So a background check downloads, verifies, and stops.

That part is deliberate and stays. Installing quits the app and every pane goes with it, so it is the user's call to make. **The bug is that we never told them there was a call to make.**

`UPDATE_AVAILABLE{downloaded}` fires once, at the moment the download finishes, and the only thing listening was the Settings panel — mounted only while Settings is open, which is almost never. So the app sat on a ready installer in silence, and the reporter's "no sufficiently obvious indication" is exactly right.

## Three places, one mistake

This subsystem has now made the same push-only mistake three times:

1. **#866's refused-install notice** — a main-side push into that same Settings-only listener. Already fixed by pulling from `AppLayout`, which is always mounted.
2. **No ready-notice at all** — fixed here.
3. **The Settings widget itself** — it subscribed to events and never asked what was already true, so opening Settings *after* a download showed a bare version number and a "Check for updates" button while a verified installer sat on disk. Pressing that button then appeared to fix things, which is the behavior being reported.

Same fix each time: read from a surface that is actually mounted.

## What changes

**A persistent bottom-right notice**, from `AppLayout`. Both halves are needed and neither is sufficient alone:

- a **read at mount** covers "it was ready before you looked";
- a **live subscription** covers "it became ready while you were working" — a poll finishing an hour into a session is the common case, and a mount-only fix stays silent for it until restart.

It persists rather than fading (`persist` already exists on the toast for exactly this: state that is still true five minutes later). A notice that fades leaves the user where they started. It names both versions and carries a one-click install, because otherwise it says something is ready and sends you to find the button in the panel you were never in.

The main-side read is a **read, not a take** — unlike the refusal marker, a pending install is still true after you look at it, and clearing it would make the notice appear exactly once per boot.

**The Settings widget** now reads state on mount and shows it as two lines instead of one:

```
Current  v3.42.0
Latest   v3.43.0 — Update ready        [Install now]
         Installing ends live sessions
```

The old copy put the running version and a status word on one line and never named the version you would move TO, so "update ready" could not be reconciled against anything — and the reporters were comparing exactly those two numbers by hand, one against `winget list`. The button said "Update ready", which is what the line above already reports; a button named after a state does not tell you what pressing it does, especially one that quits the app.

## Not changed

Background installs are still not automatic. The `#866` warning that installing ends live sessions stays. Settings remains a place to install from, just no longer the only one.

## Scope

This does not close #897. It fixes what two of the three reports describe. The third (pressed the button, app restarted, version unchanged) is a different branch — that one got as far as running the installer — and needs the reporter's `main-*.log` to tell which path it took.

## Tests

Structural, following the house pattern for `AppLayout` (no jsdom fixture). They pin the properties whose failure mode is silence, which is indistinguishable from "there was no update": the read comes from an always-mounted surface, the toast persists, it carries the action, it fires on the live event as well as at mount, it announces once, and `getPendingInstall` reports without consuming.

## Summary by CodeRabbit

* **New Features**
  * Added persistent, version-specific notifications when updates are ready to install.
  * Added an **Install now** action and clear installation failure messages.
  * Settings now retains update status across refreshes and displays current and latest versions accurately.
* **Localization**
  * Added update installation and error messaging in all 23 supported languages.
* **Bug Fixes**
  * Prevented repeated update checks from losing downloaded-update status.
  * Preserved pending update information until installation is completed or dismissed.


openwong2kim:

**Priority: P2** (fixes the bulk of #897)

This PR addresses the real failure behind #897: the background update poll downloads and verifies an installer, then says nothing, because `UPDATE_AVAILABLE{downloaded}` fired into a listener that only exists while the Settings panel happens to be open. The fix adds a `getPendingInstall` read on the main side (deliberately a read, not a take — a pending install is still-true state) and an always-mounted `usePendingInstallNotice` hook in `AppLayout` that covers both halves: what was already ready at mount, and what becomes ready mid-session, shown as a persistent toast naming both versions with a one-click install. The Settings widget gets the same correction — it now asks what is already true on mount, shows current and latest versions on separate lines, and the button says "Install now" instead of repeating a status. The structural tests pin exactly the properties whose failure mode is silence, which is the right guard against a fourth round of the push-only mistake in this subsystem. Scope is honestly stated: this closes two of the three #897 reports; the third (installer ran but the version did not change) is a different branch and still needs the reporter's `main-*.log` to diagnose.

PR #909, opened 2026-08-15
feat(terminal): add new terminal tab button

rayss868:
## Summary

Adds a discoverable `+` button to the pane surface-tab strip so users can create a new terminal tab without relying on `Ctrl+T`.

The new UI path shares the same terminal-creation helper as `Ctrl+T`, keeping startup gating, workspace profile handling, shell selection, IPC error handling, and spawned working-directory adoption consistent across both entry points.

## User-visible behavior

- A `+` button appears immediately after the final surface tab in each pane.
- Clicking `+` creates a terminal surface in the current pane and workspace.
- The newly created terminal becomes the active surface through the existing `addSurface` store action.
- The button uses the existing `pane.newTerminal` translation for its title and accessible label.
- The button stops event propagation, so clicking it does not select or activate an existing tab.
- The button remains available when the optional right-side pane action cluster is hidden.
- Existing split-right, split-down, browser, zoom, and close behaviors are unchanged.
- `Ctrl+T` continues to create terminals, but now routes through the same shared helper.

### Shared terminal creation helper

Added `src/renderer/utils/createTerminalSurface.ts` with an injected dependency contract for:

- target workspace and pane IDs;
- pane startup gate state;
- workspace profiles and global startup directory;
- default shell;
- the existing IPC invocation wrapper;
- PTY creation;
- the existing `addSurface` store action.

The helper:

1. Returns without side effects until `paneGate` is `ready`.
2. Resolves the requested workspace explicitly instead of relying on list order or global active state.
3. Resolves the initial working directory using workspace profile settings with the existing global fallback.
4. Builds PTY options using the existing default-shell and workspace-profile overlays.
5. Runs PTY creation through `useIpc().invoke`, preserving the existing toast/error behavior.
6. Adds the terminal to the requested pane only after successful PTY creation.
7. Adopts the cwd returned by the main process, falling back to an empty string when none is returned.

### Pane integration

`Pane.tsx` now wires the new `SurfaceTabs` callback to the shared helper using the rendered pane's concrete `workspace.id` and `pane.id`. This preserves correct targeting in multi-workspace and multi-pane layouts.

### Keyboard integration

The inline PTY creation logic in `useKeyboard.ts` was replaced with the shared helper. The shortcut retains its existing ready-gate, active-workspace lookup, IPC wrapper, shell, cwd, and surface-creation semantics.

### Tests

Updated the surface-tab action tests to cover:

- action ordering with `new-terminal` immediately after the tab strip;
- accessible label rendering;
- callback invocation without tab selection;
- terminal surface creation and activation in the target pane;
- continued visibility when pane actions are disabled.

Added helper tests for:

- successful terminal creation;
- startup-gate no-op;
- missing-workspace no-op;
- IPC failure handling;
- empty cwd fallback;
- explicit workspace/pane targeting when another workspace appears first;
- rejected PTY creation folded through the IPC wrapper.

Added a keyboard source-level assertion ensuring `Ctrl+T` uses `createTerminalSurface` rather than directly calling `window.electronAPI.pty.create`.

## Verification

- `npx tsc --noEmit --pretty false` — passed.
- Focused Vitest suite — passed: **3 test files, 62 tests**.
- ESLint on changed files — **0 errors**; six existing non-null assertion warnings remain.
- `git diff --check` — passed.

The broader non-runtime suite was also run:

- `npm run test:parallel` — one unrelated pre-existing failure in `src/renderer/terminal/__tests__/atlasCoherence.test.ts`:
  - `installed addon-webgl 0.19.0 still has the I1/I3 patch`

That failure is outside the files and behavior changed by this PR.

## Scope notes

- No package version changes.
- No `CHANGELOG.md` changes.
- No dependency changes.
- No changes to the optional local `.codebase-memory/` directory.

## Summary by CodeRabbit

- **New Features**
  - Added an optional `+` button after the terminal tabs for creating another terminal in the current pane.
  - Added a setting to show or hide the new-terminal button; it is disabled by default and preserved across sessions.
  - Added localized labels and OS-specific shortcut tooltips.

- **Updates**
  - `Ctrl+T` now consistently creates a terminal in the active pane.
  - Updated shortcut descriptions to distinguish terminal creation from workspace creation.

- **Documentation**
  - Added design specifications and changelog entries for the new terminal action.


openwong2kim:

This adds a visible `+` button right after the last surface tab so a new terminal can be created without knowing `Ctrl+T`, and the more valuable half is the refactor underneath: the whole PTY-create → addSurface flow is extracted into a shared `createTerminalSurface` helper that both the button and the `Ctrl+T` path now call, with focused tests for the helper (gate no-op, missing workspace, IPC failure, cwd adoption, explicit workspace/pane targeting), the tab-strip integration, and a keyboard regression guard. The helper preserves the keyboard path's contract exactly — pane-gate guard, profile/startup-cwd resolution, default-shell and workspace-profile overlays, the `useIpc` error toast on rejection, and adoption of the cwd main actually spawned in — and `Pane.tsx` passes the owning `workspace.id`/`pane.id` explicitly instead of relying on global active state, which is correct for multi-workspace layouts.

The one thing that needs a call before merge: the diff deletes the comment recording that this exact button was removed deliberately ("one pane = one terminal is the product concept; `Ctrl+T` stays as the power-user escape hatch; only the discoverable button is gone"). If that product decision still stands, this should be closed rather than merged; if the concept has loosened, the PR should say so and replace the removed rationale instead of silently dropping it, or the next contributor will re-litigate it. One small nit either way: `window.electronAPI.pty.create` is now passed as a detached reference in both call sites — fine if the preload exposes plain closures, but worth a quick confirm before it becomes the canonical creation path.

Priority: P3 — a discoverability nicety with no correctness impact either way; the helper extraction is worth keeping regardless of the button's fate.

openwong2kim:
Thanks for this — and sorry it sat for a bit. The gap you're pointing at is real, so rather than send you a list I pushed a commit reworking the form. Revert it freely if you disagree; it's your PR.

**The history worth knowing.** The pane header used to have a new-terminal button. It was removed on purpose in #451: one pane = one terminal is the shape we recommend, and splitting is the answer to "I want another terminal", so a button that adds a *second* terminal to the same pane didn't belong in the header. Ctrl+T was kept deliberately as the escape hatch — the capability stayed, only the discoverable affordance went. That reasoning lived in a comment right where the button had been, which this PR removed; I've replaced it rather than let it disappear silently.

**But you're right that something was missing.** A capability reachable only by a keystroke nobody advertises isn't much of a capability. That's a genuine problem and it's what this PR found.

**Where I landed:** the `+` opens a small menu instead of creating a terminal on click. The extra gesture is the point — it keeps "one pane = one terminal" as the default reading while putting the capability one click away, and the menu shows `Ctrl+T` next to the terminal entry so the keyboard path stops being a secret.

While there, new-browser moved out of the pane action cluster into the same menu. It creates a surface in this pane exactly like new-terminal does, so the two belong together; the cluster is left holding only what acts on the pane itself — split right, split down, zoom — and narrows from 116px to 90px.

**What I kept unchanged:** your `createTerminalSurface` extraction. Pulling the Ctrl+T path out with its collaborators injected, and documenting the behaviour contract point by point, is better than what was there before — the menu entry calls straight into it.

I also added `changelog.d/909.md`, which the PR was missing.

Verified live, not just in tests: the menu opens and closes (Escape included), `aria-expanded` tracks it, both entries create their surface in the pane, and — the part I most wanted to check — focus lands in the new terminal, so typed input goes there instead of sticking on the trigger. `tsc` clean, 3277 renderer tests pass.

If this looks right to you, say so and I'll merge it.

PR #911, opened 2026-08-15
feat(i18n): localize hardcoded sidebar & settings strings (zh)

cnxiekun:
## Summary

Routes hardcoded English strings in the sidebar `+` menu and Settings through the i18n `t()` system so the Chinese (zh) locale — and any future locale — is honored. **Translation only; no behavior change.** Also catches zh up with the two `fanout.*` keys added in #891.

## Changes

- Sidebar `+` menu: Browse Folder…, Empty workspace, and the five layout presets (previously hardcoded English) now render via `t()`.
- Settings → Accounts: the multi-account registry and add-account wizard strings now render via `t()`.
- Settings → MCP servers: status block now renders via `t()`.
- Settings → Theme section label + aria-label, and the "ESC to cancel" hint in the key-capture overlay.
- `en.ts` (source of truth) + `zh.ts`: ~60 new keys; zh ↔ en fully aligned (1270/1270).
- No English string text changed; no component logic/props/state/events touched. Agent-facing protocol strings (A2A delivery hints, MCP tool errors) intentionally left in English.

### Changed

- The sidebar **+ menu** and Settings (Accounts, MCP servers, Theme) now respect the selected locale instead of showing hardcoded English.

### Fixed

- Chinese (zh) locale catches up with the `fanout.roleLabel` / `fanout.roleNone` keys added in #891.

## Test plan

- `tsc --noEmit` clean; ESLint clean; related component tests pass (20 files / 232 tests).
- Parity gate (`check-locale.py`): zh ↔ en fully aligned, 1270/1270.

## Related issues / PRs

- Follow-up to the zh localization PR (#876).

## Summary by CodeRabbit

* **New Features**
  * Expanded localization across account management, settings, sidebar navigation, workspace presets, and fan-out role selection.
  * Added English and Chinese translations for common actions, account controls, theme settings, MCP status, and workspace layouts.
  * Localized onboarding, usage status, refresh controls, removal notices, and accessibility labels.
  * Improved consistency of translated labels and descriptions throughout the application.


openwong2kim:

This PR is a translation-only follow-up to #876: it routes the remaining hardcoded English strings in the sidebar `+` menu (PresetPicker: Browse Folder, Empty workspace, and the five layout presets) and in Settings (the multi-account registry, the add-account wizard, the MCP server status block, the Theme section label, and the key-capture overlay's ESC hint) through the `t()` system, so the zh locale — and any future locale — is honored there. It also brings zh up to date with the two `fanout.*` role keys added in #891, keeping the zh↔en parity gate at a full 1270/1270, and adds a `changelog.d` entry.

I reviewed the diff: it touches five renderer files plus the two locale files, changes no English copy, no component logic, props, state, or event handlers, and intentionally leaves agent-facing protocol strings (A2A delivery hints, MCP tool errors) in English — the right call for machine-parsed surfaces. One nit for review, not a blocker: `accounts.removedReverted` collapses the old singular/plural branch into `{n} workspace(s)`, which is slightly less polished in English than the code it replaces; the zh side is unaffected.

This matters because these were the last user-visible chrome surfaces showing English regardless of locale, and it lands cleanly at low risk. Priority: **P3** — pure polish with no behavior change.

cnxiekun:
Thanks — the singular/plural nit is addressed in d603d227 (accounts.removedReverted / accounts.removedRevertedPlural).

openwong2kim:
Merged — thanks, this is careful work. A few things I checked rather than took on faith:

- Every English string you moved is byte-identical in `en.ts`, ellipsis included — spot-checked ten of them.
- `preset.id` still drives `handleSelect`; only the label goes through `t()`. That is the trap this kind of change usually falls into and you avoided it.
- Ran it: switched to 中文 (简体) on Windows and walked the surfaces you touched. The `+` menu renders 浏览文件夹… / 空白单面板 / 左右分栏 / 三栏 / 主面板 + 侧栏 / 2×2 网格, and no raw key leaked anywhere in the UI. `tsc` clean, 220 tests across the touched components pass.
- Restoring the exact singular/plural instead of shipping "workspace(s)" was the right call, and `t(key, vars)` is used properly.

Two follow-ups, neither blocking:

1. `t(\`preset.${preset.id}.name\`)` is a dynamic key, and `t()`'s last fallback returns the key itself. All six ids have keys today, so nothing breaks — but adding a seventh preset without keys would render `preset.foo.name` verbatim in the menu, in every locale including English. `chromePresetLocale.test.ts` is the pattern for pinning that: assert every `LAYOUT_PRESETS` id has `preset.<id>.name`/`.description` in `en`. Worth adding while the reason is fresh.

2. Related: `LayoutPreset.name`/`description` are now unread by the only thing that displayed them, and `preset.single.*` can never render (the menu filters `single` out). Dead data in a shared type tends to get trusted by the next person — either drop the dead keys or say in the type that the fields are a fallback.

Also worth knowing: `check-locale.py` isn't in the repo, so your 1270/1270 parity result is a local check that CI does not reproduce. The repo only enforces parity for individual high-stakes keys (`updateSessionsLocale.test.ts` and friends). Not something you need to fix here — just so you know what is and isn't guarded when you build on this.

PR #912, opened 2026-08-15
feat(i18n): localize hardcoded sidebar & editor strings (zh)

cnxiekun:
## Summary

Routes hardcoded English strings in the sidebar and editor chrome through the i18n `t()` system so the Chinese (zh) locale — and any future locale — is honored. **Translation only; no behavior change.**

## Changes

- Sidebar → Company panel: department rows, CEO row (Active/Idle), add-member / manage / destroy confirmations, empty states, "+ Add Department".
- Sidebar → Missions section: section header, done disclosure, mission-channel tooltips/aria.
- Sidebar → workspace account submenu: vendor labels, "Default account", logged in/out badges, new-terminals note.
- Sidebar → workspace search placeholder.
- Editor panel: reload, view/edit segmented control, loading, read-error.
- File tree panel: header, refresh, empty / no-CWD states, markdown preview close, read-error.
- Error boundary: crash message + retry.
- `en.ts` + `zh.ts`: ~45 new keys; zh ↔ en fully aligned (1316/1316).
- Updated the CompanyPanel source-scan regression guard (`Sidebar.companyMode.test.tsx`) to pin the i18n key instead of the literal string.
- No English string text changed; no component logic/props/state/events touched.

### Changed

- The sidebar company panel, Missions section, workspace account submenu, workspace search, editor/file-tree panels, and the error boundary now respect the selected locale instead of showing hardcoded English.

## Test plan

- `tsc --noEmit` clean; ESLint clean for changed files; related tests pass (9 files / 94 tests).
- Parity gate: zh ↔ en fully aligned, 1316/1316.

## Related issues / PRs

- Part 2 of the localization follow-up. Stacked on #911 (this branch contains its commits; rebased onto main once #911 merges).


openwong2kim:

This PR is the second half of the zh localization follow-up: it routes the remaining hardcoded English strings in the sidebar (Company panel, Missions section, workspace account submenu, workspace search) and the editor/file-tree panels plus the error boundary through the i18n `t()` system. The diff is clean and scoped as described — every change is a string-to-key substitution, the ~45 new keys keep zh ↔ en at full parity (1316/1316), and the CompanyPanel source-scan guard was sensibly re-pinned to assert the i18n key rather than the literal text. Note the branch is stacked on #911 and carries its commits, so the Accounts/Settings/+ menu portions in this diff belong to that PR — worth keeping the rebase-onto-main plan in mind at merge time so #912's changelog fragment lands on top of #911's.

All checks are green (validate, cross-platform baselines, bench, CodeRabbit) and there is no behavior change, so this is low-risk. Marking P3 as with the rest of the localization series.

openwong2kim:
Now that #911 is in, this one has gone conflicting — and the conflict turns out to be the informative part, because it says the work already landed.

The four files unique to this PR's diff — `changelog.d/911.md`, `AccountsSection.tsx`, `SettingsPanel.tsx`, `PresetPicker.tsx` — are exactly the four that #911 touched. Same story on the keys:

- this branch adds 103 keys to `en.ts`;
- 58 of them are already on `main`, landed by #911;
- the other 45 are all in #913, and none of those 45 are on `main` yet, so #913 really is carrying them.

Diffing this branch's `en.ts` against #913's comes back empty — it's a strict subset. Resolving the conflict here would land nothing new.

So my inclination is to close this and keep #913 as the single follow-up: one branch is easier to review than two that overlap, and #913 covers the same ground. But you've been living in this split longer than I have — is there anything here you think #913 drops? I'd rather hear that before closing than after.

Either way, thanks for the work. The extraction itself was clean; where I sampled the removed literals they matched `en.ts` byte for byte, and the layout-preset contract test you added on #913 was exactly the right response to the gap I flagged.

cnxiekun:
Agreed — thanks for the analysis. #913 fully covers this branch (I verified the net diff locally: every sidebar/editor change is in it). Happy to close this one and keep #913 as the single follow-up. Thanks for the review.

PR #913, opened 2026-08-15
feat(i18n): localize remaining hardcoded strings (zh)

cnxiekun:
## Summary

Routes the last hardcoded English strings in the renderer through the i18n `t()` system so the Chinese (zh) locale — and any future locale — is honored. **Translation only; no behavior change.**

## Changes

- Toast actions (go-to-terminal, dismiss, notifications aria)
- Workspace multiview remove, notification panel aria labels, onboarding overlay (Step / Skip / Back / Done / Next)
- Relative time helpers (`timeAgo`), status-bar bell aria, usage-meter tooltip
- Deck fleet roster (shell fallback, role picker), orchestrator model "Default", diff panel
- Create-channel modal (name rules, errors, buttons), permission + execute approval dialogs, fleet card
- `en.ts` + `zh.ts`: ~50 new keys; zh ↔ en fully aligned (1386/1386)
- No English string text changed; no component logic/props/state/events touched. Agent-facing workflow step names and the cheat sheet intentionally-literal entries stay English.

### Changed

- Toast, notification panel, onboarding overlay, status bar, usage meter, deck fleet, diff panel, channel-create modal, and the approval dialogs now respect the selected locale instead of showing hardcoded English.

## Test plan

- `tsc --noEmit` clean; ESLint adds no new errors; related tests pass (exit 0).
- Parity gate: zh ↔ en fully aligned, 1386/1386.

## Related issues / PRs

- Part 3 (final) of the localization follow-up. Stacked on #911 (merged) and #912 (this branch contains its commits; GitHub shows only the net diff once #912 merges).

## Summary by CodeRabbit

* **New Features**
  * Expanded localization across workspace, sidebar, missions, company, onboarding, notifications, approvals, channels, editor, file tree, usage, and account interfaces.
  * Added Chinese translations for newly localized interface text.
  * Localized relative-time labels, accessibility text, errors, statuses, buttons, and empty states.
  * Added live locale updates for approval prompts and error screens, with safer interpolation of special characters.
* **Documentation**
  * Clarified localized layout preset naming.
* **Tests**
  * Added coverage for localization, interpolation, and layout preset translations.


openwong2kim:

This is the final part of the localization follow-up: it routes the last hardcoded English strings in the renderer through the i18n `t()` system so the Chinese (zh) locale — and any future locale — is honored across the remaining surfaces. The diff spans 31 files (+500/−129): toast actions and aria labels, the notification panel, onboarding overlay, status-bar bell aria, usage-meter tooltip, deck fleet roster (shell fallback, role picker), the diff panel, the create-channel modal, both the permission and execute approval dialogs, and the shared `timeAgo` helper, with ~50 new keys and zh↔en parity reported at 1386/1386. It stays translation-only as described — no component logic, props, state, or event handlers change — though two deliberate side effects deserve a reviewer's eye: the empty-value orchestrator model option now renders a localized "Default" label instead of its raw label, and the CompanyPanel regression guard pins `t('company.empty')` rather than the literal string. The new `layoutPresetsLocale.test.ts` is a welcome addition because it pins the dynamic `preset.<id>.name` key contract, so a future preset added without locale keys fails in tests instead of rendering the raw key verbatim in every locale. One minor note: `timeAgo`, `formatBellAriaLabel`, and `buildOkTooltip` now call the module-level `t()` directly, so their output is captured at render time and won't live-update on a locale switch until the surrounding component re-renders — acceptable for these surfaces, just worth knowing at review. Marking P3: no behavior change, but it closes out the locale story for a large share of the UI, and it should land only after #912 (it carries that branch's commits plus its own changelog entry).

openwong2kim (commented):
Thanks for the follow-up, and thanks for adding the layout-preset contract test — that is exactly the gap I flagged on #911, and pinning it in `en` is the right level.

This one I do not think should land as it stands. Four things, two of which change behaviour rather than language.

**The approval dialog is now untranslatable.** `approval.sameWsIntro` is `"An agent in"`, `sameWsMid` is `"wants to spawn another autonomous Claude CLI with"`, `sameWsEnd` is `"in the same workspace."` — the sentence is cut into fragments and its word order is held in JSX. Korean, Japanese and Turkish put the verb last; Arabic runs the other way. None of them can be produced by filling those four slots in that order, so the twenty-one locales this is meant to open up still cannot express it, and a translator receives `"An agent in"` as a unit with nothing to attach it to. That matters more here than anywhere else in the diff: this is the dialog that asks whether to spawn an agent with `bypassPermissions`. Whole-sentence keys with the emphasis carried as a token — `t('approval.sameWs', { workspace, mode })` and split on the result — keep both the meaning and the red span.

**Interpolation corrupts values containing `$`.** `t()` substitutes with `str.replace(re, String(v))`, and a string replacement treats `$&`, `$$`, `` $` `` and `$'` as patterns. Nothing hit it before because the interpolated values were counts and names; this PR is the first to route arbitrary runtime text through it — `error.crashed` takes a raw `Error.message`, `openChannelFor` takes an agent-authored title, `fleetRoleAria` takes a pane title with a path in it. Measured:

```
t('error: {message} here', { message: 'boom $& oops' })
  → 'error: boom {message} oops here'
```

The token comes back. One-line fix in `i18n/index.ts`, and it belongs in this PR because this PR is what makes it reachable:

```ts
str = str.replace(new RegExp(`\\{${k}\\}`, 'g'), () => String(v));
```

**English text changed.** `approval.fanoutCount` is `'{count} isolated task(s)'`, replacing `task{count === 1 ? '' : 's'}`. English now reads "1 isolated task(s)". That is the same thing CodeRabbit caught on #911, which you fixed properly there with `removedReverted` / `removedRevertedPlural` — worth the same treatment rather than the parenthesis.

**The dialogs will not switch language while open.** `ExecuteApprovalDialog`, `PermissionApprovalDialog` and `ErrorBoundary` call the module-level `t` rather than `useT()`. A locale change updates the store but does not re-render them, so an approval prompt already on screen stays in the previous language — and these are exactly the surfaces a user is least able to dismiss and reopen. `useT()` for the two dialogs; `ErrorBoundary` is a class, so it needs a subscription or a re-read at render.

Smaller: `deck.fleetShell` is still `'shell'` in `zh.ts` (untranslated); the `o.value === ''` → Default-label branch is duplicated across four call sites and will drift; `toasts.map((t) => ...)` in `ToastContainer` now shadows the translator you added above it, which is harmless today and a `TypeError` the moment someone calls `t('…')` inside that map; and the changelog fragments do not mention the multiview close button, the relative-time localisation that came with the `timeAgo` change, or the orchestrator model chip.

One piece of housekeeping: **#912 is subsumed by this branch.** It carries #911's commits, which is why it conflicts now that #911 is in. Every key it adds exists here, and the only files unique to its diff are the ones #911 already landed. Closing it and keeping this one is the tidier path.

The mechanical parts are good — I checked a sample of the removed literals and each is byte-identical in `en.ts`, `tsc` is clean, and 1160 renderer tests pass.

openwong2kim:
I've pushed three commits to this branch rather than leave you a list to work through — `maintainer_can_modify` was on, and it seemed friendlier to show the shape I had in mind than to describe it. Revert any of it freely; it's your PR.

**`d098e38a` — interpolation corrupting values that contain `$`.** This one is ours, not yours: `t()` has always substituted with `str.replace(re, String(v))`, and a replacement *string* reads `$&`, `$$`, `` $` `` and `$'` as patterns. It was unreachable while every interpolated value was a count or a workspace name. Your PR is what makes it reachable — `error.crashed` takes a raw `Error.message`, `openChannelFor` takes an agent-authored title. Passing a replacer function fixes it. There's a test that fails all three ways against the old line.

**`c899427a` — the approval sentences.** The `sameWsIntro` / `sameWsMid` / `sameWsEnd` split can't be translated: the word order lives in the JSX, and verb-final languages can't fill those slots in English order. Your own zh values show the cost — to fit the fixed order they became `智能体在` / `想用` / `在同一工作区中生成…`, which ends up saying "this workspace" and "the same workspace" both. Each prompt is one key again with the emphasised parts as `{slots}`, filled by a small `renderSentence()` helper, so a translator can move them.

Same commit restores the English on `approval.fanoutCount`: `'{count} isolated task(s)'` had replaced `task{count === 1 ? '' : 's'}`, so English read "1 isolated task(s)". Now singular/plural keys, matching the `removedReverted` / `removedRevertedPlural` pattern you used on #911.

**`fe86e3b0` — open dialogs not following a locale change.** The two approval dialogs and `ErrorBoundary` used the module-level `t`, so a language switch didn't re-render them. Those are the three surfaces a user can least dismiss and reopen. Dialogs take `useT()`; `ErrorBoundary` is a class so it subscribes directly. Also renamed `toasts.map((t) => …)`, which shadowed the translator you'd added just above it.

One thing I deliberately left alone: `deck.fleetShell` is still `'shell'` in `zh.ts`. Whether that reads better as `shell`, `终端` or `命令行` is your call, not mine.

`tsc` clean, 3288 renderer tests pass. I also updated `changelog.d/913.md` — it hadn't mentioned the multiview close button, the relative-time localisation, or the orchestrator model chip.

If this all looks right to you, say so and I'll merge. And separately — #912 looks fully subsumed by this branch; I left the numbers over there.

cnxiekun:
All three commits look right to me — the `$` interpolation fix was a real catch (error.crashed and openChannelFor made it reachable), the whole-sentence approval keys with renderSentence are clearly better than my split fragments, and following locale changes on open dialogs is the right behavior.

I have set deck.fleetShell to 终端 in zh.ts (pushed as 69e62d86).

This all looks good — please go ahead and merge. And agree on #912: it is fully subsumed here, happy for it to be closed.

PR #914, opened 2026-08-15
A subagent returning is not the pane finishing

openwong2kim:
`agent.subagent_stop` mapped to status `complete`, and that status flows straight through to the pane's `agentStatus` — `DaemonNotificationRouter` does `agentStatus: ev.status`. So every time a nested subagent returned, the whole pane was marked done while its main turn was still running.

## What that reaches

The pane's status is not cosmetic here; it is what the orchestration layer reasons over:

| | reads it as |
|---|---|
| `deck.rpc.ts:139` | outstanding work = `running \|\| awaiting_input` — a pane flipped to `complete` stops counting as outstanding, so a fan-out can finalize while an agent is mid-turn |
| `deckBriefing.ts:283` | `complete` after not-complete is reported to the brain as **finished** |
| `deck.handler.ts:195` | counted as `stopped` |
| `CommanderEventCoalescer` | drives wake/surface decisions off `complete` in four places |

So the symptom is not a wrong dot in the UI. The fleet reads the pane as free, the briefing tells the orchestrator that pane finished, and its next wake lands on an agent that never stopped.

## Same family as #733, opposite direction

There, a deleted guard blocked the only path that cleared a stuck status. Here, a nested event clears it too eagerly. Both end with the pane's status describing something other than the pane.

## The rest of the system already knew

Two neighbouring paths had this right and the status mapping just never got the same treatment:

- `agentLiveness.ts:100-102` special-cases it explicitly — `if (subagentStop) state = 'busy'` sits *above* the `complete → idle` branch, so the phone header was already correct.
- `driveApprovals` deliberately does not expire a pending prompt on `subagent_stop`, with a comment saying a subagent finishing says nothing about the main agent's question.

## The change

`subagent_stop` reports `running` instead of `complete`. Not dropped: the pane IS still working, and saying so is more accurate than saying nothing. The event still reaches the bus either way, so a consumer that genuinely wants subagent boundaries can still see them — only what the pane's own status is allowed to claim changes.

## Tests

The existing test asserted the old value, so it was pinning the bug; it now pins the new shape. A second test pins the **property** rather than the value: whatever `subagent_stop` reports, it must not be the same status `agent.stop` produces. Restoring symmetry between the two ("both are stops, right?") is how this comes back, and every other test stays green when it does.

## Context

Found while surveying hook insertion points for the remaining agents. Notably GitHub Copilot CLI exposes `agentStop` and `subagentStop` as separate events too, so this distinction is not a Claude Code quirk — any adapter added later inherits the same hazard, which is why the fix belongs in the ingest layer rather than per-adapter.

## Summary by CodeRabbit

- **Bug Fixes**
  - Nested subagent completion no longer incorrectly marks the parent pane as finished while its main task is still running.
  - Parent panes now continue to display a running status after a subagent finishes.

- **Tests**
  - Added regression coverage to verify the corrected pane status behavior.

### DIFF
--- a/changelog.d/914.md
+++ b/changelog.d/914.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **A pane no longer looks finished when one of its subagents returns.** When an
+  agent spawned a nested subagent, that subagent finishing marked the whole pane
+  as done — while its main turn was still running. The pane dropped out of the
+  count of busy work, the fleet read it as free, and the orchestrator was told
+  that pane had finished, so its next instruction could land on an agent that
+  never stopped. A subagent returning now leaves the pane reported as running,
+  which is what it actually is. (#914)
--- a/src/daemon/hooks/HookIngest.ts
+++ b/src/daemon/hooks/HookIngest.ts
@@ -433,12 +433,26 @@ function isMetadataKind(kind: AgentSignalKind): kind is
  */
 function eventShapeFor(
   kind: 'agent.stop' | 'agent.subagent_stop' | 'agent.awaiting_input',
-): { status: 'complete' | 'awaiting_input'; message: string } {
+): { status: 'complete' | 'awaiting_input' | 'running'; message: string } {
   switch (kind) {
     case 'agent.stop':
       return { status: 'complete', message: 'Task finished' };
     case 'agent.subagent_stop':
-      return { status: 'complete', message: 'Subagent finished' };
+      // A nested subagent returning is NOT this pane finishing. It said
+      // 'complete', which flows straight through to the pane's agentStatus
+      // (DaemonNotificationRouter: `agentStatus: ev.status`), so every
+      // subagent return marked the whole pane done while its main turn was
+      // still running — the pane drops out of "who is busy", the fleet reads
+      // it as free, and the orchestrator's next wake lands on an agent that
+      // never stopped. Same family as #733's stuck status, in the opposite
+      // direction: there a deleted guard blocked the only clear, here a
+      // nested event clears too eagerly.
+      //
+      // It reports 'running' rather than being dropped, because the pane IS
+      // still working and saying so is more accurate than saying nothing —
+      // and the event itself still reaches the bus, so a poller that wants
+      // subagent boundaries can still see them.
+      return { status: 'running', message: 'Subagent finished' };
     case 'agent.awaiting_input':
       return { status: 'awaiting_input', message: 'Awaiting input' };
   }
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR fixes the status mapping for `agent.subagent_stop` in `HookIngest.ts`: a nested subagent returning previously mapped to `complete`, which flows straight through to the pane's `agentStatus`, so every subagent return marked the whole pane done while its main turn was still running. The change makes `subagent_stop` report `running` instead — the event still reaches the bus, so consumers that want subagent boundaries can still see them — and updates the existing test plus adds a property-based regression guard asserting `subagent_stop` can never produce the same status as `agent.stop`. The blast radius is real beyond the UI: `deck.rpc.ts` stops counting a `complete` pane as outstanding work, the briefing reports it to the orchestrator as finished, and the coalescer drives wake decisions off it, so this is an orchestration-correctness fix, not a cosmetic one. Scope is tight (3 files, +58/−3), CI is green across all jobs, and the changelog entry is in place — I'd like to see this merged ahead of further hook-insertion work since any future agent adapter inherits the same hazard.

openwong2kim:
Closing: #907 already landed this, and the change here cannot run.

`handle()` returns early for `agent.subagent_stop` at `HookIngest.ts:824`, broadcasting `status: 'running'` / `decision: 'internal'` directly. `eventShapeFor` is called once, at line 850 — after that return. So the `subagent_stop` case in `eventShapeFor` is dead, and flipping it from `complete` to `running` changes nothing at runtime.

That also means the new test does not pin what the commit message says it pins. `ingest.handle(subagent_stop)` takes the early return, so the `'running'` it asserts comes from #907's hardcoded broadcast, not from the mapping this PR edits — put `'complete'` back in `eventShapeFor` and the file stays green.

The diagnosis was right, and the write-up of where the status reaches (deck outstanding-work, the briefing's finished-transition, the coalescer's wake decisions) is the reason #907 handled it. Keeping the changelog fragment would double-claim it in the release notes, which matters with twelve fragments already queued.

Two things worth salvaging separately, neither of which needs this branch:

- The definition is in two places now. `eventShapeFor` still carries a `subagent_stop` case that nothing reaches, and the live one at :824 hardcodes both status and decision. Whoever next "fixes the mapping" will edit the dead one and see no effect — which is exactly what happened here.
- `agentLiveness.ts:100` still says `subagent_stop` "also arrives as `status:'complete'`". That is now false, and the defensive branch under it is unreachable for that reason. A future reader taking that comment at face value is a real hazard.

Happy to do that cleanup as its own change, with the changelog marked n/a since neither is user-facing.

PR #915, opened 2026-08-15
Measure that wmux's hooks are harmless instead of claiming it

openwong2kim:
wmux installs observation hooks into other people's agents. #898 is what that costs when one of them says something the host reads as an instruction: the permission gate emitted `ask` on every "wmux has no opinion" path, `ask` is not neutral — it forces a prompt and overrides the session's permission mode — so a `bypassPermissions` session started asking permission to run `Read`, and the documented escape hatch (`WMUX_GATE=0`) emitted `ask` too. Nothing in the tree could have caught it, because "observation hooks are harmless" was an assumption rather than a measurement.

This makes it a measurement, and it is the gate the next agent adapters have to pass before they ship.

## The claim under test

> Installing a wmux hook must be indistinguishable, to the host, from installing a hook that provably has no opinion.

Every case runs twice — a no-op control (drain stdin, exit 0, write nothing) and the real hook, invoked byte-for-byte the way the integration's own manifest invokes it — and the two are compared on what the host would actually observe.

| Criterion | How it is measured |
|---|---|
| Approval/denial set identical | The host's decision is a pure function of (exit code, stdout). Each host's documented contract is pinned in `HOST_CONTRACTS` with its source. An observation hook must classify as `none` and write **zero bytes** — the direct #898 pin. |
| Completion and duration | The hook's own wall clock, capped at 1.5s over the control. The cap sits deliberately *below* the bridges' 2s transport cap, so a hook that waits out a dead daemon on the host's thread cannot pass. |
| No process outlives the hook | `close`, not `exit`. The host holds the hook's stdout/stderr and waits on them, so a descendant that inherits those pipes shows up as a gap between the two — no process-tree walking, and it is exactly the survivor that would matter to the host. |

Five environment scenarios, because the same hook has to be silent in all of them: wmux not installed; installed with the daemon down; a daemon that answers and has no verdict; the gate disabled; headless.

## What keeps it from becoming a green no-op

Fixtures that violate exactly one criterion each — a hook that emits `ask`, one that never returns, one that exits cleanly but leaves a process behind — must all be rejected.

And on every run the **real bridge** is copied with #898 put back into it, and the gate must catch it emitting `ask` in the two states where #898 actually bit. That check is not decoration. The first draft of this harness passed a bridge with #898 reintroduced: its sandbox had no auth token, so every bridge exited at `no-auth-token` long before reaching a decision at all. The scenarios were fixed, and this keeps them fixed — if the neutral-decision guard is ever refactored away from its current shape, the check fails loudly and asks to be re-derived rather than quietly defending nothing.

## Coverage that survives the next adapter

The matrix is derived from the shipped `hooks.json` files, and `integrations/` is **scanned** rather than listed. A hook added to a manifest is covered the moment it is added; an integration with neither a manifest nor a stated reason fails the run. That is the same property `HookIngest` needs from its `subagent_stop` filter — safety has to live where every adapter passes through, or the next adapter is unprotected by default.

opencode is not in the subprocess matrix because it is an in-process plugin, so it gets a different measurement rather than a silent skip: driven against a stand-in daemon that accepts and never answers, its event handler must still return in well under the 2s transport cap. That is the plugin's own detachment claim, and against an *absent* endpoint it would have gone untested.

## The live half

`scripts/hook-ab-live.mjs` runs the same task twice through the real agent CLI. Opt-in — it needs credentials and spends tokens — and it reports what it cannot establish instead of passing quietly:

- a headless `claude -p` never reaches the permission gate, because the agent sets `CLAUDE_CODE_ENTRYPOINT` itself when it spawns a hook, so the parent cannot present the run as interactive. Measured, after a flag built on the opposite assumption let a #898 mutant straight through;
- `--settings` merges with the user's config, so on a machine where the wmux plugin is already installed the control arm is not hook-free. It says so and exits 2 ("did not establish the claim") rather than 0.

## Measured

- CI gate: 15 tests, ~15s, every hook silent in every scenario, added latency ~110ms over control, survivor gap 0ms.
- Mutation check: the real bridge with #898 reintroduced fails in both affected scenarios; restored, it passes.
- Live A/B against Claude Code: completion equal, `permission_denials` equal, duration within 1%, injected hooks confirmed to reach the stand-in daemon (`outcome: ok`).

## Notes

- Test-only; no runtime code changes and no changelog fragment (same as #904).
- One thing found and left alone as out of scope: a >1MB `PostToolUse` payload is truncated past `MAX_STDIN_BYTES`, the JSON no longer parses, and the signal is dropped entirely. That is the documented, intended behaviour and a delivery question, not a harmlessness one — but it means a turn that read a large file loses its activity stamp.

Part of the step-3 moat work, after #903 (`pane_list` contract + blocking poll) and #914 (subagent status). Unblocks the copilot and kiro adapters.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive runtime checks to ensure integrations remain silent, neutral, responsive, and clean up processes correctly.
  * Added coverage for hangs, blocking prompts, orphaned processes, malformed events, oversized input, and historical regressions.
  * Added validation across supported hook integrations and representative daemon scenarios.

* **New Features**
  * Added an opt-in live comparison tool to measure agent behavior with and without hooks, including timing, failures, and hook activity.

### DIFF
--- a/scripts/__tests__/hookHarmlessness.runtime.test.mjs
+++ b/scripts/__tests__/hookHarmlessness.runtime.test.mjs
@@ -1,0 +1,394 @@
+// P-5 — the A/B harmlessness gate, wired into CI.
+//
+// wmux installs hooks into other people's agents. #898 is what it costs when
+// one of them says something the host reads as an instruction: the permission
+// gate emitted `ask` on every "wmux has no opinion" path, `ask` is not
+// neutral, and a bypassPermissions session started asking permission to run
+// `Read` — with the documented escape hatch emitting `ask` too. The claim this
+// file defends is narrow and testable:
+//
+//   installing a wmux hook is indistinguishable, to the host, from installing
+//   a hook that provably has no opinion.
+//
+// Every case runs a CONTROL (a no-op hook) and a TREATMENT (the real hook,
+// invoked the way the integration's own manifest invokes it) and compares what
+// the host would observe. The measurement lives in
+// scripts/lib/hookHarmlessness.mjs; the pass criteria live here.
+//
+// Two properties keep this from decaying into a green no-op:
+//   - the matrix is DERIVED from hooks.json, so a hook added to a manifest is
+//     covered the moment it is added;
+//   - the gate is re-proven against the real bridge on every run, by putting
+//     #898 back into a copy of it and requiring rejection.
+// The second one is not theoretical: the first draft of this harness passed a
+// bridge with #898 reintroduced, because its sandbox had no auth token and
+// every bridge bailed before reaching the decision path at all.
+
+import { describe, it, expect, beforeAll, afterAll } from 'vitest';
+import { rmSync, readdirSync, readFileSync } from 'node:fs';
+import { join } from 'node:path';
+
+import {
+  REPO_ROOT,
+  buildCases,
+  discoverHookManifests,
+  NON_MANIFEST_INTEGRATIONS,
+  setupScenarios,
+  makeSandbox,
+  writeFixtureHooks,
+  writeMutatedBridge,
+  runHookCase,
+  classifyDecision,
+} from '../lib/hookHarmlessness.mjs';
+
+// Kill budget. Well above any honest hook; a hang lands here.
+const KILL_BUDGET_MS = 15_000;
+// How much wall clock a wmux hook may add over the no-op control. This is the
+// live A/B's "duration within ±50%" criterion expressed against a simulated
+// host: there the baseline is the agent's task, here it is a hook that does
+// nothing. The band sits deliberately BELOW the bridges' own 2s transport cap
+// (HOOK_TIMEOUT_MS) — a hook that waits out a dead daemon on the host's thread
+// is the stall this criterion forbids, so a budget above 2s could not fail it.
+// Measured headroom: every hook lands within ~110ms of the control locally.
+// (Review: Grok, P1.)
+const MAX_ADDED_MS = 1_500;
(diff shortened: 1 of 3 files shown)


openwong2kim:

This PR turns "wmux's observation hooks are harmless" from an assumption into a measurement: every hook runs as a control/treatment A/B (a no-op control vs. the real hook invoked exactly as the shipped manifest invokes it) and the two are compared on what the host would actually observe. It is test/tooling-only — three new files under `scripts/` (the harness lib, the Vitest runtime gate, and the opt-in live A/B driver) — with no runtime code changes, matching #904's precedent of no changelog fragment. The design covers the failure modes that matter: cases are derived from the shipped `hooks.json` manifests so a new hook is covered the moment it lands, `integrations/` is scanned so an unaccounted adapter fails the run, and every run re-proves the gate by putting #898 back into a copy of the real bridge and requiring rejection. The pass criteria — byte-empty stdout on every "no opinion" path, added latency capped at 1.5s (deliberately below the bridges' 2s transport cap), and no descendant surviving the hook while holding the host's stdio — map directly to the three ways a hook can hurt its host, and the oversized-stdin case checks the host's write settles rather than wedges. I also appreciate the honest edges: the wedged-daemon scenario is excluded from the subprocess matrix because the permission gate blocking on a pending approval is the feature working, and the live driver exits 2 ("did not establish the claim") rather than passing quietly when the control arm is polluted. Since this is the gate the copilot and kimi adapters must pass before they ship, I'm tagging it P2 and want it landed ahead of that adapter work.

PR #886, opened 2026-08-14
fix(i18n): add missing prefix-action labels to en and zh

cnxiekun:
## What

Three prefix-action labels were missing from both `en.ts` and `zh.ts`, so the Settings → Shortcuts → Prefix bindings rows rendered the raw key string (`settings.prefix.renameWorkspace` / `killWorkspace` / `showCheatSheet`).
Most other locales already ship `killWorkspace` and `showCheatSheet` translations (e.g. zh-TW), so this was a gap specific to `en` + `zh`.

Added:
- `settings.prefix.renameWorkspace` → en `Rename workspace` · zh `重命名工作区`
- `settings.prefix.killWorkspace` → en `Kill workspace` · zh `结束工作区`
- `settings.prefix.showCheatSheet` → en `Show cheat sheet` · zh `显示快捷键速查表`
  
## Verification
  
- Key parity + `{var}` check: en/zh both 1206 keys, 0 missing, 0 extra, 0 mismatch
- `npx tsc --noEmit` passes
- `eslint` clean on both files
- i18n contract tests pass (51/51)

## Follow-ups found while dogfooding the zh UI (not in this PR)

Two settings sections hardcode English strings in the components and never go through i18n — so they show English in every locale, not just zh. Both are the  same straightforward fix: move each literal into a locale key, replace the literal with a `t(...)` call (no component logic changes), and add the keys to  `en.ts` + `zh.ts` (other locales catch up as they will).
  
### 1. Settings → Claude integration → Accounts (`AccountsSection.tsx`)

| proposed key | EN (current) | ZH |
|---|---|---|
| `settings.accounts.title` | Accounts | 账户 |
| `settings.accounts.empty` | No accounts yet. Add one to bind different subscriptions per workspace. | 还没有账户。添加一个，以便按工作区绑定不同的订阅。 |
| `settings.accounts.add` | + Add account | + 添加账户 |
| `settings.accounts.remove` | Remove | 移除 |
| `settings.accounts.cancel` | Cancel | 取消 |
| `settings.accounts.unregister` | Unregister (does not delete the directory) | 注销（不会删除目录） |
  
### 2. Settings → Agents → MCP Servers (`SettingsPanel.tsx`)
  
| proposed key | EN (current) | ZH |
|---|---|---|
| `settings.mcpServers.title` | MCP Servers | MCP 服务器 |
| `settings.mcpServers.notDetected` | Not detected | 未检测到 |
| `settings.mcpServers.modified` | · modified {date} | · 修改于 {date} |
[...]

PR #917, opened 2026-08-16
chore(release): 3.43.0

openwong2kim:
Folds the fourteen pending changelog fragments, bumps the version in
package.json and both package-lock sites, and regenerates
docs/api/reference.md so its baked-in version matches (the CI drift guard
checks this).

Minor rather than patch: the release carries new capabilities, not only
fixes — fan-out orchestration, a blocking events poll, and a glyph rail
for the collapsed deck.

Contents: the orchestrator can fan work out into isolated worktrees, one
call spawning N tasks each on its own branch, workspace and pane, with
per-task roles that bind to whichever agent and model you chose in
Settings — still gated behind an approval that spells out what each role
resolves to. `wmux_events_poll` can block for an event instead of being
polled on a timer, narrowed to one pane and one kind of signal.
Collapsing the deck now leaves a rail of its icons rather than nothing.

The fixes: a Claude Code permission gate that emitted `ask` on every
"wmux has no opinion" path — including its own documented escape hatch —
so a bypassPermissions session was asked to approve a Read; wmux now
detects a plugin install still running that hook and says so. The
completion alarm no longer fires on a turn that has not ended. An update
that finished downloading now announces itself instead of waiting to be
asked. The terminal reports the real Windows build number.

## Summary by CodeRabbit

* **New Features**
  * Added configurable event polling with blocking waits and filtering.
  * Added clearer update notifications, version details, one-click installation, and installation error messages.
  * Added localized session-closing warnings across supported languages.

* **Bug Fixes**
  * Improved completion alarms to account for active background tasks and verified turn completion.
  * Improved reliability of update state preservation and completion signaling.

* **Documentation**
  * Updated the API reference and application version to 3.43.0.


openwong2kim:

**P3** — This is the 3.43.0 release chore: it folds the fourteen pending `changelog.d/` fragments (#886–#911) into `CHANGELOG.md`, bumps the version in `package.json` and both `package-lock.json` sites, and regenerates `docs/api/reference.md` so its baked-in version matches what the CI drift guard expects. Minor rather than patch is justified — the release carries new capabilities (fan-out orchestration, the blocking `wmux_events_poll`, the glyph rail for the collapsed deck) alongside the fixes.

It touches no runtime code: eighteen files, and everything outside the changelog/version/reference regeneration is fragment deletion, so the merge itself is the release switch. One thing worth a pre-merge glance since CI runs on the merge commit: confirm the lockfile's two version sites and `docs/api/reference.md` are the regenerated artifacts rather than hand-edited, because the drift guard is the only thing that keeps `reference.md` honest about what actually shipped.

PR #918, opened 2026-08-16
Report Kiro turn boundaries from a hook instead of guessing

openwong2kim:
wmux answers "is this pane done?" for Kiro by scraping the terminal — a two-stage regex gate on Kiro's chrome and prompt lines (`AgentDetector.ts:329`). That holds until the model prints something prompt-shaped, or Kiro changes its UI. The answer feeds notifications, `deckBriefing`, the stop gate, the orchestrator's count of what is still busy, and `wmux_events_poll(blockMs)` — so when it is wrong, the fleet sends work to a pane that never stopped. That is the failure mode #914 just fixed for subagents.

This lets Kiro say it itself.

| Kiro trigger | wmux signal |
|---|---|
| `stop` | `agent.stop` |
| `agentSpawn` | `agent.session_start` |

## Measured on kiro-cli 2.15.1, not assumed

Running it corrected three entries of the earlier hook survey:

| Assumed | Measured |
|---|---|
| `session_id` is in the payload | **Absent.** `stop` is `{hook_event_name, cwd, assistant_response}` |
| `~/.kiro/hooks/` applies globally | **Inert** — 0 invocations, both plausible layouts |
| copilot should be first | It has no CLI here; Kiro is the one that can be verified at all |

Also measured: Kiro's neutral is no output and exit 0, the same contract as Claude Code's. A *failing* hook does not block a turn — stderr plus exit 2 still let the agent answer normally — though this bridge does neither.

Consequences, all deliberate:

- **No resume binding.** No session id means nothing to bind, so unlike the Codex bridge there is no spool. A lost signal is re-derived by the detector.
- **Pane attribution is `WMUX_PTY_ID` only**, inherited from the pane environment (verified: the hook process does inherit it). A payload without it is **dropped**, not attached to a guess.
- **`preToolUse`/`postToolUse` unregistered.** An activity stamp would cost a process spawn per tool call — the one path that actually makes anything heavier. The Claude bridge needed a 2.5s source-side throttle for exactly this.
- **`userPromptSubmit` unmapped.** Kiro has no approval-specific event, and calling "a prompt was submitted" `awaiting_input` is the conflation #898 punished. No signal beats a false one.

## Privacy

Kiro hands the hook the user's whole prompt and the model's whole reply. wmux's bridges are metadata-only, so the envelope builder reads `hook_event_name` and `cwd` and nothing else. The test asserts on the **serialized envelope** rather than field by field, so a future field that happens to carry content fails it too.

## The agent config, and why it looks like that

Kiro's hooks can only live inside an agent config, and the user's default agent is built in — there is no file of theirs to add a hook to. So wmux ships its own agent and panes opt in with `--agent wmux`. The user's default stays their default; `kiro-cli agent list` confirms it (`wmux` appears under `Global`, the `*` stays on `kiro_default`).

The rule is **equivalent to `kiro_default` except the prompt**, checked against a materialized built-in:

- `allowedTools` is `[]` in the built-in too, which is why **approval behaviour cannot change** — anything omitted is not more restrictive than the default.
- `includeMcpJson: true` and `resources` are carried over. An earlier revision dropped both, which would have cost the pane its MCP access (including wmux's own tools) and its project context — silently, since the agent keeps answering either way. The measurement that cleared the empty prompt ran a synthetic task in a temp directory, where neither could have shown up. That miss is now pinned by tests.
- The empty prompt is the one intended difference. Measured 3× on the same tool-using task: thin and built-in both 3/3 correct, 17.4s vs 18.0s, thin slightly cheaper. Cloning the built-in's 1.7KB prompt would freeze it at install time and drift.

`scripts/kiro-agent-equivalence.mjs` re-checks that against the live built-in after a Kiro upgrade — no CI can, since it needs a real kiro-cli. Current output:

```
EQUIVALENT — differs only in: name, description, prompt, hooks
```

## Installation is NOT wired up

The end-to-end path (global agent + `--agent wmux` + hook fires + daemon receives) needs a Kiro account to verify, and this machine has none. Three of its four links are verified live — hooks fire from an agent config, the global agent is discovered and validated, the default agent is untouched — but their composition is not. So the bridge ships with a manual setup guide instead of an install nobody has watched work; wiring it into `McpRegistrar`/`lifecycleIntegrations` is a small follow-up once someone can dogfood it.

If the composition turns out not to work, the cost is a silent no-op: Kiro panes keep using the detector, exactly as today.

## Harmlessness

Covered by the P-5 gate (#915) across five environment scenarios: zero bytes to stdout **and stderr**, always exit 0, ~90ms, no process left holding the host's stdio.

The gate also caught this integration on its own. Creating `integrations/kiro/` failed the coverage test before any wiring existed:

```
new integration(s) with no hook manifest and no stated reason — wire them into
the harmlessness gate, or record why: expected [ 'kiro' ] to deeply equal []
```

## Notes

- No shebang on the bridge, deliberately: Kiro always invokes it as `node "<path>"`, and Vitest cannot parse a `.mjs` that starts with one — which would leave the pure envelope builder testable only through a subprocess. The OpenCode plugin is shebang-free for the same reason.
- tsc clean · lint clean · 23 new tests · harmlessness gate 18 tests.

## Summary by CodeRabbit

* **New Features**
  * Added Kiro CLI integration with lifecycle signal support for agent stop and spawn events.
  * Added automatic Kiro agent configuration with MCP access, project resources, and supported hooks.
  * Added secure event handling that forwards metadata only and avoids transmitting conversation content.
  * Added configuration equivalence validation tooling and setup guidance.

* **Tests**
  * Added comprehensive coverage for configuration, event handling, endpoint fallback, and hook harmlessness.

### DIFF
--- a/integrations/kiro/README.md
+++ b/integrations/kiro/README.md
@@ -1,0 +1,99 @@
+# wmux ↔ Kiro CLI
+
+Turns "is this Kiro pane done?" from a guess into a fact.
+
+Without a hook, wmux infers a Kiro turn boundary by scraping the terminal
+(`AgentDetector`, a two-stage regex gate on Kiro's chrome and prompt lines).
+That works until the model prints something prompt-shaped, or Kiro changes its
+UI. With the hook, Kiro reports the turn ending itself, and wmux's turn-end
+consumers — notifications, the orchestrator's "what's still busy" count,
+`wmux_events_poll(blockMs)` — get a fact instead of an inference.
+
+**Not wired into wmux's installer yet.** The bridge, its tests and its
+harmlessness coverage ship here; automatic installation is deliberately held
+back until the end-to-end path can be verified on a machine with a Kiro
+account. Set it up by hand today with the steps below.
+
+## What it reports
+
+| Kiro trigger | wmux signal |
+|---|---|
+| `stop` | `agent.stop` — the turn ended |
+| `agentSpawn` | `agent.session_start` |
+
+Deliberately **not** registered:
+
+- `preToolUse` / `postToolUse` — an activity stamp would cost a process spawn
+  per tool call. That is the one path that actually makes things heavier.
+- `userPromptSubmit` → `agent.awaiting_input` — Kiro has no approval-specific
+  event, and "a prompt was submitted" is not "a human is being waited on".
+  Conflating those is what issue #898 punished; no signal beats a false one.
+
+## What it cannot do (measured, not assumed)
+
+- **No resume binding.** Kiro's `stop` payload is
+  `{hook_event_name, cwd, assistant_response}` — there is no session id, so
+  there is nothing to bind a resumable session to.
+- **Pane attribution comes only from `WMUX_PTY_ID`**, inherited from the pane
+  environment. A payload without it is dropped rather than attached to a guess.
+- **Only panes launched with `--agent wmux`.** A hand-run `kiro-cli chat` keeps
+  using the detector, exactly as before.
+
+## Privacy
+
+Kiro's payloads carry content: `prompt` is the user's whole input and
+`assistant_response` is the model's whole reply. wmux's bridges are
+metadata-only, so the envelope builder reads `hook_event_name` and `cwd` and
+nothing else. The test asserts on the serialized envelope, so a future field
+that happens to carry content fails it too.
+
+## Manual setup
+
+1. Put the bridge somewhere stable, e.g. `~/.wmux/hooks/wmux-kiro-bridge.mjs`.
+2. Write `~/.kiro/agents/wmux.json` with the config
+   `agent/wmuxAgent.mjs` builds, pointing `command` at that path:
+
+   ```json
+   {
+     "name": "wmux",
+     "description": "wmux-managed: kiro-lifecycle-agent",
+     "prompt": "",
+     "tools": ["*"],
+     "includeMcpJson": true,
+     "resources": ["file://AmazonQ.md", "file://AGENTS.md", "file://README.md",
+                   "skill://.kiro/skills/*/SKILL.md",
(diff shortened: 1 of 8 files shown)


openwong2kim:

**P2** — This PR replaces the regex-based Kiro turn-end detection (the two-stage chrome/prompt gate in `AgentDetector`) with a hook bridge that lets Kiro report its own turn boundaries: `stop` → `agent.stop` and `agentSpawn` → `agent.session_start`. It adds a new `integrations/kiro/` tree (the self-contained `wmux-kiro-bridge.mjs`, a `wmuxAgent.mjs` config builder, tests, and a setup README) plus `scripts/kiro-agent-equivalence.mjs` to detect drift against the live `kiro_default` after Kiro upgrades, and wires kiro into the harmlessness gate.

This matters because every turn-end consumer — notifications, the orchestrator's busy count, the stop gate, `wmux_events_poll(blockMs)` — currently trusts an inference that breaks the moment the model prints something prompt-shaped or Kiro ships a UI change; a measured signal eliminates the "fleet sends work to a pane that never stopped" failure mode that #914 just fixed for subagents. The design choices are the right ones: a metadata-only envelope (the serialized-envelope test keeps `prompt`/`assistant_response` out), dropping unattributable signals rather than guessing a pane, and deliberately not mapping `userPromptSubmit` to `awaiting_input` — exactly the conflation #998/#898 punished.

Two things to keep visible during review: the end-to-end install path is intentionally not wired up yet (needs dogfood on a machine with a Kiro account, so this ships inert and harmless), and the equivalence of the `wmux` agent against `kiro_default` is only checked on demand since CI cannot run a real kiro-cli — worth noting the follow-up for `McpRegistrar`/`lifecycleIntegrations` wiring once someone can verify the composition.

PR #923, opened 2026-08-16
Fix mouse events not reaching TUIs on Windows 10 by spawning the bundled ConPTY

### LINKED ISSUE #910: wmux 3.42.0: mouse events (click/scroll) not forwarded to terminal apps on Windows 10 (works on Windows 11) — identical version & config
### What happened?

Mouse events (click + wheel) are not forwarded to terminal programs in wmux on Windows 10, but work fine on Windows 11 with the same wmux version and identical config. In vim, the wheel scrolls the terminal's own scrollback instead of vi, and clicking doesn't move the cursor.

The failing machine is a lower-spec Windows 10 box I use for remote SSH/attach — not planned for a Windows 11 upgrade, so a wmux-side fix on Windows 10 would be much appreciated.

### Steps to reproduce

1. Open a Git Bash pane in wmux (Windows 10 22H2).
2. Open vim (with `set mouse=a`, `ttymouse=sgr`, TERM=xterm-256color).
3. Click in vim (cursor does not move); scroll the wheel (scrolls the terminal's own scrollback, not vi).
4. The SAME vim in plain Git Bash (mintty) on the same machine works fine.
5. Reset `~/.wmux` config (moved aside, rebuilt fresh) — no change.

Environment:
| | Main machine (works) | New computer (fails) |
|---|---|---|
| OS | Windows 11 (26200) | Windows 10 22H2 (19045) |
| wmux | 3.42.0 (latest) | 3.42.0 (latest) |
| GPU | — | AMD Radeon RX Vega 11 |
| vim | `set mouse=a`, `ttymouse=sgr`, TERM=xterm-256color | identical (same binary, same .vimrc) |

### Expected behavior

Mouse clicks and wheel should be forwarded to the terminal app (vim) like on Windows 11 — clicking moves the cursor, and the wheel scrolls vi instead of the terminal's own scrollback.

### wmux version

3.42.0

### How did you install wmux?

winget

PowerShell 7+ (pwsh)

### Windows version

Windows 10 22H2 (build 19045)

None / not relevant

```shell
HYPOTHESIS: wmux uses ConPTY (node-pty) on Windows; the older ConPTY on Windows 10 may swallow or fail to relay the DECSET mouse-enable sequences (`ESC[?1000h` / `?1002h` / `?1006h`) that vim sends, so xterm.js never receives them and mouse reporting stays off. Mintty works there because it bypasses ConPTY.

Could you confirm whether the PTY layer needs a Windows 10-specific path for relaying mouse-mode sequences? Happy to test a debug build or capture anything you need.
```

### Confirmation

openwong2kim:
Fixes #910

## Root cause

The in-box ConPTY on Windows 10 (and Server 2022) never forwards the
mouse-reporting DECSETs (`?1000` `?1002` `?1006`) to the output pipe —
[microsoft/terminal#15977](https://github.com/microsoft/terminal/issues/15977),
fixed in Windows Terminal 1.20. So `set mouse=a` in vim on Win10
silently does nothing and the wheel scrolls wmux's own scrollback
instead of the vim buffer. Win11 ships a fixed in-box ConPTY; mintty
bypasses ConPTY entirely, which is why both work.

## Change

- **Spawn gate** (`shared/conptyWindows.ts`): below build 22000, ask
  node-pty for its bundled conpty.dll (`useConptyDll: true`) — the
  OpenConsole build inside node-pty's prebuilds has the fix. 22000 is a
  product cut, not the upstream fix build: it covers Server 2022 (20348)
  and leaves early Win11 22000 on its own (working) in-box relay.
  Applied at both spawn sites: `DaemonSessionManager` (daemon mode) and
  `PTYManager` (local mode).
- **Renderer consistency** (`useTerminal.ts`): when the bundled DLL
  drives the PTY, report `windowsPty.buildNumber = 22621` to xterm — a
  capability token, not an OS claim, because reflow behaviour comes
  from OpenConsole. Keeps xterm's 21376 reflow boundary on the correct
  side while the OS is 19045 (guards the #897/#900 reflow fixes).
- **Damaged-install fallback**: if the bundled DLL cannot load (the five
  error shapes node-pty's `LoadConptyDll` throws, incl.
  `Cannot launch conpty` for antivirus-blocked OpenConsole.exe), retry
  once with the in-box backend so the shell still opens. Matching is
  deliberately narrow — a transient error 87 must keep failing so
  PaneSupervisor's restart backoff can absorb it, instead of silently
  demoting that pane to mouse-less ConPTY forever.
- **Logging**: every gated spawn logs which ConPTY backend actually
  started (`bundled conpty.dll` vs fallback), so a still-broken report
  is diagnosable from the daemon log alone.

`conptyWindows.ts` is a new browser-safe pure module — `platform.ts`
reads `process.platform` at module scope, which the sandboxed renderer
cannot evaluate, so importing it from `useTerminal.ts` would have
crashed the renderer bundle.

## Known trade-offs (panel review, accepted)

- The damaged-install fallback leaves xterm on the modern-build
  heuristic while the PTY runs in-box (resize reflow may misbehave) —
  rare packaging failure only, visible in the new backend log.
- node-pty 1.1.0's DLL mode skips conout-worker cleanup on natural
  exit (`_$onProcessExit` no-ops when `_useConptyDll`) — watching for
  it in dogfooding; upstream issue if confirmed.

## Verification

- `shared/__tests__/platform.test.ts`: 18 tests covering the build
  cut (19045/20348/21999 → bundled, 22000+/null → in-box), the 22621
  capability token, and the error-fragment matcher (5 load errors
  match; error 87 does not).
- tsc (app/daemon/cli), eslint, and the full vitest suite pass
  (11,393 tests).

## Dogfood checklist for Win10 19045 (reporter)

1. vim with `set mouse=a` / `ttymouse=sgr` — click moves the cursor,
   wheel scrolls the vim buffer, not wmux scrollback.
2. Resize with wrapped long lines on screen — lines re-wrap cleanly.
3. Exit a shell / restart a supervised pane — pane comes back, mouse
   still works (fallback must NOT fire on transient error 87).
4. Restart the daemon with live sessions — panes reattach, mouse works.
5. Daemon log names the backend on every spawn — absence of the line
   means the gate never fired.

## Summary by CodeRabbit

- **New Features**
  - Improved Windows 10 and Windows Server 2022 terminal support, including mouse input in Vim and other terminal applications.
  - Uses an enhanced ConPTY backend on supported Windows builds for improved compatibility.
  - Records which terminal backend is selected.

- **Bug Fixes**
  - Automatically falls back to the built-in Windows backend if the enhanced backend cannot load.
  - Preserves actionable error messages for unrelated terminal launch failures.
  - Windows 11 behavior remains unchanged.


openwong2kim:

This PR fixes #910, where mouse clicks and wheel events never reached TUIs like vim on Windows 10 while the same setup worked on Windows 11. The root cause is the in-box ConPTY on sub-22000 builds never forwarding the mouse-reporting DECSETs, so the fix spawns those panes against the bundled conpty.dll (`useConptyDll: true`) at both spawn sites — `DaemonSessionManager` and `PTYManager` — behind a shared predicate in the new browser-safe `shared/conptyWindows.ts`. The renderer side reports `buildNumber: 22621` to xterm as a capability token, which keeps the reflow boundary from #897/#900 on the correct side while the OS itself is 19045, and every spawn logs which backend actually started so a still-broken report is diagnosable from the log alone. The fallback on a damaged install is deliberately narrow — only the five DLL-load error shapes demote to in-box, so a transient ConPTY error 87 still fails into PaneSupervisor's restart backoff instead of silently disabling the mouse fix for that pane forever. Given how many users run Win10 boxes for remote-attach workflows, triaged as **P2**.

openwong2kim:
**Hardware verification of the DLL path:** [run 31947780090](https://github.com/openwong2kim/wmux/actions/runs/31947780090) executes the bundled-ConPTY path on a `windows-2022` runner — build **20348**, which is below the 22000 gate, so the gate actually fires there (unlike `windows-latest`, which is build 26100+ and never enters this branch).

Verified on real Windows hardware:

- Runner build recorded as 20348, `shouldUseBundledConpty('win32', 20348)` → `gate=true`
- `pty.spawn('powershell.exe', …, { useConptyDll: true })` — two spawns at different row counts, both produced output and exited 0
- The gate's unit tests pass on the same pre-22000 runner

This proves the bundled conpty.dll loads and drives a real shell on pre-Win11 hardware. What it cannot prove (no TUI input injection in CI): that mouse events reach vim on the reporter's exact Win10 22H2 desktop — that remains the reporter checklist on #910.

The workflow lives on scratch branch `verify/conpty-910` (never merged; deleted after #910 verifies).

openwong2kim:
**Root-cause A/B proof on real Windows hardware:** [run 31948095449](https://github.com/openwong2kim/wmux/actions/runs/31948095449) probes the exact mechanism #910 is about — whether the mouse-mode DECSETs (`?1000h ?1002h ?1006h`, what vim writes for `set mouse=a`) survive the pty round trip on a `windows-2022` runner (build 20348, below the gate):

```
dll:   relayed=true   ← bundled conpty.dll (this PR's path on Win10)
inbox: relayed=false  ← the OS in-box ConPTY
```

Two conclusions from one run:

1. **The fix works where it fires** — the bundled DLL relays the DECSETs the in-box drops. Without relay, the terminal never learns mouse tracking is on, which is precisely the reported symptom (wheel scrolls the terminal's own scrollback).
2. **Including Server 2022 (20348) in the gate was correct** — its in-box ConPTY is also broken, so the `buildNumber < 22000` cut catches it.

Combined with the earlier green run (gate fires on 20348; two real spawns via `useConptyDll` exit clean), the fix is verified end-to-end at the mechanism level on pre-Win11 hardware. The one thing CI cannot do is wiggle a physical mouse in vim on the reporter's Win10 22H2 desktop — that final user-level check is the checklist on #910.

Issue #905, opened 2026-08-15
[remote attach] Mirror panes don't adapt to the local viewport size — unused space and truncated lines


cnxiekun:
### Problem / motivation

When attaching to a remote wmux from the desktop app, the mirrored panes keep the remote pane's own grid size and don't fill the local viewport — leaving unused space and truncating long lines.

Concrete case: the remote workspace has two side-by-side panes (Claude Code left, bash right). On the local screen:
- The left pane roughly fills the left half (fine).
- The right bash pane is narrower than the right half — over-long lines are truncated at the remote pane's width, and the rest of the right area is empty.
- There is also unused space at the bottom of the mirrored area.

On the remote machine it looks fine; on the local attach viewport the fixed grid + letterboxing leaves dead space and clips text.

### Proposed solution

The mirrored panes should adapt to the available local viewport — when the local mirror box is bigger than the remote pane's grid, the pane should renegotiate its size (or the remote should resize it) so it fills the space and long lines wrap at the actual usable width.

Suggested direction: `RemoteMirrorTerminal.tsx` never resizes the remote — it applies the remote's cols/rows (`onPaneMeta`/`onPaneResize` → `term.resize(e.cols, e.rows)`) and `runFit` only shrinks the local font to fit that grid into the box (geometry has a single owner, the remote daemon). Letting the mirror renegotiate the pane's geometry with the remote (a resize request when the local box changes) would make the pane match the viewport and wrap long lines at the usable width.

### Alternatives considered

Workaround today: resize the remote pane on the remote machine to roughly match the local viewport — but the local window size differs, so it never lines up. Not practical for a remote-attach workflow.

### Additional context

- wmux desktop app (Windows), attaching to a remote wmux over `wmux web --expose`
- Remote workspace: two side-by-side panes (Claude Code + bash)
- Screenshot to follow in a comment.

### Confirmation

- [x] I searched existing issues and Discussions and this is not a duplicate.

cnxiekun:
<img width="3636" height="1973" alt="Image" src="https://github.com/user-attachments/assets/4109f8c4-2442-41bd-a327-c21c8b2d2d23" />

openwong2kim:
## [wmux-hermes] Triage Summary

**P3 — enhancement** in the remote-attach experience: mirrored panes keep the remote pane's grid size instead of adapting to the local viewport, which letterboxes the mirror area and truncates long lines at the remote pane's width even though the local box is larger.

[...]

PR #924, opened 2026-08-16
Give the attach mirror the terminal's own copy, paste, and newline keys

### LINKED ISSUE #895: [remote attach] Mirror terminals drop the local terminal's input conveniences (copy/paste, Shift+Enter)
### Problem / motivation

Huge praise for the remote attach feature first — I'm on a bare new machine (nothing configured, just wmux installed) remote-attaching to my main machine's wmux, and it genuinely works great. 🔥

But copy/paste is something I use constantly while working, and the attached mirror drops the local terminal's input conveniences. The remote-attach mirror (`RemoteMirrorTerminal`) forwards keystrokes raw to the remote pane and skips the local terminal's input handling:

| Behavior | Local terminal | Attached mirror |
|---|---|---|
| Select text | auto-copies | selection is not copied |
| `Ctrl+C` with a selection | copies the selection | forwarded to the remote as SIGINT (clears the pane's input line) |
| `Ctrl+V` | pastes from clipboard | no effect |
| `Shift+Enter` | inserts a newline | forwarded raw (sent straight to the app) |

### Proposed solution

The attached mirror should handle editing input like the local terminal: auto-copy on selection, `Ctrl+C`-with-selection copies instead of sending SIGINT, `Ctrl+V` pastes from the local clipboard, and `Shift+Enter` produces a newline.

Suggested root cause: `src/renderer/components/Remote/RemoteMirrorTerminal.tsx` creates a bare xterm instance and forwards every keystroke via `onData → paneWrite` without wiring the input handling that `useTerminal.ts` gives local panes (`autoSelectionCopy`, the `Ctrl+C`-copies-selection path, paste, and the deterministic newline keys). Wiring the mirror through the same input path (or the shared pieces) would close the gap.

### Alternatives considered

I habitually hit `Ctrl+D` to split (as in the local terminal), but in the mirror it forwards as EOF and closes the remote pane instead. And since the docs say there's no way to create/split panes from the attach view, I can't even open a fresh pane remotely to work around it — new panes have to be created on the remote machine itself.

### Additional context

- wmux desktop app (Windows), attaching to a remote wmux over `wmux web --expose`
- Local and remote both on recent builds

### Confirmation

openwong2kim:
A mirrored pane forwarded every keystroke raw. That is right for the remote app's keys and wrong for the four that never leave this machine — and #895 is the list of the four, reported by someone using attach as their daily driver from a bare laptop.

| | before | after |
|---|---|---|
| select text | copied nothing | auto-copies, debounced |
| `Ctrl+C` with a selection | SIGINT — cleared the line being read | copies |
| `Ctrl+V` | nothing at all | pastes into the remote pane |
| `Shift+Enter` | submitted | inserts a newline |

None of the four act on remote state. The selection and the clipboard are local, so handling them here makes a mirror behave like a terminal without giving the local window any claim on the remote pane — which is the property that made #905 the harder question and this one the easy one.

## Shape

The chord table is a pure decision function in `mirrorInput.ts`, the same split `mirrorFit.ts` and `newlineKeys.ts` already use, so it is tested without xterm, Electron, or a DOM. The component keeps only the adapter: clipboard, socket, terminal. The pieces the local pane already had extracted — `createAutoSelectionCopy`, `pastePtyChunked`, `resolveNewlineKeyByte`, `copySelectionWithFeedback` — are reused rather than re-derived, so a mirror copies with the same toast and pastes with the same bracketed-paste handling a local pane does.

## Two properties the tests pin as hard as the fixes

- **An empty selection still lets `Ctrl+C` interrupt.** Taking SIGINT away would be a worse bug than the one being fixed, so it has its own test on both platforms.
- **A read-only host receives nothing.** `--allow-input` off means the remote refuses writes anyway; the new direct-write path does not pass through the `onData` gate that already handled this, so the gate is on the one function that can reach `paneWrite`, not only on its callers.

Every chord matches the physical `code` as well as `key`. Under a CJK IME `key` is a composed jamo or the literal `'Process'` — exactly how these same handlers went dead for local panes, and the reason that fix is worth copying rather than rediscovering.

## Deliberately unchanged

- **App shortcuts, zoom, and custom keybindings still reach the remote pane.** A mirror has no local pane to act on, so intercepting them would trade a working remote key for a local no-op.
- **`Ctrl+D` stays EOF.** The reporter mentions habitually reaching for it as the split shortcut; taking it over would silently break EOF on the remote shell and give nothing back, since a pane cannot be created from the attach view either way. Worth its own decision, not a ride-along.
- **Image paste is skipped.** A local pane pastes the temp file's *path*, and that path names a file on this machine and nothing on the other end of an attach. The mirror stays quiet rather than typing a broken path into a live shell.
- **macOS keeps the Cmd chords** — ⌘C / ⌘V copy and paste, `Ctrl+C` stays SIGINT unconditionally, `Ctrl+V` stays readline's quoted-insert — and the mac-only native-paste race guard from `useTerminal.ts` comes along, since ⌘V arrives as an NSMenu key equivalent that `preventDefault()` cannot suppress.

## Testing

- `mirrorInput.test.ts` — 25 cases over the chord table: the four fixes, what must not be taken away (SIGINT, `Ctrl+D`, keyup, an explicit `Ctrl+J` binding), the read-only matrix, the macOS chords, and the IME cases where `key` is a jamo or `'Process'`.
- `RemoteMirrorTerminal.test.tsx` — 7 new cases over the wiring: that a copy never reaches `paneWrite`, that a direct write carries the live attach id, that the debounce holds mid-drag, and that a read-only host receives nothing from either a paste or a newline key while still being able to copy.
- Full renderer suite green (3394 passing). The one unrelated failure locally is `atlasCoherence.test.ts`, which asserts a patch-package patch is present in `node_modules`.

Closes #895

## Summary by CodeRabbit

* **New Features**
  * Mirrored remote terminals now support copying, pasting, and inserting new lines, including macOS keyboard shortcuts.
  * Local selections can be copied automatically.
  * Shift+Enter sends the appropriate newline format for the remote terminal.
  * Bracketed paste and larger pasted input are handled reliably.
  * Keyboard behavior adapts to supported remote keyboard protocols.
* **Bug Fixes**
  * Empty-selection `Ctrl+C` continues to interrupt remote processes.
  * Read-only hosts reject paste and newline input while still allowing local copying.
  * Hosts without input permission continue to reject remote input.

### DIFF
--- a/changelog.d/924.md
+++ b/changelog.d/924.md
@@ -9,4 +9,6 @@
   added a line. All four now behave the way they do in a local pane — including
   `⌘C` / `⌘V` on macOS — while an empty-selection `Ctrl+C` still interrupts the
   remote process, and a host started without `--allow-input` still receives
-  nothing. (#924)
+  nothing. `Shift+Enter` sends the newline form the remote app actually asked
+  for: apps that enable the kitty keyboard protocol get it, and anything else
+  keeps the plain carriage return it has always received. (#924)
--- a/changelog.d/924.md
+++ b/changelog.d/924.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **A mirrored remote pane now copies, pastes, and inserts newlines like any
+  other terminal.** Attaching to a remote wmux gave you a pane that forwarded
+  every keystroke straight through, so the editing habits that never leave your
+  own machine stopped working: selecting text copied nothing, `Ctrl+C` over a
+  selection interrupted the remote process instead of copying it, `Ctrl+V` did
+  nothing at all, and `Shift+Enter` submitted where a local pane would have
+  added a line. All four now behave the way they do in a local pane — including
+  `⌘C` / `⌘V` on macOS — while an empty-selection `Ctrl+C` still interrupts the
+  remote process, and a host started without `--allow-input` still receives
+  nothing. (#924)
--- a/src/renderer/components/Remote/RemoteMirrorTerminal.tsx
+++ b/src/renderer/components/Remote/RemoteMirrorTerminal.tsx
@@ -4,6 +4,7 @@ import { useT } from '../../hooks/useT';
 import { applyUnicodeWidthModel } from '../../../shared/terminalUnicode';
 import { computeMirrorFontSize, mirrorFitKey, MAX_FIT_PASSES } from './mirrorFit';
 import { decideMirrorKeyWithRepeat } from './mirrorInput';
+import { foldRemoteKeyboardState, acceptsCsiU, INITIAL_REMOTE_KEYBOARD_STATE } from './keyboardProtocol';
 import { useStore } from '../../stores';
 import { terminalFontFamilyCss } from '../../utils/terminalFont';
 import { createAutoSelectionCopy } from '../../utils/autoSelectionCopy';
@@ -97,6 +98,12 @@ export default function RemoteMirrorTerminal({ attachId, error, readOnly }: Remo
   // flip (allowInput probe resolving after mount) doesn't tear down and
   // re-subscribe the whole attach — only paneWrite needs the live value.
   const readOnlyRef = useRef(readOnly);
+  /**
+   * What the remote app has asked for in the way of key encodings, folded from
+   * its own output. A ref, not state: the key handler reads it synchronously
+   * and nothing renders from it.
+   */
+  const remoteKeyboardRef = useRef(INITIAL_REMOTE_KEYBOARD_STATE);
   readOnlyRef.current = readOnly;
   // Same reason: the key handler is installed once, at mount, and needs the
   // CURRENT attach to write to. Listing `attachId` in that effect's deps would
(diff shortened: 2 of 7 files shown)


openwong2kim:

This PR closes #895 by giving the remote-attach mirror the four local editing conveniences it was missing: auto-copy on selection, `Ctrl+C` over a selection copying instead of sending SIGINT to the remote, `Ctrl+V` pasting from the local clipboard, and `Shift+Enter` inserting a newline. It touches `RemoteMirrorTerminal.tsx` plus a new pure chord table in `mirrorInput.ts`, reusing the utilities a local pane already extracted (`createAutoSelectionCopy`, `pastePtyChunked`, `copySelectionWithFeedback`) rather than re-deriving them, so a mirror copies and pastes exactly like a local pane. The two invariants pinned by the tests matter as much as the fixes themselves — an empty-selection `Ctrl+C` still interrupts, and a read-only host receives nothing down the new direct-write path, which is gated on the one function that can reach `paneWrite`. All four behaviors act on the local selection and clipboard only, so the mirror gains terminal behavior without claiming any ownership of the remote pane. Matching every chord on physical `code` as well as `key` also carries over the CJK-IME hardening, which is where these same handlers went dead for local panes before. Triaged as **P3**, in line with the underlying enhancement.

PR #927, opened 2026-08-16
feat(sidebar): optional per-workspace color tag

### LINKED ISSUE #926: [Feature]: Optional per-workspace color tag in the sidebar (and 48px rail)
### Problem / motivation

With a long workspace list every row is the same shape: one status dot plus a monospace name at the same size and weight. Finding a specific workspace means *reading*, not *seeing* — and reading is slow when many rows share a naming pattern (three-letter role names, `client-a` / `client-b`, `api-*`, several worktrees of one repo).

The collapsed 48px rail is worse: the label is reduced to two characters (`C6`, `C7`, `C8`), so several workspaces become visually identical.

The indicators that exist cannot fill this gap, by design: the dot encodes agent status, the git lights encode git state, the badges encode PR/unread. All of them are *computed* signals — none can express "this one is mine to watch".

### Proposed solution

An optional, user-assigned color tag per workspace.

- Right-click a workspace → **Color tag** → one row of swatches (8 curated hues + "None").
- The color renders as a 3px rail on the left edge of the sidebar row, and the same rail in the 48px mini-rail.
- Default is no tag: an untagged workspace renders exactly as today.
- Purely visual — it does not encode agent/git/PR state and does not travel the metadata event/RPC bus.

Two deliberate constraints:

- **Fixed 8-hue set, not a free-form color picker.** The sidebar renders on every bundled theme, dark and light. A user-picked `#111111` disappears on the dark themes, `#ffff00` is unreadable on the light ones. A curated set stays legible everywhere and makes the persisted value trivially validatable.
- **A rail, not a tinted row background or a tinted name.** The row background already carries active/hover/multiview state and the name has to keep its contrast on every theme. A rail adds a channel instead of overloading one.

### Alternatives considered

- **Renaming workspaces with emoji/symbol prefixes** — works today, but it eats the name column, is invisible in the 48px rail (the label is the first character + index), and it changes the name that pane auto-names and copied session info are keyed off.
- **Reordering so related workspaces sit together** — helps a little, but position is not identity: after a couple of drags the mental map is stale.
- **Reading the status dot** — that dot answers "is an agent busy", not "which workspace is this". Overloading it would lose the signal that is already there.
- Prior art: VS Code's `workbench.colorCustomizations` per folder and JetBrains' project color both solve the same "which window am I in" problem with a color assigned by the user, not derived from state.

### Additional context

Implementation is done and I'm happy to open the PR: `Workspace.color?: WorkspaceColorId` persisted through the existing session snapshot (the `...ws` spread in `buildSessionData`, so no new persistence path), with `loadSession` dropping unknown ids rather than rendering them or failing the load. Multiview keeps its blue left border — when both apply the color rail shifts 2px right so the two signals stay separately visible. New i18n keys land in `en.ts` only; other locales fall back to English until translated.

Locally: `tsc --noEmit` clean, new unit tests for the palette/normalizer and the store action (set / clear / unknown id / load-time normalization) pass, and the parallel suite shows no regression against `main` on the same machine.

Happy to adjust the surface first if you'd rather have it somewhere else — e.g. picker in the workspace profile modal instead of the context menu, a different palette size, or the rail also showing in FleetView.

### Confirmation

p-poppe:
## Summary

Adds an optional, user-assigned color tag per workspace, set from the row context menu and drawn as a 3px rail on the sidebar row and on the 48px mini-rail. It is purely a visual label for telling workspaces apart — the status dot, git lights and badges keep their meanings untouched.

## Changes

- `src/shared/workspaceColors.ts` (new) — the 8-hue palette, the `WorkspaceColorId` union, and `normalizeWorkspaceColor`, the single load/save boundary that turns anything unknown into "no tag" instead of throwing.
- `src/shared/types.ts` — `Workspace.color?: WorkspaceColorId`. Persists through the existing `...ws` spread in `buildSessionData`; no new persistence path.
- `stores/slices/workspaceSlice.ts` — `setWorkspaceColor(id, color | undefined)`; sets, or deletes the field when cleared. Deliberately publishes no metadata event (a view label has no business on the RPC bus). `loadSession` normalizes every workspace's tag, dropping unknown ids rather than rendering them or failing the load.
- `Sidebar/WorkspaceItem.tsx` — the rail plus a **Color tag** submenu (8 swatches + "None", selection shown as a ring). When the workspace is also in multiview the rail shifts 2px right, clearing the blue multiview border instead of covering it — the two signals mean different things.
- `Sidebar/MiniSidebar.tsx` + `selectors/workspaceProjections.ts` — the same rail in the 48px rail, with `color` added to `WorkspaceRailSummary` and to its equality fn so the projection's reference cache still holds (no extra re-renders).
- `i18n/locales/en.ts` — new keys; other locales fall back to English until translated.

Untagged workspaces render exactly as before — every style is behind `if (tagColor)`.

Why a fixed 8-hue set rather than a free-form picker: the sidebar renders on every bundled theme, dark and light. A user-picked `#111111` disappears on the dark ones and `#ffff00` is unreadable on the light ones. A curated set stays legible everywhere and makes the persisted value trivially validatable. The hexes are literals rather than `var(--accent-*)` on purpose — the theme accents are only four and are re-tinted per theme, so a workspace tagged "green" would silently change identity when the user switches theme.

## CHANGELOG entry

Added as `changelog.d/927.md`:

### Added

- Workspaces can carry an optional color tag (right-click a workspace → **Color tag**). The color shows as a rail on the sidebar row and on the 48px mini-rail, so a long list of same-shaped names stays scannable. Purely visual — the agent status dot, git lights and badges keep their meanings.

## Stability tier impact

- [x] Additive change to a `stable` surface (new optional field on `Workspace`, new store action). No existing field or event changes shape; `color` absent = current behavior.

n/a — the tag is renderer/session state only. It is intentionally NOT on `WorkspaceMetadata`, so no PaneMetadata wire shape, EventBus envelope, permission point, pipe surface, RPC method or event type is touched.

## Test plan

- Unit tests added:
  - `src/shared/__tests__/workspaceColors.test.ts` — palette/id coverage, and that `normalizeWorkspaceColor` drops `undefined`, `null`, `''`, unknown names, a raw hex, a number, `{}`, `[]` and `'toString'` (prototype-key probe) instead of throwing.
  - `src/renderer/stores/__tests__/workspaceColor.test.ts` — set / clear (field removed, not nulled) / unknown workspace id is a no-op / `loadSession` keeps valid tags, drops invalid ones, and still loads the session.
- Manual verification: **not done — flagging this honestly.** My sandbox installed deps with `--ignore-scripts`, so there is no Electron binary here and I could not run the app. Everything above is static + unit-level. The visual bits that deserve a human eye before merge: the rail against each bundled theme (especially the light ones), the rail + multiview border sharing one row, and the picker submenu flipping near the right screen edge. If you'd rather I verify locally first, I can do a full install and come back with screenshots.
- Suite status on my machine: `npx tsc --noEmit` clean; `eslint` on the touched files reports 0 errors (1 pre-existing `shortenPath` unused-var warning in `WorkspaceItem.tsx`, untouched by this PR). The parallel suite does not fully pass in my sandbox — deps installed with `--ignore-scripts`, so native/registry/canvas-dependent tests fail. I ran it on `main` and on this branch on the same machine to compare: 46 vs 47 failing files, and the single delta is `activitySummary.test.ts`'s timing assertion (`expected 51 to be less than 50` ms), which this PR does not touch. Net: +10 passing tests, no regression. Please let CI be the real gate.

## Related issues / PRs

Closes #926

## Summary by CodeRabbit

* **New Features**
  * Added optional workspace color tags with eight color choices.
  * Workspace colors can be selected, cleared, and reset from the workspace context menu.
  * Color tags appear as visual rails in the sidebar and mini-rail without changing existing status or badge indicators.
  * Duplicated workspaces start without the original workspace’s color tag.

* **Bug Fixes**
  * Invalid or unknown saved workspace colors are safely ignored.

* **Tests**
  * Added coverage for color selection, clearing, persistence, validation, duplication, and display mappings.


openwong2kim:

I reviewed the full diff and this is a clean, well-scoped additive change: a new `Workspace.color?: WorkspaceColorId` field, a new `shared/workspaceColors.ts` module that owns the 8-hue palette and acts as the single load/save boundary (`normalizeWorkspaceColor` drops anything unknown instead of throwing — including the prototype-key probe), and a `setWorkspaceColor` store action that deletes the field when cleared rather than nulling it. The rendering side is entirely behind `if (tagColor)` guards, so an untagged workspace stays pixel-identical to today, and the multiview rail shift plus the projection equality-fn update in `workspaceProjections.ts` show the right attention to not breaking the reference cache or overloading the blue multiview border. The honest flag that manual/Electron-level verification was not possible in the sandbox is noted — the rail against each bundled theme (especially the light ones) and the submenu flip near the right screen edge are the two things that deserve a human eye before merge, so CI plus a local visual pass should be the gate. Labeling **P3** (cosmetic feature, matching #926) — nothing here blocks or destabilizes anything, and I have no structural objections to the approach.

PR #916, opened 2026-08-16
Home agent verbs on the pane and workspace, and detect Grok

openwong2kim:
## Problem

The workspace-spanning 36px agent toolbar lied about blast radius. Compose, attach, and new conversation sat on a strip that spanned every pane; Multi Task disappeared when the fleet was empty; Broadcast looked like a pane verb. Grok was not a first-class agent: the detector had no profile, and a Grok TUI that *read this repo* (or any file quoting Claude's banner/footer) opened Claude's compound gate, so the roster and pane name said Claude Code.

Grok also does not scroll. It lives on the alt screen, so xterm has no scrollback and the wheel became Up/Down — prompt history, not the transcript.

## Change

Agent verbs go home:

- **Compose (`⌘G`), attach, new conversation** live on the focused pane's tab cluster.
- **Broadcast** is a compose target (`This pane` / `All N terminals`). All N is 4-second armed.
- **Multi Task / Start agents** sit on the selected workspace card (deck header only when the sidebar is collapsed).
- The 36px workspace-spanning strip is gone.

Grok is a first-class agent:

- Detector recognizes the live TUI (startup banner, Help improve Grok, composer `Grok 4.6 (high) · always-approve`).
- Claude's compound gate now requires **per-line TUI chrome**, not a substring anywhere in the 4 KB probe. Status patterns cannot steal `lastAgent` back to another gated name.
- Hook display names go through `agentSlugToDisplay`, so a new row in `AGENT_IDENTITIES` is enough.
- Alt-screen wheel sends PageUp/PageDown, which is how Grok scrolls its own transcript.

## Test plan

- [x] AgentDetector: Claude compound gate still opens on real chrome; source dumps of this repo do not; Grok live composer gates; Grok stays Grok after a Claude-source dump
- [x] Compose popover: This pane / All N arm / snippets
- [x] Fan-out trigger still renders with an empty fleet
- [x] SurfaceTabs inject actions, pane cluster width, ⌘G shortcut, placePopover, alt-screen wheel
- [x] 106 tests in the touched files pass

## Out of scope

`TODOS.md`, the 2026-08-15 HTML mock, and `docs/internal/*` stay local — they are not part of this PR.

## Summary by CodeRabbit

* **New Features**
  * Added Grok detection and naming across rosters and terminal panes.
  * Added focused-pane Compose, file attachments, snippets, new conversations, and multi-terminal broadcast controls.
  * Added workspace-level controls for spawning tasks across terminals.
[...]

PR #932, opened 2026-08-17
Decide a pane's agent from process truth, not screen text

### LINKED ISSUE #919: Decide a pane's agent from process truth, not from what the terminal printed
## The problem

A pane's agent identity is decided by one signal: what the terminal printed.

That signal cannot distinguish an agent from a sentence about an agent. Measured on a live Claude pane:

```
"I will run codex review on this diff"   -> pane becomes Codex CLI
"compare with gemini output please"      -> pane becomes Gemini CLI
"the opencode plugin does this"          -> pane becomes OpenCode
```

The gates that do this are literally `/codex /`, `/gemini /`, `/opencode/` — a substring tested against the whole output chunk.

We keep paying for this. `claude`, `kiro` and `grok` have each been promoted, one at a time, from a substring to line-level chrome inspection, with helper regexes for box-drawing noise and for "this line is source code, not chrome". #916 promoted Grok because a Grok pane reading this repo was labelled Claude; the same PR then had to add an ownership lock, and the lock made the mislabel permanent until it was fixed again. Each promotion narrows one hole and leaves the rest open, and every new adapter starts at the bottom.

Identity feeds the pane label, the sidebar roster, status events, and the orchestrator's count of who is still busy. When it is wrong, work is routed on a lie.

## Two stronger signals already exist in the tree, and neither decides identity

**1. The process image name.** `src/daemon/AgentProcessTracker.ts` already snapshots the process table (pid/ppid/name), walks the pane shell's descendant tree, and picks the agent process. Its own comment:

```js
/** Agent launcher binaries (native installs): a descendant with one of these
 *  stems IS the agent — the strongest possible attribution. */
const AGENT_STEMS = new Set(['claude', 'codex']);
```

and

> this tracker only ever REPLACES guesswork with process truth — it never invents a state

It is used for liveness only — alive/dead for the resume chip. The table holds 2 of the agents we support. The enumeration cost is already paid (one per launch, then a ride on the existing ProcessMonitor tick), and both Windows and POSIX paths are implemented.

Prose cannot rename a process. `claude.exe` stays `claude.exe` however often the model says "codex".

**2. The hook's own report.** Every bridge sends `{ agent, ptyId }` from inside the agent's runtime, and `HookSignalRouter` already tracks authority per pane (`touchAuthority`, `isGovernedFor`). But `hooks.rpc.ts` uses `signal.agent` for latency metering, authority, alarms and notification payloads — never to set the pane's identity. A hook can prove it is Kiro and the detector will still call the pane Claude off a quoted banner.

## Proposal

Make identity a precedence, the way state already is ("hook is canonical"):

| | source | beats |
|---|---|---|
| 1 | hook self-report | everything |
| 2 | process image name | screen |
| 3 | screen chrome | — |

Sketch, in the order the work would land:

- **A.** Extend `AGENT_STEMS` to every agent we ship, and thread the picked process's stem out of `AgentProcessTracker` as an identity, not just a liveness handle. Node-hosted agents share the `node` stem — the tracker already separates `RUNTIME_STEMS` for this reason — so this step needs the command line as well (`Win32_Process.CommandLine`, `ps -o args`).
- **B.** Let a hook signal set identity for its pane, reusing the authority window that already governs state.
- **C.** Demote the screen gates to the fallback they should be: the answer when 1 and 2 are silent, which is the case that actually needs them — a user who typed `codex` into a shell pane wmux launched as a shell.

`AgentProcessTracker.statusFor()` is already tri-state, with `undefined` meaning "could not attribute, keep the old heuristic". The same discipline makes this incremental: add the stronger tiers above the existing one and delete nothing until they are proven.

## What this would have prevented

- The Grok-labelled-Claude bug #916 fixed (process is `grok`).
- The ownership regression #916 introduced and then fixed (process is `claude`; a sentence cannot outrank it).
- The next adapter starting at the bottom of the same ladder.

## Limits worth stating

- Remote/SSH panes have no local process — tier 3 stays for them.
- A user who launches a second agent inside a pane genuinely changes identity; the process tier tracks that correctly, the screen tier only guesses at it.
- Reading a command line is more than reading a name; it should be measured before it is assumed cheap.

openwong2kim:
## What

A pane's agent identity is no longer decided by what the terminal printed. It is folded from three signals into one precedence:

```
corroborated hook > fresh exact-routed hook (2-min TTL) > live attributed process > screen chrome
```

with a residue veto: a confirmed-dead same-slug screen read is sticky detector residue, not an agent — unless a fresh exact-routed hook backs the same slug, which is the relaunch itself.

Closes #919. Fixes the mislabel class from #916 (a Grok pane labelled Claude — twice) at the layer that actually caused it: the screen gates cannot tell an agent from a sentence about an agent, yet they were the only identity source.

## How

- **`src/daemon/canonicalAgent.ts` (new)** — the tier rule as a pure function. `IDENTITY_TTL_MS` (2 min) is deliberately not the 30-min hook-authority TTL: that window spans long quiet tool calls for notification suppression, but an identity claim riding a stale entry would mislabel the pane's next agent.
- **`AgentProcessTracker`** — the launch-edge process pick now also answers *which* agent: cmdline attribution with strict token discipline (quote-aware tokenizer; candidates are only the script token — `argv[1]`, or the first non-flag when the launcher prefixes runtime flags, in path form only so a bare `claude.js` is the user's own file — the non-flag after a package runner or `python -m`, and `node_modules/<pkg>` boundary segments; arbitrary ancestor directories never match). Pick = shallowest **attributed** by depth regardless of native-vs-runtime class; exact-depth different-slug ties drop the slug rather than guessing. Adds `rearm()` (forced probe when a banner names a different agent than the tracked one), a 30 s negative-result backoff so chatty banners can't re-run `ps` per event, and a state-change listener for attribution/death edges.
- **`HookSignalRouter`** — authority entries record routing provenance (`exact` vs cwd-guess); new `authorityAgentFor()` (identity input) and `expireAuthorityFor()` — the 30-min detector veto dies with the launch that earned it, so a relaunched same-slug agent with broken hooks gets its completions through.
- **`daemon/index.ts`** — every emission/feed site (`session:agent`, detector re-emission, `session:active`, `daemon.getAgentName`) resolves canonical identity. Detector events contradicted by a tier-1/2 identity are suppressed outright; **`data.agent` is never rewritten** (main's dedup keys off it). The pane label self-heals via `activity.active` / `getAgentName`; no fabricated lifecycle events. Hook signals arm the tracker so bannerless agents become corroborable; death edges expire authority and re-persist the corrected label.

`AgentDetector` gates are untouched — the screen stays the fallback for remote/SSH panes where no local process or hook exists.

## Cost

One process-table enumeration per agent **launch** per pane (arm/rearm), then a piggyback ride on the existing ProcessMonitor cadence. `ps -axo pid=,ppid=,args=` measures **0.01 s** wall on macOS (~600 processes); `identityFor`/`statusFor` are synchronous map reads. Windows rides the same single CIM query, now with `CommandLine`.

## Testing

- `canonicalAgent.test.ts` (new): all tier combinations including the #916 regression guard (live different-slug process beats even a fresh exact hook), the residue veto and its fresh-hook rescue, and non-exact routing never standing alone.
- `AgentProcessTracker.test.ts`: cmdline resolution real-world spellings and the false-positive guards (`node /work/claude/scratch.js`, `node demo.js claude`, `node claude.js` never resolve; `node --max-old-space-size=… cli.js` does), depth-regardless-of-class picking, rearm past a stale alive pick, queued-rearm replay behind an in-flight probe, backoff, listener edges, PID-reuse/stale-onDead/stale-batch guards.
- `HookSignalRouter.test.ts`: authority shape/TTL/provenance, death-scoped expiry.
- `ProcessMonitor.reuse.test.ts`: a stale monitor batch must not unwatch a replaced watch.

tsc clean, eslint clean on changed files; full `src/daemon` + `src/shared` suites pass (192 files / 3475 tests).

## Out of scope (deliberate)

- No `AgentDetector` pattern changes (screen tier must keep working for remote panes).
- No identity-flip `agent.event` fabrication — its lifecycle semantics would invent completions.
- No pgid/tty foreground enumeration; depth-first among attributed candidates approximates it.
- Main-process local mode (`PTYBridge`) identity paths untouched — daemon mode owns both detectors and ingest in one process.
- Follow-ups flagged by the pre-PR review panel, left for their own issues: (a) the desktop status label is not cleared when canonical `activity.active.data` goes null (tri-state `DaemonClient.onActive` plumbing — flicker-sensitive, needs its own change); (b) authority expiry on a **slugless** tracked death is pane-scoped, not pick-scoped (rare wrapper-orphan case; the exact fix needs a death timestamp the monitor edge does not carry today); (c) POSIX `kill(pid, 0)` liveness has no image identity check against PID reuse (pre-existing; this PR's `rearm()` at least recovers a stale-pick label on the next conflicting banner).

## Summary by CodeRabbit

* **New Features**
  * Added more reliable agent identity detection across processes, hooks, runtimes, aliases, and scoped packages.
  * Agent events now include canonical identity information and its detection source.
  * Added support for refreshing process tracking and reporting live identity state changes.

* **Bug Fixes**
  * Prevented stale monitoring results from affecting replacement processes.
  * Improved handling of ambiguous, failed, terminated, and conflicting identity signals.
  * Added expiration and cleanup for outdated hook-based identity data.
  * Improved label recovery and suppressed contradictory identity notifications.

### DIFF
--- a/changelog.d/932.md
+++ b/changelog.d/932.md
@@ -1,18 +1,21 @@
 ### Fixed
 
-- **A pane's agent label now comes from process truth, not screen text.** The
+- **A pane's agent label now prefers process truth over screen text.** The
   label, roster entry, and status events for a pane are decided by a
   precedence of signals — a corroborated or fresh hook self-report first, then
   the live attributed agent process, and only then what the terminal happened
-  to print. A confirmed-dead same-slug read on screen is treated as sticky
-  residue instead of relabeling the dead agent "alive". This fixes the class
-  of mislabels where a Grok pane was shown as Claude: the screen gates cannot
-  tell an agent from a sentence about an agent, and until now they were the
-  only identity source. (#932)
+  to print. Screen text still names the pane when nothing stronger is
+  available, which is what remote and SSH panes rely on. A confirmed-dead
+  same-slug read on screen is treated as sticky residue instead of relabeling
+  the dead agent "alive". This fixes the class of mislabels where a Grok pane
+  was shown as Claude: the screen gates cannot tell an agent from a sentence
+  about an agent, and until now they were the only identity source. (#932)
 
 - **Detector notifications contradicted by a live agent of another kind are
   suppressed.** A screen-detected "agent finished" that disagrees with the
   pane's hook- or process-backed identity no longer notifies, and the pane's
-  label self-heals on the next output burst. The 30-minute detector veto also
-  expires with the launch that earned it, so a relaunched agent with broken
-  hooks gets its completions through. (#932)
+  label self-heals on the next output burst. A banner this build cannot map to
+  a known agent is not treated as a disagreement, so an unrecognised name never
+  costs a pane its status updates. The 30-minute detector veto is scoped to the
+  agent launch that earned it and expires when that process is seen to die, so
+  a relaunched agent with broken hooks gets its completions through. (#932)
--- a/changelog.d/932.md
+++ b/changelog.d/932.md
@@ -1,0 +1,18 @@
+### Fixed
+
+- **A pane's agent label now comes from process truth, not screen text.** The
+  label, roster entry, and status events for a pane are decided by a
+  precedence of signals — a corroborated or fresh hook self-report first, then
+  the live attributed agent process, and only then what the terminal happened
+  to print. A confirmed-dead same-slug read on screen is treated as sticky
+  residue instead of relabeling the dead agent "alive". This fixes the class
+  of mislabels where a Grok pane was shown as Claude: the screen gates cannot
+  tell an agent from a sentence about an agent, and until now they were the
+  only identity source. (#932)
+
+- **Detector notifications contradicted by a live agent of another kind are
+  suppressed.** A screen-detected "agent finished" that disagrees with the
+  pane's hook- or process-backed identity no longer notifies, and the pane's
+  label self-heals on the next output burst. The 30-minute detector veto also
+  expires with the launch that earned it, so a relaunched agent with broken
+  hooks gets its completions through. (#932)
--- a/src/daemon/AgentProcessTracker.ts
+++ b/src/daemon/AgentProcessTracker.ts
@@ -1,6 +1,7 @@
 /**
  * Edge-triggered "is this pane's agent process still alive?" tracker — the
- * process-truth gate for the persistent resume chip (ResumeInfoChip).
+ * process-truth gate for the persistent resume chip (ResumeInfoChip) and,
+ * since #919, tier-2 of the pane's canonical agent identity.
  *
  * Problem it solves: on panes WITHOUT OSC 133 shell integration the chip's
  * only busy signal was a decaying activity heuristic (HOOK_RUNNING_TTL_MS).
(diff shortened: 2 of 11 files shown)


openwong2kim:

This PR implements the canonical agent-identity tier rule from #919: a pane's agent is now decided by a precedence of corroborated hook > fresh exact-routed hook (2-min identity TTL, deliberately shorter than the 30-min authority TTL) > live attributed process > screen chrome, with a residue veto so a confirmed-dead same-slug screen read no longer resurrects a label. It touches the daemon core — a new pure canonicalAgent.ts holding the tier rule, AgentProcessTracker extended from liveness-only to full cmdline attribution with strict token discipline (path-form script slot, node_modules boundary segments, no substring matches), HookSignalRouter gaining routing provenance and death-scoped authority expiry, and every emission/feed site in daemon/index.ts routed through the canonical resolver, with detector events contradicted by a tier-1/2 identity suppressed outright rather than relabeled. The ProcessMonitor stale-batch identity guard and the dedicated regression tests, including the #916 Grok-labelled-Claude case and the queued-rearm/backoff paths, make this a careful and reviewable change. Since identity feeds the pane label, the roster, notifications, and the orchestrator's busy count, fixing the layer that actually caused the #916 mislabel class matters well beyond cosmetics — I'm tagging this P2 and it closes #919.

openwong2kim:
Addressed CodeRabbit's three notes in 26c4fb80.

The `index.ts:179` one was a real defect, not a nit. `detectorSuppressedBy` vetoed whenever the canonical slug differed from the screen slug, and an unmappable display name reads as `undefined` — so a banner this build cannot name was treated as *disagreement*. These events carry status and liveness as well as a label, so the veto cost the pane its running/idle updates over a name we merely failed to parse. `canonicalIdentityFor` right above it already treats an absent slug as "no constraint" (`...(screenSlug ? { screenSlug } : {})`); the two readings disagreed. Fixed by returning early on an absent slug.

Moved the rule into `canonicalAgent.ts` beside the tier rule it guards — it is pure, that module is where the rule is unit tested, and it was unreachable from a test while private to a 4600-line file. Added the veto's five cases, including the absent-slug regression.

The two changelog notes were fair: the entry now says screen text still names a pane when nothing stronger exists (remote/SSH depend on it), and states that the detector veto is scoped to the agent launch and expires when that process is seen to die.

Daemon + shared suites green (3455 passed), `tsc` clean on both configs.

PR #934, opened 2026-08-17
feat(sidebar): lead roster rows with the pane title, not the vendor name

### LINKED ISSUE #933: [Feature]: Roster rows should lead with the pane title, not the vendor name
### Problem / motivation

In the workspace agent roster (the expandable "Agents N" list under a workspace in the sidebar), every row leads with the **vendor name**. A workspace running several sessions of the same vendor — which is the normal case for me — renders as:

```
● Claude Code · w2-127    Waiting
● Claude Code · w2-123    Complete
● Claude Code · w2-131    Waiting
```

Every readable word is identical. The only thing that differs is an opaque coordinate I have never memorized, so picking the right row means clicking through them.

Meanwhile each of those panes already has a name I gave it or that the agent set — "Zwroty", "AI", "Scalar SINOTKEN" — and that name is visible in the tab strip and in the pane corner. It just never reaches the roster.

### Proposed solution

Lead the row with the surface title when it exists, and move the vendor name into the muted trailer:

```
● Zwroty           Claude Code · w2-127    Waiting
● AI               Claude Code · w2-123    Complete
● Scalar SINOTKEN  Claude Code · w2-131    Waiting
```

- A surface with no title leads with the vendor name exactly as today — no row loses information.
- The vendor is never printed twice: it appears in the trailer only when the title took the lead. For a mixed-vendor workspace it still answers "which agent is this", just one step to the right.
- The tab position (`#2/3`) keeps its place in the trailer for leaves holding several surfaces.

### Alternatives considered

- **Renaming panes** (`paneLabel`) — works, and the roster already shows it, but it means maintaining a second name for something that already has one, and it does not help the many panes whose title the agent set itself.
- **Widening the coordinate** (`w2-127` → something mnemonic) — the coordinate is deliberately stable and opaque; making it descriptive would fight its purpose.
- Prior art: tmux and Windows Terminal both let the pane's own title win over the program name in every list that names panes.

### Additional context

The data is already in the projection: `WorkspaceAgentRosterRow.surfaceTitle` is populated in `selectWorkspaceAgentRoster`. The reason it never shows is a single condition in `WorkspaceAgentRoster.tsx` — `locationLabel()` returns `row.paneName` alone when `surfaceCount <= 1`, and appends the title only for multi-surface leaves. Single-surface panes are most workspaces, so in practice the title is invisible.

PR follows — the change is confined to the two label helpers and the row's two spans, plus unit tests for label precedence. Happy to adjust the layout (e.g. keep the vendor leading and append the title instead) if you'd rather have the emphasis the other way round.

### Confirmation

p-poppe:
## Summary

The workspace agent roster now leads each row with the session's surface title and demotes the vendor name to the muted trailer, so a workspace running several sessions of one vendor stops rendering as a column of identical rows.

## Changes

- `WorkspaceAgentRoster.tsx` — `locationLabel()` is replaced by two exported pure helpers:
  - `rosterPrimaryLabel(row)` → `surfaceTitle ?? agentName`,
  - `rosterSecondaryLabel(row)` → vendor (only when the title took the lead, so it is never printed twice) · pane coordinate · `#i/n` for multi-surface leaves.
- The row's two spans swap emphasis: the title truncates first as the sidebar narrows, the coordinate is capped at 40% width. `aria-label`/`title` carry both labels, so the accessible name keeps everything it had.
- No selector change: `WorkspaceAgentRosterRow.surfaceTitle` was already populated — `locationLabel` simply gated it behind `surfaceCount > 1`, which is why single-surface panes (most panes) never showed it.

Before / after for a workspace of same-vendor sessions:

```
● Claude Code · w2-127      →   ● Zwroty           Claude Code · w2-127
● Claude Code · w2-123      →   ● AI               Claude Code · w2-123
● Claude Code · w2-131      →   ● Scalar SINOTKEN  Claude Code · w2-131
```

A titleless surface still leads with the vendor name, unchanged.

## CHANGELOG entry

Added as `changelog.d/934.md`:

### Changed

- The workspace agent roster now leads with each session's pane title instead of the vendor name, so a workspace running several Claude Code sessions no longer renders as a column of identical "Claude Code · w2-1xx" rows. The vendor moves to the muted trailer next to the pane coordinate, and a session whose pane has no title still leads with the vendor name as before.

## Stability tier impact

- [x] No external surface touched (renderer-only label composition; no store, projection, event or RPC change).

n/a — no PaneMetadata shape, EventBus envelope, permission point, pipe surface, RPC method or event type is involved.

## Test plan

- Unit tests added — `src/renderer/components/Sidebar/__tests__/workspaceAgentRosterLabels.test.ts`: title leads when present; vendor leads and is not repeated when the surface has no title; `#i/n` survives in both shapes; and a regression case asserting that three same-vendor rows now produce three distinct primary labels.
- `npx tsc --noEmit` clean; `eslint` clean on the touched files; the roster selector's existing suite still passes (23 tests across both files).
- Manual verification: **not done** — my sandbox installed deps with `--ignore-scripts`, so there is no Electron binary here to run the app. The visual bits worth a human eye before merge: truncation behavior at a narrow sidebar width with a long title, and whether you want the vendor kept at all in the trailer for single-vendor setups.

## Related issues / PRs

Closes #933

## Summary by CodeRabbit

* **UI Improvements**
  * Workspace agent rows now prioritize pane titles for clearer identification.
  * Vendor names appear as muted secondary labels when relevant.
  * Rows fall back to vendor names when no pane title is available.
  * Multi-surface positions and pane details remain visible without clutter.
  * Long labels now truncate cleanly for improved sidebar readability.
  * Accessibility text now reflects the same clearer row labels.


openwong2kim:

This PR implements #933: roster rows now lead with the surface title (`rosterPrimaryLabel` = `surfaceTitle ?? agentName`) and the vendor moves into the muted trailer (`rosterSecondaryLabel`), so same-vendor workspaces stop rendering as a column of identical rows.
I reviewed the diff: it is confined to the two label helpers, the row's two spans, and a new unit test file — renderer-only, no selector, store, or event changes, and the coordinate span is sensibly capped at 40% width so a long title truncates first.
The test coverage matches the rule under test: title-leads precedence, vendor fallback without duplication when there is no title, `#i/n` survival in both shapes, and a regression case asserting three same-vendor rows produce three distinct primary labels; the changelog entry is in place.
One note before merge: the author flagged manual visual verification as not done (no Electron binary in the sandbox) — worth a quick human check of narrow-sidebar truncation before this ships.
Priority P3 — clean, low-risk readability improvement; will close #933 via `Closes #933`.

openwong2kim:
Thanks — merging this. I pushed one fix onto the branch (435e7aca) rather than send you round again for it.

`rosterPrimaryLabel` used `??`, which only falls back on null/undefined, while `rosterSecondaryLabel` tests truthiness. A row whose `surfaceTitle` is `''` therefore led with an empty string **and** had the vendor withheld from the trailer as already-shown, so the row carried no name at all — just the pane coordinate. The selector's `nonEmpty()` keeps that out of practice today, but these are exported pure functions now and the row type permits it, so the guard belongs in the function. Added a test for the empty case.

Also folded in two of CodeRabbit's notes: the new comment is in English (repo language rule — the pre-existing Korean comment below it is left alone), and the changelog says "terminal title" instead of "pane title", since the promoted value is the surface's and one pane can hold several.

The label split itself is the right call, and extracting the two pure functions with tests made this easy to check.

PR #937, opened 2026-08-18
Bring the agent toolbar back on approach, and give the deck one toggle

openwong2kim:
Two pieces of chrome were costing the terminals more than they returned.

## The agent toolbar is one bar again

The 2026-08-15 decision scattered the agent verbs onto the pane tab cluster and the workspace card. In use that meant one verb with three homes — the empty workspace card, the roster header, and the deck header — each with its own label and its own visibility rule, so nothing became muscle memory. A 420px fan-out form opened from a 240px sidebar column and covered the hero.

So the verbs are one workspace-spanning bar again: attach · files · snippets · rich input ⌘G · Broadcast · Multi Task · new conversation. What is different is that it **overlays** the workspace column instead of taking a row. That answers the objection the removal was built on — the bar spends no chrome module, so the terminals lose no height and no PTY is resized when it appears.

It sits exactly where the terminal's prompt line lives, so the reveal is guarded:

- a dwell delay, so crossing the band on the way elsewhere is not intent
- suppressed while a pointer button is held, so dragging out the last output row does not summon a bar over the text being selected
- retreats on a keystroke, and re-arms when the pointer next moves
- hysteresis — the arming band is 8px but a revealed bar holds across its whole height, or it hid itself the moment the cursor rose onto one of its buttons
- the bar's background is `pointer-events: none`; only its controls claim a hit area, so the two terminal rows it floats over keep their clicks and text selection
- a pin toggle for the plain always-on strip, and a 2px lip so a hidden bar is not invisible

## The collapsed deck stops costing a column

Collapsing the deck left a 36px glyph rail down the right edge: four icons at the top, an expand chevron at the foot, ~85% of a full-height column empty between them. Opening the deck is one command, so it gets one button, beside Settings. Collapsed, nothing renders on that edge and the grid runs to the window edge (measured 1244px → 1280px on a 1280px window).

The rail's per-tab badges become one dot — deliberately a boolean, not a total, because unread messages and dirty worktrees are different kinds of thing and adding them would invent a number that means nothing. At zero there is no dot, the rule the fleet vitals already follow.

Three constraints had to hold and do: the deck's state is app-global (so an app-wide row is scope-correct, unlike the pane verbs), Windows native controls are already reserved through `env(titlebar-area-*)`, and the drag gap measured 829px against 24px spent.

## Defects fixed along the way

Moving the triggers meant rewriting their call sites, which surfaced these:

- **Fan-out anchored to the viewport's left pad.** `FanOutAnchor` carried only `{top,left}` while `placePopover` right-aligns against `right`, so every anchor collapsed onto the 8px pad.
- **The fan-out dialog was measured at a fixed 560px** while it can take `70vh`, so it overflowed and clipped its own repo field and Launch button. The footer is sticky now — the primary action was below the fold of a scrolling form.
- **The per-task row left its card.** The title input was `flex-1` without `min-w-0`, the role select sizes to its longest option, and the derived branch name was `shrink-0` — all three competing for 420px. The slug moved to its own line; it is derived feedback, not an input.
- **Two compose surfaces answered ⌘G.** Restoring the bar left `ComposePopover` watching the same `toolbarPopover === 'rich'` as the bar's own Rich Input, with `useComposeShortcut` racing the bar's handler. One keypress opened both.

## Panel review

Three models reviewed the branch independently. They converged on three defects, all fixed here:

- **The file explorer did nothing.** Its popover portals to `document.body`, so a click on a row was "outside" the bar's DOM and the outside-mousedown handler tore it down before the click could run.
- **The drag guard latched forever.** It cleared only on `pointerup`, so a release outside the window, a `pointercancel`, or a native context menu left it set — and the reveal path returns early while it is.
- **The hidden bar was `aria-hidden` but not `inert`,** leaving eight buttons in the tab order.

Agreed by two: New chat fired on one click (main had grown a two-step confirmation that restoring the older bar dropped, so a stray click could send `/clear` and discard a live conversation); the bar's background took the terminal's clicks; there was no keyboard route in.

Single-reviewer and verified: ⌘G died with the bar though it is documented to survive the inject-chrome setting; the bar floated over remote workspace views while every verb targeted the local pty; opening a popover did not close the fan-out dialog; a popover left `toolbarPopover` set after its pty went away, holding the bar open with nothing rendered; the pointer watch forced a synchronous layout in xterm's paint frame.

Fixing the `inert` finding then broke ⌘G — Rich Input mounts before the host flips `revealed`, and autofocus is dropped inside an inert subtree, so focus stayed on the terminal and Escape reached xterm instead of the popover. `hold` is computed synchronously in the bar, so it gates `inert` too.

## Verification

`tsc` clean on both configs, 11,657 tests pass. New coverage: the reveal state machine (dwell, drag, typing, pin, hold, and each way a press can end without a `pointerup`), the fan-out anchor rect, the deck toggle's signal rule and arrow direction, and a guard that the pane cluster carries no agent verbs.

Driven against a live build over CDP as well — the reveal geometry, the file explorer surviving a press on one of its rows, New chat arming instead of firing, the bar background letting clicks reach the terminal, and the grid reaching the window edge once the rail was gone.

## Summary by CodeRabbit

* **New Features**
  * Added a workspace-wide agent toolbar that reveals on hover or focus and can be pinned.
  * Added toolbar actions for terminal attachment, file browsing, snippets, rich input, broadcast, fan-out, and new conversations.
  * Added command-palette actions for multi-tasking and pinning the toolbar.
  * Added a titlebar deck toggle with unread and dirty-state indicators.
* **Bug Fixes**
  * Improved popover interaction, fan-out dialog layout, and workspace sizing.
  * Prevented toolbar overlap in remote workspaces and fixed duplicate-action behavior.


openwong2kim:

This PR restores the agent toolbar as one workspace-spanning bar (attach · files · snippets · rich input ⌘G · Broadcast · Multi Task · new conversation), reversing the 2026-08-15 split, but reworks it as an overlay revealed on approach rather than a layout row — so the terminals lose no height and no PTY is resized when it appears — and it removes the collapsed deck's 36px glyph rail in favor of a single `DeckToggle` beside Settings in the titlebar, reclaiming a full-height column for the grid. Because the bar sits exactly on the terminal's prompt line, the reveal is guarded by a real state machine in `useHoverReveal` (150ms dwell, suppression while a pointer button is held or the user is typing, hysteresis keep-alive band, self-healing drag guard for a pointerup that never arrives, focus as the keyboard route in), which keeps it from fighting the user for that line. The move also surfaces and fixes a meaningful set of defects: the fan-out dialog anchoring and overflow (real trigger rect with `right`/`bottom`, sticky footer, per-task row layout), the file explorer popover being torn down by outside-click before its row click could run (`data-toolbar-owned`), ⌘G opening two compose surfaces at once, a two-step confirmation on New chat since it sends `/clear`, and `inert` rather than aria-hidden so hidden controls leave the tab order. Scope is 44 files (+1306/−1327) across AgentToolbar, Deck, Pane/SurfaceTabs, Sidebar, StatusBar, store slices, i18n, plus DESIGN.md decision rows and a changelog entry, with new tests for the reveal machine, the deck toggle, and guards that the pane cluster carries no agent verbs. Tagging P2 — it is a large chrome rework that also closes several functional bugs (the file browser silently inserting nothing being the worst of them), so it should land soon after review rather than sit.

openwong2kim:
Addressed all six in 05c26073.

The palette finding was the one worth catching. This PR stops the toolbar from mounting over a remote workspace view — its verbs target the *local* pty, so a bar floating over a remote screen would inject into a terminal you cannot see — and then I added an unguarded ⌘K route to fan-out in the same PR, reopening the hole from the other side. Same guard now applies to the command.

The rest:

- **ComposeHost resize** — the dialog was placed from the viewport once and never again, so resizing under an open dialog left it against a stale anchor. It re-renders on resize while open now.
- **Deck toggle a11y** — the signal reached `title` but not `aria-label`, so a screen reader heard "Expand dock" while sighted users could see there was a reason to press it. Both read the same string now.
- **ErrorBoundary in the flex flow** — correct: `ToolbarHost` is layout-neutral but the boundary's fallback is a plain `height:100%` block, so a crash would have squeezed the pane grid. The boundary is the absolutely-positioned element now, confined to the bar's 36px strip. Verified the reveal geometry survives the move — the host rect becomes `764..800` instead of `36..800`, but the trigger band measures from `bottom`, which is unchanged.
- **`agentToolbarPinned` validation** — fixed to a `typeof === 'boolean'` check. Its neighbour `agentToolbarEnabled` on the line above has the same `!= null` shape; left alone as pre-existing rather than widening this diff.
- **Korean comments** — right, and I should have caught it: the file is new (restored from the pre-2026-08-15 toolbar), so the comments I carried over count as new. Translated. Only the two flagged; the untouched Korean comments elsewhere stay per the repo's don't-mass-rewrite rule.

`tsc` clean, 11,657 tests pass, and the live CDP run is back to 16/16.

openwong2kim:
@coderabbitai review

Issue #935, opened 2026-08-17
[Bug]: Roster status and "N need you" disagree with what panes are doing (working panes shown as waiting/complete, ended panes as running)


p-poppe:
### What happened

In a workspace running four Claude Code sessions side by side, the roster status and the "N need you" counter regularly disagree with what the panes are visibly doing — in several different directions at once, and the disagreement persists for minutes rather than flickering for a moment.

Three snapshots of the same workspace, minutes apart. "Pane shows" is what was on screen in that pane at that moment; "Roster says" is the label in the sidebar for that same session.

**Snapshot 1** — header: `3 need you`

| Pane | Pane shows | Roster says |
|---|---|---|
| w2-127 | conversation ended, `Crunched for 3m 54s` | Complete |
| w2-123 | `Garnishing…`, shell command running, tests just passed | **Waiting** |
| w2-131 | `Cooked for 1m 14s`, idle at the prompt | Waiting (fair) |

**Snapshot 2** — header: `1 need you`

| Pane | Pane shows | Roster says |
|---|---|---|
| w2-127 | `Running 1 shell command · 46s`, `Whirlpooling… (21m 27s)` | **Waiting** |
| w2-132 | conversation ended, `Crunched for 3m 54s` | **Running** |
| w2-123 | `Working… (3m 41s)`, commands running | **Complete** |
| w2-131 | `Cooked for 1m 14s`, idle | Waiting (fair) |

**Snapshot 3** — ~6 minutes later, header: `2 need you`

| Pane | Pane shows | Roster says |
|---|---|---|
| w2-127 | `Running 1 shell command · 1m 38s`, `Whirlpooling… (22m 18s)` | **Waiting** |
| w2-132 | ended | Idle (fair) |
| w2-123 | `Working… (thinking)`, reading files | **Complete** |
| w2-131 | idle | Waiting (fair) |

So across the three: a working pane reported as waiting, a working pane reported as complete, an ended pane reported as running. The counter inherits all of it — `3 need you` when at most one session actually wanted anything.

### Environment

- wmux 3.42.0, Windows 11.
- Claude Code panes, all started in bypass-permissions mode.
- Plugin-less hook install (no marketplace plugin): bridge wired for `Stop`, `SubagentStop`, `SessionStart`, plus the `AskUserQuestion` `PreToolUse`/`PostToolUse` pair. The wide permission-gate `PreToolUse` is deliberately not installed.
[...]

PR #939, opened 2026-08-19
Stop the roster from reading a working pane as waiting

openwong2kim:
Claude Code's footer says `bypass permissions on` for the whole turn in bypass-permissions mode. It is chrome, not a turn boundary. `AgentDetector` maps it to `waiting`, and the hook-authority veto that exists for exactly this reason only covered the notification — the status broadcast ran *before* it and was deliberately left ungated. So every mid-turn repaint wrote `waiting` onto the pane's roster row, and because `countNeedsAttention` counts `waiting`, each false read also inflated the "N need you" chip. That chip is the one signal that must never cry wolf.

## What changed

The rule now lives in one place. `HookSignalRouter.governsDetectorStatus` answers whether the pane's live hook bridge owns a given detector status, and both emission sites withhold the lifecycle status when it does:

- `PTYBridge` — local panes
- `DaemonNotificationRouter` — the packaged daemon path, which is what the report was running

On the daemon path a `decision: 'veto'` event is withheld too; the daemon already applied the same rule upstream.

## What is deliberately *not* withheld

- **Identity.** `agentName` / `agentSlug` still ride every broadcast, so a pane keeps its `(Claude Code)` auto-name.
- **`awaiting_input`.** Claude wires `PreToolUse` for `AskUserQuestion` alone, so the detector is the only source for ordinary approval prompts. Gating it would leave a pane blocked on a real prompt looking idle for the full 30-minute authority TTL.
- **`running`.** A working cue, not a turn boundary.
- **Ungoverned panes.** No hook bridge, or one gone quiet past the TTL, means the detector is still the backstop it has always been.

Prior art worth naming: OpenClaude already carried this exclusion in a comment — it re-renders its status bar every frame, so `bypass permissions on` was never wired to waiting for it. Claude Code in bypass mode has the same property; the exclusion just never reached it.

## Scope — two directions from #935 are NOT fixed here

The report showed three disagreements. This closes the false-alarm one. The other two are staleness with a different root and need their own change:

- a Stop hook's `complete` survives into the next turn
- a finished pane can read `running`

Both trace to `ActivityMonitor.onActive` being edge-triggered and re-armed only after five seconds of byte silence (`IDLE_DELAY_MS`), which a TUI painting a live timer never gives. Fixing it needs `ActivityMonitor` reachable from the hook path in both main and the daemon, so it is not a rider on this PR.

## Tests

- `HookSignalRouter` — the predicate: covers `waiting`/`complete`, spares `awaiting_input`/`running`, false without authority / slug / matching agent / past TTL
- `PTYBridge.lifecycle` — governed pane gets no `waiting` status but keeps identity; ungoverned pane still gets it
- `DaemonNotificationRouter.lifecycle` — same pair on the daemon path, plus `decision: 'veto'` now withholding the status

Both regression tests were confirmed to fail with the fix disabled. Two existing tests that pinned the old "status dot stays live" behaviour were updated rather than deleted.

Full suite: 11818 passed, 0 failed. `tsc` clean on the app, daemon and mcp projects.

Closes part of #935.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude Code pane and prompt detection for cursor-rendered text and omitted whitespace.
  * Hook-controlled panes now suppress detector-generated waiting and completion statuses while preserving agent identity information.
  * Approval prompts and `awaiting_input` statuses remain unaffected.
  * Panes without active hooks continue using detector-based status behavior until lifecycle control is established.
  * Vetoed lifecycle statuses are suppressed while identity metadata remains available.

* **Documentation**
  * Added guidance to update both app and daemon components when using mixed versions.

### DIFF
--- a/changelog.d/939.md
+++ b/changelog.d/939.md
@@ -13,3 +13,22 @@
   Note for anyone running a mixed pair: on the packaged daemon path the fix
   rides the daemon's own arbitration stamp, so an older daemon with a newer
   app does not get it until both are updated. (#935, #939)
+
+- **Claude Code panes are detected again.** The Claude gate needed its footer
+  spelled with literal spaces on a line of its own. The shipping product
+  (measured on v2.1.235) provides neither: spaces are cursor advances
+  (`ESC[1C`), so the footer strips to `bypasspermissionson`, and the whole
+  bottom region — input box, statusline, footer — arrives as one newline-free
+  run whose rows are placed with cursor-position escapes. The gate therefore
+  never opened on a real Claude pane, which quietly disabled every Claude
+  detector pattern, `waiting` and the approval `awaiting_input` regexes alike.
+  Whitespace is now optional in the footer fragments (the same treatment the
+  banner pattern already carried) and candidate lines are cut on row-position
+  escapes as well as newlines. (#935)
+
+- **A freshly launched agent no longer reads as busy.** Hook authority answers
+  "a bridge speaks for this pane", which is not the same as "the hook has
+  claimed this pane's lifecycle". A bridge that had sent only `SessionStart`
+  suppressed the detector's true "ready for input" read, leaving the pane on
+  the gate's one-shot `running` until its first turn ended. The status veto now
+  waits for the hook to actually take the lifecycle over. (#935)
--- a/changelog.d/939.md
+++ b/changelog.d/939.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **The roster no longer reads a working pane as waiting.** Claude Code's
+  `bypass permissions on` footer is on screen for the whole turn in
+  bypass-permissions mode, and the detector read every repaint of it as
+  "ready for input" — which put a false `waiting` on the pane's roster row
+  and inflated the "N need you" chip while the agent was still working. On a
+  pane with a live hook bridge the Stop signal now owns the lifecycle status
+  outright, on both the local and the packaged daemon path. Approval prompts
+  (`awaiting_input`) are unaffected, and a pane with no hook bridge keeps the
+  detector as its backstop. (#935, #939)
--- a/changelog.d/939.md
+++ b/changelog.d/939.md
@@ -8,4 +8,8 @@
   pane with a live hook bridge the Stop signal now owns the lifecycle status
   outright, on both the local and the packaged daemon path. Approval prompts
   (`awaiting_input`) are unaffected, and a pane with no hook bridge keeps the
-  detector as its backstop. (#935, #939)
+  detector as its backstop.
+
+  Note for anyone running a mixed pair: on the packaged daemon path the fix
+  rides the daemon's own arbitration stamp, so an older daemon with a newer
+  app does not get it until both are updated. (#935, #939)
--- a/src/daemon/hooks/HookIngest.ts
+++ b/src/daemon/hooks/HookIngest.ts
@@ -98,10 +98,17 @@ export interface HookIngestSession {
  *   - 'emit'  → fan out the notification AND the `agent.lifecycle` tee
  *   - 'dedup' → the other source already covered this turn: lifecycle tee with
  *               `decision:'dedup'`, NO notification
- *   - 'veto'  → detector-only, hook-governed pane: metadata/status dot only,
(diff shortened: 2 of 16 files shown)


openwong2kim:

This PR closes the false-alarm third of #935: Claude Code's `bypass permissions on` footer is permanent chrome in bypass-permissions mode, and the detector was reading every mid-turn repaint of it as `waiting`, which landed on the roster row and inflated the "N need you" chip — the one signal that must never cry wolf. The fix puts the rule in one place, a new `HookSignalRouter.governsDetectorStatus` predicate, and applies it at both emission sites (`PTYBridge` for local panes, `DaemonNotificationRouter` for the packaged daemon path), withholding only the lifecycle status while identity (`agentName`/`agentSlug`) still rides every broadcast. The carve-outs are right: `awaiting_input` stays live because Claude's hooks only wire `AskUserQuestion`, so the detector is the sole source for ordinary approval prompts, and `running` is a working cue rather than a turn boundary. Regression tests cover the predicate and both emission paths, and two existing tests that pinned the old "dot stays live" behavior were updated rather than deleted.

One note on CI: the `Baseline (windows-latest)` failure is in `src/main/git/__tests__/mergeSession.test.ts` (a 5s timeout followed by an EBUSY on temp-dir cleanup) — untouched by this diff and unrelated to the hook/roster code, so it reads as the known flaky Windows git test rather than something this PR introduced. The remaining two directions from #935 (a stale `complete` surviving into the next turn, and a finished pane reading `running` via the edge-triggered `ActivityMonitor`) are correctly scoped out to a follow-up that makes `ActivityMonitor` reachable from the hook path in both main and the daemon.

Priority: P2 — same tier as #935, since it repairs the attention signal users watch most closely.

openwong2kim:
## Review round applied (7a6495f0)

A three-model panel reviewed this. The sharpest finding was that the daemon-path test was green against a branch production never takes, which is worth writing down since it also explains why the rest changed.

**`hooks.rpc` returns at the daemon relay, above its own `touchAuthority` call.** So main's `HookSignalRouter` never gains authority over a daemon-served pane — on the packaged path `governsDetectorStatus` is always false there, and the daemon's `decision: 'veto'` stamp is the entire mechanism. The original test stubbed the predicate to `true` and passed without ever exercising it.

That led to three real defects in the first commit, all from the status gate re-judging an event the daemon had already arbitrated:

- **A hook-sourced completion could be suppressed by the authority the hook itself established.** The notification veto below already splits arbitrated from unarbitrated (`if (arbitrated) … else if …`); the status gate did not. It does now: arbitrated withholds on `veto` and nothing else, unarbitrated consults main's router.
- **`awaiting_input` had no local guard.** The daemon declines to stamp `veto` on it, but that guarantee lives in another process. An older daemon paired with this app could hide a real approval prompt for the full 30-minute TTL — worse than the bug being fixed. Now refused on both arms.
- **Contract drift.** `HookIngest`'s comment still said main treats `veto` and `internal` identically. Only `veto` withholds the status now.

Tests: the daemon case is now four — the M1 arm (`decision: 'veto'`, predicate false), the pre-M1 arm, the arbitrated-hook case, and the skew defence. Both new ones were confirmed to fail against the previous logic. Full suite 11821 passed, `tsc` clean on app, daemon and mcp.

Changelog now states that an older daemon with a newer app does not get this fix until both are updated.

### One finding answered in a comment, not in code

All three models flagged that the status withhold rides the same 30-minute authority TTL as the notification veto, so a bridge that goes quiet while its agent is still alive leaves the status stale until expiry.

Real, and left as is on purpose. The veto has always accepted that exact window on the louder surface, so making the status *less* trusting than the toast would be the odd choice. A shorter TTL is not available either: a single turn can run past twenty minutes with no bridge traffic at all on a turn-boundary-only hook install (#935 measured 21m), so a short window would simply restore the bug. Confirmed process death already releases authority ahead of the TTL via `expireAuthorityFor`. The reasoning is now in the source next to the predicate rather than only in a review thread.

### Still out of scope

`decision: 'internal'` writes its status through unchanged — the alarm rejected the candidate as not-a-real-turn-end, yet a hookless Claude pane still lets that reading reach the roster. Pre-existing, and it belongs with the two staleness directions tracked in #935 rather than here.

Issue #922, opened 2026-08-16
Browser caller scope is declared, not verified: no server-side binding from a caller identity to a workspace


openwong2kim:
## What

`callerScope` in `src/main/pipe/handlers/browser.rpc.ts` has been authoritative for browser target lookup under `mcp.mode: enforce` since #873. That closed the caller #810 described: an approved third-party that omits `workspaceId` is now refused rather than resolved against whichever surface happened to register first.

Two lanes in its table were left open on purpose, and they were left open for one shared reason — **nothing in the main process records which workspace a caller belongs to.** The code comment says so at `browser.rpc.ts:117-127`; this issue is where they go now that #113 is closed.

### 1. `declared` verifies presence, not ownership

A caller with a `clientName` that sends *a* `workspaceId` is scoped to whatever it named. There is no check that it named its own, because there is nothing to check against: `mcp.claimWorkspace` (`workspace.rpc.ts:68`) forwards the claim to the renderer and records no server-side association. Workspace ids are not secret either — `workspace.list` returns `{id, name}[]` and is gated on `workspace.read`, an ordinary declarable capability. So the bar for reaching a foreign workspace is `browser.*` **+** `workspace.read`, both grantable from the same approval prompt.

### 2. `legacy` (no identity envelope) is allowed and unscoped

A caller that sends no `clientName` at all lands in `legacy`, which is allowed and unscoped. That matches `PermissionEnforcer`'s own grandfather lane, so closing it here alone would start a second, inconsistent deprecation clock — it was deliberately deferred rather than overlooked.

## Why it needs its own issue

Both need a caller-identity binding that does not exist. #113 — the track that would have supplied it — is closed, so leaving these attached to #810 meant they had no route forward and #810 could not be closed on work that was genuinely finished.

Also relevant: `browser.open` / `browser.close` route through the renderer and don't select by ownership, so they are outside `callerScope` entirely. Worth deciding whether that stays true.

## Suggested direction — a shape, not a prescription

- **Record the workspace where a caller identifies itself.** `mcp.claimWorkspace` is the natural point; today it is a pass-through. A per-request derivation is what's needed, since `sendRpc` opens a fresh socket per call (`src/mcp/wmux-client.ts`) and there is no connection to bind to.
[...]

PR #941, opened 2026-08-19
fix(browser): scope a plugin's browser calls to the workspace hosting it

snowyukitty:
PR1 of the two-PR split agreed on #922
([comment](https://github.com/openwong2kim/wmux/issues/922#issuecomment-5337705295)).
Built on current `main`, since #928 moved these lines.

## The asymmetry this closes

The iframe plugin surface is the one caller class where **neither half of the
identity is supplied by the caller**:

- `clientName` is stamped main-side from the loaded manifest
  (`pluginHost.handler.ts`), never read from the bridge envelope;
- the workspace is derived by the host component that owns the frame —
  `PluginFrame.tsx` reads `activeWorkspaceId` from the store.

#719 already used that derivation on one side: `events.poll` is scoped, so a
plugin may not **observe** another workspace's lifecycle events. But a plugin's
own requests were forwarded with their params untouched, so `browser.*` landed
in the `declared` lane and got whatever workspace it named. The live rule was
*may watch only where the user is, may act anywhere* — which reads as an
oversight rather than a decision.

## What it does

- `RpcContext.hostedWorkspace`, written only by `RpcRouter.dispatch` from a new
  dispatch option next to `firstParty`. The plugin-host IPC handler is the sole
  call site that supplies it. A request envelope cannot reach it.
- The option is typed onto the `firstParty` variant alone and declared `never`
  on the other two; the runtime guard in `dispatch` rejects it off that lane as
  an internal provenance failure rather than dropping it into a context that
  would then read as merely unscoped.
- **Presence marks the caller class; the value is only the binding.** `string`
  binds, `null` means *the plugin host dispatched this and has no workspace to
  bind to*, absent means *not a plugin-host dispatch*. The host sends the key on
  every call.
- A `hosted` lane in `callerScope`, above `declared`, with the `pinned` lane's
  exact semantics — omitted `workspaceId` resolves to the binding, a mismatch is
  refused — plus two of its own: the mirror of `pinned`'s source check (`pinned`
  must arrive on the local wire, `hosted` must arrive in-process), and a refusal
  for the unbound case.
- Refusal copy follows the existing rule: name the refusal, never name another
  workspace. The audit entry may carry the ids; the message may not.

Because both halves are derived, this lane is an ownership fact where `declared`
can only be a claim.

## Deliberately not in this PR

- Moving the binding up to dispatch, and the `useRpcBridge.ts` renderer
  fallbacks (`pane.list`, `browser.open`) — that is PR2, and it gets its own
  design pass.
- The wire lanes. `declared` and `legacy` are untouched.

## One non-change worth flagging

`PluginFrame.tsx` reads the workspace **at request time** rather than taking it
as a dependency of the bridge effect. Making it a dependency is precisely what
broke the bridge in #928: the effect's cleanup closes the `MessagePort`, and
`load` never fires again for an unchanged `src`.

## Ceiling

This confines an **approved** plugin to the scope its approval implied. It is
not a defence against hostile code already running as the user, and it does not
touch the wire lanes, which still need the peer-credential primitive. The
binding is only as strong as the host's own derivation — not cryptographic
identity.

## Effect on shipped code

None. The bundled reference plugin (`examples/plugins/hello-panel`) calls
`workspace.list` and `pane.list` only — neither is in the browser lane table.
For any other plugin: omitting `workspaceId` on a `browser.*` call now
**resolves**, where `declared` previously refused it as `workspace-unresolved`;
naming a foreign workspace is refused where it previously succeeded.
Enforcement rides the existing `mcp.mode` gate (packaged `enforce`, dev
`shadow`), so there is no new mode or rollback switch.

## Review

Second commit is the result of an independent review (codex, `gpt-5.6-sol`,
read-only) that caught the lane **fail-open on its own empty case**: with the
binding omitted when the host had no active workspace, the context was
indistinguishable from any other in-process caller and the plugin fell through
to `declared`, where its own `workspaceId` is accepted — an unbound plugin would
have been *less* confined than a bound one. That is what the tri-state above
fixes. The same review narrowed two over-broad comments and flagged the source
check as an invariant backstop rather than production telemetry; both are in.

## Verification

- `npx tsc --noEmit -p tsconfig.json` — clean. `eslint` on the changed files —
  clean.
- `vitest run src/main/pipe src/renderer/plugins src/main/ipc/handlers/__tests__/pluginHost.handler.test.ts`
  — 36 files, 755 tests, green.
- Wider sweep `vitest run src/main src/shared src/preload` — 4658 passed. Three
  files fail on this machine (`logSink`, `cliShim`, `mergeSession`); `logSink`
  fails identically on a clean `origin/main` checkout here and the other two
  pass in isolation, so they are load/environment noise, not this change.
- The unbound test was run against the pre-fix implementation: it fails there
  with `expected true to be false` — the call succeeded and was scoped to the
  workspace the plugin named — and passes here.

New coverage:

- `browser.rpc.test.ts` — seven lane-table cases (resolve-on-omit, accept-own,
  **refuse-foreign**, unbound, blank, wire-context-carrying-the-field, and
  pinned-before-hosted precedence) plus three enforce-path cases through the
  registered handler, asserting the refusals land **before** any target lookup,
  wake, or lease, and that the audit entry is written.
- `RpcRouter.dispatchProvenance.test.ts` — the option reaches the context (with
  a positive control so the test cannot pass with the feature removed), an
  envelope field never does, `null`/`''`/whitespace all normalise to `null`
  while staying distinguishable from absent, and off-lane cases hard-reject.
- `pluginHost.handler.test.ts` (new file) — the IPC seam: forwarded as an option
  not a param, params left verbatim, absent/blank host workspace forwarded as an
  explicit `null`, non-string refused.
- `PluginFrame.bridge.dynamic.test.tsx` — mounted: the same port reports `ws-1`
  then `ws-2` across a switch with no reload, pinning both the binding and the
  #928 invariant.

## Summary by CodeRabbit

- **Security**
  - Restricted hosted plugin browser requests to the host’s active workspace.
  - Rejected requests with missing, mismatched, or invalid workspace context.
  - Prevented untrusted request data from overriding the host workspace.

- **Bug Fixes**
  - Preserved correct workspace binding when switching workspaces.
  - Improved audit details for rejected hosted plugin requests.

- **Tests**
  - Added coverage for workspace inheritance, rejection scenarios, provenance validation, and workspace switching.

### DIFF
--- a/changelog.d/941.md
+++ b/changelog.d/941.md
@@ -1,0 +1,15 @@
+### Security
+
+- **A UI plugin's browser calls now stay in the workspace you are looking at.**
+  An approved iframe plugin could already only *watch* the active workspace —
+  its event feed has been scoped since #719 — but its own `browser.*` calls
+  were forwarded with whatever workspace id the plugin chose, so it could drive
+  a browser surface in a workspace you were not in. The workspace the plugin
+  host is showing now travels with each request on its own channel, out of the
+  plugin's reach, and a browser call that names a different workspace is
+  refused; one that names none resolves to the workspace hosting the plugin
+  instead of being rejected as unresolved. A plugin whose host has no active
+  workspace is refused rather than falling back to naming its own. This
+  confines an approved plugin to the scope its approval implied — it is not a
+  defence against hostile code already running as your user, and callers on the
+  local RPC wire are unchanged. (#941)
--- a/changelog.d/941.md
+++ b/changelog.d/941.md
@@ -1,15 +1,18 @@
 ### Security
 
-- **A UI plugin's browser calls now stay in the workspace you are looking at.**
-  An approved iframe plugin could already only *watch* the active workspace —
-  its event feed has been scoped since #719 — but its own `browser.*` calls
-  were forwarded with whatever workspace id the plugin chose, so it could drive
-  a browser surface in a workspace you were not in. The workspace the plugin
-  host is showing now travels with each request on its own channel, out of the
-  plugin's reach, and a browser call that names a different workspace is
-  refused; one that names none resolves to the workspace hosting the plugin
-  instead of being rejected as unresolved. A plugin whose host has no active
-  workspace is refused rather than falling back to naming its own. This
-  confines an approved plugin to the scope its approval implied — it is not a
-  defence against hostile code already running as your user, and callers on the
-  local RPC wire are unchanged. (#941)
+- **A UI plugin can no longer drive a browser surface in a workspace you are not
+  in.** An approved iframe plugin could already only *watch* the active
+  workspace — its event feed has been scoped since #719 — but its own
+  `browser.*` calls were forwarded with whatever workspace id the plugin chose.
+  The workspace the plugin host is showing now travels with each request on its
+  own channel, out of the plugin's reach: a browser call that names a different
+  workspace is refused, one that names none resolves to the workspace hosting
+  the plugin instead of being rejected as unresolved, and a plugin whose host
+  has no active workspace is refused rather than falling back to naming its own.
+  This covers the browser methods that resolve a target — navigation,
+  evaluation, extraction, input, capture and the rest. Opening and closing
+  browser surfaces (`browser.open`, `browser.close`) resolve their workspace on
+  a different path and are **not** covered yet; that is the next step on #922.
+  Either way this confines an approved plugin to the scope its approval implied
+  — it is not a defence against hostile code already running as your user, and
+  callers on the local RPC wire are unchanged. (#941)
--- a/src/main/audit/shadowRejectionLog.ts
+++ b/src/main/audit/shadowRejectionLog.ts
@@ -82,6 +82,7 @@ export type BrowserScopeShadowReason =
   | 'pinned-source-unqualified'
   | 'pinned-workspace-mismatch'
   | 'hosted-source-unqualified'
+  | 'hosted-workspace-unbound'
   | 'hosted-workspace-mismatch'
   | 'workspace-unresolved';
 
--- a/src/main/audit/shadowRejectionLog.ts
+++ b/src/main/audit/shadowRejectionLog.ts
@@ -81,6 +81,8 @@ export type BrowserScopeShadowReason =
   | 'caller-origin-unsupported'
   | 'pinned-source-unqualified'
   | 'pinned-workspace-mismatch'
+  | 'hosted-source-unqualified'
+  | 'hosted-workspace-mismatch'
   | 'workspace-unresolved';
 
 /**
@@ -95,6 +97,12 @@ export interface BrowserScopeShadowEntry {
   reason: BrowserScopeShadowReason;
   requestedWorkspaceId?: string;
   pinnedWorkspaceId?: string;
+  /**
+   * #922 — the workspace the plugin host derived for the caller. Recorded for
+   * the same reason as `pinnedWorkspaceId`: this is a local operator-owned
+   * audit file, so it may hold what a refusal message must never disclose.
+   */
+  hostedWorkspaceId?: string;
 }
 
 export type BrowserScopeShadowInput = Omit<BrowserScopeShadowEntry, 'entryKind' | 'ts'>;
--- a/src/main/ipc/handlers/__tests__/pluginHost.handler.test.ts
+++ b/src/main/ipc/handlers/__tests__/pluginHost.handler.test.ts
@@ -73,13 +73,16 @@ describe('plugins:rpc host workspace binding', () => {
     expect(request.params).toEqual({ hostedWorkspace: 'ws-forged' });
   });
 
-  it('omits the option when the host has no workspace yet', async () => {
+  it.each([undefined, '', '   '])('says "hosted, unbound" rather than going quiet (%p)', async (value) => {
     const rpc = setup();
 
-    await rpc({}, 'demo', 'workspace.current', {}, undefined);
+    await rpc({}, 'demo', 'workspace.current', {}, value);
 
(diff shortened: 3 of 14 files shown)


openwong2kim:

This is PR 1 of the two-PR split agreed on #922, and it closes the asymmetry where an iframe plugin could only *watch* the active workspace (#719) but its `browser.*` calls were forwarded with whatever `workspaceId` the plugin itself named. The mechanism is sound: `hostedWorkspace` travels as a `RpcRouter` dispatch option on the `firstParty` lane only, written solely by the plugin-host IPC handler (the workspace the host component is showing, read at request time), so a request envelope can never populate it, and the new `hosted` lane in `callerScope` applies the pinned lane's exact semantics — omitted `workspaceId` resolves to the binding, a foreign one is refused, and the unbound case is refused rather than demoted to `declared`, which is the fail-open the independent review caught.

The diff itself is tight and appropriately paranoid. I checked the load-bearing points: the tri-state (`string` binds / `null` means hosted-with-nothing-to-bind / absent means not a plugin host) is preserved end to end — handler trims blanks to `null`, router guards the option off the `firstParty` lane with a loud provenance failure, and the `browser.rpc` backstop re-checks the source for hand-built contexts. Refusals land before target lookup, wake, or lease, and the refusal copy names the lane without leaking another workspace's id. Keeping `PluginFrame`'s workspace read at request time (not as an effect dependency) is the right call given #928, and the dynamic bridge test pinning `ws-1 → ws-2` on the same port protects exactly that invariant. Verification is thorough — 755 targeted tests plus the wider sweep, with the pre-fix failure of the unbound case demonstrated.

Marking P1: this is a confinement fix for the plugin surface on the enforce path, the second half of #922's plan, and it should land ahead of PR2 rather than wait. Two notes for the merge, neither blocking: `changelog.d/941.md` says the workspace "travels with each request on its own channel" — worth confirming the shipped release notes keep the same ceiling statement as the PR description (approved-plugin confinement, not hostile-code defence); and the renderer fallbacks (`pane.list`, `browser.open` in `useRpcBridge.ts`) that resolve a workspace without reaching `scopeFor` stay uncovered here, which PR2 needs to pick up.

snowyukitty:
Read the bot review before merge this time, and checked it rather than took it.

**The reachable half is real, and the changelog was over-claiming.** `browser.open`
and `browser.close` (`browser.rpc.ts:706`, `:738`) read `params.workspaceId`
directly, take no `ctx`, and never reach `scopeFor` — so a hosted plugin can
still open or close a surface in another workspace. `browser.navigate` is an
ordinary declarable capability (`methodCapabilityMap.ts:279,282`), so that is
one approval away. Fixed the entry in `4ef94c09`: it now names what it covers
(the target-resolving methods) and says those two are not covered yet and are
the next step on #922.

**One correction to the finding:** `browser.tabs` is not in that set.
`methodCapabilityMap.ts:278` gates it on `wmux.internal`, which a plugin cannot
declare, so it is not third-party reachable and it is not part of the exposure.

**The code half stays in PR2 on purpose**, per your Q4 answer on the issue —
`browser.open` / `browser.close` get covered when the binding moves up, not by
widening this lane table. Same for the `useRpcBridge.ts` fallbacks you named;
`pane.list` is the one there that reads on user data.

Nothing else changed — the fix commits are untouched, only the changelog entry.

snowyukitty:
`validate` is red on the latest push and it is the `cliShim` flake, not this
change. Stating the evidence rather than asserting it, since "CI is flaky" is
exactly the claim that should not be taken on trust:

- The failure is `cliShim.test.ts > PATH edit (live registry, sandbox key) >
  FullLanguage: add appends without destroying existing entries`, *killed by
  SIGTERM after 20000 ms* — a timeout in the live-registry PowerShell harness.
  1 failed, 11680 passed.
- **The only difference between the green run and the red one is prose.**
  `validate` passed on `03adb870` ([run
  32225897851](https://github.com/openwong2kim/wmux/actions/runs/32225897851));
  `4ef94c09` changes `changelog.d/941.md` and nothing else —
  `git diff --stat 03adb870 4ef94c09` is 1 file, +16/−13, all inside that
  markdown.
- Same test failed in my local wide sweep under parallel load and passed when
  run alone. It is the harness #576 described and closed as
  "harness, not a regression"; the timeout budget is the recurring cause.
- Nothing in this PR touches PATH, the registry, the CLI shim, or its timeouts.

Every other check is green: baselines on all three platforms, bench, and
CodeRabbit. I cannot re-run the job from here — it needs your click, or it will
clear on the next push.

openwong2kim:
Non-blocking, noted for PR2 rather than this merge:

`5a04cb87` added `src/main/mcp/wireProvenance.ts`, but no production code imports it — its only importer is its own test. The code this PR hardens (`browser.rpc.ts:33`, `PermissionEnforcer.ts:61`) still runs the pre-existing `src/main/mcp/rpcProvenance.ts`, and the two copies differ: the new one excludes `operator: true` and is a type predicate; the live one does neither.

So `wireProvenance.test.ts`'s first case (`operator: true` → `false`) green-lights a property the running predicate does not have. Not exploitable today — an operator dispatch gets `firstParty: true` in `RpcRouter.dispatch`, which the live predicate already excludes, and the dispatch-options union makes `operator` + `externalWire` unrepresentable — but it leaves two same-named provenance predicates where only the dead one is tested, and the next importer has a coin-flip.

Suggested cleanup when PR2 touches this area: delete `wireProvenance.ts`, re-point the test at `rpcProvenance`, and if the `operator` exclusion / type-predicate shape is wanted, fold it into the file that actually runs.

The second test in that file (AST scan pinning `pipe/PipeServer.ts` as the sole production writer of `externalWire: true`) is a genuinely good invariant test and survives the re-point unchanged — worth keeping either way.

PR #943, opened 2026-08-19
fix(status): re-arm the activity cycle on turn ends, and give Claude a running signal

p-poppe:
Reopened with the handoff branch underneath and the `running` signal on top, so direction 2 of #935 lands as one change: the primitive, the local-mode consumer that needs it, and the signal that actually exercises it.

Authorship: commits 1–2 are mine (`ActivityMonitor.endTurn` + the daemon seam), commit 3 is **@openwong2kim**'s local-path wiring and middleware regression test from `handoff/943-local-turn-end`, kept intact, commit 4 is mine (the detector pattern).

## What's in it

1. **`ActivityMonitor.endTurn(ptyId)`** — re-arms the activity cycle on an authoritative turn end. Threshold-based on purpose, unlike `beginTurn`: a turn end is not proof that later output means a new turn, so the next *genuine burst* reports running and idle chrome does not.
2. **Daemon seam** — `DaemonPTYBridge.noteAgentStatus` calls it on the terminal branch.
3. **Local seam (his)** — `PTYBridge.noteTurnEnd`, called from the hook path (`hooks.rpc` turn boundary + `awaiting_input`, wired through `main/index.ts`) and from the un-vetoed detector path, sitting below the #939 veto so a governed pane's footer `waiting` can't become a cycle boundary. This is where the stuck cycle was real: the local bridge feeds the monitor ungated, so a live elapsed-time counter reschedules the idle timer forever and the cycle never re-arms on its own.
4. **The `running` signal** — `esc to interrupt` as Claude Code's `running` pattern, plus a new `AgentPattern.unless` abstention guard.

## Why (4) is the part that makes (1)–(3) matter

Without it there is still nothing that says "a new turn started" when the agent starts one itself. `checkGates`' one-shot aside, **no agent had a text pattern producing `running`** — so on a plugin-less install an autonomous turn (a background task finishing and the agent resuming) had no submitted input, no per-tool hook, and no detector signal. The re-armed cycle would sit there waiting for a promotion that never came.

`esc to interrupt` is the one piece of Claude Code chrome that exists only while a response is in flight — the file already said as much in the comment that removed it from the *waiting* list, where it was plainly wrong.

## The measurements behind the choices

Sixteen real daemon captures from my machine, 73 MB of Claude Code output in bypass mode:

| measurement | count |
|---|---|
| `esc to interrupt`, raw bytes | 462 |
| `esc to interrupt`, after ANSI strip | **462** |
| hint-bearing flattened lines | 447 |
| …of those, also carrying `shift+tab to cycle` | 138 |
| …of those, also carrying `bypass permissions on` | 153 |

So the anchoring is safe (the cursor-forward hazard never lands inside the phrase) and the merged-frame problem is real, which is what `unless` exists for: both footer markers are covered by `CLAUDE_PROMPT_RE`, because scoping on one leaves the other free to merge in. Per your reasoning I took **option 1** — abstain on a merged line rather than let narrower evidence win — so the failure mode self-heals within a spinner tick instead of reopening the daemon gate on a possibly-finished pane.

**Replay validation** (not a fixture): feeding all captures through the patched detector in 4 KB chunks — 15 non-empty captures, 10 gating as Claude Code — yields **19 `running` emissions and 18 `waiting`**, from 462 raw hint occurrences. The dedup ledger collapses the per-second repaints, so the pattern is not a notification source; it alternates with the terminal states as turn boundaries should.

## One judgement call I'd flag for you

**R3 changed shape.** `REGRESSION (R3): does NOT match "esc to interrupt"` asserted `expect(cb).not.toHaveBeenCalled()`, but its own title says *"Claude in-flight hint, not idle"*. I rewrote the assertion to that intent — never `waiting`, never `complete`, now `running` — rather than deleting or weakening it. If you'd rather R3 stay literal and the new behavior be locked by its own test, say so and I'll split them.

## Test plan

- `ActivityMonitor.test.ts` — `endTurn` block (6 cases), including the regression: 30 s of per-second sub-threshold repaints never re-arm on their own, then `endTurn` + a real burst reports running; sub-threshold chrome after `endTurn` does not.
- `PTYBridge.turnEnd.test.ts` and `hooks.rpc.turnEnd.test.ts` — his middleware-level regression (fails with the `endTurn` call removed).
- `AgentDetector.test.ts` — 8 new cases: hint alone promotes; the hint survives cursor-forward spacing; abstains on either merged footer; the merged line still reports `waiting` exactly as before; a hint-only line one tick later promotes; an ungated pane never fires; repeated hint lines collapse to one emission.
- `src/main/pty/__tests__` 21 files / 333 tests pass. `npx tsc --noEmit` clean. `eslint` on the touched files reports the same 3 errors + 1 warning as `main` does with the patch stashed (pre-existing: a control-char regex and two empty arrow functions).
- Manual verification: **not done** — no Electron binary in this sandbox. Worth a human eye: a pane that finishes a turn and immediately starts another should go amber again rather than sitting green, and a pane parked on an approval should not.

Closes the direction-2 half of #935; directions 1 (yours, #939) and 3 (level semantics — what evidence keeps `running` alive) are untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved activity status handling when an agent turn completes or awaits input.
  * Subsequent output now correctly reports an active or running status.
  * Prevented idle UI redraws from incorrectly changing activity status.
  * Improved detection of in-progress Claude responses, including the “esc to interrupt” indicator.
  * Restored reliable idle detection after completed turns.

* **Tests**
  * Added coverage for turn completion, activity reactivation, idle timers, status detection, and independent activity tracking.


openwong2kim:

This PR takes direction 2 of #935 — a pane sticking at `complete` through its next turn — by adding `ActivityMonitor.endTurn(ptyId)` and calling it from `DaemonPTYBridge.noteAgentStatus()` where the authoritative terminal status is recorded. The root cause it targets is real and well isolated: `onActive` fires once per cycle and the cycle only re-arms inside the idle timer after `IDLE_DELAY_MS` of complete byte silence, which a TUI painting a live elapsed-time counter never allows, so the `complete` written by the Stop hook survives the pane's whole next turn and an agent-initiated turn never reports `running` at all. The choice to re-arm the threshold path rather than reuse `beginTurn()` is correct — a turn end is not proof that subsequent output means a new turn, so sub-threshold idle chrome stays inert while the next genuine burst reports running.

The daemon-only scope is sensible given #939 is open in the local `PTYBridge` file; the local-mode half should follow as agreed rather than ride along here. Test coverage hits the regression case exactly (thirty seconds of sub-threshold repaints never re-arming on their own, then `endTurn` plus a real burst reporting running), along with the late-idle-timer clear, per-pty isolation, and post-`stop()` no-op, and the changelog fragment is in place. CI is still running at triage time; the change is purely additive on an internal main-process helper plus one call site, with no external contract touched. Marking P2 to match #935.

p-poppe:
Withdrawing this — the automated review flag is correct and the approach cannot work. Closing rather than patching it, because the mechanism is wrong rather than incomplete.

What I missed: `noteAgentStatus` sets `explicitTerminalStatus = true` on the same branch where I call `endTurn()`, and the `onActive` handler returns immediately while that flag is set. So the re-armed cycle fires a callback that is then dropped, and nothing reaches the status write. The patch is inert in the daemon path.

The obvious "fix" — letting a qualifying burst reopen the gate — is worse: `DaemonPTYBridge.turnPriority.test.ts` locks exactly that behavior with *ignores a later full-screen repaint entirely* and the per-status `treats %s as terminal` cases. A full-screen redraw is well over the 2 KB threshold, so byte volume cannot distinguish a repaint from the first output of a new turn. That gate is a decision, not an oversight, and reopening it would resurrect the stale-`running`-from-repaint bug those tests were written for.

Which means direction 2 of #935 needs an authoritative *turn started* signal, not a re-armed heuristic. I've put the analysis and a concrete proposal in #935 rather than guessing at a second mechanism here.

Sorry for the round trip — I should have read the turn-priority tests before writing the patch, not after.

openwong2kim:
## Live dogfood: the detector half is inert in exactly its target scenario

I ran this branch live before review (isolated dev instance, real `claude` in bypass mode inside a daemon pane), because the first iteration of this PR died on an integration gate that unit tests couldn't see — and the same class of failure is back, one layer deeper.

**Symptom.** Submitted turns are fine: `waiting → running → complete` transitions exactly as intended. Then I reproduced the actual #935 direction-2 shape — asked Claude to launch `sleep 25 && echo WAKE` as a background task and end its turn, so the completion resumes it with no submitted input. The screen clearly shows the autonomous turn (spinner, then `RESUMED`), but the pane status **stays `complete` for the whole turn**. No `running` transition — the exact bug this PR targets, still present, observed on this branch's build.

**Root cause.** The emission dedup stores `lastEmittedFor[agent:status] = match[0]`, and for `/esc\s*to\s*interrupt/` the matched text is the constant `esc to interrupt`. Both reset paths are closed in the autonomous case:

- `noteInput` → never fires (no submitted input — that's the premise of the scenario), and
- the daemon's `onActive` handler → returns on its first line while `explicitTerminalStatus` is set, so its `resetEmissionState()` call is unreachable.

So once the detector has emitted `running` in *any* earlier turn, every later autonomous hint dedupes against the same constant and is swallowed before `noteAgentStatus('running')` can reopen the gate. Minimal repro (fails on this branch):

```ts
const { det, cb } = claudeGated();
det.feed('✻ Cooking… (0m 12s · esc to interrupt)\n');   // turn 1 → running
// turn 1 ends via Stop hook: detector untouched, no reset path runs
det.feed('✻ Churning… (0m 1s · esc to interrupt)\n');   // autonomous turn 2
// expected ['running', 'running'] — actual ['running']
```

The 73 MB capture measurement couldn't catch this: it validates the pattern text, not the dedup lifecycle. And the new tests all feed hints within a single turn, so `lastEmittedFor` never crosses a turn boundary.

**Suggested direction.** `noteAgentStatus`'s terminal edge is already the place that re-arms the byte cycle (`endTurn()`); clearing the detector's `:running` dedup keys there too would make a turn end re-arm *both* heuristics symmetrically. I'd scope it to `:running` rather than a full `resetEmissionState()` — a full clear lets an idle footer repaint re-emit `waiting` over a fresh `complete`, which trades a silent failure for a noisy one on every turn end.

Happy to share the live rig details if useful — the repro is deterministic.

openwong2kim:
Closing this in favour of #965, which takes a different route to the same bug. The reasoning is worth recording, because two of the three things that were wrong here could only be found by running it.

**The re-arm was inert, as you said.** `noteAgentStatus` calls `endTurn()` on the same branch that sets `explicitTerminalStatus`, and the daemon skips `feed()` entirely while that flag is set (`DaemonPTYBridge.ts:359`) and returns on `onActive`'s first line (`:243`). Your read was right.

**The detector half had a third problem underneath the dedup one.** I measured `/esc\s*to\s*interrupt/` against 41 MB of live pane output from this machine's daemon ring buffers — 448k lines across six Claude Code panes — and it occurs **zero** times. Current Claude Code does not paint that hint at all; the working chrome is the spinner verb, the token counter and `thinking with max effort`. So the pattern was anchored on a string the product no longer draws, which is also why the dedup lifecycle bug I reported earlier was the only thing that showed up in the dogfood: the pattern was never going to fire twice regardless.

That is a general argument against the approach rather than against this patch. Chrome strings are version-specific and agent-specific, and this one had already moved.

**What #965 does instead** is let byte throughput promote a settled pane, which covers every agent and does not depend on any string. The numbers separate cleanly on a real pane at 140x41 — typing peaks at ~1.5 KB per 3 s window, working exceeds 2.7 KB, an idle pane emits ~0.5 KB per 20 s — because that TUI repaints differentially, about 13 bytes per keystroke rather than a frame. Three guards keep it honest: a cool-down after the turn end, a keystroke-quiet window, and the existing resize-redraw window.

Your `ActivityMonitor.endTurn()` is kept as written, with its rationale intact — it turned out to be exactly the right primitive once there was a path where it could run. The multi-model review of #965 also independently landed on the failure your changelog was describing (the cycle re-arming only after byte silence a live counter never gives), which is a good sign the diagnosis in this PR was sound even where the mechanism was not.

Thanks for chasing it, and for withdrawing it yourself rather than patching around the inert half.

PR #945, opened 2026-08-19
Pin only the textarea during IME composition, never the inline preedit

### LINKED ISSUE #942: [Bug]: Korean IME — the composing syllable is painted over the previously committed one (display-only regression from the #875 / #888 composition-anchor pin)
## What happened

Typing Korean in a wmux pane makes syllables disappear on screen while typing.
Typing `대한민국` shows `대한국`: as each syllable commits and the caret advances,
the next composing syllable is painted back on top of the syllable that just
committed, hiding it.

**The buffer content is fine — this is a painting bug.** Ending the composition
with space or Enter makes the full `대한민국` appear correctly. So nothing is lost
downstream; what breaks is the ability to read what you are typing, which for
Korean is every single keystroke rather than an occasional candidate list.

That self-healing on commit is itself diagnostic: it rules out the PTY and parser
path entirely and localises the fault to the composition overlay — which is
exactly the surface the `.xterm-helpers` transform moves.

This is Korean-specific in a way #874 / #901 are not. Korean Microsoft IME
(`00000412`, TIP `{a028ae76-01b1-46c2-99c4-acd9858ae02f}`) draws **no candidate
window** during ordinary typing — the syllable under composition is rendered
*inline at the caret* by xterm's `.composition-view`. So the composition overlay
is not a floating list beside the text; it *is* the text the user is looking at.

Downgrading to **v3.41.1** — the last release before PR #875 landed — makes the
symptom disappear completely. Same machine, same IME, same panes, no other change.

## Steps to reproduce

1. Windows 11 with Korean Microsoft IME as an input method.
2. Launch wmux v3.44.0 and open a pane (Windows PowerShell).
3. Switch to Korean input (Han/Eng key) and type `대한민국`.
4. Observe `대한국` — the composing syllable sits on top of the previous one.
5. Press space or Enter. The composition ends and `대한민국` renders correctly,
   confirming the buffer was right all along.

Reproduced across three separate panes in the same session, at three different
zoom levels (cellHeight 16 / 17 / 18 px).

## Expected behavior

All four syllables stay visible as they are typed: `대한민국`. The inline preedit
box should follow the caret as it advances, which is what xterm's
`CompositionHelper` does on its own.

## Version bisect

| Version | Released | Result |
|---|---|---|
| v3.41.1 | 2026-08-12 | **clean** — verified after downgrade |
| v3.42.0 | 2026-08-13 | not tested (contains #875) |
| v3.43.0 | 2026-08-16 | not tested (contains #888) |
| v3.44.0 | 2026-08-18 | **broken** — verified, logs below |

v3.41.1 is the last release before #875 merged (2026-08-13 11:39Z, shipped in
v3.42.0). #888 merged 2026-08-14 09:14Z, shipped in v3.43.0. So the regression
window is v3.42.0–v3.44.0; I confirmed only the two endpoints, not which of the
two PRs introduced it.

## Environment

- wmux version: **3.44.0** (broken) / **3.41.1** (clean)
- Install method: Setup.exe (direct download)
- Shell in the affected pane: Windows PowerShell 5.1
- Windows version: Windows 11 Enterprise 25H2 (build 26200.9106)
- AI agent involved: Claude Code was running in the workspace
- Input method: Microsoft IME (Korean), keyboard layout `00000412`

## Diagnostic log — the field evidence #888 asked for

PR #888 shipped the `[wmux:ime-anchor2]` diagnostic with an explicit condition
attached to it:

> **Known limit, accepted deliberately:** dwell time cannot distinguish the caret
> from any cell the cursor parks on for ≥ 32 ms. […] If misplacement persists with
> `src=instant` and a large `held`, the next iteration is revisit-frequency
> tracking — gated on that evidence, not shipped speculatively.

From `%APPDATA%\wmux\logs\main-2026-08-19.log`, 41 `compositionstart` records
across three panes while typing Korean on v3.44.0:

```
[wmux:ime-anchor2] pty=f68ecdf1 compositionstart ybase=0 ydisp=0 cursor=(2,40)  sel=(2,40)  src=instant held=86ms    restAge=86ms   cellHeight=16.00 correction=(0.0,0.0)
[wmux:ime-anchor2] pty=f68ecdf1 compositionstart ybase=0 ydisp=0 cursor=(7,40)  sel=(7,40)  src=instant held=131ms   restAge=150ms  cellHeight=16.00 correction=(0.0,0.0)
[wmux:ime-anchor2] pty=f68ecdf1 compositionstart ybase=0 ydisp=0 cursor=(9,40)  sel=(9,40)  src=instant held=396ms   restAge=396ms  cellHeight=16.00 correction=(0.0,0.0)
[wmux:ime-anchor2] pty=f68ecdf1 compositionstart ybase=0 ydisp=0 cursor=(2,39)  sel=(2,39)  src=instant held=20214ms restAge=-1ms   cellHeight=16.00 correction=(0.0,-16.0)
[wmux:ime-anchor2] pty=e7acce87 compositionstart ybase=0 ydisp=0 cursor=(23,0)  sel=(23,0)  src=instant held=307ms   restAge=307ms  cellHeight=17.00 correction=(0.0,0.0)
[wmux:ime-anchor2] pty=e7acce87 compositionstart ybase=0 ydisp=0 cursor=(25,0)  sel=(25,0)  src=instant held=281ms   restAge=281ms  cellHeight=17.00 correction=(0.0,0.0)
[wmux:ime-anchor2] pty=e7acce87 compositionstart ybase=0 ydisp=0 cursor=(27,0)  sel=(27,0)  src=instant held=427ms   restAge=427ms  cellHeight=17.00 correction=(0.0,0.0)
```

Aggregates over all 41 records:

- `src=instant` in **41/41** — the resting-cell tracker never promoted a cell, so
  #888's resting logic is not what is selecting the anchor here.
- `correction=(0.0, 0.0)` in **40/41** — the #875 scroll-offset correction is a
  no-op in this workload (`ybase == ydisp` throughout; nothing was scrolled back).
- One outlier: `src=instant held=20214ms restAge=-1ms correction=(0.0,-16.0)` —
  `src=instant` with a 20-second `held`, which is the exact residual case #888
  said it would wait for field evidence on.
- The caret column advances between records (`23 → 25 → 27`), i.e. each Korean
  syllable is its own composition session and each one pins to a correct cell.

That last point is why I do not think a wrong *anchor cell* is the story here, and
why the `correction=(0,0)` records are not evidence against the report.

## Why Korean breaks where the CJK candidate window does not

`imeAnchor` applies its correction as a `transform` on `.xterm-helpers`, and that
container holds **two** children with different requirements (the diagram is from
#875's own description):

```
.xterm-screen
  +-- .xterm-helpers                 <- transform: translate(dx, dy)   [wmux]
  |     +-- textarea                 <- style.top / left               [xterm]
  |     +-- .composition-view        <- style.top / left               [xterm]
  +-- canvas                         <- painted cursor                 [webgl]
```

- The **textarea** is what Chromium reports the caret rect from, so it is what the
  candidate window follows. Pinning it for the duration of a composition is the
  fix #875/#888 wanted, and it is right for Chinese/Japanese.
- The **`.composition-view`** is the visible preedit box.
  `CompositionHelper.updateCompositionElements` deliberately re-positions it to the
  **live** cursor on every `compositionupdate` (plus a recursive `setTimeout`),
  precisely so an inline preedit tracks the caret.

`imeAnchor.sync()` sets `desired = frozen ?? paintedCursorPosition(...)`, so while
`frozen` is non-null the transform is `frozen − actual`. Once the caret advances
past the pinned point, that difference pulls the whole container — the visible
preedit box included — backwards onto already-committed cells. For a floating
candidate list that is invisible; for an inline preedit it puts the glyph the user
is typing on top of the neighbouring syllable, which is the reported symptom.

In other words: the pin was designed for a surface that should *not* follow the
caret, but it is being applied to a container that also holds a surface that
*must* follow the caret.

This predicts the observed behaviour exactly, including the recovery: on
`compositionend` `imeAnchor` clears `frozen` and re-syncs, xterm hides the
preedit box, and webgl repaints the real cells — which is why space or Enter
restores the correct text every time.

Two smaller things noticed while reading the module:

- `sync()` already knows that `textarea.style.height` is repurposed during a
  composition (it only samples the cell height when `frozen === null`), but
  `readActual()` still reads `style.left` / `style.top` unconditionally during a
  composition.
- The diagnostic only fires in `onCompositionStart`, before xterm has written the
  composition-adjusted child styles. Any `dx` that develops on subsequent
  `compositionupdate` / `onRender` syncs is never logged — which is why a log full
  of `correction=(0,0)` can coexist with visible misplacement. Logging the
  correction on `compositionupdate` and `compositionend` too would make this class
  of report self-diagnosing.

## Suggested direction

Split the two surfaces instead of transforming their shared parent: pin the
textarea (candidate-window anchor) and leave `.composition-view` on xterm's own
live-cursor positioning. If keeping one transform is preferable, gate the pin on
whether a candidate window is actually in play — an inline-preedit IME with no
candidate list gains nothing from it and is actively harmed.

## Related

- #874 / #875 — IME candidate window not following the cursor (Chinese); introduced the composition pin.
- #888 — follow-up, resting-cell anchor; retagged the diagnostic to `ime-anchor2`.
- #901 — still-open Chinese candidate-window misplacement on v3.42.0 in a Kimi Code pane.
- #167 — older "IME input wipes the already-typed line", different mechanism but same visible family.

No existing issue covers Korean: a repo search for `IME` returns #874, #901, #929,
#678, #167, #252, #171 and #22, none of them Korean-specific.

## Confirmation

- [x] I am on Windows 10 or 11.

openwong2kim:
## Summary

Fixes #942 — on Korean Microsoft IME, the composing syllable was painted on top of the syllable that just committed (typing `대한민국` read `대한국`), a display-only regression introduced by the #875 composition-anchor pin.

### Root cause

The pin applied its correction as a single `transform` on `.xterm-helpers`, a container holding two surfaces with opposite requirements during a composition:

- the hidden **textarea** — the caret rect Chromium reports, i.e. the candidate-window anchor, which SHOULD stay pinned where the user started typing (#874/#875/#888);
- **`.composition-view`** — the visible inline preedit, which xterm deliberately re-anchors to the live cursor on every `compositionupdate`.

Korean IME draws no candidate window: the inline preedit IS the text the user watches. When a committed syllable's echo advanced the caret, `frozen − actual` went negative and dragged the preedit back onto the committed cell. The buffer was always correct, which is why ending the composition restored the text — exactly as reported.

### Fix

Split the correction per child:

- **textarea**: keeps the composition pin (preserves the Chinese/Japanese candidate-window behavior from #874/#875/#888);
- **`.composition-view`**: never frozen; gets the same anchor math against the live cursor (so it still receives the ydisp scroll correction xterm forgets), applied only at xterm's own composition-event cadence — not per render frame — so it cannot chase mid-repaint transient cursors while an agent streams (cause-3 protection, a 3-model review-panel finding on the first cut).

### Diagnostics

The `ime-anchor` diagnostic is retagged `ime-anchor3` and now also fires on `compositionupdate`/`compositionend` when a correction changes — #942's field log was 40/41 `correction=(0,0)` precisely because the drag developed after the start-only diagnostic had fired. Records report the transforms actually applied (not recomputed values), and the log budget is split per phase (20 start + 20 update/end) so one long composition cannot silence the diagnostic.

## Testing

- New `#942` regression suite: the committed-syllable-echo scenario (verified to fail on the previous code), scrolled-viewport preedit parity, streaming-jitter protection, change-gated diagnostic phases, missing-`.composition-view` degradation.
- Upstream-contract test now also locks `.composition-view`'s place in `.xterm-helpers`.
- Full suite: 11,659 passed / 27 skipped. `tsc --noEmit` clean.
- Reviewed by a 3-model panel (Claude / Codex / GLM); all four findings addressed in the second commit. Field verification on a real Korean Windows setup is pending — the reporter offered to verify any candidate build (they are pinned to v3.41.1 meanwhile).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Korean IME composition rendering by keeping inline preedit text aligned with the live cursor.
  - Corrected candidate-window positioning without disrupting visible composition text.
  - Preserved existing Chinese and Japanese IME behavior.
  - Improved cleanup and scrolling behavior during and after text composition.

- **Diagnostics**
  - Enhanced IME diagnostics to report composition updates, completion corrections, and preedit positioning.

### DIFF
--- a/changelog.d/945.md
+++ b/changelog.d/945.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- **Korean IME no longer paints the composing syllable over the one just
+  committed.** Since v3.42.0, typing Korean showed syllables disappearing as
+  you typed (`대한민국` read `대한국` until the composition ended) — the
+  candidate-window pin introduced for Chinese/Japanese IMEs dragged the whole
+  helper container, inline preedit included, back onto committed cells. The
+  pin now holds only the hidden textarea that anchors the candidate window,
+  while the visible inline preedit stays on the live cursor, so Korean input
+  reads correctly keystroke by keystroke and the Chinese/Japanese
+  candidate-window behavior is unchanged. The IME diagnostic log (now
+  `ime-anchor2` → `ime-anchor3`) also records mid-composition and
+  composition-end corrections when they change, so a future report of this
+  family is self-diagnosing. (#945)
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1120,13 +1120,17 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
     // (#942's field log was all zeros because the drag developed after the
     // start-only diagnostic had fired). Capped: the first few are all anyone
     // needs to diagnose, and the cap keeps it from filling main-*.log for the
-    // rest of the session. Remove the whole diagnostic once #874/#942 are
-    // confirmed fixed in the field.
-    let imeAnchorLogsLeft = 20;
+    // rest of the session. Budgeted per phase so one long composition's
+    // update/end records cannot burn the start budget (or vice versa) and
+    // silence the diagnostic for the rest of the session.
+    // Remove the whole diagnostic once #874/#942 are confirmed fixed in the
+    // field.
+    const imeAnchorLogsLeft = { start: 20, mid: 20 };
     const imeAnchor = attachImeAnchor(terminal, {
       onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, selY, selX }) => {
-        if (imeAnchorLogsLeft <= 0) return;
-        imeAnchorLogsLeft -= 1;
+        const budget = phase === 'start' ? 'start' : 'mid';
+        if (imeAnchorLogsLeft[budget] <= 0) return;
+        imeAnchorLogsLeft[budget] -= 1;
         // Mirrored into the main-side log file by src/main/index.ts's
         // console-message listener, so the user can share it. The "3" in the
         // tag marks the split-surface build so a shared log is unambiguous
@@ -1140,7 +1144,7 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
           `cursor=(${cursorX},${cursorY}) sel=(${selX},${selY}) src=${src} held=${held.toFixed(0)}ms ` +
           `restAge=${restAge.toFixed(0)}ms cellHeight=${cellHeight.toFixed(2)} pin=(${dx.toFixed(1)},${dy.toFixed(1)}) ` +
           `preedit=(${preeditDx.toFixed(1)},${preeditDy.toFixed(1)})` +
-          (imeAnchorLogsLeft === 0 ? ' (last one, diagnostic capped)' : ''),
+          (imeAnchorLogsLeft[budget] === 0 ? ` (last ${budget} record, diagnostic capped)` : ''),
         );
       },
     });
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1104,34 +1104,42 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
       },
     });
 
-    // #874: keep the IME candidate window on the cursor. xterm anchors its
-    // hidden helper textarea at the ybase-relative cursor row while the
+    // #874/#942: keep the IME candidate window on the cursor. xterm anchors
+    // its hidden helper textarea at the ybase-relative cursor row while the
     // renderer paints the cursor at the ydisp-relative one, so a scrolled-up
     // viewport offsets the candidate window by exactly that many rows, and the
     // composition path re-anchors on every keystroke so the window chases the
-    // TUI's cursor while an agent streams. See terminal/imeAnchor.ts.
-    // #874 could not be reproduced locally (no CJK IME on the dev boxes), so
-    // the anchor reports what it corrected and a reporter's log tells us
-    // whether any offset survives. A CJK user starts a composition per word,
-    // so this is capped: the first few are all anyone needs to diagnose, and
-    // the cap keeps it from filling main-*.log for the rest of the session.
-    // Remove the whole diagnostic once #874 is confirmed fixed in the field.
+    // TUI's cursor while an agent streams. The pin covers the textarea only;
+    // the inline preedit box follows the live cursor (#942, Korean IME draws
+    // its composition inline with no candidate window). See
+    // terminal/imeAnchor.ts.
+    // Neither issue could be reproduced locally (no CJK IME on the dev boxes),
+    // so the anchor reports what it corrected and a reporter's log tells us
+    // whether any offset survives. Start records fire per composition;
+    // update/end records only when a correction changed mid-composition
+    // (#942's field log was all zeros because the drag developed after the
+    // start-only diagnostic had fired). Capped: the first few are all anyone
+    // needs to diagnose, and the cap keeps it from filling main-*.log for the
+    // rest of the session. Remove the whole diagnostic once #874/#942 are
+    // confirmed fixed in the field.
     let imeAnchorLogsLeft = 20;
     const imeAnchor = attachImeAnchor(terminal, {
-      onCompositionDiagnostic: ({ baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, src, held, restAge, selY, selX }) => {
+      onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, selY, selX }) => {
         if (imeAnchorLogsLeft <= 0) return;
         imeAnchorLogsLeft -= 1;
         // Mirrored into the main-side log file by src/main/index.ts's
-        // console-message listener, so the user can share it. The "2" in the
-        // tag marks the resting-cell build so a shared log is unambiguous
+        // console-message listener, so the user can share it. The "3" in the
+        // tag marks the split-surface build so a shared log is unambiguous
         // about which release produced it. src/held/restAge/sel are the
         // cause-3 discriminator: src=resting means the composition started
         // mid-repaint and the anchor used the last resting cell instead of
-        // the instantaneous cursor.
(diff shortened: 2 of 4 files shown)


openwong2kim:

**Priority: P1** — this is the fix for #942, a display regression that has Korean users pinned to v3.41.1, and it lands on the same IME-anchor surface #875/#888 introduced.

The core change replaces the single container-level `transform` on `.xterm-helpers` — which dragged the inline preedit back onto committed cells whenever a Korean syllable's echo advanced the caret — with per-child corrections: the hidden textarea keeps the composition pin that anchors the Chinese/Japanese candidate window, while `.composition-view` gets the same anchor math against the live cursor, never frozen, applied at xterm's own composition-event cadence rather than per render frame so an agent streaming mid-composition cannot drag the preedit to transient cursors. The scroll-offset (ydisp) correction is preserved for the preedit, the missing-`.composition-view` case degrades to stock behavior instead of crashing, and the upstream-contract test now locks the preedit box's place under the helpers container so an xterm rename fails loudly rather than silently losing the correction.

The diagnostic rework is what makes this reviewable without a local Korean repro: retagging to `ime-anchor3`, firing on `compositionupdate`/`compositionend` only when the applied correction changes, reporting the transforms actually on the elements instead of a recomputed value, and splitting the log budget per phase all directly answer why #942's field log came back 40/41 zeros. Test coverage matches the failure modes the report called out — the committed-syllable-echo regression case, scrolled-viewport preedit parity, streaming-jitter protection, and the deferred-sync straddle guard on `compositionend`.

One thing before this ships: field verification on a real Korean Windows setup is still pending and the reporter offered to test a candidate build, so it is worth getting an installable build to them before cutting a release, since this whole fix family has never reproduced on the dev boxes.

openwong2kim:
Field-verified on a real Korean Windows 11 setup (Microsoft IME, dev build of this branch, isolated data dir).

**Live typing test** — typed 대한민국 with the actual IME:
- At the #942 moment (대 commits via compositionend, 하 restarts composing immediately), the preedit rendered one cell to the right of the committed syllable — screen read 대하, not 하 overlapping 대. Same held through 민 → 국 (screen read 대한민국 at every step).
- DOM-level: textarea kept its pin (`translate(-14.45px)` cancelling the committed echo's advance, caret rect stationary), `.composition-view` carried only sub-pixel correction and tracked the live cursor. `.xterm-helpers` had no transform at any point.
- After commit: composition view hidden, both transforms cleared, echoed line and buffer both correct (대한민국).

**Synthetic cadence test** — composition held open while a delayed echo advanced the cursor: preedit re-anchored to the live cursor on the next compositionupdate only, textarea stayed pinned throughout, and `ime-anchor3` fired on start/update/end with distinct pin/preedit values (the #942 field-log blind spot is covered).

LGTM — merging.

Issue #940, opened 2026-08-19
Perf gate: the confirmation re-run shares the runner it is confirming, and frameBudget is quantized so 52% of main already sits at 2x baseline


snowyukitty:
The #632 confirmation re-run has now fired on `main` three times — the first
real activations since it merged. It behaved exactly as designed each time, and
in doing so it showed the one thing its design cannot do.

## What happened

| date | commit | first sample | in-job re-run | outcome |
|---|---|---|---|---|
| 08-15 | `4eeba03` (#911), [run 31902615864](https://github.com/openwong2kim/wmux/actions/runs/31902615864) | 78.1 ms | 31.2 ms | *"failed once and passed … recorded as CI noise"* → green, no human |
| 08-17 | `24fbdb9` (#909), [run 31980627825](https://github.com/openwong2kim/wmux/actions/runs/31980627825) | 46.8 ms | 46.8 ms | reproduced → red |
| 08-18 | `72bbcc4` (v3.44.0), [run 32122353502](https://github.com/openwong2kim/wmux/actions/runs/32122353502) attempt 1 | 46.9 ms | 46.8 ms | reproduced → red |

All three are `frameBudget.*.frameDeltaMs.p95` against a 15.7 ms baseline.

On the third one, attempt 2 — the same commit, a **fresh runner** — measured
15.7 ms and passed, and that is the run that carries the release tag. So a red
that reproduced on an immediate re-run was not reproducible on another machine
20 minutes later.

The trend file behaved correctly throughout: it holds the ORIGINAL first sample
each time (78.1 / 46.8 / 46.9), never the calmer retry.

## Finding 1 — the confirmation re-run shares the runner, so it cannot refute a
sustained one

`#570 → #632` was built to separate a real regression from CI noise by
re-running the scenario family immediately. That works for a *transient* blip:
`4eeba03` is exactly the false red the gate used to throw, and it was absorbed
with no human involved. It cannot work for a runner that is degraded for its
whole lifetime — the re-run inherits the same machine, reproduces, and the gate
reports a regression that a different machine cannot reproduce at all.

Today the thing that actually distinguishes the two is a human clicking
"re-run all jobs", which is what happened before v3.44.0 was tagged.

## Finding 2 — `frameBudget.*.frameDeltaMs.p95` is quantized, so the ratio rule
has only three reachable verdicts

Across the last 25 `main` records in `bench-history`, `frameBudgetP95Ms_N16`
[...]

PR #947, opened 2026-08-19
fix(perf): gate the frame-budget family on frames, and stop calling the confirmation re-run independent

snowyukitty:
Takes the two items from #940 that your triage called safe to move on — the
free wording fix, and the frame-budget verdict shape. Leaves the third (a
fresh-runner confirmation job) alone; that one changes CI topology and wants
your call first.

## 1. The confirmation summary called a same-runner re-run "independent"

`perf-confirm.mjs` printed *"The same metric failed on a second, independent
measurement of the same code."* The re-run lands on the same runner by design,
and the codebase already says so in two places — the module header
(*"measured once more on this runner"*) and `bench/README.md` (*"it cannot
cover a host that is degraded for the whole job"*). The CI-facing string was the
only place that contradicted them, and it is the one a person reads when
deciding whether a red is a regression.

Both the summary and the `::error::` line now say **same runner**, and state
what the re-run does and does not separate. Wording only — no verdict, no gate,
no baseline change.

## 2. `frameBudget.*.frameDeltaMs.p95` is gated in frames, not on a ratio

I pulled `bench-history` rather than work from the table in the issue. Across
**all 216 records**, these p95s land in clusters one frame interval apart:

| cluster | frames | note |
|---|---|---|
| 15.7 – 15.8 | 1 | 537 samples |
| 31.1 – 31.4 | 2 | 89 |
| 46.8 – 47.0 | 3 | 14 |
| 62.4 – 62.5 | 4 | 6 |
| 78.1 | 5 | 1 |

Against the blessed 1-frame baseline (15.7 for all three N), `ratio: 2.0` puts
the threshold at **exactly 31.4** — the top of the two-frame cluster. Every
two-frame record passed only because the comparison is strictly-greater. And
the threshold drifted multiplicatively: a baseline blessed from a two-frame run
would have moved the ceiling to four frames.

Replaced with `frameMargin: 1` — FAIL when `current > baseline + one frame
interval`. The `absMargin: 8` was already additive, just half a frame and
arbitrary. Drift is now linear: a two-frame baseline allows three frames.

### The constant is 15.7, and finding out why is most of the work here

The obvious value is 15.625 ms, the Windows timer tick these numbers physically
come from. **It is wrong**, and the trend says so: the clusters have width, and
15.625 puts the threshold at 31.325 — *inside* the two-frame cluster — which
flips the two real records that measured 31.4 from PASS to FAIL. That is the
same between-quanta accident this PR exists to remove, one quantum lower.

15.7 is the top of the one-frame cluster and is what `baseline-ci.json` holds,
so the constant is read off the measurement rather than off a refresh rate. I
only caught this by replaying the change over the whole trend; the 25-record
sample in the issue does not contain a 31.4.

### Verdicts, replayed rather than asserted

Old rule vs new, over **every record in the trend — 648 samples (216 x 3 N)**:
**zero disagreements.** The dropped-frame step the original calibration comment
was written around (33.3 ms) still trips the gate, as does the 37.1 ms
`perfConfirm`'s fixtures use.

Tests lock the parts that matter rather than restating the implementation:
- the five real clusters keep their verdicts;
- both cluster edges — 31.4 passes, 46.8 fails — with a comment saying why
  15.625 would break it;
- 33.3 and 37.1 still fail, so this is a reshaping and not a loosening;
- the one deliberate behaviour change (linear drift) evaluates the **old** rule
  inline and shows it let 62.5 ms through against a two-frame baseline, then
  asserts the new one does not.

`ratio` / `absMargin` stay on those three entries for the delta columns; they no
longer decide the verdict. `bench/README.md`'s gate table is updated in the
third commit — it claims to mirror `GATES`, so a stale row there is a wrong doc
about the one gate that just changed.

## What this does not do

- No fresh-runner confirmation job (#940 item 1). That is the real fix for the
  runner-health case and it adds a dependent job; your call on whether you want
  it and in what shape.
- No baseline re-blessing. Descriptive-not-aspirational is untouched, and this
  change deliberately does not let a blessed baseline move the ceiling
  multiplicatively any more.

## Verification

`npx vitest run scripts/__tests__/` — 308 passed, 17 files. ESLint clean on the
changed files. The trend replay above is reproducible from
`origin/bench-history:history.ndjson`.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved performance-gate evaluation using a consistent one-frame tolerance.
  - Updated frame-budget failure messages to clearly report exceeded allowances.
  - Confirmation results now clarify that reruns occur on the same runner and may include runner-wide degradation.

- **Documentation**
  - Clarified frame quantization, threshold behavior, validation details, and confirmation-run limitations.

- **Tests**
  - Added coverage for frame margins, threshold boundaries, scaling behavior, and confirmation messaging.

### DIFF
--- a/bench/README.md
+++ b/bench/README.md
@@ -181,7 +181,7 @@ nothing for six weeks (#602).
 
 ## Gate semantics
 
-A metric **FAILS only when both** of these hold:
+Most metrics **FAIL only when both** of these hold:
 
 - `current > baseline * ratio`, **and**
 - `current > baseline + absMargin`.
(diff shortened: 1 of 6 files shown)


openwong2kim:

This PR implements two of the three items from #940: it rewords the perf confirmation output to say the re-run happens on the same runner rather than calling it an "independent measurement", and it replaces the ratio-based gate on the `frameBudget.*.frameDeltaMs.p95` family with a frame-margin rule (`frameMargin: 1`, FAIL when `current > baseline + 15.7 ms`). It touches `scripts/perf-compare.mjs`, `scripts/perf-confirm.mjs`, their tests, the `bench/README.md` gate table, and adds a changelog fragment — CI-tooling only, no application code paths affected.

The reasoning is sound and unusually well-evidenced. The metric is quantized to whole frame intervals, so a `2.0x` ratio against a 1-frame baseline put the threshold at exactly 31.4 ms — the top of the two-frame cluster — meaning every two-frame record passed only because the comparison is strictly-greater. The choice of 15.7 over 15.625 as the interval constant is correctly justified: 15.625 would place the threshold at 31.325, inside the two-frame cluster, flipping two real records; reading the constant off `baseline-ci.json` instead of the timer tick is the right call. The replay over all 648 samples with zero verdict changes, plus tests locking both cluster edges (31.4 passes, 46.8 fails) and confirming 33.3/37.1 ms still trip, gives good confidence this is a reshaping rather than a loosening. The one deliberate behavior change — linear rather than multiplicative baseline drift — is isolated in its own test with the old rule evaluated inline for contrast.

The item deliberately left out (a fresh-runner confirmation job) is the right omission, as it changes CI topology and warrants a maintainer decision first. Assigning **P2** — the gate shape is an accident waiting to flip verdicts the moment baselines shift, but nothing is currently red because of it.

snowyukitty:
Read the bot review and checked both findings against the code rather than
applying them. Both are real, both are mine, and both are the same mistake:
prose that lagged behind a late correction.

- **Test comments still said 15.625.** I started with the Windows timer tick,
  found on the trend replay that it flips two real records at 31.4, and moved
  the constant to 15.7 — but two comments kept the old number, including one
  that put the two-frame-baseline threshold at 46.925 where the gate computes
  47.0. The assertions around them were already correct (46.8 passes against a
  31.3 baseline, 62.5 fails), so this was a wrong explanation of right code,
  which is the worse kind of stale comment. The header now also points at the
  cluster-width case for why the interval is not the tick.
- **The changelog fragment broke the metric name across a line** inside the
  inline-code span, rendering `frameBudget.*` and `.frameDeltaMs.p95` with a
  space between them.

Fixed in `777e3d31` — comments and prose only, no gate, test, or verdict
change. `npx vitest run scripts/__tests__/` stays at 308 passed.

On the docstring-coverage warning: the one symbol this PR adds to
`perf-compare.mjs` is `FRAME_INTERVAL_MS`, which carries the longest doc
comment in the file. The 50% reading is the two changed files averaged, not a
gap this diff introduced, so I have left it.

openwong2kim:
Verified the load-bearing claim independently before merging: replayed the full `bench-history` trend (now 218 records / 654 samples — two records landed after this PR was written) against the blessed 1-frame baseline. Old rule vs new: **zero disagreements**, and the cluster edges are exactly as described — the two-frame cluster tops out at 31.4 (2 real records, which a 15.625 constant would flip) and the three-frame cluster starts at 46.8.

One note from the newer records: there is now a single off-quantum sample at 28.202 ms, so "only ever lands on whole frame intervals" has one exception in the wild. It passes under both rules and the gate code already judges off-quantum samples by wall-clock distance, so nothing to change — just flagging it so the next reader of the cluster table knows.

Thanks for an unusually well-evidenced PR — the 15.7-vs-15.625 catch is exactly the kind of thing that only falls out of replaying the whole trend. Merging.

PR #950, opened 2026-08-20
chore(mcp): keep one local external-wire predicate, and test the one that runs

snowyukitty:
## Summary

Takes the non-blocking note you left when merging #941: `src/main/mcp/wireProvenance.ts` had no production importer, and the predicate that actually runs disagreed with it. This keeps one predicate and points the test at it.

## Changes

- **Fold the `operator` exclusion into `rpcProvenance.ts`** — the file `browser.rpc.ts:33` and `PermissionEnforcer.ts:61` actually import. Comment records why it is defence-in-depth rather than a live fix: an operator dispatch already carries `firstParty: true`, which the predicate rejected already, and the dispatch-options union makes `operator` together with `externalWire` unrepresentable.
- **Delete `src/main/mcp/wireProvenance.ts`.** Its `LocalExternalWireContext` type and the undefined-tolerant type-guard shape had no importer either; production has always used the `boolean(ctx: RpcContext)` signature. Say the word if you want the type-predicate form and I will fold that in too.
- **Rename `__tests__/wireProvenance.test.ts` → `__tests__/rpcProvenance.test.ts`** and re-point the import. Both cases are unchanged. The AST scan pinning `pipe/PipeServer.ts` as the sole production writer of `externalWire: true` is untouched.

## CHANGELOG entry

no CHANGELOG — internal only. No user-visible behaviour changes: the added conjunct cannot alter a decision at either call site today.

## Stability tier impact

n/a — no wire shape, envelope, enforcement point, or RPC method changed. `rpcProvenance` is an internal helper behind the existing enforcement points.

## Test plan

- `npx tsc --noEmit -p tsconfig.json` — clean.
- `npx vitest run src/main/mcp src/main/pipe` — 51 files, 984 tests, all pass.
- `npx eslint src/main/mcp/rpcProvenance.ts src/main/mcp/__tests__/rpcProvenance.test.ts` — clean.
- Mutation check, which is the point of the change: removing `ctx.operator !== true` from `rpcProvenance.ts` now fails `rpcProvenance.test.ts` (the `operator: true` case, and only that case). Before this PR the same removal failed nothing, because the assertion was pointed at the dead copy.
- Deleting the file does not disturb the AST invariant — `wireProvenance.ts` declared `externalWire: true` only in type position (a `PropertySignature`), which `writesExternalWireTrue` does not count. Expected writers stay `['pipe/PipeServer.ts']`, verified by the test passing.
- `grep -rn wireProvenance` over the repo (excluding `node_modules`) returns nothing.

## Related issues / PRs

Follow-up to #941.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for local external-wire operations by excluding operator-dispatched contexts.
  * Retained safeguards requiring external-wire origin, local execution, and non-first-party context.
* **Refactor**
  * Consolidated provenance validation to improve consistency and fail-closed behavior.


openwong2kim:

This PR resolves the non-blocking note from merging #941: `wireProvenance.ts` had no production importer while the predicate that actually runs (`rpcProvenance.ts`, imported by `browser.rpc.ts` and `PermissionEnforcer.ts`) disagreed with it. The diff does exactly what it says — folds the `operator` exclusion into the live predicate (defence-in-depth, since `firstParty: true` already excludes operator dispatches today), deletes the dead file along with its unused `LocalExternalWireContext` type, and renames the test so its assertions target the code that ships. The mutation check in the test plan is the convincing part: removing the `operator` conjunct previously failed nothing because the test pointed at the dead copy, and now it fails exactly the one case it should. P3 — internal-only cleanup with no user-visible behaviour change, but it removes a real trap for the next person touching MCP provenance, so worth merging before further work lands there.

PR #953, opened 2026-08-20
Anchor the IME candidate window to the quiet caret while output streams

### LINKED ISSUE #951: [IME] Chinese IME candidate window still anchors to the streaming cursor, not the typing position (#945 did not fix this)
### Problem / motivation

Chinese IME (WeChat 微信输入法) candidate window still does not follow the cursor in the desktop terminal, even on current `main` with the #945 composition-anchor split. #874 was closed as fixed by #875/#888, but the field symptom is unchanged for CJK IMEs; #945 addressed the Korean IME (inline preedit) case but not the Chinese candidate-window case.

**Key observation (latest build, ime-anchor3 diagnostics):** the candidate window is correct most of the time when the pane is idle, but when Claude Code is streaming output the candidate window anchors to the WRONG position — usually the row where Claude's own cursor currently sits mid-repaint (its start or end), occasionally the bottom-right corner of the screen. The `cursor=` field in the ime-anchor3 log is the smoking gun.

### Reproduction

Windows 11, WeChat 微信输入法, desktop wmux from `main` (contains #945). Open a pane running Claude Code (streaming output), then type Chinese during/around its output.

### Diagnostics

Pane is 128×45. ime-anchor3 log from `%APPDATA%\wmux-dev\logs\main-*.log` (dev build):

```
[wmux:ime-anchor3] composition-start ybase=661 ydisp=661 cursor=(53,43) sel=(53,43) src=instant pin=(0.0,0.0) preedit=(-592.0,80.0)
[wmux:ime-anchor3] composition-start ybase=661 ydisp=661 cursor=(127,43) sel=(127,43) src=instant pin=(0.0,0.0) preedit=(592.0,0.0)
[wmux:ime-anchor3] composition-start ybase=661 ydisp=661 cursor=(127,38) sel=(127,38) src=instant pin=(0.0,0.0) preedit=(0.0,0.0)
[wmux:ime-anchor3] composition-start ybase=665 ydisp=665 cursor=(4,2) sel=(4,2) src=instant pin=(0.0,0.0) preedit=(0.0,0.0)
```

`cursor=(127,43)` is the bottom-right cell of a 128×45 pane; `(53,43)` and `(127,38)` hug the bottom/right edge. These are positions where Claude Code parked its cursor while repainting, not where the user is typing. The composition started there, so the candidate window anchored there.

### Root cause analysis

The anchor math (#875 scroll offset, #888 resting-cell, #945 preedit split) is all working — `pin` corrections are small and the preedit tracks. The unresolved part is **which cell is sampled at `compositionstart`**: `selectFreezeCell` falls back to `src=instant` (the live buffer cursor), and during a streaming TUI repaint that cursor is wherever the TUI last parked it mid-frame — a screen corner, the output row, etc. The user's caret (where they clicked to type) is NOT the buffer cursor while output is streaming. So the candidate window anchors to a cell that has nothing to do with the user's input position.

This is the original #874 symptom, still open in practice: the candidate window "follows the flickering cursor" while streaming. The #875/#888/#945 fixes refined the anchoring math but never fixed the sampled cell itself.

### Expected behavior

When a composition starts, the candidate window should anchor to where the user is actually typing (the caret / selection position), not to the buffer cursor that a streaming TUI happens to have parked somewhere.

### Environment

- wmux `main` (contains #945), dev build via `npm start`
- Windows 11, WeChat 微信输入法 (also reproduces with Microsoft Pinyin per #874)
- Terminal pane 128×45 running Claude Code
- Reproduces in the desktop terminal (not web)

### Related

- #874 (original, closed as fixed but field symptom unchanged)
- #942 (Korean IME preedit regression, fixed by #945 — different code path, same anchor file)
- #945 (textarea/preedit split — candidate-window path unchanged)

openwong2kim:
Fixes #951.

## Problem

While an agent (Claude Code) streams output, a Chinese IME composition (WeChat IME, Microsoft Pinyin) anchors its candidate window to whatever cell the streaming TUI last parked its cursor on — a screen corner or an output row — instead of the user's input position. The #951 field logs show it directly: `src=instant` selections at `(127,43)`, `(53,43)`, `(127,38)` on a 128×45 pane.

## Root cause

`selectFreezeCell` trusted any buffer cursor that had held its cell for `RESTING_MS` (32 ms). That threshold was designed to reject sub-ms inter-chunk gaps inside one repaint frame, but a streaming agent's inter-burst gaps (token pacing, network) are far longer, so the dwell heuristic certified exactly the wrong cells. The `resting` fallback derives from the same buffer cursor and shares the blind spot. The missing signal is output recency: a cursor position observed while output is still flowing is repaint state, not a caret.

## Fix

- Track output recency via xterm's public `onWriteParsed`.
- Record a **quiet caret**: the cell the cursor held through the most recent ≥ `OUTPUT_QUIET_MS` (500 ms) output-free span — the position the TUI deliberately parked it on. Stored as a **screen row**, so output scrolling the buffer does not invalidate it (a TUI addresses its input line by screen position).
- At `compositionstart`, when output is recent **and sustained** (`STREAM_SUSTAIN_MS` = 700 ms since the burst epoch began), anchor to the quiet caret (`src=caret`) instead of trusting dwell. The sustain gate keeps ordinary typing intact: a committed syllable's echo is recent output too, but it is an isolated burst — without the gate, every commit would anchor the next composition on a one-word-stale snapshot (3-model review panel finding).
- Tracker resets (resize reflow, alt-screen switch) re-seed the output clock so a snapshot can re-form after a reflow (panel finding).
- Idle panes keep the exact pre-#951 selection, which the field reports as correct; with no snapshot the selection degrades to the previous behavior, never worse.

## Diagnostics

`ime-anchor3` → `ime-anchor4`: adds `gap=` (output silence at compositionstart), `caretAge=` (snapshot age), and the `caret` source; update/end records now emit the screen row fixed at selection time instead of re-deriving it against an ybase the stream has since advanced. A follow-up field log from the reporter can confirm the fix or identify the documented residuals (mid-stream pause re-snapshot, resize/clear stranding) without another instrumentation round.

## Tests

- Pure-tracker and DOM-wiring regression tests reproducing the field scenario (corner cell held ≥ 32 ms mid-stream loses to the caret; `src=caret` did not exist before, so both fail without the fix).
- Fast-typist regression lock: an isolated echo burst keeps the freshly advanced cursor.
- Screen-row rebase across scrolled output, reset re-seeding, snapshot formation in both intra-chunk event orders, sub-quiet gap rejection.
- Full suite: 11713 passed / typecheck clean.

## Field validation

No CJK IME or Windows box is available locally (same constraint as #875/#888/#945), so the anchor selection is validated by unit tests and the diagnostic is the field instrument. Requesting a log from @cnxiekun on the next build, as with previous rounds.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Chinese IME candidate-window positioning during streamed terminal output.
  * Cursor anchoring now responds to recent output activity while preserving quiet-pane typing behavior.
  * Improved handling at terminal line ends, during scrolling, and when output continues without a recent cursor snapshot.
  * Inline preedit text now follows the live cursor except when anchored to a recent caret position.
* **Diagnostics**
  * Added output-gap, snapshot-age, and edge-condition details to IME diagnostics.
* **Tests**
  * Expanded coverage for streaming, scrolling, edge positions, and fallback behavior.

### DIFF
--- a/changelog.d/953.md
+++ b/changelog.d/953.md
@@ -1,0 +1,14 @@
+### Fixed
+
+- **Chinese IME candidate window no longer jumps to wherever a streaming
+  agent parked its cursor.** Typing Chinese (WeChat IME, Microsoft Pinyin)
+  while Claude Code streamed output anchored the candidate window to the
+  TUI's repaint cursor — a screen corner or an output row — because a cursor
+  that sat still for 32 ms was trusted as the caret, and a streaming TUI
+  parks its cursor between output bursts far longer than that. The anchor
+  now tracks output recency: while output is streaming, the composition
+  anchors to the cell the cursor held through the last output-quiet period —
+  the input line the TUI deliberately parked it on — instead of trusting
+  stillness alone. Ordinary typing into a quiet pane is unaffected, and the
+  IME diagnostic log (`ime-anchor3` → `ime-anchor4`) now records the output
+  gap and snapshot age so a follow-up report is self-diagnosing. (#951)
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1127,7 +1127,7 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
     // field.
     const imeAnchorLogsLeft = { start: 20, mid: 20 };
     const imeAnchor = attachImeAnchor(terminal, {
-      onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, outputGap, caretAge, selY, selX }) => {
+      onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, outputGap, caretAge, edge, selY, selX }) => {
         const budget = phase === 'start' ? 'start' : 'mid';
         if (imeAnchorLogsLeft[budget] <= 0) return;
         imeAnchorLogsLeft[budget] -= 1;
@@ -1143,7 +1143,7 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
         // live composition-view correction.
         console.info(
           `[wmux:ime-anchor4] pty=${ptyIdRef.current} composition-${phase} ybase=${baseY} ydisp=${viewportY} ` +
-          `cursor=(${cursorX},${cursorY}) sel=(${selX},${selY}) src=${src} held=${held.toFixed(0)}ms ` +
+          `cursor=(${cursorX},${cursorY}) sel=(${selX},${selY}) src=${src}${edge ? ' edge=1' : ''} held=${held.toFixed(0)}ms ` +
           `restAge=${restAge.toFixed(0)}ms gap=${outputGap.toFixed(0)}ms caretAge=${caretAge.toFixed(0)}ms ` +
           `cellHeight=${cellHeight.toFixed(2)} ` +
           `pin=(${dx.toFixed(1)},${dy.toFixed(1)}) preedit=(${preeditDx.toFixed(1)},${preeditDy.toFixed(1)})` +
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1127,23 +1127,26 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
     // field.
     const imeAnchorLogsLeft = { start: 20, mid: 20 };
     const imeAnchor = attachImeAnchor(terminal, {
-      onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, selY, selX }) => {
+      onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, outputGap, caretAge, selY, selX }) => {
         const budget = phase === 'start' ? 'start' : 'mid';
         if (imeAnchorLogsLeft[budget] <= 0) return;
         imeAnchorLogsLeft[budget] -= 1;
         // Mirrored into the main-side log file by src/main/index.ts's
-        // console-message listener, so the user can share it. The "3" in the
-        // tag marks the split-surface build so a shared log is unambiguous
-        // about which release produced it. src/held/restAge/sel are the
-        // cause-3 discriminator: src=resting means the composition started
-        // mid-repaint and the anchor used the last resting cell instead of
-        // the instantaneous cursor. pin= is the textarea correction,
-        // preedit= the live composition-view correction.
+        // console-message listener, so the user can share it. The "4" in the
+        // tag marks the quiet-caret build (#951) so a shared log is
+        // unambiguous about which release produced it. src/held/restAge/sel
+        // are the cause-3 discriminator: src=resting means the composition
+        // started mid-repaint and the anchor used the last resting cell;
+        // src=caret with gap= is the #951 discriminator: output was still
+        // flowing, so the anchor used the quiet-caret snapshot instead of
+        // any buffer cursor. pin= is the textarea correction, preedit= the
+        // live composition-view correction.
         console.info(
-          `[wmux:ime-anchor3] pty=${ptyIdRef.current} composition-${phase} ybase=${baseY} ydisp=${viewportY} ` +
+          `[wmux:ime-anchor4] pty=${ptyIdRef.current} composition-${phase} ybase=${baseY} ydisp=${viewportY} ` +
           `cursor=(${cursorX},${cursorY}) sel=(${selX},${selY}) src=${src} held=${held.toFixed(0)}ms ` +
-          `restAge=${restAge.toFixed(0)}ms cellHeight=${cellHeight.toFixed(2)} pin=(${dx.toFixed(1)},${dy.toFixed(1)}) ` +
-          `preedit=(${preeditDx.toFixed(1)},${preeditDy.toFixed(1)})` +
+          `restAge=${restAge.toFixed(0)}ms gap=${outputGap.toFixed(0)}ms caretAge=${caretAge.toFixed(0)}ms ` +
+          `cellHeight=${cellHeight.toFixed(2)} ` +
+          `pin=(${dx.toFixed(1)},${dy.toFixed(1)}) preedit=(${preeditDx.toFixed(1)},${preeditDy.toFixed(1)})` +
           (imeAnchorLogsLeft[budget] === 0 ? ` (last ${budget} record, diagnostic capped)` : ''),
         );
       },
--- a/src/renderer/terminal/__tests__/imeAnchor.test.ts
+++ b/src/renderer/terminal/__tests__/imeAnchor.test.ts
@@ -929,6 +929,17 @@ describe('#951 quiet-caret wiring', () => {
     });
     expect(translateOf(dom.textarea)?.dy).toBeCloseTo((40 - 43) * 17.6, 6);
     expect(translateOf(dom.textarea)?.dx).toBeCloseTo((5 - 127) * 10, 6);
+    // xterm re-anchors the inline preedit to the live cursor on every
+    // compositionupdate. A caret-sourced composition must pull it back onto
+    // the same frozen point as the textarea — the first quiet-caret build
+    // left it on the live cursor, and the field report showed the pinyin
+    // riding the agent's output rows while its candidate list sat at the
+    // input line (the two IME surfaces torn apart).
+    dom.compView.style.top = `${43 * 17.6}px`;
+    dom.compView.style.left = `${127 * 10}px`;
+    dom.textarea.dispatchEvent(new Event('compositionupdate'));
+    expect(translateOf(dom.compView)?.dy).toBeCloseTo((40 - 43) * 17.6, 6);
+    expect(translateOf(dom.compView)?.dx).toBeCloseTo((5 - 127) * 10, 6);
     handle.dispose();
   });
 });
--- a/src/renderer/terminal/__tests__/imeAnchor.test.ts
+++ b/src/renderer/terminal/__tests__/imeAnchor.test.ts
@@ -21,6 +21,8 @@ import {
   computeImeAnchorCorrection,
   createRestingTracker,
   isUsableGeometry,
+  ANCHOR_CONTINUITY_MS,
+  noteCompositionAnchor,
   noteCursorMove,
   noteOutputParsed,
   OUTPUT_QUIET_MS,
@@ -869,36 +871,42 @@ describe('#951 quiet-caret tracker (pure)', () => {
     expect(sel).toMatchObject({ absRow: 640, col: 5, src: 'resting' });
   });
 
-  it('#953 case 1-2: a last-column park is rejected even when output is quiet', () => {
+  it('#953 case 1-2: a last-column park yields to the previous composition anchor', () => {
     // Claude Code idling with an empty input box parks its real cursor at
-    // the line end — (236,47) on a 237-col pane — and a 1.6s output gap then
+    // the line end — (236,47) on a 237-col pane — and a long output gap then
     // certified it as an at-rest caret, pushing the candidate window and the
     // pinyin off the right edge. A CJK composition can never sit on the last
-    // column, so the selection must fall back to the caret snapshot.
+    // column, so the selection reuses where the user last composed.
     const t = createRestingTracker(640, 5, 1000, 40);
-    noteOutputParsed(t, 1600, 237);        // snapshot: the input-line cell
-    noteCursorMove(t, 647, 236, 1700, 47); // TUI parks at the line end
-    noteOutputParsed(t, 1700, 237);
-    const sel = selectFreezeCell(t, 647, 236, 3400, { top: 600, rows: 48, cols: 237 }, 600);
-    expect(sel).toMatchObject({ absRow: 640, col: 5, src: 'caret', edge: true });
+    noteCompositionAnchor(t, 40, 12, 2000);  // previous composition anchored here
+    noteCursorMove(t, 647, 236, 2500, 47);   // TUI parks at the line end
+    const sel = selectFreezeCell(t, 647, 236, 4000, { top: 600, rows: 48, cols: 237 }, 600);
+    expect(sel).toMatchObject({ absRow: 640, col: 12, src: 'prev', edge: true });
+  });
+
+  it('#953: with no previous anchor a line-end park degrades to the instant cell', () => {
+    // Third field pass: every quiet cursor-derived fallback (the snapshot,
+    // the resting cell) wanders between inputs — a stable right-edge anchor
+    // read as better than an unpredictable one. Cold start keeps instant.
+    const t = createRestingTracker(640, 5, 1000, 40);
+    noteCursorMove(t, 647, 236, 1100, 47);
+    const sel = selectFreezeCell(t, 647, 236, 3000, { top: 600, rows: 48, cols: 237 }, 600);
+    expect(sel).toMatchObject({ absRow: 647, col: 236, src: 'instant', edge: false });
+  });
+
+  it('#953: an expired anchor no longer participates', () => {
+    const t = createRestingTracker(640, 5, 1000, 40);
+    noteCompositionAnchor(t, 40, 12, 2000);
+    noteCursorMove(t, 647, 236, 2500, 47);
+    const late = 2000 + ANCHOR_CONTINUITY_MS + 1;
+    const sel = selectFreezeCell(t, 647, 236, late, { top: 600, rows: 48, cols: 237 }, 600);
+    expect(sel).toMatchObject({ absRow: 647, col: 236, src: 'instant', edge: false });
   });
 
   it('#953: line-end cells are never snapshotted as the caret', () => {
     const t = createRestingTracker(647, 236, 1000, 47);
     noteOutputParsed(t, 1600, 237); // spanning cell is a line-end park
     expect(t.hasCaret).toBe(false);
-    // …and with no snapshot and no resting cell, a last-column instant
-    // degrades to the pre-#953 selection instead of anchoring nowhere.
-    const sel = selectFreezeCell(t, 647, 236, 3000, { top: 600, rows: 48, cols: 237 }, 600);
-    expect(sel).toMatchObject({ absRow: 647, col: 236, src: 'instant', edge: false });
-  });
-
-  it('#953: without a snapshot, a last-column park falls back to the resting cell', () => {
-    const t = createRestingTracker(640, 5, 1000, 40);
-    noteCursorMove(t, 647, 236, 1100, 47); // promotes the input-line cell
-    const sel = selectFreezeCell(t, 647, 236, 2000, { top: 600, rows: 48, cols: 237 }, 600);
-    // held=900ms would have certified the line end; edge rejection wins.
-    expect(sel).toMatchObject({ absRow: 640, col: 5, src: 'resting', edge: true });
   });
 
   it('#953: a mid-line at-rest cursor keeps the instant selection (regression lock)', () => {
@@ -906,6 +914,28 @@ describe('#951 quiet-caret tracker (pure)', () => {
     noteOutputParsed(t, 1600, 237);
     const sel = selectFreezeCell(t, 640, 100, 3400, { top: 600, rows: 48, cols: 237 }, 600);
     expect(sel).toMatchObject({ absRow: 640, col: 100, src: 'instant', edge: false });
+  });
+
+  it('#953: consecutive streaming compositions reuse the previous anchor, not a re-taken snapshot', () => {
+    // The snapshot can be re-captured across mid-stream pauses and wander
+    // between inputs (field: rows 36 vs 43, cols 23-79). Once a composition
+    // has anchored, the next one lands at the same spot.
+    const t = createRestingTracker(640, 5, 1000, 40);
+    noteCursorMove(t, 643, 60, 2000, 43);
+    noteOutputParsed(t, 2000, 128);  // snapshot + epoch start
+    noteOutputParsed(t, 2300, 128);
+    noteOutputParsed(t, 2700, 128);
+    const first = selectFreezeCell(t, 643, 60, 2760, { top: 600, rows: 45, cols: 128 }, 600);
+    expect(first).toMatchObject({ src: 'caret', absRow: 640, col: 5 });
+    noteCompositionAnchor(t, first.absRow - 600, first.col, 2760);
+    // A mid-stream pause re-snapshots a DIFFERENT (wrong) cell…
+    noteCursorMove(t, 636, 79, 3200, 36);
+    noteOutputParsed(t, 3900, 128); // 1.2s pause ended — snapshot re-taken at (43,60)
+    noteOutputParsed(t, 4200, 128);
+    noteOutputParsed(t, 4650, 128);
+    const second = selectFreezeCell(t, 636, 79, 4700, { top: 600, rows: 45, cols: 128 }, 600);
+    // …but the previous anchor wins, so the candidate window does not wander.
+    expect(second).toMatchObject({ src: 'prev', absRow: 640, col: 5 });
   });
 
   it('a same-cell report still refreshes the screen row (scroll under a stationary cursor)', () => {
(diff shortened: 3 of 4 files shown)


openwong2kim:

This PR fixes #951: while an agent streams output, a Chinese IME composition anchors its candidate window to whatever cell the streaming TUI last parked its cursor on (screen corners, output rows) instead of the input position. The diagnosis matches the field logs exactly — `selectFreezeCell` trusted dwell time (RESTING_MS = 32 ms), but inter-burst gaps during streaming run far longer, so dwell certified repaint state as a caret. The fix adds the missing signal, output recency via `onWriteParsed`, and anchors to a quiet-caret snapshot — the cell the cursor held through >= 500 ms of output silence, stored as a screen row so scrolled output does not invalidate it — with a sustain gate (700 ms) so an isolated syllable echo still keeps the freshly advanced cursor and fast typists do not drift.

The change is tight and reviewable: about 200 lines in `src/renderer/terminal/imeAnchor.ts`, a diagnostic bump to `ime-anchor4` (adds `gap=` / `caretAge=`) in `useTerminal.ts`, a changelog entry, and a thorough pure-tracker + DOM-wiring test block covering both intra-chunk event orders, reflow reset re-seeding, the screen-row rebase across scrolled output, and the fast-typist regression lock. Idle panes keep the pre-#951 selection exactly, and the documented residuals (mid-stream pause re-snapshot, resize/clear stranding) are self-identifying in the new log fields.

Since #951 is P1 and CJK input is broken in exactly this scenario for the reporters, marking this P1 — it should go into the next release once CI is green, with a follow-up field log from the #951 reporter to confirm, as in previous IME rounds.

cnxiekun:
@openwong2kim Correcting my earlier reading — the pinyin and the candidate window are ONE IME rendering, not two independently-anchored surfaces.

**What is now right (vs before #953):** while Claude Code streams, the candidate window no longer jumps to the top output rows. Clear improvement. Good.

**What is still wrong, more precisely:**

- **When idle and "following", both the pinyin and the candidate list sit together at the caret** — that is the correct layout (see screenshot). When the candidate list is off, **the pinyin is not even visible** — they fail together, which means the whole IME surface is anchored to the wrong cell, not the two halves anchoring differently.
- **While streaming**, the pinyin appears up on Claude's streaming output rows, and the candidate list is mostly not at the input caret either. Again the whole IME surface is displaced, not split.

The correct layout (pinyin directly above the candidate list, both at the caret):

correct layout

**My reading of the cause:** the anchor cell (what the IME surface follows) is still wrong during streaming — it is not the user caret. Whether that is the quiet-caret sampling missing some cases, or the preedit following the live cursor (#945) while the candidate window follows the quiet caret, the observable symptom is a single displaced IME surface. The diagnostic fields (`src`, `gap`, `caretAge`) should show whether `src=caret` is being selected in the failing compositions or whether it falls back to `instant`/`resting` on the streaming rows.

openwong2kim:
Thanks for the field test — your reading is exactly right, and the split is now fixed in 5a892946.

**What was happening:** the two IME surfaces were anchored to different cells whenever output flowed. The candidate window follows the hidden textarea, which #953 pins to the quiet caret; the inline pinyin is xterm's `.composition-view`, which #945 deliberately left on the live cursor (that live follow is what fixed the Korean inline case, where a committed syllable's echo must advance the composing syllable). Mid-stream, the live cursor is the agent's repaint cursor — so the pinyin rode the output rows while its own candidate list stayed at the input line.

**The fix:** when a composition's freeze cell came from the quiet caret (`src=caret`, i.e. output is streaming), the preedit now pins to the same frozen point as the textarea, so pinyin and candidates stay stacked at the input position — your reference screenshot's layout. Quiet compositions are untouched: there the selection is `instant`/`resting` and the live follow stays exactly as #945 shipped it, so the Korean case cannot regress.

Would appreciate one more `ime-anchor4` log pass on this build when you get a chance — on a `src=caret` record the `pin=` and `preedit=` corrections should now be identical while streaming.

cnxiekun:
@openwong2kim Field test on the `5a892946` build — the streaming pinyin/candidate split is fixed (`src=caret` records now show identical `pin=` / `preedit=`), good. Two remaining cases remain, both on `src=instant` and both visible in the log:

**Case 1 — idle, starting to type with an empty input box: the caret sits at the bottom-right.

**Case 2 — idle, after typing some text and then deleting part of it, the caret sometimes jumps to the right end of the input line.** After the delete, the composition can start at `cursor=(236,47)` again — the TUI parked the cursor at the line end — and the same thing happens: candidate window off to the right, pinyin pushed off-screen.

wmux-paste-1787227889177-kv0enp

**Case 3 — streaming, while typing, the pinyin sometimes lands on Claude's thinking/output row above, sometimes at the far right (off-screen).** Same shape: the pinyin follows a `src=instant` cell that is the agent's repaint position, not the caret.

Screenshot (pinyin up on the streaming row):

wmux-paste-1787228134231-mdvp9r

**My reading:** `src=instant` still certifies cells that are TUI placeholders — the line-end caret after clearing or deleting, and the streaming repaint cursor — because the only gate is the output gap, which is long in all these cases. The `instant` selection has no way to tell "the caret the user actually parked" from "a cell the TUI parked". All failure paths share this: the anchor lands on a TUI-parked cell, and the pinyin is then pushed off-screen by the same correction.

Happy to capture more logs or a screen recording if it helps. The `src=caret` / `src=instant` split in the diagnostic is exactly what makes these cases self-identifying.

openwong2kim:
Great log capture again — all three cases share one signature, and 9d6486e2 targets it directly.

**The shared signature:** every remaining mis-anchor is a `src=instant` selection at the **last column** — `(236,47)` on your 237-col pane, and `(127,43)` in the original #951 logs. A TUI with nothing to point at (empty input box, mid-edit redraw, streaming repaint) parks its real cursor at the end of the line it just painted, and a long output gap then certifies that placeholder as an at-rest caret.

**The fix:** a CJK composition can never sit on the last column — the glyph alone is two cells wide — so a last-column cursor is now rejected outright, quiet or not. The selection falls back to the quiet-caret snapshot first, then the resting cell, and degrades to the old behavior only when neither exists. Line-end cells are also excluded from snapshot capture, so the placeholder cannot sneak back in through the caret path. Rejected selections log `edge=1`.

Per case:
- **Case 1 (idle, empty input box)** — the `(236,47)` park now rejects to the caret snapshot; your log line already showed `caretAge=2925ms`, i.e. a snapshot was available and pointing at the input line.
- **Case 2 (after deleting text)** — same path, same rejection.
- **Case 3 (streaming, pinyin on the thinking row / far right)** — the far-right subcase is the same last-column rejection. The thinking-row subcase is the documented mid-stream-pause residual (a >500 ms pause with the cursor parked mid-line re-snapshots a wrong caret); it should now be rarer since line-end parks no longer poison the snapshot, but a mid-line park across a pause can still slip through — if you catch one, the record will show `src=caret` with a small `caretAge` and a mid-line `sel=`, which tells us whether a further gate is worth it.

Same ask as before: one more log pass on this build — `edge=1` records landing on the input line confirm the guard; any remaining miss will be self-identifying from the src/gap/caretAge/edge combination.

cnxiekun:
@openwong2kim Field test on the `9d6486e2` build (128×45 pane). The `edge=1` guard works — last-column rejects land as expected — but three regressions are now worse, and the caret snapshot the guard falls back to is unstable.

**Case 1 — starting to type with an empty input box (idle):**
The caret now offsets to the **bottom-right of the whole wmux screen** — worse than before, where it was only the bottom-right of the claude-hud content area.
wmux-paste-1787230497399-z2np03

**Case 2 — after typing then deleting part of the text:**
The caret no longer jumps to the far right of the input line. But now the caret's **starting position varies between inputs** — each time you begin typing it lands somewhere unpredictable, not where it landed last time. During a single input it stays put; the position is just not consistent across inputs. Previous builds: idle typing was almost perfect, with only the occasional post-delete jump to the right end; this build feels worse overall.
wmux-paste-1787230720774-yxlprv

**Case 3 — while Claude Code streams:**
Similar to Case 2 — the caret's starting position is not fixed between inputs.

**Log analysis (guard works, but the snapshot it falls back to is unstable):**
The log shows `edge=1` correctly rejecting last-column (127) selections. But the caret snapshot it falls back to is itself unstable: consecutive compositions snapshot different cells (row 36 vs 43; cols 23–79), so the candidate window wanders. The snapshot is captured over the quiet span, but the TUI keeps moving the cursor, so "the cell that stayed still" changes each time.

Happy to capture a screen recording if it helps. The `edge=1` + `src`/`caretAge` fields make these cases self-identifying in the log.

openwong2kim:
Thanks for the honest ranking — you were right that the snapshot fallback was a net loss for idle typing, and 70c8a5fb replaces it.

**What the third log showed:** the `edge=1` guard fires correctly, but the quiet-caret snapshot it fell back to is re-captured every time an output-free span ends — and an idle TUI keeps twitching its cursor, so consecutive compositions snapshotted different cells (your rows 36 vs 43, cols 23–79) and the candidate window wandered. Cursor-derived fallbacks cannot be stabilized; the TUI's cursor simply isn't the caret.

**The replacement — composition continuity:** the one cell we know is right is where the *previous composition* anchored. That anchor (fresh within 60 s, refreshed on every trustworthy composition) is now the preferred fallback everywhere the cursor can't be trusted:

- **Idle, empty box / after deletes (cases 1–2):** a last-column park reuses the previous anchor (`src=prev edge=1`) — consecutive inputs land at one consistent spot. With no previous anchor (cold start) it degrades to the instant cell, i.e. exactly the pre-guard behavior you ranked "almost perfect", never the wandering snapshot.
- **Streaming (case 3):** the first composition of a session still seeds from the pre-stream snapshot (that part you confirmed fixed), and every later one reuses the previous anchor — so mid-stream pause re-snapshots can no longer move the candidate window between inputs.
- Anchors are only ever recorded from trustworthy selections (quiet non-edge instant, snapshot, reuse) — never line-end or mid-burst cells — and the pinyin pins together with the candidate window for every non-live-cursor anchor.

What the log should show on this build: `src=prev` records with a constant `sel=` across consecutive inputs, both idle-after-delete and mid-stream. The known cold-start limit (very first composition into an empty box before any successful composition) intentionally keeps the old right-edge behavior — if that case matters enough to chase further, it needs a user-side signal rather than another cursor heuristic, and I'd rather validate this build first.

cnxiekun:
@openwong2kim Field test on `70c8a5fb` (composition continuity). The log confirms the new mechanism works as designed: `src=prev` records appear, and `sel=` stays constant across consecutive inputs (e.g. `sel=(58,33)` repeated) — the wandering between inputs is gone.

However, two issues remain:
- **Idle after delete:** the caret occasionally lands mid-line (not the input position) — the reused previous anchor inherits the previous anchor's error when that anchor was not the current caret.
- **Streaming:** the candidate window still appears on the output rows above, as before.

Honest ranking across the three builds: `9d6486e2` (snapshot fallback) and `70c8a5fb` (prev-anchor reuse) each traded one problem for another. If a further fix is complex, I'd suggest reverting to **`5a892946`** — in practice the input experience there was the best of the three.

openwong2kim:
You called it right, and I've taken your suggestion: **8bb88adc restores the 5a892946 selection behavior.** Both fallback generations are gone — no snapshot rerouting on line-end parks, no previous-anchor reuse. Streaming still anchors to the pre-stream quiet-caret snapshot (the part your second pass confirmed fixed), everything else is back to the dwell selection you ranked best.

Two harmless pieces from the later rounds stay, both invisible to behavior:
- line-end cells are still excluded from snapshot capture (can only improve the streaming snapshot, never moves an anchor), and
- a last-column cursor still logs `edge=1` — purely diagnostic now, so future logs keep saying "this was a line-end park" without acting on it.

**On the two remaining defects** (idle line-end park with an empty/edited input box; streaming candidate window on output rows when the snapshot is stale): after four rounds I'm convinced no cursor-derived heuristic closes them — every signal we can derive from the cursor (dwell, output-recency, snapshots, continuity) has now been field-falsified in some flow. They need a different signal class, and I'm treating that as follow-up work rather than more iterations here:

1. **The real fix is upstream.** Claude Code draws its own caret and does not park the terminal's real cursor on it, which breaks IME anchoring in every terminal (Zed, Ghostty, Hyper report the same). Ink has shipped the API for this (`useCursor`, 6.7.0, with a dedicated IME example) but Claude Code hasn't adopted it — anthropics/claude-code#25186 is the canonical issue, and your logs here are strong evidence to add to it.
2. **Terminal-side follow-up (separate PR):** content-aware anchoring — detect the known agent's input-line marker in the screen buffer and anchor there, gated per detected agent. Other agent-focused terminals run this approach in production; it fixes the empty-input-box case that no cursor signal can, but it needs its own e2e coverage before it ships.

If you can give 8bb88adc one final pass to confirm it matches your 5a892946 experience, I'll take this PR to review as-is. Thanks for four rounds of exceptional field data — the honest ranking especially is what stopped this from over-fitting.

cnxiekun:
@openwong2kim Thanks for the revert — I understand `8bb88adc` is behaviorally meant to match `5a892946`, and the two kept pieces (line-end cells not snapshotted, `edge=1` diagnostic) should not affect normal input. But in practice the experience is still slightly worse than `5a892946` — the caret is a bit more often off-position. I can't pin down the exact detail, but it's a real difference in feel. Could you restore `8bb88adc` to be byte-for-byte identical to `5a892946`?

openwong2kim:
You were right, and the difference you felt was real — my "invisible to behavior" claim covered one piece it shouldn't have.

Of the two things I kept from the later rounds, only `edge=1` was inert. The other one — *line-end cells are never snapshotted* — does change behavior: when the pane's pre-stream quiet caret happened to sit on the last column, `8bb88adc` dropped that snapshot entirely, so a composition during streaming fell back to the dwell cursor instead of anchoring to `src=caret`. On paper that only ever removes an anchor I'd already judged wrong; in the field you measured it worse, and the field is what decides.

**`5ca90958` removes it.** The tracker now takes the quiet-caret cell exactly as `5a892946` did, with no last-column condition anywhere in the selection or the snapshot path.

What remains different from `5a892946` is a single computed boolean that is written into the log line and read by nothing else:

- `selectFreezeCell` computes `edge = instCol >= cols - 1` and carries it on the returned selection.
- `useTerminal` prints it as ` edge=1` in the `ime-anchor4` record.

No branch reads it. Every code path that chooses a cell — `caret`, `instant`, `scrolled_out`, `resting` — is byte-identical to `5a892946`, as is `noteOutputParsed`. I kept it because it costs nothing at runtime and it's what tells us "this was a line-end park" in your next log without another instrumentation round. If you'd rather have the literal byte-for-byte tree, say so and I'll drop those two lines too — I just didn't want to spend a round removing a `console.info` argument.

So the ask is the same as before, on `5ca90958`: one pass to confirm it matches your `5a892946` experience. If it still feels off from `5a892946` after this, that's genuinely useful information — it would mean the difference is somewhere I haven't found, and I'd stop guessing and bisect it properly.

cnxiekun:
@openwong2kim Field test on 5ca90958 — idle experience is back to ~`5a892946` (close and acceptable). But streaming (Claude Code working) is now DIFFERENT from before: while typing during streaming, the caret/candidate anchors to the bottom-right of the whole wmux screen — whereas before it was on the streaming output row.

The log during streaming shows src=instant edge=1 at cursor=(145,33) (last column of the pane) — the quiet-caret snapshot is not being used, so it falls back to the last-column instant and anchors bottom-right.

https://github.com/user-attachments/assets/df8a9763-74d7-4a86-9aec-46745c632ff7

https://github.com/user-attachments/assets/5f7713d9-eb5b-4381-a70e-57575b5c8699

[main-2026-08-21.log](https://github.com/user-attachments/files/31288140/main-2026-08-21.log)

openwong2kim:
Your log named the cause outright, and it is not the snapshot — it is the gate in front of it. **`f881457a`.**

The failing records read:

```
src=instant edge=1  cursor=(145,33) sel=(145,33)  gap=203ms  caretAge=602ms
```

`caretAge=602ms` means a quiet-caret snapshot existed and was usable. `gap=203ms` means output was flowing. Those are exactly the conditions the streaming path was written for, and it still chose `instant` — the line-end park, hence the bottom-right corner.

What blocked it is the sustain gate. It measures from `epochStart`, which restarts every time a quiet span of `OUTPUT_QUIET_MS` (500 ms) ends, and Claude Code's token pacing pauses longer than that between bursts. So the epoch keeps resetting and never reaches `STREAM_SUSTAIN_MS` (700 ms), the caret branch never runs, and selection falls through to the live cursor. Your own log shows both sides of it two seconds apart — `02:24:51.783` caught an epoch that had lasted long enough and selected `src=caret sel=(50,33)`, correctly; `02:24:53.058` did not and gave `src=instant sel=(145,33)`.

That also explains why this looked like a regression from `8bb88adc` when the snapshot logic was the part that changed: the snapshot was fine all along (`sel=(50,33)` in the good record), it just was not reachable often enough.

**The fix:** while output is flowing, a line-end park defers to the snapshot without waiting out the epoch. The gate exists to stop a *commit echo* from pushing the next composition onto a stale snapshot — and it has nothing to protect there, because a commit leaves the caret where the syllable landed, mid-line, and a CJK composition cannot sit on the last column at all (the glyph is two cells wide).

I want to be explicit that this is a fourth attempt at acting on `edge=1`, after you measured three losing. The difference is the condition, not the idea: all three of those rerouted a line-end park in **every** state, idle included, and idle is precisely where the plain dwell selection was already right — which is what you kept measuring. This one cannot fire unless output arrived within the last 500 ms, so the idle behaviour you just confirmed as back to `5a892946` is untouched by construction. Both halves are pinned by tests: a line-end park mid-stream now takes the snapshot, and a mid-line cursor still waits out the gate so a commit echo keeps the correctly advanced cursor.

The idle line-end park remains a known residual, unchanged.

If you have time for another pass, the two things worth watching are whether the streaming anchor now lands on the input row, and whether consecutive typing while idle still feels like `5a892946`. A log with `src=caret` appearing during streaming is the signal that this path is live.

Thanks for the video plus the log together — the `gap`/`caretAge` pair on a failing record is what made this a five-minute read instead of another guess.

cnxiekun:
@openwong2kim Field test on f881457a — the streaming path is now live (src=caret edge=1 appears, 21 times), so the line-end park correctly defers to the snapshot while output flows. But two things to report.

1. idle — clearing the input box behaves differently depending on a click:

    Clearing the input box then starting to type again puts the caret at the bottom-right (wrong). But if I click somewhere else in the pane first and then start typing, the caret lands at the start of the input box (correct). So the click resets the caret to where the input actually begins; without it, the TUI parks the caret at the line end.

 2. Streaming — the caret is still not on the input row:
    
    While Claude Code streams, the caret is still not at the input position — sometimes it is below the input row, sometimes at the bottom-right of the whole screen.

The log shows the gate fix works — streaming compositions now select src=caret edge=1. But the snapshot itself is the last column: cursor=(127,36) sel=(127,36) src=caret edge=1. The snapshot was captured during the quiet span, but that span found the TUI parked at the line end (127), so deferring to it anchors the candidate window bottom-right. This matches observation 1: after a click the caret resets to the input start, so the snapshot is correct; without a click the quiet snapshot is the line-end park.

Video and log attached below. Happy to capture more if it helps.

https://github.com/user-attachments/assets/a713e23b-91cc-4aaf-bd95-4122706b3bbb

[main-2026-08-21.log](https://github.com/user-attachments/files/31291270/main-2026-08-21.log)

openwong2kim:
Your log has it exactly. **`ed56e47c`.**

The gate fix landed — `src=caret edge=1` twenty-one times, so a line-end park does defer to the snapshot now. It moved the miss one step earlier rather than removing it, and your two observations are the same fact seen from both ends.

The log carries both shapes side by side:

```
cursor=(127,43) sel=(53,43)  src=caret edge=1   ← working: cursor at the line end, snapshot on the input column
cursor=(13,36)  sel=(127,43) src=caret          ← broken: cursor mid-line at 13, snapshot at 127
```

That second line is the one that matters. The live cursor is at column 13, so the anchor should be there — but the snapshot says 127, and the snapshot now wins. Deferring to it put the authority behind the wrong cell instead of the right one.

**The fix:** a quiet span that finds the TUI parked on the last column keeps the caret the last real span recorded, instead of replacing it. Your observation 1 is the cleanest possible isolation of that — clicking elsewhere first resets the TUI's cursor to where the input begins, so the snapshot taken after the click is right; without the click the span spans the parked cell and the snapshot goes wrong. Same code path, and the click is the only variable.

**On reverting my own revert.** This is the exclusion `8bb88adc` carried and `5ca90958` removed after you measured it worse — so I want to be explicit about why it belongs now rather than quietly putting it back. When you measured it, the streaming branch was unreachable: the sustain gate read a token-paced stream as a run of isolated echoes, so the snapshot was almost never selected and the exclusion could only change which *fallback* ran. Your call on it was correct for that build. Now the branch is live and the snapshot is what it reads, so the rule finally protects the thing that uses it. Both of your measurements were right; they were measuring different machines.

**Still open, and I do not want to claim otherwise:** your observation 2 says the caret is *sometimes below the input row* during streaming, not only bottom-right. This change addresses the bottom-right half — the wrong column. A snapshot on the right column but the wrong ROW is a separate residual (the header lists it: a mid-stream pause longer than the quiet window with the cursor parked on repaint state re-snapshots a wrong caret). If the next pass still shows a caret one or two rows off the input line with `sel` on a sensible column, that is this one and it needs its own fix.

Worth watching in a log: `sel` should now stay off the last column entirely. Any `src=caret` record whose `sel` column equals the pane width minus one means the refusal missed a path.

openwong2kim:
@cnxiekun Gentle ping on the `ed56e47c` build — no rush, and no obligation if
you are out of bandwidth.

Where this stands: seven of your passes closed seven distinct causes, and the
one I flagged as still open is the only thing I am waiting on. It needs a log I
cannot produce — the failure only shows up with a real IME on a real pane.

What would settle it, if you get a chance:

- Type during streaming and check whether the caret still lands one or two rows
  **below** the input line. That is the residual I named: right column, wrong
  row. If it is gone, this is done.
- In the log, any `src=caret` record whose `sel` column equals pane width minus
  one means the last-column refusal missed a path. There should now be none.

If you would rather stop here, say so and I will merge as is and split the
residual into its own issue — the original #951 symptom (candidate window
jumping to the output rows while an agent streams) is fixed and confirmed by
your earlier passes, and that is worth shipping on its own. Either answer is a
good outcome; I just do not want to merge over an open question you raised
without hearing which one you prefer.

Thank you for the eight rounds. The logs made every one of them tractable.

cnxiekun:
@openwong2kim Field test on ed56e47c. Two streaming behaviors remain, both captured on video, and the log pins down a shared cause: the line-end refusal does not seem to be firing — sel is still on the last column.

Observation 1 — caret on the far right of the screen, pinyin wandering:

While Claude Code streams, the caret blinks at the far right of the screen, and the pinyin jumps up/down/left/right. The log matches: sel=(137,36) on a 139-col pane — the last column — with pin=(843.0,86.7) pushing the candidate window to the right edge.

https://github.com/user-attachments/assets/f586bf7a-e98b-40a8-8250-13b62941b29d

Observation 2 — caret at the bottom-center, pinyin wandering:

In another case the caret blinks at the bottom-center of the screen. The log shows sel=(53,39) (row 39 of a 41-row pane — near the bottom), with src=caret.

https://github.com/user-attachments/assets/c3ca904f-4b0a-44e0-826f-ca449f31d669

The shared cause — the snapshot is still a line-end/bottom cell:

src=caret records still carry a last-column sel: cursor=(22,31) sel=(137,36) src=caret, 21 records at sel=137. The live cursor is at column 22 (the input), but the snapshot it defers to is column 137 (the line end) — so the refusal ed56e47c added is not stopping the line-end cell from becoming the snapshot. edge=1 does not appear on these records either, which suggests the refusal path is not the one taking the snapshot.

Videos and the full main-2026-08-23.log attached.

[main-2026-08-23.log](https://github.com/user-attachments/files/31350459/main-2026-08-23.log)

> Honest observation — this tail may be more stubborn than any of us expected, and perhaps part of it is also how Claude Code itself handles its caret, not entirely wmux's side. Either way, no rush: the current state is already a big improvement over where we started, and we have all put real work into it. Whatever remains can wait for a later round. If it is simpler to land something now, reverting to 5a892946 or any build you judge stable is fine from my side.

openwong2kim:
@cnxiekun Your log settles it, and your closing suspicion — that part of this is how Claude Code handles its caret — is exactly right. Two facts fall out of the records:

**1. The refusal checks the wrong column.** `ed56e47c` refuses a snapshot on the grid's last column — `cols - 1`, column 138 on your 139-column pane. Your park sits at column 137: the cursor rests just past the end of the text the stream last painted, and that painted line happened to end two cells short of the grid edge. The check passes it, the promotion stands, and every composition anchors there with the snapshot's authority — `sel=(137,36) src=caret`, twenty-one records. Observation 2 is the same shape with the mask off: `sel=(53,39)` is a park in the middle of the line, nowhere near any column bound. A park is "wherever the painted text ended", and that can be any column at all. A column threshold can only ever catch the full-width special case — your earlier 128-column log, where the park landed exactly on column 127, is why it wore the name "line-end park" for three rounds.

**2. There is no cursor signal left to read.** While Claude Code streams, its cursor rides the repaint and rests at the end of painted content between bursts. It is never placed on the input caret during the stream — the upstream behavior you suspected (anthropics/claude-code#25186; Ink's `useCursor` exists, Claude Code has not shipped it). Everything in this PR reads the cursor: dwell, quiet spans, column checks. When the true caret is a cell the cursor never visits, no combination of those rules can find it. The header predicted this ceiling as a known residual; your two videos are the proof it has been reached.

So I am taking your suggestion and landing this build. What it ships for good: idle anchoring correct across your seven closed causes, streaming compositions deferring to the snapshot as designed, and a diagnostic that names its own failures precisely enough that this round needed no guessing. What it does not ship: a correct anchor while output streams — the snapshot machinery cannot know where the input line is, because the cursor never goes there. That residual moves to #1016 with a different signal class entirely: reading the agent's input-line marker out of the buffer, gated per recognized agent, instead of one more cursor heuristic.

Eight field passes, seven distinct causes closed, on an IME and a language I cannot reproduce here — every fix in this PR was steered by a log you took the time to capture. Thank you. #1016 starts from everything this thread taught it, and there is no obligation whatsoever, but if it reaches a build worth testing I would be glad to ping you.

openwong2kim:
For the record on this thread: the content-signal follow-up just landed on main as 055ed239 (#1017, closing #1016). While Claude Code streams, a composition now finds the input box in the buffer itself — the `│ > ` prompt row under its border — and anchors there (`src=marker` in the new `ime-anchor5` diagnostic), instead of trusting any cursor-derived cell. Idle behavior is untouched. No obligation whatsoever, as before — but if you ever feel like one more pass on a main build, that log line is the one that will tell the story.

Issue #936, opened 2026-08-18
Codex CLI in wmux pane (Windows 10, wmux 3.43.0): mouse wheel cannot scroll prior conversation


lishimin8507-droid:
### Environment
- wmux: 3.43.0 (latest)
- OS: Windows 10 22H2, build 19045 — so the #923 bundled-ConPTY path should activate
- Agent: OpenAI Codex CLI (latest)
- Reproduction pane: any wmux pane running `codex`

### Symptom
In a wmux pane running `codex`, scrolling the mouse wheel up/down does not scroll back to view prior conversation. Keyboard `PgUp` / `PgDn` also do not move within the conversation.

### Differential reproduction — isolates wmux as the broken factor
I tested all three of these on the same Windows 10 box, same wmux install:

| # | Where | What I scrolled | Mouse wheel |
|---|---|---|---|
| 1 | wmux + plain Git Bash / PowerShell pane | shell scrollback | ✅ works |
| 2 | wmux + `codex` | prior conversation context | ❌ does NOT work |
| 3 | plain Windows Terminal + `codex` | prior conversation context | ✅ works |

The only configuration that fails is #2 (wmux + codex). Cases #1 and #3 each share one factor with #2 and both work — that isolates wmux as the broken factor for the codex case.

### Why this is not a Codex TUI design issue
- Codex's PTY output in wmux contains **zero** `\e[?1000h` / `\e[?1006h` sequences (counted across the live byte stream of `~/.wmux/buffers/daemon-*.buf`) → Codex does **not** request mouse capture, so wheel events should fall through to xterm.js's native scrollback handling.
- Codex's PTY output contains **zero** `\e[2J` (clear-screen) and very few `\e[H` → Codex's painted conversation lines accumulate in xterm.js's scrollback rather than being wiped each frame.
- wmux's renderer (`index-*.js`) has no custom `onWheel` handler — `deltaY` appears only in React DOM event-type tables, not in any actual wheel listener — so wheel events should reach xterm.js natively.
- Codex responds to wheel in plain Windows Terminal (#3) → Codex itself does support the affordance.

### Possibly related
- #910 / #923 — same class of "mouse/wheel doesn't reach the TUI on Windows 10"; #923 fixed it for vim by spawning the bundled `conpty.dll` on builds < 22000. This report is on build 19045, where the bundled-DLL path should be active.
[...]

PR #954, opened 2026-08-20
fix(terminal): render the cursor only at rest for synchronized-output TUIs

### LINKED ISSUE #929: [Bug]: Codex TUI shows two blinking cursors on Windows
### What happened?

Two independent blinking cursors are visible at the same time:                                                
                                                                                                                
  1. One cursor appears in the Codex input/composer area before                                                 
     `Use skills to list available skills`.                                                                     
  2. Another cursor blinks after the working-directory text                                                     
     `D:\project\week_report`.                                                                                  
                                                                                                                
  The second cursor is in the status path, not in a filename input field.    

### Steps to reproduce

1. Open a terminal pane in wmux.                                                                              
  2. Run `codex`.
  3. Submit a prompt, for example `我更改了什么`.                                                               
  4. Wait while Codex is executing a command.                                                                   
  5. Observe the Codex composer and the bottom status line.   

### Expected behavior

  Only the active Codex input caret should be visible. The terminal's native cursor should be hidden or disabled
  while Codex owns the full-screen TUI.

### wmux version

3.43.0

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

 Windows 11  

None / not relevant

```shell

```

### Confirmation

openwong2kim:
Fixes #929.

## What the second caret is

Byte-level work on #929's sibling investigation (#936) settled the composition of the "two blinking cursors": there is only **one** real cursor, and it teleports. Codex ends some `?2026` synchronized-output frames — status-line ticks, the "Working" shimmer — with the hardware cursor **visible and parked where the frame's last text write finished**: after the model/cwd status text (`row29 col29`, right after `… · ~`) and next to the spinner. Replaying a captured session frame-by-frame: of 336 sync frames, 115 end at the composer, 125 end mid-transcript during streaming, and 9 end on the status row — every one of them with the cursor shown. Both positions the reporters named (after the cwd path, next to "Working") appear verbatim in the held-state list.

Frame-level video analysis of the reporter's recording confirms the illusion: whenever the status-row caret is lit, the composer block is dark (`Sc` correlation in every frame) — one cursor, bouncing, integrated by the eye into two.

Timing measurements (chunk-timestamped capture): the stray states persist 17–23 ms on a fast Win11 box — 1–2 render frames — and at least a full video frame (≥42 ms) on the reporter's machine. Honoring `?2026` cannot help; these are **complete** frames whose end state is bad. Upstream reports of the same family: openai/codex#9081 (closed not-planned), openai/codex#32546 (open).

## The fix

The resting-cell idea from #888, applied to cursor rendering: every output chunk carrying a `?2026` marker gets a DECTCEM hide appended, and the cursor is re-shown only after **32 ms with no further frame traffic** — and only if the application's own last DECTCEM asked for a visible cursor (its intent is tracked before our injection and never overridden).

- Behavioral gate, not an agent list: only `?2026`-bracketing frame writers arm the guard. vim/less/htop never see it.
- Simulated over a real byte-timestamped codex session (40 s, prompt → streaming → Working): stray-position cursor visibility drops **1,939 ms → 144 ms (−93%)**; the composer caret keeps its resting visibility.
- The show-injection routes through the same ordered write path as PTY output (resync hold-out included), and the guard is disposed before the terminal (the #582 lesson).

## Tests

`restingCursor.test.ts`: pass-through without `?2026`, hide+show cycle, burst coalescing, app-hidden cursor never re-shown, last-DECTCEM-wins intent tracking, injected hide not mistaken for app intent, dispose cancels the pending show.

### DIFF
--- a/changelog.d/954.md
+++ b/changelog.d/954.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **Codex panes no longer show a second blinking cursor on the status
+  line.** Codex ends some of its synchronized-output frames with the
+  hardware cursor visibly parked after the model/cwd status text or next to
+  the "Working" spinner, and the terminal faithfully painted it there for a
+  frame or two — quasi-periodic flashes that read as a second caret blinking
+  alongside the composer's. The cursor is now rendered only at rest: frame
+  traffic hides it, and it reappears 32 ms after the last frame — at the
+  composer, where the resting position actually is. Classic TUIs (vim, less,
+  htop) never bracket frames, so their cursor behavior is unchanged. (#954)
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -33,6 +33,7 @@ import { createDeadInputWatchdog } from '../terminal/deadInputWatchdog';
 import { awaitParseBarrier } from '../terminal/parseBarrier';
 import { STALE_REPLAY_INPUT_MODE_RESETS, STALE_REPLAY_ALIVE_SHELL_RESETS, STALE_REPLAY_DISPLAY_RESETS, staleReplayResetLevel } from '../terminal/staleReplayModeReset';
 import { attachAltScreenWheel, PAGE_SCROLL_AGENTS } from '../terminal/altScreenWheel';
+import { RestingCursorGuard } from '../terminal/restingCursor';
 import {
   writeTerminalOutput,
   flushTerminalOutput,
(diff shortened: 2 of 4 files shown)


openwong2kim:

**P3** — fix for #929 (Codex TUI showing two blinking cursors on Windows).

This PR adds a `RestingCursorGuard` (`src/renderer/terminal/restingCursor.ts`) that appends a cursor hide (`?25l`) to every output chunk carrying a `?2026` synchronized-output marker and re-shows the cursor only after 32 ms of frame silence, tracking the application's own DECTCEM intent so a cursor the app hid is never forced back on. The wiring in `useTerminal.ts` routes the injected show through `deliverPtyData` so ordering with queued output holds and the guard cannot re-enter itself, and it is disposed before the terminal teardown, which correctly avoids the #582 late-inject class of bug. Scope is behavioral rather than per-agent — only `?2026` frame writers arm the guard, so vim/less/htop cursor behavior is untouched — and with the measured ~93% reduction in stray-position cursor visibility over a real captured codex session, seven focused unit tests, and green CI across validate, bench, and all three platform baselines, this is a low-risk fix that is ready to merge. The changelog entry is accurate and reads well as-is.

PR #956, opened 2026-08-20
fix(pane): keep pane decoration rings under overlays and modals

### LINKED ISSUE #946: [Bug]: Border of the terminal window lives on top of the 'new window' modal
### What happened?

Image

### Steps to reproduce

1. Have a terminal pane light up (my assumption is that means it's 'ready') (I'm using claude code here)
2. Click '+' sign to add a new workspace
3. Border appears above the modal window

### Expected behavior

1. Have a terminal pane light up (my assumption is that means it's 'ready') (I'm using claude code here)
2. Click '+' sign to add a new workspace
3. Border appears below the modal window

### wmux version

3.44.0

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

Windows 10 19045.6466

None / not relevant

```shell

```

### Confirmation

openwong2kim:
Fixes #946.

The completed-blink ring (`.pane-complete-blink::after`), the notification ring (`.notification-ring::before`), and the pane flash (`.pane-flash::after`) are pseudo-elements at `z-index: 50`/`100` on a pane root that creates **no stacking context** — so they compete document-wide. At 50 they tied with the `--z-overlay` baseline (dropdowns, the new-workspace picker) and won on DOM order, drawing the green ring straight across the picker; at 100 the flash outranked even modals and dialogs for its 500 ms.

**Fix**: all three drop to `z-index: 40` — pane decorations only need to beat pane content, and the pane-local stack caps at the toolbar's `z-30`. 40 keeps them above everything inside the pane and below every piece of app chrome (`--z-overlay: 50` and up), slotting into the documented stacking scale beside `--z-cheatsheet: 40`.

**Why not `isolation: isolate` on the pane root** (the direction floated in triage): a stacking context on the pane would also trap pane-internal popovers (`AgentToolbar`'s RichInput is 30 rem wide) that legitimately overflow onto a neighboring pane — a later-DOM sibling pane would then paint over them. Lowering the decorations fixes the reported class without that regression, and without adding a new stacking context for future authors to reason about.

**Verified live**: injected the blink+notification ring on a pane, opened the new-workspace picker over its edge — the picker paints cleanly over the ring (before: the ring crossed the picker exactly as in the report's screenshot). Pane tests (130) pass.

### DIFF
--- a/changelog.d/956.md
+++ b/changelog.d/956.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **A pane's attention ring no longer draws on top of menus and modals.** The
+  green completed-blink ring, the notification pulse, and the pane flash sat
+  at overlay-level z-index without a containing stacking context, so a pane
+  that lit up could paint its ring straight across the new-workspace picker
+  (and the flash briefly over any modal). The decorations now sit above
+  everything inside their pane and below all app chrome. (#956)
--- a/src/renderer/styles/globals.css
+++ b/src/renderer/styles/globals.css
@@ -479,7 +479,13 @@ select option {
   inset: 0;
   border-radius: inherit;
   pointer-events: none;
-  z-index: 50;
+  /* #946: pane decorations must sit above pane content (toolbar caps the
+   * pane-local stack at z-30) but BELOW the --z-overlay(50) baseline. The
+   * pane root creates no stacking context, so these pseudo-elements compete
+   * document-wide: at 50 they tied with dropdowns/pickers and won on DOM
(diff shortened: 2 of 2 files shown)


openwong2kim:
Structural follow-up filed as #957: this PR closes the collision class by ordering (decorations below the overlay baseline), but the pane root still creates no stacking context, so pane-internal z values keep competing document-wide. The containment fix (portal the toolbar popovers, then `isolation: isolate` the pane root) is scoped there — until it lands, any future pane-internal element at z≥50 would reintroduce this class.

PR #959, opened 2026-08-20
fix(rpc): stop reading the plugin host as the operator on private-event and channel reads

openwong2kim:
Part of #922.

## Scope — read this first

#922 has three tracks, and this PR is **not** the two the title of the issue names. Here is where it fits:

- **`declared` / `legacy` wire lanes** (the issue's headline): need the peer-credential primitive, deferred with the #113 grandfather clock. **Untouched here.**
- **PR1 — the `hosted` lane in `browser.rpc.ts`**: landed as #941. It derives the plugin host's workspace server-side and applies it to the `browser.*` table.
- **PR2 design pass** (snowyukitty, 2026-08-19): while enumerating the fallback family, its **section 5** found a separate, higher-value hole — `firstParty` is read as "the operator" at two read gates, and the plugin host is inside `firstParty` but is *not* the operator. The design pass explicitly asked (Q4) whether section 5 should be **its own change** rather than folded into the router work. **This PR is that change.**

So: this closes the `firstParty`-conflation leaks for the caller class whose workspace #941 already derives. The router-level mechanism (A–D: `pane.list`, `input.readScreen`, `browser.open` fallbacks) remains PR2's, unblocked by this.

## The two leaks

Both trusted bare `ctx.firstParty` as "a human operates every local workspace" — true for the renderer bridge, false for the iframe plugin host (`RpcContext`'s own doc: *"The iframe plugin host is also in-process but is NOT the operator."*).

**1. `events.poll` private-event scope** (`events.rpc.ts`). `privateSet = clientSet` on bare `firstParty`, so an approved plugin gated another workspace's PRIVATE events (`a2a.task`, `channel.*`) by naming its id. Cost: `events.subscribe`, which `hello-panel` declares. Repro from the design pass:
```
dispatch { method: 'events.poll', params: { workspaceId: 'ws-victim-a', types: ['a2a.task'] }, clientName: 'hello-panel' }
       as { firstParty: true, hostedWorkspace: 'ws-plugin' }
→ {"events":[{"type":"a2a.task","from":"ws-victim-a","to":"ws-victim-b","messagePreview":"PRIVATE"}]}
```

**2. `a2a.channel` reads** (`a2a.channel.rpc.ts`). The no-`senderPtyId` read path left the caller-supplied `verifiedWorkspaceId` in place under process-boundary trust. A plugin could read any channel a **named** workspace was a member of, clear the `ws-human` first-party gate into *every private channel with full history* (`a2a.channel.read`), and aim `task.mission.list` — titles, branches, absolute worktree paths, pane group ids — at any workspace.

## The fix

`isHostedCaller` / `hostedBindingOf` (`shared/rpc.ts`) key on the **presence** of `hostedWorkspace` (its `null` state still means "hosted, unbound"), mirroring #941's browser lane.

- **`events.poll`**: operator keeps the caller-supplied scope; a hosted caller's private scope **is** its server-derived binding (the claim is ignored, exactly the B3 agent-path rule); an unbound (`null`) hosted caller fails closed to the empty set. Lifecycle types keep every-subscriber semantics for everyone.
- **`a2a.channel` reads**: the binding is stamped over any self-claimed `verifiedWorkspaceId` (D5's rule, applied to this caller class); an unbound hosted read is refused before the daemon is reached; the `ws-human` gate is now **operator-only** rather than `firstParty`-only.

Plugin *mutations* were already refused on this path (no `senderPtyId`) — unchanged.

## Attacker model & ceiling

Threat closed: an **approved** plugin (same user, same machine) that names, or omits, a foreign `workspaceId` to read another workspace's private events / channels / missions. It controls neither half of its identity — `clientName` is stamped from the manifest, the workspace is derived by the host — so for this caller class the binding is a genuine server-side ownership fact, not a presence check.

Ceiling, per #941 and written into the changelog: this confines an approved plugin to the workspace hosting it. It is **not** containment of hostile same-user code, and it does **nothing** for the wire — external MCP callers keep landing in `declared` / `legacy` and still need the peer-credential primitive #922 is really about.

## Tests

`events.rpc.test.ts` (+4): hosted caller's claim ignored, operator unaffected (distinguishes, not narrows), unbound fails closed with lifecycle intact. `a2a.channel.rpc.test.ts` (+3, one renamed): hosted read stamped to binding across `list`/`getMessages`/`task.mission.list`, ws-human self-claim stamped not refused, unbound read `NOT_AUTHORIZED` before the daemon. `src/main/pipe` suite: 744 passed. tsc + eslint clean.

### DIFF
--- a/changelog.d/959.md
+++ b/changelog.d/959.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **An approved plugin can no longer read another workspace's private
+  activity by naming it.** The iframe plugin host dispatches through the same
+  trusted in-process channel as the app's own UI, and two read paths —
+  `events.poll` (private `a2a.task` / channel events) and the a2a channel /
+  mission reads — treated that shared channel as "the operator", so a plugin
+  granted an ordinary read capability could aim those reads at any workspace
+  by passing its id (or omitting one). Both paths now bind such a plugin to
+  the workspace it is actually hosted in, matching the scope its approval
+  implied; an unbound plugin read fails closed. This confines an approved
+  plugin — it is not a defense against hostile same-user code, and external
+  (non-plugin) callers are unchanged. (#959)
--- a/src/main/pipe/handlers/__tests__/a2a.channel.rpc.test.ts
+++ b/src/main/pipe/handlers/__tests__/a2a.channel.rpc.test.ts
@@ -234,7 +234,7 @@ describe('a2a.channel.rpc — ws-human read scope is first-party only (W1)', () 
   );
 
   it.each(['a2a.channel.list', 'a2a.channel.getMessages'] as const)(
-    'first-party %s scoped to ws-human passes through (the renderer/plugin-host read path)',
+    'OPERATOR first-party %s scoped to ws-human passes through (the renderer read path; hosted callers are stamped to their binding instead — see the #922 suite)',
     async (method) => {
       const expected = { ok: true, value: [] };
       const daemon = makeFakeDaemon((m, params) => {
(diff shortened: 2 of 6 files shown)


openwong2kim:

This PR closes the two `firstParty`-conflation leaks identified in section 5 of the #922 design pass: `events.poll` treated any first-party dispatch as the operator and trusted the caller-supplied workspace scope for private event types, and the no-`senderPtyId` read path in `a2a.channel.rpc.ts` left the caller-supplied `verifiedWorkspaceId` in place — together letting an approved plugin read another workspace's private `a2a.task`/`channel.*` events, channel history via `a2a.channel.read`, and `task.mission.list` data simply by naming the workspace id. The fix introduces `isHostedCaller`/`hostedBindingOf` in `src/shared/rpc.ts` (keyed on the presence of `hostedWorkspace`, mirroring the #941 browser lane) and applies them at both gates: a hosted caller's self-claim is stamped over with its server-derived binding, an unbound hosted caller fails closed (empty private set for events, `NOT_AUTHORIZED` before the daemon for channel reads), and the `ws-human` gate becomes operator-only rather than first-party-only. The scoping is right — the renderer/operator path is preserved and explicitly distinguished in the tests rather than narrowed, lifecycle events keep their every-subscriber semantics, and the `declared`/`legacy` wire lanes from #922's headline stay untouched for the PR2 router work. Test coverage matches the threat model (hosted claim ignored, operator unaffected, unbound fails closed, ws-human self-claim stamped rather than refused), and the changelog states the ceiling honestly: this confines approved plugins, not hostile same-user code. P2, on par with parent #922 — landing this before the PR2 router work keeps the caller-class invariants consistent across the read gates.

PR #961, opened 2026-08-20
fix(perf): confirm a reproduced red on a fresh runner, and let that job carry the verdict

### LINKED ISSUE #940: Perf gate: the confirmation re-run shares the runner it is confirming, and frameBudget is quantized so 52% of main already sits at 2x baseline

openwong2kim:
Fixes #940 — the third proposal, the one #947 deliberately left for a topology call: **confirm on a fresh runner**.

## What remained after #947

#947 took the wording fix and the frame-margin verdict shape. The unresolved case is the one the evidence table opens with: a red that **reproduces on the immediate same-runner re-run** and then goes green on a different machine 20 minutes later. The same-runner re-run is a check on the code's determinism, not on the machine's fitness to measure — and until now the only mechanism that told a real regression from a lifetime-degraded runner was a human clicking "re-run all jobs" (which is what happened before v3.44.0 was tagged).

## The design

**Escalation is deliberately narrow.** `perf-compare --escalate <path>` changes behavior in exactly one case: the same-runner confirmation RAN and the failure REPRODUCED. Then the failing plan (commit, gates, legs, bench args) is written to the marker and the step exits 0 — the verdict moves to the job that can render it. Everything else is untouched: a cleared blip needs nothing, and an unconfirmable red (correctness gate, harness failure) still fails closed on the first runner — "could not measure it again" is not a measurement question and must not ride to another machine as one.

**`bench-confirm` carries the verdict.** A dependent job (`needs: bench`, fresh machine by construction) runs `perf-confirm-fresh.mjs`, which packages the app, re-measures only the escalated legs, and judges them against the same baseline with the same fail-closed contract as the in-job re-run:

- every escalated gate passes → **green**: measured twice on one machine, not at all on another — runner-health noise. The first sample stays in the trend (it records what was measured, not what was excused).
- anything fails → **red**, reproduced on a *second machine* — the strongest claim this pipeline can make.
- anything unverifiable → **red**.

**The cross-job handshake is bound to the commit, not pretended away.** The in-job design (#632) exists partly to avoid a handshake file; a dependent job cannot avoid one, so it is bound to the only identity that matters: the escalation names the short SHA the gate measured, `bench-confirm` checks out `github.sha` explicitly (a moved PR base cannot swap the merge commit under it), and both its own HEAD and the fresh sample's recorded SHA must match or it refuses.

**An escalated red exits the first job green, so every link is load-bearing and pinned.** `perfWorkflow.test.mjs` gains contract checks + mutation coverage for the whole chain: the `--escalate` flag on the exact gate invocation, the literal existence-check step (no `if:`, tests the very file the gate writes), the `fresh-confirm-needed` output wiring, `bench-confirm`'s `needs`/`if` topology, and the single-line verdict command with nothing chained after it. 10 new mutants prove each check actually fires.

## Tests

- `perfWorkflow.test.mjs`: 30 pass (contract + mutants, incl. 10 new #940 links).
- `perfConfirm.test.mjs`: 53 pass — `confirmGate` now returns the escalation plan **only** on a reproduced red (null on cleared and on unconfirmable, with the rationale in comments), plus 4 end-to-end `--escalate` cases through the real CLI (reproduced→0+marker, not-reproduced→0+no marker, correctness red→1+no marker, green→untouched).
- `perfConfirmFresh.test.mjs` (new): 15 pass — `parseEscalation` validation, and the CLI end-to-end: non-reproducing→0 with the first sample untouched, reproducing→1, correctness break in the fresh run→1, commit mismatch (stale/replayed artifact)→1 before any bench runs, nonzero bench→1, guarded-file clobber→restored+1, missing flag→2.
- `bench/README.md` documents the topology and its posture.

Not run here: a live CI activation (needs a real regression or a degraded runner). The failure paths are all fail-closed, so the worst outcome of an undiscovered wiring bug is the pre-#940 behavior — a red that needs a human — never a silent green.

### DIFF
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -34,6 +34,13 @@ jobs:
     timeout-minutes: 45
     permissions:
       contents: write
+    # #940: `fresh-confirm-needed` is what hands a reproduced red to the
+    # bench-confirm job below. It is derived from the escalation FILE the
+    # compare step writes, never from the step's exit code — if the file could
+    # not be written, compare kept the red itself, so a missing file always
+    # means "nothing to escalate".
+    outputs:
+      fresh-confirm-needed: ${{ steps.escalation.outputs.needed }}
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
         with:
@@ -85,8 +92,29 @@ jobs:
       # anywhere and nothing chained after the command that could swallow it.
       # scripts/__tests__/perfWorkflow.test.mjs pins that, because nothing else
       # checks YAML.
+      #
+      # --escalate (#940): the one red the same-runner re-run cannot settle —
+      # it REPRODUCED here — is written to out/perf-escalation.json and this
+      # step exits 0; the bench-confirm job below re-measures on a fresh
+      # machine and carries the verdict. Every other red (transient blips that
+      # cleared, correctness gates, harness failures) keeps its exit code
+      # here, exactly as before.
       - name: Compare against baseline
-        run: node scripts/perf-compare.mjs --current out/perf-current.json --baseline bench/baseline-ci.json --summary perf-summary.md --confirm-retry out/perf-retry.json ${{ steps.histflag.outputs.value }}
+        run: node scripts/perf-compare.mjs --current out/perf-current.json --baseline bench/baseline-ci.json --summary perf-summary.md --confirm-retry out/perf-retry.json --escalate out/perf-escalation.json ${{ steps.histflag.outputs.value }}
+
+      # #940: turn the escalation file's existence into the job output that
+      # triggers bench-confirm. Nothing else — the file is the contract, and
+      # perfWorkflow.test.mjs pins that this step stays a literal existence
+      # check (an `if:` here, or a rewrite, could silently orphan an escalated
+      # red into a green run).
+      - name: Detect escalation
+        id: escalation
+        run: |
+          if (Test-Path out/perf-escalation.json) {
+            "needed=true" | Out-File -FilePath $env:GITHUB_OUTPUT -Append
+          } else {
+            "needed=false" | Out-File -FilePath $env:GITHUB_OUTPUT -Append
+          }
 
       - name: Append summary to step summary
         if: always()
@@ -103,6 +131,7 @@ jobs:
           path: |
             out/perf-current.json
             out/perf-retry.json
+            out/perf-escalation.json
             out/perf-history-line.ndjson
             perf-summary.md
           if-no-files-found: ignore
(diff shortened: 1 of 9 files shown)


openwong2kim:

This PR lands the piece #947 deliberately left as a topology call — proposal 1 from #940 — by turning the human "re-run all jobs" click into a dependent CI job. `perf-compare --escalate` now hands exactly one case upward: a red that REPRODUCED on the same-runner confirmation re-run gets its failing plan (commit, gates, legs, bench args) written to `out/perf-escalation.json` and the compare step exits 0, after which a new `bench-confirm` job re-measures the escalated legs on a fresh Windows runner and carries the gate's verdict — green with runner-health reasoning when the failure cannot be reproduced elsewhere, red when it reproduces on a second machine, and red (fail-closed) for anything unverifiable.

It touches nine files, all CI and bench-side: the `perf.yml` escalation-detection step and `bench-confirm` job wiring, `perf-compare.mjs` and `perf-confirm.mjs` (the escalation payload is produced only on the reproduced-red path; cleared blips and unconfirmable reds still fail or settle exactly as before), the new `perf-confirm-fresh.mjs` with a three-way commit binding (escalation ↔ first-run sample ↔ checkout HEAD, plus the fresh sample's own recorded SHA) and guarded-file restoration, contract and mutation coverage in `perfWorkflow.test.mjs` pinning every link of the chain, and a changelog entry. No production code changes.

This matters because all three real gate activations on `main` — including one twenty minutes before the v3.44.0 tag — ended in a human resolving a red that a different machine could not reproduce, and this removes that manual step from the release path while keeping the failure posture honest: the worst outcome of a wiring bug is the pre-#940 behavior (a red that needs a person), never a silent green. Priority P2 — release-pipeline reliability rather than a product bug, but it sits directly on the release path and the escalation design is well-scoped and pinned by tests.

PR #965, opened 2026-08-21
Let a turn the agent starts by itself report that it is running

openwong2kim:
Closes part of #935 — the direction #939 left open.

## The bug

A pane's `agentStatus` is a level, but every writer of it is edge-triggered. Inside the daemon a detector or hook terminal edge settles the pane, and from then on byte throughput was ignored entirely, so exactly two things could start the next turn: a submitted-input boundary, or an explicit `running` edge.

A turn the agent starts **by itself** produces neither. A background task finishes, the agent picks up and works, and nothing submitted any input; the only sources of a `running` edge are hook kinds a plugin-less install does not wire (`PostToolUse`, the permission-gate `PreToolUse`) and `AgentDetector`'s one-shot "agent started" at gate time. So the `complete` written when the previous turn ended stayed on the pane for the whole autonomous turn.

Reproduced live on the unfixed build — the pane held `complete` across 49 seconds of visible work, then recorded a second `complete` ending a turn it never reported starting:

```
24.5s  activity.active            <- the submitted turn
29.4s  agent.event complete       hook/agent.stop
       ... 49s of visible work, no status event at all ...
79.1s  agent.event complete       hook/agent.stop
```

## The fix

Bytes now count while a pane is settled, and a burst clearing the threshold promotes it. Three guards keep that from being the loose heuristic the settle was introduced to stop:

- **A six-second cool-down after the turn end.** A turn keeps painting after its Stop hook, and that tail is not a new turn. The length is set by the tighter of two constraints: the status a promotion writes is cleared by the byte-silence idle 5 s later, and main drops any idle landing within `AGENT_EVENT_SUPPRESSION_MS` (10 s) of the last lifecycle event — so a promotion earlier than six seconds could strand a pane reporting `running` after it finished.
- **The user's keystrokes must have gone quiet.** A pane echoing a draft message is not a pane doing work. The window covers a full activity window, so bytes banked before a keystroke cannot combine with the echo that followed it.
- **The resize-redraw window still disqualifies a repaint**, now including the repaint storm an unmute releases.

A burst failing a guard hands the cycle back instead of consuming it — otherwise a real turn arriving behind a rejected one would stay silent for as long as it kept the idle timer alive, which for a TUI painting a live counter is the whole turn.

`awaiting_input` is exempt: it means a human has to act, and only a human acting retires it. The flag survives both a later detector `waiting` off the footer under the approval box and the `running` that HookIngest projects for `agent.subagent_stop`.

The promotion deliberately does not reset the detector's emission dedup. The unsettled path clears it so a new turn's footer can speak again; doing it here would let idle chrome still on screen answer a fresh `complete` with a stale `waiting`, trading a silent wrong status for a loud one.

## Why bytes, and not a text pattern

The obvious alternative — teach `AgentDetector` a `running` pattern — was tried in #943 and is being closed in favour of this. Its anchor was `esc to interrupt`, and current Claude Code does not paint that string: **zero occurrences across 41 MB / 448k lines of live pane output on this machine**. A version-specific chrome string is also agent-specific, where the byte path covers every agent at once.

The byte path is viable because the numbers separate. Measured against a real Claude Code pane at 140x41, peak bytes per 3 s window against the 2 KB threshold:

| phase | peak 3 s | over threshold |
|---|---|---|
| idle, 20 s | 528 B | no |
| typing, 5 chars/s | 1493 B | no |
| typing, 20 chars/s | 1523 B | no |
| 600-char paste | 1374 B | no |
| **working** | **2757 B** | **yes** |
| after the turn, 12 s | 0 B | no |

That TUI repaints differentially, which is why typing is cheap — about 13 bytes per keystroke, not a frame. The keystroke guard is what keeps the separation from depending on one agent's repaint economy.

## Verification

`scripts/autonomous-turn-status-dogfood.mjs` drives the whole scenario against a private daemon instance and a real agent: it launches a background task, ends the turn, and watches what the pane reports while the completion resumes it. It also types a 60-character draft into a settled pane and asserts nothing reports running. It fails on the unfixed build and passes on this one:

```
29.4s  agent.event complete       hook/agent.stop
75.1s  activity.active            <- the autonomous turn, now reported
78.2s  agent.event complete       hook/agent.stop
```

Full suite green (11,770 passing). The regression tests that encoded the old absolute rule were rewritten rather than deleted, so what replaced it is pinned in the same place.

## Known limits, not fixed here

- **Local (non-daemon) mode is unchanged.** The gate lives in `DaemonPTYBridge`; `PTYBridge` has no equivalent, so the two modes can disagree on this case. Daemon mode is the packaged path.
- **A `running` edge still clears the pane's question state.** `HookIngest` projects `agent.subagent_stop` and the metadata kinds as `running`, and that broadcast moves the pane out of the "needs you" count on its own. That predates this change and is untouched by it; the exemption added here only stops *bytes* from doing the same.
- **An autonomous turn whose output lands entirely inside the user's typing window is delayed, not reported.** The promotion waits for the keystrokes to stop. Bounded, and strictly better than never promoting.
- **The clearing idle can still be swallowed** if a lifecycle event lands between the promotion and the idle, restarting main's 10 s suppression. The cool-down removes the case this change creates; the general one belongs in the router.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved pane activity reporting so autonomous agent work is shown as **running** after a turn completes.
  - Prevented typing echoes, terminal repainting, and minor output bursts from falsely marking settled panes as active.
  - Preserved **awaiting input** status while approval prompts remain unanswered.
  - Activity now resumes only after genuine agent output, appropriate cooldowns, and quiet input periods.

- **Tests**
  - Expanded coverage for turn transitions, autonomous activity, input handling, repainting, and status accuracy.

### DIFF
--- a/changelog.d/965.md
+++ b/changelog.d/965.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **A pane reports that it is working when the agent starts a turn by itself.**
+  A background task finishes, the agent picks up and works for a minute — and
+  the pane kept wearing the previous turn's `complete` for the whole thing,
+  because the only two things that could start a turn were you pressing Enter
+  and a hook that a plugin-less install does not wire. Byte activity now counts
+  while a pane is settled, held back by a cool-down after the turn end, by your
+  own keystrokes still echoing, and by a resize still repainting. A pane waiting
+  on an approval is left alone: only answering it retires a question.
--- a/scripts/autonomous-turn-status-dogfood.mjs
+++ b/scripts/autonomous-turn-status-dogfood.mjs
@@ -1,0 +1,264 @@
+// Dogfood #935 direction 2: does a turn the agent starts BY ITSELF report
+// `running`, or does it wear the previous turn's `complete` for its whole
+// length?
+//
+// Unit tests cannot answer this. The first attempt at this fix (#943) passed
+// its own tests and was inert in production, because the thing that decides is
+// the interaction between three live parts: the real Stop hook's arrival time,
+// the real TUI's byte pattern, and the daemon's status gate. So this drives all
+// three for real.
+//
+// Isolation: the daemon runs under a private WMUX_DATA_SUFFIX, so its control
+// pipe name and its `~/.wmux<suffix>/daemon-pipe` hint file are its own. The
+// pane inherits that suffix (DaemonSessionManager stamps it), so the hook
+// bridge inside the pane resolves the private pipe. HOME stays real — Claude
+// Code needs its credentials and its ORDINARY hook settings, which on this
+// machine are already the plugin-less shape the report describes (Stop,
+// SubagentStop, SessionStart, and the AskUserQuestion pair; no wide
+// PostToolUse). The production daemon is never addressed.
+//
+// Usage: node scripts/autonomous-turn-status-dogfood.mjs
+import net from 'node:net';
+import fs from 'node:fs';
+import os from 'node:os';
+import path from 'node:path';
+import { spawn } from 'node:child_process';
+
+const REPO = process.cwd();
+const DAEMON = path.join(REPO, 'dist', 'daemon-bundle', 'index.js');
+const TAG = Math.random().toString(36).slice(2, 8);
+const SUFFIX = `-935dog-${TAG}`;
+const DAEMON_DIR = path.join(os.homedir(), `.wmux${SUFFIX}`);
+const PTY_ID = `dog935-${TAG}`;
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+const readPipeName = () => fs.readFileSync(path.join(DAEMON_DIR, 'daemon-pipe'), 'utf8').trim();
+const readToken = () => fs.readFileSync(path.join(DAEMON_DIR, 'daemon-auth-token'), 'utf8').trim();
+
+function rpc(method, params = {}) {
+  const id = `${Date.now()}-${Math.random().toString(36).slice(2)}`;
+  return new Promise((resolve, reject) => {
+    const s = net.createConnection(readPipeName());
+    let buf = '';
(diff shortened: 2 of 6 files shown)


openwong2kim:

This PR closes the autonomous-turn blind spot from #935: a turn the agent resumes on its own (a background task completing and the agent picking it up) produced no submitted-input boundary and no `running` hook edge on a plugin-less install, so the pane kept wearing the previous turn's `complete` for the entire turn. The fix lets byte throughput count while a pane is settled and promotes a genuine burst to `running`, held back by three well-motivated guards — a 6 s cool-down after the turn end (sized against the 10 s lifecycle suppression and the 5 s idle delay so a promoted pane can never get stranded reporting `running` after it finished), a keystroke-quiet window so the pane's echo of a draft message never reads as work, and the existing resize-redraw guard, now also stamping the unmute repaint storm via `setMuted`.

It touches `DaemonPTYBridge.ts` (the settled-promotion path and the sticky `awaitingHuman` exemption), `ActivityMonitor.ts` (a new `endTurn()` that re-arms the threshold path without firing callbacks, plus the dangling-idle-timer fix when a listener hands the cycle back mid-callback), rewrites the turn-priority tests around the new rules, and adds a dogfood script that drives a real daemon and a real agent end to end. The `awaiting_input` exemption is the part I checked hardest: only `noteInput` — including the forceSubmitted approval controls — retires a question, and it survives both a footer-derived `waiting` and a subagent `running` edge, which is the correct polarity for the one signal that must not clear itself. The known limits are documented honestly (local non-daemon mode unchanged, the general swallowed-idle case still belonging to the router) and neither blocks this. Priority P2 — it repairs the "N need you" signal, the one badge that must not lie, and the verification (unit tests plus live dogfood, full suite green) is proportionate to the risk of loosening the settle.

Issue #964, opened 2026-08-21
Workflow actions are pinned to Node 20 majors that only still run because the runner forces Node 24


openwong2kim:
The v3.45.0 release build succeeded but annotated:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: `actions/checkout`, `actions/setup-node`, `actions/upload-artifact`, `softprops/action-gh-release`.
> — https://github.com/openwong2kim/wmux/actions/runs/32436021347

Right now the runner papers over it. When it stops doing that, the workflow that publishes every wmux installer is one of the things that breaks, so this is worth clearing before it is urgent rather than during a release.

## How far behind the pins are

| action | pinned | latest |
| --- | --- | --- |
| `actions/checkout` | v4 | v7.0.1 |
| `actions/setup-node` | v4 | v7.0.0 |
| `actions/upload-artifact` | v4 | v7.0.1 |
| `actions/download-artifact` | v4 | v8.0.1 |
| `softprops/action-gh-release` | v2 | v3.0.2 |

`signpath/github-action-submit-signing-request` (v2) and `vedantmgoyal9/winget-releaser` (v2) were not named in the annotation; check them in the same pass rather than assuming.

## Sites

All four workflows, ~20 `uses:` lines: `ci.yml`, `ci-cross-platform-baseline.yml`, `perf.yml`, `release.yml`.

## A drift worth fixing in the same pass

`actions/checkout` is pinned to **two different SHAs**, both commented `# v4`:

- `11d5960a326750d5838078e36cf38b85af677262` — `ci.yml`, `ci-cross-platform-baseline.yml`, `perf.yml`
- `34e114876b0b11c390a56381ad16ebd13914f8d5` — `release.yml` (3 sites)

Nothing reads the comment, so the two have been free to diverge. One pinned SHA per action.

## Notes for whoever takes it

- Keep the SHA pins — the comment is the only human-readable part and it is what went stale. Pin the SHA, comment the exact tag.
- These are major bumps, not patch bumps. `upload-artifact` v4 already changed artifact semantics once; read each changelog rather than bumping on version number alone. `perf.yml` round-trips artifacts between jobs (upload at 128/312, download at 290) and that pairing has to keep working — the perf gate's cross-job baseline comparison depends on it.
[...]

PR #966, opened 2026-08-21
ci: prove the artifact round trip, then upgrade the Node 20 action pins (#964)

openwong2kim:
Closes the Node 20 runtime exposure on the CI workflows. `release.yml` follows in a second PR.

## Why now

GitHub's runners switched to Node 24 by default on 2026-06-16 and **remove Node 20 in fall 2026**. Until then the runner force-runs our Node 20 actions and prints a deprecation annotation; after that, the workflows that build and publish every wmux installer stop. So #964 is P2, not P3.

## The part that isn't mechanical

A green PR here does **not** prove the artifact actions still work. `bench` only uploads; the only job that downloads is `bench-confirm`, gated on `fresh-confirm-needed == 'true'` — a reproduced perf red. On a normal run it skips. Upgrading `download-artifact` across four majors and calling a green PR proof would be claiming evidence this pipeline cannot produce.

So this PR is two commits, in this order:

1. **`artifact-contract` job, pins still on v4.** Downloads into the same `first/out/...` shape `bench-confirm` uses and asserts `perf-current.json` arrived. Lands and goes green first, so the upgrade lands against a check already known to work.
2. **The pin upgrade**, which that job now actually covers.

It is a separate job rather than a force-run input on `bench-confirm`: `perfWorkflow.test.mjs:257` pins that `if:` to an exact string (#940), and loosening it would break the escalation contract the test protects.

## What still isn't covered by this PR

- **`bench-history` push.** `Publish perf history` runs only on `github.event_name == 'push'`, and checkout v6 changed credential storage ([actions/checkout#2286](https://github.com/actions/checkout/pull/2286)). This path went silently broken for six weeks once already (#602). After merge, confirm the first main Perf run appended a commit:
  `gh api repos/openwong2kim/wmux/commits?sha=bench-history --jq '.[0].commit.committer.date'`
- **`release.yml`** — tag-triggered only, so nothing here touches it. Second PR.

## Verification

- `npx vitest run scripts/__tests__/workflowSecurity.test.mjs scripts/__tests__/perfWorkflow.test.mjs` — 35 passed.
- On this PR: `artifact-contract` must **run**, not skip. If it skips, the job is wrong.
- Each job's log should lose the Node 20 deprecation annotation after commit 2. That line disappearing is the acceptance signal.

### DIFF
--- a/.github/workflows/ci-cross-platform-baseline.yml
+++ b/.github/workflows/ci-cross-platform-baseline.yml
@@ -45,10 +45,10 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
       - name: Setup Node.js 22.x
-        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: 22
           cache: npm
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,9 +11,9 @@ jobs:
   validate:
     runs-on: windows-latest
     steps:
-      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
-      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: 22
           cache: npm
--- a/.github/workflows/perf.yml
+++ b/.github/workflows/perf.yml
@@ -42,11 +42,11 @@ jobs:
     outputs:
       fresh-confirm-needed: ${{ steps.escalation.outputs.needed }}
     steps:
-      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
           fetch-depth: 1
 
-      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
           node-version: 22
           cache: npm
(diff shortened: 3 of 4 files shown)


openwong2kim:

This is the first half of the #964 fix for the Node 20 action deprecation: a new `artifact-contract` job in `perf.yml` that downloads the `bench` job's artifact on every run and asserts `first/out/perf-current.json` survived the round trip in the exact `first/out/...` layout `bench-confirm` depends on. The reasoning is sound and worth naming: `bench-confirm` is the only job that downloads today and it is gated on a reproduced perf red, so on a normal green run nothing proves the cross-job artifact path still works — that gap is exactly how a four-major `download-artifact` bump could land green while silently breaking the escalation path. Adding a separate job rather than loosening `bench-confirm`'s `if:` is the right call, since `perfWorkflow.test.mjs` pins that string exactly (#940) and the pin upgrade needs a known-good contract to land against.

One status note from reviewing the branch: the PR currently contains only commit 1 (`1f631c0c`, pins still on v4), so the second commit — the actual pin upgrades — is not in the diff yet; I'll re-review when it lands, and the acceptance signal remains the Node 20 annotation disappearing from each job's log plus `artifact-contract` running rather than skipping. Two follow-ups to hold open after merge: confirm the first main `Perf` run actually appended to `bench-history` (checkout v6 changed credential storage, and this path went silently broken for six weeks once before in #602), and the tag-gated `release.yml` upgrade arriving as the announced second PR. Given GitHub removes Node 20 in fall 2026 and the release workflows stop when it goes, this is P2, not P3.

openwong2kim:
## Verification results

**The acceptance signal fired.** Node 20 deprecation annotations, counted per job:

| job | commit 1 (v4 pins) | commit 2 (v7/v8 pins) |
| --- | --- | --- |
| bench | 1 | **0** |
| validate | — | **0** |
| artifact-contract | — | **0** |
| Baseline ×3 | — | **0** |

**The new contract job did its job.** `artifact-contract` ran — not skipped — on both commits: 4s on the v4 pins, 6s on v8. That is the upload/download round trip actually exercised on a normal green PR, which is what this pipeline could not do before.

**Bonus evidence.** Commit 1's run happened to escalate a perf red, so `bench-confirm` ran for real (2m24s, pass). The download path got covered twice over on the old pins before the upgrade landed.

## Two things the review caught that would have broken this PR

1. **`workflowSecurity.test.mjs` had `# v4` as a literal** inside the mutable-tag mutation fixture. Relabelling the pins turned its `replace` into a no-op and failed `expect(mutated).not.toBe(ci.text)` — reproduced locally before fixing. Now matched by shape with the comment behind a lookahead, so the next bump does not re-break it.
2. **`perf.yml` has eight pins, not five.** The original list missed 273/278 — the checkout and setup-node inside `bench-confirm`. Following it would have left `bench` on v7 and `bench-confirm` on v4: this issue's own bug, recreated in the same file.

## On the consistency contract

`pinConsistencyViolations` is added here but its assertion over the real workflows is not. Turning it on now would go red — correctly — because `release.yml` is still on v4 while these three moved to v7, and that split is deliberate for as long as the second PR is open. The assertion lands there, where it becomes true. The two detector tests prove the function works in the meantime.

PR #967, opened 2026-08-21
Contain a pane's z-indexes inside the pane

### LINKED ISSUE #957: Pane-internal z-indexes leak into the document stack — contain them (portal the toolbar popovers, then isolate the pane root)
Follow-up to #946, which fixed the reported class (decoration rings painting over overlays/modals) by moving the decorations below the `--z-overlay` baseline. That kills every collision between pane decorations and app chrome, but the structural hole it worked around is still there and worth naming:

**The pane root creates no stacking context, so every `z-index` inside a pane competes document-wide.** Today that leak is load-bearing: `AgentToolbar`'s popovers (RichInput is 30 rem wide) render inline inside the pane and rely on escaping the pane's box to stay visible when they overflow onto a neighboring pane. That dependency is why #946 chose lowering the decorations over `isolation: isolate` — isolating the pane root today would trap those popovers under later-DOM sibling panes.

The actual containment fix is two steps, in order:

1. **Portal the pane-internal overlays** (AgentToolbar popovers, snippets menu, rich input) to `document.body` with anchored positioning, so nothing inside a pane needs to out-z its siblings.
2. **Then give the pane root its own stacking context** (`isolation: isolate`), after which no pane-internal z value — present or future — can ever compete with app chrome again, and the z scale inside a pane becomes purely local.

Until step 2 lands, any future pane-internal element given `z-index >= 50` will reintroduce the #946 class. Nothing is urgent here — #946 closed the visible symptom — but this is the difference between "the collisions we know about are gone" and "the collision cannot be expressed".

openwong2kim:
Closes #957.

## What was open

The pane root is `relative` with no `z-index`, so it creates no stacking context and every `z-index` inside a pane competes document-wide. #946 fixed the reported class — decoration rings painting over overlays and modals — by moving the decorations below the `--z-overlay` baseline. That closed the case, not the hole: anything pane-internal at 50 or above can express the same collision again.

`isolation: isolate` on the pane root ends it. The scale inside a pane becomes purely local, and the collision stops being *expressible* rather than merely absent.

One latent case goes with it. The terminal search bar sits at `z-50` — exactly `--z-overlay` — and until now that tie reached the whole document.

## The blocker, and a correction to the issue

Step 2 was gated on nothing inside a pane needing to escape it. The issue names the agent toolbar's popovers as that dependency, and it is **no longer true**: the bar moved out to the workspace column on 2026-08-18 (DESIGN.md, and #916 landed after), and now renders inside `ToolbarHost`'s own `absolute inset-0 z-20` — a positioned element with a z-index, so already its own stacking context. Its popovers are not pane descendants and never cross a pane boundary.

What actually remained was the **terminal context menu**: `className="fixed z-[var(--z-popover-top)]"` with no portal, rendered from inside the pane. That one is fixed here.

Portalling it to `document.body` is worth doing on its own terms, independent of the isolation. The menu is viewport-positioned, so being a pane descendant bought it nothing and cost it two hazards:

- `position: fixed` resolves against a **transformed ancestor** rather than the viewport, and panes carry animated rings;
- an `overflow` ancestor added to a pane later would clip it.

## Every z value in the pane subtree, enumerated before isolating

| element | z | after isolation |
|---|---|---|
| Terminal overlays | 9–10 | local, correct |
| ScrollToBottomButton, BookmarkIndicator, ViCopyMode, PaneDecorations | 20 | local, correct |
| SearchBar | 50 | local — was tying with `--z-overlay` document-wide |
| ContextMenu | `--z-popover-top` | portalled to body |

`FloatingPane`'s `z-index: 200` is unaffected — it renders at the AppLayout level, not inside a pane.

## Tests

- `ContextMenu.portal.test.tsx` (new) pins that the menu mounts on `document.body` and not in the pane subtree that rendered it, that portalling is placement-neutral, and that it unmounts cleanly. A refactor that quietly puts it back inside the tree fails here rather than in a screenshot months later.
- `composePaneClassName` now asserts `isolate` on every ring branch, so dropping it from any one of them reopens the hole.

Full suite green (11,758 passing).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed pane layering so application controls remain visible above pane content.
  - Improved terminal context menus to appear above surrounding content without clipping or shifting.
  - Preserved menu positioning and ensured menus close cleanly.

- **Tests**
  - Added regression coverage for pane layering and context-menu behavior.
  - Added live validation for stacking and menu placement scenarios.

- **Documentation**
  - Documented pane stacking and terminal context-menu improvements.

### DIFF
--- a/changelog.d/967.md
+++ b/changelog.d/967.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **A pane's own overlays can no longer paint over app chrome.** The pane box
+  did not contain its layering, so anything inside one — a decoration badge, a
+  terminal find bar — could land on top of a dialog or an overlay elsewhere in
+  the app. Panes now stack independently, and the terminal's right-click menu
+  opens above everything instead of being clipped or misplaced by the pane
+  around it.
--- a/scripts/pane-stacking-dogfood.mjs
+++ b/scripts/pane-stacking-dogfood.mjs
@@ -1,0 +1,287 @@
+// Live check for #957 — does the pane actually contain its layering, and does
+// the terminal context menu actually open outside the pane?
+//
+// jsdom cannot answer either question: it has no layout, no stacking, and no
+// hit testing, so the unit tests can only pin the class name and the parent
+// node. This drives the packaged app over CDP and reads the computed style and
+// the real hit test.
+//
+// No OS-level clicking. Every interaction is an event dispatched inside the
+// page, so it cannot land in the wrong window.
+//
+// Usage: node scripts/pane-stacking-dogfood.mjs <cdp-port>
+const PORT = process.argv[2];
+if (!PORT) {
+  console.error('usage: node scripts/pane-stacking-dogfood.mjs <cdp-port>');
+  process.exit(2);
+}
+
+const sleep = (ms) => new Promise((r) => setTimeout(r, ms));
+
+async function rendererTarget() {
+  const res = await fetch(`http://127.0.0.1:${PORT}/json`);
+  const targets = await res.json();
+  // The app window, not a devtools page or the daemon's web view.
+  return targets.find((t) => t.type === 'page' && !t.url.startsWith('devtools://'));
+}
+
+let nextId = 1;
+function connect(wsUrl) {
+  const ws = new WebSocket(wsUrl);
+  const pending = new Map();
+  ws.addEventListener('message', (ev) => {
+    const msg = JSON.parse(ev.data);
+    const p = pending.get(msg.id);
+    if (!p) return;
+    pending.delete(msg.id);
+    if (msg.error) p.reject(new Error(JSON.stringify(msg.error)));
+    else p.resolve(msg.result);
+  });
+  const ready = new Promise((res, rej) => {
+    ws.addEventListener('open', () => res());
+    ws.addEventListener('error', (e) => rej(new Error(`ws error: ${e?.message ?? 'unknown'}`)));
+  });
+  const send = (method, params = {}) => new Promise((resolve, reject) => {
+    const id = nextId++;
+    pending.set(id, { resolve, reject });
+    ws.send(JSON.stringify({ id, method, params }));
+  });
+  return { ws, ready, send };
+}
+
+/** Evaluate in the page and return the parsed value, surfacing page throws. */
+async function evalJson(send, expression) {
+  const r = await send('Runtime.evaluate', {
+    expression: `JSON.stringify((() => { ${expression} })())`,
+    returnByValue: true,
+    awaitPromise: true,
+  });
+  if (r.exceptionDetails) {
+    throw new Error(`page threw: ${r.exceptionDetails.exception?.description ?? r.exceptionDetails.text}`);
+  }
+  return JSON.parse(r.result.value);
+}
+
+const results = [];
+const record = (name, pass, evidence) => {
+  results.push({ name, pass, evidence });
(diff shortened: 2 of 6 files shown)


openwong2kim:

This closes #957 by making the pane root own a stacking context (`isolation: isolate` in `composePaneClassName`), so every z-index inside a pane is now purely local and the #946 collision class — pane decorations painting over app chrome — becomes inexpressible rather than merely patched. The prerequisite is portalling the terminal context menu to `document.body` via `createPortal`, which the diff correctly identifies as the last pane-internal element that needed to out-z its siblings; the portal also fixes two latent hazards on its own, `position: fixed` resolving against a transformed ancestor (panes carry animated rings) and future clipping by any `overflow` ancestor. I verified the premise correction against the issue: the toolbar-popover blocker is indeed gone since the bar moved to the workspace column on 2026-08-18, so the issue's two-step ordering collapses into this one change.

The z-value enumeration (terminal overlays 9–10, chrome at 20, SearchBar tying `--z-overlay` at 50, the portalled menu at `--z-popover-top`) covers the pane subtree, and `FloatingPane`'s 200 lives outside it, so nothing is missed. Test coverage pins both halves where a regression would actually surface: `ContextMenu.portal.test.tsx` fails if the menu quietly moves back inside the pane tree, and the `isolate` assertion across every ring/blink branch catches it being dropped from any one of them. Full suite green at 11,758 passing; this is ready from my side. Priority P2, matching #957.

openwong2kim:
## Live verification against the packaged app

The unit tests here pin a class name and a parent node, which is all jsdom can answer — it has no layout, no stacking and no hit testing. So `scripts/pane-stacking-dogfood.mjs` drives the packaged build over CDP and reads the real answers. Private instance throughout: its own data dir, socket and daemon; the production instance is never addressed.

```
[PASS] pane root creates its own stacking context
       computed isolation=isolate position=relative zIndex=auto
[PASS] a pane-internal z-index cannot outrank app chrome
       order=["wmux-probe-chrome","wmux-probe-inside"]
[PASS] the menu's z-index is TRAPPED while it lives inside the pane
       z:9999 inside the pane loses to chrome z:50
[PASS] the same element on document.body is on top, as the portal makes it
[PASS] without the isolation the pane-internal element wins (the bug reproduces)
       winner #probe-in-pane; isolation restored to isolate
```

The last one is the one that makes the rest mean anything: `isolation` is dropped from the pane root in place, the ordering is re-run, and the pane-internal element wins again. A containment test that passes on the broken build proves nothing.

## Two things the live run corrected

**The context menu cannot be opened at all in this build, so the rig does not try.** Its only trigger is a right-click landing on `a[href]` (`useTerminal.ts`), and the terminal renders on WebGL, where links are painted onto `.xterm-link-layer` — measured on the running instance, the pane holds **zero DOM anchors**. That is pre-existing and out of scope here, but it is worth stating plainly on this PR: it means the portal is defensive rather than load-bearing today, and the isolation is the part actually doing work. The rig verifies the property the portal exists for instead.

**A first run stacks several modals**, onboarding and then the auto-update prompt, and they own the hit test. The first version of the probe reported failures that were entirely its own fault — the element at the pane centre was the modal, not either probe. Fixed two ways: dismiss modals in a loop, and order the probes against *each other* rather than asking who is topmost overall, so residual chrome above both cannot answer a different question.

No OS-level clicking anywhere — every interaction is an event dispatched inside the page.

openwong2kim:
@coderabbitai review

openwong2kim:
@coderabbitai full review

PR #968, opened 2026-08-21
fix(hooks): remember "Don't ask again" for the hook-install prompt

snowyukitty:
## Problem

The hook-install modal had no durable refusal. Dismissing it only set a
process-local React phase, so a user who had decided against installing hooks
was asked again on every launch — and again on every raise of agent mode from
`off`. There was no way to say no and have it stick.

## Behaviour

- **New third button, "Don't ask again."** Persists in `hooks-prompt.json`
  under the wmux data dir (atomic write, `WMUX_DATA_SUFFIX` isolated, the same
  storage shape as `deck-autonomy.json`). Survives restart and upgrade, and
  suppresses both triggers.
- **"Later" is unchanged**: it hides that showing only, and a later raise of
  agent mode from `off` still prompts.
- **Reversible from Settings → integration setup**: an "Ask again" control that
  appears only when the refusal is actually stored. The preference is re-read on
  every trigger, so clearing it is picked up with no broadcast and no restart.

Fail-open by design: any uncertain read of the store — missing file, torn JSON,
a non-object, anything that is not a literal `true` — resolves to "ask". A
corrupt file resurrects the prompt rather than silently muting onboarding, which
would be invisible. Only a literal boolean writes, so no caller bug can create
or clear a refusal. On a preload too old to persist, the durable button is not
rendered at all rather than silently acting as "Later".

## Review map

- **Production**: ~280 lines across 8 files — the store
  (`src/main/hooks/hooksPromptPreference.ts`), two IPC channels, two preload
  bindings, the modal's buttons and trigger logic, the Settings card row, and
  four locale keys (en/ko/zh; `ja` has no `hooks.prompt.*` keys today and was
  left alone).
- **Tests**: ~510 lines across 3 files, 53 cases. Six properties are pinned by a
  control run — reverting that one fix makes that one test fail:
  Settings reversibility in the same session; an in-flight status probe cannot
  reopen the modal over an answer; an in-flight preference read cannot poison
  the fallback cache; the post-install Close is a dismissal like the others;
  out-of-order reads cannot overwrite a newer one; and the Settings card's
  generation guard.
- **Already run**: 53/53 in the four test files, `tsc --noEmit` clean, and
  eslint on the touched files adds no new error (the one `no-empty-function` hit
  already exists on `main` at a different line number).

## Summary by CodeRabbit

* **New Features**
  * Added a “Don’t ask again” option for the hook-installation prompt.
  * The preference persists across app restarts and upgrades.
  * Added an “Ask again” option in integration settings to re-enable the prompt.
  * “Later” and dismissal remain temporary for the current session.
  * Added localized messaging for English, Korean, and Chinese.

* **Bug Fixes**
  * Improved handling of preference read/write failures and invalid stored settings.


openwong2kim (commented):
This adds a durable "Don't ask again" to the hook-install prompt: a new `hooks-prompt.json` store in the wmux data dir (atomic write, `WMUX_DATA_SUFFIX`-isolated, same shape as `deck-autonomy.json`), two IPC channels with preload bindings, a third button on the modal, an "Ask again" row in Settings → integration setup, and four locale keys across en/ko/zh. The design choices are the right ones: fail-open on every uncertain read so a corrupt file resurrects the prompt rather than silently muting onboarding, only a literal boolean ever writes, non-boolean payloads report the stored value instead of a default, and a preload too old to persist simply doesn't render the durable button rather than acting as "Later". I read through the diff — the store, the IPC boundary tests, and the prompt component all pin the invisible-failure cases (torn JSON, out-of-order reads, status probe reopening over an answer), and CI is green across all four checks including the cross-platform baselines. Triage priority P3: real onboarding annoyance (asked on every launch and every agent-mode raise with no way to refuse durably) but no functional breakage, and the fix is self-contained with no production behavior beyond the prompt itself. Nice change — the two-lifetime split between "Later" and "Don't ask again" is documented exactly where the next reader will trip over it.

snowyukitty:
Addressed all three review comments in `ba9f5004`. Taking the Major one on its
merits rather than as written, because the mechanism it describes is real but
the remedy would make the feature worse.

**Backup recovery stays.** The finding is correct that `atomicReadJSONSync`
walks `BACKUP_SUFFIXES` when the primary yields nothing, so a torn primary with
a valid `.bak` returns the backup's value rather than "ask". But that backup is
the last value the operator actually chose. Disabling recovery here would
discard a real refusal because of an unrelated write tear — which is the exact
regression a backup exists to prevent — and would re-nag someone who had already
said no.

What was actually wrong was my header paragraph, which said a torn file resolves
to "ask" without that qualification, and the fact that nothing pinned the
distinction: the existing "unparsable → ask" case passes only because no backup
exists in that fixture. So the header now states the real invariant — doubt
means *no trustworthy answer survives* — and two tests cover both halves:

- torn primary + valid `.bak` → the refusal is recovered
- torn primary + torn `.bak` → asks

The other two are applied as suggested, with one deliberate deviation: the error
paragraph got `role="alert"` rather than `role="status"`, since it only ever
carries a failure the user just caused, and a polite live region can be held
until the user is idle. `npx eslint` is now clean — zero errors — on every file
this branch touches.

snowyukitty:
`Baseline (windows-latest)` is red on `ba9f5004`, and I do not believe it is this
branch. Posting the evidence rather than pushing a no-op commit to re-roll it —
I do not have rerun rights here.

**The failure is a timeout, not an assertion.**

```
FAIL src/main/git/__tests__/mergeSession.test.ts
  > detectConflicts — conflict detection (not exit code)
  > a conflicting merge returns the unmerged file list, a clean merge returns an empty list
Error: Test timed out in 5000ms.
```

That file shells out to real `git`, and on that runner its cases were sitting
right on the 5 s default. In the *same* run:

| case | time | result |
|---|---|---|
| `a conflicting merge returns the unmerged file list…` | 7139 ms | ✗ timeout |
| `fast-forwards base to the result after merging in the isolated worktree` | 7168 ms | ✓ |
| `rejects when an in-progress merge (MERGE_HEAD) is present` | 6118 ms | ✓ |

A sibling took *longer* than the one that failed and passed. 11794/11821 green
otherwise.

**Why it is not the change:**

- `Baseline (windows-latest)` passed on this same branch one commit earlier
  (8m34s, run 32446751187). The only delta since is
  `src/main/hooks/hooksPromptPreference.ts`, its test, and
  `HooksInstallPrompt.tsx` — none of which is reachable from `mergeSession`.
- `npx vitest run src/main/git/__tests__/mergeSession.test.ts` is 18/18 locally
  on Windows 11.
- `Baseline` on `macos-14` and `ubuntu-22.04` are both green on this commit, as
  are `validate`, `bench` and `artifact-contract`.

A re-run should clear it. If it reproduces, the durable fix is probably an
explicit `testTimeout` on that file rather than anything on this branch — the
whole file is 2–7 s per case on a loaded Windows runner.

openwong2kim:
Re-ran the failed job — **green**. Your read was right: timeout, not assertion, and nothing to do with this branch. Thanks for posting the evidence instead of a no-op commit; the sibling case taking *longer* and passing was the detail that settled it. Your suggestion of an explicit `testTimeout` on that file looks right to me too, but it belongs in its own PR rather than riding this one.

Then ran a three-way review (Claude / Codex / GLM-5.3, independently on the same diff) plus a live dogfood, since the durable-storage claim is exactly what jsdom cannot answer.

**Dogfood, 9/9** — suffix-isolated instance, real main process, real IPC:

```
preload exposes both preference calls
with no file on disk, the answer is "ask"        {suppressed:false}, file=null
setPromptPreference(true) writes to a real file  file={"suppressed":true}
reads back through a separate IPC call
non-boolean input neither writes nor changes it  "false" / 0 / null all inert
clearing writes suppressed:false
──── full process restart ────
the refusal survived on disk
the fresh process reads it back through IPC
```

The modal never mounted, which is correct — that instance already has hooks installed, so the prompt self-suppresses. Drove the bridge directly instead.

No P1 from any of the three. One thing I'd like your read on before merging.

## The shared `genRef` in IntegrationSetupSection

Codex flagged this as a merge-blocker, GLM as minor, and I confirmed the mechanism in the code — so I'm fairly confident it's real, less sure how much it matters.

`genRef` is a single counter (line 137), bumped at 150, 202, 231, and now 252. The bump in `reenablePrompt` does what your comment says it does — stops an in-flight preference probe from repainting the stale `true` after a clear. What the comment doesn't cover is that the *same* bump also invalidates the `commit(...)` callbacks for the hooks / statusline / mcp rows, since they carry the generation they were issued under.

The window: the pref probe resolves first (one small file read), the suppressed line renders, the user clicks "Ask again" while a slower row probe is still in flight. That row's commit is then dropped — and unlike the other bump sites, this path never calls `probeAll()` again, so nothing re-issues it. Codex's addition was that a concurrent failed install could sit in `working` the same way.

Not something the dogfood would catch, and closing and reopening the Settings tab clears it.

So: does a separate counter for the preference seem right to you, or do you read the coupling differently? You've been closer to this file than I have, and your last round showed you'd rather take a finding on its merits than as written — same invitation here. If you'd rather ship as-is and track it, that's a fine answer too; I'd just want it written down rather than assumed benign.

The rest was minor and I'm not asking for any of it: the store's sync read on the IPC path (same shape as `deckAutonomyStore`, so not something this PR introduced), a silent `.catch` on `reenablePrompt` where the prompt side surfaces its save failures, and a narrow case where a failed read right after a Settings clear stands on the cached `true` — the mirror image of the tradeoff your comment already names.

openwong2kim:
Separate counter, please — and I think it is this PR's to fix rather than something to track, for a reason that only shows up in the diff.

`reenablePrompt` and its `++genRef.current` are both **new here**:

```
+  const reenablePrompt = useCallback(() => {
+    const gen = ++genRef.current;
```

So this is not a pre-existing coupling the PR happens to sit next to. The three old bump sites all belong to the row lifecycle — probe, install, re-register — and every one of them re-issues the writes it cancelled. The fourth does not, and it is the one this PR adds. Introducing a bump site that drops other people's callbacks without re-issuing them is the whole defect.

**The coupling is not load-bearing in the direction that matters.** The one job the shared counter does for the preference is the one your comment names: `reenablePrompt` has to invalidate the pref probe `probeAll` may still have in flight, or a slow read repaints the stale `true` and puts the line back after the user cleared it. A `prefGenRef` read and bumped by `commitPref` and `reenablePrompt` only keeps exactly that. Nothing wants the cross-domain edges: `probeAll` issues a pref probe and three row probes together, so it can bump both counters at once, and a successful install re-probes anyway.

**Codex's `working` variant is what settles it, and it is worse than the dropped probe.** Follow `runInstall` through a failure:

```
gen = ++genRef.current          → N
commit(set, {state:'working'})  → paints the spinner at N
await install()                 → user clicks "Ask again" → genRef = N+1
commit(set, {state:'error'})    → gen N ≠ N+1, dropped
```

Note which half survives. On success `runInstall` calls `probeAll()`, which bumps again and re-issues everything, so a *successful* install self-heals through the same race. Only the failure path strands — the row keeps spinning, and the error text naming what went wrong is the thing that gets thrown away. The stuck state lands exactly on the case where the user most needs to be told something failed.

That is a different severity from a dropped probe. A dropped probe leaves a row showing its previous state, which is stale but not a claim. A dropped error commit leaves a row asserting that work is still in progress when nothing is running.

You are right that a tab close and reopen clears it, and right that the dogfood would not have caught it. Neither changes my read: a spinner that never resolves is not something I want to ship and write down.

Not asking for the other three. The store's sync read predates this PR, the silent `.catch` on `reenablePrompt` is the deliberate "leave the line up" choice and its comment says so, and the cached-`true`-after-a-failed-read case is the same tradeoff already named in the code.

Push the counter split and I will merge on green. Happy to push it to your branch myself if that is easier.

snowyukitty:
Taking you up on that offer — please push the split to the branch. The exact diff I verified is below, with what it was run against and what was not run.

I reproduced your window as a control pair, same probe result in both runs:

```
control (no click)     hooks row → installed
treated ("Ask again")  hooks row → unknown
```

`hooks.status()` resolved `installed: true` both times, so the bump at 252 dropped the commit and nothing re-issued it. Your read of the mechanism holds as written.

Reading the source rather than running it: `commitPref` is created inside `probeAll` and closes over the same `gen`, so the one counter guards the preference probe as well as the row commits. I ran that in one direction only — a slow `getPromptPreference` plus a **statusline** install returning `ok: false`:

```
preference read still in flight
  → Install clicked on the statusline row
  → that install returns ok: false
  → the suppressed line never renders, though the probe resolved suppressed: true
```

So the coupling runs both ways, not only from "Ask again" toward the rows.

I also ran the `working` trace you wrote out: Install clicked on the hooks row, "Ask again" while that install is in flight, the install then returning `ok: false`. On the current head the row stays at `working`; with the split below, the `error` commit lands.

A separate counter closed all three of the reproductions I ran:

```diff
diff --git a/src/renderer/components/Settings/IntegrationSetupSection.tsx b/src/renderer/components/Settings/IntegrationSetupSection.tsx
index 3db2e6ab..0285967a 100644
--- a/src/renderer/components/Settings/IntegrationSetupSection.tsx
+++ b/src/renderer/components/Settings/IntegrationSetupSection.tsx
@@ -135,6 +135,7 @@ export function IntegrationSetupSection({
   // repainting "not installed" over a finished install — which would put the
   // Install button back and invite a second write to the same config.
   const genRef = useRef(0);
+  const prefGenRef = useRef(0);
   const commit = useCallback(
     (set: (m: RowModel) => void, model: RowModel, gen: number) => {
       if (!mountedRef.current || gen !== genRef.current) return;
@@ -167,8 +168,9 @@ export function IntegrationSetupSection({
     // Generation-guarded like the rows: "Ask again" bumps the generation, so a
     // probe still in flight when it lands cannot repaint the stale `true` and
     // bring the line back after the user cleared it.
+    const prefGen = ++prefGenRef.current;
     const commitPref = (v: boolean | null) => {
-      if (!mountedRef.current || gen !== genRef.current) return;
+      if (!mountedRef.current || prefGen !== prefGenRef.current) return;
       setPromptSuppressed(v);
     };
     if (!api.hooks?.getPromptPreference) commitPref(null);
@@ -249,11 +251,11 @@ export function IntegrationSetupSection({
   // user would otherwise click again.
   const reenablePrompt = useCallback(() => {
     if (!api.hooks?.setPromptPreference) return;
-    const gen = ++genRef.current;
+    const prefGen = ++prefGenRef.current;
     void api.hooks
       .setPromptPreference(false)
       .then((pref) => {
-        if (!mountedRef.current || gen !== genRef.current) return;
+        if (!mountedRef.current || prefGen !== prefGenRef.current) return;
         setPromptSuppressed(pref.suppressed);
       })
       .catch(() => { /* leave the line up: the refusal is still in force */ });
```

All three probes pass with it and your 16 existing cases stay green.

What I did not run, and so cannot tell you: whether `registerMcp` behaves like `runInstall` here — I only read it; whether a successful install re-issues the dropped probe; anything about what reaches disk; and whether reopening the tab recovers either state.

openwong2kim:
Pushed — `818469f7`, your diff verbatim plus two things around it.

**A regression test for the strand.** Your third reproduction (install in flight, "Ask again" mid-flight, install rejects) is now a case in `IntegrationSetupSection.test.tsx`: the row must land on `error` with its message, not sit on `working`. It fails on `ba9f5004` and passes with the split, so the reason this counter exists is locked, not just remembered. 17/17 green locally; tsc clean; the three new non-null assertions match the file's existing idiom (warnings, not errors).

**Comments moved with the code.** The `prefGenRef` declaration now says why it is separate — the sentence is essentially your "the coupling runs both ways" finding — and the `commitPref` comment says "on its own counter" instead of "like the rows", which stopped being true the moment your diff applied.

On the two things you read but did not run:

- **`registerMcp` is covered by construction.** The strand needed `reenablePrompt` to bump `genRef`; after the split it never touches that counter, so there is no path left from the preference side into any row commit — `runInstall`'s and `registerMcp`'s alike. Nothing to verify per-callsite, which is the nice property of fixing it at the counter rather than at the sites.
- **A successful install re-issuing the dropped probe** — also moot now for the same reason, but for the record your reading was right: `probeAll()` on the success path bumps and re-issues everything, which is why only the failure path ever stranded.

Merging on green.

PR #971, opened 2026-08-21
ci(release): give release.yml a rehearsal path, then move it off Node 20 pins (#964)

openwong2kim:
Second half of #964. Follows #966, which did the CI workflows.

| action | from | to |
| --- | --- | --- |
| `actions/checkout` | v4 | v7.0.1 |
| `actions/setup-node` | v4 | v7.0.0 |
| `actions/upload-artifact` | v4 | v7.0.1 |
| `softprops/action-gh-release` | v2 | v3.0.2 |

## The rehearsal path is the point

Until now the only way to run *any* of `release.yml` was to push a `v*` tag — which creates a real public release, pushes to Chocolatey and WinGet, and burns a version number that cannot be reused if the run fails. Every change to this file had production as its first test, and that would have stayed true for every future dependabot bump of the actions it pins.

`workflow_dispatch` with a `dry_run` input builds, packages, checksums and uploads artifacts exactly as a real run does; the four publishing steps and the two release-asset uploads stand down.

**Run it before merging:** Actions → Build & Release → Run workflow → branch `chore/964-release-pins`, `dry_run` checked.

## What that covers, and what it doesn't

| pin | covered by dry_run? |
| --- | --- |
| checkout ×3 (build, macos, linux) | ✅ |
| setup-node ×3 | ✅ |
| upload-artifact `:207` | ✅ |
| upload-artifact `:63` | ❌ behind `SIGNPATH_ENABLED`, which is false without the token — **unverifiable either way** |
| **action-gh-release `:214`** | ❌ only runs when something is actually published — **first exercise is a real tag** |

Two of nine pins stay unproven at merge. That is two fewer than "all of them", which is where this file was before.

## Also in here

- **Turns on the pin consistency contract.** #966 added `pinConsistencyViolations` but not its assertion over the real workflows — the CI half had moved to v7 while this file was still on v4, so enforcing it then would have failed a green tree for a split we chose deliberately. This PR closes the split, so the assertion lands with it.
- **`.github/dependabot.yml`**, grouped by blast radius: `actions/*` (fully exercised by the PR that carries it) separate from `softprops/*` + `vedantmgoyal9/*` (the publishing path, which needs a dry_run first). `signpath` is in `ignore` — those steps are no-ops without the token, so nothing verifies them, and an update nobody can check should not arrive automatically on the path that signs the binaries users install.
- **Explicit npm cache** on the three release setup-node steps. All four CI ones already had it; these ran a cold `npm ci` every tag. Leaving it implicit also meant a future `packageManager` field would silently change what the privileged job does (setup-node v5+ auto-caches on it).

## Checked rather than assumed

`upload-artifact@v7.0.1` still exports the `artifact-id` output that the SignPath step consumes from it (verified in its `action.yml` at that tag), so bumping the producer while the consumer stays pinned is safe.

## #964 does not close with this merge

`release.yml` is still unproven for the two pins above until a real tag runs. The issue closes after that release is audited — checking each job's actual conclusion, not the asset count, since `winget`/`macos`/`linux`/symbols carry `continue-on-error` in six places and a partial publish reports success.

## Summary by CodeRabbit

- **New Features**
  - Added a manual release dry-run option to build, package, verify, and upload artifacts without publishing releases or platform packages.
  - Enabled improved dependency update management for release automation.

- **Bug Fixes**
  - Improved consistency and reliability of release workflow action versions.

- **Documentation**
  - Added release-process follow-up items covering failure reporting, validation, verification, and ownership.

- **Tests**
  - Added automated checks to ensure consistent workflow action versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,40 @@
+# Who owns the action pins from here on.
+#
+# Every pin in .github/workflows is a full commit SHA (#642), which is what
+# makes third-party action code immutable — and also what makes it invisible
+# when it ages. actions/checkout sat four majors behind, on two different
+# commits both labelled `# v4`, until a release build's deprecation warning
+# happened to be read by a human (#964). That is not a process; that is luck.
+#
+# Grouped by blast radius, not by convenience. A single catch-all group would
+# put a bump to the action that publishes every wmux installer in the same pull
+# request as a checkout patch bump — and the whole point of splitting #964 into
+# two PRs was that those two things earn very different amounts of scrutiny.
+version: 2
+updates:
+  - package-ecosystem: github-actions
+    directory: "/"
+    schedule:
+      interval: monthly
+    open-pull-requests-limit: 3
+    groups:
+      # Everyday CI plumbing. Fully exercised by the pull request that carries
+      # it: checkout and setup-node run in every job, upload/download-artifact
+      # are covered by perf.yml's artifact-contract job.
+      ci-actions:
+        patterns:
+          - "actions/*"
+      # The publishing path. release.yml only runs on a `v*` tag, so a bump
+      # here cannot be proven by its own PR. Run the workflow's dry_run first
+      # (Actions → Build & Release → Run workflow), then merge.
+      release-actions:
+        patterns:
+          - "softprops/*"
+          - "vedantmgoyal9/*"
+    ignore:
+      # Code signing. The SignPath steps are no-ops until SIGNPATH_API_TOKEN is
+      # configured, so nothing — not CI, not dry_run, not a real release —
+      # exercises this action today. An update nobody can verify should not
+      # arrive automatically on the path that signs the binaries users install.
+      # Revisit when signing is enabled and there is a test-signing canary.
+      - dependency-name: "signpath/github-action-submit-signing-request"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,7 +25,7 @@ updates:
         patterns:
           - "actions/*"
       # The publishing path. release.yml only runs on a `v*` tag, so a bump
-      # here cannot be proven by its own PR. Run the workflow's dry_run first
+      # here cannot be proven by its own PR. Dispatch the workflow manually first
       # (Actions → Build & Release → Run workflow), then merge.
       release-actions:
         patterns:
@@ -33,7 +33,7 @@ updates:
           - "vedantmgoyal9/*"
     ignore:
       # Code signing. The SignPath steps are no-ops until SIGNPATH_API_TOKEN is
-      # configured, so nothing — not CI, not dry_run, not a real release —
+      # configured, so nothing — not CI, not a rehearsal, not a real release —
       # exercises this action today. An update nobody can verify should not
       # arrive automatically on the path that signs the binaries users install.
       # Revisit when signing is enabled and there is a test-signing canary.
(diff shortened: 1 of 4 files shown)


openwong2kim:

This PR is the release.yml half of #964: it moves the last workflow off the deprecated Node 20 action majors (checkout v4→v7.0.1, setup-node v4→v7.0.0, upload-artifact v4→v7.0.1, action-gh-release v2→v3.0.2), collapses the two divergent `# v4` checkout SHAs #964 flagged into a single pin, and turns on the `pinConsistencyViolations` assertion that #966 deliberately left dormant while the two halves were legitimately split. The core addition is the `workflow_dispatch` dry_run input on release.yml — until now every change to this file had a real public release as its first test, and the dry_run path (build, package, checksum, artifact upload; publishing steps stand down behind `inputs.dry_run != true` guards) finally gives the release pipeline a rehearsal stage that does not burn a version number. The dependabot grouping is well-reasoned: `ci-actions` separated from `release-actions` by blast radius, and signpath ignored because nothing can verify an update to a step that no-ops without the token.

Mergeable, with validate and all three baseline jobs green. Per the PR's own accounting, two of nine pins (the SignPath-conditioned upload and action-gh-release itself) stay unproven until a real tag runs, so #964 correctly remains open until that release is audited. One ask before merge: run the dry_run on this branch exactly as documented in the PR body, so the rehearsal path itself is proven and not just the code behind it. Priority: P2 — this is the publishing path for every wmux installer, and the runner's Node 24 force-run that currently papers over these pins is time-limited.

openwong2kim:
## The rehearsal path works, and it earned its keep on the first run

**Run 1** ([32447862262](https://github.com/openwong2kim/wmux/actions/runs/32447862262)) failed — in exactly one place, and it was a real bug in this PR. The macOS job has *two* upload steps, and only one had been gated. `Write and upload the darwin update manifest` is deliberately the one upload with no `continue-on-error` (a missing manifest silently cuts darwin auto-updates until the next release, so it has to fail loudly), and it duly failed the whole job trying to attach an asset to a release that a dispatch never created.

That is the rehearsal path finding a defect on its first use, in a file whose previous first test was always a public release.

**Run 2** ([32448679382](https://github.com/openwong2kim/wmux/actions/runs/32448679382)) — all four jobs green:

| job | result |
| --- | --- |
| build (windows-latest) | ✅ success |
| Build macOS (macos-14) | ✅ success |
| Build Linux | ✅ success |
| Publish to WinGet | ✅ success |

And it published nothing. All six publishing steps skipped:

```
[build]        Create GitHub Release
[build]        Push to Chocolatey
[Build Linux]  Upload Linux artifacts to the release
[Build macOS]  Upload macOS artifacts to the release
[Build macOS]  Write and upload the darwin update manifest
[WinGet]       Publish to winget
```

`gh release list` still shows v3.45.0 as latest; zero releases exist with a non-`v` tag.

## The gate changed shape after review

`dry_run` as an input was a publish button on a branch — untick it, dispatch from anywhere, and `github.ref_name` is a branch name that Create GitHub Release builds against while Chocolatey and WinGet receive it. Publishing is now gated on `github.event_name == 'push'` and the input is gone, rather than kept as decoration. Dispatch is a rehearsal, always; the only way to spell "publish" involves a tag.

## Verified in this PR

- checkout v7.0.1, setup-node v7.0.0, upload-artifact v7.0.1 — **actually executed** in all three release jobs
- `pinConsistencyViolations` over the real workflows — green, so every action now resolves to one commit with one label repo-wide
- `.github/dependabot.yml` — GitHub's own validator passes it (the `.github/dependabot.yml` check, 1s)
- 38 tests in the two workflow contract files

## Still unverified at merge, by nature

- `action-gh-release@v3.0.2` — only runs when something is published; first exercise is the next real tag
- `upload-artifact` at `:63` — behind `SIGNPATH_ENABLED`, which is false without the token

Two of nine, down from nine of nine.

PR #979, opened 2026-08-21
feat(hooks): let setup-hooks install the lifecycle signals without the permission gate (#970)

### LINKED ISSUE #970: Should lifecycle signals be installable without the wide PreToolUse permission gate?
## The question

Because the permission gate only arms under `wmux web --allow-input` and stays
silently dormant otherwise, every hook install today pays the gate's
per-tool-call cost for a feature that is inert by default — should wmux support
installing the lifecycle signals **without** the wide PreToolUse gate, or is the
gate meant to ship bundled and dark until the operator opts in?

I am not proposing a patch here. The answer decides the shape entirely, and
three plausible shapes follow from three different answers, so it is worth
settling before anyone writes one.

## Why it comes up

`HOOK_SPECS` (`src/cli/commands/setupHooks.ts:88`) ships `PERMISSION_GATE_SPEC`
— a match-all PreToolUse invoking the bridge with `--permission-gate` — in the
same install as Stop / SubagentStop / SessionStart and the AskUserQuestion pair.
`integrations/claude/hooks/hooks.json` does the same for plugin users. There is
no flag, and no config, that yields lifecycle signals without that hook
existing.

Today that hook is harmless at runtime. Since #899 every no-opinion path writes
empty stdout, and `daemon/index.ts` only reaches the broker when
`webTerminalServer.canResolveGates` — `server !== null && opts.allowInput ===
true` (`WebTerminalServer.ts:984-986`). A bare `wmux web` is read-only, so the
gate stays dark. So this is **not** a report that anything is broken.

What it costs is a process. The hook is spawned on every PreToolUse regardless.
Measured on Windows 11 against an installed `~/.wmux/hooks/wmux-bridge.mjs`,
3–5 runs each:

| path | wall time |
|---|---|
| `WMUX_GATE=0` fast exit | 113–127 ms |
| armed, daemon unreachable | 128–149 ms (stdout: 0 bytes) |
| outside a wmux pane | 118–161 ms |

No daemon was running, so the live RPC leg is excluded and these are a lower
bound. The 0-byte stdout is incidental confirmation that #899's fix holds on an
installed copy.

## What already exists, so it is not confused for the same thing

`gatedTools: []` is already a durable policy off, and `gateConfig.ts:42` says
so. That is a persistent answer to *which tools are gated*. It is not an answer
to *whether the permission hook is installed at all* — and the latter is the one
a terminal-only operator, or anyone who wants "wmux cannot participate in
permission decisions" to be true by construction rather than by the continued
correctness of the bridge, would be asking for.

`setup-hooks --status` currently reports a missing gate as something to fix
(`setupHooks.ts:818-822`), which encodes an answer to this question already —
"the gate is required" — as copy rather than as a decision. If the answer here
is "supported", that string wants revisiting too.

## The three shapes

- **Supported** → an install profile split: `--signals-only`, plus a second
  hooks profile (or a self-disabling gate) for the marketplace plugin, since a
  CLI-only flag would do nothing for plugin users.
- **Bundled and dormant is the intent** → no code change; the `--status` detail
  string becomes the fix, so a missing gate reads as a configuration and not a
  defect.
- **Something in between** → a persisted opt-in the gate reads before it
  round-trips, so a dormant gate costs a stat rather than a pipe call.

Happy to implement whichever, or none.

## One thing I checked and got wrong first, in case it saves you the trip

I initially read the plugin's match-all `PostToolUse` as an oversight against
`HOOK_SPECS`' `AskUserQuestion` scoping. It is not: `0c98b739` (#435) removed it
from both paths and `cbc19375` (#462) deliberately restored it in `hooks.json`
for authority freshness. That is a separate question from this one — since #789
put a wide gate hook on both paths and `HookIngest.handle()` touches authority
on every resolved signal, #462's justification may have been overtaken — and I
will raise it separately rather than muddy this issue with it.

openwong2kim:
Closes #970 — the design question, answered as **Supported**: `wmux setup-hooks --signals-only` installs the lifecycle signals and the approval card without the wide `PreToolUse` permission gate.

## Why this shape and not the other two

**The "make the dormant gate cheaper" shape is dead.** The reporter measured 113–161 ms per tool call and correctly called it a lower bound. Reproduced here, plus the piece that settles it:

| path | measured |
|---|---|
| `node -e ""` — bare spawn floor | 81–91 ms |
| `WMUX_GATE=0` fast exit | 121–166 ms |
| armed, outside a wmux pane | 121–168 ms |

~85 ms of the ~120 is **process spawn**, and `WMUX_GATE=0` already returns before reading stdin or touching a pipe. A persisted opt-in the gate reads before it round-trips would save ~20 ms of ~120. The cost is the hook existing, so the only way to not pay it is to not install it.

**The wide hook is not inert, and that argues for the split rather than against it.** `daemon/index.ts:3030-3032` does not return quietly on the dormant path — it emits `agent.tool_started`, which `agentLiveness.ts:121` projects into the phone header's "which tool is running". So the hook does two jobs. But `emitAgentLiveness` fans out to web clients only, and gate resolution arms only under `wmux web --allow-input` (`WebTerminalServer.canResolveGates`). **Both jobs need a web surface.** A terminal-only operator pays the spawn for neither.

**There is precedent.** `setupHooks.ts:52-58` records that #435 removed a wide `PostToolUse` for exactly this reason — "a ~110ms node bridge per tool call feeding the fleet running dot, which the daemon's byte-based ActivityMonitor now drives for free". #783 reintroduced the same per-tool-call cost for the gate. And `gatedTools: []` is not the same answer: a policy of "gate nothing" still spawns the process that asks.

## What changed

- **`--signals-only` / `--with-gate`** on `wmux setup-hooks`. Signals-only writes `Stop` / `SubagentStop` / `SessionStart` plus the `AskUserQuestion` approval pair, and nothing that fires per tool call.
- **The profile is derived from `settings.json`, not stored beside it.** A marker file could go stale against the hooks it describes. More importantly, a bare `wmux setup-hooks` — which is what the app-update refresh and the in-app "install hooks" button call — must never resurrect a gate the operator removed. It now keeps whatever profile is installed. A *partial* install (a signal missing) still reads as broken-full and gets repaired.
- **The silent half is guarded.** Signals-only + `wmux web --allow-input` leaves a phone waiting on approvals nothing will ever raise: no error, no log, no timeout. `--allow-input` now checks the hook and says so on startup.
- **`--status` copy.** An absent gate is now two states. On the signals-only profile it reads as a configuration with the way back, not as something to fix. The old string encoded "the gate is required" as copy — an answer to this issue's question, written before the decision was made.

## Scoped out, deliberately

Plugin users keep the full profile. A Claude Code plugin ships one `hooks/hooks.json` with no per-install switch, and the reporter's alternative — a self-disabling gate — would still pay the spawn that is the entire cost. So `--signals-only` with the plugin active **reports that it had no effect** instead of silently doing nothing, and `integrations/claude/README.md` documents the CLI path as the way to get it. A second marketplace plugin is a distribution decision, not a patch.

## Verification

- 20 new unit tests for the profile (`setupHooks.test.ts`) + 4 for the `--allow-input` warning (`web.gateWarning.test.ts`); 372 pass across `src/cli/`, `integrations/claude/`, and the in-app hooks bridge.
- End-to-end against a built CLI in an isolated HOME: signals-only install writes zero `--permission-gate` entries; a bare re-run keeps it at zero; `--with-gate` restores exactly one; a second bare re-run keeps it at one; all three flag-conflict paths error out.
- `tsc --noEmit` clean (6 pre-existing baseline errors, none in touched files).

## Summary by CodeRabbit

- **New Features**
  - Added hook setup profiles: full installation or signals-only mode.
  - Added `--signals-only` and `--with-gate` options for managing permission hooks.
  - Hook refreshes now preserve the selected profile.
  - Web input now warns when the permission gate is missing.

- **Documentation**
  - Expanded setup guidance for profiles, status checks, plugin behavior, and removal.

- **Bug Fixes**
  - Status reporting now correctly handles intentional signals-only installations.

### DIFF
--- a/changelog.d/970.md
+++ b/changelog.d/970.md
@@ -1,0 +1,24 @@
+### Added
+
+- **`wmux setup-hooks --signals-only` installs the lifecycle signals without
+  the per-tool-call permission gate.** The remote-approval gate is a wide
+  `PreToolUse` hook, so Claude Code spawns the bridge on every single tool call
+  — ~120 ms on Windows 11, of which ~85 ms is bare `node` startup. That cost
+  cannot be optimised away once the hook is registered, and `gatedTools: []`
+  does not help: a policy of "gate nothing" still spawns the process that asks.
+  It buys nothing for a terminal-only operator either, because both things the
+  wide hook does need a web surface — approvals arm only under
+  `wmux web --allow-input`, and the tool-name liveness it feeds is fanned out
+  only to web clients. `--signals-only` installs the turn-boundary signals and
+  the approval-card pair and stops there; `--with-gate` puts the gate back. The
+  profile is derived from `settings.json` rather than stored beside it, so a
+  bare `wmux setup-hooks` — the app-update refresh and the in-app install
+  button included — keeps whatever is already installed instead of quietly
+  re-adding the gate.
+
+- **`wmux web --allow-input` now says when the permission-gate hook is
+  missing.** Enabling input is what arms the gate, so it is the only place that
+  can catch a signals-only install: without the hook no tool call ever raises
+  an approval, and nothing else reports it — the phone simply never rings.
+  `wmux setup-hooks --status` also stops calling an absent gate a defect when
+  the signals-only profile is the reason it is absent.
--- a/integrations/claude/README.md
+++ b/integrations/claude/README.md
@@ -48,9 +48,11 @@ install the same 4 hooks directly into your Claude Code user settings
 (`~/.claude/settings.json`):
 
 ```
-wmux setup-hooks            # install
-wmux setup-hooks --status   # show install + bridge freshness
-wmux setup-hooks --remove   # uninstall (leaves your other hooks intact)
+wmux setup-hooks                 # install (keeps the profile already on disk)
+wmux setup-hooks --signals-only  # install WITHOUT the per-tool-call gate hook
+wmux setup-hooks --with-gate     # put the gate back
+wmux setup-hooks --status        # show profile, install + bridge freshness
+wmux setup-hooks --remove        # uninstall (leaves your other hooks intact)
 ```
 
 This copies the bridge to a stable path (`~/.wmux/hooks/wmux-bridge.mjs`)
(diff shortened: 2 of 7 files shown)


openwong2kim:

This PR implements `wmux setup-hooks --signals-only`, answering #970 as "Supported" and splitting the hook install into two profiles: the turn-boundary signals (`Stop` / `SubagentStop` / `SessionStart`) plus the `AskUserQuestion` approval pair, with or without the wide `PreToolUse` permission gate. The rationale is well-grounded in the diff: the gate costs a ~120 ms node spawn on every tool call (~85 ms of which is bare process startup), and both things the wide hook does — resolving remote approval gates and feeding `agent.tool_started` liveness — are useless without a web surface, so a terminal-only operator currently pays for nothing.

The profile derivation from `settings.json` rather than a marker file is the right call and well tested: a bare `wmux setup-hooks` re-run (including the app-update refresh and the in-app install button) keeps whatever profile is on disk instead of silently resurrecting a gate the operator removed, while a *partial* install still reads as broken-full and gets repaired. The silent failure mode — signals-only combined with `wmux web --allow-input` leaving a phone waiting on approvals that will never be raised — is guarded with a startup warning, and `--status` now distinguishes "gate absent by configuration" from "gate genuinely missing". Plugin users keep the full profile, with `--signals-only` reporting it had no effect rather than pretending, which is the honest behavior given the plugin format has no per-install switch.

Scope is tight and appropriate: `setupHooks.ts` (+245/−33), a small `web.ts` warning hook, CLI flag wiring, README and changelog entries, plus 24 new tests across the profile matrix and the warning path. Full CI including the cross-platform baseline and perf gates is green, and the PR is cleanly mergeable. No production behavior changes for existing full-profile installs beyond the derived-profile logic, which the tests confirm preserves the default. Priority P3 — it closes a design question with no urgent user-facing breakage, and the risky edge cases (partial installs, corrupted settings, plugin conflicts) are each explicitly covered.

Prioridade: **P3**

PR #982, opened 2026-08-21
fix(themes): stop dropping xtermOverrides on every session load

### LINKED ISSUE #981: [Bug]: Custom theme and terminal appearance settings are lost after a version upgrade
### What happens

Appearance settings survive fine in normal use: pick a custom profile, set the terminal colours, hit apply, and it looks right across restarts. But after a **version upgrade** they are gone — the custom profile and the terminal colours are back to defaults and have to be entered again from scratch.

### What I checked before filing

The settings are persisted where you'd expect, and they are there right now, so this is not a case of them never being written. `%APPDATA%\wmux\session.json` (userData), mtime today:

```
theme               "custom"
customThemeColors   { bgBase, bgSurface, bgMantle, textMain, textSub, textMuted, accent, … }  (10 tokens)
terminalFontFamily  "Cascadia Code"
terminalFontSize    16
uiScale             1
locale              "pl"
```

That directory is not touched by Squirrel's `app-x.y.z` swap, so the loss cannot be a file being replaced on disk — something in-app must be overwriting or failing to read them on the first launch after an upgrade. From the outside I can only narrow it to three candidates:

1. **A default-initialised save before hydration.** If the first post-upgrade launch writes a session snapshot (`buildSessionData` carries `theme`, `customThemeColors`, the font fields) before the stored one has been read back, the defaults win and the file is overwritten with them. That would match the symptom exactly: nothing is corrupt, the values are simply replaced by defaults once.
2. **A migration that re-runs on already-migrated data.** `migrateCustomThemeColors` (legacy 37-field → 10 tokens) and `migrateThemeId` both run on load; a payload already in the new shape passing through the legacy branch would come out empty rather than unchanged.
3. **A different userData dir on first run.** `SessionManager`'s own comment mentions writing "session.json/metadata.json to the UN-suffixed userData dir", which implies suffixed variants exist. If the `--squirrel-firstrun` launch resolves a different one, it would read no settings and then save defaults over the real file.

I could not distinguish between them from here, mostly because of the next point.

### Two observations that may or may not be related

- **`%APPDATA%\wmux\app-version.txt` contains `0.8.7`** while the installed app is 3.44.0. I found no reader for that filename anywhere in the tree, so it may be a leftover from an older build — but if anything does key an upgrade path off it, it is wrong by a wide margin.
- **The backups rotate too fast to catch this.** `session.json.bak`, `.bak.1`, `.bak.2` were all written within a few minutes of each other, so by the time a user notices their theme is gone, the pre-upgrade copy is already out of the window. A single longer-lived pre-upgrade snapshot (taken by the same code that already does "Session save triggered before update install") would make this diagnosable — for this bug and for any future one about lost state.

### Repro

1. Settings → custom profile, change terminal colours, apply. Restart a few times: settings persist.
2. Take an in-app update to a new version.
3. First launch on the new version: theme is back to default, `customThemeColors` gone.

(On this machine step 2 currently cannot complete at all — see the updater issue I filed alongside this one — so I cannot bisect which of the three candidates it is without a build where updates install.)

### Confirmation

p-poppe:
## Summary

`migrateCustomThemeColors` drops `xtermOverrides` every time it runs, and `loadSession` runs it on every load — so per-slot terminal colour overrides last exactly one session. Closes #981.

## The bug

The already-migrated branch rebuilds the theme from a fixed whitelist:

```ts
if (typeof obj.xtermPaletteId === 'string' && typeof obj.xtermBackground !== 'string') {
  const ui: UIThemeTokens = { /* 11 UI tokens */ };
  return { ...ui, xtermPaletteId: obj.xtermPaletteId as XtermPaletteId };   // overrides gone
}
```

…while its doc comment says *"Idempotent on already-migrated shapes"*.

Verified by calling the shipped function on my live `session.json`:

```
before: xtermPaletteId=tokyo-night  xtermOverrides={"background":"#000000","foreground":"#4AFF00","cursor":"#4AFF00"}
after : xtermPaletteId=tokyo-night  xtermOverrides=undefined
```

Why it reads as an *upgrade* bug (how #981 was originally filed): `extractXtermColors` layers the overrides over the preset at render time, so applying them works and keeps working for the whole session. They are gone at the next launch — and an upgrade is the restart people notice. The preset id is whitelisted and survives, so only users who tuned individual ANSI colours are affected.

## Changes

- `sanitizeXtermOverrides()` (new, module-local) — keeps only keys that exist on `XtermThemeColors` with non-empty string values, the same rule `extractXtermColors` applies at render time ("unknown override keys are ignored"), enforced once at the load boundary so a hand-edited `session.json` cannot smuggle junk into the store. Returns `undefined` when nothing survives, so the key is omitted rather than persisted as `{}`.
- The already-migrated branch carries the sanitised overrides through.
- Doc comment corrected to say what the function now actually does.

The legacy 37-field branch is untouched: that shape predates overrides, so there is nothing to carry.

### Fixed

- **Terminal colour overrides survive a restart.** Customising individual ANSI colours on top of a palette preset (Settings → Terminal Palette → Customize terminal colors) worked for the rest of the session and was then lost at the next launch: the custom-theme migration that runs on every session load rebuilt the theme from a fixed field list that did not include the overrides, while documenting itself as idempotent. The preset id survived, so only users who tuned individual colours were affected.

## Stability tier impact

- [x] Change to an `internal` surface — a renderer-side migration helper. No event, RPC or persisted-shape change: `xtermOverrides` is already part of `CustomThemeColors` and already written by `buildSessionData`; this only stops the load path from deleting it.

n/a.

## Test plan

- `themes.test.ts` — four new cases: idempotence on a migrated shape **that carries overrides** (twice through, to catch erosion); only real palette slots with string values are kept; the key is omitted when nothing survives; the legacy shape still yields no overrides.
- Worth noting why this was invisible: the existing `it('is idempotent on already-migrated shape')` builds its fixture with `builtinToCustom(...)`, which never produces `xtermOverrides` — the one non-idempotent field was absent from the test asserting idempotence.
- `themes.test.ts` + `themeParity.test.ts` + `uiSlice.inspect.test.ts`: 60 tests pass. `npx tsc --noEmit` clean.
- Manual verification: **not done** — no Electron binary in this sandbox. On a real build: set a couple of ANSI slots, restart, confirm they are still applied and still listed as overriding the preset.

## Related issues / PRs

Closes #981.

## Summary by CodeRabbit

- **Bug Fixes**
  - Customized ANSI terminal color overrides now persist when sessions are reopened or restarted.
  - Theme loading filters invalid color overrides while preserving valid customizations.
  - Repeated session loads no longer alter or duplicate saved terminal color settings.

- **Documentation**
  - Updated migration documentation to reflect persistent terminal color override support.


openwong2kim:
This fixes the mechanism behind #981: the already-migrated branch of `migrateCustomThemeColors` rebuilt the theme from a fixed whitelist that omitted `xtermOverrides`, and because `loadSession` runs the migration on every load, per-slot ANSI colour overrides survived exactly one session — an upgrade is simply the restart people notice. The fix is minimal and well-scoped: a module-local `sanitizeXtermOverrides()` keeps only real palette slots carrying non-empty strings (the same rule `extractXtermColors` already applies at render time, now enforced once at the load boundary) and carries them through, omitting the key entirely when nothing survives rather than persisting an empty object. The four new tests close the exact gap that hid the bug — the previous idempotence fixture was built with `builtinToCustom`, which never produces `xtermOverrides` — and the double-pass case guards against erosion. Before closing #981 outright it is worth confirming on a real build that the reporter's other candidates (a default-initialised save before hydration, the stale `app-version.txt`) are ruled out, since the manual verification step is still outstanding.

Issue #978, opened 2026-08-21
Triage the 95 Dependabot alerts that appeared when alerts were switched on


openwong2kim:
Enabling Dependabot alerts while closing #969 surfaced 95 open alerts, **every one created at the moment the setting flipped** (`oldest=2026-08-21T09:37:31Z`, `newest=…:49Z`). This is not a deferred backlog — alerts were off, so no one had ever seen it. Filing one issue rather than reacting to 95, because the headline split is misleading and the fix is much smaller than the count suggests.

## The count, and why the runtime half is not what it looks like

| | high | medium | low | total |
|---|---|---|---|---|
| runtime | 8 | 29 | 3 | **40** |
| development | 25 | 22 | 8 | **55** |

All 40 runtime alerts trace to a single direct dependency, `@modelcontextprotocol/sdk@1.29.0`, and every one of them lives in an HTTP transport the SDK ships but wmux never loads:

```
hono 27, @hono/node-server 2   ← SDK's hono transport
express → qs 1, body-parser 1  ← SDK's express transport
router → path-to-regexp 2
express-rate-limit → ip-address 2
ajv 8 → fast-uri 5
```

wmux imports exactly three SDK subpaths:

```
14×  @modelcontextprotocol/sdk/server/mcp.js
 3×  @modelcontextprotocol/sdk/server/stdio.js
 1×  @modelcontextprotocol/sdk/types.js
```

and `dist/cjs/server/mcp.js` requires only `zod` plus internal SDK modules — no express, no hono. `streamableHttp.js` / `sse.js` are never imported, so none of this code is reachable in the app. The SDK declares them as flat `dependencies` (not optional), which is why they install and why they alert.

That is the whole runtime exposure story: **nothing reachable**. Recording it here so a future reader does not re-derive it, and so "40 runtime alerts" never gets quoted as-is.

## Most of it is a lockfile refresh, not a bump

Every one of these already sits inside its parent's declared range — `npm update` moves them with no manifest edit:

| package | installed | patched | parent's range | in range |
|---|---|---|---|---|
| hono | 4.12.8 | 4.12.34 | `^4.11.4` | yes |
| @hono/node-server | 1.19.11 | 1.19.15 | `^1.19.9` | yes |
| fast-uri | 3.1.0 | 3.1.5 | `^3.0.1` (ajv 8.18.0) | yes |
| qs | 6.15.0 | 6.15.2 | `^6.14.0` | yes |
[...]

PR #983, opened 2026-08-21
chore(ci): hand the lockfile to npm 11 on the Node 22 runtime

openwong2kim:
## Why now

All five of Dependabot's first security PRs (#972–#976) arrived dead: every job fails `npm ci` with EUSAGE, `Missing: esbuild@0.27.7 … @esbuild/* from lock file`. That is the lockfile-lineage split that has bitten this repo before (#364, #365): CI's npm 10 requires nested platform entries (`vitest/node_modules/@esbuild/*`) in the lock, while npm 11 — the npm on contributor machines and inside Dependabot — prunes them when it regenerates. Any lock touched by npm 11 fails npm 10's `npm ci`. Dependabot cannot be told which npm to use, so under npm-10 CI it cannot open a passing PR at all.

## The change

Pick one lineage — the one the tools already use — and align CI to it, **without** moving the runtime major:

- Every `setup-node` step is followed by `npm install -g npm@11`. The runtime stays **Node 22**; npm 11 supports it (`^20.17 || >=22.9`).
- `package-lock.json` renormalized once with npm 11.12.1. Deletions only — 512 pruned nested optional entries, zero resolved-version changes at the install-list level (one resolution nuance below).
- **Lineage guard** in ci.yml: after `npm ci`, renormalizing with npm 11 must be a no-op (`npm install --package-lock-only --ignore-scripts && git diff --exit-code package-lock.json`). Without it, a contributor on npm 10 who follows the EUSAGE error's own advice ("update your lock file with `npm install`") resurrects the pruned entries at whatever the registry serves that day, and CI stays green while the lineage silently regresses. Verified idempotent locally before wiring in — and its first CI run caught a real desync in this very branch (the engines edit below had not been mirrored into the lock).
- `engines`: `node` stays `>=18` (the app never needed more); the real demand is spelled `npm >=11`. CONTRIBUTING says the same in words.
- This retires the standing "regenerate the lock with `npx npm@10` only" rule carried in every delegation spec since #365.

## Why the runtime did not move

The first version of this PR moved CI to Node 24 (`node-version: 24`, then a `24.18.1` pin). Both died the same way: on windows-latest, one vitest worker fork exits silently per run — every executed test green, one test file left unreported, `Worker exited unexpectedly` surfacing only when the pool hands the dead fork its next file. Reproduced on 24.19.0 (2/2 baseline) and 24.18.1 (baseline + validate); macOS/Linux and every build path (including a full release rehearsal) were green throughout. Full evidence matrix and unblock criteria: **#984**. The lockfile goal never depended on the runtime major, so the two are decoupled and the runtime move waits on #984.

## The one-way door

After this merges, `npm ci` under npm 10 rejects the lock — the failure the Dependabot PRs show today, pointing the other way. The failing side becomes "a stale local npm", fixed with `npm install -g npm@11`, instead of "every automated lock regeneration", which nobody can fix.

## Verified

| gate | result |
| --- | --- |
| `npm ci` on the renormalized lock (npm 11.12.1) | clean install |
| `tsc` across all four projects | 0 errors |
| eslint | clean |
| vitest smoke (`imeAnchor.test.ts`) | 45/45 |
| lineage guard idempotency | no-op confirmed locally, then caught a real desync on CI |
| release path | full `workflow_dispatch` rehearsal green on all 3 OS (publishing steps stand down by design) |

Panel review (Claude·Codex·GLM, independent) ran on the Node-24 revision: Codex no findings; the accepted findings (lineage guard, engines, stale step name, release rehearsal) are in; the Node-24-specific ones are mooted by the decoupling. One noted nuance kept for the record: at the resolution level the nested `vite@8` now sees the hoisted `esbuild@0.21.5` (out of its optional-peer range) instead of a nested `0.27.4` — inert on the default rolldown path, worth knowing if vitest's esbuild path ever activates.

Refs #978. Blocks nothing; unblocks #972–#976 (rebase after merge).


openwong2kim:
This moves all four workflows (seven `node-version` pins) to Node 24 and renormalizes `package-lock.json` once with npm 11, retiring the lockfile-lineage split that has left every Dependabot security PR (#972–#976) dead on arrival with `npm ci` EUSAGE. The lockfile diff is deletions-only — pruned nested optional esbuild platform entries with zero resolved-version changes — and the vitest smoke run confirms the deduped platform binary still installs cleanly. The one-way door is stated honestly: after this lands, the failing side is an old local Node running npm 10, which a contributor can upgrade, instead of every automated lock regeneration, which nobody could fix. Two things for review: `release.yml` carries three of the pins and only truly exercises on a `v*` tag, so the release path is proven at the next cut rather than by this PR's CI; and the five Dependabot PRs will each need `@dependabot rebase` after merge to pick up the new lineage. This also unblocks the plan in #978 and moves in the same direction as #964.

Issue #980, opened 2026-08-21
[Bug]: In-app update never installs — the waiter aborts on wmux's own surviving processes, and the UI keeps offering "Install now"


p-poppe:
### What happens

Clicking **Install now** (or the update toast) restarts wmux without installing anything. The app comes back on the old version, Settings still shows *update ready → install now*, and the button then does nothing at all until the app is restarted by hand.

Two cycles on this machine, both with the same outcome, and the install root shows why.

### Environment

- Windows 11, wmux **3.44.0** installed (`%LOCALAPPDATA%\wmux\app-3.44.0`, the only `app-*` dir).
- Latest release: **3.45.0** (published 2026-08-21 01:26Z).
- Plugin-less hook install; several daemon-backed panes with `claude` running.

### Evidence

**1. The update is downloaded and verified — and then never installed.**

```
[AutoUpdater] Update downloaded + verified (sha256 match) — ready to install
[AutoUpdater] Session save triggered before update install
[AutoUpdater] install handoff: waiter=…\wmux-install-waiter-xpeSYH\wait-and-install.ps1
              watching 5 process(es) under …\wmux; force-killed 4/4,
              daemon pid 26304 left to the graceful shutdown
[AutoUpdater] quitting for install — the daemon goes down with us so the installer runs against a dead tree
[AutoUpdater] Enabled                       <- back up, same version
[AutoUpdater] install already in progress — ignoring re-entrant call.
[AutoUpdater] install already in progress — ignoring re-entrant call.
[AutoUpdater] check error: net::ERR_FAILED  (×4)
```

The Setup.exe is on disk, untouched: `%TEMP%\wmux-update-3.45.0-25728-wmux-3.45.0.Setup.exe`, 150 MB, written 03:41 today.

**2. Nothing ever ran it.** `%LOCALAPPDATA%\wmux\packages\RELEASES` still lists only `wmux-3.44.0-full.nupkg`, `packages\SquirrelTemp` is empty, and the last line in `Squirrel-Shortcut.log` is from **20.08 23:40:52** — the previous cycle.

**3. The processes the waiter waits for are still alive — a day later.** The generated `wait-and-install.ps1` waits for every handed-over PID within a 60 s budget, then aborts rather than launching Setup.exe:

```powershell
foreach ($id in @(55592,42600,26304,53660,6496)) { … }
…
if (-not $h.WaitForExit($left)) { $stuck = $true; break }
[...]

PR #985, opened 2026-08-21
feat(themes): add a Matrix terminal palette

p-poppe:
## Summary

Adds a `Matrix` xterm palette: phosphor green on green-tinted black, with the classic bright green as the cursor — the one full-intensity thing on screen.

Requested by a user (me) who had been hand-entering green overrides slot by slot; as a built-in it is also immune to the override-dropping bug in #981, since `xtermPaletteId` survives the load path while `xtermOverrides` does not.

## The two decisions worth reviewing

Making everything green is the obvious version and it is worse:

1. **Red and yellow keep their warmth.** A palette where a failing test, a `git diff -` line and a warning all render green throws away the only job colour has in a terminal. Red stays the error channel (`#FF5F56`, hue 3°); yellow stays out of the green band (`#C8FF64`, hue 81°) so a warning cannot read as success.
2. **The cool slots are separated by hue, not lightness.** My first draft had `blue` and `green` differing only in lightness, which leaves `ls` unable to distinguish a directory from an executable. Final: blue 185° teal, cyan 168° aqua, magenta 150° pale mint.

`background` is `#080D08` rather than pure black so `brightBlack` — dim text such as comments and timestamps — clears 3.9:1 instead of vanishing.

## Measured (via the repo's own `getContrastRatio`)

| slot | hex | on background |
|---|---|---|
| foreground | `#33FF66` | 14.6:1 |
| cursor | `#00FF41` | 14.4:1 |
| red | `#FF5F56` | 6.6:1 |
| yellow | `#C8FF64` | 16.8:1 |
| blue | `#009E8F` | 9.3:1 |
| cyan | `#5FFFE0` | 15.7:1 |
| brightBlack (dim text) | `#4A7A5C` | 3.9:1 |

## Changes

- `XtermPaletteId` gains `'matrix'`; the palette is added to `XTERM_PALETTES` and to `XTERM_PALETTE_OPTIONS` so it appears in the picker.
- No built-in UI theme maps to it — this is a terminal-palette-only addition, selectable from a custom theme. A matching UI theme (11 tokens + `globals.css` parity) would be a separate change if you ever want one.

### Added

- **A `Matrix` terminal palette** — phosphor green on green-tinted black, in Settings → Terminal Palette. Green carries the identity (foreground, cursor, the green slot and a green-cast white), but red and yellow keep their warmth so a failing test, a removed diff line and a warning still read as themselves, and blue/cyan/magenta are separated on the hue wheel rather than collapsed into green, so `ls` can still tell a directory from an executable.

## Stability tier impact

- [x] Additive change to a `stable` surface (a new value in the `XtermPaletteId` union). Nothing existing changes shape; unknown ids already fall back to `catppuccin-mocha` in `extractXtermColors`, so a session written by this build and opened by an older one degrades to the fallback rather than breaking.

n/a.

## Test plan

- `themes.test.ts` — eight new cases that lock the *decisions* rather than the hexes: present in the picker; every slot valid hex; AA for foreground and cursor; ≥3:1 for dim text; ≥4.5:1 for every colour slot; red inside the red hue band; yellow below 100°; and a minimum hue distance between green and blue/cyan.
- `src/renderer/__tests__` 6 files / 87 tests pass. `npx tsc --noEmit` clean, `eslint` clean on both touched files.
- Manual verification: **not done** — no Electron binary in this sandbox, so the numbers above are computed, not eyeballed. Worth a look on a real pane: `git diff`, a failing test run, and `ls --color` are the three outputs these decisions were made for.


openwong2kim:
This is the fix for #980 (stuck in-app updater), and the root cause it identifies is solid: the install handoff force-killed every wmux.exe under the install root except the daemon — which on Windows includes our own renderer, GPU and utility helpers — and then quit into a before-quit handler that awaits a session save in the renderer it had just killed, so the quit never completed and the main process held the install root open for a day. The wait-list/kill-list split is the right shape: everything under the root must still be gone before Setup.exe runs, but only foreign processes (the agent-hosted MCP servers, identified by parentage via selectOwnTreePids) are killed by us. The hardening items riding along each close a distinct silent-failure mode — the 20 s hard quit deadline armed before anything can hang, the bounded renderer save, the TickCount-to-Stopwatch fix for a waiter that went blind after ~24 days of uptime, and the quit watchdog that reports and unlatches instead of leaving "install already in progress" forever. Test coverage is proportionate (13 new cases including pid-recycle and unreadable-parent edges, runtime suite 8/8 against real processes and file locks), and deferring artifact reuse to a separate PR is the right scoping call. P1 to match #980 — this should merge as soon as review confirms, since every failed update cycle currently costs the user a 150 MB re-download and a manually restarted app.

openwong2kim:

(Reposted — the previous triage comment on this PR carried the wrong body, a summary of #986. Apologies for the noise; it has been removed.)

Additive `matrix` xterm palette: one new `XtermPaletteId` union member, one `XTERM_PALETTES` entry, one picker option, and 63 lines of tests — no changes to existing palettes or to the load path, so the blast radius is effectively zero. The two design decisions called out in the description are the right ones for a terminal (red/yellow keep their semantic warmth instead of going monochrome-green, and the cool slots are separated by hue rather than lightness so `ls` can still tell a directory from an executable), and the inline comment documents both, which is what will keep a future "make it greener" edit from quietly undoing them. Worth noting the practical motivation holds: as a built-in palette it rides `xtermPaletteId`, which survives the session-load path that drops `xtermOverrides` (#981) — though once #982 lands that workaround stops being load-bearing. P3 as labeled; low-risk, reviewable in one sitting.

PR #996, opened 2026-08-21
fix(sidebar): give roster rows their width back

p-poppe:
## Summary

Two independent width thieves in the same roster row. Reported by a user (me) whose pane titles — the thing #934 put there so rows could be told apart — were truncating to `Diagnos…`, `Zorgan…`, `E2eTE…` in a 240px sidebar while the row spent its space on text that was identical on every line.

Happy to split this into two PRs if you'd rather review them apart; they land in the same row for the same reason, which is why they are together.

## 1. The vendor name, on every row

For a workspace where every session runs the same CLI, `Claude Code` repeats what the line above already said and costs roughly a third of the row. It is now omitted for single-vendor rosters and returns the moment a workspace mixes vendors — which is when it answers a real question. `rosterHasMixedVendors()` decides once per render, not per row.

```
before   ● Zwroty          Claude Code · w2-127   Oczekuje
after    ● Zwroty          w2-127                 Oczekuje
mixed    ● Zwroty          Claude Code · w2-127   Oczekuje      <- unchanged
```

A titleless row still leads with the vendor, so no row is ever left nameless.

## 2. Polish status labels prefixed the subject

`'workspace.agentWaiting': 'Agent oczekuje'` — in the narrowest column of the layout, on a row that already names the agent, next to a coloured dot. Now bare states (`Oczekuje`, `Pracuje`, `Ukończone`, `Bezczynny`, `Błąd`), which is what `en` and `ko` already do.

While there: **`workspace.agentAwaitingInput` was missing from `pl` entirely**, so the one status a user must act on rendered as English `Needs input` in an otherwise Polish UI. Added.

**19 of 21 non-English locales have the same hole** (only `ko` and `zh` carry the key) — filed separately rather than machine-translating twenty languages inside this PR.

### Changed

- **The agent roster spends its width on the name, not on repetition.** In a workspace where every session runs the same agent, the vendor name is no longer printed on each row — it returns as soon as a workspace mixes vendors, which is when it answers something. Polish status labels lost their "Agent " prefix (the row already names the agent) and gained the missing awaiting-input label, which had been falling back to English.

## Stability tier impact

- [x] No external surface touched. `rosterSecondaryLabel` takes a new optional `opts` argument and defaults to the previous behaviour, so any other caller is unaffected; the rest is a locale file.

n/a.

## Test plan

- `workspaceAgentRosterLabels.test.ts` — 7 new cases: single-vendor vs mixed detection (and empty → false); vendor omitted while the coordinate stays; vendor kept when mixed; tab position survives the omission; a titleless row still leads with the vendor; and the default-argument case, so existing callers are pinned to the old output.
- `src/renderer/components/Sidebar/__tests__` 7 files / 56 tests pass. `npx tsc --noEmit` clean, `eslint` clean on the three touched files.
- Manual verification: **not done** — no Electron binary in this sandbox. What deserves an eye: a mixed-vendor workspace (the case where the vendor must reappear) and a very long pane title, to see where truncation now lands.

## Summary by CodeRabbit

* **Improvements**
  * Simplified agent roster labels by hiding repeated vendor names when all agents use the same vendor.
  * Vendor names remain visible when a workspace includes multiple vendors.
  * Improved Polish agent status labels and added a localized “awaiting input” status.


openwong2kim:
Small, well-scoped polish PR: omits the repeated vendor name on single-vendor roster rows (`rosterHasMixedVendors` gates it per render) and fixes `pl` status labels, including the missing `agentAwaitingInput` key. `rosterSecondaryLabel` keeps its default behavior for existing callers, so risk is minimal. P3 — nice UX polish, tests cover the gating and label cases.

PR #999, opened 2026-08-21
fix(clipboard): do not re-execute OSC 52 writes from replayed output

### LINKED ISSUE #998: [Bug]: Replaying a pane re-executes old OSC 52 writes and silently overwrites the system clipboard
### What happens

Copy something, verify it is on the clipboard, then paste in another pane — and get text you copied **days ago**, from a different session.

Concretely, what I did: selected a sentence in a Claude Code pane and copied it, checked it in Notepad (correct), opened another terminal, pasted — and got the content of an entirely different, much older copy. Nothing I did in between touched the clipboard.

### Cause

`useTerminal.ts` registers the OSC 52 write bridge on the terminal parser, so a TUI that copies via OSC 52 sets the system clipboard. That is the feature and it works.

The problem is that **those sequences are stored in the pane's buffer**, and replay writes the buffer back into the same xterm with that handler live:

- `term.write(bytes)` / `for (const chunk of st.buffer) term.write(chunk)` on the resync + legacy reconnect paths
- `terminal.write(restored)` on the scrollback-restore path

So every replay re-executes whatever OSC 52 writes the buffer happens to contain, and the clipboard is silently overwritten with text from whenever that output was produced. A reconnect, a resync, revealing a cold-parked pane, or restoring scrollback at startup is enough — the user is doing nothing clipboard-related at the time.

### Evidence from this machine

Scanning the daemon ring buffers in `~/.wmux/buffers` for `ESC ] 52 ; … ;` :

```
files containing OSC 52 writes: 10
total sequences:                70
```

Decoding the payloads, they are my own copies from the past several days — sentences from agent panes, in the same order I made them. (Content withheld; it is work material.) Any replay of those buffers re-applies one of them to my clipboard.

That also gives the bug an unpleasant second edge: **it can resurrect a clipboard entry from days ago off disk.** Anything ever copied inside a pane — a token, a password, a customer detail — is a candidate, because the ring buffer outlives the session that produced it.

### Repro

1. In a pane running a TUI that copies via OSC 52 (Claude Code, tmux, neovim), copy some distinctive text — call it A. Confirm A is on the clipboard.
2. Do something else for a while, so more output accumulates. Copy something new outside wmux — call it B. Confirm B is on the clipboard (Notepad, or `Get-Clipboard`).
3. Force a replay of the pane from step 1: reconnect it, let a resync happen, or restart wmux with scrollback restore enabled.
4. Paste. You get A.

On this machine step 3 happens by itself often enough that I hit it without knowing what "step 3" was.

### Suggested fix, and the one nuance I'd want your call on

Gate the handler rather than filtering bytes: a `replaying` flag the OSC 52 handler checks, so historical bytes cannot reach the clipboard. Stripping `ESC ] 52` out of the replay payload would work until the day a sequence straddles a chunk boundary, and a raw-stream rewrite is a worse thing to own than a boolean.

The nuance: on the resync path the same loop writes **two different kinds** of bytes — the historical replay payload, and `st.buffer`, which is *live* pty data held back while the resync was in flight. A flag that covers the whole loop would also swallow a legitimate OSC 52 that arrived during the resync window (rare, but it is a real copy the user just made). So the flag wants to be scoped to the historical payload only, with the buffered-live chunks written outside it — which is a two-line reorder rather than a one-line guard, hence asking rather than assuming.

Happy to write it either way once you say which shape you want. It is testable without a running app: feed a payload containing an OSC 52 write through the replay path with the flag set and assert `clipboardAPI.writeText` was not called.

### Confirmation

p-poppe:
## Summary

Closes #998. A copy is an act, and an act cannot be replayed — so the OSC 52 write half is now closed while xterm parses stored output, exactly as the read half always is.

Per the triage ("the proposed replay-gated handler flag is the right shape"), this is that flag, with the one nuance I flagged in the issue resolved in the conservative direction: **live bytes stay ungated.**

## Changes

- **`createOsc52Handler({ isReplaying, writeClipboard })`** in `utils/osc52Clipboard.ts` — the handler moves next to the policy it belongs to (that module already documents why reads and clears are refused; "and not from a replay" is the same kind of rule). DOM-free, so it unit-tests without a terminal.
- **`writeReplayed(term, data, depth)`** in `useTerminal.ts` — mutes the bridge, writes, and unmutes **in xterm's write callback**. Unmuting synchronously would release while the escape sequences are still queued, which is the bug itself. A counter rather than a boolean, because the replay paths write several payloads back to back.
- Three historical-payload sites routed through it: the dead-pane snapshot paint, and the restored `.txt` scrollback. (`STALE_REPLAY_*` reset constants are ours, not the pane's, so they stay plain writes.)
- Teardown force-resets the depth: a terminal disposed mid-parse never delivers its callback, and a leaked mute would silently disable copying for the next pane that reuses the hook.

## What is deliberately NOT gated

`for (const chunk of st.buffer) terminal.write(chunk)` on both the resync-complete and degrade paths. Those chunks are **live** pty data held out of xterm while a resync was in flight — a copy the user made in that window is genuine and must still reach the clipboard. This is the two-line split I asked about in the issue; gating the whole loop would have been simpler and wrong.

## Evidence this was worth fixing

On this machine before the fix: **70 OSC 52 write sequences across 10 daemon ring buffers**, the decoded payloads being my own copies from previous days. Any replay of any of those buffers re-applied one of them. The user-visible symptom that started it: copy something, verify it in Notepad, paste in another pane, get text from days ago.

### Fixed

- **A replayed pane no longer overwrites your clipboard.** Reconnecting a pane, resyncing it, or restoring its scrollback writes stored output back into the terminal — and any OSC 52 clipboard write inside those bytes was executed again, silently replacing whatever you had just copied with text from whenever that output was produced. Since the ring buffer outlives the session that produced it, this could resurrect a copy from days earlier, including one you would not want back. The clipboard bridge is now closed while historical bytes are parsed and open for everything live, so a copy made during a resync still lands.

## Stability tier impact

- [x] Change to an `internal` surface. `createOsc52Handler` is a new export in a renderer util; `decodeOsc52Write` keeps its signature and its tests. No event, RPC or persisted shape touched.

n/a.

## Test plan

- `osc52Clipboard.test.ts` — 5 new cases beside the existing 11: live sequence writes; replaying writes nothing (while still consuming the sequence); a malformed payload is not even decoded while replaying; the predicate is read per sequence so the gate reopens after the replay; reads and clears are still refused when not replaying.
- 16/16 pass. `npx tsc --noEmit` clean. `eslint` on the touched files reports the same 3 errors + 3 warnings as `main` does with the patch stashed (pre-existing empty-arrow-function and non-null-assertion sites).
- Manual verification: **not done** — no Electron binary in this sandbox. On a real build the check is the repro from the issue: copy A in a TUI pane, copy B outside wmux, force a replay (reconnect the pane or restart with scrollback restore), then paste — should be B, was A.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented restored, replayed, or resynchronized terminal history from overwriting the current clipboard.
  - Prevented delayed output from closed terminal sessions from changing clipboard contents.
  - Preserved clipboard updates for live terminal output.
  - Continued rejecting clipboard read and clear requests.

- **Tests**
  - Added coverage for live writes, replay suppression, replay completion, invalid requests, terminal transitions, nested replay handling, and stale replay protection.


openwong2kim:

Fixes #998 by muting the OSC 52 clipboard bridge while replayed output (dead-pane snapshots, restored scrollback) is parsed, via a depth counter released in xterm's write callback; live resync-buffered bytes stay ungated. Touches useTerminal.ts and osc52Clipboard.ts, with new unit tests. P2 — matches the clipboard-overwrite bug it closes.

p-poppe:
Both review findings were right, and the second one is worth spelling out because it means my PR description was wrong about the main path. Pushed fixes for what the renderer can decide; one part needs your call.

## Fixed: generation-scoped mute

This hook outlives the terminal it hosts, so a write callback from a disposed terminal could decrement the depth belonging to its replacement and unmute *that* terminal mid-replay — the same bug wearing a different hat. The mute state is now `{ depth, gen }`, callbacks only decrement while their generation is current, and teardown bumps the generation rather than only zeroing the depth.

## Fixed, and a correction to my own reasoning: the resync flush is muted now

I claimed `st.buffer` holds live bytes and left it open. That is wrong on the raw-fallback path: `pty.reconnect`'s replay arrives as ordinary `pty:data` and lands in that same buffer — the code says so two lines above the loop I edited (*"the held replay lands on the clean buffer"*). So the buffer is **mixed**, and the reconnect case — the one my own bug report started from — was still leaking. Now the whole flush goes through `writeReplayed`.

The trade is deliberate, and it is the asymmetry argument from #935 applied here: a genuine copy made inside a ~1s resync window is lost and the user can copy again; a replayed write silently replaces the clipboard with a value that can be days old and is discovered only after pasting the wrong thing somewhere. A retry beats corruption.

## Not fixed, because the renderer cannot see it: the bare reconnect

The third route has no marker at all. `reconnectPtyWithRetry` calls `pty.reconnect` **without** a resync in flight, so its replay arrives as plain `pty:data` with `st.pending === false` and goes straight to `writeTerminalOutput`. Nothing in the renderer distinguishes those bytes from the program printing something now.

I can see two shapes and neither is mine to choose:

1. **Label the replay at the source.** The daemon/main side knows it is flushing a ring buffer, so a flag on the data event (or a `pty:replay-begin` / `pty:replay-end` pair around the flush) would make the gate exact everywhere, including this path. Costs a small IPC-surface change — your contract.
2. **Heuristic in the renderer.** Treat data arriving in the first N ms after a reconnect as replay. Cheap, no protocol change, and wrong often enough that I would rather not: a long-running agent printing during those milliseconds gets its copy swallowed for no visible reason.

I lean hard on (1) — it is the only version that is actually correct, and it also lets the gate drop the mixed-buffer compromise above, because live and replayed bytes would finally be distinguishable. Say which you want and I will write it; if you would rather take this PR as the partial fix (dead snapshot, restored scrollback, resync flush) and handle the reconnect path yourself where the protocol lives, that is fine too — say so and I will note the remaining hole in the issue rather than leaving it implied.

Still 16/16 on the util tests, `tsc` clean, lint unchanged against `main`.

p-poppe:
CI was right too — I had broken five of your source-level locks. Pushed a fix; here is the accounting, since these are your regression tests and "the test was in the way" is not a reason I want to leave implied.

Every one of them reads `useTerminal.ts` as text and matches literal call sites, so routing replay payloads through `writeReplayed()` moved the ground under them. In each case I kept the assertion's intent and changed only the spelling, plus the slice window where an added comment pushed the target out of range:

| Lock | What it pins | What I changed |
|---|---|---|
| `staleReplayModeReset` | payload write precedes the mode resets | accepts `writeReplayed(term, bytes`; `completeResyncFromFlush` window 1200 → 2400 |
| `hiddenRetention` — resync | `terminal.reset()` BEFORE the held bytes are parsed | matches the muted flush; window widened |
| `hiddenRetention` — dead snapshot | discard → reset → write → clean ordering | payload write matched on the new call |
| `daemonModeRaceCancel` | the daemon-guarded `restored` value is what gets written | accepts either spelling |
| `osc52` ×2 | decode + non-null guard reached from the hook | those moved into `createOsc52Handler`, so the lock now pins that the hook goes through the factory instead of hand-rolling a clipboard path |

And one lock **added**, because it is the contract this PR introduces and precisely the thing a future refactor would drop without noticing:

```ts
expect(SRC).toMatch(/isReplaying:\s*\(\)\s*=>\s*replayMuteRef\.current\.depth\s*>\s*0/);
```

If you would rather those locks stayed literal and I introduced a helper name they could match without edits, say so and I will rework it that way — I did consider naming the helper `terminalWriteReplay` so the existing `terminal.write(` patterns kept matching, and decided that optimising source text for a regex was the wrong tail wagging the dog. Your call, it is your test suite.

Full renderer suite now 3564/3565 locally. The single failure is `atlasCoherence` asserting the `addon-webgl` patch that `postinstall` applies — my sandbox installed with `--ignore-scripts`, and it fails identically on a stashed tree, so it is my environment rather than this branch.

The open question from my previous comment still stands and is the more interesting one: the bare-reconnect path has no replay marker in the renderer, so it needs either a labelled data event from your side or a heuristic I would rather not write.

openwong2kim:
## Live dogfood: the gate never reaches the path #998 was filed about

Ran this on a real Windows build (integration branch = `main` + #985 + #982 + #1000 + #996 + this PR, daemon mode on, isolated `WMUX_DATA_SUFFIX`). The three gated sites behave as designed, but the reconnect path from the issue's own repro is still ungated, and the clipboard is still overwritten.

**Repro, 2/2:**

```
1. in a pane:  Write-Host -NoNewline ($ESC + "]52;c;" + b64("REPLAY-A-FROMPANE") + $BEL)
   -> clipboard becomes REPLAY-A-FROMPANE       (live write, correctly ungated — good)
2. outside wmux: Set-Clipboard "SENTINEL-B-EXTERNAL"
3. reconnect the pane (renderer reload -> daemon reattach)
4. Get-Clipboard
```

```
[1] before reload: SENTINEL-B-EXTERNAL
[1] after  reload: REPLAY-A-FROMPANE
[2] before reload: SENTINEL-B-EXTERNAL
[2] after  reload: REPLAY-A-FROMPANE
```

That is verbatim the symptom in #998 — *"copy something, verify it in Notepad, paste in another pane, get text from days ago"*.

**Why the gate misses it.** The reattach replay is not written through any of the three `writeReplayed()` sites. The log names the path:

```
[useTerminal] daemon reattach ptyId=daemon-b885caf3 (active-at-mount)
[lifecycle] pty.reconnect id=daemon-b885caf3 result=ok pid=5392
```

No `[wmux:reveal] mechanism=…` line — so neither the dead-pane snapshot nor `completeResyncFromFlush` ran. The RingBuffer replay arrives as ordinary `pty:data` and lands in `deliverPtyData`, which is the un-gated `writeTerminalOutput` path:

```ts
const routePtyData = (data: string) => deliverPtyData(restingCursor.process(data));
const deliverPtyData = (data: string) => {
  const st = resyncRef.current;
  if (st.pending) { st.buffer.push(data); … return; }   // <- only THIS branch is later gated
  writeTerminalOutput(terminal, data, { … });           // <- reattach replay goes here, ungated
};
```

The PR already knows this shape — the comment at the resync flush says *"the reconnect replay itself arrives as pty:data and lands here"*. It just holds only when a resync is `pending`. On a plain reattach (the common case: reload, respawn, `active-at-mount`) nothing is pending, so the bytes go straight to `writeTerminalOutput` with the bridge open.

**What still stands.** The unit tests, the handler extraction and the depth/generation scoping all look right to me, and the live write half works exactly as intended (step 1 above). This is a coverage gap, not a regression — `main` overwrites the clipboard here too. But since the reconnect case is the one the issue leads with, I would rather not land it as-is and have #998 read as fixed.

**Shape of the fix, if useful.** `deliverPtyData` is where the classification has to happen — the mute needs to cover the reattach replay window, not just `writeReplayed()` callers. Something anchored on the reconnect itself (mute from `pty.reconnect` until the replay settles, the same way `writeReplayed` unmutes in xterm's write callback) would cover it, though the mixed-buffer caveat from the resync path applies here too — a genuinely live copy inside the reattach window would be swallowed. Given the asymmetry you already argued for in this PR (a lost copy is a retry, a resurrected one is silent corruption), that trade seems consistent.

Happy to re-run the same rig against a follow-up commit — the harness is a CDP driver against a real build, so the check is a one-liner once a build exists.

p-poppe:
Your dogfood found the one path I could not see from here, and the fix is pushed.

**What I got wrong.** I gated the three sites that write replay payloads *through our own code*, then reasoned that the resync branch covered the reconnect case. It does not: on a plain reattach nothing is pending, so the RingBuffer flush walks straight through `deliverPtyData` into `writeTerminalOutput` with the bridge open. Your `[useTerminal] daemon reattach … (active-at-mount)` line with no `[wmux:reveal] mechanism=` after it is the proof, and it is also the common case rather than an edge.

**Taking your shape.** The mute now covers the reattach *window*, not just our writes. Pulled the state machine into `src/renderer/terminal/replayMute.ts` so it tests without a terminal:

| piece | when |
|---|---|
| `beginReplayWrite()` | our writes — released in xterm's write callback |
| `openReattachWindow()` | at the reconnect request, since there is no write of ours to hang it on |
| `noteReplayData()` | every chunk through `deliverPtyData` extends the window |

Two bounds, because a heuristic without bounds is a liability: **250 ms of quiet** ends the window (the flush arrives as a burst), and a **5 s hard cap** means a pane that never stops printing cannot hold the bridge shut. Everything stays generation-scoped, so a disposed terminal can neither leak a mute nor unmute its successor.

The swallowed-live-copy caveat applies here exactly as you said, and I did not try to be clever about it. A copy made inside the reattach window is lost; the user notices instantly and copies again. The alternative is a value from days ago landing in the clipboard silently.

**Tests.** 14 on the state machine (nesting, idempotent release, stale-generation release, quiet close, burst extension, the hard cap, at-most-one window, close-when-no-replay-arrives, coexistence with a write mute, teardown cancelling its timers) plus two source locks on the wiring — one of which is the reattach window itself, since that is precisely what a future refactor would drop. Renderer suite 3579/3580; the failure is `atlasCoherence` checking a `node_modules` patch my `--ignore-scripts` sandbox never applied, identical on a stashed tree.

Please re-run the rig when you get a chance. If step 4 still returns `REPLAY-A-FROMPANE` I would rather know that than have #998 read as fixed — and if it returns the sentinel, the interesting follow-up question is whether you want the window kept as a heuristic or replaced later by a labelled data event, which would let it drop the swallowing entirely.

p-poppe:
@C:/Users/PRZEME~1/AppData/Local/Temp/pr999-comment-fixed.md

openwong2kim:
## Rig re-run: the reattach replay is muted — and the control proves the rig can still see the bug

Re-ran the rig against `28a28e43`, this time on **macOS** (CDP driver against a dev build, `pbcopy`/`pbpaste` in place of `Set-Clipboard`, isolated `WMUX_DATA_SUFFIX`, pane input driven through the main pipe so focus can't confound it). So the fix is now verified on a second platform rather than only the one that filed the report.

**Result, 2/2 — step 4 returns the sentinel:**

```
[1] live-write   : REPLAY-A-FROMPANE
[1] before reload: SENTINEL-B-EXTERNAL
[1] after  reload: SENTINEL-B-EXTERNAL
[2] live-write   : REPLAY-A-FROMPANE
[2] before reload: SENTINEL-B-EXTERNAL
[2] after  reload: SENTINEL-B-EXTERNAL
```

Step 1 still lands live, so the mute did not cost the working half. The console shows the same line my previous comment pinned the miss on — `[useTerminal] daemon reattach ptyId=… (active-at-mount)` with no `[wmux:reveal] mechanism=` after it — except now the bridge is shut across that window.

**Control run, because a green rig on a new platform is worth one line of doubt.** Identical procedure against current `main` (`61fbc376`), same isolated instance, same driver:

```
[1] after reload: REPLAY-A-FROMPANE
[2] after reload: REPLAY-A-FROMPANE
```

It leaks 2/2. The pass on this branch is the fix, not the harness looking the other way.

**On your follow-up question**: keep the window as a heuristic here. It is bounded on both ends, generation-scoped, and covered by the state-machine tests — that is enough to close #998. A labelled replay event is the better end state and would let the swallowing go entirely, but it is an IPC-surface change and deserves its own PR rather than riding along in this one. I will file it as a follow-up so the swallowed-live-copy trade is tracked in writing instead of implied.

Merging. Thanks for taking the shape rather than arguing the gap.

openwong2kim:
Follow-up filed: #1014 — replacing the reattach heuristic with a labelled replay signal from the daemon, so the swallowed-live-copy trade is tracked rather than implied.

Issue #997, opened 2026-08-21
Roster chrome: the awaiting-input label is untranslated in 19 locales, and the "Agents N" row spends a row on two numbers


p-poppe:
### Problem / motivation

Two things I ran into while looking at a sidebar full of agents, both beyond what I'd fix without your call. Context: 240px sidebar, one workspace with 3–5 Claude Code sessions, which #934 gave real names to.

**1. The status a user must act on is untranslated in 19 of 21 locales.**

`workspace.agentAwaitingInput` exists in `en`, `ko` and `zh`. The other 19 locale files do not carry it, so the fallback renders English `Needs input` inside an otherwise translated UI:

```
● AI              Needs input        <- pl UI
● Wtyczk…         Agent bezczynny
```

It is the *only* status that means "stop what you're doing and look at me", which makes it an unlucky one to lose. I've added the `pl` string in #996; the remaining 18 I deliberately left alone rather than machine-translating languages I can't check.

While counting those I noticed the label style is split too: `en`/`ko` use bare states (`Waiting`, `대기 중`), while `de`/`fr`/`es`/`ja`/`pl`/`zh` prefix the subject (`Agent wartet`, `Agente en espera`, `エージェント待機中`). In a fixed narrow column, next to a coloured dot, on a row that already names the agent, the prefix costs width and adds nothing. #996 makes `pl` bare; whether the others should follow is a call for whoever owns those locales — hence this issue rather than a sweeping PR.

**2. Is the `Agents N` header row worth a row?**

Right now an expanded workspace spends one full row on the disclosure line:

```
● CTO                    ^6
  ▸ Agents 3 · Needs you 1
     ● WMUX      Claude C…  Agent oczekuje
     ● ROLA…     Claude C…  Agent ukończony
     ● Zorgan…   Claude C…  Agent oczekuje
```

That line carries two numbers and no name. Both numbers could live on the workspace row itself, where the eye already goes:

```
● CTO            3 · 1!     ^6        <- count, and needs-you
     ● WMUX      w2-127     Oczekuje
     ● ROLA…     w2-131     Ukończone
     ● Zorgan…   w2-124     Oczekuje
```

That buys a row per expanded workspace — with 11 workspaces open it is the difference between seeing everything and scrolling. The counter-argument I can see: the header is also the click target for collapse, and the chevron needs somewhere to live; hanging both numbers off the workspace row competes with the unread badge, the git line and the shortcut hint, which are already there.

[...]

PR #1000, opened 2026-08-21
i18n: translate the awaiting-input status in the remaining 19 locales

p-poppe:
## Summary

Adds `workspace.agentAwaitingInput` to the 19 locales that lacked it. Per the triage on #997 ("the locale gap is a straightforward translation pass"), this is that pass.

Before: the key existed in `en`, `ko` and `zh` only, so in every other language the status rendered as English `Needs input` inside an otherwise translated sidebar — the one status that means *stop and look at me* being the one that looked foreign.

`pl` is deliberately absent here: it lands in #996, which also makes the Polish status set bare.

## Two rules I followed

**Each locale's own convention.** Every file that lacked the key prefixes the subject — `Agent wartet`, `Agente en espera`, `エージェント待機中` — so these do too. I did not import the bare-state style from `en`/`ko` through the back door; that is the separate design question still open on #997.

**The distinction from `agentWaiting`.** This is the part machine translation gets wrong: `agentWaiting` is "idle, ready for input", `agentAwaitingInput` is "it asked you something and stopped". Several languages render both as some form of *waiting*, so where the language allowed it the wording names the **input** rather than the waiting (`Agent wartet auf Eingabe`, `Агент ждёт ввода`, `Aracı girdi bekliyor`).

## Provenance, stated plainly

These are my translations, not a native speaker's. Dropping any string is better than shipping a guess, and four deserve a second pair of eyes before they ship:

- **`ja`** — `入力待ち` is idiomatic but sits close to the existing `待機中` for `agentWaiting`; a Japanese speaker may want them further apart.
- **`fr`** — chose *une réponse* over *une saisie*, since the blocking case is normally a question rather than a text field.
- **`th`** — long even by this column's standards; a shorter form likely exists.
- **`ar`** — RTL, and I cannot check how it renders beside the Latin pane coordinate (`w2-127`) in the same row.

### Fixed

- **"Needs input" is translated again.** The status that means an agent is blocked waiting for you was missing from every locale but English, Korean and Chinese, so in the other nineteen it rendered in English inside an otherwise translated sidebar — the one status a user has to act on being the one that looked out of place.

## Stability tier impact

- [x] No external surface touched — locale files only.

n/a.

## Test plan

- Mechanical insertion (script), then verified: all 22 non-English locale files now carry the key exactly once, `pl` excepted as noted; `npx tsc --noEmit` clean.
- No new tests: this is data, and the existing type on the translation map is what pins the key name.
- Manual verification: **not done** — no Electron binary in this sandbox, so nothing here has been seen rendered. The narrow column is the risk (`th`, `ar`), not correctness of the key.

## Summary by CodeRabbit

* **Bug Fixes**
  * Translated the “Needs input” agent status across all supported locales.
  * Improved consistency of workspace status messages for international users.

* **Documentation**
  * Added a changelog entry describing the localization fix.


openwong2kim:

Adds the missing  translation to the 19 locales that lacked it, closing the gap flagged in #997. Locale data files only, no logic touched, with each string following its locale's existing label convention. P3 — cosmetic fallback fix, but worth merging to make the one actionable status readable everywhere.

PR #1003, opened 2026-08-22
fix(updater): reuse a verified installer already in temp instead of re-downloading it

### LINKED ISSUE #995: Updater re-downloads a verified installer it already has after an aborted install
Split out of #980 (ask 4), deliberately not fixed in #986: with the install path repaired, there is no repeating failure cycle multiplying this cost — what remains is one wasted download per genuinely aborted install, which is real but a different subsystem than process teardown.

## What happens

`downloadedPath` is in-memory only. After a quit-for-install (successful or aborted), the next launch has no record that a verified installer already exists, so:

- `sweepStaleArtifacts()` (`AutoUpdater.ts:283`) deletes any `wmux-update-*` temp artifact older than `TEMP_ARTIFACT_MAX_AGE_MS` (24 h) on startup — including one whose version still matches the pending update.
- Even when the artifact survives the sweep (younger than 24 h), `downloadUpdate()` only short-circuits on the in-memory `this.downloadedPath` (`AutoUpdater.ts:398`), which a restart clears — so the next check re-downloads the full ~150 MB Setup.exe regardless.

Field data from #980: three swept artifacts across the failure loop (`3.42.0`, `3.44.0`, and the `3.45.0` re-download), ~150 MB each.

## Shape of a fix

On boot (or on the next check), before downloading: if a `wmux-update-<version>-*.Setup.exe` matching the *current pending version* exists in temp, re-verify its SHA-256 against the freshly fetched manifest and adopt it as `downloadedPath` instead of downloading. Two details that matter:

1. **Re-verify, never trust.** The file sat in a world-writable temp dir across a restart; the sha256 check (`verifyUpdate.ts` digests) must run again on adoption. A mismatch deletes and falls through to a normal download.
2. **The sweep must not race the adoption.** Either adopt before sweeping, or exempt artifacts whose embedded version equals the pending release from the age-based sweep. The current name format (`wmux-update-<version>-<pid>-...`) already carries the version, so the exemption is a name parse, not new state.

No persisted state file needed — the artifact itself plus the manifest hash is the record.

## Non-goals

- Resuming partial downloads (different problem, different machinery).
- macOS: the ZIP is deleted on successful staging already and the failure loop this cost rode on was Windows-specific.

openwong2kim:
Closes #995.

## The loop

`downloadedPath` is in-memory only. After a quit-for-install — successful or aborted — the next launch has no record that a verified installer already exists, so a ~150 MB Setup.exe gets fetched again even though the identical, already-verified bytes are sitting in temp. The startup sweep made it worse by deleting that artifact on the 24 h floor. #980's field data shows three such downloads across one failure loop.

## What changed

The artifact's file name already carries its version, so the file plus the freshly fetched manifest hash is the whole record — no new state file.

- **Adopt before downloading.** `downloadUpdate()` scans temp for an artifact whose version and file name match the pending manifest and re-hashes it. A SHA-256 match becomes `downloadedPath` and the download never runs; the manifest fetch still happens (that is where the hash comes from).
- **Re-verify, never trust the name.** The file sat in a world-writable temp dir across a restart, so it goes through the same digest gate a fresh download does. A mismatch is discarded and a normal download follows.
- **Do not delete another instance's download.** A mismatching artifact written in the last minute is left alone — that is an in-flight download, not garbage. Only a quiet one is unlinked.
- **The sweep stops eating the pending installer.** An artifact for a version *newer* than the running app — the only kind adoption can reuse — is kept for 7 days instead of 24 h. Everything else keeps the old floor, and the longer bound still expires (with auto-update off, the poll that would consume it never runs, and nothing should sit in temp forever).

Naming moved next to parsing in `verifyUpdate.ts` (`artifactTempName` / `parseArtifactName`, both pure and unit-tested), so the writer and the reader of that name cannot drift apart.

## Non-goals (per the issue)

Partial-download resume, and macOS — the ZIP is deleted on successful staging already and this failure loop was Windows-specific. The change is platform-neutral, though: adoption keys off the manifest's own file name.

## Tests

`src/main/updater/__tests__/AutoUpdater.download.test.ts` grows a fake temp dir (readdir/stat/createReadStream) and four cases: reuse without downloading, tampered artifact discarded then downloaded, just-written mismatch left alone, and the sweep keeping the newer-version artifact while dropping the running-version one. `verifyUpdate.test.ts` pins the name round-trip (including a prerelease version, whose hyphens must not be read as the pid) and `isVersionNewer`.

Verified: `tsc --noEmit`, `eslint src/main/updater` (no new warnings), `vitest run src/main/updater` — 104 passed. Removing the adoption call and the sweep exemption fails three of the four new tests, so they are real gates.

### DIFF
--- a/changelog.d/1003.md
+++ b/changelog.d/1003.md
@@ -5,4 +5,11 @@
   that aborted — forgot about the ~150 MB Setup.exe sitting in temp, and the
   cleanup sweep deleted it a day later. wmux now recognizes an installer it
   already has, re-checks its SHA-256 against the current manifest before
-  trusting it, and keeps one for a version you have not installed yet.
+  trusting it, and keeps the installer for the version you have not taken yet.
+
+- **An installer is checked again in the moment it is run.** Verification used
+  to happen when an update was found, and the install could be days later —
+  long enough for the file in the shared temp directory to become a different
+  file. wmux now re-checks the bytes against the release digest immediately
+  before handing them to the installer, and refuses to run them if they do not
+  match.
--- a/changelog.d/1003.md
+++ b/changelog.d/1003.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- **An update you already downloaded is no longer downloaded again.** The
+  verified installer's path lived only in memory, so a restart — or an install
+  that aborted — forgot about the ~150 MB Setup.exe sitting in temp, and the
+  cleanup sweep deleted it a day later. wmux now recognizes an installer it
+  already has, re-checks its SHA-256 against the current manifest before
+  trusting it, and keeps one for a version you have not installed yet.
--- a/src/main/updater/AutoUpdater.ts
+++ b/src/main/updater/AutoUpdater.ts
@@ -13,7 +13,7 @@
 
 import { autoUpdater, app, type BrowserWindow, ipcMain, net, shell } from 'electron';
 import { createReadStream, createWriteStream } from 'node:fs';
-import { readdir, stat, unlink } from 'node:fs/promises';
+import { lstat, readdir, stat, unlink } from 'node:fs/promises';
 import { join, dirname, resolve } from 'node:path';
 import { createHash } from 'node:crypto';
 import { IPC } from '../../shared/constants';
@@ -93,10 +93,12 @@ const TEMP_ARTIFACT_MAX_AGE_MS = 24 * 60 * 60 * 1000;
 // It still expires, just far later: if auto-update is off the poll that would
 // consume it never runs, and nothing should sit in temp forever.
 const TEMP_ARTIFACT_PENDING_MAX_AGE_MS = 7 * 24 * 60 * 60 * 1000;
-// An artifact whose bytes fail the hash is either a corrupted leftover (delete)
-// or another instance's download in flight (leave alone). Recent writes mean
-// the second, so only a file that has been quiet this long is removed.
-const TEMP_ARTIFACT_QUIET_MS = 60 * 1000;
+// An artifact whose bytes fail the hash is either a corrupted leftover or
+// another instance's download in flight. Only age tells them apart, and the
+// age that means "nobody is writing this" is the sweep's own floor — a shorter
+// one would delete a slow download out from under the instance that owns it,
+// which on POSIX leaves that instance verifying a file whose path is gone.
+const TEMP_ARTIFACT_MISMATCH_DELETE_AFTER_MS = TEMP_ARTIFACT_MAX_AGE_MS;
 
 // In-app auto-update runs on Windows (Squirrel.Windows `.Setup.exe`) and on
 // Apple Silicon macOS (Squirrel.Mac, signed+notarized ZIP). Everything else —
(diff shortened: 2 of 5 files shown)


openwong2kim:

This reuses an already-verified installer parked in temp instead of re-downloading ~150 MB, with SHA-256 re-verified against the fresh manifest before adoption. It touches the AutoUpdater download path and the startup sweep; the fail-closed hash gate and the quiet-window guard for in-flight downloads keep the risk low. P3 — matches #995, well covered by new unit tests.

openwong2kim:
## Panel review + live dogfood

Reviewed by three models independently (Claude / Codex / GLM 5.2), then every accepted finding was re-checked against a real 3.45.0 artifact on this machine rather than only in tests.

**Taken (pushed in 47dffff0):**

| # | Raised by | Finding |
| --- | --- | --- |
| 1 | Claude (CRITICAL) | The 7-day exemption had no bound — one ~150 MB installer retained per release, and only the newest can ever be installed. The long floor now goes to the newest pending version only. |
| 2 | Codex (CRITICAL) + GLM + Claude | Verification and execution were days apart with only a path in between. The accepted digest is kept and re-checked immediately before the handoff. |
| 3 | GLM (CRITICAL) + Claude | A mismatching artifact was deleted after 60 s — short enough to be another instance's slow download, whose owner would then verify a path that no longer exists. Deletion now waits for the sweep's own 24 h floor. |
| 4 | Claude | Adoption followed symlinks (a link can be re-pointed atomically after the hash passes). Only regular files are candidates — `lstat`, not `stat`. |
| 5 | Claude + GLM | `isVersionNewer` required a bare X.Y.Z on **both** sides, so a build whose own version carries a suffix switched the whole mechanism off. |
| 6 | Claude + GLM | `parseArtifactName` matched the version as "anything": `3.46.0-1-12345-…` read as version 3.46.0 / pid 1. Now semver with a greedy prerelease. |
| 7 | Claude | `artifactTempName` sanitized the file name but not the version, which arrives over the network and is joined onto the temp path. |

**Not taken:** hashing off the main thread (Claude, conf 4). The read is chunked through a stream, so the loop interleaves rather than blocks — adopting a 117 MB artifact measured under a second end to end.

**Live dogfood** — real `AutoUpdater`, real update feed and manifest, real files in the OS temp dir (install path deliberately not run):

```
version = 3.42.0
[AutoUpdater] swept stale update artifact wmux-update-3.43.0-…      <- superseded pending, 3d old
[AutoUpdater] swept stale update artifact wmux-update-3.44.0-…      <- superseded pending, 3d old
[AutoUpdater] reusing verified installer already in temp (wmux-update-3.45.0-94526-…) — skipping download
re-verify (untouched) = true
[AutoUpdater] refusing install: staged installer no longer matches its verified digest
re-verify (bytes swapped under it) = false
re-verify (restored) = true
```

Worth noting where that adopted file came from: it was **already on this machine** — 117 MB left in temp by wmux itself on 21 Aug, verified and never installed. That is the waste this PR is about, found by the PR.

Negative control (tampered copy parked, everything else removed): `discarded temp artifact …: sha256 does not match the manifest`, followed by a real download starting — aborted at the first progress event.

110 passed / 8 skipped in `src/main/updater`; removing the re-verification or the `isFile` guard fails the new tests.

PR #1004, opened 2026-08-22
test(mcp): drop the weaker duplicate externalWire source-invariant scan

### LINKED ISSUE #958: test(mcp): drop the weaker duplicate externalWire source-invariant scan
Follow-up to #950, which deduplicated the local external-wire predicate but left a second, weaker duplicate behind — this time in the AST invariant, not the predicate.

## The duplicate

`src/main/mcp/__tests__/rpcProvenance.test.ts:72` ("keeps PipeServer as the only production writer of externalWire authority") scans `src/main` and asserts:

```ts
expect(writers).toEqual(['pipe/PipeServer.ts']);
```

`src/main/pipe/__tests__/RpcDispatchProvenance.sourceInvariant.test.ts` already pins the same invariant, and asserts:

```ts
expect(markerWriters('externalWire')).toEqual([
  'pipe/PipeServer.ts',
  'pipe/RpcRouter.ts',
]);
```

Two tests, same invariant, different expected sets. Only one of them is right.

## Why the weaker one passes

`writesExternalWireTrue` (`rpcProvenance.test.ts:23`) only matches a `PropertyAssignment` whose initializer is literally `TrueKeyword` (`:36`). The real second writer, `src/main/pipe/RpcRouter.ts:292`

```ts
externalWire: externalWire ? true : undefined,
```

is a `ConditionalExpression`, so the matcher walks straight past it. The assertion passes, but it does not mean what its test name claims — it means "PipeServer is the only file that writes a bare `true` literal." A future transport could widen the boundary with a ternary, `as const`, a shorthand, or any computed boolean and this test would stay green.

`RpcDispatchProvenance.sourceInvariant.test.ts` matches every `PropertyAssignment` plus `ShorthandPropertyAssignment` regardless of initializer, which is why it correctly lists both writers, and why it is the one to keep.

## Not a live security hole

The strong invariant is already in place and enforced, so the boundary is covered today. The cost is a misleading test name and a second AST scan that will disagree with the real one the next time a writer is added — exactly the failure mode #950 set out to remove.

## Proposed fix

- Delete the `writesExternalWireTrue` helper and the `it('keeps PipeServer as the only production writer...')` case from `src/main/mcp/__tests__/rpcProvenance.test.ts`, leaving it as a focused unit test of `isLocalExternalWireContext`.
- `RpcDispatchProvenance.sourceInvariant.test.ts` stays the single source-invariant pin.

No production code changes.

openwong2kim:
Closes #958. Follow-up to #950, which deduplicated the predicate but left a second, weaker duplicate in the AST invariant.

Two tests pinned "who may write `externalWire`", with different expected sets:

- `src/main/mcp/__tests__/rpcProvenance.test.ts` asserted `['pipe/PipeServer.ts']`, but its matcher only accepted a `TrueKeyword` initializer — so `RpcRouter.ts:292`'s `externalWire: externalWire ? true : undefined` was invisible to it. It passed while meaning "PipeServer is the only file writing a bare `true` literal", not what its name claimed.
- `src/main/pipe/__tests__/RpcDispatchProvenance.sourceInvariant.test.ts` matches every property assignment plus shorthand regardless of initializer, correctly lists both writers, and is the one that would catch a transport widening the boundary with a ternary, `as const`, or a computed boolean.

This deletes the weaker scan and its helper, leaving `rpcProvenance.test.ts` as a focused unit test of `isLocalExternalWireContext` with a pointer to where the source invariant lives. The strong invariant is untouched, so the boundary stays covered — no production code changes.

Verified: both suites pass (`vitest run src/main/mcp/__tests__/rpcProvenance.test.ts src/main/pipe/__tests__/RpcDispatchProvenance.sourceInvariant.test.ts`).

## Summary by CodeRabbit

* **Tests**
  * Refined coverage for detecting local external wire contexts.
  * Moved enforcement of external wire-writing rules to a separate test.

### DIFF
--- a/src/main/mcp/__tests__/rpcProvenance.test.ts
+++ b/src/main/mcp/__tests__/rpcProvenance.test.ts
@@ -1,7 +1,15 @@
-// The source-level invariant — which production files may write the
-// externalWire marker at all — is pinned by
-// src/main/pipe/__tests__/RpcDispatchProvenance.sourceInvariant.test.ts. This
-// file stays a unit test of the predicate itself (#958).
+// Two other tests hold the externalWire boundary, and they hold different
+// halves of it (#958):
+//   • WHICH production files may write the marker at all —
+//     src/main/pipe/__tests__/RpcDispatchProvenance.sourceInvariant.test.ts,
+//     which pins the file set (PipeServer + RpcRouter) over every property
+//     assignment and shorthand, whatever the initializer.
+//   • THAT RpcRouter's write stays conditional — RpcRouter.dispatchProvenance
+//     .test.ts ('does not inherit external-wire provenance into an unmarked
+//     nested dispatch'), which is the test that fails if that ternary is ever
+//     widened to a bare `true`. The source invariant would not: the file set
+//     is unchanged by that edit.
+// This file stays a unit test of the predicate itself.
 
 import { describe, expect, it } from 'vitest';
 import { isLocalExternalWireContext } from '../rpcProvenance';
--- a/src/main/mcp/__tests__/rpcProvenance.test.ts
+++ b/src/main/mcp/__tests__/rpcProvenance.test.ts
@@ -1,48 +1,10 @@
-import * as fs from 'fs';
-import * as path from 'path';
-import * as ts from 'typescript';
+// The source-level invariant — which production files may write the
+// externalWire marker at all — is pinned by
+// src/main/pipe/__tests__/RpcDispatchProvenance.sourceInvariant.test.ts. This
+// file stays a unit test of the predicate itself (#958).
+
 import { describe, expect, it } from 'vitest';
 import { isLocalExternalWireContext } from '../rpcProvenance';
-
-const MAIN_DIR = path.resolve(__dirname, '..', '..');
-
-function collectProductionTsFiles(dir: string): string[] {
-  const out: string[] = [];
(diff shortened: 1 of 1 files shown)


openwong2kim:

This drops the weaker duplicate externalWire scan from rpcProvenance.test.ts — its TrueKeyword-only matcher missed RpcRouter's ternary writer — leaving RpcDispatchProvenance as the single source-invariant pin. Test-only with no production changes, so risk is minimal. P3 — cleanup per #958.

openwong2kim:
## Panel review

Codex: `NO FINDINGS`. GLM 5.2: `NO FINDINGS`. Claude: one — and a correct one.

The header comment claimed the remaining source invariant pins what the deleted scan pinned. It does not: `RpcDispatchProvenance.sourceInvariant` compares the *file set*, so widening `RpcRouter`'s `externalWire ? true : undefined` to a bare `true` leaves it green.

Verified by mutating exactly that line:

```
RpcDispatchProvenance.sourceInvariant.test.ts   2 passed     <- file set unchanged
RpcRouter.dispatchProvenance.test.ts            3 failed     <- incl. 'does not inherit external-wire
                                                                provenance into an unmarked nested dispatch'
```

So coverage of that edit is preserved — by the behavior tests, not by the source invariant. The comment now names both halves (5cc4b395), so weakening the behavior test later reads as removing the value-level guard rather than as routine cleanup.

PR #1005, opened 2026-08-22
fix(shortcuts): distinguish "nothing to repair" from "the repair pass never ran"

### LINKED ISSUE #962: Flaky: shortcutHygiene.runtime 'retargets a dead versioned pin' — repairInstalledShortcuts returned [] once on CI
Observed on PR #959's `validate` job ([run 32383089890](https://github.com/openwong2kim/wmux/actions/runs/32383089890/job/96470835440)): 

```
FAIL src/main/__tests__/shortcutHygiene.runtime.test.ts
  > shortcutHygiene end-to-end (real .lnk, real PowerShell)
  > retargets a dead versioned pin to the stub and pins its icon to app.ico
AssertionError: expected [] to deeply equal [ { path: …wmux-pin.lnk, action: 'retargeted' } ]
```

`repairInstalledShortcuts` returned an empty action list. The PR's diff touches only pipe handlers (`events.rpc` / `a2a.channel.rpc` / `shared/rpc`) — nothing near shortcuts — and an unmodified re-run of the same job went green, so this is a flake, not a regression.

The test drives real PowerShell against a real `.lnk` in `%TEMP%` (the suite took 26.8 s), so the suspects are the usual runtime-test ones: the WScript.Shell COM call racing the file's creation, PowerShell startup returning before the pin is visible, or antivirus briefly holding the `.lnk`. Worth either a bounded retry around the COM read, an existence/settle wait on the pin before calling `repairInstalledShortcuts`, or capturing the PowerShell stderr into the assertion message so the next failure says why the list was empty instead of just `[]`.

One observation ≠ a rate, but this one cost a full validate re-run on an unrelated security PR; filing it now so the next hit has somewhere to accumulate evidence.

openwong2kim:
Closes #962.

## What the flake was hiding

`shortcutHygiene.runtime.test.ts` failed once on CI with `expected [] to deeply equal [ { path: …wmux-pin.lnk, action: 'retargeted' } ]`, on a PR that touched only pipe handlers, and went green on re-run. The empty list was the whole evidence — and it could not be more, because `repairInstalledShortcuts` returns `[]` for two different outcomes:

- nothing needed repair, and
- the pass never ran.

The script sets `$ErrorActionPreference = 'SilentlyContinue'`, so a refused `New-Object -ComObject WScript.Shell` — a class-factory request lost to load or to antivirus, the usual suspect on a busy runner — leaves `$sh` null, every `$sh.CreateShortcut($p)` fails quietly, every candidate is skipped, and the script prints a clean `[]` with exit code 0.

That is worse in production than in CI: a real install where COM blinks keeps its blank taskbar icon (the whole #863 symptom) and logs nothing at all.

## Changes

- **The script fails loudly.** `if (-not $sh) { [Console]::Error.WriteLine(…); exit 3 }` — a pass that could not start no longer looks like a pass that found nothing.
- **`runShortcutRepairPass(execPath, locations)`** returns `{ actions, failure }` and `console.warn`s the failure (powershell's exit status and stderr). `repairInstalledShortcuts` is now a thin wrapper over it: same signature, same never-throws contract, same call sites in `index.ts` untouched.
- **One bounded retry on the COM failure specifically.** It is transient by nature, and a fresh powershell start is its own backoff. Every other failure is reported as-is — retrying a real problem just doubles its cost.
- **The runtime suite routes every call through a helper that throws the failure**, and `makeLnk` now asserts the `.lnk` actually landed. The next hit names the layer that gave out instead of showing an empty list.

## Tests

Two unit cases in `shortcutHygiene.test.ts`: the built script carries the COM guard and its exit code, and the pass reports no failure off win32 (nothing to repair is not a broken pass). The runtime suite is `describe.skipIf(!onWindows)`, so its side of this only runs on the Windows CI leg.

Verified locally (macOS): `tsc --noEmit`, `eslint` clean, `vitest run src/main/__tests__/shortcutHygiene*` — 18 passed, 11 skipped (the Windows-only runtime suite).

### DIFF
--- a/changelog.d/1005.md
+++ b/changelog.d/1005.md
@@ -3,5 +3,7 @@
 - **A shortcut repair that could not run no longer reports success.** If
   Windows refused the COM object the icon-repair pass used, the pass quietly
   reported "nothing needed fixing" — so a taskbar icon that stayed blank after
-  an update left no trace of why. The pass now says it failed, retries once,
-  and logs what Windows returned.
+  an update left no trace of why. The pass now says it failed — whether COM
+  was refused outright or not one shortcut could be opened — retries a refused
+  COM object once inside the same time budget, and writes what Windows
+  returned to the install log.
--- a/changelog.d/1005.md
+++ b/changelog.d/1005.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **A shortcut repair that could not run no longer reports success.** If
+  Windows refused the COM object the icon-repair pass used, the pass quietly
+  reported "nothing needed fixing" — so a taskbar icon that stayed blank after
+  an update left no trace of why. The pass now says it failed, retries once,
+  and logs what Windows returned.
--- a/src/main/__tests__/shortcutHygiene.runtime.test.ts
+++ b/src/main/__tests__/shortcutHygiene.runtime.test.ts
@@ -10,7 +10,12 @@ import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import { execFileSync } from 'node:child_process';
-import { repairInstalledShortcuts, stageRootIcon, type RepairLocations } from '../shortcutHygiene';
+import {
+  runShortcutRepairPass,
+  stageRootIcon,
+  type RepairLocations,
+  type ShortcutRepairAction,
+} from '../shortcutHygiene';
 
 const onWindows = process.platform === 'win32';
 
@@ -45,6 +50,11 @@ function makeLnk(lnkPath: string, target: string): void {
     `$l.TargetPath = ${q(target)}`,
     `$l.Save()`,
   ].join('\n'));
+  // Save() is synchronous, but this suite's whole job is to catch the cases
+  // where the shell layer disagrees with the filesystem. If the link is not
+  // there, say THAT rather than letting the pass report "nothing to repair"
+  // three lines later (#962).
+  if (!fs.existsSync(lnkPath)) throw new Error(`.lnk was not written: ${lnkPath}`);
 }
 
 function readLnk(lnkPath: string): { target: string; icon: string; workdir: string } {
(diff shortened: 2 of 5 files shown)


openwong2kim:
This PR closes #962 by making the shortcut-repair pass distinguish "nothing to repair" from "the pass never ran": a refused WScript.Shell COM object now exits 3 and gets one bounded retry. The old `repairInstalledShortcuts` signature and never-throws contract are kept, so risk is confined to the win32 hygiene path. P3 — diagnostics for a best-effort pass; behavior unchanged when it succeeds.

openwong2kim:
## Panel review + live dogfood

Three independent models (Claude / Codex / GLM 5.2). Ten findings between them, all on the same theme: the first cut made *one* silent failure loud and left the others silent. Every one below was reproduced against a real spawned process before and after the fix (pushed in 6161c740).

| Raised by | Finding |
| --- | --- |
| Claude | The retry keyed on exit code 3 alone — a runtime abort also exits 3, and re-running mid-loop repeats work already applied to real `.lnk` files. The script now prints `WMUX_COM_UNAVAILABLE` and only a *marked* exit 3 retries. |
| Claude | `[Console]::Error.WriteLine` is a blocked method call under ConstrainedLanguage (WDAC/AppLocker) — the very environment most likely to refuse the COM object. The marker goes to stdout via `Write-Output`. |
| Claude | The guard only covered COM being unavailable *at all*. A per-link failure (locked/corrupt `.lnk`, WSH policy, a COM server dying mid-loop) still skipped every candidate and emitted `[]`. The script counts opened vs unopened and exits 4 when candidates existed and none opened. |
| Codex | Exit 0 with empty or non-JSON output parsed as "nothing to repair". The last statement is always `ConvertTo-Json`, so that state means it died earlier — it is a failure now. |
| Claude | The failure string carried `err.message`, i.e. the whole command line: kilobytes of PowerShell plus the user's profile paths, into a log line and every test failure. Now `status`/`signal`/`code`, which also separates a timeout from a missing interpreter. |
| Claude | The retry doubled the time budget inside a Squirrel hook that Squirrel cancels at ~15 s. Both attempts now share one 15 s deadline. |
| Codex + GLM + Claude | Two early returns (APPDATA unset, unsafe path) returned silently — the same invisible empty list this PR exists to remove. Every failure exit logs. |
| GLM | **The diagnostics went nowhere in a packaged build**: the hook path is the only production caller, it has no console, and `initLogSink()` runs far later in boot. It is now initialised at the top of the Squirrel branch (idempotent), so install-time output lands in `%APPDATA%\wmux\logs`. |

**Live dogfood** — real `execFileSync` against a stand-in PowerShell (the script itself is Windows-only, so the interpreter is what gets faked; the Node side is real):

```
com-fail    (marker + exit 3) -> failure "the WScript.Shell COM object was refused"      spawns=2   <- retried
abort-exit3 (exit 3, no marker) -> failure "…status=3 signal=none code=none"             spawns=1   <- not retried
unreadable  (marker + exit 4) -> failure "no candidate shortcut could be opened"         spawns=1
ok                            -> 1 action parsed                                          spawns=1
empty       (exit 0, no output) -> failure "the script produced no JSON [empty output]"   spawns=1
garbage     (exit 0, "Access is denied.") -> failure "the script produced no JSON […]"    spawns=1
slow        (hangs) -> failure "…signal=SIGTERM code=ETIMEDOUT" after 15005ms             spawns=1
```

`empty` is the case Codex predicted: before this change it returned `{actions: [], failure: null}` — a pass that never ran, reported as a clean one.

The Windows CI leg runs the real `.lnk` / real PowerShell suite in `validate`.

PR #1006, opened 2026-08-22
chore(release): 3.46.0

openwong2kim:
Release 3.46.0. MINOR — the fold carries `### Added` entries (`--signals-only`, the Matrix palette).

Mechanical, per the documented procedure:

- `node scripts/collect-changelog.mjs` folded 11 fragments into `CHANGELOG.md` (the `[Unreleased]` heading is re-added at release time, then renamed).
- `[Unreleased]` → `[3.46.0] — 2026-08-22`.
- `package.json` + both `package-lock.json` mirrors: 3.45.0 → 3.46.0.
- `node scripts/gen-api-reference.mjs` — the generated header bakes the version, which the CI drift guard enforces.

## What ships

**Added** — `wmux setup-hooks --signals-only` (lifecycle signals without the per-tool-call permission gate, ~120 ms per tool call saved on Windows), `wmux web --allow-input` now reports a missing permission-gate hook, and a `Matrix` terminal palette.

**Changed** — the agent roster spends its width on the name rather than repeating one vendor per row.

**Fixed** — the in-app updater actually installs again (#980) and no longer re-downloads an installer it already has (#995), with the artifact re-verified in the moment it is run; a self-started agent turn reports as running (#935); terminal colour overrides survive a restart (#981); "Needs input" is translated in the remaining 19 locales (#997); "Don't ask again" sticks on the hook-install prompt (#968); pane overlays stay under app chrome; a shortcut repair that could not run no longer reports success (#962).

## Verification

`tsc --noEmit` clean; `npm run test:parallel` 11892 passed with two failures that are local-environment only and reproduce on an untouched checkout: `atlasCoherence` (the xterm addon patch is not applied in this machine's `node_modules`) and the `THIRD_PARTY_NOTICES` drift guard (this tree predates the last dependency bumps, so regenerating would wrongly *drop* `content-type@2.1.0` — reverted, notices unchanged). Both are installed-tree state, and CI installs from the lock.

After merge: push `v3.46.0` — the installer builds hang off the tag.

## Summary by CodeRabbit

* **New Features**
  * Added signals-only hook installation with clearer missing-gate diagnostics and persistent refusal settings.
  * Added the Matrix terminal palette and improved color override persistence.
  * Improved agent activity detection, roster displays, status labels, pane stacking, and localization.
* **Bug Fixes**
  * Improved updater, installer, quit handling, cached installer verification, and shortcut-repair diagnostics.
  * Improved timing reliability for installation waiters.
* **Documentation**
  * Published release notes and updated the API reference for version 3.46.0.


openwong2kim:
Mechanical 3.46.0 release fold: 11 changelog fragments into CHANGELOG.md, version bumps in package.json and both lockfile mirrors, regenerated API reference. No code paths touched beyond docs and metadata, so regression risk is minimal. P3 — routine release choreography.

PR #1013, opened 2026-08-24
fix(status): self-heal a pane stuck at running after its turn ended (#935 direction 3)

p-poppe:
Closes the last direction of #935 — #939 and #965 closed the other two.

## The bug

`onIdle` (daemon mode) / `onActiveToIdle` (local mode) is the only thing that clears a stale `running` back to `idle` when a pane goes quiet with no precise event to explain it. It defers to a precise `waiting`/`complete` status set in the last `AGENT_EVENT_SUPPRESSION_MS` (10s), on the assumption that status is still correct and the clear would just be redundant.

But `onActive` broadcasts `running` unconditionally on any burst over the byte threshold — no deference to a recent precise status at all. So the sequence that wedges a pane is:

1. Turn ends, Stop hook fires, `agentStatus: 'complete'` — correct.
2. A short burst inside the 10s window (a final chrome repaint, a keystroke echo — not a new turn) fires `onActive`, which overwrites `complete` with `running`.
3. 5s later `onActiveToIdle`/`onIdle` should clear `running` → `idle`, but it's still inside the 10s window from step 1, so it defers — to a status that no longer exists.
4. `ActivityMonitor` already consumed its active→idle transition, and a quiet pane never bursts again, so nothing ever retries the clear. `running` is now permanent.

Same shape, same constant, in both `PTYBridge.ts` (local mode) and `DaemonNotificationRouter.ts` (daemon mode, the packaged build's path) — the wedge isn't specific to either.

(This is the case the maintainer named directly in #965's "Known limits": *"the clearing idle can still be swallowed if a lifecycle event lands between the promotion and the idle... the general one belongs in the router."*)

## The fix

Track the last `agentStatus` actually broadcast per PTY (`lastBroadcastStatus`). The suppression window now only applies while that tracked status is still the precise one — once it's `running` (meaning a burst already clobbered whatever was there), there's nothing left to defer to, and the clear goes through regardless of the window.

A withheld status (hook-governed pane, `#935` direction 1's veto) does not update the tracker — nothing landed that a later clear would need to defer to.

No change to `ActivityMonitor`, the terminal-wins ordering rule, or the resize/typing guards — this is scoped entirely to the suppression check in the notification layer, as the maintainer's own note suggested.

## Test plan

- New regression test in both modes: precise `complete` → burst re-fires `running` inside the suppression window → byte silence must still clear to `idle`. Verified both fail on the pre-fix code (reverted the fix locally, confirmed red) and pass with it.
  - `src/main/pty/__tests__/PTYBridge.staleRunning.test.ts` (new file, local mode)
  - `src/main/notification/__tests__/DaemonNotificationRouter.statusClear.test.ts` (added case, daemon mode)
- Existing #733 protection (`defers to a recent precise agent status instead of overwriting it`) still passes unchanged — a precise status that was never clobbered still gets deference.
- Full suite: `npm run test:parallel` — 11,509 passing, 9 pre-existing failures unrelated to this change (timing races in `deck.handler.loop.test.ts`, a git-checkout timeout in `worktree.handler.test.ts`, one `state.fallback.test.ts` case) — none touch `PTYBridge`, `DaemonNotificationRouter`, or anything in this diff.
- `tsc --noEmit` clean.

## Stability tier impact

None — no new public API, no change to `AgentStatus`, no schema/RPC shape change. Purely internal bookkeeping in the notification layer.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed stale **Running** statuses that could remain after activity ended.
  * Agent and terminal status now correctly return to **Idle** after inactivity, including when a recent **Complete** or **Waiting** status was temporarily overwritten.
  * Preserved precise statuses when they remain active, avoiding premature idle updates.

* **Tests**
  * Added regression coverage for status recovery after completion and short activity bursts.

* **Documentation**
  * Added a changelog entry for the status correction.


openwong2kim:

This PR stops a pane from wedging at `running` forever after its turn ended — the last open direction of #935. It tracks the last broadcast status in `PTYBridge` and `DaemonNotificationRouter` so the idle clear no longer defers to a clobbered status; internal-only, with regression tests in both modes. P2: completes the #935 roster-correctness bug with low-risk, well-scoped changes.

openwong2kim (changes requested):
Thank you for this — the wedge is real, the diagnosis matches what #965's known-limits note predicted, the both-modes coverage is exactly right, and "track what actually went out" is the correct instrument. I want to land this. Two things need to change first, both verified by execution rather than by reading, so the chains are below.

## 1. The tracker misses the hook-authority broadcast sites — and that wipes the completion status on governed panes (must fix)

`lastBroadcastStatus`'s declared invariant is *the last `agentStatus` actually broadcast per PTY*. Two broadcast sites live outside the two files this PR instruments, both in `hooks.rpc.ts`: the turn-boundary `agentStatus: 'complete'` (`buildTurnBoundaryMetadata`, broadcast around line 658) and the `awaiting_input` broadcast (line 775). On a hook-governed pane those are not corner cases — they are the *primary* path for the two statuses the suppression window exists to protect.

The chain, on a local-mode hook-governed pane at turn end:

1. The turn's output bursts fire `onActive` → tracker = `'running'`. (This site is unconditional — no hook-authority veto, `PTYBridge.ts:640`.)
2. The detector's end-of-turn `complete` is **withheld** by hook authority (#935 direction 1) — it arms `lastAgentEventAt` (the set at `PTYBridge.ts:490` is outside the withhold), but per this PR's own withhold rule it does **not** touch the tracker.
3. The Stop hook broadcasts `agentStatus: 'complete'` via `hooks.rpc` — invisible to the tracker, which still says `'running'`.
4. Five seconds of byte silence later, `onActiveToIdle` runs: `stillPrecise === false`, the armed window is ignored, and the clear **wipes the just-set `complete` to `idle`**. Same shape for `awaiting_input` — a pane asking the user something loses its indicator five seconds after going quiet.

Before this PR, step 4 was swallowed by the window armed in step 2 and never retried, so the hook status survived. Executed both ways with the identical input sequence (gate match → 3000-byte burst → withheld detector `complete` → 5.1s silence), swapping only `PTYBridge.ts`:

- **this branch:** the idle clear fires (`idleFired=true`) — the hook status is wiped
- **origin/main:** the window swallows it (`idleFired=false`) — the hook status survives

So the fix trades a permanent wedge for a deterministic 5-second loss of the completion/attention status on every governed burst turn, in local mode. (Daemon mode is safe by accident of plumbing: the arbitrated `session:agent` event writes the tracker with `ev.status` before the boundary replay at `DaemonNotificationRouter.ts:852` re-broadcasts the same value — that replay site is a third untracked broadcast, currently harmless but one refactor away from not being.)

**Suggested shape:** rather than chasing call sites, record at the single funnel — `broadcastMetadataUpdate` (metadata.handler) already sees every `agentStatus` that goes out, so updating the tracker there covers `hooks.rpc`'s two sites and the Router's boundary replay in one move, and the invariant becomes structural instead of a convention. If threading the map there is awkward, a small setter (`noteStatusBroadcast(ptyId, status)`) called from the two `hooks.rpc` sites also works; the funnel is just harder to drift from.

## 2. The "#733 still protected" test is vacuous (must fix)

The regression guard (`still defers to a precise status that has NOT been overwritten`) never exercises the guard: its total input is ~36 bytes, under `ActivityMonitor.ACTIVE_THRESHOLD` (2000), so the monitor never goes active, no idle timer is ever scheduled, and `onActiveToIdle` cannot fire regardless of the suppression logic. Verified by mutation: with the entire check disabled (`if (false && stillPrecise && ...)`), both tests in the file still pass 2/2 — the PR's only "protection preserved" evidence currently proves nothing.

A shape that actually bites: burst 3000 bytes first (active + tracker `'running'`), then the precise `complete` (tracker `'complete'`, window armed), then 5.1s of silence, then assert **no** idle broadcast. That ordering goes red when the guard is removed.

## Not requesting: distinguishing burst-`'running'` from precise-`'running'`

We considered asking for a "clobbered by burst" flag instead of the `!== 'running'` comparison (the detector's gate-match `'Agent started'` is a precise event that also writes `'running'` into the tracker) and decided your rule is the better one: the window exists to protect attention statuses, `lastAgentEventAt` never armed on `'running'` anyway, and a displayed `'running'` going idle after byte silence is precisely this handler's job — a source-split would quietly re-open a wedge (precise `'running'` within 10s of a `complete`, then eternal silence, single swallowed clear). The one nit is the justification text: "once it's `running` (meaning a burst already clobbered whatever was there)" isn't quite the mechanism, since a precise gate `'running'` lands there too — the rule stands on "`running` is never worth defending from the idle clear", and saying it that way will save the next reader the same detour.

With the funnel recording and a real deference test, this closes #935's last direction properly — the analysis in this PR is already the best writeup of this state machine's failure mode we have.

openwong2kim (changes requested):
Direction 3 is the right call for #935, but as implemented the self-heal clears *precise* statuses too — on hook-governed panes it undoes the thing the status system is best at.

**1. Hook-path broadcasts bypass `lastBroadcastStatus`, so the clear wipes `complete`/`awaiting_input` every turn (blocker)**

`lastBroadcastStatus` only records what each component itself broadcast. On panes governed by hooks, the `complete` (turn boundary) and `awaiting_input` statuses are broadcast by the hooks path directly and never touch `PTYBridge`'s map — so after any output burst the map is stuck at `onActive`'s `'running'`, `stillPrecise` comes out false, and the 5s-quiet clear overwrites the hook's precise status with `'idle'`. Net effect: the completion dot disappears ~5s after every turn, and an approval-prompt (`awaiting_input`) indicator gets hidden while the agent is actually waiting on the user. Previously these survived because the withheld detector events kept refreshing the suppression timestamp.

Fix: let the hook broadcast path record into the same map — e.g. expose a `recordBroadcastStatus(ptyId, status)` on `PTYBridge` (and the router equivalent) and call it where hooks broadcast `complete`/`awaiting_input`.

**2. The #733 no-regression test never enters an active cycle**

The second test in `PTYBridge.staleRunning.test.ts` feeds only a few dozen bytes — below ActivityMonitor's 2000-byte ACTIVE threshold — so `onActiveToIdle` can never fire and the test passes even with the deferral logic deleted. Feed a >2000-byte burst first (running → precise complete → 5s quiet) and assert no idle broadcast.

**3. Question on the intended trade-off**

Inside the 10s window, a genuinely-running new turn that goes byte-quiet for 5s (long silent tool call, response generation) now gets cleared to idle with no deferral, so roster/stopGate briefly see a false idle until the next burst. If that's the accepted cost of direction 3, a short code comment saying so would keep the next reader from "fixing" it; a cheap softener would be one deferral round when the last `'running'` broadcast is very recent.

p-poppe:
Both blockers verified by execution, not just reading — thank you for the exact chain, it made this fast to reproduce and fix.

**1. Hook-path broadcasts now feed the same tracker (blocker, fixed)**

Moved `lastBroadcastStatus` out of `PTYBridge`/`DaemonNotificationRouter` and into the one funnel every `agentStatus` broadcast already passes through — `broadcastMetadataUpdate` in `metadata.handler.ts`. Recording there instead of at each call site means `hooks.rpc.ts`'s turn-boundary `complete` and `awaiting_input` broadcasts are covered automatically, along with the boundary-replay site you flagged as "currently harmless but one refactor away from not being" (`DaemonNotificationRouter.ts:852`) — same funnel, same fix, no separate patch needed.

The withheld-status exclusion falls out for free too: a withheld status never reaches `payload.agentStatus` (the call site omits the field rather than sending it), so it never reaches the tracker either. No `if (!withheld)` needed at any call site anymore — I removed the three that existed.

One consequence worth flagging explicitly: the map is now shared and module-level (fed by both local-mode and daemon-mode broadcasts), so `DaemonNotificationRouter.stop()`'s `this.lastBroadcastStatus.clear()` would have wiped unrelated local-mode PTYs' tracked status. Removed that line — per-PTY cleanup already happens on session end via the existing delete call, so there's nothing left for `stop()` to clear.

Verified by execution: added a test that puts a hook-governed pane through the exact chain from your writeup (burst → running, detector's own complete withheld, hooks.rpc-style broadcast of complete, 5.1s silence) and confirms no idle broadcast. Ran it against a sabotaged tracker (`getLastBroadcastAgentStatus` forced to always return `undefined`, simulating the pre-funnel "separate maps" shape) to confirm it — and the *existing* burst-clobbering test — both go red without the fix, not just the new one.

**2. #733 test reshaped (blocker, fixed)**

Feeds a >2000-byte burst first (active + tracker `'running'`), then the precise `complete` (tracker `'complete'`, window armed), then 5.1s of silence, then asserts no idle broadcast — exactly the shape you suggested. Confirmed it goes red with the deferral check deleted.

**3. Trade-off: went with (a), documented in code**

Added a comment at both `stillPrecise` sites (`PTYBridge.ts`, `DaemonNotificationRouter.ts`) spelling out the accepted cost and explicitly warning against re-adding a deferral round — a genuinely-running turn that goes byte-quiet inside the 10s window will read `idle` for a moment until the next burst, and that's the trade this direction makes. Went with the plain comment over the one-round softener: any additional deferral state here is exactly the kind of "recent status protects a stale read" logic that caused this bug in the first place, and the failure mode it would reopen (a genuinely-quiet pane past the deferral matching the wedge shape) seems worse than a few-second cosmetic lag on the roster.

**Test suite**

Full `src/main/pty` + `src/main/notification` + `src/main/ipc/handlers` + `src/main/metadata`: all passing except two pre-existing timing-sensitive failures in `deck.handler.loop.test.ts` (untouched by this diff, reproduce identically against `origin/main`, self-described in their own comments as "not instant... on a loaded runner"). `tsc --noEmit` clean. ESLint clean on every file this PR touches (one pre-existing, unrelated `no-empty-function` at `DaemonNotificationRouter.lifecycle.test.ts:1131`, present on `origin/main` too).

Six other test files that mock `metadata.handler.ts` wholesale needed the two new exports added as no-op stubs to keep the import from throwing — none of them exercise the deferral logic, so a stub is the right amount of ceremony there. The two files that specifically test the deferral logic (`PTYBridge.staleRunning.test.ts`, `DaemonNotificationRouter.statusClear.test.ts`) got a faithful reimplementation of the funnel's recording behavior instead, so they'd actually catch a regression in the wiring, not just the deferral condition.

openwong2kim (approved):
Both blockers verified closed at the code level.

- **The funnel move is the structurally right fix.** Traced every `agentStatus` broadcast site: the two hook-path broadcasts I flagged (`hooks.rpc.ts` — the turn-boundary `complete` via `buildTurnBoundaryMetadata` at its `broadcastMetadataUpdate` call site, and `awaiting_input`) both pass through `broadcastMetadataUpdate`, and the record now lives inside that funnel (`metadata.handler.ts`, the `payload.ptyId && payload.agentStatus !== undefined` guard) — so hook statuses are covered without any per-site patch, and the withheld-status exclusion really does fall out for free since a withheld status never reaches `payload.agentStatus`.
- **The #733 no-regression test now earns its name** — the 3000-byte burst crosses ActivityMonitor's ACTIVE threshold, so the deferral path is actually exercised instead of vacuously passing.
- The "Accepted cost of #935 direction 3, do not 'fix'" comments at both clear sites are exactly what I asked for — the next reader won't re-introduce the deferral.
- Stubbing the two new exports in the six unrelated suites and faithfully reimplementing the funnel only in the two deferral suites is the right split of ceremony.

One honesty note on scope: I did not run a live GUI session with a real agent turn for this one — the verification above is code-tracing plus the test suite. The change is display-signal only (no data path), and daily use will exercise the completion-dot and awaiting-input flows immediately, so any residue will surface fast and loudly. Approving and merging on that basis.

PR #1015, opened 2026-08-24
Stash a pane out of the layout without killing it

### LINKED ISSUE #977: [Feature]: Fullscreen button should snap you our of dual pane view into single pane view
### Problem / motivation

I'm not sure if that's a bug or intended behavior, so will just file it as feature request.
Currently if you're in dual pane view, it's a bit non-obvious from UX perspective how to snap out of it into basic single pane view (dragging doesn't work as expected).

https://github.com/user-attachments/assets/70e72502-2257-4095-be1e-8ef7a8c2782c

### Proposed solution

When clicking on fullscreen button, you should be snapped into regular single page view.

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

openwong2kim:
Closes #977

@RudVlad473 filed this as "not sure if that's a bug or intended behavior": in a
dual-pane view there is no obvious way back to a single pane. Dragging doesn't do
it, and the only button that removes a pane is `✕` — which destroys the session
running inside it. So tidying a split and killing an agent were the same gesture,
and that is why the way out wasn't obvious: the obvious-looking way out was
destructive, so nobody wanted to press it.

The fix is a second, reversible verb.

## Stash a pane instead of killing it

`Stash` takes a pane out of the layout and leaves its session running. The
remaining pane fills the space — which is the single-pane view the issue asked
for — and the stashed one shows up in the workspace's sidebar list, gathered
under a divider and marked with a crossed-out eye, with its status still
updating as proof it is alive. Hover or focus the row and the eye turns back on:
one click brings the pane back next to its former neighbour, scrollback replayed.
A ten-second undo rides the confirmation toast.

Reachable from the pane header, `prefix` + `!` (tmux's break-pane), the sidebar,
and the command palette. Refused with a reason — never silently — when the pane
is the only one on screen, when there's no daemon to hold the session, or when
the pane holds unsaved editor/diff state that can't be replayed.

It's also a first-class RPC/MCP surface (`pane.stash` / `pane.unstash`): a
stashed pane stays fully addressable for `terminal_send`, `terminal_read` and
A2A delivery, because its PTY is alive and none of those care where the pane is
on screen. Only position-dependent calls are refused, and they answer with a
`PANE_STASHED` error carrying a `recovery` payload the caller can invoke
verbatim.

## The header had to earn its fifth button

Adding `Stash` pushed the action cluster to five buttons (142px), and that
exposed an older problem: the cluster is fixed-width while the tab strip beside
it is `flex-1 min-w-0`, so on a narrow pane the strip doesn't shrink gracefully —
it collapses to nothing. At ~200px the header was 100% buttons and 0% identity,
with the last button clipped.

Below 222px the five buttons now collapse into one `⋮` that opens them as a
vertical menu, and right-clicking a pane header opens the same menu at any width.

Wide panes keep the cluster; narrow ones collapse to a single ⋮

The `⋮` never collapses further, however narrow the pane gets. An earlier cut
dropped it below ~111px, which was wrong on this change's own argument: that is
the most cramped state there is, and the menu holds the only ways out of it
(zoom, stash). The tab strip scrolls, so identity survives the 31px either way.

The ⋮ menu open on a 98px pane

One of the five had no other entry point at all: "add a browser tab to **this**
pane" is only the header button — the palette's Open Browser passes `forceNew`,
which splits off *another* pane and makes a cramped layout worse.

Right-click works on a full-width header too

## Verification

Measured live against the running app rather than reasoned about: at a 1282px
window the panes sat at 378 / 257 / 199 / 98 / 98px in one layout, which
exercises both sides of the threshold. Confirmed there — the 98px panes render
the `⋮` and its menu opens all five actions inside the viewport (the 216px menu
escapes the 200px pane because it portals to `document.body`, the same path the
terminal context menu takes); a right-click at x=763 places the menu's left edge
at x=763; Escape returns focus to the trigger it came from.

Keyboard: arrow keys walk the menu wrapping at both ends, Home/End jump, and
disabled items stay focusable so a screen reader can read *why* they're
unavailable. Zoom is a `menuitemcheckbox`, not a button wearing `aria-pressed`.

Also in here: the 17 stash strings are translated into all 23 locales, not just
English and Korean.

Tests pin both the geometry and the contract — the `⋮` markup is held to its
31px the way the cluster's 142px already was, `paneActionsMode` never returns
"no actions" for a measured width, and one test asserts the cluster and the menu
name the *same* action set, since the next action landing on one and not the
other is exactly the failure this design exists to prevent.

## One thing reviewers should look at closely

The two new MCP tools didn't fit. The commander profile's `tools/list` view has
a 48000-byte ceiling — context the orchestrator brain pays for on *every* turn —
and measuring `origin/main` first showed it sits at exactly **48000/48000**.
There is no slack, so this branch could not have added anything: not two tools,
not even the two `includeStashed` booleans (390 bytes on their own). Raising the
ceiling would quietly tax every commander turn, so the bytes were found instead.

Roughly half came from the new code (tighter tool descriptions — the refusal
conditions are gone because each is already returned as an error that explains
itself). The other half came from three existing descriptions that were over
budget for what they carried, and **that part touches tools this PR otherwise
has nothing to do with**:

- `wmux_events_poll.types` — 1779 bytes, the largest single string on the
  surface. Kept the field names, the filterable kinds, that `a2a.task` is a
  pointer rather than a payload, and that it's dual-party (without that last
  one an unscoped poll looks broken: zero events, no error). Dropped an
  OSC-133 migration note and a per-signal restatement of pairs listed a
  sentence earlier.
- `pane_set_metadata.expectedVersion` — dropped v2.8/v2.9 version tags and a
  legacy aside; kept the failure mode, where to read the version, and that `0`
  guards a never-written pane.
- `a2a_task_update.evidence` — was a near-verbatim copy of its own tool
  description; now states the shape to pass and points there for grading.

Commander lands at **47995/48000**. Five bytes of headroom is the honest
number: the budget was full before this PR and it is full after. That's worth a
separate decision at some point — the next tool to be added faces the same wall
— but raising it silently inside a feature PR seemed like the wrong place.


openwong2kim:

Implements pane stash (#977): take a pane out of the layout with its session alive, restorable from the sidebar, plus `pane.stash`/`pane.unstash` RPC/MCP surface, `includeStashed` listings, and a collapsed `⋮` header menu below ~222px. Large diff (+5986/−429, 109 files) touching session persistence, the MCP tool catalog, sidebar/roster, and i18n — the stash-refusal paths and down-grade round-trip are the riskiest spots to verify. P2: substantial and well-tested, but the breadth warrants a careful pass before merge.

PR #1017, opened 2026-08-24
Anchor streaming IME compositions to the input line found by content

### LINKED ISSUE #1016: Streaming IME anchor needs a content signal: find the agent's input line in the buffer
### Where this comes from

#951 fixed IME candidate-window anchoring for an idle pane, and PR #953 added the quiet-caret snapshot so compositions during agent streaming defer to the last cell the cursor held through an output-quiet span. The 2026-08-23 field round (PR #953, final report) proved that machinery has hit the ceiling of what any cursor-derived rule can do:

- While Claude Code streams, its cursor rides the repaint and rests at the **end of painted content** between bursts. It is never placed on the input caret during the stream (upstream: anthropics/claude-code#25186 — Ink's `useCursor` exists, Claude Code has not shipped it).
- "End of painted content" can be **any column**. The field log shows quiet-span snapshots at column 137 of a 139-column pane and at column 53 mid-line. The last-column refusal shipped in #953 only catches the full-width special case (the earlier 128-column log parked exactly on column 127, which is why it looked like "the last column" for three rounds).
- Consequence: during a sustained stream, the snapshot is routinely a park, and `src=caret` faithfully anchors the candidate window to it. Dwell, quiet spans, and column checks all read the cursor; when the true caret is a cell the cursor never visits, no combination of them can find it.

Field evidence: PR #953 comments of 2026-08-23 (two videos + `main-2026-08-23.log`, `ime-anchor4` records `sel=(137,36) src=caret` / `sel=(53,39) src=caret` with the live cursor mid-line).

### The fix this issue tracks

A different signal class: **read the agent's input line out of the buffer** instead of inferring it from cursor behavior.

- At `compositionstart` (user-paced, so cost is irrelevant), scan the viewport bottom-up for the recognized agent's input-line marker — for Claude Code, the `│ > ` box row (top border `╭─`, prompt row, bottom border `╰─`).
- Anchor the composition to the caret position on that row (prompt offset + typed length, or simply the prompt column as a floor).
- **Gate per known agent.** wmux already recognizes which agent runs in a pane; the marker scan runs only for agents whose input-line shape we know. Unrecognized panes keep the #953 behavior.
- Marker matching must tolerate version drift in the agent's chrome (box width, margins, locale) and must not fire on streamed content that happens to contain the marker glyphs (e.g. an agent printing its own UI in a code block) — bottom-up scan plus box-frame adjacency covers the common cases; the diagnostic should record when the scan matched and where.

### Non-goals

- No more cursor heuristics. Three generations of them lost to the untouched baseline (see the `imeAnchor.ts` header for the full account), and the fourth (#953's refusal) reduced but could not remove the miss.
- Not conditional on upstream: if Claude Code ships caret parking upstream, the cursor path starts working on its own (#953's machinery already prefers it) and the marker scan becomes a fallback for agents that never park.

### Definition of done

- Composing Chinese (WeChat IME / Microsoft Pinyin) while Claude Code streams anchors the candidate window to the input box, not to a park; verified by an `ime-anchor` field log showing the marker-scan source selected during streaming.
- Idle behavior unchanged (it is already correct).
- Tests cover: marker found (anchored), marker absent (falls back to #953 selection), marker-like content in streamed output (not matched), agent not recognized (scan skipped).

openwong2kim:
Closes #1016. Refs #951, #953.

### Why

The 2026-08-23 field round on #953 (two videos + `main-2026-08-23.log`) proved the cursor signal is exhausted for the streaming case:

- While Claude Code streams, its cursor rests at the **end of painted content** between bursts and never visits the input caret (upstream: anthropics/claude-code#25186).
- "End of painted content" can be **any column** — the log holds quiet-caret snapshots at column 137 of a 139-column pane and at column 53 mid-line — so the last-column refusal from #953 only catches a painted line that happens to fill the row.
- Consequence: during a sustained stream the snapshot is sometimes the park itself, and `src=caret` faithfully anchors the candidate window to it.

Dwell, quiet spans, and column checks all read the cursor; when the true caret is a cell the cursor never visits, no combination of them can find it.

### What

A second signal class for the streaming anchor: **read the input line out of the buffer by its chrome.**

- At `compositionstart` (user-paced — a full-viewport scan is free), when the pane's agent is recognized as Claude Code **and output is flowing**, the viewport is scanned bottom-up for the input box (`╭─` border with a `│ > ` prompt row directly beneath). The composition anchors to the prompt row at the input's start column (`src=marker`), outranking both the snapshot and every dwell source.
- **The idle pane stays cursor-driven on purpose.** It is field-verified correct, and every generation that touched it lost to the untouched baseline (see the `imeAnchor.ts` header's account). The scan does not run while output is quiet.
- Gated per agent via the pane's `surfaceAgent` slug; unrecognized agents, a scan that finds nothing, and callers without a buffer line reader all keep the #953 behavior unchanged.
- Bottom-up scan plus box-frame adjacency keeps quoted chrome in the transcript (an agent printing its own UI) from matching; the bottom-most real box wins.
- A marker-sourced composition pins the inline preedit with the textarea, same as caret-sourced ones, so the two IME surfaces cannot tear apart (#951 field report).
- Diagnostic tag bumps `ime-anchor4` → `ime-anchor5`; `src=marker` records make a field log self-identifying.

### Tests

15 new tests: the scanner (prompt row found; indented box column math; bottom-most match beats quoted chrome; bare prompt line without the border refused; wrapped multi-line input; unreadable/markerless screens), the selection (marker outranks a park snapshot — the exact 137-of-139 field shape; covers a stream that never produced a snapshot; idle ignores the marker; marker without a viewport ignored), and the wiring (recognized agent anchors to the marker and pins the preedit; unrecognized agent never scans; empty scan falls back to `src=caret`; idle composition never scans). Full imeAnchor suite: 76/76.

### Verification

CJK IMEs cannot be reproduced on the dev boxes — the `ime-anchor5` log is the field discriminator, as it has been for every round of #951/#953. The testable half of #1016's definition of done is covered by the suite above; the field half stays open for any future report.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Chinese IME candidate-window positioning while Claude Code is actively streaming.
  - Input composition now stays anchored to the correct prompt line, including wrapped or indented input.
  - Preserved existing cursor-based positioning when the terminal is idle or no valid prompt is detected.
  - Added safeguards for unreadable terminal content and unsupported environments.
  - Improved positioning reliability when terminal content changes during streaming.

### DIFF
--- a/changelog.d/1017.md
+++ b/changelog.d/1017.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **Chinese IME candidate window now anchors to Claude Code's input box
+  while output streams.** The quiet-caret snapshot shipped for #951 could
+  itself capture the streaming cursor's parking spot (the end of whatever
+  the last frame painted — any column), because Claude Code never parks its
+  cursor on the input caret mid-stream. Compositions started during
+  streaming now find the input line by content — the `│ > ` prompt row
+  under its box border — and anchor there, outranking every cursor-derived
+  source. Idle panes keep the field-verified cursor behavior unchanged, and
+  the IME diagnostic (`ime-anchor4` → `ime-anchor5`) records when the
+  content scan was used. (#1016)
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1172,22 +1172,31 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
     // field.
     const imeAnchorLogsLeft = { start: 20, mid: 20 };
     const imeAnchor = attachImeAnchor(terminal, {
+      // #1016: gates the input-line content scan to agents whose chrome the
+      // scanner understands. Read per composition, so it tracks the pane's
+      // live agent identity without a re-attach.
+      getAgentSlug: () => {
+        const id = ptyIdRef.current;
+        return id ? useStore.getState().surfaceAgent[id]?.slug : undefined;
+      },
       onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, outputGap, caretAge, edge, selY, selX }) => {
         const budget = phase === 'start' ? 'start' : 'mid';
         if (imeAnchorLogsLeft[budget] <= 0) return;
         imeAnchorLogsLeft[budget] -= 1;
         // Mirrored into the main-side log file by src/main/index.ts's
-        // console-message listener, so the user can share it. The "4" in the
-        // tag marks the quiet-caret build (#951) so a shared log is
+        // console-message listener, so the user can share it. The "5" in the
+        // tag marks the input-line marker build (#1016) so a shared log is
         // unambiguous about which release produced it. src/held/restAge/sel
         // are the cause-3 discriminator: src=resting means the composition
         // started mid-repaint and the anchor used the last resting cell;
         // src=caret with gap= is the #951 discriminator: output was still
         // flowing, so the anchor used the quiet-caret snapshot instead of
-        // any buffer cursor. pin= is the textarea correction, preedit= the
-        // live composition-view correction.
+        // any buffer cursor; src=marker means the agent's input line was
+        // found by content (#1016) and outranked them both. pin= is the
+        // textarea correction, preedit= the live composition-view
+        // correction.
         console.info(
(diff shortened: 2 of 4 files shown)


openwong2kim:
Implements #1016: while output streams, compositions scan the viewport bottom-up for the Claude Code `│ >` prompt row and anchor there (`src=marker`), outranking the quiet-caret snapshot. Touches `imeAnchor.ts` selection and `useTerminal.ts` wiring; the main risk is a false marker match, mitigated by the agent gate and fallback to #953 behavior. P2 — direct fix for the streaming IME anchor bug.

PR #1019, opened 2026-08-24
feat(cli): add wmux daemon start|stop|status (#1001)

p-poppe:
## Summary

Adds `wmux daemon start | stop | status` — a headless daemon lifecycle verb, per your scope note in #1001. Exactly the v1 you specified: no first-run wizard, no web setup (that stays `wmux web`'s job), no service-manager integration.

## The extraction

Per your hard requirement ("the spawn path must not fork"): `src/main/daemon/launcher.ts`'s spawn/readiness chain now lives in `src/shared/daemon/daemonLauncherCore.ts`, which has zero import of `electron`. Both `src/main/daemon/launcher.ts` (Electron main) and `src/cli/commands/daemon.ts` (the new CLI verb) consume the *same* `ensureDaemon()` — they differ only in what they inject through a `DaemonLauncherDeps` seam:

| | Electron main | CLI |
|---|---|---|
| daemon script location | `app.getAppPath()` / `process.resourcesPath` candidates (unchanged) | resolved relative to the CLI bundle's own `__dirname` — `dist/cli-bundle` and `dist/daemon-bundle` are already built as siblings (package.json's `build:cli`/`build:daemon`, forge.config.ts's `extraResource`), in both the dev-checkout and installed-app layout, so one relative path covers both |
| version stamp | `app.getVersion()` | nearest `package.json`, same fallback idiom `doctor.ts`/`system.ts` already use for the CLI's own version reporting |
| unverified stale PID | native recovery dialog (`WMUX_NO_DIALOG=1` refuses in tests) | always refuses — structurally, not by env var, since a headless caller has no window for a dialog and no keyboard to ask either way |

`pollDaemonReady()` moved as-is, already fully dependency-injected. The four pre-existing `launcher.*.test.ts` suites (49 tests) run unmodified against the new location and pass — the public behavior for the Electron path is untouched.

`status` = `daemon.ping`, `stop` = `daemon.shutdown` — both pre-existing daemon-pipe RPCs (the same `daemon.shutdown` the auto-replacement path already uses). No new termination path.

## Verified end to end, not just unit-tested

Ran the actual CLI bundle against an isolated `WMUX_DATA_SUFFIX` instance:

```
$ wmux daemon status
wmux daemon is not running.

$ wmux daemon start
wmux daemon started (PID 24868)
pipe: \.\pipe\wmux-daemon-...

$ wmux daemon status
wmux daemon is running.
  status: "ok"
  pid: 24868
  ...
  spawnedByVersion: "3.46.0"          <- confirms the version stamp, not the 'unknown' sentinel

$ wmux daemon start          # second call
wmux daemon already running (PID 24868)   <- reuse path, no double-spawn

$ wmux daemon stop
wmux daemon stopped.

$ wmux daemon status
wmux daemon is not running.
```

## One thing I caught in my own first pass

`--json daemon start` originally echoed the daemon's `authToken` straight to stdout. Nothing downstream needs it — `status`/`stop` resolve their own token via `resolveDaemonAuthToken()` — and printing a live credential risks it landing in shell history or a captured log. Fixed before this PR: the JSON output carries `pid`/`pipeName`/`spawned` only.

## Punted, per your note

- `ClaudeWorker` (background A2A task execution without a visible pane) stays Electron-coupled — not touched here, callers fall back to daemon-native panes.
- Renderer-routed surfaces (browser panes, some MCP tools) won't function against a headless daemon — worth a docs line, not code, for v1.

## Changelog

`changelog.d/<this-pr-number>.md` to follow once the PR number is assigned — will push it as a follow-up commit.

## Test plan

- [x] `npx tsc --noEmit` — 0 errors, both `tsconfig.json` and `tsconfig.cli.json`
- [x] `npm run build:cli` && `npm run build:daemon` — both bundle cleanly (no `electron` pulled into the CLI bundle)
- [x] New: `src/shared/daemon/__tests__/daemonLauncherCore.deps.test.ts` (2 tests) — the injected deps are what's actually consulted, not a parallel hardcoded path
- [x] New: `src/cli/commands/__tests__/daemon.test.ts` (10 tests) — start/stop/status, reuse vs spawn, failure paths, the auth-token redaction
- [x] Existing `src/main/daemon/__tests__/launcher.*.test.ts` (49 tests) — unmodified, pass at the new location
- [x] Manual end-to-end run against a live isolated daemon instance (see above)
- [x] Full suite: 12294 passed / 19 failed / 27 skipped — all 19 failures are pre-existing `worktree.handler.test.ts` (Windows `EBUSY`/timeout on git-worktree cleanup, filesystem locking) and one `touchScroll.test.ts` worker-pool timeout, neither touching anything this PR changes; confirmed unrelated by inspection, not re-run against a clean tree.

Related: #1001.

## Summary by CodeRabbit

- **New Features**
  - Added headless daemon lifecycle commands: `daemon start`, `daemon status`, and `daemon stop`.
  - Added human-readable and JSON output with authentication-token redaction.
  - Daemon commands now work without launching the Electron app.

- **Bug Fixes**
  - Improved daemon startup, readiness, recovery, liveness checks, and safe shutdown.
  - Prevented termination of unrelated processes and handled startup failures more reliably.

- **Tests**
  - Added comprehensive coverage for lifecycle commands, recovery, errors, and output formats.

- **Documentation**
  - Documented the new daemon lifecycle commands.


openwong2kim:
Adds headless `wmux daemon start|stop|status` verbs by extracting `ensureDaemon()` from Electron main into shared `daemonLauncherCore.ts`; the app path rebinds the same core, existing tests unchanged. Low risk: purely additive CLI surface, auth token redacted from output, stale-PID recovery refuses structurally. P3 — scope-complete v1 per #1001, well tested, nothing urgent.

p-poppe:
Addressed all 5 findings from CodeRabbit's review.

**🔴 Critical (host-identity inference) — confirmed real, fixed by narrowing the fix, not the architecture I initially reached for.**

I first considered supplying a cross-host "expected image set" via `DaemonLauncherDeps` (each host declaring every binary name a genuine daemon could run under), but that requires each host to know the *other* host's binary name, which isn't knowable in general (a packaged Electron app's name varies per install/platform). Went with the smaller, more robust fix instead: an image-name mismatch no longer short-circuits to "stale, someone else's program." It only downgrades confidence and logs; the command-line marker check (`daemon-bundle` / `daemon/index.js` in argv) — which is host-independent, since both hosts spawn the same script — now always decides. Applied identically in `ensureDaemon`'s stale-PID verification and in `killVerifiedDaemonPid`. An unrelated process reusing a recycled PID is still correctly refused, since its cmdline won't carry the daemon markers either way — nothing about the safety net weakened, just un-gated from an assumption that stopped holding once there were two hosts.

Also replaced the adjacent `isElectron` heuristic (`nodePath === process.execPath && !nodePath.toLowerCase().includes('node.exe')`) — it was *also* execPath-based and wrong on non-Windows (the substring `node.exe` doesn't match plain `node` on macOS/Linux, so a headless CLI spawn there would've been misclassified as Electron and gotten `ELECTRON_RUN_AS_NODE=1`). Replaced with an explicit `isElectronHost()` on `DaemonLauncherDeps`, injected `true`/`false` by each consumer instead of inferred.

**🟠 Major (missing `error` listener):** added, rejects with a clean message instead of throwing uncaught. Regression test mocks `spawn` to emit `error` and asserts the promise rejects rather than crashing the process.

**🟡 Minor ×2:** `daemon status`'s unreachable-daemon branch now sets `process.exitCode = 1` (updated the existing test that had pinned the old exit-0 behavior). `DaemonInfo.pid` is now `number | null` — the reconnect path that previously fabricated `0` now reports `null`, which the CLI renders as `unknown` instead of a fake PID; updated the two `DaemonRespawnController` call sites that pass it to `killVerifiedPid` (both are only reached on the `spawned: true` branch, where pid is always real, but the type is shared so they're now null-guarded) and added an early return in `replaceStaleDaemon` for the (currently unreachable in practice, but now type-honest) null case.

**🔵 Trivial:** the 5 inline `require('child_process')` calls are now one top-level `execFileSync` import; the two intentionally-empty `catch` blocks each have a one-line comment.

**New tests:**
- `daemonLauncherCore.spawnError.test.ts` — mocks `child_process.spawn`, pins the clean-reject contract for both a bare `error` emission and `error` + `exit` racing.
- `killVerifiedDaemonPid.crossHost.test.ts` — spawns a real child process through a *renamed copy* of the current Node binary (portable across platforms, no OS mocking needed) to reproduce a genuine cross-host image mismatch, confirms the daemon is still verified and killed via the cmdline marker, and confirms an unrelated process under the same renamed binary (no marker in argv) is still refused.

Full suite: `src/shared/daemon` + `src/main/daemon` + `src/cli/commands` — 21 files, 335/335 passing. `tsc --noEmit` clean. ESLint clean on all touched files.

openwong2kim (changes requested):
Thanks — the extraction itself is exactly what #1001 asked for, and the deps-seam + e2e verification made this easy to review. Three things need to change before merge, though; the first one breaks the feature in the shipped app.

**1. Version fallback `'0.0.0'` gets the CLI-spawned daemon killed (blocker)**

In the installed layout the CLI bundle lives at `<resources>/cli-bundle`, where `__dirname/../../package.json` does not exist, so `resolveSpawnedByVersion()` returns `'0.0.0'`. The staleness gate (`src/main/daemon/daemonReplacement.ts`, `isDaemonOlder`) keeps the `'unknown'` sentinel but *parses* `'0.0.0'` and reads it as positively older — so the first GUI connection replaces (kills) the daemon the CLI just started, PTYs included. Your e2e run showed `spawnedByVersion: "3.46.0"` because it ran from a dev checkout, where `../../package.json` resolves. (doctor.ts shares the idiom, but there the fallback is display-only.)

Fix: fall back to `'unknown'` (the gate keeps it), or bake the version in at build time.

**2. `runStop` swallows every transport error as success (blocker)**

The catch in `runStop` prints "wmux daemon is not running." and returns without setting an exit code — timeouts, permission errors and connection aborts all exit 0, so `wmux daemon stop && ...` proceeds while the daemon is still alive. `runStatus` handles the same case with `process.exitCode = 1` (with a comment explaining why). Suggest: treat only not-running-shaped errors (ENOENT/ECONNREFUSED) as idempotent success; surface everything else and set `exitCode = 1`.

**3. Importing `bootTrace` makes every CLI command print to stderr (blocker)**

`daemonLauncherCore` imports `main/util/bootTrace`, whose module body calls `markBoot('js-start')` at eval time, and `src/cli/index.ts` imports all commands statically — so after this PR *every* `wmux` invocation emits a `[boot-trace]` line on stderr. Suggest injecting `markBoot` through `DaemonLauncherDeps` as an optional no-op callback (same treatment `askUserToRecoverFromStalePid` already gets).

Smaller things, fine as follow-ups:
- `ensureOk(response)` sits outside the try in both `runStatus`/`runStop` — an `ok:false` response crashes with a stack trace instead of a clean exit-1.
- The spawn `'error'` handler rejects without cancelling `pollDaemonReady`, so a failed spawn can leave the one-shot CLI hanging up to the 90s ceiling after printing the error.
- `status --json` success output is the raw ping result (no `running` field), while the failure shape is `{ok, running, error}` — scripts keying on `.running` get `null` when the daemon is up.
- The cmdline markers (`'daemon/index.js'`) are generic enough to match an unrelated `node /srv/anything/daemon/index.js` on a recycled PID; an exact path-segment match (`/daemon-bundle/index.js`) would narrow that. Relatedly, with `definitiveOnly: false` a *definitive* image mismatch plus a null cmdline still kills — refusing on that combination keeps the cross-host support while restoring the old guarantee.

p-poppe:
Addressed all three blockers, plus the smaller items — folded them in since we were already touching this code. `97be1245`.

## 1. Version fallback `'0.0.0'` → `'unknown'` (blocker)

Fixed exactly as suggested: `resolveSpawnedByVersion()` now falls back to `'unknown'` on any failure to read/parse `package.json`, which the gate already special-cases as "keep it". Verified end-to-end, not just by reading — `daemon.spawnedByVersion.test.ts` feeds the function's real output straight into the real (unmocked) `isDaemonOlder()` and asserts `older: false`, plus a companion test documenting that `'0.0.0'` (the old fallback) *would* have read as positively older, so a future revert of the fallback value fails a test even if nobody reads the prose.

## 2. `runStop` swallowing every error as success (blocker)

Exported `isConnectFailure` from `cli/client.ts` (was module-private) and used it in `runStop`: only a connect-level failure (no daemon listening) is idempotent success now. A timeout, `EACCES`, or a dropped connection surfaces as an error with `exitCode = 1`, matching `runStatus`'s existing convention. Two new tests cover the non-connect-failure path in both text and `--json` mode; had to fix the *existing* "already down" test too — it asserted on a message-only `Error('connect ENOENT')` with no `.code` set, which never matched real production errors (`client.ts` always sets `.code` on the wrapped error) but happened to pass under the old message-blind catch-all.

## 3. `bootTrace` import printing on every CLI invocation (blocker)

`markBoot` is now an optional `DaemonLauncherDeps` callback instead of a static import in `daemonLauncherCore.ts`. Electron's `launcher.ts` supplies the real one (import moved there); the headless CLI's `cliDaemonDeps` omits it, so every call site is `deps.markBoot?.(...)` — silent no-op on the CLI, unchanged behavior in Electron. Verified statically: `grep -rln "from '.*bootTrace'" src` now returns only `main/index.ts` and `main/daemon/launcher.ts`, neither reachable from `cli/index.ts`'s import graph.

## Smaller items

- **`status --json` now includes `running: true` on success**, matching the failure shape (`{ok, running, error}`). New test asserts the field.
- **Spawn `error` handler now cancels the readiness poll** instead of just rejecting — it couldn't call `readiness.cancel()` before because `readiness` didn't exist yet at registration time; hoisted a `let readiness` (declared `undefined`, assigned once `pollDaemonReady` runs) so the handler can route through it once available, falling back to a direct `reject` for the (unlikely) case an error fires before the poll even starts.
- **`killVerifiedDaemonPid`: image-mismatch + null-cmdline now refuses**, even with `definitiveOnly: false`. Neither signal alone should refuse (cross-host support needs that), but both pointing away from "ours" restores the pre-#1001 guarantee.
- **`ensureOk` "outside the try" — verified, no change needed.** Read the implementation: it calls `process.exit(1)` directly on an error response, never throws. There's no exception path to escape the try/catch here, so wrapping it wouldn't change any observable behavior. Flagging this rather than adding a no-op try for the sake of matching the suggestion.
- **Cmdline marker narrowing — attempted, reverted, flagging as a real follow-up.** Tried anchoring the generic markers as `/daemon/index.js` / `/daemon-bundle/index.js`. It doesn't actually narrow anything: `.includes()` has no path-boundary awareness, so `/daemon/index.js` is still a substring of the exact false positive it was meant to exclude (`/srv/anything/daemon/index.js`). It also broke `killVerifiedDaemonPid.crossHost.test.ts`, which correctly relies on the bare `'daemon-bundle'` marker. A real fix needs to tokenize the cmdline and compare path segments (or at minimum `endsWith` on a normalized path), not string-contains the whole line — left as-is rather than ship something that reads safer without being safer.

## Test plan

`npx vitest run src/shared/daemon src/main/daemon src/cli/commands src/cli/__tests__` — 29 files, 455 passed. `npx tsc --noEmit` clean. `npx eslint` on touched files — one pre-existing unrelated warning (`'os' is defined but never used` in `client.ts`, confirmed via `git stash` against the unmodified branch), no new errors or warnings.

p-poppe:
Fixed all 4. Point-by-point:

**1. `--json` diagnostics leaking to stdout — fixed, verified end-to-end.**

`DaemonLauncherDeps` gains optional `log`/`warn`; every `console.log`/`console.warn` inside `spawnDaemon`/`ensureDaemon` now goes through `(deps.log ?? console.log)(...)` / `(deps.warn ?? console.warn)(...)` (omitting them keeps Electron's launcher behavior byte-for-byte identical — it never supplies them). The CLI's `cliDaemonDeps` routes both to `console.error`, in both `--json` and text mode, since neither output shape has room for launcher-internal chatter.

Verified by actually building and running it, not just reading the diff:
```
$ node dist/cli-bundle/index.js daemon start --json
```
Before the fix (confirmed via `git stash` on just these two files + rebuild): stdout carried 5 `[launcher]` lines ahead of the JSON object. After: stdout is exactly the JSON object; all diagnostics land on stderr.

**2. `EPERM` no longer reported as "already stopped."**

Added `isDefinitelyNotRunning` (ENOENT/ECONNREFUSED only) in `client.ts`, next to `isConnectFailure`, with a doc comment explaining why they have to stay separate — `isConnectFailure` is correct to include `EPERM`, because it answers a different question (`sendRequest`'s own pipe→TCP retry decision, where a Windows ACL issue should try the other transport). `runStop` now uses the narrower one. New test: an `EPERM` rejection now exits 1 and does not print "not running."

**3. `oldClient` leak in the no-verifiable-pid bail-out — fixed.**

`replaceStaleDaemon` now calls `oldClient.disconnect()` before setting `lastError`/returning `null` in that branch — it was the one early-return in the function that skipped the normal `disconnectClient` cleanup `runDaemonReplacement` provides everywhere else. New test spawns the exact edge case (`ensureDaemonImpl` returning `pid: null, spawned: false`) and asserts `disconnectCalls === 1`.

**4. Cmdline marker: segment-aware instead of raw substring.**

Extracted `cmdlineMatchesDaemonScript` (quote-aware tokenizer + path-segment split), used at both remaining call sites. A segment must **start with** `daemon-bundle` (not equal it exactly — the existing cross-host fixture's `daemon-bundle-fake-index.js` needed to keep matching, and future bundle-dir suffixing shouldn't break this), or be an exact `daemon`/`index.js` (or `daemon`/`daemon`/`index.js`) segment pair. This rejects your exact example — added `/.../my-daemon-bundle-backup/index.js` as a new case in `killVerifiedDaemonPid.crossHost.test.ts`, confirmed it fails against a reverted `.includes()` version and passes against the fix, while the two pre-existing cross-host tests still pass unchanged.

Full suite (`src/cli/commands`, `src/main/daemon`, `src/shared/daemon`): 346 passed. `tsc --noEmit` and `eslint` clean (the pre-existing `daemon.test.ts` escape-char errors and `client.ts` unused-`os` warning are unchanged from `origin/feat/wmux-daemon-verb` — confirmed by linting that revision directly, not introduced here).

openwong2kim (approved):
All three blockers verified closed, live — not just by reading the diff. Built the CLI and daemon bundles at `e2e34ffb` and ran them against isolated `WMUX_DATA_SUFFIX` instances:

- **Version stamp**: copied `cli-bundle`/`daemon-bundle` as siblings into a bare directory with no `package.json` above them (the shape the installed layout has, and the one the original e2e couldn't see from a dev checkout). `daemon status --json` reports `spawnedByVersion: "unknown"` — the value the gate keeps. The failure this PR would have shipped is gone.
- **`stop` error handling**: pointed the CLI at a socket that accepts connections and never answers. Text mode prints `wmux daemon stop: Request timed out after 5 seconds.` and exits 1; `--json` gives `{"ok": false, ...}` and exits 1. Previously this exact case printed "not running" and exited 0.
- **stderr hygiene**: `wmux doctor` stderr carries zero `[boot-trace]` lines, and `daemon start --json` stdout parses as JSON with no `[launcher]` chatter.
- **Lifecycle and restart**: status(1) → start(0) → start again (reuses, no double spawn) → stop(0) → stop again (0) → status(1). `stop && start` spawns a genuinely new PID and the old process is gone, so the shutdown ack doesn't leave the restart sequence reusing a dying daemon.

Approving. One correction for the record, and one follow-up.

**The marker narrowing did not close the case I raised.** Your note says the segment-aware matcher "rejects your exact example," but the example I gave was `node /srv/anything/daemon/index.js`, not `my-daemon-bundle-backup`. The new tokenizer still matches it: the segments are `['srv','anything','daemon','index.js']`, and the `daemon` + `index.js` adjacent-pair rule fires on the last two.

I verified this by execution rather than inspection — spawned an unrelated `node <dir>/daemon/index.js` sleeper and called `killVerifiedDaemonPid(pid, { definitiveOnly: false })` against it. Result: returned `true`, and the innocent process was gone afterwards.

To be fair about severity: this is **pre-existing, not a regression from this PR** — before #1019 the same-image case reached the same `.includes()` markers and killed it too, so nothing here got worse. It also needs PID reuse to trigger in production. That's why it isn't blocking the merge.

The structural fix is to stop pattern-matching altogether: the caller already resolved the daemon script path (`deps.daemonScriptCandidates`), so the cmdline's script token can be compared against those absolute paths instead of against generic markers. `daemon/index.js` as a marker is inherently ambiguous — it's an extremely common layout — and no amount of segment awareness fixes that. Worth a follow-up issue rather than more work on this PR.

PR #1020, opened 2026-08-24
fix(a2a): surface pane titles in a2a_discover, allow readonly cross-ws metadata reads

### LINKED ISSUE #1018: [Bug] a2a/pane tools can't read pane names cross-workspace — impossible to distinguish multiple sessions in one workspace
## Problem

When a workspace has 2+ panes (e.g. multiple Claude Code sessions open under one role/workspace, like "CEO"), there is no way from another workspace to tell which pane is which before sending an a2a message.

## Reproduction

1. Workspace "CEO" has 2 panes, both Claude Code sessions.
2. `a2a_discover` returns both panes, but each just has `agentName: "Claude Code"` — no distinguishing name/title, even though the user can see distinct pane/terminal names in the wmux UI itself (screenshot on file).
3. Tried `pane_get_metadata` on a pane belonging to another workspace to see if a custom name/title was set there instead:
   ```
   pane.getMetadata: pane.validateWorkspace: leaf "pane-2c24df02-09b8-417d-bb4c-a3bf2fa11617" not in workspace "ws-c9f8167f-730a-4dcd-bf43-cee4c6040219"
   ```
   `pane_get_metadata` is hard-scoped to the caller's own workspace, so even a readonly check of another workspace's pane metadata is impossible.

## Impact

`a2a_task_send` / `send_message` can address a workspace by name reliably, but the moment a workspace has multiple live panes, picking the *right* pane (vs the active/idle one) is a guess — effectively "shooting in the dark" when trying to reach a specific one of several parallel sessions under the same role.

## Requested fix (minimal, readonly)

- Allow `pane_get_metadata` (and/or `pane_list`) to read another workspace's pane metadata in readonly mode, OR
- Have `a2a_discover` surface each pane's display name/title (whatever the UI already shows the user) instead of just the generic `agentName: "Claude Code"`.

This alone would let an agent in one workspace correctly identify and target a specific pane in another workspace before sending a message.

## Worth considering longer-term

Once pane names/titles are readable, it would also make sense for `a2a_task_send`/`send_message` to accept a `paneId` or resolved pane name as a first-class target, not just workspace — today workspace-level addressing is fine for a single-pane workspace, but doesn't extend cleanly to multi-pane ones. Not required for this issue, just flagging the natural next step.

Reported via Claude Code (@cto, account p-poppe).

p-poppe:
## Summary

- `a2a_discover` only ever labeled a pane with the generic vendor `agentName` ("Claude Code"), so a workspace running several sessions of one vendor rendered as indistinguishable entries — another agent had no way to tell which pane was which before addressing one with `send_message`/`a2a_task_send`. Each entry in `agents[].panes` now also carries `paneTitle`, sourced from the same `surface.title` the sidebar roster already leads with (#934). Additive only — `agentName` is unchanged.
- `pane_get_metadata` was hard-scoped to the calling workspace and errored `"not in workspace"` against any other pane, even though the underlying `pane.rpc` `resolveTarget` already accepts an arbitrary `workspaceId` (it only verifies the paneId belongs to it). The MCP tool wrapper was the sole thing forcing it to the caller's own id. Added an optional `workspaceId` param so `pane_get_metadata` can do a readonly cross-workspace read. `pane_set_metadata` gets no such override and stays confined to the calling workspace.

## Changes

- `src/renderer/hooks/useRpcBridge.ts` — `a2a.discover`'s per-pane `panes[]` entries gain `paneTitle: s.title?.trim() || null`.
- `src/mcp/index.ts` — `a2a_discover` tool description documents `paneTitle`; `pane_get_metadata` gains an optional `workspaceId` param (`targetWorkspaceId || (await requireWorkspaceId())`), `pane_set_metadata` untouched.
- `src/renderer/hooks/__tests__/useRpcBridge.a2aPaneIdentity.test.ts` — new structural test locking `paneTitle`/`s.title` wiring, additive to `agentName`.
- `src/mcp/__tests__/paneGetMetadataCrossWorkspace.test.ts` — new structural test locking the `workspaceId` override on `pane_get_metadata` and its absence on `pane_set_metadata`.
- `changelog.d/1020.md` — changelog fragment (guessed PR number from the latest open PR at #1019; happy to rename if it lands differently).

## Test plan

- `npx vitest run src/renderer/hooks/__tests__/useRpcBridge.a2aPaneIdentity.test.ts src/mcp/__tests__/paneGetMetadataCrossWorkspace.test.ts` — 16 passed.
- `npx tsc --noEmit` — clean.
- `npx eslint` on touched files — no new errors/warnings (one pre-existing `import/no-unresolved` on an SDK type-only import and a handful of pre-existing warnings in `useRpcBridge.ts`, both unrelated to this diff).

Fixes #1018

## Summary by CodeRabbit

* **New Features**
  * A2A discovery results now include each pane’s title for clearer identification.
  * Read-only metadata can be retrieved from another workspace by specifying its workspace ID.
* **Bug Fixes**
  * Cross-workspace metadata access validates workspace identity and required pane details.
  * Metadata updates remain restricted to the current workspace.
* **Tests**
  * Added coverage for cross-workspace access, validation, identity enforcement, and pane title reporting.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -28,6 +28,21 @@
           }
         ]
       }
+    },
+    {
+      "files": [
+        "src/mcp/__tests__/paneGetMetadataCrossWorkspace.test.ts"
+      ],
+      "rules": {
+        "import/no-unresolved": [
+          "error",
+          {
+            "ignore": [
+              "^@modelcontextprotocol/sdk/(?:client/index|inMemory)\\.js$"
+            ]
+          }
+        ]
+      }
     }
   ]
 }
--- a/changelog.d/1020.md
+++ b/changelog.d/1020.md
@@ -1,0 +1,15 @@
+### Added
+
+- **`a2a_discover` now reports each pane's own title, not just the generic
+  vendor name.** A workspace running two or more sessions of the same vendor
+  (several "Claude Code" panes under one role) used to list as indistinguishable
+  rows — another agent had no way to tell which pane was which before
+  addressing one. Each entry in `agents[].panes` now carries `paneTitle`
+  (the same title source the sidebar roster leads with, #934) alongside the
+  unchanged `agentName`, additive so existing callers are unaffected.
+- **`pane_get_metadata` can read another workspace's pane, read-only.** It
+  was hard-scoped to the calling workspace and errored `"not in workspace"`
+  against any other pane — the underlying RPC already supported an explicit
+  `workspaceId`, only the tool wrapper forced it to the caller's own. Pass
+  `workspaceId` to read cross-workspace; `pane_set_metadata` keeps no such
+  override and stays confined to the calling workspace. (#1018)
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 80000,
-      "wireResultSha256": "6af0795491ea6a16310a03f4df0649e956e0d0ffb15718c9ce718a0f904cb350",
+      "wireResultSha256": "9ffa0f077647435fd41855d7d0d8d28a7ca94acbbdc27bcf58fc2d4256134969",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
@@ -99,7 +99,7 @@
     },
     "commander": {
       "maxListBytes": 49000,
-      "wireResultSha256": "a058c76c881bf4d5eadaf79ecde266fb1447c899ee09d823276efc76cd7c38d7",
+      "wireResultSha256": "c4cd51661adaef9efa197b947282fe11a4dbd8f2d28879691ed9eee20d6a979f",
       "instructionSha256": "5a97f851c892326eba7cbcf46292f6925834d5794172461d1ced997d90ac1deb",
       "toolNames": [
         "terminal_read",
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 80000,
-      "wireResultSha256": "6fa509beaa6229f875953d69eef7a13f60a8b389d564f84656d415eab3d0d890",
+      "wireResultSha256": "3370081b43158a38ddad92ba8ed397a5254f0132f1406bf2c619c6aaced6374a",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
@@ -98,8 +98,8 @@
       ]
     },
     "commander": {
-      "maxListBytes": 48000,
-      "wireResultSha256": "211b42767254c4d8f4cdbadcd1aa24053fbc01affee4a7807c3a0b196da85031",
+      "maxListBytes": 49000,
+      "wireResultSha256": "6cc0ba1944009d7a730ac7d2e77ec16aad82ea22c628558a718ccb32c602561f",
       "instructionSha256": "5a97f851c892326eba7cbcf46292f6925834d5794172461d1ced997d90ac1deb",
       "toolNames": [
         "terminal_read",
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 80000,
-      "wireResultSha256": "3370081b43158a38ddad92ba8ed397a5254f0132f1406bf2c619c6aaced6374a",
+      "wireResultSha256": "6af0795491ea6a16310a03f4df0649e956e0d0ffb15718c9ce718a0f904cb350",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
@@ -99,7 +99,7 @@
     },
     "commander": {
       "maxListBytes": 49000,
-      "wireResultSha256": "6cc0ba1944009d7a730ac7d2e77ec16aad82ea22c628558a718ccb32c602561f",
+      "wireResultSha256": "a058c76c881bf4d5eadaf79ecde266fb1447c899ee09d823276efc76cd7c38d7",
       "instructionSha256": "5a97f851c892326eba7cbcf46292f6925834d5794172461d1ced997d90ac1deb",
       "toolNames": [
         "terminal_read",
--- a/src/mcp/__tests__/paneGetMetadataCrossWorkspace.test.ts
+++ b/src/mcp/__tests__/paneGetMetadataCrossWorkspace.test.ts
@@ -142,6 +142,21 @@ describe('pane_get_metadata — cross-workspace read (#1018), behavioral', () =>
     }
   });
 
+  it('an empty-string paneId is rejected by the schema, not silently treated as "omitted" (active-pane fallback)', async () => {
+    // CodeRabbit, PR #1020: paneId: '' passed the old `paneId === undefined`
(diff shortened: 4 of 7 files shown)


openwong2kim:

Adds paneTitle to a2a_discover and an optional workspaceId for read-only cross-workspace pane_get_metadata, closing #1018. Changes are confined to useRpcBridge.ts and mcp/index.ts, additive by design — agentName and the pane_set_metadata write path are unchanged, with structural tests locking both. P3: low-risk discoverability fix, no schema or behavior breaks.

openwong2kim (changes requested):
The read-only/additive shape is right, but the identity gate doesn't survive the implementation — that one is a blocker.

**1. `targetWorkspaceId || (await requireWorkspaceId())` short-circuits the identity gate (blocker)**

The comment above it (and the PR body) say `requireWorkspaceId()` "still runs first", but with `||` it never runs when an override is present. `requireWorkspaceId()` is the thing that rejects callers whose workspace identity can't be determined — an identity-less external MCP caller that used to be stopped with "Workspace identity unknown" can now read any workspace's pane metadata just by passing an id from `a2a_discover` (or by probing ids blindly). And the new test ("prefers the explicit workspaceId over the caller's own") pins this exact pattern with a regex, so the bug is currently enforced by CI.

Fix:

```ts
const own = await requireWorkspaceId();   // identity gate always runs
const workspaceId = targetWorkspaceId ?? own;
```

and update the test to match.

**2. `workspaceId` without `paneId` silently reads the target workspace's *active* pane**

The schema text asks for `paneId` alongside the override but nothing enforces it, and `resolveTarget`'s active-pane fallback then depends on what the user happens to have focused in that workspace — a silently-wrong-target read. Require `paneId` whenever `workspaceId` is present and error otherwise.

Smaller:
- `||` also treats an empty-string `workspaceId` as absent, silently falling back to the caller's own workspace; `??` plus `.min(1)` on the zod shape closes both.
- The new test greps `index.ts` source with regexes rather than exercising behavior — it survives no refactor and (see above) happily pins a bug. An integration-style test that actually calls the tool with a foreign workspaceId would prove the feature.
- `scripts/mcp-protocol-baseline.json`: commander `maxListBytes` moved 48000 → 49000 without a note. That 48KB was a deliberate budget for the commander surface; if the longer tool descriptions pushed it over, either trim the text back under budget or call the bump out explicitly in the PR so it's a decision, not a drive-by.

The `paneTitle` addition to `a2a_discover` itself looks fine — titles are already sanitized (control chars stripped, 64-char cap) and were already exposed via `pane_list`. Worth one line in the tool description noting the title is a pane-chosen, untrusted string, so consuming agents treat it as data.

p-poppe:
Fixed all points. Thanks for catching the identity-gate bypass — that was a real hole, not a nitpick.

**1. Identity gate (blocker) — fixed.**
```ts
const own = await requireWorkspaceId();   // always runs
const workspaceId = targetWorkspaceId ?? own;
```
exactly as you specified. Verified this actually closes the hole, not just quiets the assertion: `git stash`'d this file back to the pre-fix code and re-ran the new test file — the identity-less-caller test fails against the old code (`isError` is `undefined`, meaning the RPC was reached instead of rejected), passes with the fix.

**2. `workspaceId` without `paneId` — fixed.** Now throws `"paneId is required when workspaceId is set"` before the RPC call, rather than falling through to `resolveTarget`'s active-pane fallback.

**3. Smaller items:**
- `||` → `??`, plus `.min(1)` on the `workspaceId` zod field, so an omitted and an empty-string value are both rejected consistently (no silent fallback to the caller's own workspace either way).
- Replaced the source-regex test with a behavioral one: a real `McpServer` from `createWmuxServer()`, talking to a real `Client` over `InMemoryTransport.createLinkedPair()`, with only `wmux-client`'s `sendRpc` mocked at the RPC boundary (identity resolution runs unmocked — `resolveWorkspaceId()`'s real fallback chain naturally resolves to `''` with no env hint and no daemon, and to the env hint when one is given and `workspace.list` is unreachable, which is exactly its documented behavior when the daemon is briefly down). Six tests: identity-less caller + foreign workspaceId is rejected pre-RPC, `workspaceId` without `paneId` is rejected, empty-string `workspaceId` is rejected, a valid override reads the *other* workspace, omitting `workspaceId` still reads the caller's own workspace unchanged, and a source lock confirming `pane_set_metadata` carries no override at all. Ran all three new assertions against the pre-fix code first to confirm they actually fail there (see point 1) rather than being vacuously true.
- `maxListBytes`: rather than bump it further (my new description text would have pushed the commander surface to 49238 bytes against the existing 49000 budget), I trimmed the added wording — commander surface now measures 48987 bytes, so the budget from your last review stays untouched. Re-ran `scripts/probe-commander-surface.mjs` clean (updated the two `wireResultSha256` values in the baseline for the new tool-description text; `instructionSha256` and tool names/ordering are unchanged).
- `a2a_discover`'s `paneTitle` field description now says it's untrusted, pane-chosen text — kept short to stay inside the byte budget above.

**One thing worth a second look:** the new behavioral test imports `@modelcontextprotocol/sdk/client/index.js` and `@modelcontextprotocol/sdk/inMemory.js` directly, which `import/no-unresolved` doesn't know how to resolve (same underlying gap as the existing ignore for `scripts/probe-commander-surface.mjs`, not a real unresolved import — `tsc` and vitest both resolve it fine). I extended `.eslintrc.json` with a second override scoped to just this test file rather than widening the existing one, so it doesn't loosen the rule anywhere else. Happy to fold it into the existing override's `files` array instead if you'd rather have one ignore list than two.

Full suite: `npx vitest run src/mcp src/renderer/hooks/__tests__/useRpcBridge.a2aPaneIdentity.test.ts` — 371 passed, 1 skipped (pre-existing, unrelated). `tsc --noEmit` clean. `eslint` on touched files clean except the one pre-existing `import/no-unresolved` on `src/mcp/index.ts` line 2 you already flagged as unrelated in the original PR description.

p-poppe:
Fixed — empty-string `paneId` had the same hole `workspaceId` did: it passed the `paneId === undefined` check and fell through to `resolveTarget`'s active-leaf fallback, silently reading whatever pane the target workspace's user happened to have focused. Same fix, same reasoning: `.min(1)` on the zod field, plus a test that fails against the pre-fix schema and passes now (confirmed by reverting the schema change locally and re-running).

p-poppe:
CI failure (`validate` — MCP protocol probe) was on me: the empty-paneId fix added a `minLength` constraint to the wire schema, which pushed the `commander` profile 1 byte over its 49000 budget (49001). Trimmed the `paneId` description (7 chars: "Omit to use the active pane in the calling workspace. Required when workspaceId is set." → "Omit for the active pane in the calling workspace. Required with workspaceId.") to land back under budget rather than raising it again, and refreshed both profiles' `wireResultSha256` to match. `commander` now measures 48991 bytes. Verified locally: `npm run build:mcp && node scripts/probe-commander-surface.mjs` passes clean.

openwong2kim (approved):
Verified the identity gate live, against a real stdio MCP server rather than the diff — and A/B'd it against the pre-fix bundle so the result isn't vacuous.

Built `dist/mcp-bundle` from this branch and from `HEAD~1` (the commit carrying `targetWorkspaceId || (await requireWorkspaceId())`), then drove both with a real MCP `Client` over stdio with `WMUX_WORKSPACE_ID` empty — i.e. exactly the identity-less caller the gate exists to stop.

Pre-fix, `pane_get_metadata({ workspaceId: 'ws-someone-else', paneId: 'pane-x' })` came back with `wmux auth token not found` — meaning the tool never consulted the identity gate at all and the call reached the RPC transport. It stopped there only because my probe process couldn't read the token; an agent running on the user's own box reads it fine, and the cross-workspace read would have completed.

Post-fix, the same call returns `Workspace identity unknown...`, rejected at the tool layer before any RPC. The empty-string case now fails schema validation (`.min(1)`) instead of silently falling back to the caller's own workspace, and with an identity present, `workspaceId` without `paneId` returns `paneId is required when workspaceId is set` — so the active-pane fallback really is closed on the cross-workspace path.

Replacing the source-regex test with the in-memory `McpServer`/`Client` pair was the right call — the old one was pinning the buggy line as if it were the contract. Trimming the descriptions to stay under the existing 48KB commander budget instead of raising it again is also the answer I was hoping for.

Approving.

PR #1022, opened 2026-08-24
feat(workspace): more color tags, and reach the pane header and tabs

p-poppe:
### Summary

Two extensions to the #927 workspace color tag (a purely visual, optional accent), both requested by a user running a dozen-plus workspaces at once:

1. **More colors.** Eight ran out fast once real usage hit ~10+ concurrent workspaces — two unrelated workspaces would end up sharing a tag with no way to tell them apart by color alone.
2. **The tag was only reachable from the sidebar.** Once you're inside a multiview tile or looking at a pane's tab strip, there was no visual link back to which workspace you're in.

### Changes

- **`src/shared/workspaceColors.ts`** — extends `WORKSPACE_COLOR_IDS` from 8 to 15: adds `amber, lime, mint, cyan, indigo, magenta, rose`. Each new hue is WCAG-contrast-checked against both bundled theme extremes (`#11111b` / `#eff1f5`) the same way the original eight were, and kept at least 12° apart in hue from every other color (test-enforced, see below) so no two tags read as "the same color" at a glance. `amber` sits between yellow (45°) and lime (83°) rather than near orange, since orange/yellow are only 16° apart — not enough room for a third hue within 12° of both.
- **`src/renderer/components/Sidebar/WorkspaceItem.tsx`** — the color picker now wraps at 8 per row instead of one long row, so 15 colors + the "none" swatch (16 total) don't risk running off the sidebar's edge.
- **`src/renderer/components/Layout/WorkspaceViewport.tsx`** — a multiview tile's header (the workspace name shown above its panes when multiple workspaces are on screen at once) now shows a small dot in the workspace's tag color before the name. No-op when the workspace has no tag (the common case).
- **`src/renderer/components/Pane/SurfaceTabs.tsx`** — each pane tab's label now carries a thin colored underline in the workspace's tag color. This is independent of the two existing signals on the same strip — the pane-focus indicator (an inset shadow on the *whole* header row) and the active/inactive tab background swap — so the three never compete for the same pixels. No-op when the workspace has no tag.
- **`src/renderer/i18n/locales/en.ts`** — adds labels for the 7 new ids, following the existing pattern of English-only labels with automatic fallback for the other 21 locales (same as the original eight — none of them carry translated color names either).

### Screenshots

Not attached — happy to add if useful; the change is small enough (a dot, an underline, 7 new swatches) that it may be faster to eyeball on a checkout.

### Test plan

- `src/shared/__tests__/workspaceColors.test.ts` — existing tests are generic over `WORKSPACE_COLOR_IDS` and passed unchanged. Added two new tests: every pair of colors is ≥12° apart in hue (would have caught `amber` landing 9° from `orange` on the first attempt — see commit history), and every color clears the same contrast floor the original eight live in.
- `src/renderer/components/Pane/__tests__/SurfaceTabs.colorTag.test.tsx` (new) — no underline for an untagged workspace, correct color for a tagged one, and the underline follows a workspace prop update (the reactive path a real parent re-render would take).
- `src/renderer/components/Layout/__tests__/MultiviewWorkspaceSlot.colorTag.test.tsx` (new) — same untagged/tagged pair for the tile header dot. Exported `MultiviewWorkspaceSlot` (previously module-private) to make this reachable, mirroring how `WorkspaceSlot` is already exported for its own memo test.
- Full suite: 12347 passed, 17 failed — all in `src/main/ipc/handlers/__tests__/worktree.handler.test.ts` (git-worktree cleanup + one timeout), untouched by this change and already flagged as pre-existing Windows flakiness on other recent PRs in this batch.
- `tsc --noEmit` clean, ESLint clean on all changed files.

### CHANGELOG entry

Will add `changelog.d/<this-pr-number>.md` once the PR number is known (per the repo's no-conflict convention).

## Summary by CodeRabbit

* **New Features**
  * Added seven workspace color tags: Amber, Lime, Mint, Cyan, Indigo, Magenta, and Rose.
  * Color tags now appear as clear indicator dots in multiview tile headers and pane-tab labels.
  * Workspace color options are displayed across two rows for easier selection.
  * Untagged workspaces retain their existing appearance.

* **Bug Fixes**
  * Improved color distinction and contrast across workspace tags.


openwong2kim:
This PR grows the workspace color tag palette from 8 to 15 WCAG-checked hues and shows the tag in multiview tile headers and pane tab underlines. Purely visual and additive — untagged workspaces render unchanged and unknown ids are dropped at the save boundary — so risk is low. P3: polish on #927 with no functional surface touched.

openwong2kim (commented):
Functionally this checks out — hue spacing, `normalizeWorkspaceColor` bounds, and the stated contrast numbers all verify, and en-fallback for the new keys matches the existing i18n convention. Two visual-grammar issues worth fixing, neither structural:

**1. The tab underline collides with the focus-cue grammar**

The 2px tag-colored underline on `SurfaceTabs` has the same shape and position as the focused-pane cue (`inset 0 -2px 0 var(--accent-blue)`). With the blue/indigo/cyan tags, tabs in *unfocused* panes now wear something that reads as the steel-blue "where you are" signal — DESIGN.md reserves that underline grammar exclusively for focus. It also only spans the truncated label width on long titles, so the signal changes size per tab. Suggest using the same dot treatment as the tile header instead of an underline.

**2. The tag dot is invisible on the active multiview tile**

The active tile header is filled with `var(--accent-blue)`, and against that fill all 15 tag colors land at ~1.00–1.45 contrast (teal and green at exactly 1.00) — the dot vanishes precisely on the tile you're looking at. The new contrast test only covers the theme extremes, not this fill. A 1px `var(--bg-base)` ring on the dot (and a test case for the accent-blue background) covers it.

Minor, take or leave: the popover's fixed `w-[164px]` / 8-per-row layout is tuned to exactly 15+none — a max-width + flex-wrap would absorb the next color addition without re-tuning.

openwong2kim:
Went ahead and pushed the two visual fixes from the review directly (8b901bfc), since they were small and mechanical: the tab strip now carries the tag as a single dot at its start (the underline shape is reserved for the focus cue, and the tag is workspace-level anyway), and the tile-header dot got a 1px `--bg-base` ring so it stays legible on the active tile's accent-blue fill. Tests updated to pin both. Will merge once CI is green — thanks for the feature!

Issue #546, opened 2026-07-22
launcher reuse path can SIGKILL a daemon that is still cold-recovering (not wedged)


openwong2kim:
### Context

Follow-up from the #537 investigation. #543 fixed the primary trigger (the *spawn* path now waits out a live-but-recovering daemon instead of timing out at 15s). This is the sibling hole on the *reuse* path, which #543 deliberately left out of scope.

### The hole

When `ensureDaemon` (`src/main/daemon/launcher.ts`) finds an existing `daemon.pid` that is alive but doesn't answer a ping, it walks the image + cmdline verification and, for a confirmed-wmux daemon, does a short escalated re-ping before SIGKILLing it:

```
tryEscalatedReping((t) => pingDaemon(pipeName, token, t), [500, 1000], 200, sleep)  // ~1.9s worst case
```

A daemon that is **cold-recovering a large session set** is alive, has written `daemon.pid` (at `acquireLock`, boot start), but has **not** opened its control pipe yet (the pipe file is written only after `recoverSessions`, `src/daemon/index.ts` ~3666 → ~3770, measured ~19–23s for 30–35 sessions). So it fails every ping. After ~1.9s of escalated re-ping the launcher SIGKILLs it and spawns a fresh one, which restarts the same multi-second recovery, and can thrash across the respawn budget.

### Why it's not urgent

After #543, the primary upgrade flow reaches the *spawn* path (old daemon already dead → spawn fresh → wait on child liveness), which no longer hits this. This reuse-path kill only bites when a recovering daemon is **discovered** rather than spawned by the current call, e.g. a fast app relaunch / double-launch over a daemon that is still cold-recovering from a previous boot. So it's a narrower, harder-to-hit case, but the same class of bug (treating "recovering" as "wedged").

### Suggested direction

Give the reuse path the same "recovering, not wedged" awareness the spawn path got in #543. The tricky part: on the reuse path the launcher didn't spawn the daemon, so it has no child handle to check liveness on. Options to weigh:

- Have the daemon expose a boot/recovery heartbeat (e.g. touch `daemon.pid` or a `daemon-booting` marker periodically during `recoverSessions`); the launcher treats a freshly-updated marker as "actively booting → keep waiting, don't kill", and only kills once the heartbeat goes stale.
- Or write the pipe file / start listening earlier and pre-register recovering sessions so the daemon can answer a ping while recovery continues (bigger change, overlaps with the renderer reconcile guards).

[...]

PR #1023, opened 2026-08-25
Give the #596 restart suite a witness the re-harden cannot move

openwong2kim:
## The bug in the test

`webRestart.runtime.test.ts` asserted that a restore left `web-state.json`
untouched by checking its mtime, on the premise its own comment stated —
*"Async ACL re-hardening leaves mtime put."*

It does not. `scheduleTokenFileReHarden` (`src/shared/security.ts`) hands the
file to `rewriteThroughFreshInodeAsync`, which stages a replacement and
`renameSync`s it into place. Measured directly on an already-hardened file:

```
mtimeMs before : 1787586542106.389  ino: 87257242780872620
t=+50ms          1787586542106.389  ino: 87257242780872620    SAME
[security] deferred token ACL re-harden took 80ms (web-state.json)
t=+150ms         1787586542188.7268 ino: 12666373952643324    *** MOVED ***
```

Both mtime and inode move ~80 ms into every boot, whether or not the restore
wrote anything. The assertion passed only by observing the file before that
rewrite landed — a `sleep(1500)` ahead of the same line failed it every time:

```
AssertionError: expected 1787586732240.7578 to be 1787586730197.0908
```

So the check was a coin flip on how long the rest of the test happened to
take, and the TLS case — which does more work before it looks — is the one
that came up red.

## The fix

Assert on the daemon's own log, which is the only witness that says *which
path ran*. A redundant `saveWebState` prints a **synchronous** harden line for
that file; the deferred re-harden a restore does perform prints a different
one. The suite now waits for the deferred line (which also proves the restore
path ran at all), then requires that no synchronous harden of `web-state.json`
happened.

Negative control — forcing the redundant rewrite in `restoreWebServer`:

```
× restores the listener AND the token after the daemon is killed and replaced
× restores native HTTPS without key material or plaintext downgrade
AssertionError: expected [ Array(1) ] to deeply equal []
```

Both cases now catch it. Under the mtime check, whether either one caught it
depended on timing.

## The reason nobody saw it

Neither this suite nor the #546 boot-marker suite has ever run in CI. Both
skip when `dist/daemon-bundle/index.js` is absent, and no workflow ever built
it — `validate` runs `build:mcp`, never `build:daemon`. A skip reads as green,
so both files have been decorative since the day they were written.

`validate` now builds the bundle, and declares `WMUX_REQUIRE_DAEMON_BUNDLE=1`.
Each suite carries a prerequisite test, deliberately outside its own skip
guard, that fails when the variable is set but the bundle is missing. Deleting
the build step turns both files red instead of quietly retiring them.

## Verification (Windows, real bundled daemon)

- `webRestart` + `bootMarker`, bundle present: **11 passed**
- Bundle hidden, `WMUX_REQUIRE_DAEMON_BUNDLE=1`: **2 failed, 9 skipped** — the
  guard fires, nothing pretends to pass
- Forced redundant persist: **2 failed** (both restore cases)
- `tsc --noEmit` clean; `eslint --max-warnings=0` clean on both files

No product code changes.

https://claude.ai/code/session_01XJtrPSP68cZPmp2mUyJqGX

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon restart validation to confirm deferred security hardening completes successfully.
  * Added coverage for both plaintext and native HTTPS restart scenarios.

* **Testing**
  * Windows CI now builds the daemon bundle before testing.
  * CI fails when the required daemon bundle is unavailable instead of silently skipping runtime tests.
  * Enhanced diagnostics capture daemon output to make restart failures easier to identify.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,5 +57,17 @@ jobs:
       - name: Type check push relay
         run: node_modules/.bin/tsc --noEmit -p relay/tsconfig.json
 
+      # `npm test` ends in `test:runtime`, and two of those suites drive the
+      # REAL bundled daemon (#546 boot marker, #596 web restore). Both skip
+      # when `dist/daemon-bundle/index.js` is absent — which it always was
+      # here, so neither has ever run in CI. A skip reads as green, so this
+      # step is what gives those suites any signal at all.
+      - name: Build daemon bundle (the runtime suites drive it)
+        run: npm run build:daemon
+
       - name: Test
         run: npm test
+        env:
+          # Tells those suites the bundle was promised, so removing the step
+          # above fails a test instead of silently retiring both files.
+          WMUX_REQUIRE_DAEMON_BUNDLE: '1'
--- a/src/daemon/__tests__/bootMarker.runtime.test.ts
+++ b/src/daemon/__tests__/bootMarker.runtime.test.ts
@@ -40,6 +40,18 @@ afterAll(() => {
 // fresh checkout without `npm run build:daemon` doesn't red the suite.
 const hasBundle = fs.existsSync(bundle);
 
+/** Set by the CI job that promises to build the bundle first (see
+ *  webRestart.runtime.test.ts). Outside the skipIf below on purpose: it has to
+ *  run precisely in the case where the suite itself cannot, so a workflow that
+ *  stops building the bundle goes red instead of quietly retiring this file. */
+const bundlePromised = process.env.WMUX_REQUIRE_DAEMON_BUNDLE === '1';
+
+describe.runIf(bundlePromised)('#546 boot marker suite prerequisites', () => {
+  it('has the daemon bundle the promising job was supposed to build', () => {
+    expect(hasBundle).toBe(true);
+  });
+});
+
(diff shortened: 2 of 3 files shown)


openwong2kim:
Replaces the #596 webRestart suite's flaky mtime assertion with a log-based witness — the deferred re-harden must appear, a synchronous harden must not — and builds the daemon bundle in CI so both runtime suites actually run. Test- and CI-workflow-only; no product code touched. P3 — pure test quality fix, no user-facing risk.

openwong2kim:
CI now actually runs them. Same job, before and after:

**main** (run 32741324074, `validate`):
```
↓ src/daemon/web/__tests__/webRestart.runtime.test.ts (7 tests | 7 skipped)
↓ src/daemon/__tests__/bootMarker.runtime.test.ts   (2 tests | 2 skipped)
```

**this PR** (run 32793853278, `validate`):
```
✓ src/daemon/web/__tests__/webRestart.runtime.test.ts (8 tests) 24156ms
✓ src/daemon/__tests__/bootMarker.runtime.test.ts    (3 tests)   440ms
```

Nine skips became eleven executed tests, and the 24 s is a real daemon being
started, killed and restarted twice. `validate` went 6m44s → 8m12s.

All checks green.

PR #1024, opened 2026-08-25
fix(usage): let the expired-token status clear itself after a re-login

### LINKED ISSUE #1012: [Feature]: 'Token expired' message to have a timeout or update its logic
### Problem / motivation

Currently even when I'm logged in into claude, I still get token expired message that just hangs in the top of the screen indefinitely.

My wmux version is 3.46.0
Windows is Windows 10 19045.6466
Installed wmux via direct github install

https://github.com/user-attachments/assets/31d3c25a-db91-4d5c-a252-f80f7e954d18

### Proposed solution

Ideally I'd want it to have some kind of timeout (appear for 1-2 minutes and then dissapear even if token is still expired) or update the logic to properly catch your authentication status.

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

snowyukitty:
Closes #1012.

## The bug

`UsagePoller` treats a 401/403 from the usage probe as terminal:

```ts
// src/main/claude/UsagePoller.ts (before)
if (err.detail.kind === 'unauthorized') {
  // STOP the interval — the credential is bad and pummeling
  // Anthropic with 401s helps nobody. User must re-login then
  // re-toggle.
  this.stop();
  this.setState({ status: 'unauthorized', ... });
```

The reason it gives is right — re-sending a credential Anthropic just refused
helps nobody. The instrument is what fails: `stop()` clears the interval, so
after the reporter logged back into Claude Code there was nothing left running
to notice. The titlebar chip kept reading `● Token expired — log in to Claude
Code again` indefinitely, which is what #1012 shows.

Nothing else recovers it either:

- `setWindowVisible(true)`'s catch-up kick is gated on `this.timer`, which is
  now null — so hiding and re-showing the window does nothing.
- The only ways out are the Settings toggle off/on and the manual refresh
  button, neither of which announces itself from a chip that just says the
  token expired.

The wedge is also more likely than it looks in wmux specifically: the re-login
usually happens *in a wmux pane*, so the window never hides, and this is a
long-lived app — the chip sits there for the rest of the session.

`AccountUsageService` (the per-account meter in Settings) does not have this
shape: its probes are driven by `agent.stop` past a cooldown, so an
`unauthorized` entry there re-probes on its own. Only the global poller latches.

## The fix

Keep the interval armed and pin the token that was refused.

- `rejectedAccessToken` records the `accessToken` Anthropic answered 401/403
  for. A tick that reads that same token back off disk returns *before*
  `fetchUsage`, so the property `stop()` was protecting is now enforced by
  token identity instead of by killing the loop.
- Two bounds release that skip, one for each cause a 401 can have. If the
  token really is bad, no request can succeed until the credential changes,
  and the recheck notices that within minutes for free. If the refusal was
  **not** about the token — an auth-service blip, an org policy since reverted
  — then no credential will ever change, and pinning on identity alone would
  strand the chip exactly the way `stop()` did. So the skip also expires after
  `intervalMs`: one retry per poll period, which is precisely what a healthy
  poller spends. An unauthorized poller can therefore never cost more traffic
  than a working one, and it recovers from either cause without the user doing
  anything.
- The skip is those two facts and nothing else — not what the chip currently
  reads. Keying it on the displayed status as well looked safer and was not: a
  credential read that fails once moves the status off `unauthorized` while the
  credential underneath is unchanged, and the skip would then release early and
  re-send the very token it exists to withhold. Nor does the pin strand any
  other status, because every one of them recovers by way of a *different*
  credential, and a different token fails the comparison on the first tick that
  reads it.
- `cadenceMs()` derives the interval period from that same freshness window —
  the recheck period while a refusal is fresh, the ordinary poll otherwise —
  and `syncCadence()` is the only thing that re-arms, once, at the end of a
  tick. So the fast cadence cannot outlive its evidence, and no outcome can
  leave the wrong period behind. `arm()` is the single owner of the timer;
  re-arming at the period already running is a no-op.
- A credential that reads as missing for one tick **keeps** the pin. That read
  is what a re-login looks like mid-write, and dropping the fast cadence there
  would strand the new token for a whole poll interval at the exact moment
  recovery was arriving. It would also release the skip on the tick after
  that, since a credential coming back unchanged is not a reason to re-send
  it.

The recheck means there are now background ticks in a state that previously
had none. That is what makes the lifecycle rules below load-bearing: before
this, a tick had nowhere to land except the hourly poll every other state
already ran.

- A tick captures a generation before it does anything else — including
  before any wait — and abandons its work if `start`/`stop`/`dispose` bumped it
  across an await. Without that, a `stop()` landing inside a recheck's
  credential read resumes into a cleared pin, finds the skip no longer true,
  and sends the very token the skip exists to withhold — after the user turned
  the meter off. (The same guard stops a fetch from repainting a status over
  the `idle` that `stop()` just set, which is a pre-existing race on `main`:
  both are the same missing check.)
- The re-entry guard yields to a peer but not to a corpse. Standing down for a
  pass from the current session is right — it is about to produce the same
  answer. Standing down for one left behind by a session that has since been
  stopped is not: that pass abandons its own result at its first `isStale`
  check, so a toggle off and back on during a slow read used to cost the new
  session its opening fetch and leave the widget empty for a whole interval.
- A forced tick never stands down; it waits for the slot in a loop, so several
  queued clicks each get their own pass. Ordinary ticks coalesce instead: if
  the slot was used by a pass of their own generation while they waited, that
  pass already asked their question, and N interval firings piled up behind one
  slow read stay one request rather than N.

`refreshNow()` runs its tick with `force`. An explicit click outranks both
skips: it re-sends even a pinned token (the user may have fixed something the
credential file cannot show), and it is honored while the window is hidden past
the skip threshold — which is what that skip's own comment already claimed and
the code did not do.

## What did not change

- No new status, no change to `PollerState`, no IPC or schema change. The
  renderer is untouched.
- `token-missing`, `read-error`, `http-error` and `network-error` keep their
  existing retry semantics.
- The hourly cadence, the hidden-window cost control, and the 30 minute skip
  threshold are unchanged for every state except a fresh refusal.

## Three things I looked at and deliberately left alone

- **`setState` delivers a superseded state to later subscribers** if an earlier
  subscriber calls `stop()` from inside its callback: the nested broadcast
  lands first, then the outer loop continues handing out the old value. It is
  pre-existing, unreachable with the one production subscriber (which forwards
  to the renderer over IPC), and the fix belongs to `setState` rather than to
  this change. Happy to send it separately.
- **The elapsed bound reads wall clock** (`this.now()`, like the hidden-window
  threshold above it), so a large clock correction can release the retry early
  or hold it late. The consequence is bounded at one extra probe or one delayed
  one, and the primary recovery path — the credential changing — does not
  depend on the clock at all. A second, monotonic clock for this seemed worse
  than the exposure.
- **A waiting tick blocks until the in-flight pass settles.** Both awaits it can
  block on are externally bounded — a 5s `execFile` timeout on the macOS
  credential read, a 20s abort on the request — so this is a bounded wait and
  not a barrier a wedged read could hold open forever.

## Test plan

- `src/main/claude/__tests__/UsagePoller.test.ts` — 27 cases, 14 of them new or
  rewritten. The ones that carry the argument:
  - `clears unauthorized by itself once Claude Code writes a new token (#1012)`
    — the regression test. Asserts the recovery fetch carried
    `Bearer fresh-token`, i.e. it is the new credential that went out, and that
    the poller returned to the normal cadence afterwards.
  - `keeps re-reading the credential on every recheck while it withholds the
    request` — counts credential reads as well as requests, so "it stays armed"
    is proven rather than assumed.
  - `retries a still-refused token once per poll interval, not once per
    recheck` — the second bound, from both sides, across two full periods.
  - `does not let a momentary credential miss cost a whole poll interval` and
    `keeps the fast cadence through an unreadable credential too` — the
    re-login window, which is exactly when a read is most likely to fail.
  - `still withholds a refused token that reappears unchanged after a failed
    read` — the skip is the pin and its age, not the chip's current text.
  - `abandons a tick whose session was stopped while it read the credential`
    and `... while it awaited the request` — the lifecycle guard on both awaits.
  - `does not run a queued tick for a session that was stopped while it waited`
    — the guard on the wait itself.
  - `gives every queued manual refresh its own pass, not just the first`,
    `does not let a stopped session's in-flight tick swallow the next session's
    first fetch`, and `coalesces the interval ticks that pile up behind a
    stopped session's read` — the three re-entry rules.
  - `refreshNow() re-sends even a token that was already refused` and
    `refreshNow() is honored while the window is hidden past the skip
    threshold`.
- Red/green verified by putting `origin/main`'s `UsagePoller.ts` under the new
  test file: 12 fail, the other 15 pass. With the fix, 27/27. Every behavioural
  change above was also checked red against the revision immediately before it,
  so nothing in here is passing for a reason other than the code it names.
- `npx vitest run src/main/claude src/main/account` — 120 passed.
- `npx tsc --noEmit -p tsconfig.json` — clean. `eslint` on both files — clean.
- One drive-by, called out so it is not mistaken for part of the fix: the
  pre-existing `fires an immediate fetch on start (queueMicrotask, not
  interval)` was renamed to say `setTimeout(0)`, which is what `start()`
  actually uses and has used since it was written.

## Known limit, and the follow-up it suggests

An account whose token works for Claude Code but that the usage probe is
refused for — wrong beta scope, an org restriction, a non-subscription login —
keeps the chip up. That is right, I think: it is a status readout, not a toast,
and the meter genuinely cannot read usage. But the *wording* is then wrong,
because nothing has expired.

`ClaudeCredential` already carries `expiresAtMs`, so the two are separable: a
401 on a credential whose stored expiry is still in the future is "Anthropic
refused this token", not "Token expired". I left that out of this PR on purpose
— it is a new string in 20 locales and a new `PollerStatus`, which is a
different review from this one. Happy to open it as a follow-up if you want it,
and it would be worth asking the reporter of #1012 whether their token ever
changed, since that is what tells the two cases apart.

## Stability tier impact

None. `UsagePoller` is main-process internal; `unauthorizedRecheckMs` is a new
constructor option with a default, used by tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Anthropic usage monitoring now continues after an unauthorized response instead of stopping.
  * Rejected credentials are temporarily withheld to prevent redundant requests.
  * Credentials are rechecked periodically and retried sooner when updated.
  * Expired-token notices clear automatically after a valid token is detected.
  * Forced refreshes continue to work even while normal polling is paused or skipped.

### DIFF
--- a/changelog.d/1024.md
+++ b/changelog.d/1024.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **"Token expired" clears itself once you log back in.** The Anthropic usage
+  meter stopped polling the moment Anthropic refused its token, so the titlebar
+  kept reading `Token expired — log in to Claude Code again` after a successful
+  re-login — there was nothing left running to notice. It now keeps looking, and
+  the notice goes away on its own within a few minutes of Claude Code writing a
+  new token. It still will not sit there re-sending a credential it already
+  knows is bad: while the notice stands, a check that finds the same token on
+  disk stops before the request, and a genuine retry happens no more often than
+  the ordinary hourly check would have — so a refusal that was never about your
+  token clears itself too. The refresh button still forces a real attempt
+  whenever you want one. (#1024)
--- a/src/main/claude/UsagePoller.ts
+++ b/src/main/claude/UsagePoller.ts
@@ -13,8 +13,17 @@
 //     called or the credential file is observed to appear (future
 //     follow-up — for now manual refresh is the recovery).
 //   - 401/403 (token expired or revoked) → emit 'unauthorized' status
-//     and STOP the interval. User must re-login and toggle Settings
-//     off/on (or click manual refresh).
+//     and remember the token that was rejected. The interval KEEPS
+//     RUNNING, on a shorter cadence, but a tick that reads back that
+//     same rejected token returns before the network call — so we
+//     never re-send a credential we already know is bad, and the
+//     status clears by itself as soon as Claude Code writes a new one
+//     (a re-login, or its own token refresh). That skip also expires
+//     after `intervalMs`, so a 401 that was never about the token at
+//     all still gets one retry per poll period — the same traffic a
+//     healthy poller spends, and never more. Stopping the interval
+//     outright was the old behaviour, and it left the widget reading
+//     "Token expired" forever after a successful re-login (#1012).
 //   - Network / 5xx → emit 'network-error' / 'http-error' with the
 //     last error. The interval KEEPS RUNNING; next tick is the retry.
 //     Avoids the failure mode where a transient outage permanently
@@ -36,7 +45,9 @@ export type PollerStatus =
   | 'ok'
   /** Claude Code is not logged in / credential file missing. */
   | 'token-missing'
-  /** Anthropic returned 401/403. Poller paused. */
+  /** Anthropic returned 401/403. The poller keeps running, but it will
+   *  not re-send the token that was refused until the credential on
+   *  disk changes (or a whole poll interval has passed). */
   | 'unauthorized'
   /** Non-auth HTTP failure. Poller keeps running, retrying. */
   | 'http-error'
@@ -66,6 +77,20 @@ export interface PollerOptions {
   intervalMs?: number;
   /** Skip a tick if the window has been hidden longer than this. Default 30 min. */
   hiddenSkipThresholdMs?: number;
+  /** While a refusal is still fresh, re-read the credential this often
+   *  instead of waiting out `intervalMs`. A recheck costs one
+   *  credential read and sends nothing unless the token changed (or a
+   *  whole `intervalMs` has passed since the refusal), so it can be
+   *  far shorter than the poll interval. Clamped to `intervalMs`,
+   *  since a recheck slower than the poll it stands in for would only
+   *  lengthen the wrong-status window. Default 5 min.
+   *
+   *  "Fresh" is the pin's age, not anything about the displayed
+   *  status: a credential that reads as missing for one tick during a
+   *  re-login moves the status to 'token-missing' without giving up
+   *  either the fast cadence or the skip, because that tick is
+   *  precisely when recovery is landing. */
+  unauthorizedRecheckMs?: number;
   /** Injectable for tests. */
   now?: () => number;
   /** Injectable for tests. */
@@ -76,6 +101,7 @@ export interface PollerOptions {
 
 const ONE_HOUR_MS = 60 * 60 * 1000;
 const THIRTY_MIN_MS = 30 * 60 * 1000;
+const FIVE_MIN_MS = 5 * 60 * 1000;
 
 /**
  * Owns a single in-process interval. The poller is created once and
@@ -86,6 +112,7 @@ const THIRTY_MIN_MS = 30 * 60 * 1000;
 export class UsagePoller {
   private readonly intervalMs: number;
   private readonly hiddenSkipThresholdMs: number;
+  private readonly unauthorizedRecheckMs: number;
   private readonly now: () => number;
   private readonly fetchImpl: typeof fetch;
   private readonly loadCredential: () => Promise<LoadResult>;
@@ -97,8 +124,36 @@ export class UsagePoller {
     subscriptionType: null,
   };
   private timer: ReturnType<typeof setInterval> | null = null;
+  /** Period the live interval was armed with, so `arm()` can tell a
+   *  cadence change from a no-op re-arm. Null exactly when `timer` is. */
+  private timerPeriodMs: number | null = null;
   private immediateTimer: ReturnType<typeof setTimeout> | null = null;
+  /** The access token Anthropic answered 401/403 for. A tick that reads
+   *  this same token back off disk returns without spending a request.
+   *  It deliberately outlives the 'unauthorized' chip: a credential read
+   *  that fails mid-relogin moves the status on without making the token
+   *  underneath any more sendable. Cleared by start(), by stop(), and by
+   *  any tick that gets past the skip — never by a status change. */
+  private rejectedAccessToken: string | null = null;
+  /** When that answer came back. The skip above expires after
+   *  `intervalMs`, so a 401 that was never about the token (an auth
+   *  outage, an org policy since reverted) still gets one retry per
+   *  normal poll period — never more traffic than a healthy poller. */
+  private rejectedAtMs = 0;
   private inflight = false;
+  /** The run currently in flight, so a tick that must not be dropped
+   *  can wait the slot out instead of returning. */
+  private inflightRun: Promise<void> | null = null;
+  /** The generation that run belongs to. A tick only stands down for a
+   *  pass from its own session; one left behind by a stopped session is
+   *  going to abandon its result anyway. */
+  private inflightGeneration = -1;
+  /** Bumped by every start(), stop() and dispose(). A tick captures it
+   *  on entry and abandons its result if the session it belonged to
+   *  ended while it was awaiting. Without it a fetch can outlive its
+   *  own poller: repainting a status the toggle already turned off, or
+   *  sending a credential the recheck deliberately withholds. */
+  private generation = 0;
   private windowVisible = true;
   private windowHiddenAtMs = 0;
   private disposed = false;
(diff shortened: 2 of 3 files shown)


openwong2kim:

Fixes #1012: the UsagePoller stays armed after a 401/403 and pins the rejected token, so the "Token expired" chip clears itself after a re-login. Touches only main-process UsagePoller plus tests and changelog; the pin and retry bounds cap extra traffic at one request per poll period. P3 — main-process only, and the new tests cover the lifecycle races.

snowyukitty:
Two checks are red and neither is reachable from this diff, which touches only `src/main/claude/UsagePoller.ts` and its test file. Recording what they were rather than asking for a re-run:

- **Baseline (windows-latest)** — `src/main/ipc/handlers/__tests__/worktree.handler.test.ts:91`, `Test timed out in 5000ms` followed by `EBUSY: resource busy or locked, rmdir ...\wmux-wth-Wul66m\repo`. Git worktree cleanup losing a race with a Windows file lock.
- **validate** — `src/daemon/__tests__/bootMarker.runtime.test.ts` `#546 boot marker lifecycle (real daemon)`: `expected NaN to be 5136`, i.e. the poll never observed the marker file during boot, so `parseInt('')` gave `NaN`.

Both pass locally on Windows 11 against this branch:

```
npx vitest run src/main/ipc/handlers/__tests__/worktree.handler.test.ts   → 12 passed
npx vitest run src/daemon/__tests__/bootMarker.runtime.test.ts            → 2 passed, 1 skipped
```

The worktree one has the same signature #1019 reported hitting on Windows. I have not seen the boot-marker one before, and its shape — a real daemon whose marker was never caught inside the window — reads as timing rather than logic, but I can only say it is not this diff, not that it is nothing.

What this branch does cover: `npx vitest run src/main/claude src/main/account` → 120 passed, `tsc --noEmit` and `eslint` clean, and the 12-red/15-green red-green check described in the PR body.

openwong2kim:
Reviewed this closely — the core design is genuinely good. Keeping the interval armed while the pin (token identity + age) decides whether a request goes out means retry traffic can never exceed a healthy poller's, and `arm()` as the single timer owner closes the leak/double-schedule risks that usually come with cadence switching. The generation guards and the coalescing/waiter logic also held up under scrutiny.

One behavior question before merging, because I can't tell from the code whether it's a deliberate trade-off or a gap:

**After a 401, a transient credential read failure strands the chip on the wrong status.** Sequence: 401 pins the token and the chip reads `unauthorized` → one tick hits `EBUSY` (or any transient read failure) → chip moves to `read-error` → the next recheck reads the *same rejected token* back cleanly, takes the skip branch — and returns without touching state. The poller knows at that moment that the operative cause is still the 401, but the chip keeps saying `read-error` for up to `intervalMs`.

Your own test walks this exact sequence — `still withholds a refused token that reappears unchanged after a failed read` asserts the fetch count and (per its comment, "however the chip currently reads") deliberately doesn't re-assert the status. I follow the reasoning that the *skip decision* must not key on the displayed status — the comment defending that is right. But the display is a separable axis: the skip branch could restore `unauthorized` on its way out without the skip itself ever consulting the status.

If leaving the display as-is was intentional (both statuses do point the user at roughly the same remedy), happy to merge as-is — just say so. If it's a gap, it looks like a ~3-line restore in the skip branch plus one assertion in that test; also fine as a follow-up after merge if you prefer.

Three smaller observations, none blocking, take or leave:

- **Persistent-401 cadence never decays.** If the token stays refused (revoked account, org policy), each `intervalMs` retry re-pins and re-opens the fast window, so the 5-minute recheck runs indefinitely — 12× the credential reads per hour, and on macOS each read is a `security find-generic-password` subprocess. The "never more traffic than a healthy poller" argument counts network requests, which it wins; it doesn't count credential reads. An exponential decay toward `intervalMs` after N same-token re-pins would cap it.
- **`rejectedAccessToken` holds the raw OAuth token in a long-lived field**, and in the "401 wasn't about the token" case that's a still-valid credential sitting in the heap. It's only ever compared for identity, so a sha256 digest would behave identically.
- **`unauthorizedRecheckMs` has an upper clamp but no lower/NaN guard.** Unreachable today (index.ts passes no options), but as an exported option a `Math.max(1000, …)` + `Number.isFinite` check is cheap.

CI note: the earlier `validate` / windows-Baseline failures were the known `bootMarker.runtime.test.ts` timing flake — both passed on re-run, nothing here.

openwong2kim (approved):
Merging as-is — the design holds up under the scrutiny it got (the pin-and-age skip means retry traffic can never exceed a healthy poller's, `arm()` closes the timer-leak class, and the generation guards survived a deliberate hunt for interleavings), and the one behavior gap from my earlier comment is bounded: the mislabeled chip after a 401 + transient read failure self-corrects within one poll interval, both labels point the user at credentials either way, and it is strictly narrower than the permanent wedge this PR removes. Happy to take the ~3-line status restore (and any of the three smaller notes) as a follow-up whenever you get to them — none of it needed to hold this back.

PR #1026, opened 2026-08-25
feat(pane): fold the auto label on single-tab panes, rename from the menu

### LINKED ISSUE #1021: Pane header: the pane label duplicates the surface tab title on single-tab panes, and isn't clickable
### Problem / motivation

Two related things that both cost width in a narrow pane, filed together because they're the same underlying complaint ("I run many narrow vertical splits and every pixel of chrome matters").

**1. The pane label is dead weight when a pane has exactly one surface tab.**

Every pane header shows two independent names side by side: the pane label (`w{ws}-{ordinal}(agent)`, or your own rename) and the surface tab title (e.g. `WMUX`) — two different stores (`MetadataStore` vs `surface.title`), intentionally kept separate so their renames never collide. That's a reasonable design when a pane holds *multiple* tabs and you want to name the pane itself independently of any one tab. But the common case is one terminal, one tab — and there the pane label is a second name for the same thing the tab title already says.

It's also not a live control: a single click on the label does nothing (`data-pane-label` only wires `onDoubleClick` for rename — no click handler, no tab-switch). So in the single-surface case it's reserved width (`max-w-[170px]`, `shrink-0`, ahead of the scroll region) that a user can look at but not otherwise interact with, duplicating information the adjacent tab already carries.

Not proposing to remove pane naming — it's clearly load-bearing once a pane has more than one surface. The ask is narrower: when `surfaces.length === 1`, could the pane label fold into (or sit inline with) that one tab instead of reserving its own segment? Interested in your read on whether that's a real simplification or whether there's a reason (e.g. multi-surface panes gaining/losing a tab shouldn't cause a layout jump) that keeps them visually separate always.

**2. (Confirmed NOT an issue, for the record) — the pane-move grip (`::`, #645) stays as-is.** Raised it in the same pass since it's adjacent chrome, but it's load-bearing (drag-to-move, and the only right-click surface reachable at any width) — no change requested there.

### Additional context

Found while looking at #997 (roster row width) from the other direction — same instinct ("does this row need two things where the eye only needs one"), this time in the pane header instead of the sidebar.

### Confirmation

- [x] I searched existing issues and didn't find a duplicate.

openwong2kim:
Closes #1021 (item 1; item 2 was confirmed no-change in the issue).

## What

With exactly one surface tab and no user rename, the pane label now folds away. It comes back the moment any of these hold:

- the pane gains a second tab (the layout shift lands on a deliberate structural act, not on idle chrome — the concern raised in the issue),
- the user names the pane (explicit intent to see it),
- the rename editor is open.

## The affordance the fold removes, and its replacement

Double-click-on-label was the *only* way to name a pane — folding the label on unnamed single-tab panes would have made them permanently unnameable. The pane-actions menu now carries a **Rename pane** item (`startPaneRename`, same editor), which is reachable at any width via the header right-click and arguably more discoverable than the double-click ever was. New `pane.rename` key added to all 23 locales, following the full-coverage convention of the other `pane.*` menu keys.

## Test contract changes

- New `SurfaceTabs.paneLabel.test.tsx` (4 tests): hidden / user-named / multi-tab visibility states, plus menu-rename opening the editor on a label-less pane.
- The cluster/menu parity test (`paneClusterWidth.test.ts`) becomes directional: every cluster action must exist in the menu — the menu is the surface that must never lose an action, since at ⋮-only widths it is all the user has — and the menu-only set is pinned to exactly `['rename-pane']` so future menu-only additions are a decision, not drift.
- `SurfaceTabs.actions.test.tsx`: menu key list and arrow-walk indices updated for the sixth item; the native-context-menu-on-rename test names the pane first, since an unnamed single-tab pane no longer renders the label it double-clicks.

## Test plan

- `npx tsc --noEmit` — clean
- Pane `__tests__` + i18n suites: 22 files, 312 passed

## Summary by CodeRabbit

* **New Features**
  * Added a **Rename pane** action to the pane-header context menu.
  * Pane names can now be edited even when the pane label is hidden.
  * Automatic labels are hidden for unnamed, single-tab panes and remain visible for named or multi-tab panes.
  * Added localized rename controls across supported languages.

* **Tests**
  * Added coverage for pane-label visibility, renaming, and menu navigation.

### DIFF
--- a/changelog.d/1026.md
+++ b/changelog.d/1026.md
@@ -1,0 +1,7 @@
+### Changed
+
+- The pane header no longer shows the auto label (`w2-1`) on a pane with a single tab and no custom name — the tab title already says it. Naming the pane, adding a second tab, or opening the rename editor brings it back. (#1021, #1026)
+
+### Added
+
+- **Rename pane** in the pane-actions menu (right-click the pane header) — pane renaming no longer requires knowing the double-click gesture. (#1026)
--- a/src/renderer/components/Pane/SurfaceTabs.tsx
+++ b/src/renderer/components/Pane/SurfaceTabs.tsx
@@ -12,7 +12,7 @@ import { computePaneAutoName, paneDisplayName } from '../../utils/paneNaming';
 import { findPane } from '../../../shared/paneUtils';
 import PaneDragGrip from './PaneDragGrip';
 import { FOCUS_RING } from '../focusRing';
-import { IconSplitRight, IconSplitDown, IconBrowser, IconEyeOff } from '../icons';
+import { IconSplitRight, IconSplitDown, IconBrowser, IconEyeOff, IconPencil } from '../icons';
 import { displayPath } from '../../utils/displayPath';
 import { workspaceColorHex } from '../../../shared/workspaceColors';
 import PaneActionsMenu, { PANE_ACTIONS_MENU_WIDTH, type PaneActionItem } from './PaneActionsMenu';
@@ -402,12 +402,7 @@ export default function SurfaceTabs({
     {
       key: 'rename-pane',
       label: t('pane.rename'),
-      // Zoom's precedent: a mono glyph span where no drawn icon exists yet.
-      icon: (
-        <span aria-hidden="true" className="font-mono text-[13px] leading-none">
-          ✎
-        </span>
-      ),
+      icon: <IconPencil size={14} />,
       onSelect: startPaneRename,
     },
     {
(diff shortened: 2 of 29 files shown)


openwong2kim:

Folds the auto pane label on single-tab panes and adds a Rename pane item to the pane-actions menu, closing #1021 item 1. Renderer-only: SurfaceTabs.tsx plus a `pane.rename` key in all 23 locales, with visibility and menu-parity tests updated. P3 — UI polish with no daemon or IPC surface, and the fold conditions are well guarded.

Issue #1025, opened 2026-08-25
killVerifiedDaemonPid can SIGKILL an unrelated process whose argv contains daemon/index.js


openwong2kim:
`killVerifiedDaemonPid` decides whether a PID belongs to wmux by looking for generic markers in the process command line. Since #1019 that check is segment-aware rather than a raw substring match, which removed one class of false positive — but the underlying ambiguity is still there, and it is demonstrable.

## What still matches

`src/shared/daemon/daemonLauncherCore.ts`, `cmdlineMatchesDaemonScript`:

```ts
if (segments[i].startsWith('daemon-bundle')) return true;
if (segments[i] === 'daemon' && segments[i + 1] === 'daemon' && segments[i + 2] === 'index.js') return true;
if (segments[i] === 'daemon' && segments[i + 1] === 'index.js') return true;
```

Any process whose argv carries `<anything>/daemon/index.js` satisfies the last rule. That is an extremely common project layout, so it is not a wmux-specific signal at all.

## Reproduced by execution

Spawned an unrelated `node <tmpdir>/daemon/index.js` sleeper (a stand-in for someone else's program that happens to use that layout), then called:

```ts
killVerifiedDaemonPid(pid, { definitiveOnly: false })
```

Result: returned `true`, and the innocent process was dead afterwards. The image-name check does not save this case — a plain `node` daemon and a plain `node` third-party process share the same image basename.

## Severity

Low in practice, and **not a regression** — before #1019 the same-image path reached the same markers via `.includes()` and killed it too. Reaching this code requires wmux to already believe the PID is its daemon, so a live false positive needs PID reuse (our daemon dies, the OS hands its PID to a process with that layout). Narrow, but it is exactly the "tier-1 wtf is wmux doing" outcome the original comment on `getProcessImageName` was written to prevent.

## Suggested fix

Stop pattern-matching. The caller already resolved the daemon entry script — `DaemonLauncherDeps.daemonScriptCandidates` holds the absolute paths, and the one that won is known at spawn time. Comparing the cmdline's script token against those absolute paths turns a heuristic into an identity check. No marker list can fix `daemon/index.js` being generic; only comparing against the path we actually launched can.

Related: #1019 (where this was raised and partially narrowed), #1001.

openwong2kim:
## [wmux-hermes] Triage Summary

[...]

Issue #537, opened 2026-07-22
[Bug]: After upgrade (3.27→3.30), session restore leaves panes blank and spawns duplicate agent sessions instead of reattaching


p-poppe:
### What happened?

After the app auto-updated from **3.27.0 → 3.30.0** and relaunched, restoring the previous session is broken:

- There is **no "restore session" prompt/button**. On launch the app **hangs** for a while, then the workspace layout and pane tabs reappear (with the previous task titles) — but the **terminals inside are blank**: a cursor is shown, nothing runs, nothing loads. The panes never reattach to their processes.
- **Strong evidence that restore re-launches instead of reattaching:** for each role workspace, a **brand-new agent session was created at launch time** — a fresh session containing only the startup init + the role prompt (~15 log lines) — sitting **alongside** the original, now-dead pane. So each workspace ends up with a blank dead pane **and** a freshly-spawned duplicate. In other words, restore appears to re-run the pane's startup command rather than re-attaching the existing pty.
- I had to manually recover each pane.

(No data was lost on my side because the agent persists its own conversation history to disk — but that's independent of wmux; from wmux's side the panes came back unusable.)

### Steps to reproduce

1. In 3.27.0, have several workspaces with running agent panes (Claude Code) each launched via the workspace's startup command.
2. Let the app auto-update to 3.30.0 and relaunch.
3. Observe: no restore prompt; a long hang; panes/tabs reappear with the old titles but **blank** (cursor only), not reattached.
4. Check the agent's session store: **new duplicate sessions were spawned at launch time** (startup command re-run) instead of the existing ones being reattached.

### Expected behavior

Restored panes should **reattach to their existing sessions** (or clearly offer a restore action) rather than coming back blank, and restore should **not spawn duplicate fresh sessions** by re-running the startup command.

### wmux version

3.30.0 (auto-updated from 3.27.0)

### How did you install wmux?

Setup.exe (direct download)

### Shell in the affected pane

PowerShell 7+ (pwsh)

### Windows version

Windows 11 Pro (build 26200)

### AI agent involved (if any)

Claude Code

### Logs, screenshots, or recordings

Daemon is alive after relaunch; `bridge.log` shows `SessionStart`/`Stop` hooks firing `ok`, so hooks work — the failure is in reattaching/restoring the pane processes. Screenshots of the blank restored panes (one workspace with two fully-blank panes; another with a blank tab + cursor) available on request.

### Confirmation

[...]

PR #1028, opened 2026-08-25
Revert "fix(daemon): identify the daemon by its script path, not a generic marker"

openwong2kim:
Reverts #1027; reopens #1025.

Post-merge panel review found the fix traded a narrow, old false positive for a worse failure mode, so it comes out while the real fix is built.

**The disqualifying finding.** #1027's safety argument was "a false negative is safe — refusing to kill degrades to the respawn budget." That holds in `killVerifiedDaemonPid`, but at the `ensureDaemon` call site a cmdline mismatch takes branch (b): *delete `daemon.lock`/`daemon.pid`/`daemon-pipe` and spawn fresh over the live process*. Under the old markers, a mismatch reliably meant "some other app on a recycled PID," so that cleanup was sound. #1027 narrowed the matcher so a mismatch can now also mean "our own daemon that we failed to prove" — e.g. any install path containing a space, since POSIX cmdline probes return argv joined by spaces and the token-equality identity check shatters the path — and then the cleanup path spawns a second daemon over a live one. That is the #537/#543 split-brain this module exists to prevent, reachable without PID reuse; the false positive #1027 fixed needs PID reuse plus a colliding layout.

Also found, for the record (all to be addressed in the redo): the `daemon-bundle` fallback still matches `daemon-bundler`/`daemon-bundle-backup` and runs unconditionally; identity compares *every* token, so another program merely holding our script path as an argument (`vim .../daemon-bundle/index.js`) verifies as the daemon; the Electron-seam wrappers call the candidate resolver outside the core's try/catch, breaking its "never throws" contract; and the new test injected candidates by hand, leaving the actual new wiring unexercised.

The redo needs: entry-script *position* comparison rather than any-token, argv preserved as an array on Linux (NUL-split `/proc/<pid>/cmdline`), the fallback as an exclusive branch with an exact-shape match, and — separately reviewed, since it is the split-brain-sensitive piece — the `ensureDaemon` mismatch branch refusing instead of cleaning + spawning.


openwong2kim:
This reverts #1027, restoring the segment-based cmdline matcher in daemonLauncherCore and dropping the scriptCandidates plumbing, tests, and changelog. The identity check could false-negative on POSIX paths containing spaces, driving ensureDaemon to spawn a second daemon over a live one. P2: it removes a split-brain regression reachable without PID reuse, at the cost of reopening the narrower #1025.

PR #1029, opened 2026-08-25
i18n(pl): fill in the remaining 1075 missing Polish strings

p-poppe:
## Summary

Polish had drifted to 374 of 1464 keys (26% coverage) — the exact same stalled point 20 of 22 locales sit at (only `ko` at 71% and `zh` at 98% are actively maintained). Settings alone was missing 219 strings; Deck, Channels, the first-run wizard, and most of the rest of the app fell back to raw English inside an otherwise-Polish UI.

This is deliberately **Polish only**, not a mass machine-translation of all 20 stalled locales — per your note on #997 that you won't machine-translate a language you can't verify the quality of. I can read/write Polish, so this one I can actually stand behind. The other 20 stay untouched.

## Changes

- **`src/renderer/i18n/locales/pl.ts`** — filled in all 1075 missing keys, mirroring `en.ts`'s structure and section comments 1:1 so future diffs stay easy to review. Followed the existing 374 keys' established terminology (`workspace` → `obszar roboczy`, `pane` → `panel`, bare status labels per #996/#1000 — did not reintroduce the `Agent `-prefixed form) and preserved every `{placeholder}` exactly (verified programmatically, see test plan).
- Dropped one orphaned key, `fanout.errPromptRequired` — it validated a prompt being required, but the fan-out prompt became optional upstream (`fanout.envOnlyHint` now covers that case) and the key has no `en.ts` counterpart anymore.
- **`src/renderer/i18n/__tests__/plLocaleCoverage.test.ts`** (new) — locks this from drifting silently again: every `en.ts` key has a `pl.ts` counterpart, no orphans, every `{placeholder}` set matches exactly, no key is an empty string. Scoped to `pl` only — the other 20 locales are a known, accepted gap, not something this test polices.
- **`changelog.d/1029.md`** — guessed the PR number from the latest merged PR; happy to rename if it lands differently.

## Test plan

- New test: `npx vitest run src/renderer/i18n/__tests__/plLocaleCoverage.test.ts` — 4/4 passed.
- Full i18n suite: `npx vitest run src/renderer/i18n src/renderer/components/Onboarding` — 140/140 passed, nothing broken.
- Placeholder consistency verified with a standalone script (loads both locale objects, diffs the `{name}` set per key) before writing the test — 0 mismatches across all 1464 keys.
- `npx tsc --noEmit` — clean.
- `npx eslint` on the changed files — clean.

## Summary by CodeRabbit

* **New Features**
  * Expanded Polish localization across workspace navigation, settings, accounts, terminals, notifications, approvals, sharing, onboarding, Git tools, themes, and remote workspaces.
  * Added Polish translations for status, warning, error, accessibility, confirmation, and progress messages.
* **Bug Fixes**
  * Removed an obsolete message and corrected dynamic placeholder usage.
* **Quality Improvements**
  * Added automated checks for complete translation coverage, matching placeholders, orphaned keys, and nonblank Polish text.
  * Completed 1,091 Polish translations, bringing Polish localization into full alignment with English.


openwong2kim:
This PR fills the 1075 missing Polish keys in `pl.ts` and adds a coverage test locking key/placeholder parity against `en.ts`. Renderer-only locale change, no logic touched — lowest-risk kind of diff, and the placeholder check guards the one subtle failure mode. P3: content-only i18n improvement, safe to merge on review.

p-poppe:
Both fixed.

**1. "chrome" mistranslated as the Chrome browser.** `settings.agentToolbar`/`agentToolbarShow` now read `Pasek narzędzi agenta`, matching the established `toolbar.*` term. Grepped the rest of `pl.ts` for `Chrome` — no other occurrences.

**2. Changelog's historical count didn't add up** (374 + 1075 − 1 ≠ 1464). The real number of net-new strings is **1091**, not 1075 — fixed the entry to state that and made the parity claim explicit (pl.ts now matches en.ts at 1464 keys exactly).

p-poppe:
All 9 fixed, plus one more of the same bug class the diff range missed.

- `workspace.agentComplete`: neuter → masculine (`Ukończony`)
- `workspace.agentNeedsAttention`: count now leads — checked the other `{count}`-first siblings in this file (`fleet.needsAttention`, `deck.fleetNeedsYou`, `strip.needsYou`, `deck.briefing.clause.blocked.one`) and this key was the only outlier
- `accounts.macosManualLogin`: keychain phrase case agreement
- `settings.baseOnPreset`: names the preset explicitly, matching adjacent `ustawienie wstępne` terminology
- `workspaceProfile.secretKeyWarning` / `warningPlaintext`: `jawnym tekstem` → `w jawnym tekście`
- `deck.tabCommander`: `Agent` → `Orkiestrator`, matching `deck.commander` and the feature's own naming comment
- `palette.cmd.newSurface`: `Nowa powierzchnia` → `Nowa karta`, matching `pane.addSurface` for the same action
- `fanout.agentLabel`: nominative → genitive (`polecenie agenta`)
- The 6 flagged ASCII-quote-closing-a-„-opening pairs fixed to `”` — and grepped the whole file for the same pattern, found a 7th instance outside this PR's diff range (a static help string) and fixed it too

Key parity re-verified: 1464/1464, 0 missing, 0 orphaned.

openwong2kim (dismissed):
This is the right shape — a speaker-verified single-locale pass instead of the mass machine translation #997 declined, the blast radius is exactly three files, and the coverage test is genuinely behavioral. The orphan-key removal checks out (zero references outside locale files, no `en.ts` counterpart, `fanout.envOnlyHint` is the successor). Suite and `tsc` verified green at `91f000f9`.

One thing before merge — and it's a source-sync matter, not a Polish one, so being the speaker doesn't save you here:

**`update.readyToInstall` (pl.ts:1352) translates the OLD English source.** `Instalacja zamknie uruchomione sesje` mirrors the pre-#897 sentence; current `en.ts` says "Installing closes every pane," reworded deliberately because on macOS the detached daemon means sessions *survive* the install (on Windows the installer deletes the install root, so there they don't — "every pane" is the phrasing that's true on both). `Instalacja zamyka każdy panel.` and this is done. It's also the one drift class the coverage test can't see, since the key exists and the placeholders match.

Everything else in the pass reads as your call to make — you're the speaker, and the wording is yours. Fix the one line and I'll merge.

openwong2kim:
Correction on item 1, before you spend time on my framing: "tells the user the opposite of what the feature does" was too strong, and I want the record straight.

What's actually true, verified in the updater code: on macOS the daemon is detached and sessions survive the install/relaunch (`AutoUpdater.ts`, the Squirrel.Mac comment); on Windows the installer deletes the install root the daemon runs out of, so before-quit takes the full `daemon.shutdown` branch and sessions genuinely do end (`index.ts`, the #866 hook). So "closes every pane" is the phrasing that's true on both platforms — which is presumably why #897 chose it — while your `zamknie uruchomione sesje` is right on Windows and wrong on macOS.

The ask itself stands, but for the ordinary reason, not the dramatic one: the Polish for `update.readyToInstall` should mirror its current English source (`Instalacja zamyka każdy panel.`), same as any other key.

Two things that fell out of checking this, neither yours to fix: `en.ts` itself is internally split — `settings.updateEndsSessions` still says "Installing closes your running sessions," which is the macOS-false phrasing on a key that renders next to the same widget — and that's an en-side cleanup I'll take separately.

openwong2kim:
Pushed the one line to your branch (`1ce9f90a`) rather than making you round-trip for it — the wording is the one from the review, so there was nothing for you to decide. `update.readyToInstall` now reads `Instalacja zamyka każdy panel.` Suite and `tsc` green locally.

While I was in there I went looking for the rest of that drift class, since "the key exists and the placeholders match" means neither your coverage test nor CI can see it. Method: blame `en.ts` and `pl.ts` per key on `main`, and flag any key whose English was last touched *after* the Polish was. 13 candidates on the pre-existing 374, 6 of which are real:

| key | en.ts today | pl.ts today |
|---|---|---|
| `settings.fontFamilyDesc` | Any installed font — type a name or pick one | Czcionka o stałej szerokości dla terminala |
| `fanout.title` | Multi Task — N parallel isolated tasks | Fan-out — 1 prompt → N izolowanych zadań |
| `fanout.promptLabel` | Shared prompt (optional — common context for every task) | Prompt |
| `fanout.promptPlaceholder` | Common context every task receives… (optional when each task has its own prompt) | Prompt, który otrzyma każde zadanie… |
| `fanout.titlesLabel` | Per-task title & prompt (editable — branch/slug identity) | Tytuły poszczególnych zadań (edytowalne — tożsamość branch/slug) |
| `update.readyToInstall` | (fixed in `1ce9f90a`) | |

The five open ones all predate this PR — they are in the original 374, not among the 1075 you filled — so they are outside what you signed up for, and the wording is yours to decide, not mine to guess. The fan-out four are the same event: the feature was renamed to Multi Task and its prompt became optional, and the Polish still describes the old dialog. `settings.fontFamilyDesc` is a straight leftover: it still says "monospace font for the terminal" while the setting now takes any installed font.

Want to take them in this PR, or would you rather they went in a follow-up? Either is fine by me — I am merging this one now either way, since none of the five are regressions this PR introduced and holding a 1075-key pass on them would be silly.

The other seven candidates were false positives (the English line moved or was reformatted, translation still accurate). `workspace.departed` → `poza granicą` I checked specifically, since it looked like drift: it is the boundary-warning badge, so "outside the boundary" is right.

One note for whoever picks up the follow-up: this class needs tooling, not a test. A coverage test can only see structure — missing keys, orphans, placeholders. Catching "the English changed and the translation didn't" needs the blame comparison above, which would fit as a CI advisory rather than a gate (it has a real false-positive rate, as the 7-of-13 here shows).

Thanks for doing this the narrow way. A speaker-verified single locale with a test behind it is worth more than twenty machine-translated ones, and the section-comment mirroring makes the next diff readable.

PR #1034, opened 2026-08-26
Let the inline preedit track the caret while an agent streams

### LINKED ISSUE #1032: [Bug]: Korean IME regression is back in v3.47.0 — composing syllable painted over the committed one again (different code path than #942/#945)
### Summary

Same visible symptom as #942 (`대한민국` reads `대한국` while composing) is back after upgrading v3.46.0 → v3.47.0. v3.46.0 was field-verified clean (see #942's closing comment). This is **not a regression of the original #945 fix** — the `.composition-view` split #945 made is still intact. It is a **new code path**: the `src=caret` content-anchor added by #951 / #1016 for Chinese IME recomputes `preedit` from the same correction as `pin`, which reintroduces exactly the coupling #945 removed.

### Environment

- wmux v3.47.0 (upgraded from v3.46.0)
- Windows 11 Enterprise 10.0.26200
- Korean Microsoft IME (no candidate window — every keystroke composes)

### Diagnostic evidence

`ime-anchor5` records from today's session, `%APPDATA%\wmux\logs\main-2026-08-26.log`:

| | count |
|---|---|
| total records | 80 |
| `src=instant` | 44 |
| `src=caret` | 36 |
| `src=caret` with `preedit ≠ (0,0)` | 27 |
| `src=caret` with `pin == preedit` exactly | **27** |
| `src=instant` with `preedit ≠ (0,0)` | 3 (stale carry-over on `composition-start`, not the live-drag pattern — see below) |

75% of `src=caret` events show `pin` and `preedit` at the *identical* value. Representative line:

```
composition-update cursor=(11,34) sel=(2,34) src=caret pin=(-72.0,0.0) preedit=(-72.0,0.0)
```

Compare with the working v3.46.0 log (attached to #942's closing comment), where negative `pin` values never appeared on `preedit` — the split held in every one of the 8 reproduction-geometry records observed there.

The 3 `src=instant` outliers are `composition-start` events with `pin=(0.0,0.0)` but nonzero `preedit` — a stale value carried from the composition's own prior `src=caret` correction, not a fresh miscalculation. Flagging for completeness, but the `src=caret` statistic above is the actual bug.

### Hypothesis

#951 / #1016 added a second anchor source (`src=caret`, content-scan of the agent's input line) specifically to fix Chinese candidate-window positioning while Claude Code streams output. That machinery appears to compute one correction and apply it to **both** the textarea pin and the visible `.composition-view` preedit — the same anchor-shared-between-two-surfaces mistake #945 fixed for the `src=instant`/pin path, reintroduced on the new path.

Korean draws no candidate window, so the textarea-anchor correction has nothing to serve for Korean — but the preedit is still getting dragged by it whenever `src=caret` fires, which per the field data above is most of the time while a Claude Code pane is actively streaming (i.e. most of the time this app is used for its intended purpose).

### Suggested fix direction

Whatever computes `src=caret`'s correction should keep the same separation #945 established: apply the correction to the textarea/candidate-window anchor only, and let `.composition-view` continue tracking the live cursor independent of which source (`instant` or `caret`) produced the pin.

### Offer

Same as #942 — I'm on Windows with Korean IME and can verify any candidate build against this exact reproduction (compose during active Claude Code streaming output). Downgrading to v3.46.0 in the meantime; happy to test a fix before it ships.

cc @openwong2kim (per #942/#945) — the `ime-anchor5` diagnostic tag should make this self-diagnosing per the design intent noted in #1016's fix.

openwong2kim:
Closes #1032

Korean inline composition is painted over text that already committed again on v3.47.0 — typing into a pane while an agent streams puts the syllable being composed near the *front* of the line, and the first syllable disappears from view until the space commits. The buffer is never wrong, only the paint, same as #942. But this is a different code path: #945's split is intact, and what broke it is the streaming anchor added for Chinese candidate windows.

## What actually happens

`src=caret` / `src=marker` exist because while an agent streams, its cursor is not reliably the input caret — so the anchor is taken from a quiet-caret snapshot or from the input line found by content. Both answer a **line-level** question: which row the input is on. #951 then pinned the inline preedit to that same cell, because following the live cursor tore the two IME surfaces apart (pinyin riding the agent's output rows while the candidate list sat at the input line).

The gap: typing into an agent's input reaches that streaming branch too. Every keystroke's echo is output, and a normal typing rhythm keeps the gaps under `OUTPUT_QUIET_MS` for longer than `STREAM_SUSTAIN_MS`. There the snapshot is one quiet-span stale, the live cursor **is** the caret, and pinning the preedit to the snapshot column drags the composing syllable backwards — further with every keystroke. The reporter's numbers show it exactly: `pin.x = (sel_col − cursor_col) × cellWidth` on every record, a correction sized to cancel the caret's real position.

## The fix

The pin reaches the preedit only when the live cursor is on a **different row** than the frozen cell — that cursor is the agent's repaint state, which is the tear #951 fixed. A cursor on the anchor's own row is the caret advancing along the input line, and the preedit follows it, which is the field-clean v3.46.0 behavior. Evaluated per composition event, so a chunk that parks the cursor off-row mid-composition pins for that event and the follow resumes when it returns.

The second commit extends the same gate to the row *range* of a wrapped input box, found by `scanClaudeInputLine`. This one is defensive rather than field-driven — see the note at the bottom.

The diagnostic tag moves to `ime-anchor6`, and logs `span=` when the box spans more than one row, so a shared field log is unambiguous about which build produced it.

## Evidence

86 unit tests pass, including the reporter's geometry (stale snapshot column, caret advancing on that row), a cross-row chunk mid-composition pinning for that event and resuming, a wrapped box, and a scrolling stream where ybase advances under a sticky input row.

Live, against a **real Claude Code pane** (v2.1.246) streaming a real turn, typing Korean into its input while it streams, and firing real composition events on xterm's helper textarea. Only the renderer module was swapped between runs.

The painted result, clipped from the input row mid-composition — `민` is the syllable being composed:

```
on main:      › 민확히어떻게해야하는지알려주
with this PR: › 정확히어떻게해야하는지알려주민
```

That is the reported symptom in pixels: the composing syllable lands on top of the first one, which vanishes.

The correction behind it, same run, caret advancing while the snapshot stays at column 2:

| caret | anchor | preedit correction on `main` | with this PR |
|---|---|---|---|
| `cursor=(8,42)` | `sel=(2,42)` | `-38.4` | `-0.06` |
| `cursor=(20,42)` | `sel=(2,42)` | `-115.2` | `-0.14` |
| `cursor=(38,42)` | `sel=(2,42)` | `-230.4` | `-0.27` |
| `cursor=(30,42)` (pixel capture) | `sel=(2,42)` | `-179.2` | `-0.2` |

The textarea pin is unchanged in both builds — the candidate window still sits where the anchor put it. Only the visible preedit stops being dragged.

A cross-row streaming cursor still pins both surfaces together, so #951 is unaffected.

## Note on the wrapped-box commit

While dogfooding I checked what current Claude Code actually draws, and it does not draw the `╭─ / │ > / ╰─` box `scanClaudeInputLine` matches — v2.1.246 renders a horizontal rule and a `›` prompt. So `src=marker` never fires against it: every record in this dogfood, and every record in the reporter's log (44 `instant`, 36 `caret`, 0 `marker`), comes from the cursor path. The rowSpan change is therefore correct but currently unexercised in the field, and the input box wrapping does not break the fix on this version anyway, because the caret keeps its screen row when the box grows (measured: column wrapped 50 → 4, row stayed 42).

The scanner being stale is its own defect and I have filed it separately rather than widening this PR.

@changinho25 — the reproduction above is yours (compose while output streams, without pausing long enough to end the composition). If you can run a build of this branch: `ime-anchor6` records should read `src=caret` with `preedit=(0.0,0.0)`; a nonzero `preedit` on a `src=caret` record means it is still wrong, and the `cursor=`/`sel=` fields on that same line will say why. Thank you for the correction comment — the first report's framing would have sent this at the wrong geometry.

### DIFF
--- a/changelog.d/1034.md
+++ b/changelog.d/1034.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **Korean inline composition no longer drags backwards while an agent
+  streams.** Typing a multi-syllable word into a pane with output flowing
+  painted the syllable being composed near where the word started, so
+  `정확히 어떻` read `떻확히 어` until the space committed it. The anchor added
+  for Chinese candidate windows pins both IME surfaces to the input line it
+  finds, and fluid Korean typing reaches that path too — commit echoes are
+  output — so the visible preedit was held at a column one pause stale while
+  the caret moved on. The candidate-window pin stays where it was; the inline
+  preedit now follows the caret whenever the cursor is on the anchor's own
+  row, including the continuation rows of a wrapped input box. (#1034)
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1213,8 +1213,8 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
         if (imeAnchorLogsLeft[budget] <= 0) return;
         imeAnchorLogsLeft[budget] -= 1;
         // Mirrored into the main-side log file by src/main/index.ts's
-        // console-message listener, so the user can share it. The "5" in the
-        // tag marks the input-line marker build (#1016) so a shared log is
+        // console-message listener, so the user can share it. The "6" in the
+        // tag marks the row-gated preedit-follow build (#1032) so a shared log is
         // unambiguous about which release produced it. src/held/restAge/sel
         // are the cause-3 discriminator: src=resting means the composition
         // started mid-repaint and the anchor used the last resting cell;
@@ -1225,7 +1225,7 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
         // textarea correction, preedit= the live composition-view
         // correction.
         console.info(
-          `[wmux:ime-anchor5] pty=${ptyIdRef.current} composition-${phase} ybase=${baseY} ydisp=${viewportY} ` +
+          `[wmux:ime-anchor6] pty=${ptyIdRef.current} composition-${phase} ybase=${baseY} ydisp=${viewportY} ` +
           `cursor=(${cursorX},${cursorY}) sel=(${selX},${selY}) src=${src}${edge ? ' edge=1' : ''} held=${held.toFixed(0)}ms ` +
           `restAge=${restAge.toFixed(0)}ms gap=${outputGap.toFixed(0)}ms caretAge=${caretAge.toFixed(0)}ms ` +
           `cellHeight=${cellHeight.toFixed(2)} ` +
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1208,7 +1208,7 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
         const id = ptyIdRef.current;
         return id ? useStore.getState().surfaceAgent[id]?.slug : undefined;
       },
-      onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, outputGap, caretAge, edge, selY, selX }) => {
+      onCompositionDiagnostic: ({ phase, baseY, viewportY, cursorY, cursorX, cellHeight, dx, dy, preeditDx, preeditDy, src, held, restAge, outputGap, caretAge, edge, rowSpan, selY, selX }) => {
         const budget = phase === 'start' ? 'start' : 'mid';
         if (imeAnchorLogsLeft[budget] <= 0) return;
         imeAnchorLogsLeft[budget] -= 1;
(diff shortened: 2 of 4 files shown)


openwong2kim:
@codex review

openwong2kim:

Row-gates the #1032 streaming pin so the inline preedit follows the caret on the anchor's own row and pins only cross-row. Touches `imeAnchor.ts`, the `ime-anchor6` diagnostic in `useTerminal.ts`, plus tests for the reported geometry, wrapped boxes, and scrolling streams. P2 — closes the v3.47.0 Korean IME regression; the #951 tear risk is test-covered.

Issue #1035, opened 2026-08-26
The IME anchor's Claude input-line scan looks for chrome Claude Code no longer draws, so src=marker never fires


openwong2kim:
## What

`scanClaudeInputLine` (added in #1016, `src/renderer/terminal/imeAnchor.ts`) looks for Claude Code's input box as a rounded box:

```ts
const CLAUDE_PROMPT_ROW = /^(\s*)│ [>!#] /;
const CLAUDE_BOX_TOP = /^\s*╭─/;
```

Claude Code v2.1.246 does not draw that. It renders a horizontal rule with a `›` prompt under it — no side borders, no rounded corners, no `│ > `. So the scan never matches, `src=marker` never wins, and every composition falls back to the cursor path the marker exists to override.

## Evidence

Live, in a dev pane running a real Claude Code v2.1.246 turn, with the pane recognized as an agent (sidebar shows `Agents 1 · Claude Code`, so `getAgentSlug()` returns `'claude'` and the scan does run):

- 25+ `ime-anchor` records across four dogfood runs while output streamed: every one `src=caret` or `src=instant`, never `src=marker`.
- The field log in #1032 says the same thing independently: 44 `src=instant`, 36 `src=caret`, **0 `src=marker`**.

Clipped from the pane, the input area as it actually renders:

```
────────────────────────────────
› 정확히어떻게해야하는지알려주
⚠ Transcript saving is off …
```

versus what the scan is looking for:

```
╭──────────────╮
│ > …          │
╰──────────────╯
```

## Why it matters

The whole point of #1016 was that no cursor-derived rule can find the caret while an agent streams, because the TUI does not park its cursor there — the content scan was the answer. On the current Claude Code that answer is unreachable, so the anchor is back to the quiet-caret snapshot with its documented residuals (a snapshot that is itself a park, a stale snapshot after a mid-stream pause).

It fails safe: the scan returns null and selection degrades to the pre-#1016 behavior, which is why nothing looked broken. But the machinery is inert, and a reader of the code would reasonably assume it is doing work.

Worth noting from the same dogfood: v2.1.246 keeps its cursor **on the caret** while streaming (`cursor=(2,42)` = the `›` prompt column, unmoved for six seconds of output). If that holds generally, the premise the streaming anchor was built on has partly expired for this agent, and the honest fix might be to narrow the machinery rather than teach the scanner new chrome.

## Options

[...]

PR #1038, opened 2026-08-26
chore(release): 3.47.1

openwong2kim:
Folds the two pending fragments into `CHANGELOG.md`, bumps to 3.47.1, and regenerates the API reference (its header bakes the version, which the CI drift guard enforces).

PATCH because both fragments are `Fixed`: the full Polish locale pass (#1029) and the Korean inline-preedit anchor (#1034). No API surface moved, so the reference diff is the version line alone.

Contents: #1029, #1034.

One edit before folding: `1034.md` credited the fix with covering the continuation rows of a wrapped input box. The code does that, but the content scan that reaches that path looks for chrome current Claude Code no longer draws (#1035), so no user can observe it today — the clause came out rather than ship a release note nobody can verify.

`## [Unreleased]` had to be re-added before running the collector, since each release consumes the heading. Same papercut as last time, already in TODOS.md.

Verified: `tsc --noEmit` clean, generated reference reads `Generated from wmux v3.47.1 sources`, `package-lock.json` mirrored to 3.47.1 in both places alongside `package.json`.

After merge the `v3.47.1` tag goes on the release commit — installer builds hang off the tag.

One thing to weigh before tagging: #1032's reporter has not field-verified #1034 yet. The fix is proven against a real Claude Code pane in a dev build (real streaming turn, real typing, the painted preedit compared against `main` — see the PR), but a real Korean IME driving real composition events is the one gap a dev-side dogfood cannot close, and this issue family has burned a field verification before (#875). Tagging ships it to everyone; holding the tag keeps #1029's Polish pass waiting too.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected Korean inline text composition positioning during streaming output.

* **Localization**
  * Completed Polish locale coverage and improved translation parity.

* **Documentation**
  * Updated the API reference and package version to 3.47.1.


openwong2kim:
Release chore: folds the #1029 (Polish locale) and #1034 (Korean inline preedit) fragments into CHANGELOG.md and bumps to 3.47.1 across package.json, the lockfile, and the regenerated API reference. No runtime code changes, so risk is limited to release mechanics. P3 — routine patch-release bookkeeping; worth deciding on holding the tag until #1034 is field-verified.

PR #1040, opened 2026-08-26
fix(terminal): don't anchor the preedit to the previous composition at start

openwong2kim:
Field follow-up on #1032. The reporter verified v3.47.1 fixes the drag, and flagged the one record that did not match the healthy signature:

```
composition-start … src=caret pin=(-72.0,0.0) preedit=(24.0,0.0)
```

with the same `24.0` recurring on unrelated `src=instant` composition-start records. Their read was right — it is not the old `pin == preedit` coupling — and it is worth chasing, because the cause is real even though the symptom is not visible.

## Cause

xterm positions the preedit only in `updateCompositionElements`, called from `compositionupdate`. Its `compositionstart` handler does this and nothing else:

```ts
public compositionstart(): void {
  this._isComposing = true;
  this._compositionPosition.start = this._textarea.value.length;
  this._compositionView.textContent = '';
  this._dataAlreadySent = '';
  this._compositionView.classList.add('active');
}
```

So at `compositionstart` the element still carries the **previous** composition's `style.left/top`, and `syncPreedit` computed `desired - actual` against those. A stale anchor gives a stale delta, and since the leftover position is the same each time, so is the number — which is exactly the recurrence the reporter noticed.

Nothing is painted there: an empty `.composition-view` has no padding or `min-width` in xterm's stylesheet, so it is zero-width until the first update fills it. That is why this never reached the screen in either the dogfood or the field.

## Why fix a cosmetically inert record

The diagnostic exists to be read by reporters — it is the whole reason `#874`/`#942`/`#951`/`#1032` were diagnosable from a shared log rather than a repro none of us can run. A record that reports an offset nobody can see is indistinguishable in a log from a real drag, and it already cost a reporter time on a build that was working correctly.

## The change

The preedit correction begins at the first update, where the position is real. The streaming pin is unaffected — the first update is also where xterm gives the preedit a position for the pin to correct — and a test now pins that specifically.

One existing test moved with it. #942's scrolled-viewport case asserted the preedit correction immediately after `compositionstart`, against coordinates the test had placed on the element itself, which is a state xterm never produces at start. It now asserts after the update, so it still guards what it was written to guard (the ydisp term reaches the preedit) without encoding a timing that does not exist. Calling that out explicitly rather than letting an existing green assertion quietly change meaning.

## Verified

`npx vitest run src/renderer/terminal/__tests__/imeAnchor.test.ts` — 88/88, including two new cases: a second composition starting with a zero preedit correction instead of one measured off the first, and a caret-sourced pin still landing on the first update.

`tsc --noEmit` and `eslint` clean.

No changelog fragment: the user-visible behaviour does not change, only what the diagnostic reports.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved text input behavior for IME compositions by deferring preedit positioning until the first update.
  - Prevented stale positioning from previous compositions from affecting new input sessions.
  - Preserved accurate caret alignment when composition text spans multiple rows.

### DIFF
--- a/src/renderer/terminal/__tests__/imeAnchor.test.ts
+++ b/src/renderer/terminal/__tests__/imeAnchor.test.ts
@@ -519,6 +519,13 @@ describe('#942 inline preedit (Korean IME) — the pin never drags the preedit',
     dom.compView.style.left = '120px';
     dom.textarea.dispatchEvent(new Event('compositionstart'));
     expect(translateOf(dom.textarea)?.dy).toBeCloseTo(8 * 17.6, 6);
+    // The preedit's correction lands on the first compositionupdate, not at
+    // start: xterm's compositionstart writes only textContent and the `active`
+    // class, so at start the element holds no position for this composition
+    // (nothing is painted there either — an empty view is zero-width). This
+    // used to assert at start against coordinates the test had placed itself,
+    // a state xterm never produces (#1032 field follow-up).
+    dom.textarea.dispatchEvent(new Event('compositionupdate'));
     expect(translateOf(dom.compView)?.dy).toBeCloseTo(8 * 17.6, 6);
     handle.dispose();
   });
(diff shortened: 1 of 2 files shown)


openwong2kim:
This defers the preedit correction in `imeAnchor.ts` to the first `compositionupdate`, because xterm still carries the previous composition's coordinates at `compositionstart` — the phantom `preedit=(24.0,0.0)` from the #1032 field log. Screen behavior is unchanged: the source change only clears a stale correction at start, and the #942 test assertion moves to the update. P3: narrow, well-tested, no user-facing change.

PR #1041, opened 2026-08-26
Close the two review-round findings on the IME sync timer and the boot-marker poll

openwong2kim:
Post-merge 3-way review round (Claude / GLM-5.3 / Grok) on #1039 and #1040 — both had merged without independent review. No P1s; two findings survived verification and this PR closes both.

## 1. The deferred sync outlives a composition that never ends (`imeAnchor.ts`)

`compositionupdate` arms a `setTimeout(0)` re-sync, mirroring xterm's own recursive reposition. Only compositionEND disarmed it. A composition that dies without its `compositionend` event — focus loss mid-composition — leaves the timer armed, and it fires during the **next** composition's pre-update window, correcting against coordinates xterm wrote for the previous composition. That is precisely the stale-anchor class #1040 fixed at `compositionstart`, surviving through a side door.

The reviewer who found it proposed a latch; the disarm-at-start is simpler and closes the same hole, because `syncPreedit` has no callers outside the composition handlers (`onRender`/`onScroll` touch only the textarea path) — so the leftover timer was the *only* thing that could fire in that window. It is also symmetric with the end handler, which already disarms for the mirror-image reason.

The regression test is proven non-vacuous: with the disarm removed it fails (`translate(60px, 0px)` where `''` is expected), with it it passes.

Narrow trigger, invisible result (the transform lands on a zero-width empty view and self-heals on the first update) — but the module's whole diagnostic story depends on the log not lying, and this was a path where it could.

## 2. The boot-marker poll accepted torn reads (`bootMarker.runtime.test.ts`)

All three reviewers converged on this one independently, each from the same observation: #1039 skipped only **empty** reads, so a torn read (`"90"` of `"9052"`) still escaped the poll and reached the PID assertion as the same NaN-class flake — while the comment justified the check by claiming equivalence with `parseBootMarker()`, which rejects *anything* unparseable. The claimed equivalence was not the implemented one.

The poll now uses the production parser itself as its predicate, against the spawned child's PID: empty and torn reads alike re-poll as "not yet", and the comment's claim is now literally true. A daemon that never writes a valid marker still fails `expect(sawMarker).not.toBeNull()`, so the witness keeps its teeth.

## Review-round claims that did *not* survive verification, for the record

- "Re-syncs between start and first update reapply stale corrections via cursor/write/render/scroll" — those paths call `sync()` (textarea only), never `syncPreedit`. The defect above is real but its mechanism is the leftover timer, not those callers.
- "The test can pass without ever observing a marker" — `expect(sawMarker).not.toBeNull()` fails that case.

## Verified

- `imeAnchor.test.ts` 89/89 (new straddle test included), fails-without-fix proven.
- `WMUX_REQUIRE_DAEMON_BUNDLE=1` boot-marker runtime suite 3/3 against the real bundle.
- `tsc --noEmit`, `eslint` clean.

No changelog fragments: a test-only change and a not-user-observable timer disarm, same standing as #1039/#1040.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input-method editor behavior when a composition loses focus unexpectedly, preventing stale positioning from affecting subsequent text input.
  * Improved daemon startup marker handling to reject incomplete or invalid reads and confirm the marker matches the active daemon process.

* **Tests**
  * Added regression coverage for interrupted text compositions and daemon boot-marker validation.

### DIFF
--- a/src/daemon/__tests__/bootMarker.runtime.test.ts
+++ b/src/daemon/__tests__/bootMarker.runtime.test.ts
@@ -19,6 +19,7 @@ import fs from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
 import net from 'node:net';
+import { parseBootMarker } from '../../shared/daemon/daemonLauncherCore';
 
 const SUFFIX = '-i546-test';
 const dir = path.join(os.homedir(), `.wmux${SUFFIX}`);
@@ -61,6 +62,8 @@ describe.skipIf(!hasBundle)('#546 boot marker lifecycle (real daemon)', () => {
       env: { ...process.env, WMUX_DATA_SUFFIX: SUFFIX },
       stdio: ['ignore', 'ignore', 'ignore'],
     });
+    const spawnedPid = child.pid;
+    expect(spawnedPid, 'spawn returned no pid').toBeDefined();
 
     // (1) The marker appears, and names the daemon's own PID. Poll densely —
     // boot with zero sessions to recover is ~300 ms, and the marker is written
@@ -74,12 +77,15 @@ describe.skipIf(!hasBundle)('#546 boot marker lifecycle (real daemon)', () => {
       try {
         const raw = fs.readFileSync(markerFile, 'utf-8');
         // `writeFileSync` creates the file and then writes it, so a 2 ms poll
-        // can win that race and read the empty intermediate — which parsed to
-        // NaN and reddened `validate` on PRs that never touch the daemon.
-        // An empty read is not "the marker appeared"; it is the same "not yet"
-        // as ENOENT, and production agrees: parseBootMarker() rejects anything
-        // unparseable rather than treating it as a boot claim.
-        if (raw.trim() !== '') {
+        // can win that race and read a not-yet-complete intermediate — which
+        // parsed to NaN and reddened `validate` on PRs that never touch the
+        // daemon. The production parser is the predicate here, not an
+        // approximation of it: an empty read AND a torn one ("90" of "9052")
+        // are equally unparseable boot claims, and both re-poll as "not yet",
+        // exactly the way parseBootMarker() refuses them in the launcher.
+        // (3-way review consensus on the first cut, which skipped only empty
+        // reads and let a torn read through to the PID assertion below.)
+        if (spawnedPid !== undefined && parseBootMarker(raw, spawnedPid)) {
           sawMarker = raw;
           break;
         }
--- a/src/renderer/terminal/__tests__/imeAnchor.test.ts
+++ b/src/renderer/terminal/__tests__/imeAnchor.test.ts
@@ -1770,3 +1770,53 @@ describe('#1032 field follow-up — compositionstart never corrects against the 
     handle.dispose();
   });
 });
+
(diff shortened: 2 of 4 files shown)


openwong2kim:
Follow-up to the #1039/#1040 review round fixing two verified findings: the leftover `setTimeout(0)` sync in `imeAnchor.ts` is now disarmed at `compositionstart`, and the boot-marker runtime test polls with the production `parseBootMarker` so torn reads can't reach the PID assertion. Mostly test/hardening — the timer fix targets a narrow, self-healing focus-loss path. P3: no user-visible behavior change, regression tests included.

openwong2kim:
Post-merge live verification of both halves of this PR, with one correction to put on the record.

## The boot-marker race is real — observed, not inferred

Eight launches of the real bundled daemon, polling the marker as fast as the loop allows: 9,993 reads, and round 1 caught the intermediate — `distinct=["","5300"]`, one empty read between the file's creation and its write. The pre-#1039 poll would have taken that `""` as the marker and produced exactly the CI `NaN`. Torn partial-PID reads: zero in 9,993, consistent with a single small write being effectively atomic — the `parseBootMarker` predicate stands as defense for the unobserved case, not as the fix for an observed one.

## The straddle window is theoretical-only — and my stated justification for it was wrong

The commit message and PR text justify the `compositionstart` disarm with *"a composition that dies without its compositionend event (focus loss mid-composition) leaves it armed."* Measured against a real composition in the running app: **focus loss fires `compositionend` before `blur`** (`compositionupdate(정확) → compositionend(정확) → blur`). That justification is false and I withdraw it.

Full instrumentation of the timer (every arm/fire tagged in-page, driven by real CDP compositions through commit-and-recompose cycles, cancel, and commit paths):

- 29 arms across two sessions — **zero fired after a subsequent `compositionstart`**. Every timer either fired within 6–16 ms or was cleared by `compositionend`/the next update's re-arm.
- Every path that ends a composition produced `compositionend` (syllable commit, `insertText` commit, focus loss). The one path that did not — a synthetic CDP composition left uncancelled by Escape — also produced **no next `compositionstart`**: the composition just continues, so the straddle's precondition (a new start with no prior end) fails twice over.
- Input-priority check: a key event dispatched immediately after arming still arrived ~6 ms **after** the timer fired — no evidence that input tasks can preempt an armed 0 ms timer here.
- The margin that matters: end→next-start gaps run as low as 4 ms against a 6–16 ms timer lifetime — so the **end-handler disarm is the load-bearing guard**, and it held in every observed sequence. The start-side disarm added here is symmetry behind it.

The regression test proves the code (remove the disarm and it fails) but reaches the window only via fake timers holding the callback across dispatches — a state no observed real sequence produced.

The disarm stays: it is correct, free, and symmetric, and a UA or future refactor that ever violates the end-before-start ordering hits it instead of the stale-anchor bug. But it is hardening against an unreached state, not a fix for a reachable one, and the record should say so rather than let the commit message's claim stand.

(Also corrected during this round: `Input.imeSetComposition` works fine against this build — the renderer-wedge note from the #874 era did not reproduce — which is what made driving real compositions possible at all.)

Issue #1043, opened 2026-08-26
Update install has no feedback during the wait, and a repeat click while a waiter is live is silently a no-op


p-poppe:
### Problem

The install-teardown machinery from #866/#980 (`installTeardown.ts`, the detached PowerShell waiter) is working correctly, but the user-facing experience around it has no feedback, which reads as "the update button did nothing" even when the underlying mechanism is behaving exactly as designed.

Reproduced from a real session, timestamps from `main-2026-08-26.log`:

1. **08:44:32** — clicked "update". `[AutoUpdater] install handoff: waiter=...IUmr9G...` fires, app quits. The detached waiter (`wait-and-install.ps1`) starts its up-to-~120s window (60s handle-wait + up to 60s lock-probe) with **zero UI**: no window, no tray icon, no progress of any kind.
2. During that window the closed app looks identical to "nothing is happening" from the outside — because, from the outside, nothing visibly is.
3. **08:45:50** (~78s later) — the user reopened wmux manually, reasonably assuming the update had failed or wasn't going to happen. This re-locks the install root's binaries (a live `wmux.exe` is now running out of the same tree the waiter is watching), directly working against what the waiter is polling for in `Test-RootLocked`.
4. **08:46:18** — clicked "update" again. Log shows the *second* handoff had to `force-killed 1/1 foreign` — something the first attempt's cleanup left behind. This second handoff is where the report's "clicked again, nothing happened" lines up: `buildWaiterScript`'s single-instance gate (`installTeardown.ts`, the named mutex) is `if (-not $mtxCreated) { exit 5 }` — **silent, by design**, no marker file, nothing for the app to report on next boot, if a previous waiter for the same root is still alive. From the user's seat this click and the first click are indistinguishable: both just make the window disappear.
5. Gave up, manually downloaded and ran `Setup.exe` — which of course works, since it bypasses this whole handoff.

None of this is data corruption or a wrong outcome — the mechanism's whole job is to refuse rather than risk destroying the install, and it did that correctly both times. The gap is entirely in feedback:

- **No indication *during* the wait** that anything is happening at all (~1-2 minutes, easily long enough for a reasonable person to conclude it's stuck and intervene).
[...]

PR #1044, opened 2026-08-26
fix(updater): give the install wait some feedback

p-poppe:
## Summary

Addresses one of the three parts of #1043 — the other two turned out to already be handled (see below). The install-teardown mechanism from #866/#980 (`installTeardown.ts`, the detached PowerShell waiter that confirms every process has actually released the install root before letting Squirrel run) is correct — it fails safe by refusing rather than risking a half-deleted install. The gap #1043 reported was entirely in feedback around it.

## Changes

**A best-effort "installing the update, please wait" window for the wait itself.**

Between quit and `Setup.exe` actually launching, the waiter spends up to ~1-2 minutes (60s handle-wait + up to 60s lock-probe, both generous on purpose — see the comments on `INSTALL_LOCK_BUDGET_MS`) with zero UI. From the outside that's indistinguishable from a failed update, which is what triggered the report behind #1043: a manual reopen of the app mid-wait re-locked the install root, and a subsequent click hit the single-instance mutex refusal (see "considered and reverted" below).

The window can't live in the Electron app's own process — that process is precisely what has to fully release the install root, so anything meant to render for the whole wait has to run outside that tree, same reasoning as why the waiter script itself already lives outside the install root (see the comment on `spawnInstallWaiter`). So it's built into the waiter: a minimal always-on-top WinForms window, shown non-modally right after the mutex check, kept responsive via `Application.DoEvents()` woven into the existing wait loops (the handle-wait loop is now a short poll instead of one long blocking `WaitForExit`, same deadline/contract, just finer-grained so the form doesn't freeze for the whole slice), and closed right before `Start-Process` (Squirrel has its own UI from there).

**Safety property:** every use of `$form` is null-checked and wrapped in its own `try`/`catch`. A machine where `Add-Type -AssemblyName System.Windows.Forms` fails (locked-down policy, no display subsystem) degrades to exactly today's silent-but-correct wait — the indicator can never gate, delay, or fail the install itself. Verified the generated script parses with `[System.Management.Automation.Language.Parser]::ParseFile` (not just eyeballed).

Because this window is owned by the still-running *incumbent* waiter and stays on screen (topmost) for its whole lifetime, it also covers the reopen case from the report: a user who closes-and-reopens wmux while a waiter is still alive sees it still sitting there, rather than a closed window and no explanation.

## Considered and reverted: writing a marker on the mutex refusal

First pass at this PR made `buildWaiterScript`'s single-instance guard (`if (-not $mtxCreated) { exit 5 }`) write an abort marker before exiting, so a repeat click while a waiter was already live would have *something* to report on the next boot, via the existing pull-based notice (`AppLayout.useRefusedInstallNotice`, #866).

Reverted after checking it against what the *incumbent* waiter can still do: it has up to another ~120s to decide the real outcome. If it goes on to **succeed**, it writes no marker at all — the newcomer's "another install is already in progress" marker would then sit there unclaimed and get read on the next boot as "your update was refused," which would be false. If the incumbent later fails, it overwrites whatever the newcomer wrote anyway, so the newcomer's write only ever mattered in the one case where it was wrong. This exact tradeoff is why the mutex branch was silent by design in the first place (see the pre-existing comment above it) — the wait window above is the right fix for the visibility gap instead, since it doesn't require guessing at an outcome that isn't decided yet.

## Part 2 of #1043's ask — already fixed, not touched here

`AutoUpdater` already tracks `isInstalling` in-process and answers a repeat "install now" with an immediate `UPDATE_ERROR` ("An update install is already in progress...") instead of re-running the quit/waiter cycle — this already covers a fast double-click within the same running session. It does *not* cover the report's actual scenario (the app was fully closed and manually reopened between clicks, which drops all in-memory state) — that gap is what the wait window above closes instead, from the waiter side rather than the app side.

## Part 3 of #1043 (proactive refusal notice) — already fixed, not touched here

The issue's diagnosis leaned on `AutoUpdater.takeRefusedInstall()`'s doc comment, which says the only listener lives in the Settings panel (mounted only while Settings is open). That's stale documentation: `AppLayout.useRefusedInstallNotice` already pulls and surfaces this on every boot, unconditionally, as an error toast — the same fix already applied to `getPendingInstall()` for #897. Flagging this rather than silently no-op'ing it: didn't want to claim credit for something that was already working, and didn't want to touch the doc comment's historical framing without being sure that's welcome scope for this PR.

## Test plan

- `installTeardown.test.ts`: 2 new tests for the wait indicator (code shape — build wrapped in try/catch, closed before `Start-Process` on every exit path including both abort branches, never gates the wait/probe loop). All pre-existing tests in the same file, including the two that pin the mutex branch's silence (`yields silently (exit 5) — the incumbent owns the install AND the marker`) and the `WaitForExit` int-cast guarantee (updated to match the new per-slice call), pass unchanged in intent.
- Full suite run for real this time: `npm install` + `npx vitest run src/main/updater` — **137 passed**, `installTeardown.test.ts` alone **52/52**. `npx tsc --noEmit` clean. `npx eslint` on both touched files — 0 errors, 4 pre-existing warnings (`no-non-null-assertion`, same pattern the rest of the file already uses).
- Generated the actual script with the real function and validated it with PowerShell's own parser (`[System.Management.Automation.Language.Parser]::ParseFile`) — 0 syntax errors.

## Not in this PR

The budget sizes themselves (60s + 60s) aren't touched — they're already documented as deliberately generous (a daemon flushing large scrollback buffers can hold files a while, and waiting is free while refusing costs the update), and #1043 didn't ask to shrink them, just to make the wait visible.

## Summary by CodeRabbit

* **New Features**
  * Added an always-on-top installation progress window during Windows handoff.
  * The window remains responsive while installation processes and file locks are checked.
  * Updated the progress window with branded dark styling and improved readability.

* **Bug Fixes**
  * Ensured the window closes after successful, aborted, timed-out, or interrupted installation attempts.
  * Improved graceful fallback when the window cannot be displayed.
  * Prevented distinct installation locations from being incorrectly treated as the same concurrent installation.

### DIFF
--- a/changelog.d/1044.md
+++ b/changelog.d/1044.md
@@ -1,7 +1,3 @@
-### Fixed
-
-- **A repeat "Install now" click while an update was already installing used to do nothing at all.** The single-instance guard in the install waiter (#980) is correct — a second Squirrel install racing the first is exactly the corruption it exists to prevent — but it refused silently: no marker, so nothing to report on the next boot. It now writes a marker before refusing, so the existing "why was my update refused" notice (#866) picks it up like any other refusal instead of leaving the click looking like the button did nothing. (#1043)
-
 ### Added
 
 - **A best-effort "installing the update, please wait" window during the install handoff.** Clicking "install now" used to close the app and go silent for up to ~1-2 minutes while the install waiter confirmed every process had actually let go of the install root — indistinguishable, from the outside, from the update having failed. The waiter (which already has to run outside the install root it is watching) now shows a small always-on-top window for that wait, and closes it right before Squirrel's own installer UI takes over. Failing to show it (a locked-down machine, no display subsystem) degrades to today's silent-but-correct behavior — it can never block or fail the install itself. (#1043)
--- a/changelog.d/1044.md
+++ b/changelog.d/1044.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- **A repeat "Install now" click while an update was already installing used to do nothing at all.** The single-instance guard in the install waiter (#980) is correct — a second Squirrel install racing the first is exactly the corruption it exists to prevent — but it refused silently: no marker, so nothing to report on the next boot. It now writes a marker before refusing, so the existing "why was my update refused" notice (#866) picks it up like any other refusal instead of leaving the click looking like the button did nothing. (#1043)
+
+### Added
+
+- **A best-effort "installing the update, please wait" window during the install handoff.** Clicking "install now" used to close the app and go silent for up to ~1-2 minutes while the install waiter confirmed every process had actually let go of the install root — indistinguishable, from the outside, from the update having failed. The waiter (which already has to run outside the install root it is watching) now shows a small always-on-top window for that wait, and closes it right before Squirrel's own installer UI takes over. Failing to show it (a locked-down machine, no display subsystem) degrades to today's silent-but-correct behavior — it can never block or fail the install itself. (#1043)
--- a/src/main/updater/__tests__/installTeardown.runtime.test.ts
+++ b/src/main/updater/__tests__/installTeardown.runtime.test.ts
@@ -17,6 +17,7 @@ import * as fs from 'node:fs';
 import * as os from 'node:os';
 import * as path from 'node:path';
 import { spawn, spawnSync, execFileSync, type ChildProcess } from 'node:child_process';
+import { createHash } from 'node:crypto';
 import {
   buildWaiterScript,
   spawnInstallWaiter,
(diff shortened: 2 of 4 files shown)


openwong2kim:
#1044 gives the install wait feedback (#1043): the waiter's mutex refusal now writes a marker, and a best-effort always-on-top wait window runs inside the detached waiter. It touches installTeardown.ts, the install-safety machinery, and the UI half is untestable on CI — worth one manual run on Windows. P3: feedback-only on a P3 issue, fail-safe by construction.

p-poppe:
Both fixed.

**1. Mutex name collision.** `$mtxName` now derives from `SHA256($root)` instead of `$root -replace '[^A-Za-z0-9]', '_'` — added a test computing the same digest in Node's `crypto` over the exact adversarial pair (`C:\wmux-a` vs `C:\wmux_a`) that the old lossy replace collided on, confirming SHA256 does not.

**2. Form leaked visible on a mid-setup throw.** The `catch` block now closes $form before nulling it (if it got far enough to exist) instead of just dropping the reference.

Also restyled the wait window — dark background/text matching the app's own default chrome (`bgMantle`/`textMain`/`accentSecondary` from `themes.ts`'s `amber` theme) instead of a plain default WinForms gray box, after actually seeing it live: running `npx vitest run src/main/updater` directly (rather than `npm run test:parallel`, which correctly excludes `*.runtime.test.ts`) picked up `installTeardown.runtime.test.ts` and popped the real window on a real desktop mid-verification — safe (fake setup.cmd, self-contained sandbox, confirmed no leftover processes/files after), but a good look at the default styling in the wild.

129/129 (`test:parallel`, the runtime suite correctly excluded this time), `tsc` clean, ESLint clean, generated script re-verified with PowerShell's own parser.

PR #1047, opened 2026-08-26
feat(i18n): detect system locale on first run, fall back to English

p-poppe:
## Summary

Feature request from Simply.IN HQ (not a filed bug): a fresh install/session should default to the OS's own language when it's one of the 23 we ship, instead of always opening in English until the user finds Settings.

## What's there today

`workspaceSlice.loadSession` only ever sets `state.locale` when `data.locale` is present in `session.json` — and the only writer of that field is `setLocale` (Settings panel), so a user's explicit choice was the only thing that ever set it. Nothing read `app.getLocale()`/`navigator.language` anywhere in the codebase before this PR — confirmed by grep, not assumed.

## Changes

- **`src/renderer/i18n/index.ts`** — `detectSupportedLocale(systemLocale, supported = SUPPORTED_LOCALES)`: pure, total, never throws. Exact match first (catches `zh-TW`/`pt-BR`, the two ids that carry a region themselves), then primary-subtag match ignoring region (`pl-PL` → `pl`, `de-AT` → `de`). `zh` gets dedicated handling: a Traditional/Taiwan/HK/Macau signal (`Hant`, `-TW`, `-HK`, `-MO`) routes to `zh-TW`, everything else starting `zh` routes to `zh` — those two scripts aren't mutually readable, so a wrong guess is worse than falling back to English would be for any other language. No match at all → `'en'`.
- **`src/preload/preload.ts`** — `systemLocale: navigator.language` added as a static field on `electronAPI`, same pattern as the existing `platform`/`windowsBuildNumber` fields right above it (sandboxed preload keeps the full Web platform surface even though most Node builtins are restricted — see that block's own comment). No IPC round-trip to `app.getLocale()` needed; Electron already seeds `navigator.language` from the OS locale, and it can't change mid-session so there's nothing to keep in sync.
- **`src/renderer/stores/slices/workspaceSlice.ts`** — `loadSession`'s existing `if (data.locale) { ... }` gained an `else` branch: only reached when `data.locale` was never written (a genuinely first-ever run, or a `session.json` predating this field), calls `detectSupportedLocale(window.electronAPI?.systemLocale ?? '')` and applies the result. An explicit choice already on record — the user's own pick, or a previous run's detection result, both get written back to `session.json` by `setLocale` — always wins; this branch cannot re-fire for a session that's already been through it once.

## Test plan

- New `src/renderer/i18n/__tests__/detectSupportedLocale.test.ts` — exact/region/case-insensitive matching, the zh-TW vs zh script routing, English fallback for both an unsupported language and garbage/empty input, and that an injected `supported` list is respected (used to keep the fixture stable if we ever add/drop a locale).
- New tests in `workspaceSlice.loadSession.test.ts`: detects a supported OS locale on a session with no `locale` field, falls back to English for an unsupported OS locale, does **not** override an explicit `data.locale` even when the OS locale would map to something else, and is inert (defaults to English, never throws) if `window.electronAPI` isn't available yet.
- `npm run test:parallel -- src/renderer/i18n src/renderer/stores/slices/__tests__/workspaceSlice.loadSession.test.ts` — **217/217 passed**. (Used `test:parallel`, not a bare `vitest run` — this repo also has real-execution `*.runtime.test.ts` files elsewhere under `src/main/updater` that a bare `vitest run` would pick up and that visibly launch PowerShell/WinForms windows on a real Windows box; irrelevant to this PR's files, but worth the right command out of habit.)
- `npx tsc --noEmit` — clean.
- `npx eslint` on all five touched files — 1 pre-existing error + 22 pre-existing warnings, all confirmed present in the identical, untouched lines of `upstream/main` before this PR (checked via `git show upstream/main:<file> | eslint --stdin`) — none introduced by this change.

## Not in this PR

Region-to-region nuance beyond the `zh`/`zh-TW` split (e.g. `pt-PT` maps to `pt-BR` via the generic primary-subtag pass, since `pt-BR` is the only Portuguese we ship) — flagging it here rather than silently deciding it's fine: happy to special-case it out to English instead if that's preferred, but `pt-BR` read as closer to a Portugal user than English felt like the more useful default.

`changelog.d/1047.md` — guessed the PR number from the latest open PR at filing time; happy to rename if it lands differently.

## Summary by CodeRabbit

- **New Features**
  - The app now detects your system language during first launch and selects the closest supported language automatically.
  - Regional and Chinese script variants are recognized, including Traditional Chinese locales.
  - Unsupported, invalid, or unavailable language settings fall back to English.
  - Previously saved or explicitly selected language preferences remain unchanged.
  - Language detection works reliably even when system information is unavailable.

### DIFF
--- a/changelog.d/1047.md
+++ b/changelog.d/1047.md
@@ -1,0 +1,3 @@
+### Added
+
+- **First-run system-locale detection.** A session that has never had a language chosen (fresh install, or a `session.json` from before the `locale` field existed) now picks up the OS locale via `navigator.language` and maps it to one of the 23 shipped languages — a region variant like `pl-PL` or `de-AT` matches its base language, and Traditional-vs-Simplified Chinese is routed by script/region rather than defaulting to one. A language nobody ships (or a locale string that doesn't parse) falls back to English, same as today. Detection runs exactly once: any explicit choice already on record in `session.json` — the user's own pick, or a previous run's detection — is left untouched.
--- a/src/preload/preload.ts
+++ b/src/preload/preload.ts
@@ -66,6 +66,16 @@ const electronAPI = {
       typeof process.getSystemVersion === 'function' ? process.getSystemVersion() : null,
     )
     : null,
+  // Raw OS locale (e.g. 'pl-PL', 'zh-Hant-TW') for first-run language
+  // detection (workspaceSlice's loadSession). `navigator.language` and not an
+  // IPC round-trip to `app.getLocale()`: sandboxed preload keeps the full Web
+  // platform surface (only Node builtins are restricted — see the
+  // getSystemVersion comment above), and Electron already seeds
+  // `navigator.language` from the OS locale, so this is available
+  // synchronously with no main-process hop. Static like the two fields above:
+  // the OS locale cannot change mid-session, so there is nothing to keep in
+  // sync.
+  systemLocale: typeof navigator !== 'undefined' && navigator.language ? navigator.language : 'en',
   pty: {
     // `exec`/`supervision` (X8): set by the AppLayout funnel for a supervised
     // wmux.json leaf — `exec` runs the command as the pane's ROOT process and
--- a/src/renderer/i18n/__tests__/detectSupportedLocale.test.ts
+++ b/src/renderer/i18n/__tests__/detectSupportedLocale.test.ts
@@ -1,0 +1,55 @@
+import { describe, it, expect } from 'vitest';
+import { detectSupportedLocale, LOCALE_OPTIONS } from '../index';
+
+describe('detectSupportedLocale', () => {
+  it('matches a plain supported code exactly', () => {
+    expect(detectSupportedLocale('pl')).toBe('pl');
+    expect(detectSupportedLocale('en')).toBe('en');
+  });
+
+  it('strips a region subtag when the base language is supported', () => {
+    expect(detectSupportedLocale('pl-PL')).toBe('pl');
+    expect(detectSupportedLocale('de-AT')).toBe('de');
+    expect(detectSupportedLocale('fr-CA')).toBe('fr');
+  });
+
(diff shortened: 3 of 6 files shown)


openwong2kim:

Adds first-run system-locale detection: `detectSupportedLocale` maps `navigator.language` to a shipped locale (with zh script routing) and `loadSession` applies it only when no explicit choice exists. Additive and low risk — explicit `data.locale` always wins, the pure helper never throws, well tested. P3: quality-of-life i18n default, not blocking.

openwong2kim:
Pushed one commit to your branch (`9ddde238`) rather than bouncing it back: the red CI was a single mechanical gap, not a design issue.

`loadSession` also runs under node-environment suites (`workspaceColor`, `workspaceProjections`, `paneOrdinal`, `remoteWorkspacesSlice`), where `window` does not exist at all — so the bare `window.electronAPI?.systemLocale` read was a `ReferenceError` that took 16 tests across those four files down with it. Your guard covered `electronAPI` being absent but not `window` itself; the preload half of the PR got this exact distinction right with `typeof navigator !== 'undefined'`, so the fix just applies the same shape on the renderer side (`typeof window !== 'undefined' ? window.electronAPI?.systemLocale : undefined`), which is also the pattern the codebase already uses in `uiSlice` and `events/publisher`.

All 16 previously failing tests plus your 124-test targeted set pass locally with the guard; `tsc` clean. This is also why the targeted `test:parallel` run in your test plan couldn't have caught it — the four broken suites are outside the paths the PR touches, and only a full run walks through them.

On the design itself: no notes. The detect-once semantics (an explicit `data.locale` always wins, detection result gets persisted by `setLocale` so the branch never re-fires), the zh script routing with English as the safe miss, and flagging `pt-PT`→`pt-BR` instead of silently deciding it — all read right to me, and `pt-BR` over English for a Portugal user is the call I'd make too.

PR #1048, opened 2026-08-26
Say when an install half-completed, from the two places that can still speak

### LINKED ISSUE #1046: P1: fresh install can silently half-complete — no icudtl.dat, no Update.exe, no error anywhere
### Problem

A fresh first-time install can silently fail partway through, on a real Windows machine, leaving an install directory that looks present but is unusable: the app never starts, uninstalling doesn't work, and nothing during the install itself told the user anything went wrong.

### Reproduction (real machine, real logs)

`%LOCALAPPDATA%\wmux\SquirrelTemp\Squirrel-Install.log` (from the actual failed install):

```
System.UnauthorizedAccessException: Odmowa dostępu do ścieżki
"C:\Users\<user>\AppData\Local\wmux\app-3.47.1\wmux.exe"
   at System.IO.File.InternalDelete(...)
   at Squirrel.Utility.<>c__DisplayClass13_1`1.<<ForEachAsync>b__1>d.MoveNext()
   ...
   at Squirrel.Update.Program.<Install>d__4.MoveNext()
```

The install throws an unhandled `UnauthorizedAccessException` trying to delete `wmux.exe` and aborts. The likely trigger: `wmux.exe` here is **~213 MB**, freshly written and never seen before by this machine — Windows Defender's real-time protection holds a transient lock on new, unrecognized executables of that size while it scans them. Squirrel's delete has no retry/backoff around a transient `ERROR_ACCESS_DENIED`, so one unlucky timing window kills the whole install. No AV *detection* was logged (`Get-MpThreatDetection` empty for the window) — this isn't a false-positive quarantine, just a lock race.

### What was left behind (diagnosed from the live machine)

```
PS> Test-Path "$env:LOCALAPPDATA\wmux\Update.exe"
False
PS> Test-Path "$env:LOCALAPPDATA\wmux\app-3.47.1\resources\icudtl.dat"
False
```

- `wmux.exe` and `resources\` got copied, but `icudtl.dat` (a required Chromium resource) never did — every launch attempt fails immediately with `[ERROR:base\i18n\icu_util.cc:232] Invalid file descriptor to ICU data received.` and **no window ever appears**, repeatedly, with nothing telling the user why.
- `Update.exe` (Squirrel's own stub, normally written near the end of a successful install) was never created — `fixPinnedExecutables`/shortcut-fixup/icon registration never ran either (confirmed: both the Start Menu entry and the desktop shortcut show a generic blank window icon instead of the app icon).
- Settings → Installed apps → Uninstall fails outright, pointing at the missing `Update.exe`:
  `System Windows nie może odnaleźć pliku „...\wmux\Update.exe"`.

None of this — not the ICU failure, not the missing uninstaller, not the icon — gave any indication of *why*, on a machine that had never run wmux before. The only usable diagnostic was the raw Squirrel log, which a typical user would never think to open.

### Suggested directions (not prescribing an exact fix)

1. **Retry the delete** (or whichever Squirrel operation raises the transient `UnauthorizedAccessException`) with a short backoff before giving up — this is a known class of Windows AV/indexer race, not a permissions problem, and a couple of retries over ~1-2s would likely absorb it.
2. **Detect an incomplete/corrupt install at launch** and say so, instead of four-plus silent ICU errors to a console nobody's watching: e.g. check for `icudtl.dat` (or just catch the resulting crash) before/around app init and, if missing, show *something* — even a plain MessageBox saying "the last install didn't finish, please reinstall" — rather than a window that never opens at all.
3. If retries still don't fully close the race, consider whether Squirrel's own error handling can be told not to abort the whole install on this specific step, so a transient AV lock degrades to "install completed, this one non-essential step failed" instead of "install silently half-happened."

### Impact

P1-shaped from the user's side: fresh install, never worked, no error surface anywhere in the app itself — only diagnosable by reading a Squirrel-internal log file. Confirmed fixed manually by deleting `%LOCALAPPDATA%\wmux` + `%LOCALAPPDATA%\SquirrelTemp` and reinstalling (AV didn't repeat the race on the retry), but nothing in the product would have told an average user to do that.

### Confirmation

- [x] Searched existing issues for `icu_util`, "Invalid file descriptor", "fresh install", "Update.exe" — no duplicate found.

openwong2kim:
Closes #1046

A Squirrel install can throw partway — the field log shows Defender's transient lock on the freshly written 213 MB exe turning a delete into `UnauthorizedAccessException` — and abort with the app dir half-copied: `wmux.exe` present, `icudtl.dat` never written, `Update.exe` never created. Nothing in the product said anything, at install time or ever after; the only diagnostic was Squirrel's own log file.

## The constraint that shapes the fix

An installation missing `icudtl.dat` dies before one line of our code runs — Chromium's ICU init fails first, so **no in-app check can ever fire on the machine that needs it**. Whatever detects this has to be a process that is still alive when the installer finishes. There are exactly two candidates, and this PR uses both:

**1. The install waiter (update path).** It already outlives the app by design. It now captures `Setup.exe` with `-PassThru`, waits for its exit (10-minute ceiling), and verifies what was left behind: the newest `app-*` dir must contain `icudtl.dat`, and the root must contain `Update.exe`. A stable corpse gets the abort marker (read as a toast on the next boot that can boot) **and** a MessageBox (the only surface that reaches a machine that cannot), then exits with its own code (6).

Fail-safe posture mirrors #1044's wait window: verification can only *add* a warning. An installer still running at the deadline means "cannot judge" and exits 0 exactly as before; so does a throw anywhere in the checks; the 30-second settle poll means the alarm fires only on a corpse that stayed a corpse. The MessageBox is gated on the Forms assembly having actually loaded, and marker/box failures are independent — a machine that cannot show UI still gets the marker, one that cannot write the marker still gets the box.

**2. A boot-time integrity check (the runs-but-broken class).** The reported machine also had the other shape available: an install can copy far enough to *run* but never get `Update.exe` — after which every in-app update fails and Windows' own uninstall entry points at a missing file, forever, silently. `installIntegrity.ts` detects the Squirrel layout from `process.execPath` (one stat, `path.win32` so the pure functions behave identically across the CI matrix) and surfaces one `dialog.showErrorBox` with reinstall guidance. English on purpose: an installer-corruption diagnostic for someone about to reinstall does not warrant 23 locale files plus the pl parity gate.

## What this deliberately does not do

- **Fresh installs stay uncovered.** `Setup.exe` run by hand has no waiter and no bootable app — no process of ours exists on that machine. That half of #1046 needs Squirrel-side retry or code signing (which shrinks the Defender scan window), both upstream of this repo. Said in the issue rather than silently narrowed here.
- **No retry of the installer.** The reporter's option 1 targets Squirrel's internals (`Squirrel.Update.Program.Install`), which we ship as a prebuilt binary.

Reporter's options 2 and 3 from the issue are what this implements, at the two layers that can actually speak.

## Verified

- `installTeardown.test.ts` + `installIntegrity.test.ts` — 64/64. New shape tests pin the fail-safe ordering (`-PassThru` → `WaitForExit` → cannot-judge exits 0 → checks → guarded marker → guarded MessageBox → `exit 6`), and the pre-existing ordering assertions (form closed before `Start-Process`, close-site count, mutex-before-anything) pass unchanged.
- Generated script validated with PowerShell's own parser — 0 errors.
- `tsc --noEmit` clean; eslint clean on all touched files (the 4 pre-existing `no-var-requires` in `index.ts` are on untouched lines).

### DIFF
--- a/changelog.d/1048.md
+++ b/changelog.d/1048.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **A half-completed install now says so, from the two places that can still
+  speak.** An antivirus scan holding the freshly written executable can make
+  the Windows installer abort partway, leaving an installation that either
+  never starts (a required Chromium resource was never copied) or starts but
+  can never update or uninstall (Squirrel's `Update.exe` was never written) —
+  in both cases silently, with the only diagnostic buried in the installer's
+  own log. On the update path, the install waiter now stays for the
+  installer's exit and verifies what it left behind: a broken result gets a
+  visible warning right there and a notice on the next boot that can boot.
+  An installation that runs but is missing `Update.exe` is detected at
+  startup and explained once, with reinstall guidance, instead of every
+  future update failing with no visible reason. Fresh installs run by hand
+  remain the installer's own responsibility — no process of the app exists
+  yet on such a machine. (#1048)
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -93,6 +93,7 @@ import { setConfiguredFirstPartyClients } from './mcp/firstParty';
 import { readConfiguredFirstPartyClients } from './mcp/firstPartyConfig';
 import { ClaudeWorker } from './a2a/ClaudeWorker';
 import { AutoUpdater } from './updater/AutoUpdater';
+import { warnOnInstallIntegrityGap } from './updater/installIntegrity';
 import { readDaemonPid } from './updater/installTeardown';
 import { McpRegistrar } from './mcp/McpRegistrar';
 import { BrokerSupervisor, isMcpBrokerEnabled } from './mcp/BrokerSupervisor';
@@ -429,6 +430,16 @@ const autoUpdater = new AutoUpdater(() => mainWindow, {
   onInstallRequiresFullShutdown: () => { fullShutdownRequested = true; },
   getDaemonPid: () => readDaemonPid(getWmuxDir()),
 });
+
+// #1046: a Squirrel install can half-complete (an AV lock race inside the
+// installer aborts it mid-copy) and leave an installation that RUNS but can
+// never update or uninstall -- Update.exe was simply never written. Detect
+// that shape once the app is up and say so, instead of letting every later
+// update fail with no visible reason. The dead-on-arrival half of the class
+// (icudtl.dat missing) cannot reach any in-app check by definition; the
+// install waiter's post-exit verification (installTeardown.ts) covers it on
+// the update path.
+void app.whenReady().then(() => { try { warnOnInstallIntegrityGap(); } catch { /* best-effort */ } });
 
 const rpcRouter = new RpcRouter();
 markBoot('pre-pipe-server-ctor');
--- a/src/main/updater/__tests__/installIntegrity.test.ts
+++ b/src/main/updater/__tests__/installIntegrity.test.ts
@@ -1,0 +1,68 @@
+// #1046 — the boot-time half of the half-completed-install detection: an
+// installation that RUNS but never got Update.exe written can never update
+// or uninstall, and before this check nothing ever said so. Pure functions
+// only — the dialog wiring is a thin best-effort shell over these.
+import { describe, it, expect } from 'vitest';
+import { squirrelInstallRootFor, findInstallIntegrityGap } from '../installIntegrity';
+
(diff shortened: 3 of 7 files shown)


openwong2kim:

Detects half-completed Squirrel installs (#1046) from the two places that can still speak: post-exit verification in the install waiter, plus a boot-time `Update.exe` check. Touches `installTeardown.ts` and adds `installIntegrity.ts` with tests; both paths are additive and fail-safe, so risk is low. P2 — matches the underlying issue; the new tests pin the fail-safe ordering well.

PR #1050, opened 2026-08-26
Let an authoritative running edge re-arm the activity cycle

### LINKED ISSUE #1045: Roster shows idle for a pane genuinely mid-turn when its tool call's own output is sparse
### Problem

A pane genuinely mid-turn — the CLI's own status footer visibly ticking ("Crunching… 7m 39s · ↓ 6.4k tokens") — can sit in the roster as idle/waiting for the whole duration of a long tool call whose own PTY output is sparse (a slow poll loop, long-running build, anything that writes well under 2 KB per 3 s).

Reported live: a pane running a bounded poll loop (`for i in $(seq 1 16); do gh api .../reviews; sleep …; done`, ~7+ minutes in) showed **Bezczynny** (idle) in the sidebar roster while the pane's own footer showed active elapsed/token counters the whole time — the app's *own* UI had two disagreeing opinions about whether the same pane was working.

### Root cause (traced, not guessed)

`ActivityMonitor` (`src/main/pty/ActivityMonitor.ts`) requires **2000 bytes within a 3 s window** to (re-)enter `active`; once active, *any* further byte resets the 5 s idle countdown, but getting IN requires clearing that bar. A CLI status line re-rendering itself once a second is a few dozen bytes — nowhere near 2000/3s — so if the cycle has already gone idle (or never started), sparse output like this can never cross back into `active` on byte throughput alone.

There are two documented bypasses for "let a smaller signal count":

- `beginTurn()` — armed from `noteInput()` (`DaemonPTYBridge.ts:180-207`) on a submitted keystroke boundary: "the next output byte is enough to prove this turn is running."
- `endTurn()` — armed from `noteAgentStatus()`'s non-running branch, added for #935 direction 3.

But `noteAgentStatus()`'s **`running`** branch does neither:

```ts
// DaemonPTYBridge.ts:216-224
noteAgentStatus(status: AgentEventStatus): void {
  if (status === 'running') {
    this.explicitTerminalStatus = false;
    this.settledStatus = null;
    // awaitingHuman deliberately survives...
    return;
  }
  ...
}
```

This is the hook path — `agent.tool_started` and friends route here (`hooks.rpc.ts`'s `agent-turn` category, `'Tool call started'` / `'Running a tool'` display text) — so the app *does* know, authoritatively, that a new tool call just began. But this branch only clears the settled-state flags and returns; it never calls `activityMonitor.beginTurn()`. So the one piece of code that could re-arm ActivityMonitor to accept the next (small) byte as proof of activity is never told this is exactly that situation. If the pane's byte throughput for this tool call never crosses 2000/3s on its own — a real, common shape for a polling loop — `onActive` never fires again, no fresh `running` ever reaches the roster, and the pane is stuck showing whatever it last was (here: idle) until either the loop happens to produce a real burst or the turn ends and a *different* mechanism (the hook's own terminal status) eventually settles it.

`onActiveToIdle`'s guard (`DaemonPTYBridge.ts:334-341`, `if (this.explicitTerminalStatus) return;`) doesn't help here either — the `running` branch is specifically the one that *clears* `explicitTerminalStatus`, so a stale idle firing from a cycle that predates this tool call is not blocked by it, and there's no promotion path once already idle.

### Why I'm filing rather than sending a patch

The obvious-looking fix — call `this.activityMonitor.beginTurn(this.sessionId)` inside the `running` branch, mirroring `noteInput()` — has a real risk I don't have confidence resolving alone: `beginTurn()` unconditionally resets `activeFired = false`, and `onActive`'s one-fire-per-cycle dedup exists specifically as "IPC spam protection." If `agent.tool_started`-class hook events fire more than once inside an already-active cycle (plausible — a turn with several sequential tool calls), an unconditional `beginTurn()` on each one would re-open `onActive` mid-cycle and could reintroduce duplicate `running` broadcasts, which is exactly the class of regression #1013 went through two review rounds to avoid. A correct fix likely needs to gate the re-arm on "cycle is not currently active" rather than fire unconditionally, and I'd rather that call be made by whoever has the clearest view of every caller of `noteAgentStatus('running')` and how often it actually fires per turn.

### Confirmation

- [x] Searched existing issues — this is the false-**idle** direction; #935's three directions and #1013 were all the opposite (false-**running**). Not a duplicate.
- Scoped to daemon mode: `src/main/pty/PTYBridge.ts` (local/main mode) has no `noteAgentStatus` at all, so this specific gap doesn't apply there — worth confirming that's intentional rather than a parity gap of its own.

openwong2kim:
Closes #1045

The report's trace was complete and correct, so this PR is small: `noteAgentStatus`'s `running` branch cleared the settled flags but never told the `ActivityMonitor` anything, so once a pane idled mid-turn, a tool call whose own output stays under the 2 KB/3 s threshold could never earn `running` back — the roster said idle while the pane's own footer ticked.

## The fix, and whose design it is

The edge now re-arms the byte cycle through a new `ActivityMonitor.beginTurnIfInactive()`: identical to the submitted-input re-arm (`beginTurn`), but a **no-op while the cycle is already active**. That gate is the reporter's own risk analysis adopted verbatim — running-class hook events arrive many times per turn (every metadata kind projects as `running`), and an unconditional `beginTurn` on each would reset `activeFired` mid-cycle and re-open the once-per-cycle `onActive` dedup that exists as IPC spam protection, which is the regression class #1013 spent two review rounds avoiding. The mid-cycle no-op keeps that dedup byte-for-byte.

## A test moved with the contract, deliberately

`DaemonPTYBridge.turnPriority.test.ts`'s "an explicit running edge re-opens the gate" asserted `[]` after the edge plus one sub-threshold byte — which pinned exactly the reported bug (the edge unlocked the gate but the threshold still applied). It now expects the broadcast on the first byte, and keeps pinning the dedup through the following burst. Calling this out rather than letting a green assertion quietly change meaning.

New coverage: monitor-level (idle→edge→sparse byte fires; mid-cycle edges are no-ops; unknown pty safe) and bridge-level (the field scenario end-to-end; repeated edges broadcast exactly once).

## Scope note, as the reporter scoped it

Daemon mode only. Local-mode `PTYBridge` has no `noteAgentStatus` at all — whether that is intentional or a parity gap of its own is the reporter's open question and stays open; this PR does not guess at it.

## Verified

- `ActivityMonitor.test.ts` + `DaemonPTYBridge.turnPriority.test.ts` — 65/65.
- `tsc --noEmit` clean; eslint clean on the changed lines (the one pre-existing `no-empty-function` in the test file's harness, line 43, predates this PR).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed roster panes incorrectly appearing idle while an agent was actively working mid-turn.
  - Sparse agent output following a running status now correctly restores the pane’s active state.
  - Repeated running signals no longer cause duplicate activity notifications within the same turn.


openwong2kim:

Fixes #1045: an authoritative running edge now re-arms the activity cycle via `beginTurnIfInactive`, so sparse tool-call output earns `running` again. Touches `ActivityMonitor` and the `DaemonPTYBridge` running branch only; the inactive gate preserves the once-per-cycle dedup (#1013), and coverage spans both levels. P2 — corrects a user-visible false-idle in the roster with a small, well-guarded change.

PR #1052, opened 2026-08-27
Stop shelling out to PowerShell for the port watch — Defender reads it as a trojan

### LINKED ISSUE #1051: [Bug]: wmux flagged as Trojan by Windows Defender
### What happened?

I've been trying to deploy the latest release of wmux on my company laptop for orchestrating Claude Code instances. Every time I try to run it, Windows Defender flags it as a Trojan/virus, quarantines and then removes the .exe.

This is the content of one of the quarantine messages as far as what's trying to execute:

CmdLine: C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -Command $p=Get-CimInstance Win32_Process -ErrorAction SilentlyContinue | Select-Object ProcessId,ParentProcessId;$c=Get-NetTCPConnection -State Listen -ErrorAction SilentlyContinue | Select-Object LocalPort,OwningProcess;@{procs=@($p);conns=@($c)} | ConvertTo-Json -Depth 3 -Compress

Image

### Steps to reproduce

To reproduce, you just need to have Windows Security enabled with Windows Defender. After that, simply install the wmux software and then try to run it. Sometimes it will come up. Sometime Defender will flag it immediately. 

If it does come up, just wait a minute and you should receive a threat notification pop-up.

### Expected behavior

The application should not be flagged as a Trojan. 

### wmux version

3.47.1

### How did you install wmux?

Setup.exe (direct download)

Not applicable

### Windows version

Windows 11 Version 24H2 (OS Build 26100.9106)

None / not relevant

```shell

```

### Confirmation

openwong2kim:
Closes #1051.

A reporter's Defender quarantined wmux outright, and the command line in the
quarantine log is ours, byte for byte:

```
powershell.exe -NoProfile -Command $p=Get-CimInstance Win32_Process ...;
$c=Get-NetTCPConnection -State Listen ...; ConvertTo-Json
```

That is `portWatch`'s snapshot, and it ran **every 10 seconds** for as long as
a single pane was open. An unsigned executable sweeping every process and every
listening socket on a fixed heartbeat is a textbook recon signature; the sidebar
port chips are simply not worth looking like malware for. Signing would blunt
this, but that is still waiting on SignPath's production policy, so this fixes
the part we control: the behaviour itself.

## What changed

The snapshot now runs in-process through the same user-mode APIs the cmdlets
wrapped — `GetExtendedTcpTable` (AF_INET and AF_INET6) and
`CreateToolhelp32Snapshot`, via koffi's prebuilt Node-API binary. No child
process, no command line, nothing for a behavioural heuristic to match.

The PowerShell path is **deleted, not kept as a fallback**. The detection
matched the command bytes, so leaving the string in the bundle would re-ship the
signature — the same call we made for the ACL path in GHSA-8fj2-47w9-jxq3. A
test scans both sources for spawnable PowerShell artifacts so it cannot creep
back. When the native path is unavailable the watcher degrades to the existing
quiet-absence policy: the chips just don't render.

koffi and its `@koromix/koffi-win32-x64` prebuild sidecar ship beside node-pty
in both `app.asar` (scope unpacked so dlopen works) and
`daemon-bundle/node_modules`. Packaging a Windows build without them fails
loudly rather than shipping a dead feature.

## Review round

Three reviewers (Claude, GLM-5.3, Grok) ran independently. They converged on one
real defect in the first commit: a native snapshot that couldn't run resolved to
an *empty* table, which reads as "this machine has no listening ports" — wiping
live chips, leaving the failure backoff unreachable, and killing a2a.rpc's
skip-the-retry branch. It rejects now, matching what the subprocess path did.
Grok separately caught two silent-failure doors: `Process32FirstW` returning 0
was read as end-of-list (a failed walk reported success with an empty table),
and the listener-table size — an out-param — was fed to `Buffer.alloc` uncapped.
Both are closed, and a snapshot that omits our own pid is now rejected, which
doubles as a runtime guard against struct-offset drift.

One P1 claimed `th32ParentProcessID` sits at offset 28. It doesn't: offset 32
matched the parent pid for 274/274 processes cross-checked against an
independent source, offset 28 for 1. Offset 32 stands.

## Verification

- Unit tests for the buffer decoders, the failure contract, and the backoff;
  a runtime test drives the real Win32 calls and asserts our own pid/ppid and
  that a listener we open on both `127.0.0.1` and `::1` is attributed to us.
- Packaged build: koffi loads and calls successfully from inside `app.asar`
  under the Electron runtime, and from `daemon-bundle` under system Node — the
  two runtimes that actually host this code.
- Live dogfood: a dev server in a pane produces its `:5099` chip, Ctrl+C clears
  it, and a 90 s process-table watch shows **zero** enumeration spawns under the
  new build — while the installed 3.44.x daemon on the same machine kept
  spawning one every ~4.5 s, which is the shipped bug in miniature.
- Cost measured rather than assumed: median 5.7 ms, p95 7.3 ms on a 287-process
  machine, about 0.06% of a 10 s tick.

## Not in scope

`paneResources` (4 s, only while Fleet View is open) and `AgentProcessTracker`
(event-driven, and it needs `CommandLine`, which native enumeration can only get
by reading another process's memory — worse than what we're removing) still use
PowerShell, as does the updater's `-ExecutionPolicy Bypass` waiter. Those are
follow-ups; this PR removes the one that was actually flagged.

Users already quarantined on 3.47.x will need to restore or reinstall. And the
honest caveat: this removes the trigger that matched, not every reason an
unsigned binary can be flagged — the reputation half needs signing.

## Summary by CodeRabbit

* **Bug Fixes**
  * Windows listening-port detection no longer launches PowerShell, reducing false-positive antivirus detections.
  * Workspace listening-port chips continue to display as before.
  * Polling pauses temporarily after repeated detection failures and resumes automatically.

* **Documentation**
  * Updated technical documentation and changelog details for Windows detection improvements.

* **Chores**
  * Updated third-party license notices and Windows packaging for newly included components.


openwong2kim:

Replaces the 10 s PowerShell port-watch sweep with in-process koffi FFI (GetExtendedTcpTable, CreateToolhelp32Snapshot), removing the Defender recon signature from #1051 with no fallback path. It adds a native dependency to both bundles, and all three Baseline typechecks currently fail on `import.meta` in `winSnapshotNative.test.ts`. P2 — direct fix for the quarantine report, blocked until that typecheck error is cleared.

PR #1054, opened 2026-08-27
docs(todos): register the changelog collector's [Unreleased] papercut

openwong2kim:
Two consecutive release commits (3.47.1 and 3.48.0) claimed this papercut was already tracked in `TODOS.md`. It wasn't — the only `[Unreleased]` line there is about duplicated grapheme entries from #671, which is unrelated. 3.48.0 repeated the claim from 3.47.1 without checking it, which is on me.

This adds the entry those commits were pointing at, so the next release cut finds it where it was promised to be.

The papercut: `collect-changelog.mjs` folds the fragments under `## [Unreleased]` and leaves nothing in its place, so the next release has to re-add the heading by hand or `applyToChangelog` throws (`changelog-fragments.mjs:117`). Restoring an empty stub after the fold would close it.


openwong2kim:

This PR registers the changelog collector papercut in TODOS.md — `collect-changelog.mjs` consumes the `## [Unreleased]` heading, so each release re-adds it by hand or `applyToChangelog` throws. It is docs-only, one line, so nothing can break. P3: bookkeeping that just makes the real fix findable next release.

PR #1059, opened 2026-08-27
fix(updater): make the install waiter's launch provable, and survivable

### LINKED ISSUE #1056: Update install can silently no-op: the detached waiter's PowerShell process appears to die before it ever runs, right as the app quits
### Problem

On a real Windows machine (v3.47.1), clicking "Update" produced the pre-#1044 symptom exactly: the app quit and nothing happened afterward — no install, no marker on next boot, stuck on the old version. Traced with Windows Event Log evidence (`Microsoft-Windows-PowerShell/Operational`), not just re-reading the code.

### Timeline (from `logs/main-2026-08-27.log` + PowerShell Operational log, times in local CEST)

- `11:54:07` — user clicks Update.
- `11:54:12` and `11:54:14` — two short-lived PowerShell engines start and reach "ready for user input" within ~1s each (event 40961 → 53504 IPC thread → 40962). These match `collectInstallRootProcesses()`'s two `execFileSync` calls to `Get-CimInstance Win32_Process`.
- `11:54:16` — app log: `[AutoUpdater] install handoff: waiter=C:\...\Temp\wmux-install-waiter-aQOgl1\wait-and-install.ps1 watching 4 process(es)...`, immediately followed by `[AutoUpdater] quitting for install` and `app.quit()`. Event log shows a **third** PowerShell engine event 40961 ("starting up") at the same second.
- After `11:54:16`: **nothing**. No `40962` for that third engine — ever. No IPC-thread event for it. Compared to the two engines 4 seconds earlier that both completed their startup in ~1s, this one never got past "starting up," and the PowerShell Operational log goes completely silent for it through at least `11:56:30`.

Consistent with this: `wait-and-install.ps1` exists on disk (written before spawn, as expected) with its original timestamp, but no abort marker was ever written (`update-install-aborted.txt` — absent), no Squirrel log exists (`SquirrelTemp` doesn't exist), and the app is still on `app-3.47.1`. Every one of the three explicit abort paths in `buildWaiterScript` (`exit 2/3/4`, all preceded by `Set-Content $marker`) requires reaching at least the mutex line — and even `exit 5` (silent mutex-collision bailout, the one abort path that's deliberately marker-less) requires the engine to be far enough along to evaluate a `New-Object System.Threading.Mutex`. None of that shows up. The evidence points at the process dying before executing a single line of the script — i.e., before the engine finished spinning up.

### Leading hypothesis (not confirmed — flagging for whoever picks this up to verify against Node's actual Windows behavior before trusting it)

`spawnInstallWaiter` (`src/main/updater/installTeardown.ts`) does:

```ts
const child = spawn(
  powershell,
  ['-NoProfile', '-NonInteractive', '-ExecutionPolicy', 'Bypass', '-File', scriptPath],
  { detached: true, stdio: 'ignore', windowsHide: true },
);
child.unref();
```

immediately followed by `app.quit()` in the caller. On Windows, `detached: true` in Node's `child_process.spawn` sets `CREATE_NEW_PROCESS_GROUP` — it does **not** request breakaway from any Job Object the parent belongs to (that would need `CREATE_BREAKAWAY_FROM_JOB`, which Node doesn't expose). Electron/Chromium processes are commonly Job Object members; if that job has `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` and doesn't explicitly permit breakaway, then a child spawned by a job member is added to the *same* job by default, and the whole job — including a just-spawned "detached" child — dies with the parent. The timing here (waiter engine starts the same second `app.quit()` fires, and never gets further) fits that mechanism exactly, though I haven't independently confirmed wmux's own process is in such a job.

### Why this is worth a look now

This predates everything the script does after startup — #1044's wait window and #1048's Setup.exe verification both live *inside* the script, so neither would ever have helped here; the process never got that far. It also isn't a regression: v3.47.1 (pre-#1044) has the same unconditional `spawn(...detached...)` immediately before `app.quit()`, so this class of failure — silent update no-op with zero trace — has likely been live since this mechanism was introduced.

### Confirmation

- [x] Searched existing issues for "detached", "job object", "CREATE_BREAKAWAY", "waiter" + "quit" — no duplicate found.

### LINKED ISSUE #1060: Waiter heartbeat refusal can strand a slow-but-alive waiter (follow-up to #1057)
Follow-up to #1057 (merged), from its review. Not a regression — the heartbeat fail-safe is strictly better than what shipped before — but one window is left open on exactly the machine class #1056 lives on.

## The gap

`WAITER_HEARTBEAT_BUDGET_MS` is 3s, on the reasoning that "a live process writes one file in well under a second." On an AV-heavy machine that premise fails: PowerShell 5.1 cold start plus AMSI/Defender scanning a freshly written `.ps1` in TEMP can exceed 3s — and those are precisely the machines where the waiter transport misbehaves. The cost of the misread is asymmetric:

1. `performInstall` concludes the waiter never started and refuses (correct UX, nothing torn down) — but
2. the waiter is actually alive, and nothing cleans it up. Nothing was killed and no quit happened, so its handle-waits burn the full 60s `lockBudgetMs` while the #1044 TopMost wait window sits over an app the user is still using, and it then writes the `exit 3` marker ("a process under the install root would not exit") — so the **next boot reports a refusal for an install the app already refused live**, with a reason that was never true.

On the fast-kill machine measured in #1056 this window does not open (the waiter dies pre-first-line and leaves zero residue — verified by live-firing the merged code there). It only opens on slow-start-but-alive machines, which cannot be ruled out on the evidence we have.

## Suggested fix

- Raise the budget to ~6–8s (still well under the 20s before-quit deadline, which is not armed at that point).
- On the refusal branch, tree-kill the spawned waiter pid (`terminatePids` already exists and `/T` reaches it) and clear the ready/abort markers it may have written, so a late-blooming waiter cannot double-report or park its window over the running app.

Both fold naturally into the #1059 rebase (its transport work already carries a refusal-path cleanup of the same shape), so this issue can be closed by that PR — filed separately so the gap is tracked even if #1059 changes shape.

Also worth carrying in whichever PR lands second: #1057's two new `UPDATE_ERROR` sends (the stale-marker refusal and the did-not-start refusal) predate #1058's `source: 'install'` tagging and need the tag added post-merge, or those two errors stay invisible to the toast surface on one-shot installs.

openwong2kim:
Rebased onto #1057 (the heartbeat fail-safe) — the two changes now layer instead of overlapping:

- **#1057 (merged)** owns the outer gate: `performInstall` refuses to tear anything down until the waiter's `$ready` heartbeat proves a process ran.
- **This PR** owns the transport underneath: making a waiter exist for that heartbeat to come from, on the machines where today's spawn dies.

## The transport problem

On current Windows builds a PowerShell spawned directly with `detached: true` dies before executing its first line — measured on two real machines (the reporter's event log in #1056 and a second machine): engine event 40961 with no 40962, exit 0, empty stderr, no AV detection, parent lifetime and Job Objects irrelevant; pwsh 7 dies identically; detached node.exe/cmd.exe children survive. With #1057 alone those machines get a *correct refusal* forever — updates still can't complete. This PR makes them complete:

- **Transport A — cmd.exe trampoline.** A PowerShell created by a detached cmd survives where a directly-detached one dies (measured, including under a deliberate `KILL_ON_JOB_CLOSE` job). argv-array spawn: libuv does the CRT quoting, and `cmd /c`'s strip-first-quote rule cannot fire because the System32 powershell path carries no spaces.
- **Transport B** — the direct detached spawn, kept for machines where A is unavailable but B still works.
- Each transport proves itself via a **per-transport launch stamp** (written right after the `$ready` heartbeat, still before the mutex — so a transport verified inside `spawnInstallWaiter` has already written `$ready`, and #1057's outer check passes instantly). Distinct stamp files per transport, so the log can never misattribute which one worked.
- **No stamp from either → tree-kill both attempts, clear what they wrote, return null** — `performInstall`'s existing refusal fires with nothing killed and no quit armed. This closes #1060: the slow-but-alive waiter that #1057's 3s outer budget could strand (later writing a contradictory exit-3 marker under a TopMost window) is now killed on the refusal path, and the inner gate gives transport A a 6s budget sized for AV cold starts.
- Once the mutex is held, the waiter writes an **"interrupted" sentinel** over the abort-marker path and removes it on both success exits — a waiter killed mid-wait reports on the next boot instead of vanishing.
- Both transports pin `cwd` outside the install root; the `Atomics.wait` sync sleep degrades per-step to a bounded busy-wait.

Alternatives measured and rejected: WMI `Win32_Process.Create` passed every gate (user env/session/token intact, WinForms renders under `ShowWindow=0`) but adds an ASR-blockable LOLBin pattern for zero benefit over the trampoline; a one-shot scheduled task also survives but needs lifecycle management; a conhost wrapper dies like the bare spawn.

## Tests

Script-shape pins for the stamp (after the heartbeat, before the mutex; best-effort; absent when not requested; double-quote paths refused) and the sentinel (after the mutex gate, removed on both success exits); the updated #1046 pin; and a runtime smoke suite that runs the **real `spawnInstallWaiter`** end to end — asserting the waiter executes with the caller's environment intact (`Setup.exe` resolves its target from `%LOCALAPPDATA%`) and survives a TEMP containing spaces and an apostrophe, plus the mid-wait sentinel pin and the temp-dir leak fix in the script-placement test. Billed honestly: on runners where the detached spawn works this suite is green either way; what it closes is that no test anywhere ran the real transport.

Unit + platform suites and the full runtime suite green on a real Windows box after the rebase; tsc and eslint clean. Plan reviewed twice (initial + delta) before implementation.

The notice layer that surfaces these failures is #1058. Closes #1056. Closes #1060.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Windows installer reliability by verifying that the post-quit update helper launches successfully before proceeding.
  - Updates are now refused when the helper cannot be confirmed, preventing incomplete or unsafe installations.
  - Interrupted updates are detected and reported the next time the app starts.
  - Added improved handling for installer completion, timeouts, temporary paths, and cleanup.

### DIFF
--- a/changelog.d/1059.md
+++ b/changelog.d/1059.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **Clicking Install can no longer quit wmux into a silent no-op.** On
+  current Windows builds the helper process that runs the installer after
+  wmux quits could die before executing a single line, leaving the old
+  version in place with no error and no trace anywhere. The helper's launch
+  is now verified before wmux commits to quitting — if it cannot be proven
+  to run, the update is refused on the spot with the app still open — and a
+  helper that dies partway through now leaves a note that wmux shows on the
+  next start. (#1059)
--- a/src/main/updater/__tests__/installTeardown.runtime.test.ts
+++ b/src/main/updater/__tests__/installTeardown.runtime.test.ts
@@ -113,7 +113,10 @@ describe.skipIf(!onWindows)('install waiter (real processes, real locks)', () =>
     const { timedOut } = runWaiter(12_000);
     expect(timedOut).toBe(true);
     expect(fs.existsSync(setupStamp)).toBe(false);
-    expect(fs.existsSync(abortMarker)).toBe(false);
+    // #1056 — mid-wait the incumbent's interrupted sentinel is ON DISK by
+    // design (it is what reports a waiter killed before any terminal); the
+    // pin here is that no TERMINAL reason has been written yet.
+    expect(fs.readFileSync(abortMarker, 'utf-8')).toContain('interrupted before it could report');
   }, 120_000);
 
   it('gives up instead of waiting forever on a process that will not die', () => {
@@ -225,7 +228,17 @@ describe.skipIf(!onWindows)('install waiter (real processes, real locks)', () =>
     const written = spawnInstallWaiter(plan([process.pid], 1_000));
     expect(written).not.toBeNull();
     expect((written as string).toLowerCase().startsWith(root.toLowerCase())).toBe(false);
-  }, 30_000);
+    // #1056/P2-6 — this call now really launches a waiter (it terminates
+    // itself in ~1s: its handle wait is against our own live pid on a 1s
+    // budget). Reap the temp dir this test used to leak every run; retried
+    // briefly because the dying waiter can hold its script file a moment.
+    const leaked = path.dirname(written as string);
+    const rmDeadline = Date.now() + 10_000;
+    for (;;) {
+      try { fs.rmSync(leaked, { recursive: true, force: true }); break; }
+      catch { if (Date.now() > rmDeadline) break; execFileSync(PS, ['-NoProfile', '-Command', 'Start-Sleep -Milliseconds 500'], { windowsHide: true }); }
+    }
+  }, 60_000);
 
   it('enumerates nothing for a root no process runs from', () => {
     expect(collectInstallRootPids(root)).toEqual([]);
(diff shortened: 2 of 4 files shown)


openwong2kim:

Fixes #1056's silent update no-op: the waiter launches via a cmd.exe trampoline (direct spawn fallback) and must write a launch stamp before wmux quits, else the install is refused. Touches the core update handoff in installTeardown.ts; the risk of falsely refusing slow transports is covered by generous budgets and runtime tests. P2, matching the issue.

openwong2kim:
Converted to draft: #1057 (which was open first and lands the same execution-proof idea at the right layer) is the base for this issue. After it merges, this PR gets rebased down to the additive parts only — the cmd-trampoline transport, the interrupted-sentinel, and the runtime transport smoke tests. See the coordination note on #1057.

PR #1063, opened 2026-08-28
feat(browser): snapshot auto-diff, selector scoping, inline lifecycle events

openwong2kim:
## What

Compresses the agent's act→verify loop in the MCP browser toolset — three techniques measured to dominate real agent-browser workloads (diff-sized verify reads, interactive-only trees, no polling for page state):

- **Snapshot auto-diff** — `browser_snapshot` returns a line-LCS diff against the previous snapshot of the same surface when that is smaller. First line always declares `[snapshot: full]` or `[snapshot: diff …]`; `full:true` forces the complete tree. Baselines are per-connection, keyed by workspace+surface+attrs, carry the page URL (a diff is never produced across different URLs), and are invalidated on navigation.
- **Scoping** — new `selector` param scopes the snapshot to a subtree via the shared DOM expression on both transports (Playwright and packaged RPC); new `filter:"interactive"` strips non-interactive nodes up front instead of only on overflow.
- **Inline lifecycle events** — `BrowserCaptureManager` enables the Page domain and records main-frame navigations (including SPA `navigatedWithinDocument`), load events, and guest closure; a new `browser.lifecycle.get` RPC drains them destructively and `withAutomationLease` prepends them to every `browser_*` tool result, so page changes surface inline instead of requiring a polling round-trip.

## Hardening (multi-model panel review)

URL guard on baselines, `filter` with zero interactive nodes reports instead of returning the full tree, `Page.enable` failure can no longer take down console/network capture, selector misses are errors and never become baselines, `lastLifecycleTarget` is bounded, and the diff DP bound is 800 lines to keep the synchronous LCS off the hot path.

## Tests

19 new tests: diff generation/fallback/bounds, URL guard, interactive filter, selector scoping + stale-ref wipe, lifecycle capture/cap/destructive drain/close synthesis, RPC handler cases, lease injection + baseline invalidation. Existing fallback suites updated to the new drain contract. `tsc` clean; targeted suites 380/380; full `npm test` green except the two known symlinked-node_modules false failures (notices, atlasCoherence).

## Summary by CodeRabbit

* **New Features**
  * `browser_snapshot` now supports compact automatic diffs, full snapshots, selector-scoped views, and interactive-only filtering.
  * Browser tool results now include recent navigation, load, and tab-closure events inline.
  * Added lifecycle retrieval through `browser.lifecycle.get`.
* **Documentation**
  * Updated the changelog and API reference to document the new snapshot and lifecycle capabilities.

### DIFF
--- a/changelog.d/1063.md
+++ b/changelog.d/1063.md
@@ -1,0 +1,4 @@
+### Added
+
+- `browser_snapshot` now returns a compact diff against the previous snapshot of the same page when that is smaller (`full:true` forces the complete tree), and accepts `selector` to scope the snapshot to a subtree and `filter:"interactive"` to strip non-interactive nodes.
+- Browser tool results now report page lifecycle changes inline — navigations (including SPA route changes), page loads, and tab closure that happened since the last tool call — so agents no longer need to poll for page state.
--- a/docs/api/reference.md
+++ b/docs/api/reference.md
@@ -26,7 +26,7 @@ returns `EPERM`. Wire framing: newline-delimited JSON, one object per line.
 
 ## RPC methods
 
-Total: **151** methods (`ALL_RPC_METHODS` in
+Total: **152** methods (`ALL_RPC_METHODS` in
 `src/shared/rpc.ts`). Capability and risk class are read from
 `src/main/mcp/methodCapabilityMap.ts`:
 
@@ -139,6 +139,7 @@ Total: **151** methods (`ALL_RPC_METHODS` in
 | `browser.screenshot` | `browser.screenshot` | `browser` |
 | `browser.evaluate` | `browser.evaluate` | `browser` |
 | `browser.console.get` | `browser.read` | `browser` |
+| `browser.lifecycle.get` | `browser.read` | `browser` |
 | `browser.network.get` | `browser.read` | `browser` |
 | `browser.responseBody.get` | `browser.read` | `browser` |
 | `browser.type.cdp` | `browser.type` | `browser` |
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 80000,
-      "wireResultSha256": "9ffa0f077647435fd41855d7d0d8d28a7ca94acbbdc27bcf58fc2d4256134969",
+      "wireResultSha256": "d2a28f9455afd75a7c92c050863084e373746bd7ece5c132ab47603c4eb8a407",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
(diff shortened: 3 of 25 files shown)


openwong2kim:

Adds snapshot auto-diff, selector scoping, and inline lifecycle events to the MCP browser toolset, including a new `browser.lifecycle.get` RPC. It changes `browser_snapshot`'s default output and grows the RPC surface to 152; hardening bounds the diff DP and URL-guards baselines. P2 — broad additive change to a core agent tool; well tested, but the default output change needs close review.

openwong2kim:
## Live dogfood (Windows, dev build of this branch)

Ran a dev instance off this branch and drove the **dev build's own MCP server**
(`dist/mcp/mcp/entry.js`, stdio) against a local test page, so every call below is
the real act→verify path rather than a unit harness.

### Held up

| check | result |
|---|---|
| first snapshot / repeat snapshot | `[snapshot: full]` → `(no changes since previous snapshot)` |
| diff hunk on a renamed button | `- [ref=0] button "Alpha"` / `+ [ref=0] button "RENAMED-ALPHA"`, 163B vs 345B full |
| diff hunk on an added button | correct insert + renumber of the following ref |
| `full:true` | forces the complete tree, and still re-baselines |
| diff not worth it (+900 nodes) | falls back to full instead of a diff bigger than the tree |
| `selector:"#panel"` | scoped listing with a `Scope:` line |
| selector miss | `isError`, and the miss does **not** become the baseline (verified with a repeat scoped snapshot right after) |
| lifecycle — real navigation | `- navigated: …/page2` + `- loaded`, and the snapshot came back full (baseline correctly invalidated) |
| lifecycle — SPA `history.pushState` | `- navigated: …/page2?spa=1` |
| lifecycle — non-snapshot tool | prepended to `browser_evaluate` too |
| lifecycle — destructive drain | next call was clean |

### Two papercuts, both on the RPC/webview transport

**1. `(no changes since previous snapshot)` is a claim about the snapshot's
representation, not about the page.** I changed an `<li>`'s `textContent`, confirmed
via `browser_evaluate` that it had actually landed (`MUTATED-ROW-3`), and the next
snapshot still said `(no changes since previous snapshot)`. Same for anything past
the DOM listing's `.slice(0, 100)` cap — I renamed element #500 of ~900 and got
"no changes" again.

Nothing here regressed: that listing never carried non-interactive text. But before
this PR a repeat snapshot returned the tree and the agent drew its own conclusion;
now we actively assert "nothing changed", which is a stronger statement than we can
back. Something like *"no changes to the interactive elements in this snapshot"*
would keep the savings without the overclaim.

**2. `filter:"interactive"` is silently ignored on that transport.** The RPC branch
calls `buildDomSnapshotExpression()` with no filter plumbing, so `filter:"interactive"`
came back byte-identical to the full snapshot (345B/13 lines both). Harmless in
effect — that listing is already interactive-only — but the `selector` branch is
explicitly honest in the same situation (`(note: format/filter are ignored when
selector is given)`), and this branch deserves the same line.

### One cross-PR note

Under the `chrome` backend from #1064 the inline-lifecycle feature goes **completely
silent** — `browser.lifecycle.get` has no chrome fallback and `drainLifecycleEvents`
swallows the failure by design. Measured and written up in #1064.

PR #1064, opened 2026-08-28
feat(browser): 'chrome' backend — dedicated real-Chrome instance over CDP

openwong2kim:
## What

Adds a third `BrowserBackend`, **`chrome`**: wmux launches the user's installed Chrome with a dedicated `--user-data-dir` (persistent logins, fully separate from their daily browser) and `--remote-debugging-port`, and the existing MCP `browser_*` toolset drives it through the already-shipped `connectOverCDP` path. Electron-webview sign-in blocks (Google) do not apply, and automation/ban risk stays isolated to a dedicated profile. Stacked on #1063.

- **ChromeLauncher** — binary discovery (`WMUX_CHROME_PATH` override), singleflight lazy launch with `/json/version` readiness, tab registry (only wmux-opened tabs are addressable, workspace-tagged), `/json/new|list|close`, dispose wired into both quit paths.
- **RPC** — three-way backend forks; `browser.open/tabs/navigate/cdp.info` chrome branches; chrome `cdp.info` reports the Chrome port with registry targets and no `shellUrl`; webview-RPC-fallback handlers fail with a dedicated `CHROME_BACKEND_RPC_UNSUPPORTED` contract error; the navigate fallback takes the caller-verified scope, never a body `workspaceId` (#810 guard).
- **Engine** — `isShellPage` is a no-op under the chrome marker (localhost pages are real tabs, not the app shell); a stale `shellUrl` is dropped on a builtin→chrome flip; scoped-target selection and the auto-open→pin contract carry over unchanged (`surfaceId` = Chrome targetId).
- **Settings** — third option (options derive from `BROWSER_BACKENDS`), en/ko/zh/pl strings.
- **PortAllocator** — env-var override name param so the Chrome range (18900–18999) cannot conflict with `WMUX_CDP_PORT`.

Out of scope (follow-ups): console/network capture for chrome tabs via Playwright page listeners, per-tab lifecycle events, addressing manually-opened tabs, M144 live-Chrome attach.

## Tests

13 new: ChromeLauncher spawn/singleflight/relaunch/registry-pruning (4), chrome RPC branches incl. contract error + empty lifecycle + missing-launcher message (6), engine marker behavior (2), settings mirror (1). MCP wire surface unchanged (`probe:mcp` green). Full `npm test`: green except the two known symlinked-node_modules false failures (notices, atlasCoherence).

## Summary by CodeRabbit

* **New Features**
  * Added a dedicated Chrome browser backend for workspace browsing.
  * Added per-workspace Chrome profiles, profile creation, and profile selection.
  * Added Live Chrome integration with a consent prompt for connecting to an existing Chrome session.
  * Added Chrome support for tabs, navigation, screenshots, CDP connections, and lifecycle tracking.
* **UI Improvements**
  * Added Chrome backend selection in Settings and Chrome profile controls to workspace menus.
  * Added localized labels and guidance for Chrome profiles and Live Chrome access.


openwong2kim:

Adds a 'chrome' browser backend — a dedicated real-Chrome instance with its own profile, driven over CDP via the existing Playwright path (ChromeLauncher, RPC forks, settings, i18n). Risk sits in the browser.rpc backend forks and tab-ownership scoping; the #810 navigate guard holds, with 13 new tests. P2 — security-adjacent RPC surface deserving prompt review, though additive and fail-closed.

openwong2kim:
Scope grew (same feature line): per-workspace Chrome profiles landed on this branch — one named profile = one user-data-dir/instance, workspace→profile binding from the workspace card menu (bind-only, never agent-selectable), per-workspace CDP port in cdp.info, engine reconnect guard so a live connection can never pin workspace B to workspace A's browser. 12 new tests; full suite green except the two known symlinked-node_modules false failures.

openwong2kim:
## Live dogfood (Windows, dev build of this branch)

Dev instance off this branch, driven through the **dev build's own MCP server**
(`dist/mcp/mcp/entry.js`, stdio) against a local test page. Backend flipped
`builtin` ↔ `chrome` at runtime through the same IPC the Settings panel uses, so
every result below has a builtin control run.

### Held up

- Backend switch and cold Chrome launch: first `browser_open` **5.5s**, second one no relaunch.
- **Tab registry is real**: I created a tab directly against Chrome (`PUT /json/new?http://example.org/`) and `browser_tabs` never listed it, while the three wmux-opened tabs were all there.
- **Cross-workspace isolation holds**: a second MCP process carrying the *other* workspace's identity got `"tabs": []`. (My first attempt passed `workspaceId` in the tool params and saw everything — that was my harness, not the code; params are self-asserted and correctly ignored.)
- **Per-workspace profiles work** (Phase 2.5 — worth adding to the PR body, it's only in the changelog): after binding workspace B to a new `dogfood2` profile, opening there launched a *second* Chrome on port **18905** with `user-data-dir=…\wmux-dev\chrome-profiles\dogfood2`, while workspace A stayed on **18900** / `chrome-agent-profile`.
- **22 of 24** `browser_*` tools green: snapshot, smart_snapshot, extract_text, extract_data, evaluate, console, network, click, type, fill, press_key, hover, scroll, scroll_into_view, highlight, cookies, storage, resize, emulate, wait, tabs, navigate.
- Force-killing the app took every Chrome process with it — no zombies (19 → 0).

### P1 — `browser_navigate` opens a new tab instead of navigating the target, and reports success

Clean Chrome, one tab:

```
1) open                              -> surfaceId 4D75A1EA   url .../
2) browser_tabs                      -> 4D75A1EA = .../
3) evaluate location.href            -> http://127.0.0.1:18321/
4) browser_navigate({surfaceId: 4D75A1EA, url: .../page2})
                                     -> "Navigated to http://127.0.0.1:18321/page2"
5) browser_tabs                      -> 4D75A1EA = .../     F58869E8 = .../page2   <-- new tab
6) evaluate location.href            -> http://127.0.0.1:18321/                    <-- still page one
7) snapshot URL line                 -> http://127.0.0.1:18321/                    <-- still page one
```

The chrome fallback's comment assumes *"pinned-tab navigation rides the engine's
Playwright path and never lands here"*, but `browser_navigate` in
`src/mcp/playwright/tools/navigation.ts` always goes through
`sendScopedBrowserRpc('browser.navigate', …)` — *"Use RPC for fast, reliable
navigation (bypasses Playwright CDP discovery)"*. So it always lands there, and the
fallback calls `openTab(navUrl, scope)` without ever looking at `params.surfaceId`.

Two consequences: the agent navigates and then reads the **pre-navigation** page —
precisely the act→verify loop #1063 exists to compress — and every navigate leaks a
tab. The builtin control run navigates the same surface correctly.

### P1 — the inline lifecycle events from #1063 are silent under `chrome`

`browser.lifecycle.get` is a `registerLeased` method with no `chromeFallback`, so it
throws `CHROME_BACKEND_RPC_UNSUPPORTED`, and `drainLifecycleEvents` swallows that by
design (older mains must not break tools). Measured: navigate, then the next tool
call carries **no** `[browser events since last tool call]` block. Same sequence on
builtin: block present. So merging this on top of #1063 quietly removes one of that
PR's three features whenever the user picks this backend.

### P2 — full-page `browser_screenshot` is impossible under `chrome`

`browser_screenshot` only takes the Playwright path when `ref` is given; otherwise it
goes to RPC by design (*"bypasses Playwright CDP discovery"*), and `browser.screenshot`
has no chrome fallback → `CHROME_BACKEND_RPC_UNSUPPORTED`. Element screenshots do
work (`ref=0` returned a 976-byte PNG), so it's the whole-page case that's out — the
one agents use most for visual verification.

### P2 — `browser_navigate_back` fails under `chrome`

Same shape: RPC-only tool, `browser.goBack` has no chrome fallback.

### P2 — non-first-party MCP clients lose automation entirely under `chrome`

`cdp.info` only discloses `cdpPort` behind `canDiscloseBrowserAttachInfo`, which ends
at `isFirstPartyClient(ctx.clientName)`. A client whose name isn't in
`FIRST_PARTY_CLIENT_NAMES` therefore can't `connectOverCDP`, every tool falls to the
RPC path, and the chrome branch rejects all of them. I hit this myself: with my
harness announcing an arbitrary client name, `browser_open` and `browser_tabs`
worked and **snapshot / evaluate / screenshot all failed**; renaming the client to
`claude-code` made them pass. On `builtin` that same client works fine through the
RPC fallback, so this is a capability cliff that only appears when the user switches
backend.

The message also points the wrong way — *"If page resolution failed, open a page
first with browser_open"* — when the page is already open and the real cause is that
the caller was never given the port.

### Suggested shape

P1/P2 all share one root: three tools (`navigate`, `navigate_back`, `screenshot`)
are hardcoded to prefer or require RPC, and four RPC methods (`goBack`,
`screenshot`, `lifecycle.get`, plus `navigate`'s surface-targeting) have no chrome
path. Either give those methods chrome fallbacks that act on the registry target, or
let the three tools take the Playwright path when a page resolves. The navigate one
is worth fixing before merge — silently reading the old page is the kind of failure
an agent can't detect.

openwong2kim:
## Follow-up: the backend description is English-only

Separate from the dogfood findings above, and much smaller — but it ships in the
same UI change, so worth catching before merge.

`settings.browserBackendDesc` gained a sentence for the new option in `en.ts`:

> Chrome launches a dedicated real-Chrome instance with its own persistent profile
> (sign in once, logins stick) and gives agents the full automation toolset —
> separate from your daily browser.

`ko.ts`, `zh.ts` and `pl.ts` each got the five new keys
(`settings.browserBackendChrome` + the four `chromeProfiles.*` strings) but their
`browserBackendDesc` was left untouched, so it still describes only two backends.

Net effect for a non-English user: the dropdown offers three options — the label
itself is translated fine — while the description right under it explains Built-in
and External and says nothing about the third one. Nothing breaks and no key is
missing, so this won't trip the i18n guard; it's the copy that's out of sync.

Same for the picker's discoverability, incidentally: the per-workspace profile
binding lives in the workspace card's context menu, which the backend description
never points at. One clause there ("each workspace can bind its own profile from
its card menu") would connect the two surfaces this PR adds.

openwong2kim:
Dogfood findings addressed (thanks — the report was exactly right that all P1/P2s share one root):

- **P1 navigate**: `browser_navigate`/`browser_navigate_back` now resolve the workspace backend (one cached `cdp.info` read) and, under `chrome`, act on the resolved Playwright page (`goto`/`goBack`). The main-side chrome fallback additionally refuses a pinned `surfaceId` loudly instead of opening a new tab and reporting success. Builtin keeps its RPC lane byte-identical (contract test updated for the one extra `cdp.info` call).
- **P1 lifecycle**: the engine now mirrors main-frame navigations/closes per scope for chrome pages and the lease drain merges that queue with the RPC drain — `[browser events since last tool call]` is back under `chrome`.
- **P2 screenshot**: whole-page shots go over the Playwright page under `chrome` (element `ref` path unchanged).
- **P2 first-party cliff**: kept as a deliberate security boundary for now, but the contract error names it explicitly so a non-first-party client sees the real cause instead of "open a page first".

Also landed since your run: zombie-instance adoption (a crashed wmux's Chrome is reused via its DevToolsActivePort instead of dying on the SingletonLock — reproduced live on macOS) and the mechanism-credit comment. A re-run of your navigate sequence (steps 4–7) under `chrome` should now show the pinned tab's URL moving and no tab leak.

openwong2kim:
## Second live dogfood — all five findings fixed, two new P1s in the added features

Same setup as before: dev build of this branch on Windows, driven through the dev
build's own MCP server, `builtin` ↔ `chrome` flipped at runtime for controls.

### Every finding from the first round is fixed

| finding | verdict |
|---|---|
| P1 `browser_navigate` opened a new tab and reported success | **FIXED** — tabs 1 → 1, and the same surface reports `…/page2` afterwards |
| P1 lifecycle events silent under chrome | **FIXED** — `[browser events since last tool call] - navigated: …` now arrives |
| P2 full-page screenshot impossible | **FIXED** — 115,824-byte PNG |
| P2 `browser_navigate_back` | **FIXED** — `Went back. Current URL: …` |
| P2 non-first-party message pointed the wrong way | **FIXED** — now names the attach-disclosure cliff |

Chrome smoke is **24/24** (was 22/24). Builtin smoke shows no regression: its two
failures are the pre-existing hidden-workspace `#529` screenshot refusal and a
`ref=0` miss that I traced to my own smoke ordering (`browser_smart_snapshot`
re-tags refs 1-based between the snapshot and the click — main behaves the same).

Window labelling also checks out: a fresh profile came up as `wmux · labeltest` in
Local State, and an existing profile carrying Chrome's own name was left alone.

### P1 (new) — the SingletonLock adoption can never fire

`adoptExisting()` reads `DevToolsActivePort` from the profile dir. Chrome only
writes that file when remote debugging is started with **port 0**; with a fixed
`--remote-debugging-port` it writes nothing. Measured on Chrome 151.0.7922.175:

```
--remote-debugging-port=0     -> DevToolsActivePort: 50982 | /devtools/browser/5517…
--remote-debugging-port=19010 -> no such file
```

And wmux passes a fixed port — here is a live wmux-spawned Chrome, still running:

```
chrome.exe --user-data-dir=…\wmux-dev\chrome-profiles\labeltest
           --remote-debugging-port=18905 --remote-allow-origins=* …
    -> …\chrome-profiles\labeltest\DevToolsActivePort : No such file or directory
```

So the file the recovery path depends on is never produced by the instance it is
meant to recover. Reproduced end to end: with a Chrome already holding
`chrome-agent-profile`, `browser_open` fails with **`ChromeLauncher: Chrome exited
during startup`** — the exact SingletonLock symptom the commit set out to fix.

Fix shape: launch with `--remote-debugging-port=0` and read the port back out of
`DevToolsActivePort` (that also makes adoption work for free and retires the port
probing), or have wmux write its own `(port, pid)` sidecar next to the profile and
adopt from that.

### P1 (new) — live-Chrome attach can open tabs but cannot drive them

Verified against an **isolated stand-in** Chrome (own `user-data-dir`, started with
`--remote-debugging-port=0` so it produces the same port + secret-WS-path file the
M144 flow does), pointed at via `WMUX_LIVE_CHROME_DIR`. I deliberately did not
touch a real daily browser or click through Chrome's permission dialog — that is a
human consent gate.

Attach itself works: binding the workspace to `live` and calling `browser_open` put
a real tab in the stand-in, and `browser_tabs` lists its tabs including the ones it
already had.

Everything else fails:

```
 ok  | browser_tabs        | {...}
FAIL | browser_snapshot    | CHROME_BACKEND_RPC_UNSUPPORTED
FAIL | browser_click       | CHROME_BACKEND_RPC_UNSUPPORTED
FAIL | browser_screenshot  | CHROME_BACKEND_RPC_UNSUPPORTED
FAIL | browser_navigate    | browser_navigate: no chrome page resolved for this scope.
```

The engine log shows the connection is fine and the targets are visible — it is the
page match that fails:

```
[PlaywrightEngine] Connected to CDP at ws endpoint
[PlaywrightEngine] CDP targets: page:about:blank, …, page:http://127.0.0.1:18321/page2, …
[PlaywrightEngine] No webview target found in Target.getTargets
[PlaywrightEngine] No page found, reconnecting... (1/3)
```

Same result with a pinned `surfaceId` from `browser_open`, with a `surfaceId` taken
straight out of `browser_tabs`, and with no `surfaceId` at all. My read is that this
is the "live client seeds no targets" decision meeting the engine's target-matching,
which has no candidate list to match against on this path.

Worth weighing because the changelog already promises the user-facing version of
this: *"let its agent work inside your own running browser — your real tabs and
logins"*. Right now it can open a tab there and enumerate tabs, and that is all.

### Still open from the first round

`settings.browserBackendDesc` remains English-only — `ko` / `zh` / `pl` still
describe two backends while the dropdown offers three.

### Not verified

- The genuine M144 path (a real Chrome with `chrome://inspect/#remote-debugging`
  enabled and its per-connection permission dialog) — human consent gate, left alone.
- `dispose()` on the normal quit path; the renderer exposes no quit API, so I only
  confirmed a force-kill takes the dedicated Chromes down with it (19 → 0).

PR #1065, opened 2026-08-28
fix(browser): chrome-backend port discovery via DevToolsActivePort + live-tab driving + backend-desc i18n

openwong2kim:
## What

Three follow-ups from the post-merge dogfood on #1064 (see the round-2 comment there for the measurements).

**Adoption could never fire (P1).** Chrome writes `DevToolsActivePort` only when launched with `--remote-debugging-port=0` — a fixed port produces no file at all (measured on Chrome 151). ChromeLauncher passed a fixed port, so the file `adoptExisting()` reads was never written by the instance it was meant to recover, and a crashed session's Chrome still bricked the backend on the SingletonLock (`Chrome exited during startup`, reproduced live). The launcher now launches with port 0 and resolves the port from the file's fresh write — freshness by content, since the secret `/devtools/browser/<uuid>` path is new every boot. That also retires the port-probe allocator and the per-profile port slices (each instance picks its own ephemeral port; nothing to slice). An env-pinned launch (`WMUX_CHROME_CDP_PORT`) keeps its fixed port, and adoption probes the pin directly since pinned instances write no file.

**Live attach could open tabs but not drive them (P1).** The live client seeded zero `cdp.info` targets, so the engine's registry match had no candidates and every page tool failed after a successful `browser_open` — and the no-surface path auto-opened a fresh tab in the user's browser on each retry. Two halves:

- `LiveChromeClient` now seeds only the tabs **wmux opened** (owner-tracked, dead-pruned, same semantics as `ChromeLauncher`) — a random user tab still never becomes the default pin, and a bare `cdp.info` with nothing opened never dials the user's Chrome.
- The engine matches a pinned `surfaceId` against Chrome's own target list **when the response carries a `wsEndpoint`** — under live attach the `browser_tabs` full-exposure list is the grant (consent model), so a surfaceId taken from it may name a pre-existing tab the wmux-opened registry cannot know. Deliberately not extended to port-based (dedicated) instances, where the registry match is what keeps workspace B off workspace A's tab on a shared profile.

**i18n.** The browser-backend description gained its Chrome sentence in English only; ko/zh/pl still described two backends under a three-option dropdown. All four locales now describe the third option and point at the workspace-card menu for per-workspace profiles.

## Verified live (Windows)

- Zombie adoption: pre-existing Chrome over the agent profile dir → `browser_open` adopts it (the tab lands in the zombie instance, 4.1s), then navigate / snapshot / full-page screenshot / back all pass on the adopted instance.
- After killing the adopted instance: next open relaunches via port 0, resolves the fresh `DevToolsActivePort` (57606 ≠ dead 56966), and inline lifecycle events still arrive.
- Live attach against an isolated port-0 stand-in: snapshot / click / screenshot / navigate on the wmux-opened tab, and evaluate / snapshot on a **pre-existing** tab pinned from `browser_tabs`. (The real M144 consent dialog remains untested by design — human gate.)

## Tests

ChromeLauncher suite reworked for port-0 discovery (file-based resolution, stale-file freshness, env-pin launch + pin adoption, registry without slices); LiveChromeClient owner-seeding, no-dial-before-open, workspace filter + dead-prune; two engine tests for the live direct match (hit and strict miss). Suites: browser-session + playwright 315/315, browser rpc 129/129, i18n 147/147. `tsc` (root + mcp) clean; `probe:mcp` unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Chrome sessions can now recover and reuse instances left running after a crash.
  * Agents can interact with both opened and pre-existing tabs in live Chrome sessions.
  * Tab targeting is more reliable, with stale tabs automatically removed.

* **Documentation**
  * Expanded browser backend setting descriptions, including Korean, Chinese, and Polish translations.
  * Clarified support for separate Chrome profiles and independent workspace logins.


openwong2kim:

Fixes the two #1064 dogfood P1s: zombie-Chrome adoption via port-0 launches resolved from DevToolsActivePort, and live-attach tab driving via owner-seeded targets plus the engine's wsEndpoint-gated direct match; backend-desc i18n rides along. Main risk is the PlaywrightEngine match change, correctly gated off dedicated instances to preserve cross-workspace isolation. P1 — it restores crash recovery and working page tools to the chrome backend.

PR #1067, opened 2026-08-28
feat(remote): bootstrap a new workspace on a paired remote host (#1001)

p-poppe:
Closes the concrete gap flagged in #1001: pairing a headless remote daemon worked end-to-end, but there was no way to create the FIRST pane of a NEW workspace on it — only to add a sibling pane to a workspace the desktop app had already created. "New workspace on this host" in the Attach Remote modal closes it.

## Server contract (quoted, not guessed)

From the maintainer's decision in #1001:

> Server contract I'd take: keep the shape check for everyone; relax only the existence check when `principal.kind === 'operator'`. Mint the uuid on the desktop side — the renderer owns the registry today, so the server accepts an operator-supplied new id rather than generating one. No separate `POST /api/workspaces` mint endpoint: if it records anything it invents the registry the daemon deliberately doesn't have, and if it doesn't, it is the same relaxation with more API surface.
>
> Stated limitation, so nobody later mistakes A for more than it is: this is an audit-and-revocability boundary, not a hard one — a shell on the daemon's own host can read the operator token, and a phone-only headless bootstrap still needs the operator credential once. Both deserve a line in the `rejectWorkspaceId` comment when this lands.

Implemented exactly that — both halves in one PR, as offered.

## Server-side gate

`WebTerminalServer.rejectWorkspaceId` now takes the caller's `WebPrincipal`. Shape (`^[A-Za-z0-9_-]{1,64}$`) is still enforced unconditionally; the existence check (some live pane already carries the id) is skipped only when `principal.kind === 'operator'`. A paired device's behavior is byte-for-byte unchanged — see the regression test below. The rationale from #1001 (identity operation, not an execution one; the thing that owns the registry may extend it) plus the stated audit-and-revocability limitation are now in the comment.

Credential path, confirmed rather than assumed: `RemoteHost.token` (what `RemoteHostClient` sends as `Authorization: Bearer`) is always the operator token — parsed from the pasted `wmux web` URL or minted by the code-pairing exchange — never a paired device's credential. That was established in #1001's own discussion thread and is what makes every call through this client already qualify.

## Desktop half

- `RemoteHostClient.createWorkspace(workspaceId, cwd?)` — `POST /api/sessions` with the operator token and a desktop-minted id, returns the new pane's `sessionId`.
- `remote:workspace:create` IPC route (main) + `window.electronAPI.remote.workspaceCreate` (preload/types), wired the same way `remote:workspaces:list` already is.
- `AttachRemoteModal`: a "New workspace on this host" button, shown whenever a host is selected — including the empty-workspaces state the #1001 VPS test actually hit ("no workspaces on that host"). Success mints a UUID, creates it on the host, and calls `attachRemoteWorkspace` exactly like picking an existing workspace does.

## Test plan

- `WebTerminalServer.test.ts`:
  - operator bootstraps a brand-new `workspaceId` → 201, reaches `lifecycle.create`
  - operator's exception still enforces shape → 400 `invalid-workspace-id`
  - a **paired device** on the same unknown id → still 400 `unknown-workspace-id`, unchanged (regression guard — this is the test that used to run against the operator token; split so the device case keeps its original assertion)
- `RemoteHostClient.test.ts`: request method/URL/headers/redirect/timeout, cwd included only when given, daemon-supplied error detail surfaced, non-JSON error body falls back to a generic message, missing-session-id response rejects.
- `remote.handler.test.ts`: success + sessionId passthrough, cwd forwarding, client rejection mapped to `{ok:false}` (never throws across IPC), unknown host reported without touching a client.
- `AttachRemoteModal.test.tsx`: mint → create → attach end to end (checks the store gets the right `AttachedRemoteWorkspace`), button offered on an empty workspace list, failure path shows the error and attaches nothing.
- `plLocaleCoverage.test.ts`: the two new `remote.*` strings are translated (Polish is the only locale with enforced coverage; the other 20 are an accepted gap per #997).
- `npm run test:parallel`: 811 passed / 7 failed test files — every failure is in `worktree.handler.test.ts` (EBUSY / timeout, git-worktree fixture cleanup racing under load) and `channelsSlice.test.ts` (unrelated timeout), none touching a file this PR changes; confirmed by `git diff --name-only` against the failing files.
- `tsc --noEmit` and `eslint` on every changed file: clean (pre-existing warnings/errors elsewhere untouched).

House rules followed: English, no version bump, changelog fragment to follow once the PR number is known.

## Summary by CodeRabbit

* **New Features**
  * Create a new workspace directly on a paired remote host.
  * “New workspace on this host” is available even when no workspaces exist.
  * Optionally select the workspace’s initial working directory.
  * Clear error messages are displayed when creation fails.

* **Bug Fixes**
  * Operators can now bootstrap new remote workspaces, while device access continues to require an active workspace.

### DIFF
--- a/changelog.d/1067.md
+++ b/changelog.d/1067.md
@@ -1,0 +1,3 @@
+### Added
+
+- Remote workspace attach: "New workspace on this host" in the Attach Remote modal bootstraps the first pane of a brand-new workspace on a paired remote daemon — previously only a sibling pane on a workspace the desktop app had already created could be added.
--- a/src/daemon/web/WebTerminalServer.ts
+++ b/src/daemon/web/WebTerminalServer.ts
@@ -2336,7 +2336,7 @@ export class WebTerminalServer {
       const workspaceId = typeof b.workspaceId === 'string' ? b.workspaceId.trim() : '';
       const cwd = typeof b.cwd === 'string' ? b.cwd.trim() : '';
       if (workspaceId) {
-        const bad = this.rejectWorkspaceId(workspaceId);
+        const bad = this.rejectWorkspaceId(workspaceId, principal);
         if (bad) return this.json(res, 400, bad);
       }
       lifecycle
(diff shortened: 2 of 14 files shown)


openwong2kim:

Implements the agreed #1001 contract: an operator-authenticated `createWorkspace` path plus a "New workspace on this host" button in the Attach Remote modal. The risk point is `rejectWorkspaceId` now skipping the existence check for operators; shape stays enforced and paired devices keep the old 400. P2 — it touches the daemon's auth boundary, so prompt, careful review is warranted.

openwong2kim:
Merged — thanks. One follow-up question rather than a blocker.

The "New workspace on this host" button stays enabled when the selected host has \`allowInput === false\`, but \`POST /api/sessions\` requires input permission, so pressing it on a read-only host can only fail after the round trip. The workspace list right below already tags that host as read-only, so the button is the one control that does not reflect the permission.

Was that deliberate — e.g. \`allowInput\` being stale until a probe runs, so you preferred letting the server be the authority — or just an oversight? If it is the latter, the fix is one line:

```
disabled={creatingWorkspace || selectedHost?.allowInput === false}
```

Happy to fold it into the next change that touches this modal if you would rather not open a PR for it.

Issue #353, opened 2026-07-07
[Bug] Heavy SPAs (Gemini): browser_snapshot exposes no refs on background surfaces + browser_press_key never reaches the webview


vanlong2109:
## Summary

Driving the MCP browser tools (`mcp__wmux__browser_*`), two related problems make it impossible to automate heavy SPAs like Google Gemini via the documented ref-based flow — even though the same tools work fine on simple pages:

1. **`browser_snapshot` returns no interactive refs for heavy SPAs.** On Gemini it returns only `- RootWebArea "Google Gemini"` (no `[ref=N]`), so `browser_click`/`browser_type` can't target the composer or the Send button. `browser_smart_snapshot` returns page text but no indexed elements, and `format:"aria"` is empty too. A simple page (example.com) returns refs normally **in the same session**.
2. **`browser_press_key` doesn't reach the browser webview at all.** An in-page `keydown` listener captures nothing after `browser_press_key`, on both google.com and Gemini — so the keyboard-driven fallback (type / Enter-to-send) also fails.

Net effect: you can `browser_open` / `browser_navigate` / `browser_evaluate` / `browser_extract_text` on Gemini, but you cannot type-and-send a prompt through the ref-based flow.

## Environment
- **wmux 3.17.0**, Windows 11
- Driver: Claude Code via the wmux MCP server (`mcp__wmux__browser_*`)

## Repro
1. `browser_open {url:"https://gemini.google.com/app"}` (logged-in profile)
2. `browser_wait {selector:"div.ql-editor"}` → succeeds (DOM is present)
3. `browser_snapshot` → returns only `- RootWebArea "Google Gemini"`. `browser_smart_snapshot` → page text, no indexed elements. `browser_snapshot {format:"aria"}` → empty too. Still empty after full render (toolbar buttons present) and after focusing the composer.
4. **Contrast:** `browser_open {url:"https://example.com"}` → `browser_snapshot` → `[ref=0] a "Learn more"` (refs work).
5. **Keyboard check:** via `browser_evaluate`, `window.__k=[]; document.addEventListener('keydown', e=>window.__k.push(e.key), true)`, focus an `<input>` (confirm `document.activeElement` + `document.hasFocus()`), then `browser_press_key {key:"z"}` → `window.__k` stays `[]` and the input value is unchanged (on google.com, not just Gemini).

## Expected
[...]

PR #1068, opened 2026-08-28
fix(browser): honor filter on browser_snapshot's DOM paths instead of silently dropping it

### LINKED ISSUE #1066: browser_snapshot silently drops format and filter on the DOM fallthrough, so filter:"interactive" is a no-op
`browser_snapshot` advertises two output-shaping params:

> `selector` scopes to a subtree, `filter:"interactive"` strips non-interactive nodes

and `format:"aria"` — *"returns the full tree"* vs the default `ai`.

On a live Chrome-backend session both `filter` and `format` are **no-ops**, and unlike the `selector` path there is no note saying so. The caller gets a full, unfiltered tree while believing it asked for a reduced one.

## Live evidence (v3.48.1, `backend: chrome`)

**`filter:"interactive"` alone, on a rich page** — `https://www.wikipedia.org/`, `full:true`, once without `filter` and once with. Output was **byte-identical**: the non-interactive `H1: Wikipedia` / `H2: 1,000,000+` / `H3: We owe you an explanation.` lines survived the filter in both.

**`format:"aria"`** — on `https://example.org/`:

```
[snapshot: full]
Page: Example Domain
URL: https://example.org/

H1: Example Domain

Interactive elements (use ref number for click/fill/type):
  [ref=0] a "Learn more" -> https://iana.org/domains/example
```

That is the `ai`-style ref listing, not an aria tree. `format` was not honored either.

**Control — `selector` genuinely works**, so this is not a general breakage:

| call | interactive elements returned |
|---|---|
| `selector:"div"` | 1 (`a "Learn more"`) |
| `selector:"h1"` | 0 |

**The honest path, for contrast** — with `selector` present the tool *does* say what it dropped:

```
(note: format/filter are ignored when selector is given)
```

Used alone, `filter` and `format` are dropped with no such note.

## Root cause

`src/mcp/playwright/snapshot.ts:343` — the DOM fallthrough returns **before** either param is ever consulted:

```ts
if (!tree || isRootOnly(tree)) {
  const domSnapshot = (await page.evaluate(buildDomSnapshotExpression())) as string;
  pageRefMaps.set(page, []);
  return domSnapshot;          // <-- early return, line 349
}

// never reached on this path:
if (options?.filter === 'interactive' && format === 'ai') { ... }   // line 365
let output = serializeTree(effectiveTree, format, depth, refs);     // line 375
```

`buildDomSnapshotExpression()` takes no `format`/`filter` argument, so both are structurally unreachable once the a11y tree is null or root-only. Per `getAccessibilityTree` (`snapshot.ts:274`) that happens whenever there is no CDP session, `Accessibility.getFullAXTree` throws, or the tree stays root-only after the 250 ms retry — i.e. background surfaces and any page whose a11y tree collapses.

The same drop exists on the second DOM path, `src/mcp/playwright/tools/inspection.ts` (the `else` RPC fallback when no Playwright page resolves): it also calls `buildDomSnapshotExpression()` and never forwards `format`/`filter`.

I could not tell from the MCP surface which of the two DOM paths my calls took — which is part of the problem, since neither emits any signal.

Note the fallthrough itself looks deliberate (the comment at `snapshot.ts:337-342` explains why a root-only aria tree should degrade to the interactive listing, ref #353). The bug is not the fallthrough — it is that the params are dropped **silently** on it, while the structurally identical `selector` DOM listing at `inspection.ts:352` is honest about exactly the same limitation.

## Why this is worse than a hard failure

`filter:"interactive"` exists to bound context. A silent no-op means an agent asks for the small tree, is handed the big one, and has no way to notice. Compare the deliberate care already taken one branch below:

```ts
// Zero interactive nodes must NOT fall back to the full tree (review consensus:
// the filter would silently invert into maximum output) — say so instead.
```

That exact failure mode is what the DOM fallthrough reintroduces, just earlier in the function.

## Suggested fix

1. Minimum: emit the existing note on the DOM fallthrough too, so `filter`/`format` are never dropped without a word — in `snapshot.ts` before the early return, and on the `inspection.ts` RPC fallback.
2. Better: teach `buildDomSnapshotExpression()` an interactive-only mode and honor `filter` on the DOM path, where the listing is already interactive-centric and the strip is cheap.

There is existing coverage in `src/mcp/playwright/__tests__/snapshot.filter.test.ts` and `snapshot.fallthrough.test.ts`; the gap is a case that exercises **filter together with** the fallthrough, which is where it dies.

## Also seen in the same session

Two smaller observations from the same run, filed here only so they are not lost — say the word and I will split them out:

- `browser_navigate` returns just `Navigated to <url>`; the lifecycle event arrives one call late, prepended to the *next* tool result as `[browser events since last tool call] - navigated: <url> (8s ago)`. Reproduced 2/2. No `load` event observed, only `navigated`.
- `browser_tabs list` reports `"selected": false` for the only surface in the workspace, the one every other browser tool acts on.

openwong2kim:
Fixes #1066.

## What was broken

`browser_snapshot`'s `filter:"interactive"` and `format:"aria"` were silent no-ops on every DOM-listing path:

- **`snapshot.ts:343` root-only fallthrough** — early `return domSnapshot` fired before the filter block (365) or `serializeTree(…, format, …)` (375) were reached, so both params were structurally unreachable whenever the a11y tree was null or root-only.
- **`inspection.ts` RPC fallback** (no Playwright page) — same drop: `buildDomSnapshotExpression()` was called with neither param.
- The `selector` path was honest (`(note: format/filter are ignored…)`) but didn't honor `filter` either.

Live repro on v3.48.1, backend=chrome: wikipedia.org snapshot with and without `filter:"interactive"` was **byte-identical**; `format:"aria"` returned the ai listing unlabeled.

## The fix

1. `buildDomSnapshotExpression(rootSelector?, opts?)` grows `opts.filter:"interactive"` — drops the h1–h3 heading block from the listing (the DOM-path equivalent of `stripNonInteractive`). The `Page:`/`URL:` header stays: the auto-diff URL guard parses the `URL: …` line out of this text.
2. All three DOM call sites forward `filter`: the fallthrough, the selector path, and the RPC fallback.
3. `format:"aria"` remains unfulfillable on a DOM listing, so those paths now **say so** with a note instead of returning the ai listing unlabeled — the same honesty rule the selector path already applied. The selector-path note narrows to aria only, since `filter` is now honored and `ai` is the listing's native shape.

The #353 fallthrough design (root-only aria tree degrades to the interactive listing) is untouched — only the silent drop on top of it is fixed.

## Tests

The existing suites covered `filter` (`snapshot.filter.test.ts`) and the fallthrough (`snapshot.fallthrough.test.ts`) separately but never crossed them — which is exactly where this died. Added:

- fallthrough × `filter`: the param reaches the evaluated expression (`const interactiveOnly = true`), plus a no-filter control
- fallthrough × `aria`: note asserted verbatim (was `toBe('DOM-FALLBACK')`)
- builder-level: filter drops headings, keeps interactives + `URL:` line + ref tagging; composes with `rootSelector`; no-filter control

`vitest` 37/37 on the six touched/adjacent suites, `tsc --noEmit` clean, eslint 0 errors (7 pre-existing `no-explicit-any` warnings).

Not live-verified against a running build — the targeted fallthrough path is unit-covered above; happy to dogfood on the next canary if wanted.

## Summary by CodeRabbit

* **New Features**
  * Added an interactive filter for DOM snapshots, showing interactive elements while retaining page and URL details.
  * Interactive filtering now works with selector-scoped snapshots and fallback snapshot paths.
  * Added clear notices when aria-format snapshots are unavailable and a DOM listing is returned.

* **Bug Fixes**
  * Preserved heading content when no interactive filter is selected.

### DIFF
--- a/src/mcp/playwright/__tests__/dom-intelligence.selector.test.ts
+++ b/src/mcp/playwright/__tests__/dom-intelligence.selector.test.ts
@@ -44,3 +44,38 @@ describe('buildDomSnapshotExpression(rootSelector)', () => {
     expect(out).toBe('No element matches selector: #nope');
   });
 });
+
+describe('buildDomSnapshotExpression filter:"interactive" (#1066)', () => {
+  it('drops the heading block but keeps interactives and the Page/URL header', () => {
+    document.body.innerHTML = '<h1>Title</h1><h2>Section</h2><button id="b1">Go</button>';
+
+    const out = run(buildDomSnapshotExpression(undefined, { filter: 'interactive' }));
+
+    expect(out).not.toContain('H1: Title');
+    expect(out).not.toContain('H2: Section');
+    expect(out).toContain('"Go"');
+    // The auto-diff URL guard parses the "URL: …" line — it must survive the filter.
+    expect(out).toMatch(/^URL: /m);
+    expect(document.getElementById('b1')?.getAttribute('data-wmux-ref')).toBe('0');
+  });
+
+  it('composes with a rootSelector: scoped AND heading-free', () => {
+    document.body.innerHTML =
+      '<main><h2>Inside</h2><button id="b1">In</button></main>' +
+      '<h2>Outside</h2><button id="b2">Out</button>';
+
+    const out = run(buildDomSnapshotExpression('main', { filter: 'interactive' }));
+
+    expect(out).toContain('Scope: main');
+    expect(out).not.toContain('H2: Inside');
+    expect(out).not.toContain('H2: Outside');
+    expect(out).toContain('"In"');
+    expect(out).not.toContain('"Out"');
+  });
+
+  it('without the filter, headings remain (control)', () => {
+    document.body.innerHTML = '<h1>Title</h1><button>Go</button>';
+    const out = run(buildDomSnapshotExpression());
+    expect(out).toContain('H1: Title');
+  });
+});
--- a/src/mcp/playwright/__tests__/snapshot.fallthrough.test.ts
+++ b/src/mcp/playwright/__tests__/snapshot.fallthrough.test.ts
@@ -72,11 +72,34 @@ describe('generateSnapshot — root-only fallthrough (#353)', () => {
     expect(out).toBe('DOM-FALLBACK');
   });
 
-  it('falls through for aria format too, not just ai', async () => {
+  it('falls through for aria format too, with a note instead of a silent drop (#1066)', async () => {
     const { page } = makePage({ nodes: ROOT_ONLY, evalResult: 'DOM-FALLBACK' });
     const out = await generateSnapshot(page as never, { format: 'aria' });
     expect(page.evaluate).toHaveBeenCalledTimes(1);
+    // aria has no DOM-listing equivalent — the caller must be told, not
+    // handed the ai-style listing as if it were the aria tree.
+    expect(out).toBe(
+      '(note: aria format unavailable — the a11y tree collapsed, returning the DOM interactive listing)\nDOM-FALLBACK',
+    );
+  });
+
+  it('forwards filter:"interactive" into the DOM expression on fallthrough (#1066)', async () => {
+    const { page } = makePage({ nodes: ROOT_ONLY, evalResult: 'DOM-FALLBACK' });
+    const out = await generateSnapshot(page as never, { format: 'ai', filter: 'interactive' });
+    // The param must reach the expression instead of dying on the early return.
+    expect(page.evaluate).toHaveBeenCalledWith(
+      expect.stringContaining('const interactiveOnly = true'),
+    );
+    // ai + filter needs no note: the filtered listing IS what was asked for.
     expect(out).toBe('DOM-FALLBACK');
+  });
+
+  it('fallthrough without filter keeps the heading block enabled', async () => {
+    const { page } = makePage({ nodes: ROOT_ONLY, evalResult: 'DOM-FALLBACK' });
+    await generateSnapshot(page as never, { format: 'ai' });
+    expect(page.evaluate).toHaveBeenCalledWith(
+      expect.stringContaining('const interactiveOnly = false'),
+    );
   });
 
   it('serializes the a11y tree (no DOM fallback) when the tree has children', async () => {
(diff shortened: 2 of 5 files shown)


openwong2kim:

This PR fixes #1066 by honoring `filter:"interactive"` on all three `browser_snapshot` DOM-listing paths and noting when `format:"aria"` cannot be fulfilled, with crossed fallthrough×filter tests added. Risk is low — output-only change, and the auto-diff URL guard's `URL:` line is preserved. P2, matching the issue: agents get the reduced tree they asked for instead of a silent full one.

openwong2kim:
## Live dogfood on real Chrome — filter verified, with one honest caveat

Ran the **exact generated expression** from this branch inside the live Chrome page (backend=chrome, `https://www.wikipedia.org/`) via `browser_evaluate`, filter off vs on. This is the real DOM, not jsdom.

| probe | `len` | `headings` | `refs` | `hasUrlLine` |
|---|---|---|---|---|
| `filter` off (`interactiveOnly = false`) | 5222 | **7** | 100 | true |
| `filter` on (`interactiveOnly = true`) | 5084 | **0** | 100 | true |

- headings 7 → 0: the filter **works in a real browser**, not just jsdom.
- refs 100 → 100: no collateral loss of interactive elements.
- `hasUrlLine` stays true: the auto-diff URL guard still finds its `URL: …` line, which was the main regression risk of touching this output.

### Caveat worth recording

The saving is **138 bytes on 5222 — 2.6%**. On the DOM path the listing is already interactive-centric by construction, so headings are nearly all there is to strip. The real output weight is the 100-element ref cap, not the heading block.

So this PR makes `filter:"interactive"` *honest and correct* on the DOM paths — it no longer silently claims to do something it never did — but callers should not expect the DOM path's filter to buy much context. The a11y path's `stripNonInteractive` is where the param actually pays off. If we want real reduction on the DOM path, that's a separate change to the 100-element cap or the per-element `desc` verbosity, and I'd rather not smuggle it into a fix PR.

### Path confirmation

Also confirmed the production build really does take a DOM listing path, so this fix is on the executed code, not a theoretical branch: `serializeTree`'s ai format emits `ref="N"` (`snapshot.ts:195`), whereas the live v3.48.1 output was `[ref=0] a "…" -> //ko.wikipedia.org` — the DOM expression's format, verbatim.

Not covered by this run: the note strings on the aria paths and the `generateSnapshot`/`inspection.ts` plumbing above the expression — those stay unit-only until a dev build is stood up.

PR #1069, opened 2026-08-28
feat(browser): show the workspace Chrome-profile submenu only on the chrome backend

openwong2kim:
## What

The workspace right-click → **Chrome 프로필** submenu (Phase 2.5) rendered on every backend. But a binding is only ever consulted on the chrome path — `requireChrome` → `chromeRegistry.forWorkspace` in `browser.rpc.ts` — so under `builtin` or `external` the submenu offered a profile choice that changed nothing observable.

This gates it on `browserBackend === 'chrome'`.

## Why hide rather than disable

`WorkspaceAccountMenu`, the component this one was cloned from, already hides itself when there is nothing to bind ("Hides itself when no accounts are registered"). Following that idiom keeps the two bind-only submenus consistent instead of introducing a third state.

## Notes

- **No hydration wait needed.** `readInitialBrowserBackend` reads main's authoritative value synchronously at store-module load, so the mirror is correct before first render — no flash, no gate flipping after boot.
- **Also skips the boot IPC.** `reload()` fired `chromeProfiles.list()` on mount once per rendered workspace row. Off the chrome backend none of those rows would show the menu, so the call is now skipped rather than made and discarded.
- The older-preload guard (`if (!window.electronAPI?.browser?.chromeProfiles) return null`) is unchanged and still applies first.

## Tests

New `WorkspaceChromeProfileMenu.backendGate.test.tsx` — 6 cases: renders on chrome; renders nothing on builtin and on external; boot IPC skipped off chrome and fired on chrome; still hidden on an older preload even when the backend is chrome.

Verified the tests actually catch the regression: reverting the component alone fails exactly the 3 gate cases (builtin, external, skipped IPC) and passes the other 3.

`vitest` 136/136 across the Sidebar suites + `uiSlice`, `tsc --noEmit` clean, eslint clean on both touched files.

Not live-verified in a running build — the gate is pure renderer logic and unit-covered above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Restricted the Chrome profile submenu and reload actions to supported Chrome backends.
  * Prevented unsupported browser backends from triggering unnecessary profile startup actions.
  * Hid the submenu when Chrome profile functionality is unavailable.

* **Tests**
  * Added coverage for Chrome, built-in, and external backends, including missing profile support.

### DIFF
--- a/src/renderer/components/Sidebar/WorkspaceChromeProfileMenu.tsx
+++ b/src/renderer/components/Sidebar/WorkspaceChromeProfileMenu.tsx
@@ -1,6 +1,7 @@
 import { useState, useEffect, useCallback } from 'react';
 import { IconChevron } from '../icons';
 import { useT } from '../../hooks/useT';
+import { useStore } from '../../stores';
 
 /**
  * Workspace right-click → Chrome profile submenu (Phase 2.5, BIND ONLY).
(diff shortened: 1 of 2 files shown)


openwong2kim:
This PR gates the workspace Chrome-profile submenu on the chrome backend, since bindings are only consulted on the chrome path. It touches only WorkspaceChromeProfileMenu renderer logic plus tests, hiding an inert menu and skipping its boot IPC — low risk. P3: small UX-consistency polish with no functional behavior change.

PR #1070, opened 2026-08-28
fix(browser): attribute inline lifecycle events to the tool call that caused them

openwong2kim:
Root-cause fix for the #1063 "inline lifecycle events" landing one call late (live repro 2/2: `browser_navigate`'s own `navigated` event arrived prepended to the **next** tool result). Release-gating: #1063 ships for the first time in the next release, so the contract is corrected before it ships.

## Two root causes

1. **`browser_navigate`/`browser_navigate_back` were never wrapped in `withAutomationLease`** — the 30 other browser tools all are. The tools most certain to cause a navigation structurally could not report one.
2. **The drain ran before the tool body** (`automationLease.ts`, both the normal and late-acquire branches), so events the body itself caused were always deferred to the next call.

Note this was *documented* behavior ("since the last tool call" in the code comment and changelog fragment) — so this PR is a contract change (Changed), not a bugfix, and updates the wording everywhere (`[browser events]` header, comments, fragment).

## Why dual drain instead of moving the drain after the body

An after-only move has two verified regressions:
- `invalidateSnapshotBaseline` ordering inversion — a `browser_snapshot` writes its baseline, then the post-drain nukes it (diff-cache self-destruction).
- Loss of the lazy `Page.enable` warm-up — on builtin, the pre-drain's `browser.lifecycle.get` is what turns capture on in the first place.

So: **pre-drain stays byte-identical** (warm-up + between-calls events + unconditional invalidation), and a **post-drain** (still inside the lease bracket — `browser.lifecycle.get` is a leased RPC; skipped when the body throws, leaving events for the next op) picks up the body's own events. Destructive drains + mutually exclusive sources (main cannot see chrome tabs) make double-reporting impossible — asserted in tests.

The post-drain invalidates the baseline **conditionally** via new `invalidateSnapshotBaselineIfStale`: a baseline the body just wrote for the navigation's final URL survives (snapshot-during-navigation); mismatch or unknown URL drops it — also closing the hole where a URL-less baseline survives forever because the read-side guard is fail-open.

## Self-echo

A wrapped `browser_navigate` would report the navigation its own result text already states. Suppression is deliberately narrow: only the **last** merged event, only `navigated`, only on **exact** URL match — so a redirect chain (`navigated: A` then `B`) keeps A visible. Builtin reports the requested URL, so a server redirect suppresses nothing (conservative). The builtin lane also gets a 150ms settle after the commit-resolved RPC (`navigateAwaitingCommit`, #756) so the post-drain catches its own events; chrome needs none (in-process mirror). A missed late event (`loaded` fires after commit) is a delay — next op's pre-drain — never a loss.

## Tests (18 files, 192/192 green)

- `automationLease.events.test.ts`: mock router rebuilt as a **closure queue** modeling the ring's destructive drain (the old fixed-array router was insensitive to drain ordering — which is exactly how this shipped). New: during-body attribution, pre→post merge order, self-echo suppression, redirect-chain preservation, conditional baseline preserve/invalidate, throw-skips-post-drain.
- `automationLease.test.ts`: RPC order pinned to `acquire → pre-drain → BODY → post-drain → release`.
- `navigation.tabs.test.ts`: leased-router helper; exact-order goBack assertion updated; new end-to-end cases (redirect hop visible + final echo suppressed; plain navigation carries no block).
- `snapshotCache.test.ts`: `invalidateSnapshotBaselineIfStale` — match preserves / mismatch drops / unknown-URL drops (both directions).

**Regression-catch verified**: reverting `automationLease.ts` alone fails exactly 7 of the new/updated cases.

`tsc --noEmit` clean; eslint clean on changed files (3 pre-existing `no-empty-function` errors in untouched lines of `automationLease.ts` exist on origin/main too and are left alone).

## Not covered

Live dogfood of this branch comes next (dog1066 rig) before release. `browser_tabs` stays unwrapped (Phase 2: list is high-frequency read-only; close's `closed` is already recovered by the `lastLifecycleTarget` target-tolerant path). Cursor-based non-destructive drain (fixes pre-drain loss on body throw + multi-consumer event theft) is also Phase 2.

## Summary by CodeRabbit

- **New Features**
  - Browser tool results now include lifecycle events caused by the current action, including navigations, page loads, and tab closures.
  - Navigation results report the final landed URL, including redirects and single-page app route changes.
  - Navigation events are shown immediately on the tool result that triggered them.

- **Bug Fixes**
  - Removed redundant navigation notifications when the result already reports the same URL.
  - Snapshot information is refreshed when navigation makes the previous page state outdated.


openwong2kim:
## Live dogfood (dog1066 rig, builtin backend, new MCP bundle from this branch)

| scenario | result |
|---|---|
| T1 `browser_navigate` wikipedia (plain) | **PASS** — navigate now runs leased; self-echo suppressed (no `[browser events]` block, correct) |
| T2 snapshot right after — **the original symptom** | **PASS** — no stale `navigated` on the next call any more. It carried only `- loaded (6s ago)`: the documented delayed-`loaded` fallback (`loaded` fires seconds after the commit the navigate RPC resolves on) |
| T3 click-induced navigation attributed to the click result | **delayed, not lost** — on builtin the click dispatch returns before the commit, so `navigated`+`loaded` landed on the next call's pre-drain (page verified moved to iana.org). This is the fallback the design accepts; on the chrome backend the in-process mirror makes post-drain attribution structural (no race), unit-covered by the during-body case |
| T4 snapshot ×2 diff (post-drain must not self-destruct the baseline) | **PASS** — second snapshot still returns `(no changes since previous snapshot)` |

Net: the headline bug (events one call late after `browser_navigate`) is gone end-to-end; the remaining builtin click race degrades to next-call delivery exactly as documented. Settle tuning for interaction tools stays Phase 2.

openwong2kim:
CodeRabbit findings ①②③ all applied in the follow-up commit: suppression scoped to the post-drain slice (new test: a delayed same-URL pre-drain navigated stays visible), builtin navigate now reads `location.href` after the settle and reports the landed URL (redirect echo now suppressible; fallback to requested URL on read failure), fragment renamed to `changelog.d/1070.md`. The 2 failed validate tests (`phase-2-2-dynamic`, `killVerifiedDaemonPid.crossHost`) are unrelated to this diff and pass locally 13/13 — CI flake.

openwong2kim:

This fixes #1063's one-call-late lifecycle event attribution via a post-body drain in `withAutomationLease`, plus wrapping `browser_navigate`/`browser_navigate_back` in the lease. It changes the lease bracket all browser tools share and snapshot-baseline invalidation, so RPC-ordering and redirect-chain cases are the risk. P2 — a contract change gating #1063's first release, well covered by tests here.

PR #1073, opened 2026-08-29
fix(browser): replace the dead window.prompt in the Chrome-profile menu with an inline form

openwong2kim:
## Problem

Electron's renderer implements no `window.prompt` — the call throws — so the workspace menu's "New profile…" row silently did nothing and no Chrome profile could ever be created (dogfood 2026-08-29). This repo already knew the pattern: `DiffPanel.tsx` documents the same measurement and `BroadcastPopover` replaced its own dead prompt.

## Fix

- The row swaps to a small inline form inside the submenu: Enter/Create submits, Escape cancels, IME-composition Enters are guarded (`isComposing` / `keyCode 229`), a `sendingRef`-style lock prevents double submit.
- Create failures — invalid names, the reserved `live`, the 20-profile cap — show inline under the field with the typed name kept, replacing the equally unavailable `window.alert`.
- Hover-close is suppressed while the form is open so a drifting pointer cannot eat a half-typed name; when the form ends with the pointer already outside, the submenu closes with it (cross-model review finding).
- i18n: `chromeProfiles.newPlaceholder/newCreate/newCancel/newFailed` added and `newPrompt` removed across en/ko/zh/pl (pl coverage test enforced).

## Tests

New `WorkspaceChromeProfileMenu.test.tsx` (9 cases, including the review finding's stuck-open regression, mutation-verified); existing backendGate and pl-coverage suites pass — 88 tests in the Sidebar suite.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed Chrome profile creation so the “New profile…” option now opens an inline form instead of failing silently.
  - Added clear validation errors for invalid names, the reserved `live` name, and the profile limit.
  - Preserves entered names when creation fails.

- **Usability**
  - Added Create and Cancel actions, keyboard support, and localized text across supported languages.


openwong2kim:
The workspace menu's "New profile…" row was dead — Electron's renderer has no `window.prompt`, so the call threw; it's replaced by an inline form in `WorkspaceChromeProfileMenu` with validation errors, IME guards, and a double-submit lock. Renderer-only, plus i18n keys in en/ko/pl/zh and 9 new tests. P2: restores a completely unreachable feature, low risk.

PR #1074, opened 2026-08-29
fix(browser): stable chrome-backend surface ids — minted, persisted, and tracked through page-target succession

openwong2kim:
## Problem

The chrome backend handed Chrome's CDP targetId to agents as the MCP surfaceId, violating the `shared/browserTabs.ts` contract ("never a CDP target id"). Chrome replaces the target behind a tab on its own — the first-run sync flow does exactly this (dogfood 2026-08-29) — and the in-memory `tabOwners` map silently pruned any id missing from `/json/list`, so the agent's handle died and the tab went unreachable. A main restart lost every mapping outright, and `browser_close` never worked on chrome tabs at all (the renderer bridge was a silent no-op for them).

## Fix (two commits, one design)

**Stable identity + persistence** — the launcher mints `chrome-<uuid>` per tab; the CDP targetId is demoted to mutable state on a `ChromeSurfaceRecord`, persisted per profile and re-bound after a crash via the adopt path. A tab missing from `/json/list` is withheld from listings but kept for a 5-minute orphan TTL. Revival is deterministic-only — by tab anchor or surviving targetId, never by URL match (doctrine documented in the store header). `browser.cdp.info` ships `{surfaceId, targetId}` as distinct values, which the engine's registry-match-then-dial flow already supports, so **PlaywrightEngine needs no change** (regression tests prove it). `browser.close` gains its chrome branch with workspace-checked ownership on both the primary path and the `ownerOfSurface` fallback.

**Deterministic succession tracking** — Chrome 111+ tab targets survive page-target succession. The launcher anchors every wmux-opened tab to its tab target over a dedicated browser WebSocket (`CdpSocket`, extracted from LiveChromeClient — fixing its silent drop of every no-id CDP frame) and re-binds the record the moment `detachedFromTarget`/`attachedToTarget` announce a swap. The watcher is an overlay, not a precondition: any failure warns once and leaves the polling path untouched; `WMUX_CHROME_TARGET_WATCHER=0` disables it.

Also: `browser_session_status` now reports the active backend, bound profile, and live CDP port via a side-effect-free read (previously it described the unused Electron session).

Cross-model review findings fixed in the final commit (cross-workspace close on a shared profile, a watcher-latch race across Chrome restarts, CdpSocket singleflight + dispose races) — each regression test verified to fail against the pre-fix code.

## Verification

- 269 unit tests across browser-session / browser.rpc.backend / PlaywrightEngine (16 files), tsc clean.
- Live dogfood on a real build: `browser_open` returns a `chrome-` id addressable immediately; `cdp.info` shows the id split; the persisted record carries a live `tabTargetId` anchor; a hard-killed wmux re-adopts the surviving Chrome and the tab returns under the same surfaceId; `browser_close` closes it.

## Summary by CodeRabbit

* **New Features**
  * Chrome tabs now use stable handles that survive page replacements and application restarts.
  * Closing a Chrome tab now works reliably, including workspace-scoped closes.
  * Browser session status now shows the active backend, profile, running state, and CDP port.
  * Chrome automation can reconnect after browser restarts and preserve existing tab references.

* **Bug Fixes**
  * Improved tab recovery when Chrome temporarily loses or replaces page targets.
  * Added clearer errors when attempting to close unavailable tabs.


openwong2kim:
Stable `chrome-<uuid>` surface ids for the chrome backend, persisted in a new `ChromeSurfaceStore` and re-bound through page-target swaps by a tab-target watcher; `browser_close` finally works on chrome tabs, workspace-checked. Wide main-process surface (launcher, RPC handlers, quit wiring) but 269 tests plus live dogfood. P2: fixes dead agent handles and a no-op close.

PR #1075, opened 2026-08-29
fix(browser): splice ignored a11y nodes, and scope selector snapshots through the a11y tree

openwong2kim:
## Problem (dogfood 2026-08-29, root cause confirmed via CDP)

Chrome hangs a chain of ignored "uninteresting" wrappers (html → body → generic) directly under the RootWebArea of every real page. `buildTree()` dropped an ignored node together with its whole subtree, so the root lost its only child, `isRootOnly()` came back true, and **every snapshot silently demoted to the layout-blind DOM fallback** — taking `format:"aria"` and `filter:"interactive"` down with it, and handing out refs to invisible elements (a naver.com run burned all 100 refs on hidden autocomplete layers; measured: 8 ignored nodes out of 1126 collapsed the tree to zero children).

Separately, a `selector`-scoped snapshot took the DOM listing unconditionally even with a healthy tree.

## Fix (two commits)

**Splice, don't drop** — an ignored node disappears but its children take its place (the Playwright/Puppeteer rule), flattening arbitrarily deep chains and preserving sibling order. The same early-return bug in dom-intelligence's `collectInteractiveElements()` (the smart-snapshot path) is fixed the same way. Regression fixtures now model the real wrapper chain the previous fixtures idealised away — which is why no existing test caught this.

**Selector scoping through the a11y tree** — the selector resolves to a backendNodeId (three cheap DOM calls, before the tree fetch so a miss stays free) and the subtree serializes from the same full AX tree the unscoped path already fetches; any miss falls open to the existing DOM listing. Scoped refs resolve inside the selector only (so a same-role/name element outside the scope can't be clicked), and scoped auto-diff keys carry the rendering route so a transient fallback never diffs across formats.

## Tests

17 new cases (ignored-chain fixtures, sibling order, root-ignored, scoped a11y/aria/filter, fallback nulls, scoped ref confinement), all mutation-verified — reverting the fix fails exactly the new tests. Full `src/mcp/playwright` suite: 214 passed.


openwong2kim:
Fixes `browser_snapshot` silently collapsing to the DOM fallback on virtually every real page: ignored a11y wrapper nodes are now spliced rather than dropped, and selector-scoped snapshots go through the a11y tree. Touches `snapshot.ts` tree building, `dom-intelligence.ts`, and scoped ref resolution — the core browser MCP path, with 17 mutation-verified tests. P2: restores broken functionality (`format:"aria"`, `filter:"interactive"`, scoped refs) everywhere.

Issue #1072, opened 2026-08-28
Lifecycle self-echo suppression is order-fragile: a loaded landing inside the settle window defeats it


openwong2kim:
Follow-up to #1070, found in the final pre-tag dogfood of v3.49.0 (released-main MCP bundle, builtin backend).

## What happened

`browser_navigate https://www.wikipedia.org/` (plain, no redirect) returned:

```
[browser events]
- navigated: https://www.wikipedia.org/ (1s ago)
- loaded (0s ago)
---
Navigated to https://www.wikipedia.org/
```

The `navigated` line is the self-echo #1070 suppresses — but suppression checks only the **last** event of the post-drain slice (`suppressSelfEcho`, `src/mcp/playwright/automationLease.ts`). In this run `loaded` arrived within the 150ms settle window, so the slice was `[navigated, loaded]`, the last event was `loaded`, and the echo survived. Earlier dogfood runs passed only because `loaded` had landed after the window.

## Impact

Cosmetic only — a redundant but truthful line. No information is lost, attribution is correct (arguably stronger: `loaded` now shows inline). Filed as P3 noise reduction, not a correctness bug.

## Suggested fix

In the post-drain slice, drop the matching `navigated` even when trailing non-navigated events (`loaded`) follow it — e.g. find the LAST `navigated` in the slice and remove that entry iff its URL matches `redundantNavigationUrl`. Keep the existing narrowness otherwise (post-slice only, exact URL match, single removal). Unit case to add: slice `[navigated X, loaded]` + result URL X → `[loaded]` survives alone.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

openwong2kim:
## [wmux-hermes] Triage Summary

The `browser_navigate` self-echo suppression checks only the last event of the post-drain slice, so a `loaded` landing inside the 150ms settle window lets the redundant `navigated` line survive. Cosmetic output noise in `automationLease.ts`; nothing is lost. P3 — the suggested last-match removal plus the `[navigated X, loaded]` unit case is the right fix.

openwong2kim:
Shipped. #1076 (`be6e2617`) is in **v3.49.1**, released 2026-08-30, and listed in the changelog.

`suppressSelfEcho` now scans backwards for the last `navigated` entry and drops it on an exact URL match, so a `loaded` landing inside the 150ms settle window no longer defeats suppression. Exact-URL match and single removal kept as suggested; the `[navigated X, loaded]` unit case is covered.

PR #1076, opened 2026-08-29
fix(browser): four dogfood papercuts — hidden-content extraction, error prefixes, lifecycle newline, self-echo order (#1072)

openwong2kim:
## Fixes (chrome-backend dogfood 2026-08-29 + #1072)

- **browser_extract_text strips hidden content**: the in-page serializer prunes `display:none` / `visibility:hidden` / zero-rect elements (`display:contents` excepted), using the same visibility predicate as `browser_wait`. Previously only `[aria-hidden]` was excluded, so hidden banners/overlays led the extraction while smart_snapshot (innerText-based) stayed clean.
- **Tool errors stop leaking bundled-internal frame names** (`mcpServer.executeToolHandler: …`, `automationLease: …`): `describeToolError()` strips the leading dotted-identifier prefix at the boundary — the prefixes come from Playwright's apiName derivation against wmux's bundled frames, so there is no in-repo source to fix. Tool names and error classes are preserved.
- **The inline `[browser events]` block ends with a newline** again ("`…(5s ago)Navigated to…`").
- **#1072**: `suppressSelfEcho` drops the last matching `navigated` entry even when a trailing `loaded` landed inside the settle window; exact-URL match and single removal kept.

12 new tests (including a jsdom-safe harness that executes the real in-page script); full repo suite 12,668 passed.


openwong2kim:
Four browser-tool papercuts from dogfooding: hidden-content filtering in `browser_extract_text`, internal error-prefix stripping (new `toolError` module), a newline after the events block, and the #1072 self-echo fix. Everything is output-shaping — markdown-extractor, automationLease, tool catch blocks — with 12 new tests, so risk is low. P3: agent-facing polish, no behavior change.

PR #1079, opened 2026-08-29
feat(browser): keep password values out of every tool result

openwong2kim:
## Why

An agent driving a logged-in page needs to see the login form — not the password. Measured against a real Chrome, four paths carried password values into the model's context:

- the a11y snapshot leaked `autocomplete="new-password"` fields **in plaintext** (Chrome only bullet-masks `type=password`) and re-exposed both kinds through a StaticText descendant of the input;
- smart_snapshot's DOM/RPC route read `el.value` verbatim;
- `browser_type` echoed the typed text back into the transcript;
- `browser_response_body` returned login responses that restate the submitted value.

## What

A single predicate (`type=password` or `autocomplete` current-/new-password — `src/mcp/playwright/redact.ts`) drives:

- a **backendNodeId bridge** in the a11y tree: value replaced with `[redacted:password]` AND the subtree dropped so the shadow-editor text cannot resurface. Field structure, labels, refs, and filled-ness stay visible; a sibling username field stays untouched; a CDP failure fails open to Chrome's own behavior.
- the same in-page predicate on the DOM routes, a masked type echo, and **key-scoped** masking (`password`/`passwd`/`pwd`/`pass_word` in JSON, form bodies, query strings, and `user:pass@host` userinfo) on network output, console text, and every URL a tool result renders — navigate echoes, tabs listing, DOM-fallback URL lines, lifecycle events.

The cross-model review round was closed with live measurement: the real bypass was **open shadow roots** (their fields reach the AX tree while document-root `querySelectorAll` cannot cross the boundary) — fixed via `DOM.performSearch`, which pierces shadow roots and frames at a twentieth of the cost of a pierced full dump, and which the live run revealed returns zero unless `DOM.getDocument` primes the node map first. Truncation landing inside a password value no longer un-masks its head.

Deliberately out of scope: `browser_evaluate` (arbitrary JS is the tool's purpose, already gated), and a credential-vault/handle login action (owner-decision design, future work).

## Verification

tsc clean, golden probe PASS with **unchanged** baseline (tool surface untouched), `src/mcp` 507 tests pass (27 new redaction cases), live E2E against Chrome 141 covering light DOM + shadow root fields.

## Summary by CodeRabbit

* **Security**
  * Password values are now masked in browser snapshots, smart snapshots, typing results, navigation URLs, tabs, console messages, network details, response bodies, and extracted content.
  * Credentials in query parameters, form data, and URL authentication details are replaced with `[redacted:password]`.
  * Usernames, field labels, names, references, and other non-sensitive information remain visible.


openwong2kim:
This PR redacts password values (`type=password` and autocomplete-marked fields, shadow roots included) from every browser tool result — snapshots, `browser_type` echo, network/console output, and URLs. Main risk is over-redaction, but the new tests pin byte-for-byte pass-through of unrelated content, and `browser_evaluate` stays deliberately untouched. P2: closes a real credential-into-transcript leak with thorough test coverage.

PR #1078, opened 2026-08-29
perf(mcp): cut browser tool definitions by a fifth without touching a schema

openwong2kim:
## Why

Every wmux agent session pays the tool-definition surface up front — the same class of overhead measured at **14.4k tokens** for Playwright MCP's default list (microsoft/playwright-mcp#1290). wmux's 39 `browser_*` tools carried paragraph-length descriptions.

## What

- **Measured first**: new `scripts/measure-mcp-tokens.mjs` drives the built bundle over stdio and reports per-tool chars/description/schema splits (`--save`/`--compare` for before/after diffs).
- **Prose −43.5%** (11,501 → 6,499 chars); browser_* definitions overall **−20.6%** (~6.4k → ~5.1k tokens). The 30% stretch goal was arithmetically out of reach: 55% of the definition bytes are JSON-Schema structure, which is untouched.
- **Schemas byte-identical** — verified mechanically by stripping descriptions and diffing. Parameter names, types, enums, required flags all unchanged.
- **Behavioral contracts kept**: per-tool surfaceId default-target hints, browser_tabs' opaque-id addressing rules, evaluate's blocked-pattern list, sensitive-domain redactions and escape hatches, screenshot's needs-browser_open caveat. What went was duplication, restated parameter names, and implementation notes a caller cannot act on.
- Golden-wire baseline re-recorded for the intended description change (full profile only; commander hash untouched).

Next diet target (out of scope here): non-browser tools carry 2.8× the prose (channel_post alone is 3.7k chars).

## Verification

tsc clean, golden probe PASS, `src/mcp` 436 tests pass, schema-stripped diff empty. Cross-model review round applied (measurement-script arg/exit hardening).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RRxH4qJUX5pkr2DpoHTAX6

## Summary by CodeRabbit

* **Improvements**
  * Simplified browser tool descriptions for clearer, more concise guidance.
  * Preserved all tool parameters, validation, behavior, and compatibility.
  * Reduced prompt-token usage across the browser tool surface.

* **Developer Tools**
  * Added a command to measure and compare MCP tool token usage by profile and tool group.

PR #1080, opened 2026-08-29
fix(mcp): say what full does, and that password values come back redacted

openwong2kim:
## Why

Live dogfood of #1078 and #1079 — a fresh agent driving the slimmed tool surface against a real login form. It completed the task with **zero failed calls**, so the compression did not cause misuse. But it surfaced two things the descriptions should have said, both introduced by those two PRs:

- It inferred the **wrong rule** for redaction (guessed a label/name heuristic; the predicate is `type=password` / `autocomplete` current-|new-password) — the behavior shipped undocumented.
- It burned a round-trip re-snapshotting to find out whether `browser_type` replaces or appends.

## What

- **`full` gets its description back.** The diet deduplicated it against the tool text and left the parameter with none at all, which is worse than a short one.
- **`browser_snapshot` states the redaction**: password values read as `[redacted:password]`, the field stays listed and fillable, and a redacted value means the field *is* filled — answering the "did my input fail?" question the dogfood raised.
- **`browser_type`** says it replaces rather than appends, and that a password echo comes back redacted.

Cost: ~88 tokens against the ~1300 the diet saved. Golden-wire baseline moves with the intended description change.

## Verification

tsc clean, probe PASS after re-baseline, `src/mcp` 507 tests pass, token measurement re-run.

Pre-existing gaps the same dogfood found (console/network collection start, `browser_tabs` selected-vs-active, aria+filter combination, evaluate's return contract) are filed separately — they are not regressions from these PRs.

### DIFF
--- a/changelog.d/pr-g-snapshot-doc-regressions.md
+++ b/changelog.d/pr-g-snapshot-doc-regressions.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- `browser_snapshot`'s `full` parameter documents itself again — the tool-description
+  slimming in the previous release left it with no description at all, so nothing said
+  what it was for.
+- The browser tools now say that password field values read as `[redacted:password]`.
+  The redaction shipped without a word about it, and an agent reading a snapshot could
+  reasonably conclude its input had failed, or guess the wrong rule for which fields are
+  masked. `browser_type` likewise notes that it replaces an existing value rather than
+  appending to it.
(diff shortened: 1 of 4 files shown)


openwong2kim:
PR #1080 restores the `full` parameter description and documents password redaction and replace-not-append typing in `browser_snapshot`/`browser_type`. Description-only changes plus a golden-wire baseline re-hash, so no behavior is touched and risk is minimal. P3 — documentation polish, no functional impact.

Issue #1084, opened 2026-08-29
Update refused ("a process under the install root would not exit") on 3.48.1 — daemon quit fast, something else in the 5 watched processes didn't


p-poppe:
### What happened

On wmux 3.48.1 (Windows 11, build 26200.9168), clicking "Update" produced the new #1058/#1057 UX correctly — the toast ("wmux update / Installing the update... This window closes on its own.") and a terminal window appeared — but the install still failed. Screenshots: toast + terminal, and Task Manager showing a stranded `wmux` process.

### Evidence — `%APPDATA%\wmux\logs\main-2026-08-29.log`

**Attempt 1 (10:26:20 UTC):**
```
[AutoUpdater] Session save triggered before update install
[AutoUpdater] install handoff: waiter=...wmux-install-waiter-rvbWf3\wait-and-install.ps1
              watching 5 process(es) under C:\Users\PrzemekPoppe\AppData\Local\wmux;
              force-killed 0/0 foreign, 5 of our own left to app.quit(),
              daemon pid 28236 left to the graceful shutdown
[AutoUpdater] quitting for install — the daemon goes down with us so the installer runs against a dead tree
[Main] Full shutdown — racing daemon.shutdown (8000ms budget)
[2026-08-29T10:26:20.194Z] [info] [main] full-shutdown: racing daemon.shutdown
[Main] daemon.shutdown ack received (elapsed=634ms)
```
— then nothing further from this session. On the next launch:
```
[AutoUpdater] previous install was refused: install-aborted: a process under the install root would not exit
```

Task Manager at the time (screenshot) shows a single `wmux` process idling at 0% CPU / 0 MB/s disk / 0 Mb/s network, 87 MB — alive but doing nothing. I killed it by hand; only then did a retry succeed.

**Attempt 2 (10:29:10 UTC, after killing the stuck process, reinstalling 3.49.0 which had been released in the interim):**
```
[AutoUpdater] swept stale update artifact wmux-update-3.48.1-15108-wmux-3.48.1.Setup.exe
[AutoUpdater] reusing verified installer already in temp (wmux-update-3.49.0-47676-wmux-3.49.0.Setup.exe) — skipping download
[AutoUpdater] Update downloaded + verified (sha256 match) — ready to install
[AutoUpdater] Session save triggered before update install
[Main] Full shutdown — racing daemon.shutdown (8000ms budget)
[Main] daemon.shutdown ack received (elapsed=469ms)
```
[...]

PR #1085, opened 2026-08-29
fix(updater): force-kill a hung own-tree process instead of refusing the install (#1084)

p-poppe:
Addresses #1084, from the [concrete fix proposal](https://github.com/openwong2kim/wmux/issues/1084#issuecomment-5461923271) posted there.

## The incident

On 3.48.1, an update refused with the #980-era message ("a process under the install root would not exit"), even though the daemon acked its own `daemon.shutdown` RPC in 634ms. Task Manager showed a bare `wmux` process idling at 0% CPU well past that — killed by hand before a retry succeeded.

## Why this happens

`terminatePids`'s own doc comment in `installTeardown.ts` says why the daemon is spared from the pre-quit force-kill: "it gets the graceful shutdown so it can flush scrollback." The rest of `survey.ownTree` — the main window, renderer, GPU/utility helpers — ended up in the same wait-only treatment as a side effect of #980's fix (which correctly stopped force-killing them so `app.quit()`'s renderer-side session save could complete), not because they had anything left to protect. That session save (the `beforeunload` dispatch `performInstall` triggers up front) already finished before any of this runs. If one of them hangs anyway, the waiter had no escalation — it just sits out the full `INSTALL_LOCK_BUDGET_MS` (60s) and refuses.

## What this changes

`AutoUpdater.ts` now computes `forceKillEligiblePids = survey.ownTree` minus the daemon pid, and hands it to the waiter along with a new `OWN_TREE_FORCE_KILL_GRACE_MS` (5s). In the generated script, a handle for an eligible pid gets `min(graceBudget, budget)` off the *same* shared clock as everything else — not a fresh timer — so one reached late (after the clock already ran past the grace window on earlier handles) force-kills immediately instead of getting an unearned extra window. On expiry it runs the identical `taskkill /PID /T /F` the JS side already uses on foreign processes, gives it a short `WaitForExit`, and moves on **without** setting `$stuck` — that verdict stays reserved for a pid this waiter genuinely cannot end itself (the daemon, or anything eligible-but-still-alive after the kill attempt somehow leaves a lock behind — the file-lock probe (`Test-RootLocked`) right after this loop is untouched and remains the real backstop either way).

## Scope: daemon is deliberately NOT touched here

The original idea in the #1084 comment was to also force-kill the daemon once its `daemon.shutdown` RPC ack is in hand, since that ack already proves the flush completed. Reading `AutoUpdater.ts` closely showed that doesn't hold today: `spawnInstallWaiter` is called and its heartbeat verified **before** `onInstallRequiresFullShutdown()` even runs, so the ack (which happens later, inside the app's normal quit sequence) has no channel to reach an already-spawned, already-detached PowerShell process — that would need a new cross-process signal (e.g. the app writing a marker file the waiter polls for), which felt like a bigger, separate change with its own timing edge cases. This PR stays scoped to the set that never needed such a signal in the first place: own-tree processes that already got `app.quit()` and were never carrying anything to flush. The daemon keeps exactly today's contract — full budget, refuse on timeout — unchanged.

## Test plan

- Two new real-process runtime tests in `installTeardown.runtime.test.ts` (`describe.skipIf(!onWindows)`, real handles, real file locks, real `taskkill`):
  - `#1084 — force-kills a hung force-kill-eligible pid instead of refusing over it`: holds a real file lock via a spawned process, marks that pid force-kill-eligible with a 3s grace, confirms the waiter kills it, launches the fake `Setup.exe`, and exits 0 with no abort marker.
  - `#1084 — still refuses a hung pid that is NOT force-kill-eligible (the daemon path)`: same shape, empty eligible set — pins that the daemon's contract is byte-for-byte what it was before this PR.
- Existing `WaiterPlan` fixtures across `installTeardown.test.ts` and `installTeardown.runtime.test.ts` updated for the two new required fields (`forceKillEligiblePids`, `forceKillGraceMs`), all defaulting to `[]` / unchanged behavior where the test isn't about this feature.
- `npm run test:parallel` (excludes `*.runtime.test.ts`): 158/158 in `src/main/updater`.
- Real runtime suite (`vitest.runtime.config.ts`, actual PowerShell/Windows processes): 13/13 in `installTeardown.runtime.test.ts`.
- `npx tsc --noEmit`: clean. `eslint` on changed files: 0 errors (pre-existing warnings only, unrelated to this diff).

No version bump. Changelog fragment to follow once the PR number is assigned.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved the in-app update process by force-closing eligible application processes that remain open briefly after being asked to quit.
  - Updates no longer fail unnecessarily because of a hung main window, renderer, or helper process.
  - The background service continues to receive its existing wait period to finish saving state.

### DIFF
--- a/changelog.d/1085.md
+++ b/changelog.d/1085.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **An update could refuse itself over a hung process that had already been
+  asked to quit.** The in-app update waiter watches every process under the
+  install root and refuses to launch the installer if one of them does not
+  exit within its budget — necessary for the daemon, which needs time to
+  flush scrollback before it goes down, but the main window, renderer and
+  GPU/utility helpers were held to the same patient wait even though they
+  carry nothing to flush and had already been told to quit. If one of them
+  hung — a real machine showed a bare `wmux` process idling at 0% CPU well
+  past when it should have exited — the update refused entirely, and
+  recovering meant noticing the stray process in Task Manager and killing it
+  by hand before a retry would succeed. The waiter now gives that kind of
+  process a short grace period and force-kills it if it is still there,
+  instead of waiting out the full budget and giving up. The daemon's own
+  wait is unchanged.
--- a/src/main/updater/AutoUpdater.ts
+++ b/src/main/updater/AutoUpdater.ts
@@ -91,6 +91,17 @@ const INSTALL_QUIT_WATCHDOG_MS = 30_000;
 // takes time — a live process writes one file in well under a second, so a
 // few seconds of silence already means it is gone, not just busy.
 const WAITER_HEARTBEAT_BUDGET_MS = 3_000;
+// #1084 — how long the waiter gives an own-tree, non-daemon process (main
+// window, renderer, GPU/utility helpers) to exit on its own before force-
+// killing it, instead of sitting in the full INSTALL_LOCK_BUDGET_MS wait and
+// refusing. Unlike the daemon, these have no flush to protect — the session
+// save this function triggers up front is the only state that matters for
+// them, and it already happened before performInstall got this far. Short,
+// like WAITER_HEARTBEAT_BUDGET_MS: a cooperating process that already got
+// app.quit() exits in milliseconds, so a few seconds of silence means it is
+// hung, not still working. Kept well under INSTALL_LOCK_BUDGET_MS so the
+// daemon (never in this set) still gets the long, patient wait it needs.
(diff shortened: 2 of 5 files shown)


openwong2kim:

Closes #1084: the install waiter now force-kills hung own-tree processes (never the daemon) after a 5s grace, instead of refusing the update. Scope is tight — shared clock, fail-closed plan validation, daemon contract pinned by a runtime test. P2: the bug is P3-shaped, but new force-kill logic in the update path deserves prompt review.

PR #1087, opened 2026-08-29
fix(browser): strip the skip-navigation run browser_extract_text promised to remove

### LINKED ISSUE #1077: browser_extract_text does not strip skip-navigation links its own description promises to remove
Found in the post-merge dogfood of #1076 (chrome backend, real build, driven by a Claude Code session in a wmux pane).

## What happened

`browser_extract_text` on `https://www.naver.com/` returns this as its opening content:

```
[상단영역 바로가기](#topAsideButton)[서비스 메뉴 바로가기](#shortcutArea)[새소식 블록 바로가기](#newsstand)
[쇼핑 블록 바로가기](#shopping)[관심사 블록 바로가기](#feed)[MY 영역 바로가기](#account)
[위젯 보드 바로가기](#widgetboard)[보기 설정 바로가기](#viewSetting)
```

Eight skip-navigation links, then the newsstand tab labels. The first real article
(`[네팔 대홍수] 정부, 오늘 합동신속대응팀 9명 추가 파견…`) starts around **character 440**.

## Why it matters

The tool describes itself as *"stripping navigation and noise"*. A caller that
previews the first 200–300 characters to decide whether a page has content — a
common and reasonable use — concludes the page is empty or broken when it is not.

This is **not** the bug #1076 fixed. That one was about elements the browser does
not render (`display:none` / zero-rect); these skip links *are* rendered and are
legitimately in the a11y tree. This is structural navigation noise, and
`NOISE_SELECTORS` (`src/mcp/playwright/markdown-extractor.ts`) only catches it when
it is wrapped in `<nav>` / `<header>` / `[role=navigation]`, which these are not.

## Suggested direction

Either teach the extractor about skip links (in-page anchor links to an id, in the
first N nodes, no other content), or soften the description so it stops promising a
filter it does not apply. The first is better — the description is what makes the
current output surprising.

## Not in scope here

`aside` is also in `NOISE_SELECTORS` and is dropped wholesale. That is pre-existing
and deliberate as far as the code shows, but it is the same class of question (what
counts as "noise") and worth deciding together.

openwong2kim:
Closes #1077.

## Why

`browser_extract_text` advertises *"stripping navigation and noise"*, but on naver.com its eight skip links owned the **first 440 characters** — long enough that a caller previewing 200–300 characters to decide whether a page has content concludes it is empty.

These are not the hidden elements #1076 removed. Skip links render, and they belong in the accessibility tree (they exist for screen-reader users). They are structural navigation that `NOISE_SELECTORS` only catches when wrapped in `<nav>`/`<header>`.

## How

Walk the document from the start, removing skip links (and wrappers holding nothing else) until the first node producing real content — then stop for good. No tuned "first N nodes" threshold was needed: every table of contents measured sits *after* a heading, so "first content node" is both more accurate and self-adjusting.

Guards against over-removal:
- descends only through block wrappers with no text of their own, so an anchor opening the first paragraph survives;
- an anchor resolving to a **heading** counts as content, not a skip link — this protects a TOC even when one leads the page;
- bails entirely past twelve links rather than partially strip an unfamiliar shape (measured: naver 8, MDN 2, Wikipedia 1);
- hidden nodes do not end the run, since the promo banner #1076 removes sits ahead of naver's skip links.

Deliberately **not** a condition: whether the anchor target id exists. naver's eighth link points at an id the document does not contain, so requiring resolution left exactly that link behind. Target resolution now only ever argues for *keeping* a link.

## Measured

| Page | Result |
| --- | --- |
| naver.com | first article moves from char 440 → **~200** |
| MDN | skip links gone, `# Element: getClientRects()` leads |
| Wikipedia | `Jump to content` gone, article text leads |

Over-removal control (byte-identical before/after): Python docs page whose TOC is pure in-page anchors, an RFC, a GitHub repo page. MDN's mid-body `#browser-compatibility` anchor survives.

## Verification

tsc clean, `src/mcp/playwright` 308 tests pass (8 new), golden probe passes with **unchanged** baseline (no tool description was touched).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved extracted markdown by removing leading “Skip to content” navigation links.
  - Preserved tables of contents, regular in-page links, and meaningful page content during extraction.
  - Handles pages with multiple skip links and nested navigation structures more reliably.

- **Documentation**
  - Added a changelog entry describing the improved text extraction behavior.

### DIFF
--- a/changelog.d/pr-h-extract-text-skip-nav.md
+++ b/changelog.d/pr-h-extract-text-skip-nav.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **`browser_extract_text` now strips the skip-navigation links it promised to
+  remove.** Pages open with hidden-until-focused "skip to content" anchors for
+  keyboard users, and they came through as the first thing in the extracted
+  markdown — on naver.com eight of them pushed the first headline past
+  character 440, so a caller that previews the opening few hundred characters
+  read the page as empty. The leading run of in-page anchors is now dropped,
+  while a table of contents and links inside the text are kept. (#1077)
--- a/src/mcp/playwright/__tests__/markdown-extractor.skiplinks.test.ts
+++ b/src/mcp/playwright/__tests__/markdown-extractor.skiplinks.test.ts
@@ -1,0 +1,158 @@
+// @vitest-environment jsdom
+//
+// Skip-link filtering for browser_extract_text (issue #1077).
+//
+// Like the hidden-content tests, the evaluator here runs the real in-page
+// serialisation script against a jsdom document, because that is where the
+// filter lives.
+//
+// Dogfooding naver.com (2026-08-29) opened the extracted markdown with eight
+// "바로가기" skip links and pushed the first headline past character 440, so a
+// caller previewing the first few hundred characters saw an empty page. The
+// risk in filtering them is over-removal: a table of contents and a footnote
+// reference are the same `a[href^="#"]` shape, so these tests pin both sides.
+import { describe, it, expect, beforeAll, afterEach } from 'vitest';
+import { extractMarkdown } from '../markdown-extractor';
+
+// jsdom has no layout engine: getClientRects() is unconditionally empty, which
+// would make the script judge EVERY element hidden. Model layout instead —
+// an element renders a box unless it opts out with data-zero-rect.
+beforeAll(() => {
+  Object.defineProperty(Element.prototype, 'getClientRects', {
+    configurable: true,
+    value(this: Element) {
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR fixes #1077 by stripping the leading run of skip-navigation links from `browser_extract_text` output via a new in-page stripper in `markdown-extractor.ts`. The main risk is over-removal (TOCs and footnote anchors share the same shape), but the cap, heading-target guard, and eight jsdom tests pin both sides well. P2 — direct fix for #1077 with solid guards and tests.

Issue #1086, opened 2026-08-29
Remote workspaces: one click doesn't return to a local workspace, and remote rows lack rename/color/startup parity with local ones


p-poppe:
### Bug 1: switching back to a local workspace after visiting a remote one needs two clicks

Repro (real, on 3.49.0 with #1067's remote-bootstrap): pair a remote host, create a new workspace on it (or just click an existing remote entry in the sidebar's bottom "attached hosts" section) — the remote pane shows. Click a local workspace row (e.g. "CTO") that was active before: nothing happens, still showing the remote pane. Click a *different* local workspace, then click "CTO" again: now it switches correctly.

What I traced so far (static read, not yet reproduced live):

- Local selection (`activeWorkspaceId`) and remote selection (`activeRemoteKey`) are two independent pieces of state (`workspaceSlice.ts` / `remoteWorkspacesSlice.ts`).
- `WorkspaceCenter.tsx` decides what's visible by checking `activeRemoteKey` **first** — a local workspace's pane is `display: none` whenever `activeRemoteKey` is set, regardless of `activeWorkspaceId`.
- There's a documented convention for this exact hazard: `clearRemoteSelection()` in `workspaceSlice.ts`, with a comment saying it must be called at *every* site that assigns `activeWorkspaceId`, "so selecting/promoting a local workspace always drops any remote mirror selection. Without this, WorkspaceCenter... keeps showing a remote mirror while the sidebar highlights the newly-active local workspace."
- `setActiveWorkspace` (`WorkspaceItem`'s `onSelect`) *does* call `clearRemoteSelection(state)` — so on paper, one click on a local row should already un-hide it. It doesn't, live. Whatever's actually eating the first click needs a real repro (dev app or an integration test that simulates the click sequence), not more static reading — I didn't want to guess further without seeing it happen.

### Bug 2 / feature gap: remote workspaces are second-class in the sidebar

Once a workspace is bootstrapped on a remote host (#1067) or attached (pre-existing feature), it:
- can't be renamed (no context-menu item — compare `WorkspaceItem`'s rename vs. `RemoteWorkspaceItem`'s single "Detach" item)
- can't get a color tag (same gap)
- has no strong visual "this is remote" marker beyond a small subtitle line showing the host label — asked for something closer to parity with local workspaces plus a clear remote indicator, not a stripped-down row
[...]

PR #1089, opened 2026-08-29
fix(sidebar): clear activeRemoteKey on a pane-row jump within the already-active workspace

p-poppe:
Addresses #1086 (the click-through bug half — Bug 2 there, rename/color/marker/startup parity, is a separate follow-up per the issue thread).

## Root cause

`activatePaneTarget` (`src/renderer/hooks/useNotificationListener.ts`) guards the only call to `setActiveWorkspace` in the function on a workspace-id inequality:

```ts
if (target.workspaceId !== state.activeWorkspaceId) {
  state.setActiveWorkspace(target.workspaceId);
}
```

`setActiveWorkspace` is also the only place that calls `clearRemoteSelection` — unconditionally, once past its own existence guard (`workspaceSlice.ts`). `activeWorkspaceId` never changes when a remote mirror becomes the visible pane: `WorkspaceCenter` checks `activeRemoteKey` first, ahead of the local tree, and `activeRemoteKey` is a wholly separate piece of state.

So: click a workspace's own top-level sidebar row (`WorkspaceItem`'s `onSelect={setActiveWorkspace}`, no guard) — works, always calls `setActiveWorkspace`. Click a *nested* pane/agent row inside that same, already-active workspace (`WorkspaceAgentRoster` → `focusPaneByPtyId` → `focusNotificationTarget` → `activatePaneTarget`) while a remote mirror is showing — the guard above sees `target.workspaceId === state.activeWorkspaceId` (true — the workspace never stopped being "active" in the local-only sense), skips `setActiveWorkspace`, and the remote mirror stays on screen with no way back except clicking a *different* workspace first.

Traced live against the reporter's running app via CDP (MutationObserver on the pane-grid wrapper's `display` style + a click listener), not from static reading alone — the original issue thread has the full trace, including two wrong turns before landing on this.

## Fix

Widen the guard to also fire `setActiveWorkspace` when a remote mirror is currently showing:

```ts
if (target.workspaceId !== state.activeWorkspaceId || state.activeRemoteKey) {
  state.setActiveWorkspace(target.workspaceId);
}
```

`activeRemoteKey` is added to `FocusTargetState` as optional (same convention as `unstashPane` — minimal test fixtures without the remote slice mounted stay terse, and `undefined` correctly reads as "no mirror showing").

## Test plan

- New test `AT4` (`useNotificationListener.test.ts`): `activeWorkspaceId` already equals the jump target, `activeRemoteKey` set — asserts `setActiveWorkspace` is now called and the mirror selection clears. Existing `AT2` (no mirror showing → still a no-op) is untouched and still passes, confirming the widened guard doesn't regress the plain case.
- `npm run test:parallel` on the touched files: 41/41 (`useNotificationListener.test.ts`) plus 145/145 across `Notification`, `FleetView`, `Sidebar`, `Toast` component suites (both call sites of `activatePaneTarget`/`focusNotificationTarget`) — all pass.
- `npx tsc --noEmit` and `eslint` on both changed files: clean.

Bug 2 from #1086 (rename/color/remote-marker/startup-config parity for remote workspace rows) is left untouched, per the prior diagnostic comment recommending it as a separate follow-up.

## Summary by CodeRabbit

* **Bug Fixes**
  * Clicking nested panes or agent rows now switches from a remote workspace mirror back to the local view in one click.
  * Selecting an item in the currently active workspace correctly clears the remote mirror and displays the local workspace tree.

### DIFF
--- a/changelog.d/1089.md
+++ b/changelog.d/1089.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Clicking a pane or agent row nested inside the currently-active workspace no longer leaves an attached remote workspace's mirror on screen — the jump now correctly switches back to the local view in one click, matching the workspace's own sidebar row (#1086).
--- a/src/renderer/hooks/__tests__/useNotificationListener.test.ts
+++ b/src/renderer/hooks/__tests__/useNotificationListener.test.ts
@@ -699,7 +699,7 @@ describe('focusNotificationTarget', () => {
 // ─── S-C1 — focusPaneByPtyId (Fleet card click → pane jump) ─────────────────
 
 describe('focusPaneByPtyId (Fleet View jump)', () => {
-  function makeState(opts: { workspaces: Workspace[]; activeWorkspaceId: string }) {
+  function makeState(opts: { workspaces: Workspace[]; activeWorkspaceId: string; activeRemoteKey?: string | null }) {
     const spies = {
       setActiveWorkspace: vi.fn(),
       setActivePane: vi.fn(),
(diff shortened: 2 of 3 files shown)


openwong2kim:

This PR fixes the #1086 one-click return bug by widening the `activatePaneTarget` guard to also fire `setActiveWorkspace` when `activeRemoteKey` is set, dropping the remote mirror. It touches only `useNotificationListener.ts` plus one new test, so risk is low. P3 — narrow UX fix with an existing workaround, cleanly scoped and tested.

PR #1090, opened 2026-08-29
fix(browser): start collecting console and network when the page opens, not when someone asks

### LINKED ISSUE #1081: browser_console and browser_network only start collecting on their first call, so the messages that matter are already gone
Found in the post-merge dogfood of #1078/#1079 (chrome backend, real build, MCP tool path).

## What happens

```
browser_navigate  http://127.0.0.1:8731/     # the page logs on load
browser_console                              # -> "No console messages collected."
browser_evaluate  console.log('after-attach …')
browser_console                              # -> the message appears
```

The listener attaches on the **first** `browser_console` call. Anything the page
logged before that — which is every load-time error, every failed request, every
uncaught exception — is not recorded anywhere and cannot be recovered. The same
sequence reproduces for `browser_network`: requests issued before the first
`browser_network` call are absent, and a "prime" call is what makes later ones
appear.

## Why it matters

The moment an agent reaches for `browser_console` is *after* something looked
wrong. That is exactly the window the tool cannot see. Worse, the empty result is
indistinguishable from a clean page: an agent can truthfully report "no console
errors" about a page that threw on load.

Same shape for `browser_network` — an agent investigating a failed login cannot
see the request that failed, only requests made after it started looking.

## Suggested direction

Attach the console/network listeners when the page is first resolved (open /
navigate / attach), not on first read, with a bounded ring buffer so an
unattended page cannot grow them without limit. If eager attachment is too
costly to make unconditional, the fallback is to make the emptiness honest —
distinguish "collection started at T, nothing since" from "collection had not
started" in the tool result, so an agent stops reporting the second as the first.

Noted while filing: the tool description ("Read console messages collected from
the page. They accumulate; clear:true resets.") does not say when accumulation
begins, which is what made this ambiguous to the agent that hit it.

openwong2kim:
Closes #1081.

## Why

The listeners attached on the **first** `browser_console` / `browser_network` call — so the load-time error and the failed request an agent goes looking for were already gone. Worse, an empty result was indistinguishable from a clean page: an agent could truthfully report "no console errors" about a page that threw on load.

Investigating found the two transports failed **differently**:

| Lane | Console | Network |
| --- | --- | --- |
| Playwright | lost | lost |
| main-process RPC | *survived by accident* — Chrome replays stored messages on `Runtime.enable` | lost |

Worst case was a dev `builtin` build, where `browser_navigate` never resolves a Page and `browser_console` was the call that created one.

## What

Capture attaches where each lane first knows about the page — `getPageForScope()` for Playwright (idempotent per Page) and `WebviewCdpManager.register()` for main, which fires on did-attach, before the guest loads. `builtin` now reads the main buffer on every build, so dev and packaged behave alike.

Attaching early means buffers fill for pages nobody reads, so entries are bounded where they were not: 4KB per console message, 2KB per URL, and a **4MB total body budget with oldest-first eviction** — the Playwright lane previously had no body budget at all (a theoretical 256MB). Worst case is now ~10MB per page.

Emptiness is honest either way: results state what they have been collecting *since*, and say so explicitly when a page was already open when collection started.

Also fixed while measuring: Playwright reports the warning level as `'warning'` while the filter and the main lane use `'warn'`, so `level:'warn'` silently matched nothing on the chrome backend.

## Review round

Two real leaks: an in-flight `response.text()` could attach a body to a ring-evicted entry, whose bytes then counted toward the budget forever until it evicted *live* bodies; and eager attachment pointed body retention at `text/event-stream`, where `.text()` never settles. Both closed, plus lifecycle URL truncation, `missedBefore` clearing on a fresh document, and listener removal on close.

And the claim the change was built on — that a page throwing on load ends up in the buffer — was **false in both lanes** (Playwright delivers uncaught exceptions as `pageerror`, main had no `Runtime.exceptionThrown` case). Rather than soften the sentence, both lanes now capture them.

## Verification

tsc clean, golden probe passes, **1443 tests** across `src/mcp` + `src/main/browser-session` + `src/main/pipe/handlers`. Live against Chrome 151: the uncaught `TypeError` and its stack appear in both lanes from a page that threw before anyone looked; the SSE request is listed with no body retained.


openwong2kim:

Console/network capture moves from first-query to page-open attachment across both lanes (Playwright and main-process RPC), with bounded buffers and honest emptiness semantics. Touches the browser-session capture lifecycle and memory budgets; the review-round leak fixes keep risk contained. P3 — correctness improvement for browser tooling, verified by 1443 tests and live runs.

PR #1092, opened 2026-08-29
docs(mcp): state four browser tool contracts the descriptions left to guesswork

### LINKED ISSUE #1082: Browser tool descriptions leave four contracts unstated (selected vs active, aria+filter, evaluate's return, reserved ref)
Found in the post-merge dogfood of #1078/#1079. A fresh agent drove the browser
tools against a real form and completed the task with zero failed calls, then was
asked what it could not tell from the descriptions alone. Four answers, all
pre-existing (none introduced by the tool-description slimming — the two that
were are fixed in #1080).

## 1. `browser_tabs`: `selected` vs "the active surface"

A workspace with exactly one browser surface lists it as `selected: false`, while
every other tool's `surfaceId` says "Omit for the active surface". Nothing
connects the two words, so it is unclear whether omitting `surfaceId` is safe
when the only surface reports `selected: false`. (It is — omitting worked.)
Either the field is misnamed for what it reports, or the relationship needs one
sentence.

Worth keeping as-is by contrast: `select moves UI focus only and does NOT
retarget the other browser tools` was called out as unusually clear.

## 2. `browser_snapshot`: what `format:"aria"` + `filter:"interactive"` does

`filter` is only applied on the `ai` format (`snapshot.ts`, the strip is gated on
`format === 'ai'`), so passing it with `format:"aria"` is silently ignored. The
same file already sets the honest precedent for the neighbouring case — a
selector with `aria` emits a note explaining the fallback — so silence here is
inconsistent with the file's own convention. Whether `aria` output carries refs
is also unstated.

## 3. `browser_evaluate`: no return contract

The blocked-pattern list is well documented. What is missing: how a non-scalar
return is serialized, whether top-level `await` is available, and whether the
dangerous-pattern check is string matching or AST analysis — the last one
determines whether an innocuous expression that merely *contains* a blocked
substring will be refused.

## 4. `ref` is advertised but not implemented

`browser_snapshot`'s `ref` parameter is described as `Reserved; not implemented
yet` and still appears in the schema, so an agent reads it on every first call
and has to decide to ignore it. Either drop it from the shape until it works, or
say what it will do.

## Not in this issue

Collection-start semantics for `browser_console`/`browser_network` are #1081 —
that one is a behavior gap, not a wording gap.

openwong2kim:
Closes #1082.

## Why

A fresh agent drove the browser tools against a real form, completed the task with zero failed calls, and was then asked what it *could not* tell from the descriptions alone. Four answers — each now a clause, not a paragraph.

## The four, each verified against the code (and Chrome, for #3)

**1. `browser_tabs`' `selected` reports UI focus, not tool targeting.** It is `false` whenever a terminal pane holds focus, and hardcoded `false` on the chrome backend, while omitting `surfaceId` targets the workspace's first registered CDP session regardless. An agent seeing `selected: false` on its only surface had no way to know omitting the id was still correct.

**2. `filter` is gated on the `ai` format**, so pairing it with `aria` silently did nothing. The same file already emits an honest note when a selector cannot be scoped for aria, so silence here contradicted its own convention. `aria` now says the filter was ignored — filtering would break the very contract `aria` names — and the schema records that aria output carries no refs.

**3. `browser_evaluate`'s return contract.** Measured against a real Chrome, not inferred:

| Input | Result |
| --- | --- |
| top-level `await` | `SyntaxError` |
| returned Promise | awaited (`42`) |
| object | JSON |
| `undefined` | `undefined` |
| DOM node / Map / function | `{}` |

Blocking is a **text scan**, so a blocked name inside a string or comment is refused too — worth knowing before an agent tries to work around a false positive.

**4. `ref` removed from the snapshot schema.** It was advertised as reserved-but-unimplemented, so every agent read it on every call and had to decide to ignore it. The handler never destructured it; nothing in the repo reads it.

## Cost

+335 characters across the browser surface (+84 tokens), inside the 400 budget — `snapshot` actually *shrank* 36 characters because removing `ref` took its schema with it. Well within what #1078's diet bought.

## Verification

tsc clean, `src/mcp` 555 tests pass (3 new pinning the aria note), golden probe re-baselined for the intended surface change (commander profile untouched).

## Summary by CodeRabbit

- **Changed**
  - Clarified browser snapshot behavior: `"aria"` format returns the complete accessibility tree without refs, and interactive filtering is ignored with an explanatory note.
  - Documented `browser_evaluate` return behavior, including Promise handling and top-level `await` limitations.
  - Clarified that browser tab selection reflects UI focus, not tool targeting.
- **Removed**
  - Removed the `ref` parameter from the browser snapshot interface.
- **Documentation**
  - Updated tool descriptions and release documentation to reflect these contracts.


openwong2kim:

Docs PR closing #1082: states four browser tool contracts (selected vs targeting, aria+filter, evaluate return, drops the reserved `ref` param) in the tool descriptions. Mostly schema/description edits, with one small behavior change in `snapshot.ts` — aria now notes an ignored `filter` instead of staying silent — plus a re-baselined golden probe hash. P3: low risk, clarity-only, within the token budget.

PR #1093, opened 2026-08-29
feat(browser): mark focus, iframe boundaries, and what an overlay is blocking

openwong2kim:
## Why

Three things an agent needed from a snapshot and could not get. Each is added **only where it is the exception**, so a page with none of them serializes byte-identical — GitHub, Hacker News, StackOverflow, Wikipedia and MDN all measured at **+0 characters**.

| Signal | Cost | Why it was missing |
| --- | --- | --- |
| `focused` | +8 chars, one node per page | Already in the AX tree, simply never printed |
| iframe boundary | +104 chars for two frames | Arrived as a childless leaf, indistinguishable from an empty one |
| `clickable` + overlay note | +240 chars on Bootstrap's modal demo | An agent learned occlusion from a 30s click timeout |

## The occlusion work is mostly restraint

Chrome already models the spec'd cases: `showModal()` and `inert` both remove the blocked subtree from the tree (measured 15→9 and 13→6 nodes). The only real gap is what the web actually uses — a plain high-z-index `div`.

A naive centre-point hit test was measured and **thrown out**: it called 83 of GitHub's 177 controls blocked and 27 of Wikipedia's 95, all clickable. Telling an agent a reachable control is blocked is worse than the status quo — a timeout is recoverable, a false negative makes it give up silently.

So the hit test runs only behind a gate. The review round found the first gate still fired on `position:fixed; inset:0` app shells; **reproduced on openstreetmap.org** (1 misfire out of 40 controls was enough to condemn the page) before fixing. Two guards now: a layer containing >50% of matched controls is a layout root, and a layer blocking ≤50% of what was tested is not covering the page. The second exists because excalidraw's layer is a `canvas` that contains no controls at all, so the first cannot see it.

Measured after: **0 detections across 8 app-shell sites** (OSM, excalidraw, diagrams.net, Google Maps, Linear, Stripe, Observable, dashboard fixture) while both true positives still fire.

Known cost — recall: a dialog holding most of a page's controls is now missed. That is the safe direction for a signal whose false positives are silent.

## Also from the review round

The `finally` block awaited `Runtime.releaseObjectGroup` with no timeout, so a wedged renderer that had just timed out the evaluate would hang the entire snapshot — the one thing the module header promises not to do. Plus: one shared 2s budget (worst case was ~106s), batched `describeNode`, a hard loop bound against a page patching `Array.prototype`, a clipped layer label counted against `maxLength`, `pointer-events:none` treated as blocked, and the frame note attached only to a genuinely childless iframe.

Gate behaviour is pinned by a harness that runs the **shipped probe source** against a stubbed DOM — layout-dependent decisions cannot be reached through a CDP mock.

## Verification

tsc clean, `src/mcp` 580 tests pass, golden probe passes with **unchanged** baseline (no tool surface change). Live measured against 20+ production pages.

Follow-ups left open: the tool description does not yet teach the three markers, frame-aware ref resolution (which must land before iframe contents can be stitched in without minting dead refs), and out-of-process iframes.


openwong2kim:

Adds focused, iframe-boundary, and clickable/overlay markers to the ai snapshot, with a guarded hit-test so false occlusion reports stay at zero across app shells. Snapshot serialization only; clean pages measured byte-identical. P3 — additive signal for browser agents, pinned by a DOM-stubbed harness and live measurements.

Issue #1091, opened 2026-08-29
A workspace bootstrapped on a remote host is a flat read-only mirror grid, not a real pane tree like a local workspace


p-poppe:
### What happens

After #1067, "New workspace on this host" creates a brand-new workspace on a paired remote daemon. Opening it shows a single cell in \`RemoteWorkspaceView\`'s grid (\`src/renderer/components/Remote/RemoteWorkspaceView.tsx\`) — a read-only SSE mirror of the remote session, capped at \`MAX_MIRRORS = 6\` with a "+N more" note beyond that.

There's no split, no add-pane, no zoom, no interactive pane management at all — it's a fixed grid of mirrors, not a workspace shell. A user creating a workspace on a remote host reasonably expects it to work like a local one: start with one pane, split/add more, arrange them, interact normally.

### Why this is worth a look

This predates #1067 (\`RemoteWorkspaceView\`/the mirror-grid model existed for the older "attach an existing remote workspace to watch it" use case, where read-only mirroring of whatever the remote GUI/CLI already has running is the right model). #1067 changes the use case: bootstrapping a *brand-new* workspace from the desktop app implies you want to *work* in it, not just watch it — and the mirror-grid model doesn't fit that.

### Scope question for whoever picks this up

This is bigger than the rename/color/marker gaps already tracked in #1086 (Bug 2) — those are cosmetic; this is "does a remote-bootstrapped workspace get a real, locally-rendered pane tree talking to the remote daemon's PTYs directly (like the desktop app's own panes do to its own daemon), or does the mirror-grid model get extended with interactivity (add/split/close mirrors)?" Two different architectures, not a small patch — flagging for a direction call rather than guessing.

### Confirmation

- [x] Searched existing issues for "mirror grid", "RemoteWorkspaceView", "remote pane split" — no duplicate found; related to #1086 (Bug 2) but distinct in scope.

openwong2kim:
## [wmux-hermes] Triage Summary

This asks whether a workspace bootstrapped on a remote host (#1067) should become a real interactive pane tree or stay the read-only mirror grid in `RemoteWorkspaceView.tsx`. The decision is architectural (direct remote PTY panes vs. extending mirrors with interactivity), not a small patch. P3 — nothing is broken today, but the choice gates remote workspace UX.

openwong2kim:
## Direction call (owner decision)

[...]

PR #1094, opened 2026-08-29
feat(remote): a resizable split tree for remote workspaces, with grow/shrink (#1091)

p-poppe:
## Update: the "real split tree" half is now in this PR too

The user asked directly for parity with a local workspace's own split/resize, not just grow/shrink in a fixed grid — so the second half of #1091's scope question is answered by building it, not by deferring further.

- **`remotePaneTree.ts`** — a pure leaf/branch tree (split/close/resize/reconcile), completely independent of both the server's flat pane list and the local workspace's own pane tree (`shared/types.ts` `Pane`) — zero shared code, zero risk to local workspaces.
- **`RemotePaneContainer.tsx`** — the same `react-resizable-panels` primitives (`Group`/`Panel`/`Separator`) the local `PaneContainer` already uses, so drag-to-resize works identically, for free.
- "Add pane" became "Split right" / "Split down" — split targets whichever pane was last clicked; close removes a leaf and collapses its sibling branch, same discipline a local workspace's pane-close uses.

**A real bug found and fixed along the way**: collapsing a two-pane split down to one pane changed the returned element's *type* at that tree position (a `Group` becomes a bare `Fragment`-wrapped leaf) — React unmounts the whole subtree on a type change, so closing one of two panes was tearing down and re-attaching the survivor's live SSE mirror for no reason. Fixed with a render-time-only wrapper (`wrapForRender`) that keeps the outer element type constant across the 1-pane <-> N-pane transition — the *stored* tree still collapses to a bare leaf for a single pane, keeping `splitLeaf`/`removeLeaf`'s own invariants and tests simple; only the render call site always wraps.

Also fixed in the same branch, unrelated to this feature: two `vi.fn()` mocks in `RemoteHostClient.test.ts` had no declared parameter signature, so TS inferred the empty-tuple call-args type (same class of bug as #1057) — this failed `tsc` for the whole PR. And two pre-existing test files (`RemoteWorkspaceView.paneManagement.test.tsx`, `.attach.test.tsx`) needed the same `ResizeObserver` stub `PaneContainer.moveSizes.test.tsx` already uses, since they now actually mount a `react-resizable-panels` `Group`.

**Tests**: `remotePaneTree.ts` gets 23 pure unit tests; `RemoteWorkspaceView` gets 6 new component tests (`RemoteWorkspaceView.splitTree.test.tsx`) covering split/close/reconcile against the real render tree. Full `test:parallel`: 12772/12778 passed — the 4 failures verified pre-existing and unrelated (playwright MCP + deck loop timing tests) via `git stash` on this PR's prior commit.

---

## Original PR (first commit): grow/shrink in the existing fixed grid

Addresses the concrete "does this need to be tracked separately" bit of #1091: a workspace bootstrapped on a paired remote host (#1067's "New workspace on this host") sat as a fixed grid of whatever panes happened to already be there, with no way to add another one or close one — unlike a local workspace's add/close-pane controls. **This is the "grow/shrink like local" half of #1091, not the "real drag-resizable split tree instead of a mirror grid" half** — see #1091's own scope question for why those are two different architectures; this PR deliberately keeps the existing `RemoteWorkspaceView` grid layout and only adds the missing add/close actions on top of it, so the change stays isolated and low-risk rather than reworking the view's rendering model.

## What changed

**Server-side: nothing.** `POST /api/sessions` with an already-live `workspaceId` already passes `rejectWorkspaceId`'s existence check — the check doesn't distinguish "the first pane of this workspace" from "a later one", so growing an existing remote workspace was already possible from the wire's point of view. `DELETE /api/sessions/:id` (`handleSessionDelete`) already exists too, gated on `mayInput` exactly like session creation.

**`RemoteHostClient.closeSession`** — the one destroy/delete call this class makes. Its class-level doc comment used to say "observer + input only: it never calls a destroy/delete endpoint on the remote host" — that was written for the older attach-to-an-*existing*-workspace-to-watch-it use case, where destroying someone else's pane from a mirror would be a real surprise. #1067 added a second use case: bootstrapping a *brand-new* workspace on a paired host to actually work in it. A workspace you just created has no "someone else" to surprise — closing one of its own panes is exactly what a local workspace's close button does, so I'm treating this as a deliberate, narrow, documented exception rather than quietly working around the old comment. Flagging this explicitly in case it reads as more architecturally sensitive than I'm judging it to be.

**Two new IPC channels:**
- `REMOTE_WORKSPACE_PANE_ADD` — an alias over the exact same `RemoteHostClient.createWorkspace` call `REMOTE_WORKSPACE_CREATE` (#1067) already makes. Kept as its own channel (not a reuse of the existing one) purely so the renderer's *intent* — "grow this workspace" vs. "bootstrap a brand-new one" — stays distinct in the code, even though the wire request is identical.
- `REMOTE_SESSION_CLOSE` — wraps `closeSession`.

**`RemoteWorkspaceView`**: a "+ Add pane" button in a small footer bar, and a "×" close button in each pane cell's header. Both are hidden on a read-only host (`allowInput === false`) — the server would 403 a `mayInput`-gated call anyway, so there's no point offering it. Both apply the result to the store immediately via the existing `setRemoteWorkspacePanes` (an add appends the new sessionId, relying on `mergePaneSets` to keep it; a close is a `.filter()` of the current list) rather than waiting up to `POLL_INTERVAL_MS` (10s) for `useRemoteAttachmentsLifecycle`'s own refetch to notice — the workspace visibly grows/shrinks the moment the request that did it succeeds.

## What's deliberately NOT in this PR

The mirror-grid rendering model itself (fixed `MAX_MIRRORS = 6`, auto-arranged 1/2/2×2/3×2 layout, no manual split direction, no drag-resize, no per-pane zoom) is unchanged. That's the bigger architecture question #1091 raises — a real, locally-rendered pane tree talking to the remote daemon's PTYs the way local panes do, vs. extending the mirror-grid model further — and I don't think it should get decided implicitly by me picking a direction inside this PR. This PR is scoped to the smaller, unambiguous parity gap: you couldn't add or remove panes at all, regardless of which rendering model wins that question later.

## Test plan

- New: `RemoteHostClient.test.ts` — `closeSession` (#1091) describe block: DELETE with Bearer auth, URL-encodes the session id, treats 404 as success, surfaces the daemon's error detail on other non-OK statuses, falls back to a generic message when the error body isn't JSON.
- New: `remote.handler.test.ts` — `workspacePaneAdd` (mirrors `workspaceCreate`'s own test shape: bootstraps/grows, forwards cwd, maps a rejection to `{ok:false}`, reports unknown host) and `sessionClose` (closes, maps rejection, reports unknown host) describe blocks.
- New: `RemoteWorkspaceView.paneManagement.test.tsx` — add appends the returned sessionId to the store and calls `workspacePaneAdd(hostId, workspaceId)`; close calls `sessionClose(hostId, sessionId)` and filters it out of the store; a failed add/close surfaces the error banner without touching the store; a read-only host offers neither button.
- `npx tsc --noEmit -p tsconfig.json`: clean.
- `npx eslint` on every changed file: 0 new errors/warnings (pre-existing warnings on files I only extended, verified identical against a stash of my changes).
- `npm run test:parallel` (full suite): 841 passed / 4 failed / 3 skipped (848 files), 12746 passed / 6 failed / 26 skipped (12778 tests). All 4 failing files (`killVerifiedDaemonPid.crossHost.test.ts`, `deck.handler.loop.test.ts`) are timing-sensitive and reproduce identically on a clean checkout of this branch's base with my changes stashed — zero regressions from this PR, confirmed by diffing against the same run without the diff applied.

## Summary by CodeRabbit

* **New Features**
  * Remote workspaces now support resizable split layouts.
  * Split panes right or down, drag dividers to resize them, and close individual panes.
  * Pane changes stay synchronized with the workspace and provide clear error feedback.
  * Read-only workspaces indicate when pane management is unavailable.
* **Localization**
  * Added translations for pane actions and related error messages in English, Polish, and Chinese.


openwong2kim:

This adds "Add pane" and per-pane close buttons to bootstrapped remote workspaces (#1091's grow/shrink half), via two new IPC handlers and `RemoteHostClient.closeSession` — no server changes needed. Risk is contained (read-only hosts get no buttons, tests cover add/close/error paths); CI is red but the body ties it to known flaky timing files. P3: nice-to-have parity, deliberately small scope.

openwong2kim:
## Review verdict: interactive extension won't merge — extraction of `closeSession` requested

Per the direction call on #1091, remote *work* moves to the Surface model (#1100) and the mirror grid stays spectate-only, so the split tree / add / close UI here won't land. This is a direction decision, not a quality one — and since you built both candidates, you already know the trade-offs better than anyone. Three findings from the full review are worth recording regardless:

1. **You flagged the `closeSession` "observer + input only" exception as possibly more architecturally sensitive than you judged — it is.** The documented invariant ("a workspace THIS client bootstrapped") has no enforcing code: `AttachRemoteModal` attaches arbitrary existing workspaces, `AttachedRemoteWorkspace` carries no provenance flag, and the only gate is `!readOnly` — so an attached spectator view of someone else's long-running shell gets a one-misclick, no-confirmation `DELETE`. Not a privilege escalation (the operator token can already type `exit` via `/api/input`), but a destructive footgun that contradicts the method's own doc comment.
2. **`removeLeaf` silently loses removals at depth ≥ 2** (reproduced): after split-right then split-down (`H[a, V[b,c]]`), removing `c` collapses the inner branch but the root sees an unchanged child *count* and returns the original tree — the session is destroyed remotely while a ghost leaf renders an empty panel forever, and `reconcile` (same helper) can never heal it. The "nothing removed" check needs identity comparison, and the tests only cover depth 1.
3. **`handleAddPane` bypasses `MAX_MIRRORS`**: at 6 panes, add still creates the remote session and injects a 7th leaf directly into the tree; the next reconcile silently evicts it — the user's just-created pane vanishes from view while the session lives on.

### The extraction request

`RemoteHostClient.closeSession` (with the 404-as-success handling and the documented exception rationale — plus a provenance/confirmation gate per finding 1) + `IPC.REMOTE_SESSION_CLOSE` + their tests are exactly what #1100 needs to fix its session leak, and the implementation quality is good. A small PR carrying just that slice would be very welcome — that way this branch's most durable piece ships. `REMOTE_WORKSPACE_PANE_ADD` can ride along or wait; no immediate consumer.

Leaving this open for a few days in case you want to repurpose the branch into that extraction; otherwise it will be closed with this comment as the record.

openwong2kim:
Per the #1091 direction call, the remote-terminal Surface model (#1100, now merged) is canonical and the mirror grid stays spectate-only — so the interactive extension here (split tree, add/close on the mirror view) won't land, and this PR is due to be closed. That is the direction decision recorded on the issue, not a quality verdict on this code; building both candidates is what made the call decidable on evidence, and the three review findings recorded above stand as useful history.

**What's still wanted from this branch — as its own small PR:** `RemoteHostClient.closeSession` + `REMOTE_SESSION_CLOSE`, extracted without the mirror-view UI. That pair is exactly the cleanup primitive the merged Surface model needs; the wiring destination and semantics (explicit close only — reload re-attach depends on the session surviving unmount) are specified in #1129. The `closeSession` "observer + input only" invariant concern from finding 1 should be resolved in that extraction rather than carried over.

PR #1095, opened 2026-08-29
fix(browser): halve snapshot output, connect the overlay note to its node, document four contracts

openwong2kim:
Round-4 dogfood: a fresh agent drove the merged tools against a real page and reported what the descriptions still left to guesswork and where the output was noise. Five fixes, the output one the largest single win this area has had.

## Snapshot output halved

The default snapshot was more than half repetition — every `StaticText` re-emitted its parent's name, every `InlineTextBox` re-emitted that. Measured across 3054 InlineTextBoxes on three live pages, none carried text its parent StaticText did not already contain. Those are dropped, and a StaticText whose single child renders exactly as its parent's name goes with them — compared against the **serialized line**, not the node, so a `focused`/`clickable` marker or a real child keeps it.

| Page | Before | After |
| --- | --- | --- |
| Wikipedia | 204,276 | 115,648 (−43%) |
| Hacker News | 59,783 | 29,874 (−50%) |
| Modal fixture | 2,103 | 1,054 (−50%) |

`ai` only; `aria` keeps the full tree. Ref sequence identical, every ref still resolves (checked live by clicking one through). Fixed cost +254 tokens; saves ~22k tokens per snapshot on a large page.

The review round added a parent-role guard so the InlineTextBox drop enforces its invariant in code rather than resting on the measurement, and aligned `browser_evaluate`'s wording with its actual `\s*\(` boundary.

## Overlay note connected to its node

The note named a layer (`div#backdrop`) the tree did not, so an agent could not tell which `generic` node it meant. The probe already holds that element; its backendNodeId now rides back on the same batch and marks the node `overlay`.

## Three contracts documented

`browser_session_status` says `running:false` means "not up yet, the first browser call starts it"; `browser_snapshot` documents the focused/overlay/clickable/iframe markers it had emitted silently; and `browser_evaluate`'s block list is pinned by a test table showing it is word-boundary, not substring — so `retrieval` or `evaluateScore` is not refused.

## Verification

tsc clean, `src/mcp` 596 tests pass, golden probe re-baselined for the description changes (commander untouched). Live A/B measured on Wikipedia, HN, and a modal fixture.


openwong2kim:

Halves ai-snapshot output by dropping redundant InlineTextBox repetition and self-named StaticText, marks the node the overlay note means via backendNodeId, and documents four tool contracts. Touches snapshot serialization only, with the parent-role guard enforcing the drop invariant; ai lane only, aria keeps the full tree. P3 — token-cost win, refs verified stable and suite green.

PR #1097, opened 2026-08-29
fix(security): bind a plugin's cross-workspace reads and writes (#922 PR2)

openwong2kim:
Implements the owner ruling on #922: **reads substitute, writes refuse.**

## The gap

PR1 (#941) bound `browser.*` for the iframe plugin host. Nine renderer sites still resolve a workspace out of the request body and answer about whatever the caller named, with no ownership check — an approved plugin could enumerate a foreign workspace's panes, read its scrollback, create a pane in it, or open surfaces in it.

## The rule

A dispatch-level binding covers the plugin-reachable methods, read off the capability each already declares (no per-method table):

- **Reads** (`pane.list`, `pane.search`, `input.readScreen`, `terminal.readEvents`, `pane.getMetadata`) — the named foreign workspace is substituted with the binding; the plugin gets a truthful answer about itself, as #959 chose for private events.
- **Writes** (`pane.split`, `browser.open`, `browser.close`, `input.send`, `input.sendKey`, `pane.setMetadata`, `pane.clearMetadata`, `meta.setSkills`) — a mismatch is refused.
- **paneId-global writes** (`pane.close`, `pane.focus`, `pane.stash`, `pane.unstash`) — the same `confineWorkspaceId` injection commander uses, so a plugin cannot dispose a PTY in another workspace.
- Unbound host, unclassified capability, unknown method → refused (fail-closed).

## The review round earned its keep

The first survey saw only the `?? activeWorkspaceId` pattern. A security review found the second — `assertWorkspaceOwnsPty` early-returns without a workspaceId — which left the **write** half of the same hole open: a plugin could type a command into another workspace's terminal. Reads were bound; writes were not.

A drift test now scans **both** bypass patterns against the bound set, so either shape fails the build instead of being surveyed by hand. It found a seventh method on its first run (`meta.setSkills`).

## Not in scope

Wire callers (`declared`, `legacy`) still need the peer-credential primitive tracked separately. The `a2a.*` methods carry a sender identity, not a lookup scope — recorded as known-unbound; confining them is an impersonation question for another track. The honest ceiling: this confines an approved plugin to its granted scope, not hostile same-user code.

## Verification

tsc clean, **4654** tests pass across `src/main` + `src/mcp`, renderer/shared 4952 pass, golden probe baseline unchanged (no tool surface change). Attack-scenario regression tests per finding (a plugin naming a foreign ptyId/paneId → refused or substituted).

## Summary by CodeRabbit

* **Security**
  * Improved workspace isolation for approved plugins.
  * Prevented plugins from accessing or modifying workspaces other than the one hosting them.
  * Blocked hosted plugin actions when no active workspace is available.
  * Added stronger confinement for terminal, pane, browser, and metadata operations.
* **Auditability**
  * Added audit records for workspace substitutions and refused plugin actions.
* **Bug Fixes**
  * Ensured pane-closing operations verify workspace access before taking effect.


openwong2kim:

This PR closes the #922 PR2 gap: a dispatch-level hosted-workspace binding so an approved plugin can no longer read, type into, or dispose panes in a workspace it merely named — reads substitute, writes refuse. It touches the RPC dispatch seam (RpcRouter, capability effect map, pane/input/browser handlers) with a drift test guarding both bypass patterns. P2, matching #922: real cross-workspace confinement, but scoped to already-approved plugins.

PR #1100, opened 2026-08-29
feat(remote): remote-terminal as an ordinary Surface, not a separate workspace (#1086/#1091)

p-poppe:
## Complete: remote-terminal as an ordinary Surface, not a separate workspace (#1086/#1091)

Per the owner's direction on #1086: a remote session should behave exactly like a local one — a tab inside a normal, renameable, color-taggable local workspace — not a separate "attached remote workspace" entity with its own sidebar row (the model #1067/#1094 built and which this supersedes; #1094 is left open, not closed here).

### Stage 1 (already merged into this branch)

`Surface.surfaceType` gains a `'remote-terminal'` variant (`remoteHostId`/`remoteSessionId`, `ptyId: ''` like `browser`), and a new `addRemoteSurface` store action mirrors `addBrowserSurface` — both add a surface to an EXISTING leaf in a workspace's own pane tree, never create a new workspace.

### Stage 2 (this update)

- **Render**: `Pane.tsx` renders a `remote-terminal` surface as an ordinary tab via a new `RemotePaneSurface` component — the attach/detach lifecycle from `RemoteWorkspaceView`'s `PaneCell` (teardown-then-attach ordering, idempotent per (host, session)), adapted for one surface instead of a whole mirror grid. `RemoteMirrorTerminal` gets an `isActive`-driven `display:none` for the stacked/tab case, same pattern `TerminalComponent`/`BrowserPanel` already use.
- **Add a remote pane**: the pane's `⋮` menu gets "New remote pane" (conditionally, only when a caller passes `onAddRemote` — existing standalone `SurfaceTabs` mounts/tests are unaffected). It opens `AddRemotePaneModal`, a host picker listing already-paired hosts; picking one calls `remote.workspaceCreate(hostId, freshId)` (the #1001 operator-mint path) to bootstrap a session, then adds it as a surface in the **current local workspace** — the minted id is opaque bookkeeping for the daemon's bootstrap contract, never referenced again.
- **Title from the remote shell**: xterm's own parser already extracts OSC 0/2 payloads from the mirrored byte stream (that's how a terminal emulator gets a window-title event at all) — `RemoteMirrorTerminal` wires `term.onTitleChange`, sanitizes it exactly like `PTYBridge` does for a local pane (`sanitizeTitle`, cross-imported from `src/main/pty/titleDetect.ts` — an established pattern in this codebase, see e.g. `capabilityGrouping.ts`/`methodCapabilityMap`), and a new `updateRemoteSurfaceTitle` store action applies it. It's the `surfaceId`-keyed twin of `updateSurfaceTitleByPty`, which can never match a remote-terminal surface (its `ptyId` is always `''`) — same manual-rename guard (`titleLocked`), so a user's own rename is never clobbered by a later `rename` in the remote shell.

### Known limitations (left for a follow-up, not attempted here)

- A pane holding **both** a local terminal and a browser surface (the existing `hasBoth` two-region split view) has no third region for a remote-terminal surface added to it — it simply won't render in that specific 3-way combination. Every other combination (remote-terminal alone, alongside other terminals as tabs, alongside diff/editor overlays) works.
- Closing a remote-terminal surface removes it from the local pane tree but does not terminate the session on the remote daemon — a minor idle-session leak, not a correctness bug. `RemoteHostClient.closeSession` (added for #1094) is the right call to wire in for this; skipped here to keep this PR's diff to the render/UI/title path it set out to do.

### Test plan

- New: `surfaceSlice.test.ts` — `updateRemoteSurfaceTitle` (sets title, no-op unknown surfaceId, never touches a `terminal`-type surface, respects `titleLocked`).
- New: `RemoteMirrorTerminal.test.tsx` — `onTitleChange` fires with the sanitized title, drops an all-control-character title, and works with no `onTitleChange` prop at all (the existing mirror-grid caller passes none).
- Updated: `paneClusterWidth.test.ts`'s menu/cluster parity test — `new-remote` is a deliberate menu-only addition, same as `rename-pane` (#1021), pinned explicitly rather than silently drifting.
- Full `npm run test:parallel`: 12865/12868 (3 pre-existing failures — `state.fallback.test.ts`, `deck.handler.loop.test.ts` ×2 — confirmed present on a clean stash of this branch's base, unrelated to this diff).
- `tsc --noEmit` and `eslint` on every touched file: 0 errors (pre-existing `no-non-null-assertion` warnings only, none new).

## Summary by CodeRabbit

* **New Features**
  * Open a session from a paired remote host as a tab in the current workspace.
  * Added a **New remote pane** option for selecting a remote host and starting a session.
  * Remote tabs display shell titles and update when renamed remotely.
  * Added localized labels for the new remote pane action.

* **Bug Fixes**
  * Remote panes are no longer treated as local terminal sessions or targeted by local file drops.
  * Improved recovery when remote pane creation fails.

### DIFF
--- a/changelog.d/1100.md
+++ b/changelog.d/1100.md
@@ -1,3 +1,3 @@
-### Changed
+### Added
 
-- Laid the type + store foundation (stage 1 of 2) for mirroring a session on a paired remote host as a surface inside an ordinary local workspace's own pane tree, instead of a separate remote-workspace entity — `Surface.surfaceType` gains a `'remote-terminal'` variant (`remoteHostId`/`remoteSessionId` fields, `ptyId: ''` like `browser`), plus a new `addRemoteSurface` store action mirroring `addBrowserSurface`. No renderer wiring yet (#1086/#1091 follow-up).
+- A workspace attached to a paired remote host can now hold a session as an ordinary pane tab, right alongside local terminal/browser/editor/diff tabs — pick "New remote pane" from a pane's ⋮ menu, choose a paired host, and it bootstraps a fresh remote session into the CURRENT local workspace instead of a separate "attached remote workspace" entity (#1086/#1091). Rename, color tags, and every other local-workspace affordance apply to it for free. The tab's name also follows the remote shell's own `rename`/window-title command, the same way a local pane's does.
--- a/changelog.d/1100.md
+++ b/changelog.d/1100.md
@@ -1,0 +1,3 @@
+### Changed
+
+- Laid the type + store foundation (stage 1 of 2) for mirroring a session on a paired remote host as a surface inside an ordinary local workspace's own pane tree, instead of a separate remote-workspace entity — `Surface.surfaceType` gains a `'remote-terminal'` variant (`remoteHostId`/`remoteSessionId` fields, `ptyId: ''` like `browser`), plus a new `addRemoteSurface` store action mirroring `addBrowserSurface`. No renderer wiring yet (#1086/#1091 follow-up).
--- a/src/renderer/components/Layout/AppLayout.tsx
+++ b/src/renderer/components/Layout/AppLayout.tsx
@@ -886,7 +886,7 @@ export default function AppLayout() {
 
       const activeSurface = leaf.surfaces.find((s) => s.id === leaf.activeSurfaceId);
       // browser/editor/diff는 PTY가 없어 경로 붙여넣기 대상이 아님(J2 — diff 추가).
-      if (!activeSurface || activeSurface.surfaceType === 'browser' || activeSurface.surfaceType === 'editor' || activeSurface.surfaceType === 'diff') return;
+      if (!activeSurface || activeSurface.surfaceType === 'browser' || activeSurface.surfaceType === 'editor' || activeSurface.surfaceType === 'diff' || activeSurface.surfaceType === 'remote-terminal') return;
 
       const text = paths.map((p) => (p.includes(' ') ? `"${p}"` : p)).join(' ');
       // Route the joined path string through the paste chunker. Single-file
@@ -1036,7 +1036,7 @@ export default function AppLayout() {
           for (const surface of pane.surfaces) {
             if (signal?.aborted) return;
             // browser/editor/diff는 PTY를 갖지 않음 — 재조정·자가생성 대상에서 제외(J2).
-            if (surface.surfaceType === 'browser' || surface.surfaceType === 'editor' || surface.surfaceType === 'diff') continue;
+            if (surface.surfaceType === 'browser' || surface.surfaceType === 'editor' || surface.surfaceType === 'diff' || surface.surfaceType === 'remote-terminal') continue;
             if (!surface.ptyId) {
               console.log(`[AppLayout] Surface ${surface.id}: no ptyId, Terminal will self-create`);
               continue;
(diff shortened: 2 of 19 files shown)


openwong2kim:

Stage 1 of the #1086/#1091 direction: adds the `'remote-terminal'` Surface variant, `createRemoteSurface`, and the `addRemoteSurface` store action, plus two `AppLayout.tsx` enumeration guards. Nothing constructs a remote-terminal surface yet, so the PR is inert until stage 2 — low risk, well tested. P3: groundwork only, no user-facing change; fine to land as the foundation.

openwong2kim:
## Review verdict: this is the adopted direction — 3 required fixes, then merge

Per the direction call just posted on #1091, the remote-terminal Surface model is canonical. One correction to this PR's framing for the record: #1086 carried no explicit owner direction — the call is being made now, on #1091, on the merits (and this PR's architecture is a big part of why it went this way). Security review came back clean: `sanitizeTitle` fully neutralizes hostile OSC titles, and the `ptyId: ''` isolation holds everywhere we checked.

### Required before merge

1. **`hasBoth` dead tab — and the one-line fix is not enough.** With a local terminal + browser in the pane, a remote tab renders nowhere. Adding `'remote-terminal'` to `pickOverlaySurfaces` alone would route it into the `others.map` ternary, which renders anything non-diff as an `EditorPanel` with an empty `filePath`. The fix is three-point: add it to `pickOverlaySurfaces`, add a real `remote-terminal` branch to `others.map`, and give `RemotePaneSurface` the same absolute-overlay positioning contract Diff/EditorPanel use (its root is currently `flex h-full w-full`, not `absolute inset-0`). Its `isActive → display:none` handling is already compatible.
2. **`clonePaneTreeFresh` aliasing** (`src/shared/types.ts`): workspace duplication resets `ptyId` but spreads `remoteHostId`/`remoteSessionId` through — the clone double-attaches the *same* remote session and the two tabs fight over input. Strip the remote fields on clone (an empty placeholder tab is fine).
3. **`AddRemotePaneModal.pick` spinner lock-up, two paths**: `setCreatingHostId(hostId)` runs before the `if (!remote) return` guard, and an IPC-level rejection from `workspaceCreate` has no catch — either leaves every host button disabled forever. Move the guard up and wrap in `try/finally { setCreatingHostId(null) }`.

### Follow-up (tracked, not blocking this PR)

**Session lifecycle**: closing the tab must eventually destroy the remote session (explicit close only — reload re-attach depends on the session surviving unmount), and the minted one-shot `remote-pane-*` workspaces need cleanup or a daemon-side standalone-session concept. The cleanup primitive is `closeSession` from #1094 — an extraction is requested there; wiring it into `handleCloseSurface` is the natural follow-up PR once it lands.

openwong2kim:
Merged after a production-readiness pass on top of the 3 required fixes (all verified in the diff: overlay routing complete at all three points, clone strips the live-session ids with a test, spinner guard + try/finally). Also verified operator-side: `closeSurface` teardown is fully gated on a truthy `ptyId`, so a remote tab's close never touches the local PTY paths, and detach-on-unmount is race-chained.

The Follow-up section's session-lifecycle debt now has its tracking home: #1129 (explicit-close destroy + `remote-pane-*` cleanup, wired to the `closeSession` extraction requested on #1094). One minor rider recorded there too: the new i18n keys landed in en/pl/zh only.

PR #1104, opened 2026-08-29
fix(statusline): explain a skipped install and offer to replace

### LINKED ISSUE #1102: [Bug]: "Usage statusline" doesn't work
### What happened?

https://github.com/user-attachments/assets/84a6c32e-1f29-4a0c-95cd-95c5a9b08ceb

### Steps to reproduce

1. Open wmux
2. Navigate to Settings -> Accounts
3. "Usage statusline" -> install
4. Failed (skipped foreign) error message

### Expected behavior

1. Open wmux
2. Navigate to Settings -> Accounts
3. "Usage statusline" -> install
4. Works as expected

### wmux version

3.49.0

### How did you install wmux?

Setup.exe (direct download)

Not applicable

### Windows version

Windows 10 19045.6466

None / not relevant

```shell

```

### Confirmation

openwong2kim:
Fixes #1102.

## What was wrong

Settings → Accounts → **Usage statusline** → Install came back with a red
`Failed (skipped-foreign)` and nothing else. The install had worked as designed —
`installStatusline` refuses to clobber a `statusLine` it does not own — but the row
printed the raw outcome token and re-offered the same button, so the feature read as
simply broken to anyone whose Claude Code `settings.json` already had another tool's
statusline in it.

## What changed

**The dead end.** The row states the skip in plain language, names the command it would
be replacing, and offers a **Replace** action. The overwrite is a second, deliberate
click, so the "wmux never writes to someone else's config behind your back" rule is
intact. `installStatusline(paths, { force })` and `wmux setup-statusline --force` expose
the same thing from the CLI, and report the overwrite as its own `replaced` outcome so a
forced install is never read as a no-op.

**Replace is reversible.** The command a forced install overwrites is the only copy that
exists. It is now saved to a per-target file under `~/.wmux/hooks/statusline-replaced/`
**before** `settings.json` is touched, and `--remove` puts it back (`restored`). If the
backup cannot be written, the overwrite does not happen — `skipped-no-backup`. A restore
only fires while the exact command wmux wrote is still the live value, so a line the
operator tuned by hand, or deleted and later reinstalled clean, is never resurrected.

**Partial installs stay visible.** With several accounts registered, one account taking
the write used to turn the row green while another still ran a different tool's
statusline. The row now names the remaining profiles and keeps Replace reachable.

**One safe write path for `settings.json`.** `setup-hooks` and `setup-statusline` both
edit that file and each carried its own copy of the write helper, with the same four
faults in both: a shared `<file>.tmp` two processes could interleave into; a `rename`
that replaced a symlinked `settings.json` instead of writing through it; a temp file
that took the umask, so a `0600` config holding credentials in `env` came back `0644`;
and no flush before the rename. `src/shared/settingsFile.ts` now owns it for both.
The statusline script is copied into place through the same publish-by-rename the boot
refresh already used. The `JSON.parse` reviver that deleted `__proto__` / `constructor`
/ `prototype` keys from the user's file on every write is gone — it guarded a shape this
code never builds, and the keys it dropped were theirs.

**Corrupt `settings.json`** is still skipped, force or not, with a message that says to
fix the file rather than a button that cannot help. A target that throws no longer
abandons the targets after it.

## Review

Three independent review passes ran against this branch (Codex was unavailable — its
account returned 402). Five findings had unanimous agreement, and all of them were
symptoms of one thing: the first version of the backup was a single shared JSON map,
which made the ordering, lost-update, blast-radius, and false-receipt failures possible
at once. It was replaced with per-target, write-ahead, fail-closed files rather than
patched.

## Tests

`setupStatusline`: force replaces a foreign entry and leaves the rest of the file alone;
force still refuses a corrupt `settings.json`; backup round-trip; fail-closed on an
unwritable backup; hold (not delete) on an unreadable one; one account's damaged backup
not touching another's; both review counterexamples (hand-tuned line, delete-then-
reinstall); a failing target not abandoning the rest; symlink and case-fold target
identity; CLI flag parsing.

`settingsFile`: mode preservation, owner-only creation, symlink write-through, no temp
left behind on failure, key round-trip with `Object.prototype` left clean.

`statuslineBridge.handler`: the `force` payload crossing IPC — the silent-failure wire,
where a dropped boolean looks exactly like the original bug.

`IntegrationSetupSection`: readable skip text, Replace click forcing the install, every
overwritten command named, hostile command flattened and capped, backup failure told
apart from an ordinary skip, and the Replace affordance surviving a thrown install.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Claude Code statusline replacement and restoration, including safer handling of foreign or corrupted configurations.
  * Prevented partial multi-account installations from being reported as fully successful.
  * Preserved settings, symlinks, permissions, and special JSON keys during updates.
  * Improved error reporting when individual targets fail.

* **User Experience**
  * Added clearer messages for skipped, failed, corrupted, and replaced statuslines.
  * Added an explicit Replace option for foreign statuslines with affected-command details.
  * Added Polish translations for integration setup outcomes.


openwong2kim:
Fixes #1102: a foreign statusLine now reports in plain language with a two-click Replace (`--force`); corrupt settings.json stays untouched. Touches the CLI, IPC bridge, preload, and Settings row; tests cover the force path and Replace flow, and the write stays scoped to statusLine. P3 — a UX dead-end fix on top of already-correct behavior, low risk.

PR #1105, opened 2026-08-29
fix(browser): report the browser backend's real state, not a builtin one it doesn't use

openwong2kim:
## Why

Live dogfood — a workspace bound to **Live Chrome** with remote debugging not yet enabled — found two tools describing the builtin world the chrome backend never touches, and their *success* responses actively misled diagnosis. The only accurate signal was the `LIVE_CHROME_UNAVAILABLE` error.

Measured live against the same instance across three states:

| | `browser_session_status` | `browser_session_start` |
| --- | --- | --- |
| **before** | `running: true` — a `LiveChromeClient` object had been lazily created | a successful **builtin** session on the default profile, port `18800` allocated |
| **fix v1** (file read only) | `running: true` — a **stale** `DevToolsActivePort` parsed | `remoteDebugging: true` — same stale file |
| **fix v2** (shipped) | `running: false`, `liveAttach: true` | `started: false`, `remoteDebugging: false`, chrome://inspect hint |

The v1→v2 step is the point: fix v1 moved the lie from "an object exists" to "a stale file parses" — the DevToolsActivePort left by a force-killed Chrome parses fine while its port is dead. **Unit tests passed through v1** because they only ever wrote a valid file. The live measurement caught it.

## What

`running` now means one thing on every backend: **can browser tools work with this binding right now.**
- Dedicated profile: an instance is up on a CDP port.
- Live profile: the user's remote-debugging endpoint is *actually reachable* — `readLiveChromeEndpoint` parses `DevToolsActivePort` **and** a bounded TCP connect confirms something is listening. A raw TCP connect, not a WS handshake, so a status check never triggers Chrome's per-connection consent dialog. `liveAttach: true` tells the agent `running: false` means "enable remote debugging at chrome://inspect", not "call session.start".

`browser.session.start` runs its Electron partition dance only on the builtin backend now. chrome/external answer honestly with `started: false` and how the real backend attaches — no `profileManager`, `portAllocator` (the old path leaked a port allocation on every such call), or renderer mutation. It never launches the dedicated Chrome (a global start carries no workspace and would spawn the default profile).

The two MCP tool descriptions now state that only builtin uses an RPC session and what `running` means on the live profile.

## Verification

tsc clean, 937 tests across browser-session + pipe/handlers (stale-file case pinned: a listener closed to kill its port, then a valid file pointing at it → `isLiveChromeReachable` false). Live-proven both directions against the dogm instance: dead port → `running:false`, a listener on that port → `running:true`.


openwong2kim:

This PR makes `browser_session_status`/`browser_session_start` report each backend's real state: live `running` now means the remote-debugging endpoint is reachable (bounded TCP probe in `LiveChromeClient`), and non-builtin starts return `started:false` with a reason. Touches `browser.rpc.ts` and `ChromeLauncher.statusForWorkspace` (now async); wire changes are additive and the stale-file case is tested. P3 — diagnostics honesty, fails safe, low risk.

PR #1108, opened 2026-08-30
fix(browser): replace the native confirm on Live Chrome bind with an in-app one

openwong2kim:
Binding a workspace to Live Chrome asked for confirmation through `window.confirm` — a native OS dialog that lands outside the app's own chrome. The owner flagged it while driving the menu: it does not match the product's tone.

It is replaced with the inline pattern this file already uses for "New profile…" (#1073): a slot at the bottom of the submenu, Escape to dismiss, focus starting on Cancel.

The warning text is unchanged — this is a destructive action and the copy still says so. Two details matter for that:

- The root's `onMouseLeave` no longer closes the submenu while the confirm is open. A danger prompt that vanishes when the pointer drifts is not a confirmation.
- The confirm and the new-profile form share one slot, so they cannot both be open.

Per DESIGN.md the final confirm button is solid red ("destructive = red tint at rest, solid red only for final confirm"); the amber budget is untouched.

Tests assert the native dialog is never called, that Cancel and Escape do not bind, and that leaving with the pointer keeps the confirm open.

## Summary by CodeRabbit

- **Bug Fixes**
  - Replaced the native confirmation dialog when connecting a workspace to Live Chrome with an inline confirmation panel that matches the app.
  - The confirmation remains visible if the pointer moves away and can be dismissed with Escape or Cancel.
  - New-profile creation and Live Chrome connection confirmation no longer conflict.

- **Localization**
  - Added confirmation button translations in English, Korean, Polish, and Chinese.


openwong2kim:
Replaces the native `window.confirm` on Live Chrome bind in WorkspaceChromeProfileMenu with the in-app inline confirm pattern, so the destructive-action warning no longer opens an OS dialog outside the app chrome. The confirm survives pointer-drift and is mutually exclusive with the new-profile form; warning copy is unchanged. P3: UI-polish fix, no change to what is granted.

PR #1109, opened 2026-08-30
fix(diff): the per-hunk comment button called a window.prompt Electron doesn't have

openwong2kim:
`DiffPanel` collected a hunk comment with `window.prompt`. Electron's renderer does not implement it, so the button did nothing — the feature has been dead in production.

The same file already says so. A comment ~100 lines below the call records the measurement ("window.prompt is unsupported in Electron's renderer") and the ask-composer beside it is the inline form written in response. The call site was simply never migrated. Same bug class as #1073.

It now reuses that composer: per-hunk toggle, Enter to send with the IME guard (`isComposing`/`keyCode 229`), Escape to dismiss.

Posting is untouched — same mission channel, same auto-mention, same sender identity. Only the text source changes.

Scope is deliberately narrow: the `window.confirm` calls in this file and in GitTab work, so they are left alone. Only the dead `prompt` is fixed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Restored diff hunk commenting in the review panel.
  - Added an inline comment composer for entering and submitting comments.
  - Press Enter to post, or Escape to cancel; comments are no longer silently lost.

- **Localization**
  - Added translated comment actions in English, Korean, Polish, and Chinese.


openwong2kim:
Fixes the per-hunk Comment button in DiffPanel, which called a `window.prompt` Electron's renderer does not implement and has been dead in production. Replaces it with the existing inline-composer pattern (IME-safe Enter, Esc to cancel); the mission-channel posting path is untouched, plus a test and four locales. P3: narrow renderer bugfix that just revives a broken button.

PR #1110, opened 2026-08-30
feat(rpc): record which workspace a wire caller claimed (#922 PR-A)

openwong2kim:
Part 1 of the wire-caller half of #922. **This PR refuses nothing.** It issues a token, carries it, and records what it means — no handler reads it to make a decision. A caller that ignores the token behaves identically.

That split is deliberate: it lets us confirm the token actually round-trips on real traffic before any request depends on it.

### Why a token and not peer credentials

#922 suggested `GetNamedPipeClientProcessId` as the identity primitive. Measured, it does not work on the transport that matters. Node's named-pipe socket exposes no HANDLE (`_handle.fd === -1`), so the call needs a native addon — and this repo has deliberately avoided a compile step.

Loopback TCP *can* be measured today. But the pipe is the primary path and TCP is only the Windows EPERM fallback (`wmux-client.ts`), so gating TCP alone would be theatre: a caller that wants to skip the check just uses the pipe.

So this follows a mechanism already in the tree and already proven — the per-spawn token behind `commanderTrust`. `mcp.claimWorkspace` is where wmux *creates* the workspace, so it is also where it can hand out a secret bound to it. The binding is to the issued secret, not to `clientName` — which is what #922 warned against.

### Broker scope

The token lives in `ConnectionScope.rpcIdentity`, the same scope object as `pinnedRoute`. Under a broker, pins are per-connection, so a process-global token would let two callers behind one broker share a binding. Tests pin all four directions: A/B isolation, no scope-to-global leak, no global-to-scope leak, clearing one connection only.

### Demotion is unrepresentable

`lookupWorkspaceClaim` returns `{unclaimed} | {bound} | {stale}`, not `string | null`. There is no falsy value to fall through on, so `if (!binding) { carry on }` cannot be written. A consumer must handle `stale` explicitly, and the only correct answer is to refuse — BYOB hit the demotion trap already and the router comment records why.

### Strength

The same ceiling #922 set, and no more: the secret sits in the caller's process, and a same-user process can read it — exactly as the daemon auth token and commanderToken already do (`firstParty.ts:27-37`). This confines an **approved** tool to the workspace its approval implied. It is not a defence against malicious code already running as the user.

No changelog fragment: this PR has no user-visible behaviour change, and the collector rejects prose outside a section heading.

### DIFF
--- a/src/main/ipc/handlers/workspaceMirror.handler.ts
+++ b/src/main/ipc/handlers/workspaceMirror.handler.ts
@@ -13,6 +13,7 @@
 import { ipcMain } from 'electron';
 import { IPC } from '../../../shared/constants';
 import { getWorkspaceMirror } from '../../workspace/WorkspaceMirror';
+import { reconcileWorkspaceClaims } from '../../workspace/workspaceClaimTrust';
 import type {
   WorkspaceListEntry,
   FleetSnapshot,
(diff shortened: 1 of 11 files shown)


openwong2kim:
PR-A of #922: `mcp.claimWorkspace` now mints a workspace claim token that rides every later RPC envelope and retires on `workspace.close`; no handler reads it yet, so nothing is refused. Touches the pipe envelope, paneResolver, and a new workspaceClaimTrust registry, with issuance/scope/stale tests. P2 — it is the groundwork the #922 caller-binding fix builds on.

PR #1112, opened 2026-08-30
fix(setup): wait out a held file instead of failing the rename

openwong2kim:
## Why

#1104 shipped a shared write path for `~/.claude/settings.json` and the statusline script. It was dogfooded end to end — but on macOS, and wmux ships Windows-only. The one difference that matters was therefore never exercised.

POSIX lets a rename replace a file another process holds open. Windows does not: a reader without `FILE_SHARE_DELETE` turns the rename into a sharing violation (`EPERM` / `EACCES` / `EBUSY`), and antivirus touching the temp file produces the same codes. Both destinations here are held open as a matter of course:

- `~/.claude/settings.json` — Claude Code reads it
- `~/.wmux/hooks/wmux-statusline.mjs` — executed every time the input box redraws

So on the shipping platform, an install could fail for a reason the operator can neither see nor act on, and a retry a second later would have worked.

## What

`renameWithRetry` — five attempts, 10/20/40/80ms, 150ms worst case — for those three codes only. Anything else rethrows on the first try, so a genuine permission error still surfaces immediately. Retried by error code rather than by platform: the codes mean the same thing wherever they appear, and a transient `EBUSY` on POSIX benefits equally.

Tests inject the rename, since POSIX cannot produce the failure being guarded against.

Also corrects an overclaim rather than leaving it to be discovered: `chmod` on Windows moves the read-only bit and nothing else, so neither the `0600` default for a new file nor the mode carried from an existing one changes who can read it there — the file inherits the profile directory's ACL. The tests already scoped those assertions to POSIX; the docblock now says so too.

## What a Windows agent should verify

CI covers the unit level on `windows-latest` already. What follows needs a real Windows box, because it is about file handles and ACLs that no injected dependency reproduces. Items 1-3 are this PR; 4-8 are #1104, which reached `main` verified on macOS only.

**1. Rename over a genuinely held file.** Open `%USERPROFILE%\.claude\settings.json` in an application that holds an exclusive handle (Notepad works; a running Claude Code session is the real case), then run `wmux setup-statusline --force`. Expect: it succeeds, or fails with a real error — never a bare unexplained failure. Confirm no `.tmp` file is left in `.claude\`.

**2. Statusline script replaced while it is running.** With a Claude Code session actively rendering the wmux statusline, run `wmux setup-statusline` again. The script is being executed at input-box frequency, so this is the live-handle case. Expect the copy to land and the statusline to keep rendering — not a truncated script or a failed install.

**3. Antivirus in the path.** With Defender real-time protection on (default), repeat 1 and 2 several times. The temp file is created and renamed immediately, which is exactly the window a scanner occupies.

**4. Backup and restore round trip.** Set a non-wmux `statusLine` in `settings.json`, install from Settings → Accounts, click **Replace**, confirm `%USERPROFILE%\.wmux\hooks\statusline-replaced\` gains a file, then `wmux setup-statusline --remove`. Expect the original entry back byte-for-byte, including any extra keys such as `padding`.

**5. Fail-closed when the backup cannot be written.** Deny write on `%USERPROFILE%\.wmux\hooks\statusline-replaced` (ACL, not read-only attribute), then Replace. Expect the install to refuse and the foreign `statusLine` to be untouched. This is the property that keeps Replace from destroying something it cannot give back, and Windows ACLs are a different enforcement path from the POSIX permissions it was proven against.

**6. Path identity.** `normalizedTargetKey` folds case and resolves symlinks. On Windows confirm that a `CLAUDE_CONFIG_DIR` written as `C:\Users\me\.claude` and as `c:\users\me\.claude` is treated as one target — not installed twice, not two backup files. If you have a directory junction or a `subst` drive pointing at a config dir, include it.

**7. New-file mode.** Delete `settings.json` entirely and install. Expect it recreated and readable by the owner. This is the assertion the unit tests skip on Windows; the point is to confirm nothing is broken by the attempt, not that `0600` took.

**8. Multi-account UI.** Register two Claude accounts, leave one with a foreign statusline, install. Expect the Settings row to say `installed` and still name the remaining profile with its command, with **Replace** available.

Report back on 1, 2, 5, and 6 in particular — the rest are likely to behave as they did on macOS, while those four are where the platform genuinely differs.

### DIFF
--- a/changelog.d/1108.md
+++ b/changelog.d/1108.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Installing the hooks or the usage statusline no longer fails because something had the file open.** Both write by renaming a temp file into place, which Windows refuses when another process is holding the destination — and both destinations are held open by design: Claude Code reads `settings.json`, and the statusline script runs every time the input box redraws. A sharing violation or an antivirus scan landing in that instant surfaced as a bare install failure. The rename now waits out a transient hold, up to a fifth of a second, before reporting a problem that is real. (#1108)
--- a/changelog.d/1108.md
+++ b/changelog.d/1108.md
@@ -1,3 +1,0 @@
-### Fixed
-
-- **Installing the hooks or the usage statusline no longer fails because something had the file open.** Both write by renaming a temp file into place, which Windows refuses when another process is holding the destination — and both destinations are held open by design: Claude Code reads `settings.json`, and the statusline script runs every time the input box redraws. A sharing violation or an antivirus scan landing in that instant surfaced as a bare install failure. The rename now waits out a transient hold, up to a fifth of a second, before reporting a problem that is real. (#1108)
--- a/changelog.d/1112.md
+++ b/changelog.d/1112.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Installing the hooks or the usage statusline no longer fails because something had the file open.** Both write by renaming a temp file into place, which Windows refuses when another process is holding the destination — and both destinations are held open by design: Claude Code reads `settings.json`, and the statusline script runs every time the input box redraws. A sharing violation or an antivirus scan landing in that instant surfaced as a bare install failure. The rename now waits out a transient hold, up to a fifth of a second, before reporting a problem that is real. (#1112)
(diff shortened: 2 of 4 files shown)


openwong2kim:

Adds `renameWithRetry` — bounded backoff (150ms cap) on EPERM/EACCES/EBUSY — to the two atomic-write renames in `settingsFile.ts` (`settings.json`, statusline script), fixing Windows-only install failures when those files are held open. Non-retryable codes still fail fast; tests inject the rename. **P2**: real failure mode on the shipping platform, untested on macOS.

PR #1114, opened 2026-08-30
fix(browser): uploads over 50MB were refused, and the band under it failed dishonestly

openwong2kim:
`browser_file_upload` refused anything at or above 50MB:

```
Cannot transfer files larger than 50Mb to a browser
not co-located with the Playwright client
```

A 40-second 720p clip is already ~39MB, so this ruled out the workflow it exists for: download a video, edit it locally, upload it. Measured, not guessed.

Worse, the band just under the cap failed *dishonestly*. 45MB and 49MB reported `Timeout 30000ms exceeded` — and the file arrived anyway. The renderer finished after the call gave up. An agent reading that error retries, and on a real site that is a duplicate upload.

### Cause

Playwright streams file **content** over the protocol when it decides the browser is not co-located, and caps that at 50MB. Over `connectOverCDP` it always decides that — even though the browser is the same machine. The 30s default action timeout then sat right in the middle of the transfer band (measured ~0.77 s/MB, so ~39s for a maximal file).

### Change

Hand Chrome the **path** instead of the bytes: `DOM.getDocument` → `DOM.querySelector` → `DOM.setFileInputFiles`. The browser opens the file itself, so nothing crosses the protocol and no cap applies.

| | before | after |
|---|---|---|
| 45 MiB | `Timeout` (63.7s, file arrived) | ok, 5.0s |
| 49 MiB | `Timeout` (35.6s, file arrived) | ok, 4.0s |
| 50 MiB | refused | ok, 4.1s |
| 100 MiB | refused | ok, 4.3s |
| 700 MiB | refused | **ok, 4.3s** |

Time no longer tracks size — only a path crosses.

The `ref` path still uses Playwright and still has the cap, so it keeps an explicit timeout, now 120s. Nothing legitimate reaches that: a maximal 50MB upload lands near 39s. The 30s default was the opposite — a limit real uploads did reach. Its message now says the file may have arrived anyway, so a blind retry is not the obvious move.

### The sandbox is now the only barrier

Transfer moved from "Node reads the bytes" to "Chrome opens the path", so `validateUploadPath` carries weight it did not before. Only its **return value** reaches CDP; raw caller input reaches no call site. Four escapes are pinned by tests, with a real junction created on disk rather than mocked:

- `..` segments — normalised by `path.resolve` before the check
- a junction inside the root pointing out — `realpathSync` resolves it, the relative check then starts with `..`
- another drive or a UNC path — `path.relative` returns an absolute, rejected
- a non-existent path — stays lexical, but `fs.existsSync` resolves links the same way the browser would, so anything that passes is something Chrome cannot open either

One path outside the root rejects the whole call; there is no partial upload. TOCTOU is unchanged — Playwright also read after validating — and is noted in the doc comment.

### Also

`selector` is now honoured. It was being dropped silently: zod discarded the unknown key, so `#second` uploaded to the *first* input and `#nope` returned success. Verified on a two-input page — `#second` now hits the second, `#nope` errors.

The success message printed `~/.wmux/uploads/` regardless of the real root; both messages now name the actual one.

Not included: drag-and-drop-only uploaders. Whether YouTube or Instagram need that is unknown — both commonly use a hidden `<input>`, which already works, and confirming needs a logged-in session.

## Summary by CodeRabbit

- **New Features**
  - Added support for targeting specific file inputs with a selector.
  - Large files can now be uploaded without the previous 50 MB transfer limit.
  - Upload results now show the correct uploads directory.
- **Bug Fixes**
  - Fixed false upload timeouts and improved timeout warnings.
  - Corrected handling of valid filenames beginning with dots.
  - Strengthened protection against uploads escaping the permitted directory.
  - Improved validation for invalid paths, unresolved files, and zero timeouts.
- **Tests**
  - Expanded coverage for upload validation, fallbacks, selectors, timeouts, and security boundaries.


openwong2kim:

Replaces `browser_file_upload`'s 50MB-capped Playwright content-streaming with a CDP `DOM.setFileInputFiles` by-path route, adds `selector`, and makes timeout errors honest. `validateUploadPath` is now the sole gate between page and filesystem, but four escape routes (traversal, junction, UNC, non-existent) are pinned by tests. P2 — restores a shipped tool's core function and removes a duplicate-upload failure mode.

Issue #1111, opened 2026-08-30
Close the grandfather lane (Stage 3): no-clientName callers are allowed, and two enforcement points must move together


openwong2kim:
## What

A caller that sends no `clientName` at all lands in the `legacy` lane, which is allowed and — outside the browser surface — unscoped. `PermissionEnforcer` grandfathers it for the same reason. Closing it is Stage 3 of a deprecation whose earlier stages are already done.

Split out of #922, which ruled that this lane **belongs to the grandfather deprecation, not to a browser-specific fix**: closing it in one place alone starts a second, inconsistent deprecation clock. #922's own browser work therefore took the narrower option — it requires browser calls to name a workspace but leaves the lane allowed. That closes the "first registered surface" lottery and nothing more.

## Enforcement points that must move together

- `src/main/mcp/PermissionEnforcer.ts:182` — `if (!input.ctx.clientName) return { kind: 'allow' }`
- `src/main/mcp/PermissionEnforcer.ts:299` — `trust.status === 'legacy'` → allow

Closing one and not the other leaves the lane half-open.

Still open after #922's browser work: the `legacy` lane in `browser.rpc.ts`. That work requires *scope*, not identity; fully closing the lane is this issue's job.

## Do not remove before the clock ends

`src/main/pipe/RpcRouter.ts:122-123, 415-424` — `legacyRecorder` (one trust-DB record per process) and `legacyTrafficCounter` (per-method counts). This is the only evidence of who still uses the lane. Deleting it before the close removes the ability to know what the close breaks.

## Groundwork already done

`src/main/mcp/internalCli.ts` — the CLI already sends a `wmux-cli` clientName, with a source-lockstep test. That migration is what Stage 2 existed for.

## What actually breaks

1. **Perf benches.** `scripts/perf-bench.mjs:303` states the dependency outright: "No clientName → recorded 'legacy' and grandfathered by RpcRouter, so mutating calls run against the production enforce-mode app without an approval dialog." `scripts/substrate-bench.mjs` carries the same note and the same model.
2. **The bundled example.** `examples/event-recorder/recorder.mjs` `--legacy` (`:75,88,114,181,301,305`) and the `smoke` script at `examples/event-recorder/package.json:12`.
[...]

PR #1115, opened 2026-08-30
fix(security): scope a wire caller to the workspace it claimed (#922 PR-B)

openwong2kim:
Part 2 of the wire-caller half of #922, on top of #1110. This is where refusal starts.

### The lane

`operator → pinned → hosted → verified (new) → declared → legacy`

A caller carrying a token minted by `mcp.claimWorkspace` (#1110) is scoped to the workspace that token was bound to. A caller without one is unchanged — it stays in `declared`.

A **stale** token refuses the call. It never demotes to `declared`: the registry answers `{unclaimed} | {bound} | {stale}` rather than a nullable string, and the router gate treats anything that is not a live binding as stale. Two independent defences, so neither one regressing reopens the hole. BYOB hit exactly this demotion trap before.

Refusal happens **in the lane, not globally**. Global refusal deadlocks: a claim goes stale precisely when its workspace closes, and the way out is to call `mcp.claimWorkspace` again — which a global refusal would also block, wedging the MCP server until restart. The router only translates; the lane decides. An end-to-end test pins the recovery path.

### The `legacy` decision

#922 ruled that the `legacy` lane belongs to the grandfather deprecation, not here — closing it in one place alone starts a second, inconsistent deprecation clock. So this takes the narrower option the owner chose: **the lane stays allowed; browser calls must name a workspace.**

That turns out to be one branch. `legacy` already scopes to a `workspaceId` when one is sent — so a legacy caller that names its workspace is byte-identical to before. Only the *omitted* case changes: it used to fall through to a workspace-agnostic lookup that returned whichever surface registered first, and now it is refused.

The refusal is actionable rather than flat, because the reader is someone who built against the documented lane:

> `name the workspace this call belongs to — send workspaceId in the params. workspace.current returns the one you are in; workspace.list returns every id. Do not retry unchanged.`

Both methods exist and a legacy caller can reach both — verified, and pinned by a test so the message cannot come to name something that does not.

Closing the lane itself is tracked separately in #1111, with the enforcement points, the instrumentation that must outlive the clock, and the nine documentation sites that would break.

Shadow mode is untouched: this branch sits inside `scopeFor`, which is already mode-gated, so pre-#810 behaviour survives there and only `enforce` refuses.

### What this may claim

An approved tool driving a browser without naming a workspace used to get **whichever surface registered first** — possibly one belonging to a workspace it has nothing to do with. That call is now refused, and told what to add.

Not "every caller is identified." The lane is not closed, and `wmux-cli` sits in `NON_IDENTIFYING_CLIENT_NAMES` — anyone can send it. The ceiling from #922 stands: this confines an approved tool to the scope its approval implied, not malicious code already running as the user.

### Regressions pinned

A named legacy caller is byte-identical; an omitting one is refused with the new reason; shadow still passes; a stale token is refused without a registry lookup and without demotion; `null`/number/object/empty-string stamp as stale end-to-end while a genuinely absent field stays `unclaimed`.

Two existing tests were updated rather than deleted — both pinned the pre-decision behaviour, and one said so in its comment ("closing it later is a deliberate, visible change rather than a drift"). They did their job; they now carry the new intent.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved workspace scoping for browser and RPC calls.
  - Valid workspace claims now consistently bind requests to the claimed workspace.
  - Mismatched or expired workspace claims are rejected instead of accessing the wrong workspace.
  - Unscoped legacy browser calls are rejected with guidance to provide a workspace ID.
  - Explicitly workspace-scoped legacy calls continue to work as before.
  - Added clearer auditing for rejected requests caused by workspace resolution issues.


openwong2kim:
Implements the #922 verified lane: a caller with a claim token is scoped to its claimed workspace, and stale claims are refused, not demoted. Also narrows `legacy` to refuse an omitted workspaceId; shadow mode and named legacy callers are unchanged. P2 — a real security boundary with a clean rollback lever and strong regression coverage.

PR #1116, opened 2026-08-30
fix(browser): session status asked about a workspace it never named

openwong2kim:
`browser_session_status` sent no `workspaceId`:

```ts
async () => callRpc('browser.session.status'),
```

Every other browser tool resolves one and passes it. This one asked the server a workspace-scoped question without saying which workspace.

Measured on a workspace bound to the `live` profile: the tool answered `profile: "default"`. The same RPC on the same daemon, with `workspaceId` supplied, answered `profile: "live", liveAttach: true`.

### Not an edge case

The first read of this was that it needed an unusual caller — one whose identity the server failed to derive. It does not.

`callerScope` derives a workspace from `ctx` only for the operator, pinned, and hosted lanes. An ordinary MCP agent — `clientName` present, no `workspaceId` — lands in `declared` and is rejected as `workspace-unresolved`. Under `enforce` that throws; under `shadow` the request falls through to `profileFor(undefined)`, which is `'default'`.

So a correctly-placed agent gets the same wrong answer. The tool never sent the parameter, so the server never had anything to scope by.

That matters because of what this field answers: *am I about to drive the user's own Chrome?* #1105 made the handler honest about that. The handler was answering honestly — about the wrong workspace.

### Change

The handler resolves its workspace and passes it. In `shadow`, `scopeFor` returns the parameter as given, so supplying it is what makes both modes correct.

It routes through `resolveScopedReadWorkspaceId()` — the fail-soft read resolver `surface_list` and `pane_list` use — not the raw resolver, which a source invariant locks to exactly three call sites to keep reads and writes apart. `requireWorkspaceId` would throw, and on the builtin backend, where session state really is global, that throw would be spurious.

When identity does not resolve the parameter is omitted, so builtin keeps working as before.

`start`, `stop` and `list` stay global. A stale comment on `browser_session_start` claiming status is "likewise global" is corrected — it has been per-workspace since #1105.

### Interaction with #1115

Complementary, not conflicting: different files (this is `src/mcp/index.ts`; #1115 is the server side and the client transport). Once the verified lane lands, a claimed caller is scoped by its token, and this parameter either agrees with it or the call is refused — which is the safer of the two. The verified lane also closes the residue here: a caller whose identity does not resolve.

Fixing it now rather than waiting: today's `main` has no verified lane, so the parameter is the only thing that can scope this call.

No baseline rebaseline — descriptions and schemas are unchanged, because the behaviour was what was wrong, not the description.

### Honest limit on the measurement

This was not reproduced live from a correctly-placed agent. The instance available for testing ran the builtin backend on a pre-#1065 main, where session status is legitimately global and the bug cannot appear. The conclusion rests on the code path above, plus the measurement of tool-vs-explicit-RPC on a live-bound workspace.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed browser session status reporting so it displays the profile associated with the current workspace instead of incorrectly defaulting to the “default” profile.
  - Improved workspace-aware routing for browser session status requests.
  - Preserved graceful behavior when workspace information cannot be resolved.

- **Tests**
  - Added coverage confirming browser session status uses workspace-scoped routing.


openwong2kim:
What: `browser_session_status` sent no workspace identity, so the chrome backend answered for the default profile and misreported a live-bound workspace. Touches only the handler plus a source-invariant test, routed through the fail-soft read resolver so unresolvable identity still passes. P2 — it restores the per-workspace honesty #1105 promised, at near-zero risk.

PR #1117, opened 2026-08-30
chore(browser): drop the unwired applyAntiDetection instead of wiring it

openwong2kim:
`applyAntiDetection()` forged automation signals — spoofing `navigator.webdriver`, plugin lists and the like — to make an automated browser look human. It was never called. Not once in the repository's history: it arrived in the initial commit and no commit ever referenced it from outside its own file.

Removing it rather than wiring it, for two reasons.

**It is the wrong answer to the problem it addresses.** Forging signals only matters if you are trying to push automated *login* past a site's detection. wmux has a legitimate path for that and does not need to: the `live` backend drives the user's own Chrome, which the user logged into themselves, behind Chrome's own per-connection consent. Login is done by a person; automation runs after. Leaving a signal-forging helper in the tree tells the next reader we chose the other path.

**It would not even work.** Measured on Chrome 151: `navigator.webdriver` is set by `--remote-debugging-port` alone and is already `true` before anything attaches. An init script that flips it back is not defence against a detector that has other tells, and this helper does not run at all.

The file kept its surviving export, `evaluateWithGesture` (used by `browser_evaluate`), and is renamed `user-gesture.ts` so its name stops implying an anti-detection capability the code does not have.

If a real need to drive a logged-in session without a debug port ever appears, the answer is the Playwright browser extension path or the `live` backend — not resurrecting this.

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unused browser signal spoofing that could have altered automation detection.
  * Preserved user-gesture-based browser evaluation behavior.
  * Clarified the sign-in flow: authentication occurs interactively before automation begins.


openwong2kim:
This deletes the never-called `applyAntiDetection()` and renames `anti-detection.ts` to `user-gesture.ts`, keeping the used `evaluateWithGesture` for `browser_evaluate`. Net effect is dead-code removal plus one import update in `inspection.ts`, so behavior risk is near zero. P3 — low-risk hygiene chore with no functional change.

PR #1118, opened 2026-08-30
fix(security): route browser open/close/tabs through the verified lane (#922 PR-C)

openwong2kim:
Part 3 of #922, on top of PR-B. It folds `browser.open`, `browser.close`, and `browser.tabs` into `scopeFor`, the way every other target-resolving browser method already routes — so the verified lane (PR-B) actually reaches them.

### What folding does, and what it deliberately does not

For a **verified / hosted / pinned** caller: naming another workspace is now refused instead of answered. That is the constraint PR-B's changelog promised; these three methods were the gap where it wasn't yet true.

For a **legacy** caller: naming another workspace still passes. That is not an oversight — closing it means closing the grandfather lane, and #922 ruled that the grandfather belongs to one deprecation (#1111), not a browser-local fix. Doing it here would start the second, inconsistent clock the issue forbids. So the current behaviour is pinned by a **known-gap test** that references #1111, rather than quietly changed.

This is worth stating plainly because it bounds the security claim: **folding constrains the workspace a caller resolves to; it does not decide whether a caller may reach the method at all.** A legacy caller skips the capability check entirely (`PermissionEnforcer` grandfathers a no-clientName caller before capability is consulted), so it reaches `wmux.internal` methods — `browser.tabs` among them — and `scopeFor` then hands back the workspace it named. `scopeFor` only constrains the workspace axis; the reach is a separate gate. No workspace-layer change substitutes for closing the grandfather. That is the note left on #1111.

### The exception list shrinks

`browser.close` left the scope-coverage allowlist in PR-C-past; now `browser.open` (except its `external` branch, which hands a URL to the OS browser and owns no wmux surface) and `browser.tabs` leave it too. The list is meant to shrink toward empty, and the "wait for traffic evidence" note on these entries is resolved: the verified lane answers the claimed-caller's omitted case, so folding is an exception removal, not a new refusal.

### Two things kept on purpose

- **The dispatch-level hosted binding (#1097) stays.** Folding a method into `scopeFor` does not remove PR2's binding, because dispatch runs regardless of `mcp.mode` while this table is inert under shadow. Drop it and a shadow-mode plugin regains the reach #1097 took away. They don't conflict: dispatch fills the workspace as a binding, and the hosted lane then reads it as a request that named its own workspace.
- **`external` stays out.** No surface to scope; refusing an unscoped caller there breaks a working path.

### Regressions pinned

Bundled MCP is unaffected — `navigation.ts` resolves the workspace before dispatch and fails closed on its own, so every `browser.tabs` reaching main carries a non-empty id. Tests: a claimed caller scoped; a claimed caller naming another workspace refused; the bundled-MCP declared path unchanged; the legacy known gap held; a legacy omit refused.

## Summary by CodeRabbit

* **Bug Fixes**
  * Browser commands now consistently use the caller’s verified workspace.
  * Requests targeting another workspace, or without a resolvable workspace, are refused with clear guidance.
  * `wmux open` and `wmux browser close` now work outside wmux panes by using the active workspace.
  * Browser tab scope refusals are reported as terminal errors with their remedy text instead of temporary outages.
* **Compatibility**
  * Operator workspace switching, external browser use, and shadow mode behavior remain supported.


openwong2kim:

PR-C of #922 (on top of #1115): folds `browser.open`/`browser.close`/`browser.tabs` into the shared `scopeFor` lane table, so claimed callers are held to their workspace. Enforce-path behavior change only: legacy callers omitting `workspaceId` are now refused (naming one unchanged, deferred to #1111); the `external` backend and shadow mode untouched. P2 — security follow-through in the #922 series, regression-pinned by tests.

PR #1120, opened 2026-08-30
fix(browser): a cross-origin download no longer strands the page, and downloads report their filename

openwong2kim:
Two download-side fixes found while measuring the video workflow end to end. Builds on the upload PR (#1114).

### A download that navigates instead of downloading no longer strands the page

A cross-origin media link is the common case: Chrome ignores the link's `download` attribute across origins and **navigates** to the file instead. No download event fires, the call times out, and — the real damage — the tab is now on the media URL, so the original page is gone. Anything it held is lost, and later steps run against the wrong page.

`browser_download` now detects the navigation, sends the tab **back** to where it was, and returns an error that says what happened and what to do:

> The click navigated instead of downloading… The tab was sent back to `<url>`; the page was reloaded, so anything it held in memory is gone. That URL is the file — fetch it from a terminal if it needs no session…

`goBack` is tried first, `goto` as fallback. That ordering was measured, not guessed: with a three-entry history, `goto` leaves a stray entry so a later `browser_navigate_back` re-enters the media URL, while `goBack` returns history to where it belongs. The error names that when it has to fall back.

(One wrinkle handled along the way: on a wmux-driven tab the `goBack` promise doesn't resolve even though the navigation lands, so success is judged by polling `page.url()`, not by awaiting the promise. The wedged-promise root cause is noted but not resolved — a URL check sidesteps it.)

### `browser_download` now reports the filename, and its timeout is bounded

- The result carries `suggestedFilename` (and `url`). Before, only a GUID path came back, so an agent couldn't tell what it had downloaded before handing it to ffmpeg.
- `timeout` is now a schema field, `.int().min(1)`, on `browser_download` and `browser_wait_for_download`. `0` meant "no timeout" in Playwright — a tool whose whole job is to give up would hang forever. The default is unchanged and still bounds the wait for the download to **start**, not to finish: a 60s transfer under the 30s default still completes (measured).
- Also fixed: `timeout` was being silently dropped before, the same way `selector` was on upload — an unknown key discarded by the schema.

Progress reporting during a blocking download is deliberately **not** here: it needs a non-blocking start + a pollable status tool, a tool-surface change out of this PR's scope. Left as a separate track.

Measured live: cross-origin failure now restores the original page in the snapshot; a slow 40 MB download still completes at 54s under the default.

### DIFF
--- a/changelog.d/pr-q-download-page-state.md
+++ b/changelog.d/pr-q-download-page-state.md
@@ -1,0 +1,28 @@
+### Fixed
+
+- **A download that turns into a navigation no longer strands the agent on
+  another page.** Chrome ignores a link's `download` attribute across origins
+  and renders the target instead, so clicking a link to a video on someone
+  else's host navigates rather than downloading. `browser_download` timed out —
+  correctly — but left the tab sitting on the media URL, and the next snapshot
+  described a completely different page with nothing to say the original had
+  gone. The tab is now put back where it was, and the error names the cause, the
+  URL the browser ended up at, whether the page was recovered, and what to do
+  instead. Restoring goes through history rather than a fresh load: measured on
+  Chrome 141, reloading by URL leaves the stray entry in place, so a later
+  `browser_navigate_back` walks straight back into the failure.
+
+### Added
+
+- **`browser_download` now says what it downloaded.** It returned only the saved
+  path — an extension-less temp name when no `filename` was given — so the
+  browser's own filename and the source URL were reachable only by driving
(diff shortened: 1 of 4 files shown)


openwong2kim:

This PR fixes `browser_download` so a cross-origin link that navigates instead of downloading restores the original page, and adds filename/URL reporting plus a bounded `timeout`. Changes are confined to the Playwright file tools with a new test file and a protocol-baseline hash update; risk is contained to the failure path. P3: solid robustness fix, no security or data-loss angle.

PR #1121, opened 2026-08-30
fix(hooks): stop projecting 'complete' for a turn still holding on background agents (#1096)

### LINKED ISSUE #1096: Roster shows a pane 'Completed' while its backgrounded agent is still working (`Waiting for N background agent(s) to finish`)
### What happens

A pane's roster row/status shows the pane label with **Completed** (Polish locale: "Ukończony"), but the pane's own transcript is genuinely mid-turn: an \`Agent(...) Sonnet 5\` backgrounded subagent is still running (\`↳ Backgrounded agent (↓ to manage · ctrl+o to expand)\`), and the transcript explicitly shows the orchestrator waiting on it (\`✳ Waiting for 1 background agent to finish\`), then resuming it, then waiting again for the same or another background agent — none of that is "done."

### Evidence (screenshot)

Sidebar: workspace "Ziomek", pane "WtyczkiFixy" tagged **Ukończony**.

Transcript at the same moment (bottom of frame): \`✳ Waiting for 1 background agent to finish\`, immediately preceded by real, in-progress work narration ("Agent zatrzymał się w pół drogi — czekał na jednego z recenzentów zamiast dokończyć. Kazałem mu iść dalej." / "Robota jest, tylko niezacommitowana... Dam znać, jak będzie zielone").

### Relation to prior issues

- #935 (closed) is the umbrella report for roster/status disagreeing with actual pane activity.
- #1045 (closed, fixed by #1050) covered one specific trigger: sparse tool-call output failing to re-cross \`ActivityMonitor\`'s active bar.

This looks like a **different** trigger in the same family: a *backgrounded* subagent (the async \`Agent(...)\` / Task-tool-style background spawn, not a foreground tool call) appears not to keep the parent pane's activity/status "alive" while it runs, so the pane reads as settled/complete even though the parent turn is still blocked waiting on that background agent.

### Confirmation

- [x] Searched existing issues for "backgrounded agent" + completed/idle status — no exact duplicate found; related to #935/#1045 but a distinct trigger.

openwong2kim:
Fixes #1096.

## Mechanism

The daemon already knew the truth and ignored it at the last step:

1. The Claude bridge stamps a lead `agent.stop` with `wmux_leftover_work` when background agents are still running (`integrations/claude/bin/wmux-bridge.mjs`).
2. The #907 verdict gate (`CompletionAlarm`) already judges that cue correctly — "The turn has not actually ended: background work is running" — and drops the completion alarm.
3. But the status projection didn't follow the verdict: `HookIngest` broadcast `eventShapeFor`'s `'complete'` with `decision:'internal'`, and `'internal'` deliberately leaves the status dot live. So the roster wore **Completed** for the entire `Waiting for N background agent(s) to finish` hold — exactly the report.

## Change

- **`HookIngest`** — a leftover-work stop now broadcasts `status:'running'` (`decision:'internal'`), mirroring the adjacent `subagent_stop` branch's precedent ("the pane's own turn keeps going, so the broadcast says RUNNING"). The cue still feeds the alarm so `seenWorking` lets the eventual real stop (leftoverWork 0) pass the gate; no ledger write, no approval expiry — identical to today's drop path in everything but the projected status. As a side effect the daemon bridge settle now calls `noteAgentStatus('running')`, which re-arms `beginTurnIfInactive` — so the sparse spinner output during the hold (well under the 2 KB/3 s threshold) keeps the pane honest too.
- **`daemon/index.ts` webhook sink** — the outbound turn-end ping is now guarded on the projected status: the one stop that is NOT a turn end must not ping "finished" while the roster it just corrected says running.
- **`hooks.rpc.ts` (local-mode parity)** — the pre-gate boundary broadcast stamped `agentStatus:'complete'` on every `agent.stop` unconditionally. `buildTurnBoundaryMetadata` now skips the completion stamp for a leftover-work stop; activity clear and `pendingQuestion` behavior are unchanged, and the real stop after the hold stamps `'complete'` as before.

Daemon-mode main needs no change: `DaemonNotificationRouter` applies the status dot before its `'internal'` early-return and never replays boundary metadata for it, and the `running`+`internal` shape is already exercised in production by the `subagent_stop` branch — so old-main/new-daemon and new-main/old-daemon pairings both behave.

## Deliberately untouched

The `ActivityMonitor`/`onActive` once-per-cycle dedup (#1013's regression area). This change is projection-only.

## Tests

- `HookIngest.test.ts` — the existing leftover-work test now pins the projection: the drop broadcasts `status:'running'` / `'Waiting on background agents'`, and the subsequent clean stop still confirms with `status:'complete'`.
- `hooks.rpc.emit.test.ts` — new: a stop with `wmux_leftover_work: 2` clears activity but does NOT carry `agentStatus`, alongside the existing session_start twin.

Gates: `tsc --noEmit` clean, `npm run build:daemon` clean, `npm test` 13057 + 160 passed / 0 failed (fresh `npm ci`).

### DIFF
--- a/changelog.d/1121.md
+++ b/changelog.d/1121.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **A pane waiting on background agents no longer shows Completed.** When an
+  agent's turn ended with background agents still running (`Waiting for N
+  background agent(s) to finish`), the roster marked the pane Completed for the
+  whole hold — genuinely in-progress work read as done. The daemon already knew
+  the turn had not really ended (that knowledge is what kept the completion
+  alarm quiet); now the pane's status says the same thing: it stays running
+  until the real turn end, and the outbound turn-finished webhook waits with it.
--- a/src/daemon/hooks/HookIngest.ts
+++ b/src/daemon/hooks/HookIngest.ts
@@ -872,6 +872,30 @@ export class HookIngest {
       });
     }
 
+    // #1096 — a lead stop with background work still running is not a turn
+    // end, and the STATUS must say so, not just the alarm. The #907 verdict
+    // gate already judges this cue correctly ("treat as working evidence"),
+    // but the projection below would still broadcast eventShapeFor's
+    // 'complete', and 'internal' leaves the status dot live — so the roster
+    // wore Completed for the whole `Waiting for N background agent(s)` hold.
+    // Same shape as the subagent branch above: the pane's own turn keeps
+    // going, so the broadcast says RUNNING. observe() still feeds the cue so
+    // `seenWorking` lets the eventual real stop (leftoverWork 0) pass the
+    // gate; no resume closure — this candidate must never write the ledger or
+    // expire approvals, exactly like today's drop path.
+    if (signal.kind === 'agent.stop' && cue.class === 'stop' && cue.leftoverWork > 0) {
+      this.alarm.observe(sessionId, signal.agent, cue);
+      return this.broadcast(sessionId, {
+        agent: agentSlugToDisplay(signal.agent),
+        status: 'running',
+        message: 'Waiting on background agents',
+        source: 'hook',
+        hookKind: signal.kind,
+        decision: 'internal',
+        signal,
+      });
+    }
+
     // Awaiting input keeps its phone card IMMEDIATELY (R5): a remote device
     // must see the question while the provisional window runs, not 1.5s later.
     if (signal.kind === 'agent.awaiting_input') {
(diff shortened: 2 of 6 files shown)


openwong2kim:
Fixes #1096: a lead agent.stop with background agents still running now projects running instead of complete, so the roster stops showing Completed during the wait. Projection-only across HookIngest, the daemon webhook sink, and local-mode hooks.rpc — no ledger or approval changes, with tests and a changelog entry. P2: low-risk, merge-ready correctness fix in the #935 status family.

PR #1123, opened 2026-08-30
fix(browser): let the Live Chrome handshake outlast the person approving it

openwong2kim:
Binding a workspace to Live Chrome could not connect, and did not say why.

Measured: a real Chrome with remote debugging on, port 9222 listening, `DevToolsActivePort` fresh — and the handshake times out at 10s with no explanation and no visible prompt.

### Why it could never work

`CdpSocket` had **one** `timeoutMs` serving two different jobs: the per-request budget and the connection handshake (`:94`, `:156`). Live Chrome passed 10s, so the handshake budget was 10s too.

But Chrome holds the handshake open while it asks the user for permission — that dialog is the consent model for this flow. Ten seconds is not enough for a person to see a prompt and click it, so the attach was structurally impossible: we hung up before the human could answer.

Split the two. Requests keep 10s; the Live handshake gets **3 minutes**. A timeout now also closes the socket, so a late approval cannot open a connection nobody is holding.

While a handshake is pending, `onConnectPending` fires at 5s and logs that Chrome is asking — the only place that can reach the person while the caller is blocked inside the dial. Honest limit: a tool call is a single response, so this cannot be streamed to the agent mid-wait. The agent gets either the result, or an error that names the wait.

### Failures now say what to do

`:156` (timeout) and `:159` (socket error) both threw the *same* `connectError` string, so "waiting for approval" and "port is gone" were indistinguishable. The dial now carries the reason through, and the diagnosis probes whether the port is actually listening before choosing:

- listening + timeout → `LIVE_CHROME_AWAITING_APPROVAL` — click Allow in Chrome
- listening + error → `LIVE_CHROME_REFUSED` — endpoint moved; retry re-reads it
- not listening → `LIVE_CHROME_UNAVAILABLE` — enable at `chrome://inspect`

If the diagnosis itself fails it falls back to the old string, so a bug in the probe cannot bury a real failure.

### The status line stopped over-promising

`browser_session_status` said a reachable endpoint means "the browser attaches when you first drive it". It does not — Chrome still asks. It now says it will **try** to attach and to watch for the prompt.

The reachability check stays a TCP probe on purpose: upgrading it to a WS handshake would make a mere status call raise the consent dialog, which the existing design explicitly forbids.

### What this deliberately does not do

No advice to relaunch Chrome with `--remote-allow-origins`. That was the working theory going in, and measuring killed it: the 403 it produces only appears when the client sends an `Origin` header, and the global `WebSocket` this code uses does not send one. Confirmed against a Chrome launched without the flag — raw handshake without `Origin`, the global `WebSocket`, and `ws` all upgrade fine; only an `Origin`-bearing request gets 403. Shipping that hint would have pointed at the wrong cause and told users to re-launch their own browser with a flag, breaking the very consent model this path is built on.

And nothing here clicks the prompt. A person approving each connection *is* the model; the fix is to wait for them and explain, not to route around them.

### Tests

Thirteen new ones, one per behaviour: a request outliving the 10s budget and succeeding on a late accept; the 3-minute ceiling closing the socket; the 5s notice firing (and not at 4s); each diagnosis branch plus a throwing probe; and all three codes surviving `describeToolError` — which they do only because they contain an underscore, an accident now pinned by a test.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Live Chrome attachment reliability with a dedicated three-minute connection window.
  * Added progress notifications when approval is still pending.
  * Clarified connection errors for pending permission, unavailable Chrome, and refused endpoints.
  * Updated guidance to explain that attachment is attempted on the first browser action and may require clicking **Allow**.


openwong2kim:

Fixes Live Chrome attaches timing out silently: the WebSocket handshake shared the 10s request timeout, so wmux hung up before the user could approve Chrome's prompt. The change splits the budgets (3-minute handshake), separates timeout/refused/unreachable failures, and only LiveChromeClient opts in — other CdpSocket callers keep current behavior. P2: makes an unusable path work, well-tested, low risk.

PR #1124, opened 2026-08-30
fix(browser): stable snapshot refs, stale-ref rejection, and the documented extract_text default

openwong2kim:
Three defects on the browser MCP surface, found by dogfooding it against a live GitHub PR page and confirmed in the code.

## 1. `browser_extract_text` gave the opposite of its documented default

The tool schema says `includeLinks` — "Preserve hyperlinks (default false)", and the extractor defaulted it to `true`. An agent that read the description and omitted the flag got every href on the page, paying exactly the token cost it was trying to avoid, with nothing to indicate it.

`extractMarkdown` has a single caller (`browser_extract_text`) which passes the flag straight through, so flipping the extractor default is the whole fix. `browser_extract_data` goes through `extractStructuredData` and never reads `includeLinks`.

The tool description now also says outright that hrefs are dropped, so anyone who was harvesting links notices the change instead of quietly getting anchor text back.

## 2. Snapshot refs were renumbered on every snapshot, so a stale ref clicked the wrong element

Refs were a running count over the accessibility walk, so a single inserted node shifted every ref after it by one. Reproduced on a PR page: one click inserted one node, and every later ref moved. Replaying a ref from the snapshot taken moments earlier clicked the neighbouring element — the ref resolved, it just resolved to something else, with no error anywhere.

Refs now come from a per-document number space keyed on `backendDOMNodeId`, the id CDP keeps stable for a DOM node's lifetime. A node that is still there keeps its number, a new node takes the next unused one, and a number is never reissued — so a ref that named a removed element cannot come back pointing at a different one. Moving to a different document clears the identity map but not the counter, so a ref carried across an A → B → A navigation is still rejected rather than re-resolved.

`resolveRef` now throws `StaleRefError`, telling the agent to re-snapshot, in the three cases where a ref can be shown not to name what the caller thinks it names:

- the page navigated since the snapshot (compared without the URL fragment, so scroll-driven `location.hash` rewrites on docs pages are not mistaken for navigation);
- the ref is absent from the latest snapshot;
- the role+name population it was numbered against has changed. This replaces a `Math.min()` clamp that turned "the page changed" into "click whichever element is last".

That third check is deliberately narrow. The snapshot enumerates a depth-capped, possibly filtered accessibility tree while the locator sweeps the whole page, so the two counts are not measured the same way. It applies only where the index is load-bearing — a named entry the snapshot itself listed more than one of — because for anything else a mismatch would block valid clicks without buying safety.

Every ref tool already resolves inside a try/catch that renders the message as the tool result, so no call site changed.

## 3. Snapshot auto-diff was dead, and comes back

`DIFF_WORTHWHILE_RATIO` is 0.5, and renumbering rewrote nearly every line, so no diff ever came in under it — measured with two snapshots either side of one click, both returned `[snapshot: full]`. With refs holding still, an unchanged line stays byte-identical and the diff is adopted. On a 33-line snapshot with one node inserted, the rendered result went from 1255 characters to 131.

## Known residuals

- Resolution is still `getByRole(...).nth(i)`, not a direct handle from `backendDOMNodeId`. A list that re-renders in place while keeping the same role+name population can still resolve a ref to a different row. Resolving the backend id directly needs `DOM.pushNodesByBackendIdsToFrontend` plus a DOM tag round-trip and is left for a follow-up.
- Unnamed entries (icon-only buttons) are exempt from the population check, because the locator runs with no name filter and counts named siblings too. Their index was already positional before this change.
- The same-document test is a URL comparison, not a loader identity, so a same-document `pushState` to a new path reads as a navigation (safe: refs are retired) and a POST landing back on the same URL does not (caught by the refMap checks instead).

## Verification

`npx tsc --noEmit` clean; `npx vitest run src/mcp/playwright` — 39 files, 445 tests pass. Full `npm run test:parallel` passes except two pre-existing environment-dependent failures unrelated to this diff (the THIRD_PARTY_NOTICES drift guard and the atlas addon patch check, both reading installed `node_modules` state).

New tests in `snapshot.refstability.test.ts`: refs surviving an insertion, numbers never reissued after a removal, old numbers not reissued across documents, a fragment change keeping refs live, all three stale rejections, a uniquely named ref surviving a count mismatch, correct resolution while the page is unchanged, and diff adoption after an insert into a 40-row page. Reverting the identity keying fails three of them, including the diff one.

## Summary by CodeRabbit

* **New Features**
  * Snapshot references now remain stable for unchanged page elements across snapshots.
  * Stale references are clearly rejected after navigation, removal, or ambiguous page changes.
  * Repeated snapshots show only changed lines.

* **Behavior Changes**
  * Extracted text now omits link URLs by default while retaining anchor text.
  * Link URLs can be included by enabling the `includeLinks` option.


openwong2kim:

This fixes silent wrong-element clicks: snapshot refs are now keyed on backendDOMNodeId and stale refs are rejected instead of renumbered, plus extract_text honors its documented includeLinks default. It touches the core snapshot/ref-resolution path every browser ref tool depends on, so it needs the new ref-stability suite to stay green. P2 — a real correctness fix, well tested, worth merging soon.

PR #1125, opened 2026-08-30
Give agents a persistent Node REPL instead of typing into a terminal

openwong2kim:
## The gap

wmux gives MCP callers no structured runtime. The two things that look like one aren't:

- **`terminal_send`** types keys into a PTY. No return value, no error channel, no types — you scrape the screen and hope the prompt came back.
- **`browser_evaluate`** keeps page-scope globals, but they die on navigation, and it can't reach the filesystem, the network, or `require()`.

So an agent doing real work re-derives its context on every call: re-read the file, re-parse the JSON, re-connect to the thing. A persistent REPL is the single highest-use agent surface in comparable tools; wmux had zero of it.

## What this adds

Three tools — `repl_run`, `repl_reset`, `repl_sessions` — backed by a persistent Node runtime whose **variables, required modules, and open handles survive between calls**.

```
repl_run  { code: 'let rows = JSON.parse(fs.readFileSync("data.json"))' }
repl_run  { code: 'rows.filter(r => r.failed).length' }   → 3
```

Top-level `await` works. Return values come back inspected. stdout, stderr, and the result are separate labelled blocks.

## Design notes

**A child process, not `node:vm` in-process.** REPL code is code the caller wrote seconds ago and has never run: it calls `process.exit()`, it throws from native callbacks, it blows the heap. The MCP broker hosts *every* agent's connection in one process, so an in-process runtime would turn one agent's typo into every agent's outage. A child makes the blast radius exactly one session.

**The eval protocol rides the child's IPC channel**, which leaves stdout/stderr free to carry the user code's own output verbatim — nothing to frame, nothing to escape, and no way for a printed line to forge a protocol message.

**`vm.runInThisContext`, not a fresh `vm` context.** A fresh context has no `setTimeout`, no `fetch`, no `TextEncoder` (measured, not assumed), so it would need a hand-curated global list that goes stale every Node release. The child's real context gives the whole standard library, and top-level `let`/`const` still persist because a Script's lexical declarations live in the context's global lexical scope. Only `require` is installed explicitly, bound to the session cwd.

**Two timeout layers, because one isn't enough.** The vm watchdog stops a synchronous runaway *while keeping session state*. A parent hard deadline SIGKILLs the child for what the watchdog cannot see — a promise that never settles, a blocked native call — and that state loss is always reported, never papered over.

**Bounded output.** stdout/stderr are retained head-and-tail per eval with an exact elided byte count. A 2 MB flood comes back as a 65 KB tool payload with both ends intact.

## Lifecycle, stated plainly

Sessions are scoped to **the caller's MCP connection**, held in `ConnectionScope` beside the `PlaywrightEngine` for the same reason: a process-global map would hand one agent another agent's live state. They die on reset, idle (15 min), hard timeout, connection close, or wmux restart. None of that is survivable, and the tool descriptions say so — an agent that believes a variable survived when it didn't writes code against a fiction.

## Security posture

**No new authority.** A caller holding `terminal_send` already drives an arbitrary shell as the user in its own pane. This changes ergonomics, not the trust boundary. It takes no RPC and needs no workspace identity — nothing crosses the substrate, so there's nothing for the daemon to authorize.

The child's env is **narrower** than that shell's: the gated-automation filter (wmux/Electron internals, credential-shaped names), plus `CLAUDE*` / `ANTHROPIC*` / `AI_AGENT` dropped on top — the markers that make a nested agent silently stop persisting its transcript, or quietly move work onto metered auth. Withheld credential names (not values) are reported so a failing script is diagnosable in one step.

No filesystem or network sandbox is claimed, and the tool description says that too.

**Workspace boundaries do not reach into the REPL.** Because this runs arbitrary JavaScript with full filesystem and network access, the workspace scoping that #922's verified lane applies to workspace-addressed tools has no purchase here: nothing stops an approved caller from using `repl_run` to read another workspace's files. That is not a contradiction of #922, whose contract is to keep an approved tool inside its approved scope rather than to defend against same-user code — but the boundary is worth stating plainly rather than leaving to be discovered.

## Review

A multi-model panel reviewed the diff (Claude + GLM; Codex was unavailable — `402 deactivated_workspace`). Five findings had cross-model agreement. Everything accepted is fixed and covered by a test. The ones worth naming:

- **User code could forge its own eval result.** It shares the child's globals and therefore `process.send`, so a script could report success, collect the answer, and clear the hard-deadline timer while spinning forever in the shared broker. Replies now must carry the in-flight eval id.
- **The top-level-await rewrite could silently run a different program.** It turned the last statement into a `return`, guarded only by "does it still compile" — but no semicolon is inserted before a newline followed by `[`, so `const a = await f()\n[0]` is *one* expression that split into two compilable halves and answered `[ 0 ]`.
- **The per-connection cap wasn't a host bound.** Four sessions × N agents in one broker. There's now a process-wide ceiling.
- **A scopeless call fell back to a shared registry**, which would have handed one agent another's live session. Broker mode now fails closed.
- **The output buffer retained the caller's chunks**, pinning whole pooled blocks and holding tens of megabytes behind a 64 KB cap — defeating the guarantee it exists to make.

## Verification

- `tsc --noEmit`, `npm run lint` on the new files, full `npm test` (two failures are the known symlinked-`node_modules` worktree artifacts: `generateNotices` and `atlasCoherence`, both unrelated).
- `npm run build:mcp && npm run probe:mcp` green; protocol baseline updated. Note the full tools/list is now **79796 / 80000 bytes** — 204 to spare. Rebasing onto #1124 had left only 27, so the REPL descriptions were trimmed rather than the shared budget raised; the next tool added will still need that budget revisited.
- 34 runtime tests drive real child processes: state persistence, top-level await, both timeout layers, hard-kill state loss, child self-exit, orphan reaping via the IPC watchdog, output truncation, env scrub, connection isolation, and every accepted review finding.
- Dogfooded end to end against the compiled `dist/` handlers.

## Not in scope

Multi-language runtimes (Python, Ruby), daemon-hosted sessions shared across panes or workspaces, persistence across a wmux restart, remote execution, a UI panel, streaming partial output, and any filesystem/network sandbox.

### DIFF
--- a/changelog.d/1125.md
+++ b/changelog.d/1125.md
@@ -1,0 +1,14 @@
+### Added
+
+- **Agents get a real REPL.** Three new MCP tools — `repl_run`, `repl_reset`,
+  `repl_sessions` — give an agent a persistent Node runtime where variables,
+  required modules, and open handles survive between calls, so it can build up
+  context instead of re-deriving it every time. Top-level `await` works, return
+  values come back inspected, and stdout, stderr, and the result arrive as
+  separate labelled blocks. Until now the only ways to run code were
+  `terminal_send`, which types keys at a shell and leaves you scraping the
+  screen for the answer, and `browser_evaluate`, whose page globals vanish on
+  navigation and can never touch a file or the network. Each session is its own
+  child process, so a runaway loop or a `process.exit()` costs you that session
+  and nothing else; runs have a timeout, output is capped with both ends kept,
+  and sessions live as long as your MCP connection. (#1125)
--- a/docs/api/inventory.md
+++ b/docs/api/inventory.md
@@ -180,6 +180,16 @@ The wmux MCP server (hosted in-process, named-pipe transport to the daemon) expo
 | `wmux_search_panes` | `pane.search` | |
 | `send_message` | inter-workspace messaging — send a message to another workspace. Backed by the same handler as `a2a_task_send` (which is registered as a literal alias). NOT `input.send` semantics. | |
 
+### REPL surface (experimental)
+
+Backed by **no RPC method**: the sessions are child processes of the MCP server itself, so nothing crosses the substrate boundary and there is nothing for the daemon to authorize. Scope is the caller's MCP connection — a session is not shared between panes or workspaces and does not survive a wmux restart. `full` profile only; deliberately absent from the commander surface.
+
+| MCP tool | Backs RPC method | Description |
+|---|---|---|
+| `repl_run` | *(none — in-process child)* | Evaluate JavaScript in a persistent Node runtime; variables, required modules, and open handles survive between calls. |
+| `repl_reset` | *(none — in-process child)* | Kill a session and its state; the next `repl_run` starts a fresh runtime. |
+| `repl_sessions` | *(none — in-process child)* | List the sessions this connection holds (cwd, pid, age, busy). |
+
 ### A2A surface (stable)
 
 | MCP tool | Backs RPC method | Description |
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -94,7 +94,10 @@
         "surface_new",
         "surface_close",
         "pane_stash",
-        "pane_unstash"
+        "pane_unstash",
+        "repl_run",
+        "repl_reset",
+        "repl_sessions"
       ]
     },
     "commander": {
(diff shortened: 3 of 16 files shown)


openwong2kim:
This PR adds persistent per-connection Node REPL tools (repl_run / repl_reset / repl_sessions) backed by child processes scoped in ConnectionScope like the PlaywrightEngine. Review focus: the no-new-authority claim (terminal_send as the ceiling), child lifecycle and dual-timeout handling, and the tools/list budget now at 79866/80000 bytes. P2 — a large new agent-facing surface that merits prompt, careful review.

PR #1126, opened 2026-08-30
fix(terminal): stop OpenCode synchronized-output flicker

imhalawa:
## Summary

Stop OpenCode's rapidly repeated DEC 2026 synchronized-output frames from making the terminal pane flash between content and blank states. The fix is a narrow xterm 6.0 dependency patch that paints a completed synchronized frame before the next frame can suppress its deferred render.

## Changes

- Patch `@xterm/xterm` 6.0.0's UMD, ESM, and source artifacts through the repository's existing patch-package install path.
- Add a serial Chromium/WebGL runtime regression that emits 30 OpenCode-style full-screen frames and requires continuous paints plus correct final buffer content.

### Fixed

- **OpenCode no longer flashes rapidly between its content and a blank terminal pane.** wmux now presents each completed synchronized-output frame before OpenCode begins the next one, instead of letting xterm defer and repeatedly suppress the paint.

## Stability tier impact

n/a

## Test plan

- `npx tsc --noEmit` — pass.
- `npx vitest run src/renderer/terminal/__tests__` — 22 files, 371 tests pass.
- `npx vitest run --config vitest.runtime.config.ts scripts/__tests__/opencode-sync-render.runtime.test.mjs` — pass with the patch; fails when the patch is reversed (4 paints for 30 frames versus the required 22).
- `npm test` — the parallel phase hit an unrelated existing 5-second timeout in `channelsSlice.test.ts`; the exact file passed immediately in isolation (61/61). The command stopped before runtime tests, which were run separately above.

## Related issues / PRs

- Tracks upstream xterm.js issue https://github.com/xtermjs/xterm.js/issues/6071 and PR https://github.com/xtermjs/xterm.js/pull/6073.
- OpenCode-side symptom report: https://github.com/anomalyco/opencode/issues/44866.

## Reviewer checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed rapid flashing between OpenCode content and a blank terminal pane during synchronized output.
  - Ensured completed terminal frames appear before the next frame begins rendering.

- **Tests**
  - Added automated coverage for rapid synchronized output, confirming intermediate and final frames render correctly with WebGL enabled.


openwong2kim:

This PR fixes rapid content-to-blank flashing in OpenCode panes by patching @xterm/xterm 6.0.0 so completed DEC 2026 synchronized frames paint immediately. It touches the renderer core via a vendored patch-package diff plus a Chromium/WebGL runtime regression, so the patch needs revalidating on every xterm bump. P2: user-visible rendering breakage for OpenCode, narrow and well tested.

openwong2kim:
Thanks a lot for this, @imhalawa. The root-cause write-up is spot on — xterm 6.0 deferring the completed synchronized frame through requestAnimationFrame and then dropping the paint when the next frame opens first is exactly the kind of thing that’s painful to track down, and OpenCode’s continuous full-screen redraw hits it every time. I really appreciate that you kept it to a narrow vendored patch instead of forking, and added a real Chromium/WebGL runtime regression test on top. Great contribution — thank you! 🙏

PR #1127, opened 2026-08-31
Cut 5,677 bytes from tools/list and ratchet the budget to the design target

openwong2kim:
Every agent session pays for `tools/list` before it does any work — the host copies each tool's name, description and JSON schema into the system prompt. The non-browser surface had grown a second copy of itself: error-code enumerations the runtime already reports with an action hint, field semantics spelled out both in the tool description and in that field's own schema description, and version stamps written for maintainers rather than for callers.

This trims the text and then ratchets the CI budget onto the ground the trim cleared, so the design target becomes an enforced floor instead of a comment.

## Result

| Profile | Before | After | Delta | New budget | Headroom |
|---|---|---|---|---|---|
| full | 79,796 B | 74,119 B | −5,677 | 75,000 (was 80,000) | 881 |
| commander | 48,991 B | 43,314 B | −5,677 | 45,000 (was 49,000) | 1,686 |

The two budgets are the numbers `docs/design/mcp-vnext-2026-07-28.md` set as the goal for this surface. Every tool touched is on the commander surface, so each cut is counted once in each profile.

Unchanged, and asserted by `npm run probe:mcp`: tool names, tool ordering, profile membership, every input schema's shape, and the server instructions (`instructionSha256` is byte-identical in both profiles). No invoke logic moved.

## Per-tool

| Tool | Before | After |
|---|---|---|
| channel_post | 3,725 | 2,533 |
| a2a_task_update | 3,272 | 2,656 |
| wmux_events_poll | 2,924 | 2,714 |
| fanout_start | 2,748 | 2,387 |
| pane_set_metadata | 2,509 | 1,934 |
| send_message | 2,210 | 2,029 |
| a2a_task_send | 1,999 | 1,814 |
| channel_mission_start | 1,410 | 1,218 |
| terminal_read | 1,448 | 1,212 |
| terminal_send_key | 1,347 | 1,142 |
| deck_ask_decision | 1,373 | 1,214 |
| deck_resolve_decision | 1,325 | 1,157 |
| wmux_search_panes | 1,307 | 1,172 |
| channel_ack | 1,286 | 1,112 |
| channel_read | 1,237 | 1,101 |
| channel_invite | 1,238 | 1,109 |
| terminal_send | 1,220 | 890 |
| pane_get_metadata | 1,213 | 933 |
| channel_create | 1,257 | 1,170 |
| a2a_task_query | 907 | 830 |
| channel_join | 1,114 | 1,046 |
| terminal_read_events | 1,205 | 1,199 |

Browser tools are untouched — they were dieted in an earlier pass and are out of scope here.

## Where every deleted sentence went

Nothing was dropped on the grounds that it was merely long. Each removal has a destination.

| Removed from the wire | Went to | Why that is safe |
|---|---|---|
| `channel_post`: `PERSIST_FAILED` / `CHANNEL_ARCHIVED` / `CHANNEL_MENTIONS_TOO_MANY` enumeration | error remedy (already present) + source comment | `callChannelRpc` returns every failure as `isError` with `Error [CODE]: message`, and `ChannelService` supplies both halves ("Channel is archived", "Post exceeds 64 mentions"). The caller learns the code at the moment it matters. |
| `channel_post`: "U2 maintainer directive: do not swallow saveImmediate errors on the post path" | source comment | Maintainer rationale. It tells a caller nothing about how to call. |
| `channel_post`: `client_msg_id` idempotency restated in the description | deleted as duplicate | Already the whole content of `client_msg_id`'s own `.describe()`. |
| `channel_post` `pane_id`: literal refusal reasons `"pane_not_in_workspace"` / `"pane_not_live"` | source comment; behavior kept on the wire | The strings arrive as machine-readable `droppedMentions[].reason`. The description still states what a refusal *does* — reported, mention still lands badge-only, never redirected to a sibling. |
| `a2a_task_update`: `completion_evidence_*` reason-code list | error remedy (already present) + source comment | `A2aTaskService.evidenceGateHint` pairs every rejection code with a sentence naming what to attach and retry with. |
| `a2a_task_update`: "you cannot jump straight from submitted to completed… take submitted -> working FIRST" | `status` field `.describe()` (which already lists the full transition map) | Field-level detail belongs to the field. The description keeps what the map cannot say: terminal states are final, and a rejection names the allowed next states. |
| `pane_set_metadata`: per-mode prose for `merge` / `replace` / `replaceShared`; `expectedVersion` prose | the `mergeMode` and `expectedVersion` `.describe()`s | Same words, said once, next to the field they govern. The description now points at them. |
| `pane_set_metadata`, `pane_get_metadata`: `(v2.9.0+)`, `v2.8.x`, `#1018` | source comments | Version and issue provenance is maintainer context. |
| `channel_create`: "(plan KTD10)" | source comment | Same. |
| `terminal_send`: the `submit: true` mechanic explained a second time in the description | the `submit` `.describe()` | The description keeps only the surprise — text is written uncommitted by default. |
| `wmux_search_panes`: regex flag caveat restated in the description | the `regex` `.describe()` (unchanged, still carries `[Ee]rror`) | Duplicate. |
| `channel_ack`: the `member_id` requirement restated in the description | the `member_id` `.describe()` | Duplicate. |
| `channel_mission_start`: `idempotency_key` retry semantics restated | the `idempotency_key` `.describe()` | Duplicate. |
| `terminal_read_events`: shell-integration caveat moved behind the "use this when" sentence | kept, reordered | No loss; the useful half now reads first. |

Everything counter-intuitive enough to be worth its bytes stayed on the wire: a channel post is a notification and not a delivery; `terminal_send_key` is not a submit mechanism and `ok` does not mean anything was submitted; a poll cursor can fall out of the ring and `a2a.task` is dual-party; `fanout_start` is never auto-approved and derives its repository from verified identity; a completion with no verified item is accepted but graded unverified.

## Two corrections found in review

`since_seq`'s description claimed a limited page returned "the most recent `limit` of the remainder". It does not, and has not since the paging split: with a cursor, `readVisibleMessages` returns the **oldest** `limit` rows at or after the floor, precisely so an ack-and-repeat drain cannot skip the rows between the cursor and a tail-sliced page. `channel_read`'s own tool description already said OLDEST, so the field contradicted its own tool, and an agent that believed the field would write the lossy loop the split exists to prevent. The wording predates this branch; it is corrected here because the diet rewrote that line.

`deck_ask_decision` lost "do not act further" during compression. Ending the turn and taking no further action are two promises to the operator, and only the first survived. Restored.

## Verification

- `npx tsc --noEmit` clean.
- `npx vitest run src/mcp src/shared/__tests__/commanderSurface.test.ts` — 62 files, 724 tests, all passing. No test asserts any modified string.
- `npm run build:mcp && npm run probe:mcp` green across all three protocol versions and all three handshake layouts, against the ratcheted budgets.
- Dogfooded standalone: `node dist/mcp-bundle/index.js` driven through a raw `initialize` → `tools/list`, byte counts read off the wire rather than estimated, and each trimmed description re-read to confirm it still carries its contract.

## Note on headroom

`full` lands 881 bytes under its new ceiling — roughly one average tool. That is deliberate: 75,000 is the design document's target, not a number picked to be comfortable. A PR that adds a tool to the full surface will have to either buy the room back or raise the budget with a reason, which is the pressure this ratchet exists to apply.

## Summary by CodeRabbit

- **Documentation**
  - Refined descriptions for channel, fanout, terminal, messaging, task, event, and command tools.
  - Clarified parameter guidance, task limits, polling outcomes, and identity-based command behavior.
  - Updated protocol profile limits and wire-format metadata for the full and commander profiles.
  - Simplified error information presented in channel tool documentation.
  - Documented reduced tool-list sizes and tightened validation limits.

- **Bug Fixes**
  - Corrected channel read cursor behavior for continuous acknowledgment and repeat draining.

### DIFF
--- a/changelog.d/1127.md
+++ b/changelog.d/1127.md
@@ -1,0 +1,7 @@
+### Changed
+
+- **The MCP tool list got 5,677 bytes lighter, and the budget guarding it tightened to match.** Every agent session pays for `tools/list` before it does any work — the host copies each tool's description and schema into the system prompt — and the wmux surface had grown a second copy of itself: error codes the failure already reports, field semantics written out twice, and version stamps meant for maintainers. The text is now said once, in the place that owns it, and the counter-intuitive contracts an agent actually needs are all still there: that a channel post is a notification and not a delivery, that `terminal_send_key` is not a submit mechanism, that a poll cursor can fall out of the ring. The full surface drops from 79,796 to 74,119 bytes and the commander surface from 48,991 to 43,314, with the CI budgets ratcheted from 80,000/49,000 to 75,000/45,000 so the saving cannot quietly be spent again.
+
+### Fixed
+
+- **`channel_read`'s `since_seq` no longer describes paging backwards.** With a cursor, a limited read returns the oldest messages at or after it, so an ack-and-repeat drain stays contiguous. The field description claimed the opposite — the most recent of the remainder — contradicting the tool's own description and inviting exactly the lossy drain loop the paging rule exists to prevent.
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -2,8 +2,8 @@
   "schemaVersion": 2,
   "profiles": {
     "full": {
-      "maxListBytes": 80000,
-      "wireResultSha256": "eb6cdfcae7a3bac1a2efdd9970ebaec56c7a9ded13764c9efbe5b373f65b471b",
+      "maxListBytes": 75000,
+      "wireResultSha256": "8ab858611f9ff986915bef6eb364cb80ded1ad7f4ea13f5ea63f3a98d731c0bc",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
@@ -101,8 +101,8 @@
       ]
     },
     "commander": {
-      "maxListBytes": 49000,
-      "wireResultSha256": "c4cd51661adaef9efa197b947282fe11a4dbd8f2d28879691ed9eee20d6a979f",
+      "maxListBytes": 45000,
+      "wireResultSha256": "b13d93dbc0b19487d6332a7bb2c531ca03f9f58e68924d12b0f8c9c9e82ff676",
       "instructionSha256": "5a97f851c892326eba7cbcf46292f6925834d5794172461d1ced997d90ac1deb",
       "toolNames": [
         "terminal_read",
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 75000,
-      "wireResultSha256": "8ab858611f9ff986915bef6eb364cb80ded1ad7f4ea13f5ea63f3a98d731c0bc",
+      "wireResultSha256": "f10cfab999d35edb6246064a3dee9111ac9c0f321743f3c9b73ba45f4228bba2",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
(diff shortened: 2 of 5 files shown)


openwong2kim:

Trims 5,677 bytes of tool/field descriptions across `channels.ts`, `fanout.ts`, `index.ts`, and ratchets the tools/list budgets in `mcp-protocol-baseline.json` (full 80k→75k, commander 49k→45k). Also corrects a real doc bug: `since_seq` now says pages return the oldest rows at/after the cursor. Text-only with no invoke logic moved and probe-asserted name/schema stability, so risk is low: P3.

PR #1128, opened 2026-08-31
Add the launch-time core MCP tool profile

openwong2kim:
Implements the `core` MCP tool profile that `docs/design/mcp-vnext-2026-07-28.md` reserved and `WmuxToolProfile` has named since the catalog landed, with no implementation behind it.

`--core` registers the full surface minus `browser_*` (39) and `company_a2a_*` (6) — **47 of 92 tools** — for agents doing terminal, pane, workspace, channel, and delegation work that never opens a browser or crosses the company A2A bus.

| profile | tools | `tools/list` bytes | budget |
| --- | --- | --- | --- |
| `full` | 92 | 74,119 | 75,000 |
| **`core`** | **47** | **46,665** | **49,000** |
| `commander` | 41 | 43,314 | 45,000 |

Roughly 27,500 bytes (~6,900 tokens) off every session that opts in.

## core is an optimization profile, not a security boundary

This is the load-bearing distinction and it is written into the code, not just the PR.

`commander` is a role: it mints a per-spawn token, claims that role on every outbound RPC, and narrows `PermissionEnforcer`'s allow lane. `core` does none of that. It shortens `tools/list` and changes nothing else — no token, no role claim, no RPC lane, no `TEARDOWN_DENY_METHODS`. A core-mode child keeps exactly the authority an ordinary pane agent has.

Two consequences follow, both deliberate:

- `setCommanderRole` stays inside the `commander` branch. Adding role plumbing behind `--core` would protect nothing (a caller who wants the wider surface just omits the flag) while implying protection that is not there.
- The shim's new `coreMode` handshake field is **additive**: an old broker ignores it and hosts `full`, which is *wider*. That fallback is safe here precisely because nothing is being fenced off — a version skew costs token savings, not a guarantee. `commanderMode` must never be treated this way.

## Selection

Launch-time argv only, never env, immutable for the process lifetime — the same rule `--commander` follows, for the same reason: a host that strips the environment must not be able to silently change the surface.

`--commander --core` together resolves to `commander` (fail closed) with one stderr warning. On the broker path that warning lands on the daemon's stderr rather than the agent host's, so the broker's handshake log line carries `core=` too.

## Drift gates

A new profile is only worth having if a new tool cannot quietly fall out of it. Three interlocking assertions:

1. `coreSurface.test.ts` derives the expectation from `scripts/mcp-protocol-baseline.json`'s `full.toolNames` — a new tool on the full surface **fails the build** until someone explicitly classifies it.
2. That same test pins `CORE_TOOL_SURFACE` to `baseline.core.toolNames`. Without it, dropping `'core'` from one catalog spec fails the probe once, and a reflexive baseline regeneration makes both gates green again with a tool silently missing.
3. `expectCoreCatalogLockstep` (mirroring the commander version) keeps the legacy manifest and the migrated specs' `profiles` arrays naming the same tools, called from all four domain suites.

The probe is generalized from two profiles to three via a profile→argv table: every narrower profile must be a strict, **order-preserving** subset of `full`; `commander ⊆ core`; and `core` carries no `browser_*` or `company_*`. The commander token is injected for `commander` only — `core` claims no role, so a token there would probe a topology that never ships.

## Review findings folded in

A panel review caught three things worth calling out:

- **Instructions advertised tools the profile never registers.** The handshake text was keyed off `commanderMode`, so `core` shipped full's *"wmux provides terminal, browser, ..."* and *"Discover opaque IDs ... with `browser_tabs`"* — pointing the agent at a tool absent from its own `tools/list`. `getWmuxMcpServerInstructions` now takes the resolved profile; `core` gets its own hash rather than reusing full's.
- **`tomlBlock` hardcoded `args = [scriptPath]`** instead of going through `wmuxMcpEntry`, so a profile argument would have reached JSON hosts and not TOML hosts. Both writers now derive args from one function.
- **Prefix skew** between the test (`company_a2a_`) and the probe (`company_`) — equivalent today, contradictory the day a company tool lands under another sub-prefix.

## Out of scope

- **No production caller selects `core` yet.** `wmuxMcpEntry(scriptPath, profile?)` is a stub with an optional argument and no behavior change for its existing callers; `upsertMcpServer` takes no profile, so every registration path still writes the `full` entry. Threading a profile choice through `mcpTargets` and the settings UI is a separate change. A consequence noted in the code: until that lands, a hand-added `--core` in a user's config is replaced on the next upsert.
- No version bump, no `CHANGELOG.md` edit (fragment only).

## Verification

- `npm run build:mcp && npm run probe:mcp` green — 3 protocol versions × 3 handshake layouts × 3 profiles.
- Live dogfood against the built bundle, raw `initialize` → `tools/list`: `--core` → 47 tools / 46,665 B; `--core --commander` → 41 tools + the warning; no flags → 92 tools.
- `npx tsc --noEmit` clean; `src/mcp` + `src/shared` suites green.

## Summary by CodeRabbit

- **New Features**
  - Added an optional `--core` MCP profile with a streamlined tool catalog, reducing discovery payload size.
  - Core mode includes REPL, workspace, channel, pane, and related tools while excluding browser and company tools.
  - REPL functionality is now available in both full and core profiles.

- **Bug Fixes**
  - Updated profile-specific instructions so unavailable browser tools are no longer recommended.

- **Documentation**
  - Documented core profile behavior and expanded REPL profile availability details.


openwong2kim:

Adds a launch-time `core` MCP tool profile: full minus `browser_*`/`company_*` (47 of 92 tools), argv-selected, with additive handshake fields and drift-gate tests pinning the manifest to the protocol baseline. Explicitly an optimization, not a security boundary, and no registration path selects it yet, so existing callers are unchanged. P3 — additive and well-gated, no behavior change until threaded through the UI.

PR #1132, opened 2026-08-31
fix(daemon): identify the daemon by its argv entry script — the #1027 redo (#1025)

openwong2kim:
Redo of #1027 (reverted by #1028), implementing all four requirements #1028's post-merge panel set. Reopens nothing; closes the class #1025 reproduced by execution: an unrelated `node <anything>/daemon/index.js` on a recycled PID was verified as wmux's daemon and SIGKILLed.

## The four requirements, and where each landed

1. **Entry-script position, not any-token.** `argvIdentifiesDaemonScript` locates the first non-flag token after the executable and compares only that. A process merely *holding* our script path as an argument (`vim .../daemon-bundle/index.js`-class, #1027's third defect) no longer verifies. Pinned by a unit case and a real-process execution test.
2. **argv preserved as an array.** `getProcessArgv` replaces the string-returning probe: Linux NUL-splits `/proc/<pid>/cmdline` and never round-trips it through spaces — the exact mechanism (#1028) by which a spaced install path shattered and turned wmux's *own* daemon into a "mismatch". Windows keeps the quote-aware CIM tokenization; macOS `ps` output is inherently space-joined, so the matcher compensates with a progressive re-join against known candidate paths (unit-tested with a `/Applications/wmux 2.app/...` shape).
3. **Fallback exclusive + exact shape.** The cross-host branch accepts exactly `daemon-bundle/index.js` at the entry position — segment equality, no `startsWith` — so `daemon-bundler/…` and `daemon-bundle-backup/…` (#1027's second defect) are out. A cross-host **dev** daemon (tsc layout, no bundle) is deliberately not covered: refusing degrades to the respawn budget, while the generic `daemon/index.js` marker is the very false positive #1025 executed. The crossHost test's marker path moves to the exact shape accordingly.
4. **`ensureDaemon` mismatch branch refuses instead of cleaning + spawning.** This is the split-brain-sensitive piece #1028 asked to have reviewed separately — please look at it in isolation. Under identity matching, a mismatch can mean "our own daemon we failed to prove", so branch (b) now runs the same recovery dance as the blocked-probe paths: **escalated re-ping first** (a process that answers on the authed pipe is our daemon regardless of path proof) → **user approval** → only then the cleanup + spawn. Default (headless, or declined) is a refusal that names the pid file and the manual command. Behavior change accepted with this: the genuine sibling-app-on-recycled-PID case, which used to self-heal silently, now asks — one dialog, and the #537/#543 spawn-over-live-daemon path is unreachable without explicit consent.

Also fixed from the #1028 findings list: the Electron seam calls the resolver **inside** a try/catch (a throwing `app.getAppPath()` degrades to the shape-only fallback instead of crashing a "never throws" path), and the tests reach the matcher **through `deps.resolveDaemonScriptCandidates`** — the real wiring #1027's tests bypassed — plus real-process execution tests asserting on the real OS probes.

## Tests

- 10 new unit cases on the matcher (position, flags, re-join, exact shape, #1025 layout, win32 normalization)
- 4 real-process execution tests (innocent refused, trailing-argument refused, candidate killed, `daemon-bundler` vs `daemon-bundle` split)
- 2 `ensureDaemon` branch tests through injected deps (refuse-by-default: nothing killed, nothing cleaned; approval: falls through to clean+spawn without killing the occupant)
- Full `src/shared/daemon` + `src/main/daemon` suites: 158 passed; `tsc --noEmit` clean; eslint at main's baseline

## Follow-up: spaced install paths on macOS (dogfood finding)

Live dogfooding on macOS found the identity branch still failing whenever the
app path contains a space (`/Applications/wmux 2.app/...`, the name macOS gives
a second copy). `ps -o command=` joins argv with spaces, so the **executable**
fragments too, not just the script; the entry position then landed inside a tail
fragment of the executable (`2.app/Contents/MacOS/wmux`) and the progressive
re-join could never reach a candidate. Requirement 4 kept it safe — it refused
rather than killed — but such an install could never prove its daemon's
identity, so every launch raised the approval dialog.

Fix: read `ps -o comm=` alongside the command line. It returns argv[0] on one
line with spaces intact, so stripping that prefix reconstructs `[exe, ...rest]`
and the entry position is correct again. A missing or non-prefix `comm` falls
back to the historical whitespace tokenization (refuse-safe). The alternative —
generalizing the re-join start position — was rejected because allowing a match
at an arbitrary offset re-admits #1027's third defect (`node innocent.js <our
path>`), and no cheap check distinguishes a fragmented executable from a real
extra argument.

Known limitation, accepted: the cross-host fallback still inspects the entry
token alone, so a cross-host daemon under a spaced path remains unprovable on
macOS. Widening it to re-joined spans would re-admit that same trailing-argument
false positive, and an unproven identity refuses rather than kills. Documented
in the matcher's doc comment.

New regression tests (same file): the real dogfood argv — executable *and*
script both fragmented — matches its candidate; the trailing-argument case is
re-asserted false through the same probe path, including with a fragmented
executable in front; plus the probe normalizer's fallbacks. Verified live
against a real spaced-path process and the real `ps` probe: fragmented case
`true`, trailing-argument case `false`, old tokenization reproduces the `false`.

A two-model panel review of that follow-up added two hardening changes: the
optional `comm=` probe gets a 1s budget of its own (it is a refinement, not
the primary read, and two 3s probes back to back doubled the worst-case
synchronous block on the launch path), and the prefix strip is refused unless
what follows is an absolute path or a flag — a platform whose `comm` is
truncated can end mid-path and still pass the prefix test, cutting at the
wrong boundary. Anything else falls back to the historical tokenization.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved daemon identification to prevent unrelated processes from being terminated.
  * Prevented duplicate daemon launches when an existing process cannot be verified.
  * Added a confirmation prompt before cleanup and restart when daemon identity is uncertain.
  * Improved recognition of executable paths containing spaces on macOS.
  * Ensured daemon matching requires the exact launched entry script, reducing false matches.

### DIFF
--- a/changelog.d/1132.md
+++ b/changelog.d/1132.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- wmux could SIGKILL an unrelated program that happened to be running a
+  script at `.../daemon/index.js` if the operating system had recycled the
+  daemon's process id onto it. The daemon is now identified by the exact
+  entry script wmux itself launches (compared at the argv entry position),
+  and when that identity cannot be proven, wmux asks before cleaning up and
+  spawning — it no longer risks starting a second daemon over a live one.
+  (#1025, #1132; redo of the reverted #1027 per #1028)
--- a/changelog.d/1132.md
+++ b/changelog.d/1132.md
@@ -7,3 +7,10 @@
   and when that identity cannot be proven, wmux asks before cleaning up and
   spawning — it no longer risks starting a second daemon over a live one.
   (#1025, #1132; redo of the reverted #1027 per #1028)
+- On macOS, wmux failed to recognize its own daemon when the app was
+  installed under a path containing a space (for example `wmux 2.app`, the
+  name macOS gives a second copy): the process listing joins arguments with
+  spaces, so the executable path itself arrived shattered and the daemon's
+  identity could never be proven — leaving that user with a confirmation
+  dialog on every launch. The executable is now read separately and kept
+  whole, so the daemon identifies correctly. (#1132)
--- a/src/main/daemon/launcher.ts
+++ b/src/main/daemon/launcher.ts
@@ -15,8 +15,8 @@ import {
   pollDaemonReady,
   isDaemonPipeGone,
   ensureDaemon as ensureDaemonCore,
-  killDaemonByPidFile,
-  killVerifiedDaemonPid,
+  killDaemonByPidFile as killDaemonByPidFileCore,
+  killVerifiedDaemonPid as killVerifiedDaemonPidCore,
   type DaemonInfo,
   type DaemonLauncherDeps,
   type ProcessLiveness,
@@ -48,8 +48,6 @@ export {
   DAEMON_READY_HARD_CEILING_MS,
   pollDaemonReady,
   isDaemonPipeGone,
-  killDaemonByPidFile,
-  killVerifiedDaemonPid,
 };
 export type {
   DaemonInfo,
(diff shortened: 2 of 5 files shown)


openwong2kim:

Redo of the reverted #1027, closing #1025: the kill gate now matches the daemon's argv entry script exactly, and `ensureDaemon` refuses instead of cleaning over an unproven PID. The split-brain-sensitive mismatch branch deserves isolated review, as #1028 asked. P3, matching #1025 — narrow trigger (PID reuse), strong test coverage.

PR #1137, opened 2026-08-31
fix(sidebar): clear the listening-port badge when the ports go away (#1135)

### LINKED ISSUE #1135: Sidebar listening-port badge stays stuck on a workspace after the port actually closes
### What happens

The sidebar's workspace-context line ("PID-tree-scoped listening ports", `src/renderer/components/Sidebar/WorkspaceItem.tsx`) keeps showing two ports on a workspace ("Product Ow…" — Product Owner) that are confirmed **not** actually listening anymore.

### Confirmed on the live machine (Windows 11, wmux 3.49.1)

```
PS> Get-NetTCPConnection -LocalPort 18856,63361 -State Listen
(no output)
PS> netstat -ano | Select-String ":18856 |:63361 "
(no output)
```

Neither port has a listener right now, yet the sidebar badge for those two ports is still showing under that workspace and won't clear (screenshot available on request — small enough that OCR of the row reads `:18856 :63361`).

### One suspicious path found, NOT confirmed as the cause

`src/main/pty/portWatch.ts`'s own doc comment on `snapshotWindowsNative`:

> "An unavailable native path REJECTS rather than resolving to an empty snapshot... `tick()` skips the diff pass entirely, so an already-rendered chip is left alone instead of being cleared by a phantom 'no ports' reading."

If the native FFI snapshot (`winSnapshotNative`, #1051/#1052) throws/rejects repeatedly on this machine, every tick would skip the diff and the chip would stay frozen indefinitely — matching this report exactly. I could not confirm this is what's actually happening: `%APPDATA%\wmux\logs\main-2026-08-31.log` has no `portWatch`/`snapshot unavailable` lines to check against (may not log through that channel, or the snapshot really is succeeding and something else — e.g. stale session/pid matching after pane churn — is the actual cause). Flagging the hypothesis, not asserting it.

### Impact

Minor/cosmetic (a stale badge, not a functional break) but persistent — nothing observed clears it short of restarting the app (untested whether restart actually clears it either).

### Confirmation

- [x] Searched existing issues for "listening port", "portWatch", "stuck badge" — no duplicate found.

openwong2kim:
Fixes #1135.

The sidebar's listening-port chip could outlive the ports — surviving even a full app restart. Three compounding leaks, all in the "how does this number ever go back down" direction:

1. **`listeningPorts` was persisted.** It rode along in workspace metadata into the session file and was restored verbatim on boot. Nothing ever contradicted it, because `PortWatcher` deliberately treats its first empty observation for a session as a no-op. This is what survived the reporter's v3.49.1 update restart.
2. **The workspace chip is a union over per-surface ports, and closing a pane/surface never recomputed it** — no surviving surface's update can subtract another surface's ports.
3. **A session whose process died was dropped silently** by the watcher, with no final empty emit.

Changes:
- `listeningPorts` is no longer persisted, and any already-saved value is stripped on session load (a stuck chip self-heals on first launch with this build).
- Both pane/surface teardown paths recompute the union via a new shared `recomputeWorkspacePorts`/`unionSurfacePorts` helper (single source of truth with the live write path).
- A vanished session emits one final empty port set before its diff state is dropped.
- Found while dogfooding: the daemon outlives the app, so a server already listening before an app restart was never re-announced to the new window (previously masked by the stale persisted value). `PortWatcher.resync()` now re-emits current port sets when a client subscribes.

Verified by unit tests (portWatch, store) and live dogfood in a sandboxed instance: kill-server / close-pane / restart-with-chip / hand-written stale session / app-only-restart-with-live-server all behave (clear ≤1 tick; recompute path 186 ms; resync 1 s).

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed stale listening-port badges that could remain visible after ports closed.
  - Cleared outdated port indicators when closing panes, surfaces, or sessions.
  - Prevented listening-port information from persisting across app restarts.
  - Restored accurate port indicators for servers already running when the app reconnects.

- **Tests**
  - Added coverage for session restoration, teardown, badge clearing, and port resynchronization.

### DIFF
--- a/changelog.d/1137.md
+++ b/changelog.d/1137.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- The sidebar's listening-port badge could stay lit long after the port
+  closed — even across a full app restart. The stale value was being saved
+  into the session file and restored verbatim, closing the pane that owned
+  the ports never recomputed the workspace badge, and a session whose
+  process died was dropped without a final "no ports" signal. All three
+  paths now clear the badge, an already-saved stale value is stripped on
+  load, and a server that was listening before an app restart is
+  re-announced to the new window instead of staying invisible. (#1135, #1137)
--- a/src/daemon/index.ts
+++ b/src/daemon/index.ts
@@ -2868,7 +2868,12 @@ function registerRpcHandlers(
   pipeServer.onRpc('daemon.events.subscribe', async (_params, ctx) => {
     // A false result is an expected first step of the compatibility handshake,
     // so do not warn here. DaemonClient identifies and immediately retries.
-    return { ok: pipeServer.subscribeEvents(ctx.clientId) };
+    const ok = pipeServer.subscribeEvents(ctx.clientId);
+    // #1135: a subscribing app has an empty context mirror — re-emit the live
+    // port sets so an already-listening dev server shows up without waiting
+    // for it to change.
+    if (ok) { try { resyncContextWatchers?.(); } catch { /* best effort */ } }
+    return { ok };
   });
 
   pipeServer.onRpc('daemon.events.unsubscribe', async (_params, ctx) => {
@@ -4815,6 +4820,8 @@ function wireContextWatchers(
     gitWatcher.update(s.id, s.cwd);
   }
 
+  resyncContextWatchers = () => { portWatcher.resync(); };
+
   return () => {
     sessionManager.off('session:created', onCreated);
     sessionManager.off('session:cwd', onCwd);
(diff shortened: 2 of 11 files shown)


openwong2kim:
This fixes #1135 by making the listening-port badge a live fact: `listeningPorts` is no longer persisted (stale saved values self-heal on load), pane/surface teardown recomputes the workspace union, and `PortWatcher` emits a final empty set plus a resync on subscribe. Touches the daemon event path, but changes are additive and unit/dogfood tested. P3 — cosmetic badge bug, and the fix includes its own migration.

PR #1138, opened 2026-08-31
fix(sidebar): drop the remote mirror on every activeWorkspaceId assignment path (#1086)

openwong2kim:
Closes the remaining click-return half of #1086 (bug 1). The rename/color/startup parity half (bug 2) stays open on the issue.

The reporter's root-cause analysis was correct and had already landed for one site (bb036898, #1089) — but the identical bare `!==` guard survived a second time in the same file, and two store sites violated the documented `clearRemoteSelection` contract:

- `focusNotificationTarget`'s workspaceId-only fallback (Fleet View cards that can't resolve by ptyId, workspace-level notification rows, OS toasts) silently did nothing while a remote mirror was showing and the target workspace was already active. The guard now also fires on `state.activeRemoteKey` (same fix shape as AT4).
- `companySlice.destroyCompany` / `removeDepartment` assign `activeWorkspaceId` directly when promoting a survivor and never called `clearRemoteSelection`, leaving the mirror painted while the sidebar highlighted the promoted local workspace.

Tests: new J7b case, symmetric destroyCompany/removeDepartment promotion cases (each fails on the pre-fix code in isolation). Live dogfood in a sandboxed instance with a real (seeded, boot-restored) remote attachment: all three scenarios pass, and an A/B run with only the production edits reverted reproduces both failures visibly.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed workspace navigation so selecting an already-active local workspace reliably switches away from a displayed remote mirror.
  * Corrected navigation from Fleet View cards, notifications, operating system alerts, and workspace promotion after company or department removal.
  * Ensured remote workspace selections are cleared when the active workspace changes.

### DIFF
--- a/changelog.d/1138.md
+++ b/changelog.d/1138.md
@@ -1,0 +1,8 @@
+### Fixed
+
+- While a remote workspace mirror was on screen, some jumps back to the
+  already-active local workspace silently did nothing: Fleet View cards and
+  notification rows that target a workspace rather than a pane, OS toast
+  clicks, and the promotion that happens when a company or department is
+  torn down all left the mirror painted while the sidebar highlighted the
+  local workspace. One click now always surfaces the local tree. (#1086, #1138)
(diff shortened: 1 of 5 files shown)


openwong2kim:
This fixes the click-return half of #1086: the workspaceId fallback in `useNotificationListener` now fires when a remote mirror is showing, and `companySlice`'s destroy/remove paths call `clearRemoteSelection` on promotion. Renderer-only with regression tests for each site, so risk is low. P3 — UI correctness fix; the rename/color parity half of #1086 stays open.

PR #1130, opened 2026-08-31
docs: start the #1111 legacy-lane deprecation clock (closes 2026-09-30+)

openwong2kim:
Stage 3 of the clientName deprecation (#1111) is gated on a notice period that had not started — nothing in the docs said the lane would ever end, so the clock stood at day zero. This PR is the notice; it changes no behavior.

## The announced terms

- **The `legacy` (no-`clientName`) lane closes in the first release on or after 2026-09-30.** A calendar date rather than a release count, because at the current cadence (several releases a week) "N releases from now" tells an integrator nothing.
- After the close, envelope-less requests are refused by the permission gate. `wmux-cli` remains available as the honestly-labelled limited lane (curated allowlist, no approval dialog) — per #1111's own framing, the unlimited lane disappears and a limited one remains.

## Where the notice landed

Every location #1111's audit lists as documenting the lane (one, `docs/api/inventory.md`, no longer mentions it — line drift since the audit): `PROTOCOL.md` (both mentions), `mcp-plugin-spec.md`, `reference.md` (via its generator, regenerated), `connect-to-wmux.md`, `build-a-substrate-plugin.md` (both mentions), `event-recorder/README.md`, and the two internal docs — `substrate-perf.md` now says the bench must migrate before the close, and `browser-tabs-workspace-contract.md`'s "no second deprecation clock here" note now points at the single shared clock.

The migration work itself (perf benches, the `--legacy` example flag, the two PermissionEnforcer enforcement points moving together) stays on #1111 and lands with the close, not now.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RRxH4qJUX5pkr2DpoHTAX6

## Summary by CodeRabbit

* **Documentation**
  * Announced the deprecation of legacy connections that omit a client name.
  * Legacy requests will be refused beginning with the first release on or after September 30, 2026.
  * Updated guides, API references, tutorials, and examples to recommend named client identities and capability declarations.
  * Clarified that `wmux-cli` remains available as a limited curated connection option.

PR #1139, opened 2026-08-31
fix(security): close the grandfather lane — envelope-less wire callers are refused (#1111)

openwong2kim:
Stage 3 of #1111. **DRAFT — do not merge, and do not include in any release dated before 2026-09-30** (the announced close date; notice was #1130).

Both `PermissionEnforcer` enforcement points move in one commit, per the issue's rule:
- the `!input.ctx.clientName` allow (no clientName envelope), and
- the `trust.status === 'legacy'` allow (grandfathered trust row).

Both now reject with `identity-status`/`status: 'legacy'`, no `pendingApproval`. Because the old no-clientName allow sat before the capability gate, closing it also closes the `wmux.internal` capability bypass (32 methods reachable envelope-less).

Exempt, because neither was ever the grandfather:
- **Trusted in-process surfaces** (`ctx.operator` — the renderer IPC bridge, `ctx.firstParty` — the iframe plugin host). `invokeRendererRpc` sends no clientName, so closing the lane flat would have rejected the entire UI under the production enforce default — a breakage the dev/test shadow default hides completely. Both markers are dispatch-option arguments, unforgeable from the wire.
- **The commander lane**, whose check now sits above the reject: its auth is the validated per-spawn token, not the clientName.

Second commit migrates every in-repo envelope-less caller (perf/substrate bench, 6 dogfood/verify scripts, the event-recorder example) to the `wmux-cli` curated lane; it can be cherry-picked into its own PR to land ahead of this one.

**Open decision before merge:** five methods the benches use are outside `WMUX_CLI_METHODS` (`events.poll`, `pane.getMetadata`, `pane.setMetadata`, `pane.close`, `mcp.claimWorkspace`). Extend the curated allowlist (widens the lane for anyone, since `wmux-cli` is non-identifying) or accept that those bench sections skip. Confirmed empirically in enforce-mode dogfood: exactly these five fall through to normal enforcement, nothing else fails.

Verification: mcp+pipe suites 1186 passed / 0 failed; tsc + eslint clean; enforce-mode dogfood on a packaged sandboxed build — 18/18 (UI fully alive through the operator lane, envelope-less wire refused with the new message, `wmux-cli` lane works).


openwong2kim:
This closes the #1111 grandfather lane: both PermissionEnforcer points reject envelope-less and `legacy` callers in one commit, with the operator/firstParty and commander lanes exempt. Benches and dogfood scripts move to the `wmux-cli` lane; the five-method allowlist gap is the one open call before merge. P2 — security hardening on the announced 2026-09-30 clock, not an active break.

openwong2kim:
**Retracting the three `high` findings from the AI review above (`daemon.shutdown`). All three are wrong, and they share one mistaken premise.**

The premise: `daemon.shutdown` maps to the reserved `wmux.internal` capability, is not in `WMUX_CLI_METHODS`, and therefore the dogfood cleanup RPC will be rejected by the enforcer once this PR closes the grandfather lane. The first two facts are right; the conclusion does not follow, because the enforcer never sees these calls.

`RpcRouter.ts:154-160` says so in the source:

> control-pipe-only RPCs (`daemon.*` / `lanlink.*`) are dispatched by the daemon control pipe, never registered here

Checked against this PR's head:

**`a2a-eventbus-dogfood.mjs:275` and `a2a-silent-default-dogfood.mjs:214`** — these call `rpcCall(pipeName, …)` with `pipeName = readDaemonPipeName()` (L271 / L213), i.e. the **daemon control pipe**, which has no enforcer. Nothing about this lane changes when the grandfather closes. Both also already have the fallback the finding says is missing: `app.proc.kill('SIGKILL')` (L266 / L208) and `process.kill(daemonPid)` → `SIGKILL` (L282 / L217). So neither "the sandboxed app is left running" nor "the temp HOME is non-removable" holds.

**`same-ws-a2a-dogfood.mjs:263` and `a2a-symmetric-reply-dogfood.mjs:336`** — these do go over the main pipe (`rpcCall` → `net.createConnection(mainPipe)` at L115 / L111). But `daemon.shutdown` has no main-pipe handler at all: the only registration is `pipeServer.onRpc('daemon.shutdown', …)` in `src/daemon/index.ts`, and neither `RpcRouter.ts` nor `PermissionEnforcer.ts` carries one at this head. Those two calls are therefore already dead today, swallowed by their own `.catch()` — a pre-existing cleanup wart, not a regression this PR introduces. Worth tidying on its own schedule; it is not a blocker and not `high`.

Also note all three were reported against lines outside the diff hunks, which is why they failed to post inline.

Unrelated to the above, the same run stopped early on `token_budget_reached` (1,545,050 / 1,500,000) while grouping the four dogfood scripts, so its coverage of this PR is partial.

Issue #1140, opened 2026-08-31
Remote pane (#1100): real-usage feedback — want split-into-new-pane alongside the tab, and roster/agent counting for a Claude session running remotely


p-poppe:
### Context

Dogfooding #1100 on a real paired VPS right after it shipped in v3.49.2. The tab-within-a-pane model works exactly as built and documented ("New remote pane" from a pane's ⋮ menu → new tab alongside local terminal/browser/editor/diff). Two pieces of real-usage feedback, not bug reports — the behavior matches what was deliberately built.

### 1. Want split-into-a-new-pane too, not only the tab

Local terminals can be split into their own pane (new region in the layout), independently resizable/arrangeable, in addition to living as tabs. A remote session today can only be a tab inside an existing pane's tab strip — there's no way to give it its own pane region the way a local terminal split does. Asking for parity: both affordances available, same as local, not a replacement for the tab option.

### 2. A remote pane running an actual Claude Code session isn't counted in the roster

`src/renderer/stores/selectors/workspaceAgentRoster.ts:82`:
```ts
if ((surface.surfaceType ?? 'terminal') !== 'terminal') return;
```
`'remote-terminal'` is excluded from the roster/"Agenci" count — consistent with how browser/editor/diff surfaces are already excluded, so this isn't a regression, just the existing convention applied to the new type. In practice, though, the whole point of a remote pane is often to run Claude Code (or another agent) on the far end — confirmed live: `/rename testremote` and Claude's own CLI chrome (weekly-limit banner, mode toggle) render inside it exactly like a local agent pane would. A genuinely agent-driving remote terminal gets none of the attention/status tracking a local one gets.

### Screenshots

Available on request — one showing the remote pane as a third tab (`w19-5`, `Claude Code`, `testremote`) inside workspace "COO", content showing live Claude Code CLI chrome running over the SSH-mirrored session.

### Confirmation

- [x] Searched existing issues — no duplicate; distinct from #1129 (session-lifecycle/close cleanup) and #1086/#1091 (already resolved by #1100).

openwong2kim:
## [wmux-hermes] Triage Summary
Dogfood feedback on the #1100 remote pane: split-into-new-pane parity with local terminals, plus roster counting for agents in remote-terminal surfaces (workspaceAgentRoster.ts currently excludes them). Both are enhancements to deliberate new behavior, not functional breaks. P3: parity and tracking polish, well-timed to fold into the #1129 remote-surface follow-ups.

[...]

PR #1141, opened 2026-08-31
feat(remote): split a remote session into its own pane, alongside the #1100 tab

p-poppe:
## Why

#1100 shipped remote-terminal as an ordinary Surface — a tab on an existing pane's tab strip, deliberately not a separate "attached remote workspace" entity (#1086/#1091). Real usage right after it shipped, on a paired VPS, surfaced one gap: local terminals can *also* be split into their own pane (a new layout region, independently resizable), and a remote session had no equivalent — only the tab.

Filed as #1140 with two asks. This PR is the first.

## What

**"Split right — remote" / "Split down — remote"** join the ⋮ menu, alongside the existing "New remote pane" (unchanged). Same `AddRemotePaneModal` (pick a host, mint a session) — the only difference is where the minted session gets attached:

- **New remote pane** (existing): the clicking pane, as another tab.
- **Split right/down — remote** (new): a freshly split pane, same as splitting a local terminal.

`resolveRemoteAttachPaneId` (new, pure, exported alongside the existing `pickSplitShownSurfaces`/`pickOverlaySurfaces`) is the whole decision:

```ts
export function resolveRemoteAttachPaneId(
  direction: 'horizontal' | 'vertical' | null,
  currentPaneId: string,
  splitResult: string | false,
): string | null {
  if (direction === null) return currentPaneId;
  return splitResult || null;
}
```

`null` direction is the #1100 tab flow, untouched. A direction targets whatever `splitPane()` already returned. `splitPane` creates an *empty* leaf pane — `EmptyLeafFunnel` would otherwise race to spawn a local PTY into it — so `addRemoteSurface` runs in the same synchronous tick as `splitPane` (no `await` between them), before React can commit and let the funnel's effect see an empty leaf. A blocked split (per-workspace pane cap, `splitPane` returns `false`) resolves to `null` — attach nowhere — rather than silently falling back onto the pane the user asked to split *away* from.

## What's NOT here

Issue #1140's second ask — count an agent-driving remote pane in the roster/"Agenci" — is not addressed. `workspaceAgentRoster.ts` keys agent detection on `ptyId`, fed by local PTY hook events (`PreToolUse`/`PostToolUse` etc. reported through the daemon's local IPC). Remote surfaces carry `ptyId: ''` by design (same as browser/editor), so there is no existing signal path a real agent running on the far end could report through — even removing the `surfaceType !== 'terminal'` filter changes nothing, since the very next line (`if (!ptyId...) return`) excludes it anyway. Building a heuristic on buffer content (scanning for CLI chrome, a rename command, etc.) would be fragile and guessed. Left for a direction call on #1140 rather than inventing a detection channel here.

## Test plan

- New: `resolveRemoteAttachPaneId` pure-function tests (`Pane.splitVisibility.test.tsx`) — tab flow ignores `splitResult`, split flow targets the new pane id, a blocked split resolves to `null` rather than falling back.
- Updated: `paneClusterWidth.test.ts`'s pinned menu-only-actions list now includes `split-right-remote`/`split-down-remote`.
- i18n: en/pl/zh (matching #1100's own scope for this feature's strings).
- `npx tsc --noEmit` clean, `eslint` clean on changed files.
- Full `npm run test:parallel`: 13195/13202 passed (7 pre-existing failures — playwright device presets, `deck.handler.loop` timing, `worktree.handler` git/EBUSY — reproduced identically on a clean `git stash` of this branch, unrelated to this diff).

## Summary by CodeRabbit

* **New Features**
  * Added options to open remote sessions in a new pane split to the right or below.
  * Remote sessions now attach to the newly created pane after splitting.
  * Added localized labels for the new remote pane actions.
  * Added a warning when the pane limit prevents creating another split.

* **Bug Fixes**
  * Improved cleanup when remote session attachment or pane closure is interrupted.

* **Tests**
  * Added coverage for remote session attachment and split-menu behavior.

### DIFF
--- a/changelog.d/1141.md
+++ b/changelog.d/1141.md
@@ -1,0 +1,3 @@
+### Added
+
+- **A remote session can now split into its own pane, not just a tab.** "Split right — remote" / "Split down — remote" join the pane's ⋮ menu alongside the existing "New remote pane" — same host picker, same minted session, attached to a fresh pane instead of another tab on the one you clicked from.
--- a/src/renderer/components/Pane/Pane.tsx
+++ b/src/renderer/components/Pane/Pane.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useState, useMemo } from 'react';
+import { useCallback, useEffect, useState, useMemo, useRef } from 'react';
 import { Panel, Group, Separator } from 'react-resizable-panels';
 import type { PaneLeaf, Workspace } from '../../../shared/types';
 import { maybeDelegateExternalBrowser } from '../../utils/browserPaneActions';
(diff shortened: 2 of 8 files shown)


openwong2kim:

Implements #1140's split ask — "Split right/down — remote" menu items attach a minted remote session to a fresh pane via `resolveRemoteAttachPaneId`. Touches Pane.tsx/SurfaceTabs.tsx and en/pl/zh strings; the EmptyLeafFunnel race is handled by keeping split and attach in one synchronous tick. P3 — additive parity feature, the existing tab flow is unchanged.

PR #1142, opened 2026-08-31
fix(updater): hide the install waiter behind a wscript launcher (#1136)

### LINKED ISSUE #1136: The cmd.exe/PowerShell install-waiter transport shows a visible Windows Terminal window despite windowsHide:true
### What happens

On update install (v3.49.1, Windows 11 with Windows Terminal set as the default terminal app), the waiter's cmd.exe trampoline (`spawnInstallWaiter`, `src/main/updater/installTeardown.ts`, transport A per #1059) shows a **visible** window — a real Windows Terminal window/tab (title bar reads `C:\WINDOWS\System32\Wind...`, empty black pane), sitting next to the (intentional, by-design) "wmux update / Installing the update…" toast.

### Screenshot

Windows Terminal window, tab titled `C:\WINDOWS\System32\Wind...`, empty terminal content, positioned at top-left of screen; the intentional "wmux update" toast is separately visible bottom-right. (Available on request — same screenshot used to diagnose #1084.)

### Why this looks like a real gap, not expected behavior

The spawn call already sets `windowsHide: true`:

```ts
const a = spawn(
  'cmd.exe',
  ['/d', '/c', powershell, ...psArgs, scriptPathA],
  { detached: true, stdio: 'ignore', windowsHide: true, cwd: systemRoot },
);
```

`windowsHide` maps to the classic `SW_HIDE`/`CREATE_NO_WINDOW` Win32 mechanism. On Windows 11, when **Windows Terminal is set as the system default terminal app** (Settings → Privacy & security → For developers → Terminal, or the classic `HKCU\Console\%%Startup` registry key), console-app creation is instead routed through Windows Terminal's own handoff/hosting mechanism — which is known to not always respect `CREATE_NO_WINDOW` from the spawning process, because it operates at a different layer than the traditional window-visibility flag. That would explain a window appearing despite `windowsHide: true` being set correctly in code.

### Impact

Cosmetic but confusing during an otherwise-silent, self-closing update flow — a flash of a real terminal window next to (or overlapping) the intentional toast, on a machine with Windows Terminal as default. Doesn't affect functional correctness of the update itself.

### Confirmation

- [x] Searched existing issues for "windowsHide", "Windows Terminal default", "cmd trampoline window" — no duplicate found.

openwong2kim:
Fixes #1136.

**Root cause (measured, not inferred).** `windowsHide: true` was never the deciding flag: libuv maps it to `CREATE_NO_WINDOW`, but `detached: true` also sets `DETACHED_PROCESS` — the two are mutually exclusive console-creation flags, and `DETACHED_PROCESS` wins. The detached `cmd.exe` trampoline therefore allocates its own console, the allocation goes through Win11's default-terminal delegation, and Windows Terminal opens a real window. A/B on a Win11 26200 box with WT default: same `cmd.exe`, same `windowsHide: true` — `detached: true` → visible `WindowsTerminal` window, without → none. Only the two detached waiter transports are affected; the module's non-detached `execFileSync` calls are untouched.

**Fix.** New transport W tried ahead of A/B: `wscript.exe //B //Nologo` running a two-line VBScript that starts the same PowerShell via `WshShell.Run(cmd, 0, True)`. wscript is a GUI-subsystem host — nothing reaches the delegation layer. `bWaitOnReturn = True` keeps wscript parked as the waiter's parent so the refusal path's `taskkill /T` still reaches the PowerShell child. The #1056 provable-launch property is preserved: W has its own launch stamp, script, and budget; A and B are byte-identical behind it (a machine with WSH disabled by policy falls through — a visible window is cosmetic, a refused install is not; `//B` keeps that fallback silent instead of popping WSH's error dialog). `conhost.exe --headless` was tried first and rejected: it exits 0 without ever launching the child, in every argument form.

Review hardening in the second commit: `'error'` listeners on all three transport spawns (an unlistened async spawn failure was an uncaught exception at install-handoff time), stamp waits gated on spawn success, stricter bare-argument validation in the VBS builder, the BOM as an explicit escape, and a WSH-availability skipIf on the runtime identity test.

**Evidence on the reporting hardware** (Win11 26200, WT default): transport W shows no visible window, stamps in ~440 ms, runs to completion, survives the parent's exit, and is reachable by `taskkill /T`; the unchanged cmd transport reproduces the visible window; direct-spawn B never stamps (reproducing #1056). With W force-disabled, the runtime identity test goes red and the cmd-trampoline end-to-end proof still passes — the fallback is intact. `src/main/updater` suites: 177 passed; tsc/eslint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Windows 11 installations with Windows Terminal set as the default terminal so no visible terminal window flashes during installation.
  * Improved updater fallback behavior when Windows Script Host is unavailable.
  * Prevented failed background launch attempts from interrupting the installation process.

### DIFF
--- a/changelog.d/1142.md
+++ b/changelog.d/1142.md
@@ -1,0 +1,7 @@
+### Fixed
+
+- On Windows 11 with Windows Terminal set as the default terminal, clicking
+  Install flashed a visible terminal window while the updater's install
+  waiter launched. The waiter now starts through a hidden GUI-subsystem
+  launcher that never allocates a console, with the previous transports kept
+  as fallbacks for machines where Windows Script Host is disabled. (#1136, #1142)
--- a/src/main/updater/__tests__/installTeardown.runtime.test.ts
+++ b/src/main/updater/__tests__/installTeardown.runtime.test.ts
@@ -36,6 +36,40 @@ function q(p: string): string {
   return `'${p.replace(/'/g, "''")}'`;
 }
 
+// #1136 — is Windows Script Host actually usable on this machine?
+//
+// WSH being disabled by enterprise policy is an explicitly SUPPORTED case:
+// the hidden transport fails and the cmd.exe trampoline behind it carries the
+// install (a visible window is cosmetic, a refused install is not). So the
+// "the hidden transport wins" assertion below is only meaningful where WSH
+// runs, and asserting it unconditionally would turn a correctly-working
+// fallback into a red test. Probed by actually running a script rather than
+// by reading the registry: the policy lives in two hives plus a per-extension
+// association, and only an execution answers the question the test is asking.
+function wshUsable(): boolean {
+  if (!onWindows) return false;
+  const dir = fs.mkdtempSync(path.join(os.tmpdir(), 'wmux-wsh-probe-'));
+  try {
+    const out = path.join(dir, 'ok.txt');
+    const vbs = path.join(dir, 'probe.vbs');
+    fs.writeFileSync(vbs, '\uFEFF' +
+      `Set fso = CreateObject("Scripting.FileSystemObject")\r\n` +
+      `fso.CreateTextFile("${out}").Close\r\n`, 'utf16le');
+    const r = spawnSync(
+      path.join(process.env.SystemRoot || 'C:\\Windows', 'System32', 'wscript.exe'),
+      ['//B', '//Nologo', vbs],
+      { windowsHide: true, timeout: 30_000 },
+    );
+    return r.status === 0 && fs.existsSync(out);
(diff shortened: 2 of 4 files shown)


openwong2kim:

Fixes #1136: the detached cmd.exe waiter flashes a Windows Terminal window because `DETACHED_PROCESS` overrides `windowsHide`; this adds a wscript launcher (transport W) that never allocates a console. Transports A/B remain as fallbacks and the VBS builder fails closed, keeping the install-handoff risk contained. P3 — the defect is cosmetic and the fix is well-tested with fallbacks intact.

PR #1143, opened 2026-08-31
fix(remote): closing a remote-terminal tab destroys the session it minted (#1129)

### LINKED ISSUE #1129: Remote-terminal surface follow-up: closing the tab never destroys the remote session, and one-shot remote-pane-* workspaces accumulate on the host
Follow-up debt from #1100 (remote-terminal as an ordinary Surface), declared "tracked, not blocking" in the merge verdict — this issue is that tracking.

## What's owed

1. **Explicit close must eventually destroy the remote session.** Today closing the tab only detaches (`paneDetach` on unmount); the session survives on the host. That is deliberate for reload re-attach — the session *must* survive unmount, so the destroy may only hang off an explicit user close, never the unmount path.
2. **One-shot `remote-pane-*` workspaces need cleanup.** `AddRemotePaneModal.pick` mints `remote-pane-<ts>-<rand>` per session; nothing ever deletes them, so repeated open/close accumulates workspaces on the remote host. Either cleanup on (1)'s destroy, or a daemon-side standalone-session concept.

## The primitive

`RemoteHostClient.closeSession` + `REMOTE_SESSION_CLOSE` from #1094 — extraction requested on that PR (its interactive-extension UI won't merge per the #1091 direction call, but this pair is exactly the cleanup primitive). Wiring destination: `handleCloseSurface`, per the #1100 verdict.

## Minor rider

`pane.newRemote` / `remote.noHostsHint` landed in en/pl/zh only — the other 20 locales fall back to English (#997-class). Worth folding into the next locale pass.

Refs: #1100 (verdict comment, Follow-up section), #1094 (extraction source), #1091 (direction call).

openwong2kim:
Fixes #1129 (follow-up to #1100/#1091).

**Root cause.** `Pane.handleCloseSurface` disposes `surface.ptyId` and nothing else, and a `remote-terminal` surface's ptyId is `''` by design — so closing the tab was detach-only. The `remote-pane-*` accumulation is the same bug seen from the host: the daemon derives `/api/workspaces` from live sessions' `WMUX_WORKSPACE_ID`, so the phantom workspace row lives exactly as long as the never-destroyed session. Ending the session reaps the row.

**Semantics.** Close destroys, unmount never does — wired into the explicit close paths only (tab ✕, Ctrl+W, prefix kill-pane/kill-workspace, workspace delete in sidebar and settings, the `surface.close`/`pane.close`/`workspace.close` RPCs), never a React cleanup, so remount/reload/tab-switch still re-attach. Only sessions this desktop minted are destroyed: a new `Surface.remoteOwned` flag set by the mint path and absent everywhere else, so a tab merely viewing someone else's session is non-destructive by construction. Decisions recorded in code: no confirmation prompt (local-pane semantics; ownership is the safety boundary), and sessions leaked before this fix are not auto-reaped (ownership can't be proven retroactively — clean host-side).

**Primitive.** `RemoteHostClient.closeSession` (`DELETE /api/sessions/:id`, the #1094 extraction the #1091 direction call asked for) behind a `remote:session:close` IPC. Destroy-first, detach-after: tearing streams down before the DELETE would be failure-destructive. Best-effort at call sites but observable — a refused/failed close warns with host and session id.

**Verified live** against a real paired headless host (`wmux web --allow-input`): mint → close via tab ✕ / Ctrl+W / sidebar workspace close / `workspace.close` RPC each destroyed the session and its `remote-pane-*` row on the host (host's own API read before/after each step); an unowned view survived its close while an owned tab closed in the same shot removed only its own session; host-offline close left the tab closed, one warning logged, and the session intact after the host returned. Suites: 88 files / 1,246 tests green; tsc/eslint clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Closing remote-terminal tabs, panes, workspaces, or derived workspaces now cleans up sessions created by the desktop.
  * Sessions started externally and viewed in the desktop remain available when their tabs are closed.
  * Failed cleanup attempts preserve active connections so they can be retried safely.
  * Missing or already-removed sessions are handled without disrupting the desktop.
  * Existing leaked sessions from earlier versions require manual host-side cleanup.

### DIFF
--- a/changelog.d/1143.md
+++ b/changelog.d/1143.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- Closing a remote-terminal tab only detached from the session, so the shell
+  it started — and the one-shot `remote-pane-*` workspace row derived from
+  it — kept running on the remote host forever. Closing the tab (or its
+  pane, or its workspace, by any route) now destroys the session it minted.
+  A tab that merely views a session started elsewhere is never destructive,
+  and sessions leaked before this fix are left for host-side cleanup, since
+  their ownership cannot be proven retroactively. (#1129, #1143)
--- a/src/main/ipc/handlers/__tests__/remote.handler.test.ts
+++ b/src/main/ipc/handlers/__tests__/remote.handler.test.ts
@@ -551,7 +551,7 @@ describe('remote.handler — sessionClose (#1129)', () => {
     expect(client.closeSession).toHaveBeenCalledWith('web-9');
   });
 
-  it('detaches every live stream on that session first — no reconnect target left behind', async () => {
+  it('detaches every live stream on that session once the destroy SUCCEEDS — no reconnect target left behind', async () => {
     const store = fakeStore([host]);
     const client = fakeClient(host);
     registerRemoteHandlers({ store: store as never, attachments: fakeAttachments() as never, clientFactory: () => client });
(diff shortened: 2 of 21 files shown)


openwong2kim:

Makes closing a remote-terminal tab destroy the session this desktop minted (#1129); views of foreign sessions stay non-destructive via the new `remoteOwned` flag. Touches every explicit close path plus a `closeSession` DELETE on `RemoteHostClient`; risk is bounded by destroy-first ordering, best-effort failure logging, and 1,246 green tests. P3 — tracked follow-up debt, a leak not a correctness or security break.

Issue #1107, opened 2026-08-29
Codex has lifecycle hooks now — verify, then stop guessing at Codex turn state


openwong2kim:
## The gap

For a Claude Code pane, wmux knows what happened because the hook bridge tells it: turn ended, permission prompt raised, subagent activity. For a Codex pane it does not. The only thing wmux installs into Codex is the `notify` bridge (`registerCodexNotify`, `src/shared/configIO.ts`), which is effectively a single completion ping. Everything else — turn boundaries, approval prompts, subagent starts — falls back to reading the screen: regex detection plus the byte-rate promotion from #935.

That fallback is why completion detection on Codex panes is a heuristic with tuned thresholds instead of a fact.

## What changed upstream

codex-cli 0.149.1 ships lifecycle hooks that look structurally like Claude Code's. Observed event names in the binary:

`PreToolUse` · `PostToolUse` · `Notification` · `UserPromptSubmit` · `SessionStart` · `SessionEnd` · `PreCompact` · `PostCompact` · `SubagentStart` · `SubagentStop` · `Stop`

with matching UI copy for each ("Before a tool executes", "When permission is requested", "After a tool executes", "Before/After context compaction", "When a new session starts", "Right before a session ends", "When the user submits a prompt", "When a subagent is created", "Right before a subagent ends its turn", "Right before Codex ends its turn"). The payload field names match too — `hook_event_name`, `stop_hook_active`, `permission_mode`, `tool_name`, `tool_input`, `agent_transcript_path`, `last_assistant_message`.

There is also a trust model around them: `[hooks]` in config.toml, `plugin_hooks`, `allow_managed_hooks`, and a review UI ("New hook - review required" / "Modified since last trusted" / "Managed hooks are always on"). Sources are enumerated as Admin config, User config, Project config, Session flags, Cloud-managed config.

If those behave the way the names suggest, the Codex side of completion detection stops being a guess.

## This is not verified yet

The above is **binary string inspection, not a working hook**. This repo has been wrong before by reading a competitor's surface and declaring a gap, so nothing here gets promoted to a track until someone actually installs a hook against a live Codex session and watches it fire. Blocking questions:

- What is the on-disk shape of a `[hooks]` entry in `~/.codex/config.toml`, and does it accept a command the way Claude's `settings.json` does?
- What exactly arrives on stdin per event, and is `Stop` a real turn boundary or a session-level signal?
[...]

PR #1144, opened 2026-08-31
feat(codex): report Codex turn boundaries from lifecycle hooks (#1107)

openwong2kim:
Addresses #1107. The spike is answered empirically, and the bridge is implemented — installation is deliberately not wired up (see below).

**How it was verified without a working Codex account.** A hook needs a completed *turn*, not a live account: codex-cli 0.151.0 was driven against a stub Responses endpoint on localhost with an isolated `CODEX_HOME`. The stub supplies only the model's side of the wire; the hook machinery under test is entirely local.

**Three of the issue's assumptions were wrong:**
- **Config is TOML, not `hooks.json`** — `[[hooks.<Event>]]` in `config.toml`. A `.codex/hooks/hooks.json` is inert.
- **Version floor is 0.141.0, not 0.149.1** (bisected). Worse, *no capability probe works*: 0.135.0 reports `hooks stable true`, accepts `--dangerously-bypass-hook-trust`, parses `[[hooks.*]]` clean — and fires nothing. Only the version distinguishes them, so `codexSupportsHooks()` gates on version and fails closed.
- **The trust gate is silent** — an untrusted hook does not run and `codex exec` prints no warning and exits 0.

**Confirmed:** `Stop` is a turn boundary, not a session one (fires once per turn with that turn's `turn_id`; a run whose model call failed produced SessionStart → UserPromptSubmit → SessionEnd and no `Stop`). The payload envelope is Claude Code's verbatim, `WMUX_PTY_ID` reaches the hook, and resume binding is possible (`source: "resume"`, stable `session_id`).

**Still blocked:** `PermissionRequest` — the event that would retire the three transcribed approval regexes in `AgentDetector.ts`. `codex exec` forces `approval: never`, so an approval pause needs an interactive TUI. Left unmapped rather than guessed at; the regexes stay, with a comment recording why.

**Bridge:** `SessionStart` / `UserPromptSubmit` / `Stop` → wmux signals, metadata-only (allowlist enforced in `buildCodexHookEnvelope`, asserted on the serialized envelope so a future content-bearing field fails the test). Identity follows the #1111 hook-bridge lane (`wmux-hook-bridge`, `hooks.signal` only), with a local lockstep guard; when #1111 lands, this bridge must be added to the canonical `hookBridge.lockstep.test.ts` BRIDGES list.

**A real bug found while wiring it up:** `hooks.rpc.ts` dropped an unclaimed `agent.user_prompt_submit` *before* the alarm feed, while the daemon path feeds it first. Since this bridge deliberately doesn't map PreToolUse, that is the only 'working' cue for a Codex pane — so the drop made CompletionAlarm's turn gate swallow the same turn's `agent.stop`, and it also blocked relaying to the daemon. Fixed with two non-vacuous regression tests.

Installation is not wired into the installer: Codex requires an operator to trust the hook and says nothing when they haven't, so an installer that wrote the block and reported success would be reporting a lie.

### DIFF
--- a/changelog.d/1144.md
+++ b/changelog.d/1144.md
@@ -1,0 +1,15 @@
+### Added
+
+- **Codex panes can report their turn boundaries from lifecycle hooks.** wmux
+  previously inferred whether a Codex pane was working by scraping its
+  terminal. A new hooks bridge lets Codex tell wmux directly when a session
+  starts, when a prompt is submitted, and when a turn ends. It needs
+  codex-cli 0.141.0 or newer, and the hook has to be trusted once in Codex —
+  see `integrations/codex/README.md` for setup.
+
+### Fixed
+
+- A lifecycle signal that arrived on the main pipe without a matching brain
+  pane was discarded before it could mark the pane as working, so the turn's
+  completion signal was discarded too — no notification, no lifecycle event.
+  (#1107, #1144)
--- a/integrations/codex/README.md
+++ b/integrations/codex/README.md
@@ -186,6 +186,25 @@ If you use this bridge, remove the `notify = [...]` line — otherwise every tur
 reports `agent.stop` twice. The `HookSignalRouter` dedup window swallows the
 duplicate, so nothing breaks, but the second spawn is pure waste.
 
+## Identity on the main pipe (#1111)
+
+Both bridges send `clientName: 'wmux-hook-bridge'` on every request. On the
+daemon control pipe it is ignored — there is no enforcer there. On the MAIN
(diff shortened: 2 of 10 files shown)


openwong2kim:

Implements #1107 as a Codex lifecycle-hooks bridge mapping SessionStart/UserPromptSubmit/Stop to agent signals, metadata-only, version-gated at 0.141.0, plus a hooks.rpc.ts fix for a dropped user_prompt_submit. Main risk is the #1111 coupling: once #1139 lands, this bridge must join the lockstep BRIDGES list or its signals are silently refused. P2 — closes a P2 gap with one cross-PR ordering to respect.

PR #1148, opened 2026-08-31
fix(remote): AddRemotePaneModal closes on Escape and names the flow that opened it

openwong2kim:
Two P3s surfaced by the #1141 dogfood run — both pre-existing from #1100, but the split flows tripled how often this dialog shows, so they are worth fixing now:

- **Escape closes the modal.** It binds the same document-level `keydown` listener `AttachRemoteModal` already uses; until now the backdrop click was the only way out.
- **The heading names the flow.** `AddRemotePaneModal` takes an optional `title` prop; `Pane.tsx` passes the menu label of whichever entry opened it ("Split right — remote" / "Split down — remote"), and the tab flow keeps the "New remote pane" default. No new i18n keys — the existing menu-label keys are reused.

Tests: two new cases in `AddRemotePaneModal.test.tsx` (Escape closes and the listener does not outlive the modal; caller-provided title renders). `tsc --noEmit`, ESLint, and the modal's 5 tests pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * The remote host-picker dialog can now be dismissed by pressing **Escape**.
  * Dialog headings now accurately describe the selected remote split action, such as “Split right — remote” or “Split down — remote”.

### DIFF
--- a/changelog.d/1148.md
+++ b/changelog.d/1148.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **The remote host-picker dialog closes on Escape and says which action opened it.** Escape now dismisses the dialog (backdrop click was the only way out), and its heading shows "Split right — remote" / "Split down — remote" when a #1141 split entry opened it instead of always reading "New remote pane".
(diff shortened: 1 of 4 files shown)


openwong2kim:
Remote host-picker modal (`AddRemotePaneModal`) now closes on Escape and its heading names the entry that opened it (tab vs split right/down). Renderer-only — a document keydown listener with cleanup plus an optional title prop wired from Pane.tsx, with tests for both. P3: small UX polish, no protocol or data-path risk.

PR #1149, opened 2026-08-31
feat(browser): frame-aware refs and iframe/OOPIF snapshot grafting

openwong2kim:
## What

`browser_snapshot` stopped at the iframe element and every ref resolved through `page.getByRole()`, which searches the main frame only. A payment form, a consent banner or a login widget was an invisible dead end: the agent could see that a frame existed and could not act on anything inside it. This makes frame contents reachable, and makes a ref minted inside a frame resolve there.

- **Frame-hop routes on every ref.** A ref carries a `FrameHop[]` route from the main frame — each hop the `<iframe>`'s document-order index among its host's frames, how many the host had, and the child document's URL. The ref itself stays an integer, so no tool gained a parameter and the MCP schema is byte-identical (probe unchanged across all three profiles). An empty route is the main frame, which is every ref that exists today.
- **Three fail-closed checks per hop.** The host must still hold the same number of frames, the frame at that position must still have a reachable document, and its live `url()` must still be the one the ref was minted against. Each failure is a `StaleRefError` naming what moved, never a substituted element. The last check is what catches a frame that navigated on its own, which no page-level URL comparison can see.
- **Same-process and out-of-process grafting.** Child trees are fetched by `frameId` and injected in place into the host's `<iframe>` node. A cross-origin frame is a separate CDP target, so it gets a session bound to its own target and its own password-field search. The route is built from what CDP and Playwright both see, cross-checked; if the two enumerations ever disagree, nothing is grafted. Bounded at depth 5, 100 frames, with a visited-frame guard.
- **One shared frame budget.** All grafted frames draw down a single pool of 40% of the snapshot, so a chatty ad frame cannot push the host document's own buttons past the truncation point, and a nested frame can only spend what its parent has not.
- **Boundary notes by reason.** An iframe now distinguishes not-attached, still-navigating, already-grafted, past the depth cap, past the frame budget, and read-but-empty. A frame that was never read keeps the original wording.
- **The RPC lane refuses a frame ref.** `data-wmux-ref` tags only the main document, so a frame ref there matches nothing — or worse, whatever an earlier DOM snapshot tagged with that number. Refused per browser surface, so one surface's frames never block another's DOM refs.
- **The page-facts footer stopped contradicting the tree.** pageFacts measures the main document, so a page whose controls all live in a frame read as "nearly empty" under a tree listing seven refs.

## Review & verification

Two-model panel raised 8 fixes and 2 documentation items, all applied. The panel disputed whether `DOM.performSearch` reaches a same-process iframe without `pierce`; that was settled by measurement rather than argument — against real Chrome, one search on the page session returned `resultCount 3` on a host+frame fixture (both of the child document's fields included), and a plaintext `autocomplete=new-password` field inside a frame comes back `[redacted:password]` end to end. No code change was needed; a regression test records the measurement. The other seven were real: a per-frame budget that was not a cap, a registry holding Pages as strong Map keys, an empty child URL that made mid-navigation the one moment a ref was accepted without proof, a globally-scoped RPC refusal, `parseInt` accepting `12abc`, a visited frame misreported as a failure, and a dead per-frame stamp write.

Live dogfood on an isolated instance (`WMUX_DATA_SUFFIX`), chrome backend, real Chrome, ref-based tool calls only — 7 scenes, all passing:

1. Two iframes loading one document — separate refs, and each click landed in its own frame (`clicked in slot 1` / `clicked in slot 2`), refs stable across three snapshots.
2. A genuinely cross-site frame (`https://example.com` inside `http://127.0.0.1`) — contents in the snapshot with working refs.
3. Two-level nesting — click into the innermost frame, depth continuing across the boundary.
4. Frame-only navigation — frame one's ref reported stale by name with both URLs, frame two's ref still clicked, page URL never changed; the re-snapshot renumbered frame one alone.
5. A dynamically inserted iframe — grafted on re-snapshot, pre-existing refs kept their numbers.
6. Credentials inside a frame, same-origin and cross-origin — both redacted, host field untouched.
7. A page with no iframes — refs, click, page scroll and the scrollable-container footer all unchanged.

Gates on every commit: `tsc --noEmit`, `build:mcp` + `probe:mcp` with the tool-schema hashes unchanged, and the full test suite. `snapshot.refstability.test.ts`, `snapshot.density.test.ts` and `snapshot.password.test.ts` pass **unmodified** — zero diff against `main` — which was the design's own regression gate. 20 new cases in `snapshot.frames.test.ts`, plus dogfood fixtures under `docs/dogfood/frames/`.

## Known limitations

Two cases where the checks are honest but not total, both documented next to the code that makes the trade:

- A frame **replaced** at the same position by the same URL reads as unchanged. All three hop checks pass, and the ref resolves inside a different instance of the same page, where role+name+nth is as good a locator as it ever was. Detecting the swap needs a per-document identity CDP does not expose to a locator walk.
- The walk is **not atomic** — the enumeration, the count cross-check and the per-frame trees are separate round trips. A page churning frames mid-capture can cost a re-snapshot.

Neither can produce a wrong element; both end in a stale-ref error.

## Attribution

`THIRD_PARTY_NOTICES` now names Stagehand (MIT) alongside browser-use for the frame-owning-session idea and the bounded-walk caps, with the borrow sites naming the file they mirror. No code from either project was taken; the selector-resolution and XPath machinery was deliberately not adopted.

## Release

No version bump. Entries live in `changelog.d/`.

## Summary by CodeRabbit

* **New Features**
  * Browser snapshots now include content from nested and cross-origin iframes.
  * Snapshot references can target elements inside frames and remain isolated per frame.
  * Frame content is bounded to prevent a single iframe from overwhelming results.

* **Bug Fixes**
  * Stale or invalid references are safely rejected after frame navigation or changes.
  * Snapshots now clearly indicate when iframe content is unavailable or still loading.
  * Passwords in framed content remain redacted.

* **Documentation**
  * Added browser fixtures and guidance for testing nested, dynamic, and cross-origin frames.


openwong2kim:
Frame-aware refs and iframe/OOPIF grafting for browser snapshots, with three fail-closed hop checks and a shared frame budget in `snapshot.ts`/`interaction.ts` (+2.1k lines). Main risk is the ref-resolution core; the unmodified refstability/density/password tests and 7-scene dogfood cover it well. Priority P2 — core automation path, worth a careful merge review of the budget and hop-check edge cases.

PR #1150, opened 2026-09-01
feat(browser): record successful actions and replay them deterministically

openwong2kim:
## What

An agent that repeats the same web flow — log in, filter a report, export a CSV — pays for a full accessibility snapshot on every step, every time, plus the reasoning to pick the right element out of it. This records a flow that worked and repeats it without reading a snapshot at all.

- **Automatic recording.** Every successful action from the ten replayable browser tools enters a 40-slot ring on the connection. There is no "start recording": an agent only learns a flow was the right flow *after* it worked, by which time a start call would have had to be made before there was any reason to make it — and an open recording session has no owner when the agent moves on.
- **One tool, four actions.** `browser_replay` with `list` / `save` / `run` / `forget`. One entry rather than four, because the full profile's `tools/list` payload is the first thing every host pays for.
- **The element axis is the snapshot's own RefEntry 4-tuple** (role, name, sameNameIndex, sameNameTotal) plus its frame key — not an XPath. It is what `resolveRef` already counts against, so a replayed step inherits the existing staleness contract instead of a second, weaker one. A page whose refs were renumbered by a restart, or restructured around the button, still replays.
- **Flows persist in main, not in the MCP session.** `browser.actionCache.{list,get,put,stats,forget}` over the existing broker, keyed by workspace. The MCP process dies with the agent's session — precisely when a flow starts being worth something. One workspace can never see another's.
- **A replay returns no snapshot.** It takes them internally to re-resolve elements; none are returned. That is the saving.
- **Self-heal, not silent failure.** A step whose element is gone stops the run there and reports which step, why, and how the page compares to the recording, leaving the page exactly there so the flow can be finished by hand and re-saved.
- **A landing hint.** After a navigation, one line names the recorded flows for that page — only proven, unquarantined ones. Deliberately not a snapshot footer, which would spend more context than the feature saves.
- **Variables.** A value that should vary is stored as `{{placeholder}}` and supplied at run time.
- **Credentials are never stored.** A `browser_type` or `browser_fill` into a password field is a marked hole whose value was never captured; `browser_fill` decides this per field, so a login form's ordinary fields still record normally. URLs carrying credentials are stripped and holed the same way. Coordinate clicks are not recorded at all — a coordinate does not survive a re-render.

Needs the chrome browser backend: flows are keyed on what the accessibility snapshot calls each element, and the builtin webview can fall back to a DOM snapshot that mints none.

## Review and verification

**Two-model panel — 13 findings, all fixed** across three commits: per-field password detection in `browser_fill`; URL credential stripping on both record and replay; `validateNavigationUrl` on replayed URLs (a stored URL is untrusted input by the time it runs); the recording ring made per-connection with no process-wide fallback and per-surface cut boundaries; replay refused from a page the flow was not recorded on; the cache RPCs fail-closed in both enforcement modes; a changed same-name population now refuses when the recorded element was not the first; a debounced-write race that dropped mutations; write retry with backoff; a shutdown flush; success history inherited only when the steps match; sanitiser holes instead of dropped steps.

**Live dogfood — 10 of 10**, against a real isolated instance (separate data dir, real Chrome, real daemon), driving the bundled MCP server over stdio:

| | | |
|---|---|---|
| 1 | open → navigate | chrome surface, landed |
| 2 | one snapshot → click / type / click | all three succeeded off a single snapshot |
| 3 | `save` | 4 steps, `email` variable detected |
| 4 | **full electron + daemon restart** | cache file md5 identical across it |
| 5 | `list` from a fresh MCP process | flow still there, ref maps long gone |
| 6 | `run` ×3 | 3/3, **no snapshot text in any result**, page state proved per-run variable substitution |
| 7 | button relabelled → `run` | stopped at the right step, named the reason, page left intact, failure counted |
| 8 | `forget` | removed from the file |
| 9 | password field via `browser_fill` | **plaintext absent from every file in the instance (grep)**; neighbour field recorded normally; the flow refuses to run |
| 10 | hint pipe | shown for a proven flow, silent with no flow, silent for a flow with a password hole |

Two defects the dogfood found are fixed in the last commit: the shape baseline and the live shape were measured at different points in the flow, so every successful replay reported a spurious mismatch. They are now measured at the same point, and the comparison is skipped rather than faked when either side is missing.

**Tests:** 172 cases across 8 files (6 new). Full suite green.

**Budget:** the full profile had 348 bytes of headroom — less than one tool's schema — so no tool could be added at all. Raised 75,000 → 78,000 in its own commit, with the reasoning recorded next to the budget's only consumer. Now 76,158 bytes across 93 tools. `core` and `commander` are byte-identical; the new tool is full-only.

## Attribution

Three operating principles of Stagehand's action cache (MIT) were adopted and recorded in `THIRD_PARTY_NOTICES`: best-effort (a cache miss may never break the underlying action), replay-fallback self-heal, and variable placeholders. No code was taken, and the axis is deliberately different — Stagehand keys on XPath, this keys on the snapshot's role/name/index tuple.

## No version bump

`package.json` is untouched; the changelog is a fragment under `changelog.d/`.

## Summary by CodeRabbit

* **New Features**
  * Added browser flow replay for recording, saving, listing, repeating, and deleting successful browser actions.
  * Flows persist across sessions and support variables for changing values during replay.
  * Provides clear stop reports when pages or elements no longer match, with manual fallback.
  * Displays replay hints after navigating to pages with available flows.
  * Passwords, credentials, and unsupported actions are excluded from recordings.

* **Documentation**
  * Added a comprehensive browser-flow replay guide and updated the API reference.


openwong2kim:
Adds `browser_replay` (save/run/list/forget): successful actions record into a per-connection ring and persist per workspace through five new `browser.actionCache.*` RPCs and a debounced atomic store. Risk contained — fail-closed workspace scoping, password/credential holes, 172 tests — but it is a large new persisted RPC surface, and dogfooding exposed #1151. P2: substantial and well-verified; merge alongside the #1151 McpRegistrar suffix fix.

PR #1153, opened 2026-09-01
feat(browser): promote proven flows into permanent, hint-surfaced skills

openwong2kim:
## What

Recorded browser flows (#1150) are a cache: they expire after 30 idle days, they can be pushed out by the per-workspace cap, and nothing surfaces one unless an agent thinks to call `browser_replay {action:"list"}`. That is the right contract for a cache and the wrong one for the few flows that have actually proven themselves.

This adds promotion.

- **`promote` / `demote` on the existing `browser_replay` action enum** — no new tools. `tools/list` is the first thing every host pays for, which is why the tool was one tool to begin with.
- **Permanent storage** in `~/.wmux/promoted-skills/<workspace>/<slug>.json`, one file per flow, main-owned (the MCP process dies with the agent's session; main is the long-lived single writer). `getWmuxDir()` folds in `WMUX_DATA_SUFFIX`, so an isolated instance gets its own tree.
- **A snapshot of the steps travels with the record**, so a promoted flow survives the expiry of the recording it came from. `run` falls back to that copy automatically and says it did. This is what makes promotion mean permanent rather than "listed for longer".
- **Push on navigation.** Landing on a promoted flow's page prepends a `[skill]` line carrying a one-line contract and a literal, runnable call, through the same hint pipe that already announces servable traces. Capped at the three most recently run flows plus a counted overflow line — a page with a dozen promoted flows would otherwise spend more context announcing them than the replay saves.
- **A stricter gate than the hint's**: `successCount >= 3`, not quarantined, no unreplayable step. A hint costs one attempt; a promoted flow is permanent and volunteered on every landing. A refusal names the shortfall ("2 successful run(s) and promotion needs 3") so the agent runs the flow again rather than giving up.
- **Lifecycle**: any run (successful or not) resets the idle clock; 30 idle days moves the flow to `~/.wmux/promoted-archive/`, and a later sweep deletes it from there. Nothing is deleted straight out of the live tree — every flow gets a stop in the archive first.
- **Injection defence.** The hint is instruction-adjacent text in the agent's context, so it admits exactly one page-derived string: the host, character-whitelisted and length-capped. The contract is re-derived at render time rather than read off the record. The flow name is validated against the trace-name alphabet (no quote, brace, backslash or newline) and a name that fails drops the whole line. Variable names are omitted unless they match the placeholder alphabet.

## Design note

The original design put promoted flows in `~/.claude/skills/` as `SKILL.md` files, so an agent would discover them through its own skill mechanism.

That was measured before it was built, and it did not hold up: empirically, agent-side skill auto-invocation proved unreliable — a user-source skill was discovered and matched but not actually invoked across 13 headless trials, with the harness varied four ways to rule out artifacts of the probe. Naming the skill explicitly worked every time, so the gap was autonomy, not discovery.

Discovery therefore rides wmux's own deterministic navigation hints instead. The mechanism is not a suggestion the agent may decline: it is the existing `prependReplayHints` pipe, already proven in #1150. Writing to a global, cross-instance directory is also dropped along with the approach, which removes a class of isolation and multi-instance concerns entirely.

## Review and verification

Two-model panel review produced 12 changes, the most serious of which were silent-failure bugs rather than hardening gaps:

- the 90-day deletion rung was unreachable, because archiving moves a file out of the live tree and only the live tree was swept;
- the orphan cleanup deleted any file it could not parse, including files written by a *newer* build — so running an older build once would have destroyed flows it could not read but the newer one could;
- restored runs lost the page-shape baseline, disabling the "the page changed underneath this flow" warning on exactly the runs most likely to be stale.

Live dogfood against an isolated instance (`WMUX_DATA_SUFFIX`, chrome backend, real Chrome, real pages) covered nine items, all passing:

1. record → save → 3 runs → promote, with the gate refusing at 0 runs and the record landing only under the suffixed path;
2. landing push fires on the promoted page and is silent elsewhere;
3. four promoted flows render as three lines plus `1 more`;
4. after the recording is forgotten, `run` restores from the promoted copy and succeeds — and is still refused from the wrong page;
5. `demote` succeeds with the browser surface closed (promote/demote deliberately run outside the automation lease: the moment a flow most needs demoting is when the session that recorded it has died);
6. hints and runs survive an app restart;
7. hand-poisoned records — a name carrying a JSON break-out, a contract carrying a directive, a spoofed host — produce no injected text; the poisoned record is dropped and the contract and host are re-derived;
8. a planted `version: 99` file is left byte-identical by the sweep, while a corrupt one is archived rather than deleted;
9. non-promoted pages keep their existing `[replay]` hint, and ordinary snapshot/click/save/run are unaffected.

Dogfood also found two defects, both fixed in the final commit: a promoted flow's `.bak` sidecars survived every removal path — holding a full copy of the steps, and readable through `atomicReadJSONSync`'s backup fallback, so a demoted flow could be resurrected — and one corrupt file printed a stack trace per read.

Tests: +139 (`13,653` passing). Token cost measured with `scripts/measure-mcp-tokens.mjs --compare`: `+62` tokens (`+0.3%`), full-profile wire `76,156` bytes against the `78,000` budget; the protocol baseline is regenerated for the new full-profile hash, and core/commander are byte-identical.

## Attribution

`THIRD_PARTY_NOTICES` credits two MIT projects whose approach was reimplemented and whose code is not bundled: **Hermes Agent** (Nous Research) for the lifecycle shape — usage counters kept beside the artifact rather than inside the cache that produced it, and an idle ladder that archives before it deletes — and **skill-forge** (BrowserAct) for the document shape, a short situational contract paired with a runnable invocation. The borrow sites carry the same attribution in a comment.

## Version

No version bump; one `changelog.d` fragment, per the repo's release policy.

## Summary by CodeRabbit

- **New Features**
  - Added `browser_replay promote` and `demote` actions for preserving proven browser flows.
  - Promoted flows remain available after recordings expire and can be restored when replayed.
  - Promoted flows appear as runnable suggestions when visiting relevant pages.
  - Added RPC methods for promoting, demoting, and listing promoted flows.
  - Added automatic archival after 30 days of inactivity and deletion after 90 days.

- **Documentation**
  - Added user guidance covering promotion requirements, storage, lifecycle, demotion, and captured values.
  - Updated the API reference and changelog.


openwong2kim:

This PR adds `promote`/`demote` to `browser_replay`, making proven flows permanent and hint-surfaced on navigation. The risk is instruction-adjacent hint text and permanent plain-text storage of typed values — both sanitised, tested, and documented. P3: self-contained feature, +139 tests, protocol baseline regenerated.

PR #1156, opened 2026-09-01
fix(settings): keep "Check for updates" visible while an install is staged

openwong2kim:
The update widget replaced the "Check for updates" button with "Install now" once a download was staged, so a newer release published afterwards was invisible until the 30-minute background poll caught up — observed today with 3.49.2 staged and 3.49.3 just released.

Both buttons now render when an install is staged; the check button is unconditional and the install button is gated on `downloaded`. Re-checking over a staged install is safe because a check always offers the latest.

Test pins the two-button layout (no ternary between them). No version bump; changelog fragment in `changelog.d/`.

## Summary by CodeRabbit

* **Bug Fixes**
  * The Settings panel now keeps **Check for updates** visible alongside **Install now** after an update has been downloaded.
  * You can check for newer releases without waiting for the background update check.

* **Tests**
  * Added coverage to verify both update actions remain available when an installation is staged.


openwong2kim:
Small Settings fix: "Check for updates" now stays visible next to "Install now" while an update is staged, so a newer release isn't hidden until the 30-minute poll catches up. Renderer-only (SettingsPanel.tsx), with a test pinning the two-button layout and a changelog fragment. P3 — minor UX polish, low risk.

PR #1157, opened 2026-09-01
fix(mcp): an isolated instance (WMUX_DATA_SUFFIX) no longer writes the production agent configs

### LINKED ISSUE #1151: Isolated instance (WMUX_DATA_SUFFIX) still writes the production ~/.claude.json MCP registration
Observed during live dogfooding of #1150 on an isolated instance launched with `WMUX_DATA_SUFFIX=-replay`:

```
[McpRegistrar] Claude Code: wrote wmux -> /Users/wong2kim/.claude.json
```

The isolated instance rewrote the real Claude Code MCP registration to point at the worktree build. Data dirs and the socket honor the suffix, but McpRegistrar targets the production config regardless, so a demo/dogfood instance silently retargets the user's daily MCP entry (it was restored by hand afterwards).

Expected: with a non-empty `WMUX_DATA_SUFFIX`, external registrations should either be skipped or written to a suffixed/scoped location — never the production `~/.claude.json`.

Repro: launch any build with `WMUX_DATA_SUFFIX=-x` and watch the McpRegistrar log line on startup. Likely fix area: src/main/mcp/McpRegistrar.ts (registration path selection does not consult the suffix).

openwong2kim:
Fixes #1151.

Every `MCP_TARGETS` config path (`~/.claude.json`, `~/.codex/config.toml`, `~/.gemini/settings.json`, plus Codex notify and the OpenCode plugin) is suffix-blind, so a `WMUX_DATA_SUFFIX` boot — dev's automatic `-dev` included — rewrote the user's production `wmux` entry to point at the isolated instance's bundle, silently retargeting their daily agents. The entry is only a script-path slot (which instance a pane talks to is decided by the `WMUX_DATA_SUFFIX` env the pane inherits), so the isolated instance gained nothing by claiming it — and when the checkout is a disposable worktree, the entry dangles after cleanup.

- `register()` / `forceUnregister()` skip all external targets when a suffix is set; the auth token keeps being written to its already-suffixed path. `WMUX_MCP_REGISTER_EXTERNAL=1` opts back in.
- The policy helper lives in `shared/mcpTargets.ts`; the daemon skips, while the `wmux mcp` CLI (an explicit user action) warns on stderr — in `--json` mode too — and proceeds.
- Settings → MCP Re-register/Unregister throw a visible error (toast) under an isolated instance instead of silently no-oping; first-run's MCP step short-circuits to ok before the production pre-flight.

**Review team (2-way, Codex 402-deactivated):** GLM-5.3 found no P1; its two P2s (CLI warning swallowed in `--json` mode; Settings buttons silently no-oping) and two of three P3s are folded in.

**Dogfood (isolated boot, `-dog1151`):** skip log line present, `Registered wmux MCP` absent, suffixed auth token created, production `mcpServers.wmux` byte-identical across the boot, codex/gemini configs untouched. The run also caught this bug in the wild: this morning's #1141 dogfood instance had already overwritten all three production configs with a since-deleted scratchpad path (now restored to the stable `~/.wmux/mcp` bundle).


openwong2kim:
Closes #1151: WMUX_DATA_SUFFIX boots no longer rewrite the production agent configs — register/unregister skip external targets, the CLI warns on stderr, and Settings surfaces a visible error. Tests cover the guard, the opt-back-in env, and unregister symmetry, so risk is low. P2 — this stops silent retargeting of production MCP entries.

PR #1165, opened 2026-09-01
Remove the six company_a2a_* MCP tools

openwong2kim:
## What

Removes the six `company_a2a_*` MCP tools (`whoami`, `send`, `broadcast`, `inbox`, `ack`, `status`) from every profile, and deletes the dead standalone server `src/company/mcp/index.ts` that shipped the same six names.

## What this costs

**Company mode's UI is unaffected** — the panel, spawn, and the `company.*` provisioning/mutation RPCs behind them are untouched.

**Company A2A messaging is not.** These six tools were the *only* caller of the `company.a2a.*` pipe handlers (`src/main/pipe/handlers/company.rpc.ts` → renderer `rpcHandlers`); no renderer, preload, or daemon path reaches them. So after this PR, member-to-member messaging (send / broadcast / inbox / ack / status) has no entry point. The pipe handlers are **retained, not deleted**, pending the Company mode re-evaluation.

Only two of the six have a real workspace-level equivalent:

| Removed | Replacement |
|---|---|
| `company_a2a_whoami` | `a2a_whoami` |
| `company_a2a_broadcast` | `a2a_broadcast` |
| `company_a2a_send` | **No equivalent** — resolved by department → lead, member name, or `"CEO"`; `send_message` / `a2a_task_send` only resolve workspaces. |
| `company_a2a_inbox` | **No equivalent** — the member inbox is a Company-mode store; `channel_read` / `channel_unread` read channels, a different store. |
| `company_a2a_ack` | **No equivalent** — `channel_ack` acknowledges a channel `seq`, not the inbox `messageId` this took. |
| `company_a2a_status` | **No equivalent** — returned the department tree with per-member roles; `a2a_discover` + `workspace_list` enumerate workspaces, not that graph. |

Department/CEO name routing, the member inbox, and the department tree were Company-mode-only features and are removed from MCP outright.

## Why

Nothing drove these tools. No prompt, skill, or documented flow called one, and `src/company/mcp/index.ts` had no bin or build entry — nothing imported it (only two stale comments referenced it, now fixed). Meanwhile each tool cost schema bytes in the `tools/list` that every session pays for before it does any work.

### Least-privilege follow-on

The `company.a2a.*` grants also leave `FIRST_PARTY_METHODS`. That set is a name-recognition bypass for reserved `wmux.internal` methods the bundled server actually calls; once no tool calls them, keeping the grants would leave stale attack surface for a `clientName` impersonator. This was not planned work — the existing least-privilege assertion in `firstParty.test.ts` failed and named the fix. The reserved methods the first-party lane still grants are `surface.list`, `surface.new`, `surface.close` and `browser.tabs`; comments and `docs/api/mcp-plugin-spec.md` §2.4 that used `company.a2a.*` as the example are corrected to that list.

## Result

| Profile | Tools | tools/list wire bytes |
|---|---|---|
| full | 93 → **87** | 76,410 → **73,610** |
| core | 47 (unchanged) | 46,665 |
| commander | 41 (unchanged) | 43,314 |

`core` and `commander` never carried these tools, so both are byte-identical to before.

The full-profile `maxListBytes` budget is deliberately left at 78,000; lowering the ratchet is a separate decision from removing tools.

## How tested

- `npx tsc --noEmit` — clean
- `npm run build:mcp && npm run probe:mcp` — green. Baseline regenerated the sanctioned way: `full.toolNames` pruned, then `full.wireResultSha256` updated from the probe's own failure output. `instructionSha256` is unchanged for all three profiles, and the probe's own subset/ordering/`company_`-prefix invariants all still hold.
- `npm test` — 891 files / 13,675 tests passed, 1 skipped; runtime suites 14 passed / 6 skipped.

Tests touched (all pre-existing, updated because they named these tools as examples rather than because new coverage was needed): `coreSurface.test.ts` (company_ prefix rule comment), `firstParty.test.ts` (`ALLOWED_RESERVED_FIRST_PARTY`), `PermissionEnforcer.firstParty.test.ts` and `RpcRouter.firstParty.enforce.test.ts` (swapped the `company.a2a.*` sample methods for `surface.new`/`surface.close`, which are the remaining reserved first-party methods).


openwong2kim:
Removes the six `company_a2a_*` MCP tools from every profile and deletes the dead standalone `src/company/mcp` server, pruning the matching `company.a2a.*` first-party grants. Four of the six have no workspace-level equivalent, so external callers of the documented lane break; docs and the protocol baseline are updated. P3 — deliberate cleanup of an experimental, undriven surface; company-mode UI and handlers untouched.

PR #1166, opened 2026-09-01
Give browser_smart_snapshot identity-stable refs and an auto-diff

openwong2kim:
## What

`browser_smart_snapshot` now numbers its refs off DOM node identity (Playwright/CDP lane) and returns a diff against its previous output instead of the whole indexed listing every time. `full: true` opts a call out.

Two ordered commits, because the second is only worth having once the first lands:

1. **Identity-stable smart refs.** Refs were a running 1-based count over the accessibility walk, so one node inserted above an element renumbered it and everything after it. An agent replaying a `smartRef` from the previous snapshot clicked the neighbour, with nothing to notice. Refs are now minted off `backendDOMNodeId` — the same identity `browser_snapshot` adopted in #1124 — so an unchanged node keeps its ref, and numbers are retired rather than recycled when an element goes away.
2. **Auto-diff.** Reuses the existing baseline store (`snapshotCache.ts`) and the existing formatter (`formatSnapshotResult` in `snapshotDiff.ts`) — no diff logic is duplicated — so the same 50% / 800-line fallback applies and the first line always declares whether a full listing or a diff came back.

## Why

`browser_snapshot`'s auto-diff only became worth adopting once refs held still: a renumber rewrites near enough every line, and nothing came in under the "is this diff actually smaller" ratio. `browser_smart_snapshot` was still on positional refs and still re-sending its whole listing, which on an act-then-verify loop is the same few hundred lines over and over.

Ref stability is also a correctness fix in its own right, independent of the diff.

### Lanes

Only the Playwright/CDP lane gets identity refs and diffing. The RPC lane (packaged build, no Playwright `Page`) keeps positional refs: the only place it could hold identity is the `data-wmux-ref` attribute, and `browser_snapshot`'s RPC fallback strips every one of those document-wide and renumbers from 0 on each of its own scans (`buildDomSnapshotExpression`). An interleaved `browser_snapshot` would therefore either wipe the identity or leave 0-based numbers behind for the next smart scan to adopt as its own. That lane always returns the full listing; it still writes its baseline, because the contract is a diff against the last listing the tool returned, whichever lane produced it.

### Two things worth a second look

- **`browser_click({ smartRef })` on the CDP lane was broken.** `getSmartSnapshot` stores a locator as the source snippet `getByRole('button', { name: 'OK' })`, and `interaction.ts` handed that string to `page.locator()`, which parses it as CSS. Resolution goes through a new `resolveSmartRefLocator`, which rebuilds the call through the real `getByRole` API pinned with `.nth()` to the instance the ref was numbered against. It had to change with `getLocatorByRef` anyway — the cache is no longer a dense `1..n` range, so `cache[ref - 1]` is wrong.
- **Baselines are namespaced by tool.** The store held one entry per surface, so the two snapshot tools evicted each other on a shared surface. The attrs guard kept the output correct, but alternating them would have meant neither ever diffed. The per-surface cap doubles (8 -> 16) so the split does not halve the surfaces a diff survives across.

`browser_extract_text` is deliberately untouched.

## How tested

- New `src/mcp/playwright/__tests__/dom-intelligence.refstability.test.ts` (20 cases), mirroring `snapshot.refstability.test.ts` and `resolveRefViaAxMap`: an insertion above leaves surviving refs alone, a removed node's number is never reissued, a different document does not rewind, a fragment change is the same document, the name filter is exact, an unnamed element indexes against the whole-role population, `.nth()` pins the right instance among duplicate role+name, lookup is by stored ref rather than position, a node with no `backendDOMNodeId` is skipped, a toggled element keeps its ref, one accessibility tree is read per snapshot, and every staleness case (navigated, reloaded, element gone, population resized, wrong page) is refused rather than resolved.
- New `src/mcp/playwright/__tests__/extraction.diff.test.ts` (11 cases): the first call is full and the repeat is a diff, an unchanged page says so, a URL change falls back to full, `full: true` forces full without switching the diff off, a changed `maxContentLength` does not diff across, the RPC lane never diffs and never leaves a baseline the CDP lane will use, a same-URL reload and a second tab on the same URL both fall back to full, and a diff over truncated page text says so.
- New `src/mcp/playwright/__tests__/interaction.smartref.test.ts` (3 cases): the click lands on the element the ref names, the recorded axis is a replayable ref axis rather than the `getByRole` source text, and a stale ref is reported instead of clicking a substitute.
- Full local run of the CI `validate` job: `tsc --noEmit`, relay `tsc --noEmit`, lockfile lineage guard (no-op), `npm run build:mcp && npm run probe:mcp`, `npm run build:daemon`, `npm test` — 13,711 parallel tests and 197 runtime tests pass.
- MCP protocol baseline regenerated for the one changed tool description and the added `full` param. Full profile: 76,563 bytes against the 78,000 budget (was 76,410).

## Review round

A three-model panel read the first two commits; the third commit answers it, and the fourth is a follow-up correction. Both are in the diff above.

- **A stable ref number is not a stable element.** Every check `resolveRefViaAxMap` makes for `browser_snapshot` is now mirrored: the name filter runs with `exact: true` (a ref for "Save" was matching "Save draft"), an unnamed element is counted against the whole-role population that `getByRole(role)` actually returns, and a ref whose page navigated or reloaded, whose element is gone, or whose population changed size is refused with `StaleSmartRefError`. A click aimed at a page other than the one the snapshot was taken on is refused too — ref numbers restart per page and the cache is per connection.
- **Recorded clicks are replayable.** A smartRef click stored the source text of a `getByRole` call as a css axis, which `page.locator()` cannot parse: the live click succeeded and every replay of that step failed. It records the same ref axis `browser_snapshot` does.
- **A diff never crosses a lane, a document, or a page.** The attrs key carries the lane and, on the CDP lane, a page token that changes on every main-frame navigation.
- **The identity map is trimmed at the cap, not on every snapshot.** An intermediate version pruned to the nodes each walk saw, which charged for the bound on every ordinary interaction: a menu that closes and reopens, or a row scrolled out of a virtualised list, came back with a new ref — a diff line per cycle. The trim now waits for `SMART_REF_IDENTITY_CAP`; over it, entries the walk did not see go first, and only a live page that alone exceeds the cap drops the map whole. Document boundaries still clear it outright.
- **Frames need no change here, and that is measured, not assumed.** `Accessibility.getFullAXTree` on a page target stops at the `<iframe>` element — `childIds: []`, the child document absent, same-origin or not (snapshot.ts, measured on Chrome 141). `browser_snapshot` reaches frame contents only by grafting them with an extra per-frame call, which this walk never makes. So every `backendDOMNodeId` it sees came from one document (the flat identity map is right) and its `sameNameIndex` population is the same set `page.getByRole` sweeps. A test pins the single `getFullAXTree` call.
- Smaller: nodes with no `backendDOMNodeId` are skipped rather than drawing a fresh number every snapshot, the population key is NUL-separated, and a diff over a page text cut at `maxContentLength` says the comparison is partial.

## Summary by CodeRabbit

- **New Features**
  - Smart snapshots now return indexed-listing diffs between calls, with complete-listing fallbacks when needed.
  - Use `full: true` to request a complete snapshot and refresh the comparison baseline.
  - Snapshot responses indicate whether they contain a full listing or a diff.
  - Smart references remain tied to page elements across updates and support replayable saved flows.
  - Clicking elements by smart reference now works reliably on live pages.

- **Bug Fixes**
  - Stale or ambiguous references are now rejected instead of targeting the wrong element.
  - Complete listings are preserved when element identity is unavailable.

### DIFF
--- a/changelog.d/1166.md
+++ b/changelog.d/1166.md
@@ -1,0 +1,19 @@
+### Changed
+
+- **`browser_smart_snapshot` returns a diff, and its refs hold still.** The
+  tool re-sent its whole indexed listing on every call, which on an
+  act-then-verify loop is the same few hundred lines over and over. A repeat
+  call now returns only what moved, against the same baseline machinery
+  `browser_snapshot` uses — with the same fallback to the full listing when the
+  diff stops being a saving, and a first line that always says which of the two
+  you got. Pass `full: true` for the complete listing.
+
+  What made that possible is a fix worth having on its own: smart refs were a
+  running count over the page, so one element appearing above another
+  renumbered it and everything after it. Replaying a ref you were given a
+  moment ago clicked the neighbouring element instead, and said nothing. A ref
+  is now tied to the element itself, so it means the same thing until that
+  element goes away — and it never comes back naming a different one. Clicking
+  by `smartRef` against a live page was broken outright and now works. In
+  packaged builds, where the underlying identity is not available, the listing
+  stays complete rather than diffed. (#1166)
--- a/changelog.d/1166.md
+++ b/changelog.d/1166.md
@@ -14,6 +14,12 @@
   moment ago clicked the neighbouring element instead, and said nothing. A ref
   is now tied to the element itself, so it means the same thing until that
   element goes away — and it never comes back naming a different one. Clicking
-  by `smartRef` against a live page was broken outright and now works. In
-  packaged builds, where the underlying identity is not available, the listing
-  stays complete rather than diffed. (#1166)
+  by `smartRef` against a live page was broken outright and now works, and a
+  ref that no longer names one element on the page you are looking at — the
+  page navigated or reloaded under it, the element went away, several elements
+  now answer to the same description, or the click is aimed at a different tab —
+  is refused with an explanation instead of clicking whatever moved into its
+  place. Saved flows record these clicks in a form that can actually be
+  replayed; before, the step always failed on replay while the live click
+  looked fine. In packaged builds, where the underlying identity is not
+  available, the listing stays complete rather than diffed. (#1166)
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 78000,
-      "wireResultSha256": "e82d7dfd4f92ab9425f69a2a5648f18c57ab64f9dbc4b56b948153d9311aa1a1",
+      "wireResultSha256": "d65b9f3f8bf957dd1095d1358bd3628911640c8c4595c6079669a68788811661",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
--- a/src/mcp/browser-replay/actionRing.ts
+++ b/src/mcp/browser-replay/actionRing.ts
@@ -10,6 +10,7 @@ import {
   refEntryToAxis,
   refMapShapeHash,
   stripUrlUserinfo,
+  type RefEntryLike,
   type ReplayableTool,
   type StepAxis,
   type TraceStep,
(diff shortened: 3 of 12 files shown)


openwong2kim:
Gives `browser_smart_snapshot` identity-stable refs keyed on `backendDOMNodeId` plus an auto-diff, and along the way fixes `browser_click({smartRef})` on the CDP lane, which had been feeding a `getByRole` snippet to `page.locator()` as CSS. Touches dom-intelligence, snapshotCache, and interaction, with the MCP baseline regenerated close to budget (74.3k/75k). P3 — correctness/efficiency work, well tested, no user-facing regression risk.

PR #1167, opened 2026-09-01
Opt into the core MCP tool profile, and stop overwriting the choice

openwong2kim:
## What

Makes the `core` MCP tool profile selectable, and stops automatic re-registration from overwriting the choice. The default stays `full`.

## Why

The `core` profile shipped in #1128 but nothing could select it. The `profile` argument stopped at `wmuxMcpEntry`; `upsertMcpServer` took no profile, so every registration path wrote the `full` entry. The old comment in `configIO.ts` admitted this and named the consequence:

> NOT YET THREADED THROUGH THE WRITERS. […] which also means a hand-added `--core` in a user's config is replaced on the next upsert.

So the profile was both unreachable through any supported surface *and* actively hostile to the workaround (hand-editing the config), because the next app boot reverted it.

## How

**Threading.** `profile` now runs `configIO` → `mcpRegistration` → callers. `DEFAULT_HOST_PROFILE` is the single constant deciding what a fresh registration gets.

It stays **`full`**, deliberately. The browser tools are first-class here — an agent driving the workspace browser is a headline capability, and it works with nothing to wire up precisely *because* the default registration carries `browser_*`. `core` is the opt-in for hosts that do not need that surface and would rather not pay ~27 KB of `tools/list` schema for it.

**Opt-in surface.** `wmux mcp register --profile core|full` (both `--profile core` and `--profile=core`) — one flag on the command that already owns registration. No new settings key (none exists for this today) and no settings UI.

**Preservation (the overwrite fix).** An explicit `--profile` always wins, so `--profile full` walks a config back off `core` — the opt-in is reversible. Omitting the flag means *"no opinion"*, and the entry then **keeps** the profile already on disk. Every automatic caller omits it (`McpRegistrar` on boot, bundle-path refresh after an upgrade), so a routine re-registration can no longer undo a user's `--core`. `--commander` survives the same path, even though no host-config writer ever emits it.

**Only recognised argv is rewritten.** `wmuxEntryArgs` replaces profile flags and passes **every other token through verbatim**. A token wmux did not write — a user's own addition, or a flag from a newer wmux sharing the config — is never dropped. This is what keeps the whole-array comparison below from turning "an arg I don't recognise" into "delete it on every boot".

**Idempotency.** `registerTarget`'s "already up to date" check compares the whole `args` array instead of `args[0]`. With the profile living in `args`, a path-only comparison would report no-op and silently drop a profile change. Combined with pass-through above, a repeat call is a true no-op rather than an oscillation.

**Visibility.** `readTargetStatus` and the register result now carry the profile, so `wmux mcp check` prints `registered (core) → …` (also in `--json`) and register reports which profile it wrote or preserved. Previously nothing could confirm a `--profile core` had landed.

### Behavior matrix

| `--profile` | Existing entry | Written |
|---|---|---|
| *(none)* | *(absent)* | `full` — the default |
| *(none)* | `--core` | `--core` — **preserved** |
| *(none)* | `--commander` | preserved |
| *(none)* | `--core --verbose` | `--core --verbose` — custom token kept |
| `core` | anything | `--core` + any custom tokens |
| `full` | `--core --verbose` | `--verbose` — only the profile flag drops |

## Known limitations

- **Registrar ↔ CLI read-modify-write race.** Two writers can still interleave on one config. Pre-existing (the writes are atomic per file, but not serialized against each other); file locking is out of scope for this PR.
- **`--profile` is `register`-only.** `unregister` and `check` ignore it rather than erroring.
- **`wmuxMcpEntry` remains** alongside `wmuxEntryArgs`. It is the simple "build a fresh entry" helper and has callers/tests; folding the two together is not attempted here.

## How tested

- `npx tsc --noEmit` — clean
- `npm run build:mcp && npm run probe:mcp` — green, unchanged: full 93 / 76,410 B, core 47 / 46,665 B, commander 41 / 43,314 B. This PR touches registration, not the surfaces themselves, so the baseline must not move — and does not.
- `npm test` — 891 files / 13,701 passed, 1 skipped (+26 over base); runtime suites 14 passed / 6 skipped.

Tests added:

`src/shared/__tests__/configIO.test.ts`
- defaults to the one profile constant every writer resolves through
- reads the profile flags back off an existing entry
- an explicit profile wins; omitting one preserves the existing entry
- carries unrecognized argv tokens through instead of dropping them
- de-duplicates a hand-edited repeated profile flag
- writes the default (full) entry when no profile is given
- writes `--core` when the caller opts in (json + toml)
- PRESERVES a hand-edited `--core` across a profile-less rewrite (json)
- PRESERVES a hand-edited `--core` across a profile-less rewrite (toml) — also asserts the surgical-TOML byte guarantee still holds
- PRESERVES `--commander`, which no host writer ever emits itself
- an explicit full profile walks a `--core` entry back

`src/shared/__tests__/mcpRegistration.test.ts`
- registers the full surface by default
- registers `--core` when the caller opts in
- a profile-less re-register PRESERVES an opted-in `--core`
- a profile-less PATH refresh keeps `--core` while updating the script
- switching profile is a WRITE even when the script path is unchanged
- reports the profile in the register result and in readTargetStatus
- PRESERVES a custom argv token across a profile-less re-register, and stays idempotent
- keeps a custom argv token when a PATH refresh rewrites the entry
- an explicit `--profile full` drops only `--core`, never a custom token
- (TOML target) registers `--core` when the caller opts in
- (TOML target) a profile-less PATH refresh PRESERVES `--core` in the surgical TOML rewrite

`src/cli/commands/__tests__/mcp.test.ts`
- is undefined when the flag is absent (= preserve what is on disk)
- parses both valid values, anywhere in the argv
- parses the attached `--profile=<value>` form
- returns null for an unknown or missing value so the caller can error out

### DIFF
--- a/changelog.d/1167.md
+++ b/changelog.d/1167.md
@@ -1,0 +1,24 @@
+### Added
+
+- **The `core` MCP tool profile can now be selected: `wmux mcp register --profile core`.**
+  The profile itself shipped earlier, but nothing could actually choose it — every
+  registration path wrote the full entry regardless. `core` drops the `browser_*`
+  family, cutting roughly 27 KB of schema off the `tools/list` an agent reads
+  before it does any work; it is an optimization, not a permission boundary, so a
+  core-mode agent keeps exactly the authority an ordinary one has. Use it for
+  hosts doing terminal, pane, workspace, channel, and delegation work that never
+  touch the browser. `--profile full` moves a host back.
+
+  The default is unchanged and stays `full`: the browser tools are first-class,
+  and an agent driving the workspace browser works with nothing to wire up
+  precisely because the default registration carries them.
+
+### Fixed
+
+- **A profile you chose is no longer overwritten the next time wmux registers.**
+  Re-registration happens on its own — at app boot, and whenever an upgrade moves
+  the bundle path — and it used to rewrite the entry as the full profile every
+  time, silently undoing a hand-added `--core`. Registration without an explicit
+  `--profile` now preserves whatever profile the entry already carries, and only a
+  brand-new entry gets the default. `--commander` entries are preserved the same
+  way.
--- a/docs/api/inventory.md
+++ b/docs/api/inventory.md
@@ -156,6 +156,18 @@ The full list lives in `src/shared/rpc.ts` (`ALL_RPC_METHODS`). For the MCP-faci
 
 The wmux MCP server (hosted in-process, named-pipe transport to the daemon) exposes a curated subset of the RPC surface as MCP tools. Tool names match `mcp__wmux__<tool>` when used from a Claude Desktop / Claude Code client.
 
+### Tool profiles
+
+The surface a server registers is chosen once, by a launch argument in the host config's `args` — never by an environment variable, so an env-stripping host cannot silently change it.
+
+| Profile | Launch arg | Contents |
+|---|---|---|
+| `full` | *(none)* | Every tool. **The default every registration writes.** |
+| `core` | `--core` | `full` minus `browser_*` (and any `company_*`), for agents that never touch the browser. Saves roughly 27 KB of `tools/list` schema per session. An optimization, not a permission boundary — a core-mode agent has exactly the authority an ordinary one does. |
+| `commander` | `--commander` | The orchestrator role surface. Set by the deck brain adapters at spawn time, not by host-config registration. |
+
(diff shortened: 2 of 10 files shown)


openwong2kim:

Makes the `core` MCP tool profile selectable via `wmux mcp register --profile`, and automatic re-registration (boot, path refresh) now preserves an existing profile instead of overwriting it with `full`. Touches configIO/mcpRegistration writers, the CLI, and McpRegistrar, with the whole-args idempotency check and TOML surgical-write guarantee tested. P3 — well-tested opt-in optimization plus a niche overwrite fix; no default behavior change.

PR #1172, opened 2026-09-01
Browser tools run their own page scripts in an isolated world

openwong2kim:
## Problem (measured)

Every script the browser tools inject ran in the page's **main world** — the
same JavaScript realm the site's own code lives in. Measured live against
`bot-detector.rebrowser.net` on Chrome 152 / playwright-core 1.58.2, driving
the exact snapshot-style DOM scan the tools perform:

```
🔴 sourceUrlLeak       Error stack contains UtilityScript.. You're using unpatched Playwright.
🔴 mainWorldExecution  You've called document.getElementsByClassName() in the main world.
typeof dummyFn = "function"     ← the page's own global, callable from our script
```

`dummyFn` is defined by the page. Being able to call it is the same fact from
the other side: the page can equally see and redefine everything our scripts
touch, so a hostile site can both observe the agent and hand it a doctored DOM.

## Design

`src/mcp/playwright/isolated-eval.ts`:

- **World.** One isolated world per page via CDP `Page.createIsolatedWorld`
  (`grantUniveralAccess: false` — parity with the main world, not extra
  privilege) on the page's main frame. The world name is
  `util` — bland on purpose; it never reaches the page, and nothing
  product-specific is written anywhere a devtools listener could read it.
- **Caching.** One long-lived `newCDPSession` per `Page`, plus the world's
  `executionContextId`, both cached in a `Map<Page, …>` and dropped on
  `page.on('close' | 'crash')`. Concurrent first calls share one creation
  round trip.
- **Invalidation.** The cached context id is cleared on
  `Runtime.executionContextsCleared`, `Runtime.executionContextDestroyed` for
  our id, and `Page.frameNavigated` for the **main** frame only. A call that
  still races a navigation retries once on a stale-context error.
- **Evaluation.** A function goes through `Runtime.callFunctionOn`; a string
  goes through `Runtime.evaluate({expression, contextId})`, because wrapping a
  string in `return (…)` would reject the multi-statement scripts
  `page.evaluate(string)` always accepted. Both with `returnByValue` and
  `awaitPromise`, and `NaN` / `±Infinity` / `-0` decoded back out of
  `unserializableValue`. `evaluateIsolated(page, fn|expression, arg?)` keeps the
  `page.evaluate(fn, arg)` / `page.evaluate(expression)` call and return shape,
  and re-throws a page exception with the page's own message.
- **Degradation.** If no isolated world can be had — no CDP session, or a
  target that refuses `createIsolatedWorld` — the call falls back to
  `page.evaluate` with the same arity. Losing the privacy property is bad;
  losing every browser tool is worse.
- `waitForIsolated` is the isolated-world stand-in for `page.waitForFunction`:
  a 100 ms poll, `timeout: 0` meaning "forever" as Playwright does, and the
  first evaluation allowed to throw straight through so a malformed predicate
  still fails fast instead of expiring as a timeout.

## Per-site migration

| Site | World | Reason |
| --- | --- | --- |
| `snapshot.ts` DOM fallback | isolated | DOM read + `data-wmux-ref` tagging; the DOM is shared |
| `pageFacts.ts` `collectPageFacts` | isolated | DOM/layout reads only |
| `page-eval.ts` `pageEvaluator` | isolated | `browser_extract_text` / smart-snapshot serialiser |
| `page-eval.ts` `evalFunctionOrRpc` | isolated | `browser_extract_data`; `arg` still the single parameter |
| `occlusion.ts` overlay probe | isolated | `Runtime.evaluate` now takes the world's `contextId`; the handles it returns still resolve through `DOM.describeNode` |
| `inspection.ts` `readDpr` | isolated | `devicePixelRatio` is a frame property, same value in either world |
| `inspection.ts` `browser_evaluate` | isolated (default) | new `mainWorld` opt-out, below |
| `interaction.ts` viewport bounds | isolated | `window.innerWidth/innerHeight` are per-frame |
| `interaction.ts` page scroll | isolated | `window.scrollBy` acts on the shared frame |
| `wait.ts` text predicate | isolated | reads `document.body.innerText`; a page that hooks it must not get to answer the poll |
| `wait.ts` custom `fn` predicate | isolated | same, with `page.waitForFunction(string)` semantics preserved |
| `interaction.ts` password-field check | **main** | element-scoped: an `ElementHandle` belongs to the world it was resolved in and cannot be adopted into an isolated context. Reads two properties off the given node, no page global |
| `interaction.ts` element scroll | **main** | element-scoped, same reason |
| `inspection.ts` `browser_highlight` | **main** | element-scoped; writes two inline styles the page sees in the DOM either way |
| `file.ts` file-input detection / proximity handle | **main** | element-scoped, and the *result* is a handle Playwright then passes to `setInputFiles`, so it has to live in Playwright's world |

## `browser_evaluate`

Runs in the isolated world by default. New optional `mainWorld: boolean`
(default `false`) routes back through the existing `evaluateWithGesture` path
for expressions that need the page's own globals (`window.__NEXT_DATA__` and
friends). One sentence added to the tool description.

`tools/list` after the change: **full 76,690 B** against a 78,000 B budget
(93 tools). `core` and `commander` are byte-identical — `browser_evaluate` is
full-only — so only the `full` `wireResultSha256` moved in
`scripts/mcp-protocol-baseline.json`.

## Verification (live)

Same launcher, same Chrome flags, same page, same script — once through
`evaluateIsolated`, once through `page.evaluate` as a control:

```
mode "isolated":  { "seesDummy": "undefined", "scanned": 94,  "gotNode": true }
⚪️ mainWorldExecution   Call document.getElementsByClassName('div') to trigger this test.
                        If you did and the test wasn't triggered, then you're running it
                        in an isolated world, which is safe and not detectable.
⚪️ sourceUrlLeak        Call document.getElementById('detections-json') to test sourceUrl leak.
🟢 runtimeEnableLeak     No leak detected.
🟢 pwInitScripts         No window.__pwInitScripts detected.

mode "main" (control):  { "seesDummy": "function", "scanned": 103, "gotNode": true }
🔴 mainWorldExecution   You've called document.getElementsByClassName() in the main world.
🔴 sourceUrlLeak        Error stack contains UtilityScript..
🟢 runtimeEnableLeak     No leak detected.
```

Both runs really made the calls the detector hooks (`getElementsByClassName`,
`getElementById('detections-json')`) and both got real answers back — the
isolated run simply was not seen. `runtimeEnableLeak` stays green in both.

## Out of scope

- **The RPC lane.** `page-eval.ts`'s `rpcEvaluator` (packaged builds, the
  Electron webview guest) has no Playwright `Page` to attach a session to, so
  it is untouched.
- **Bypassing `Runtime.enable`.** It measured green on Chrome 152 with stock
  playwright-core, so no patched fork and no bypass. Our session enables
  `Runtime` exactly as Playwright's own session already does.
- **Playwright's internals.** Its locator/click/fill machinery already runs in
  its utility world and is deliberately left alone.

## Tests

16 added: 14 in `__tests__/isolated-eval.test.ts` (world creation and its
parameters, the non-branded world name and `grantUniveralAccess: false`,
`arg` passing, context caching, invalidation on `executionContextsCleared`,
page-exception surfacing, main-world fallback arity, the four
`waitForIsolated` outcomes) (plus a multi-statement string, `unserializableValue` decoding, the
stale-context retry, that a failed world is retried after a navigation rather
than latched, and that a stale first poll is waited out) and 2 in
`__tests__/inspection.evaluate.world.test.ts` (`browser_evaluate` routing with
and without `mainWorld`). The 556 existing `src/mcp/playwright` tests stay
green unchanged.

## Review follow-ups (second commit)

- String scripts run through `Runtime.evaluate`, so multi-statement scripts
  keep working.
- A failed world creation is retried rather than latching the page into the
  main world for its lifetime; a fallback warns once per page.
- The occlusion probe reuses the module's cached per-page session — a session
  per snapshot minted an isolated world per snapshot, which accumulates in a
  long-lived SPA's renderer. Measured: three consecutive probe requests return
  the same `contextId` (5, 5, 5). The probe's `Runtime.evaluate` carries
  exactly one execution-context field (`contextId`).
- `unserializableValue` (`NaN`, `±Infinity`, `-0`) decoded instead of dropped.
- `waitForIsolated` waits out a stale context on its first poll; a malformed
  predicate still fails fast.
- Close/crash detaches the session, and an already-closed page is dropped at
  open time.
- An invalidation during an in-flight creation no longer caches the old
  document's context id (epoch check).
- `grantUniveralAccess: false`.
- The RPC lane (no Page) appends "ran in the page's main world: this backend
  has no isolated world" and the tool description now says "On the Chrome
  backend it runs in an isolated world…".

### DIFF
--- a/changelog.d/1172.md
+++ b/changelog.d/1172.md
@@ -1,0 +1,9 @@
+### Changed
+
+- **Browser tools now run their own page scripts in an isolated world.** A
+  website can no longer see the scripts the agent injects to read the page, nor
+  hand them doctored results by redefining the DOM methods they call — the
+  snapshot, extraction, wait and scroll scripts share the page's DOM but not
+  its JavaScript, so they are invisible to it. `browser_evaluate` runs there
+  too by default and gained a `mainWorld` option for the expressions that
+  genuinely need the page's own globals. (#1172)
--- a/changelog.d/1172.md
+++ b/changelog.d/1172.md
@@ -6,4 +6,5 @@
   snapshot, extraction, wait and scroll scripts share the page's DOM but not
   its JavaScript, so they are invisible to it. `browser_evaluate` runs there
   too by default and gained a `mainWorld` option for the expressions that
-  genuinely need the page's own globals. (#1172)
+  genuinely need the page's own globals — as does the backend that has no
+  isolated world to offer, which now says so in its answer. (#1172)
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 78000,
-      "wireResultSha256": "d65b9f3f8bf957dd1095d1358bd3628911640c8c4595c6079669a68788811661",
+      "wireResultSha256": "db19fc959c10612d3c8dc1a00e9e46c319b22e9db4dfaaba7397264c666919ef",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
--- a/src/mcp/playwright/__tests__/inspection.evaluate.world.test.ts
+++ b/src/mcp/playwright/__tests__/inspection.evaluate.world.test.ts
@@ -1,0 +1,75 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+// Which world browser_evaluate runs the caller's expression in. The default is
+// the isolated one — the page must not be able to watch or doctor an agent
+// script — and `mainWorld:true` is the deliberate opt-out for expressions that
+// need the page's own globals (window.__NEXT_DATA__ and friends).
+
+const { getPage, resolveWorkspaceBackend, isolated, gesture } = vi.hoisted(() => ({
+  getPage: vi.fn(),
+  resolveWorkspaceBackend: vi.fn(),
+  isolated: vi.fn(async () => 'from-isolated'),
+  gesture: vi.fn(async () => 'from-main'),
+}));
+
+vi.mock('../../wmux-client', () => ({
+  sendRpc: (method: string) =>
+    method.startsWith('browser.lease.') || method === 'browser.lifecycle.get'
+      ? Promise.resolve({ token: null })
+      : Promise.resolve({}),
+}));
+
+vi.mock('../PlaywrightEngine', () => ({
+  PlaywrightEngine: {
+    getInstance: () => ({ getPageForScope: getPage, resolveWorkspaceBackend }),
+  },
+}));
+
+vi.mock('../snapshot', () => ({
+  resolveRef: vi.fn(),
+  generateSnapshot: vi.fn(),
+  generateScopedSnapshot: vi.fn(),
+  markDomRefsActive: vi.fn(),
+}));
+
+vi.mock('../isolated-eval', () => ({ evaluateIsolated: isolated }));
+vi.mock('../user-gesture', () => ({ evaluateWithGesture: gesture }));
+
+import { registerInspectionTools } from '../tools/inspection';
+
+type ToolResult = { content: Array<{ type: string; text?: string }>; isError?: boolean };
+type ToolHandler = (args: Record<string, unknown>) => Promise<ToolResult>;
+
+const deps = { resolveWorkspaceId: vi.fn(async () => 'ws-test') };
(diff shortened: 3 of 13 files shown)


openwong2kim:

Moves the browser tools' page scripts (snapshot, extraction, wait, scroll) into a CDP isolated world so pages cannot see or tamper with them; `browser_evaluate` gains a `mainWorld` opt-out. Touches the full `src/mcp/playwright` surface (13 files, +691/−29) with a main-world fallback and 10 new tests. P2 — broad behavioral change to every browser tool, but live-verified with safe degradation.

PR #1173, opened 2026-09-02
Make agent browsing behave less mechanically

openwong2kim:
Five behavioural fingerprints in the browser tools. All of them are about what a
page can *observe* while the agent works — not about what the tools say about
themselves, and not about hiding that automation is present.

## Per item

| # | Item | Files | What changed | Test |
|---|------|-------|--------------|------|
| 1 | Typing delays were uniform-random | `src/shared/humanRhythm.ts` (new), `src/mcp/playwright/human-typing.ts`, `src/main/browser-session/HumanBehavior.ts` | Both lanes drew from a flat `Math.random() * (max-min) + min` over 50–150 ms. One shared definition now produces a log-normal base around the band's midpoint, an allowance after sentence (`+120 ms`) / clause (`+60 ms`) / word-break (`+40 ms`) characters, and a ~4% chance of a 300–700 ms pause. Clamped to `0.6×min … 5×max`. RNG is injectable (defaults to `Math.random`). Public option names `minDelay`/`maxDelay` unchanged. | `src/shared/__tests__/humanRhythm.test.ts` — median band, skew (mean > median), pause frequency band, clamp, punctuation ordering, seeded determinism, 40-char total-time band, custom band |
| 2 | Clicks were instantaneous centre hits | `src/shared/pointerPath.ts` (new), `src/mcp/playwright/pointer-path.ts` (new, per-page tracking), `src/mcp/playwright/tools/interaction.ts`, `src/main/pipe/handlers/browser.rpc.ts` | Chrome lane: before clicking, the pointer walks from its last position on that page (or a plausible viewport-relative start) to a landing point inside the inner 60% of the element box, with `stepsForDistance()` giving 8–25 points. Playwright still performs the click itself via a `position` offset, so actionability / `force` / `timeout` semantics are untouched. RPC lane: `browser.click.cdp` now emits the same interpolated `Input.dispatchMouseEvent mouseMoved` run before `mousePressed`. Plain interpolation with perpendicular jitter — nothing more is claimed. | `src/shared/__tests__/pointerPath.test.ts` — monotone progress along the axis, exact endpoint, step-count band, off-line mid-flight, zero-length move, click point inside the inner fraction, non-centre |
| 3 | `isTrusted=false` on hover / drag in the RPC fallback | `src/mcp/playwright/tools/interaction.ts`, `src/main/pipe/handlers/browser.rpc.ts`, `src/shared/rpc.ts`, `src/main/mcp/firstParty.ts`, `src/main/mcp/methodCapabilityMap.ts` | Hover and drag dispatched synthetic `MouseEvent`/`DragEvent` via `evaluate`, so every handler saw `isTrusted === false`. Two new verbs — `browser.hover.cdp` (interpolated `mouseMoved`) and `browser.drag.cdp` (move → `mousePressed` → held-button moves → `mouseReleased`) — route both through CDP `Input.*` instead. Both map to the existing `browser.click` capability. | Covered by the existing `interaction.fallback` / `interaction.newtab` suites plus `tsc`; the CDP verbs themselves are exercised by the live smoke's chrome-lane equivalent |
| 4 | No `Referer` on agent navigations | `src/shared/referer.ts` (new), `src/mcp/playwright/tools/navigation.ts`, `src/main/pipe/handlers/browser.rpc.ts` | When the page being left is a real http(s) document and the destination differs, `page.goto` gets `referer:` and `webContents.loadURL` gets `httpReferrer:`. `about:blank`, `chrome://`, `devtools://`, `file://`, `data:`, an unparseable URL, a first navigation, and a same-URL reload all still send none. | `src/shared/__tests__/referer.test.ts` — 6 cases covering each branch |
| 5 | `browser_emulate` UA inconsistency | `src/shared/uaMetadata.ts` (new), `src/mcp/playwright/tools/state.ts`, `src/main/pipe/handlers/browser.rpc.ts` | The UA string was overridden alone, so `navigator.userAgentData` and the `Sec-CH-UA*` headers kept answering out of the real browser — an emulated phone announced itself as a phone in one place and as this desktop in the other. `Emulation.setUserAgentOverride` now carries `userAgentMetadata` (brands at the UA's major version, platform, platformVersion, architecture, model, mobile) and `acceptLanguage`, derived from the preset's own UA. Applied on the preset path **and** on device reset, so a reset does not leave the preset's hints behind. Only the emulate path is touched; the default non-emulated UA is unchanged. | `src/shared/__tests__/uaMetadata.test.ts` — desktop Chrome on macOS and Windows, iPhone 13 (no brands, since Safari exposes no Client Hints), Pixel 5, Galaxy S9+ (Build token stripped from the model), acceptLanguage, and a mobile-flag consistency sweep |

`browser_select` in the RPC lane deliberately keeps the DOM assignment: a native
`<select>` opens an OS-drawn popup that lives outside the page, so CDP `Input`
events go to the document and never reach it. There is no mouse sequence that
picks an option reliably. The code now says so, and that the resulting `change`
event carries `isTrusted === false`.

## Live smoke

Real Chrome via `connectOverCDP`, pages served off a local http server (`data:`
URLs cannot carry a Referer), driving the new helpers directly.

**(a) Typing — 40 chars into a `<textarea>` recording `keydown` timestamps**

```
text (40 chars): "the quick brown fox jumps over lazy dog."
observed keydown intervals (ms):
  121.4, 134.4, 102.6, 122, 116.8, 163.5, 99.6, 119.4, 119.5, 124.1, 83.2,
  115.2, 171.4, 102.6, 111.1, 155.5, 100, 154.2, 119, 149.1, 177.6, 58.3,
  142.1, 673.6, 231.9, 226.3, 133.7, 82.3, 116.8, 145.8, 112.9, 116.3, 63.6,
  89.4, 88.5, 134.7, 512.3, 104.3, 67.8
  count=39  median=119.4  mean=147.8  max=673.6  min=58.3
  total wall time = 6.60s
  textarea value matches typed text: true
```

Mean above median, two thinking pauses (673.6 ms, 512.3 ms) — the shape the old
uniform draw could not produce, where every interval sat inside 50–150 ms. Wall
time is a touch over the 3–6 s band on this sample because two pauses landed in
one 40-char run and each keypress also costs a CDP round trip; the unit test
asserts the band on the schedule itself, averaged over 300 seeds.

**(b) Click — page counting `mousemove` and recording `isTrusted` + offset**

```
  box = {"x":400,"y":300,"width":200,"height":60}
  pointer start = (410, 202)  ->  landing = (505.5, 327.8)  steps = 12
  mousemove events the page saw = 13   (a bare locator.click() produces 1)
  click.isTrusted = true
  offset from centre = (7, -1) px   inside box = true
  inner-60% bound = (+/-60, +/-18)
```

**(c) Navigation — `document.referrer` on B after A → B**

```
  refererFor('about:blank', 'http://127.0.0.1:60401/a') = undefined
  after goto A, document.referrer on A = ""
  refererFor('http://127.0.0.1:60401/a', 'http://127.0.0.1:60401/b') = "http://127.0.0.1:60401/a"
  on B: referrer=http://127.0.0.1:60401/a
  reload of B sends referer = undefined  (same URL -> none)
```

## Known limitations

- **`browser_select` in the RPC lane stays untrusted.** Explained above; there is
  no `Input.*` route to a native `<select>` popup.
- **The pointer's remembered position is per page and in-process.** A page that
  is replaced, or a fresh MCP process, starts from the default start point
  rather than from where the pointer really was.
- **`Emulation.setUserAgentOverride` on the chrome lane is best-effort.** If the
  CDP session cannot be opened, the UA header is still applied and the result
  reports `clientHints=unavailable` rather than failing the call.
- **Brand grease is a fixed placeholder.** Real Chrome varies the greased brand
  entry between releases; a stable value keeps the builder deterministic.

## Out of scope (untouched, deliberately)

Electron webview UA string; CAPTCHA; fingerprint spoofing (canvas / WebGL /
fonts); `Runtime.enable` bypass; headless mode; and anything that forges
automation signals to get a login past a site's own protections — the repo rule
in the header of `src/mcp/playwright/user-gesture.ts`.

## Gates

- `npx tsc --noEmit` clean
- `npm run build:mcp && npm run probe:mcp` clean; **no tool description or input
  schema changed**, so `scripts/mcp-protocol-baseline.json` does not move.
  `tools/list` `full` = **76,410 B** (cap 78,000), unchanged.
- `npx vitest run src/mcp/playwright src/main/browser-session src/shared/__tests__`
  → 123 files, 1743 passed, 4 skipped. **36 new tests** across 4 new files.
- `eslint` clean (0 errors) on every new and changed file.

---

# Review round

Findings from the review panel, and what each turned into.

| # | Finding | Outcome |
|---|---------|---------|
| 1 | `refererFor` sent the full URL everywhere | **Done.** Now mirrors Chrome's default `strict-origin-when-cross-origin`: same-origin → full URL, cross-origin → `origin + '/'`, https→http → nothing, credentials and fragment always stripped, reload compared on normalised URLs. |
| 2 | Emulate: session detached in `finally`; only the current page; reset did not undo the UA | **Done.** New `src/mcp/playwright/ua-emulation.ts` holds one CDP session open per page (the override dies with its session), re-applies on `context.on('page')`, and `device: null` now restores the real UA + metadata and drops the `User-Agent` extra header. Verified live below. |
| 3 | No total-duration budget on a typing schedule | **Done.** Capped at 120 ms/char + 1.5 s slack, scaled down proportionally when exceeded — the shape survives the squeeze. Tested with a 2,000-char string. |
| 4 | `pointerPositions` never pruned | **Done.** Entry dropped on the WebContents `destroyed` event; the misleading "pruned on move" comment is gone. |
| 5 | `brands: []` for Safari presets is a novel fingerprint | **Done.** Non-Chromium presets omit `userAgentMetadata` entirely. Chromium presets gained `fullVersionList`, `bitness`, `wow64`. |
| 6 | `clickWithApproach` hardening | **Done.** Scrolls before measuring; centre-aims controls under 24 px; skips the approach when the target is outside the viewport; retries once without the approach on failure (which also covers rotated elements). New `interaction.clickApproach.test.ts`, 10 cases. |
| 7 | `elementCenter` zero-size rects; `buttons: 1`; verify the target after the move | **Done.** Zero-area rects resolve as not-found; `mousePressed` carries `buttons: 1`; `approachElement` checks `document.elementFromPoint` after the walk, recomputes once, then fails explicitly rather than reporting a silent mis-click. |
| 8 | Raw mouse events cannot drive HTML5 drag without `Input.setInterceptDrags` | **Rejected — measured false.** See the raw-CDP smoke below: on Chrome 145, a press/move/release sent from a raw CDP session with no interception fires `dragstart` **and** `drop` on a `draggable` fixture. An implementation that branched to `DragEvent` synthesis for draggable sources was written, measured, and then removed: it would have traded real input for `isTrusted === false` on exactly the elements the change exists to fix. One pointer mechanism now covers both fixture kinds. |
| 9 | `docs/api/reference.md` + `docs/api/inventory.md` stale | **Done.** Both regenerated/updated; `gen-api-reference.mjs --check` passes. |
| 10 | hover/drag moved to the `browser.click` capability — intended? | **Intended, and now stated in the changelog fragment.** Hover and drag *are* pointer input, which is what `browser.click` grants; routing them through `browser.evaluate` would mean handing out arbitrary script execution in order to move a pointer. |
| 11 | Item 6 of the brief (`data-wmux-ref` residue) | **Deferred**, as the brief permits. It lives in `tools/inspection.ts` and `snapshot.ts`, both changed by #1172; touching them would conflict and block both PRs. Current lifecycle is unchanged by this PR: `dom-intelligence.ts` clears every `[data-wmux-ref]` at the *start* of each snapshot, so stale refs never accumulate, but the tags from the most recent snapshot stay in the DOM because the RPC interaction path resolves refs through them. |
| 12 | Rotated/transformed elements | **Covered without the extra probe.** Reading `getComputedStyle().transform` costs a round trip on every click; finding 6(c) already handles it — a point inside the axis-aligned box but outside a rotated element makes Playwright's positioned click fail, and the retry-without-approach lands it on the centre. |

## Live smoke, round 2

**(a) Typing** — 40 chars, `keydown` timestamps:

```
  intervals: 165.9, 190.1, 88.9, 124.7, 55, 54.7, 74.5, 141.8, 62.4, 163.5,
  507.5, 102.2, 85.7, 118.3, 99, 137, 122.4, 451.3, 73.4, 115.9, 122.4, 128.3,
  81.6, 113.1, 104.9, 110.1, 104.9, 157.6, 188.7, 122.4, 251, 74.3, 66.6,
  137.3, 57.2, 143.6, 84.4, 127.5, 777.1
  count=39 median=118.3 mean=151 max=777.1 min=54.7 wall=6.16s
```

**(b) Click**:

```
  steps=11 mousemove events page saw=12 (bare click = 1)
  isTrusted=true offset=(-52, -15) inside=true innerBound=(+/-60, +/-18)
```

**(c) Referer**, same-origin vs cross-origin vs downgrade:

```
  about:blank -> /a : refererFor = undefined
  same-origin /a -> /b : sent "http://127.0.0.1:60679/a"
    on B: referrer=http://127.0.0.1:60679/a
  cross-origin 127.0.0.1 -> localhost : sent "http://127.0.0.1:60679/"
    on B: referrer=http://127.0.0.1:60679/
  https -> http downgrade : undefined
```

**(d) Emulate** — `navigator.userAgentData` before, after apply (post-reload, so
the override provably survived the tool returning), in a tab opened *during* the
emulation, and after reset:

```
  before: platform=macOS mobile=false brands=["Chromium","Not?A_Brand","Google Chrome"]
  applyUserAgentEmulation -> true
  after apply (post-reload, session still held open):
    navigator.userAgent mobile token = true
    platform=Android platformVersion=11.0.0 mobile=true bitness=""
    brands=["Not_A Brand","Chromium","Google Chrome"]
    fullVersionList=["24.0.0.0","145.0.7632.6","145.0.7632.6"]
  new tab opened during emulation: platform=Android mobile=true
  after reset: platform=macOS mobile=false brands=["Chromium","Not?A_Brand","Google Chrome"]
    UA back to desktop = true
```

**(e) Drag** — raw CDP session, no `Input.setInterceptDrags`, i.e. exactly what
`wc.debugger` sends:

```
RAW CDP drag, pointer-event fixture (#pes -> #pet):
  mousedown=true  moves=14  mouseup-on-target=true
RAW CDP drag, HTML5 draggable fixture (#h5s -> #h5t):
  dragstart=true  drop=true
DragEvent synthesis, HTML5 draggable fixture:
  dragstart=true  drop=true
```

Both fixtures work through real pointer input, which is why finding 8 was
rejected. (Driving the same fixtures through Playwright's `page.mouse` cannot
answer this question — Playwright enables drag interception and synthesises the
drag events itself.)

## Gates, round 2

- `npx tsc --noEmit` clean
- `npm run build:mcp && npm run probe:mcp` clean; still no tool description or
  schema change, `tools/list` `full` = **76,410 B** (cap 78,000), baseline unmoved
- `node scripts/gen-api-reference.mjs --check` passes
- `npx vitest run src/main/pipe src/mcp/playwright src/main/browser-session src/shared/__tests__ scripts/__tests__`
  → **180 files, 3039 passed, 5 skipped**. **46 new tests** across 5 new files.

## Summary by CodeRabbit

* **New Features**
  * Typing now follows a more natural rhythm, including pauses around punctuation and word breaks.
  * Clicking, hovering, and dragging use realistic pointer movement and input.
  * Navigation requests now include an appropriate referrer when applicable.
  * Device emulation synchronizes user-agent, platform, language, and related browser details.
  * Added browser hover and drag capabilities for supported automation workflows.
* **Documentation**
  * Updated the API reference and changelog with the latest browser capabilities and behavior changes.

### DIFF
--- a/changelog.d/1173.md
+++ b/changelog.d/1173.md
@@ -11,7 +11,12 @@
   travels from wherever the pointer was left to a point somewhere inside the
   element, which is what the page sees when a person clicks it. Hovering and
   dragging in the builtin browser now go through the same real pointer input,
-  rather than events synthesised inside the page. (#1173)
+  rather than events synthesised inside the page. Two new pipe RPC methods back
+  this — `browser.hover.cdp` and `browser.drag.cdp` — and they are gated on the
+  `browser.click` capability rather than `browser.evaluate`. A caller granted
+  hover and drag is being granted pointer input, which is what `browser.click`
+  means; routing them through `browser.evaluate` would have required handing out
+  arbitrary script execution to move a pointer. (#1173)
 
 - **Navigations carry the page they came from.** Following a link from one page
   to another now sends a `Referer`, as a real click-through does. A first
--- a/changelog.d/1173.md
+++ b/changelog.d/1173.md
@@ -1,0 +1,26 @@
+### Changed
+
+- **The agent types with a human rhythm instead of a metronome.** Inter-keystroke
+  delays used to be drawn flat from a 50–150 ms range, which no typist produces.
+  They now cluster around a median with a long tail, run longer after full stops,
+  commas and word breaks, and occasionally pause the way someone does mid-sentence.
+  (#1173)
+
+- **Clicks move the pointer there first, and land off-centre.** A click used to
+  teleport the pointer onto the exact centre of the target and press. It now
+  travels from wherever the pointer was left to a point somewhere inside the
+  element, which is what the page sees when a person clicks it. Hovering and
+  dragging in the builtin browser now go through the same real pointer input,
+  rather than events synthesised inside the page. (#1173)
+
+- **Navigations carry the page they came from.** Following a link from one page
+  to another now sends a `Referer`, as a real click-through does. A first
+  navigation, a reload, or a jump out of a blank or browser-internal page still
+  sends none. (#1173)
+
+- **Emulated devices are consistent about what they claim to be.** `browser_emulate`
+  used to change the User-Agent string alone, leaving the browser's Client Hints
+  answering with the real machine — so an emulated phone described itself as a
+  phone in one place and as your desktop in another. The user-agent data,
+  platform and Accept-Language now match the preset, and resetting the preset
+  restores all of them together. (#1173)
--- a/docs/api/inventory.md
+++ b/docs/api/inventory.md
@@ -131,7 +131,7 @@ Validation limits live in `src/shared/types.ts` (PANE_METADATA_MAX_BYTES, PANE_M
 | `browser.open` | `{ url? }` | experimental | The browser/CDP surface backs the MCP `browser_*` tools. Wire shapes may evolve before v3.0. Currently the primary AI-agent capability driver, but not part of the substrate identity. |
 | `browser.navigate`, `browser.goBack`, `browser.close` | various | experimental | |
 | `browser.session.{start,stop,status,list}` | various | experimental | |
-| `browser.type.humanlike`, `browser.type.cdp`, `browser.click.cdp`, `browser.press.cdp` | various | experimental | |
+| `browser.type.humanlike`, `browser.type.cdp`, `browser.click.cdp`, `browser.hover.cdp`, `browser.drag.cdp`, `browser.press.cdp` | various | experimental | |
 | `browser.cdp.target`, `browser.cdp.info` | various | experimental | |
 | `browser.screenshot`, `browser.evaluate` | various | experimental | |
 
@@ -350,7 +350,7 @@ The capability column below summarises the table. Three sentinels:
 | `meta.setStatus` / `meta.setProgress` / `meta.setSkills` | `meta.write` | — | metadata |
 | `system.identify` / `system.capabilities` | `null` (bootstrap) | — | — |
 | `browser.navigate` / `browser.open` / `browser.goBack` / `browser.close` | `browser.navigate` | — | browser |
-| `browser.click.cdp` | `browser.click` | — | browser |
+| `browser.click.cdp` / `browser.hover.cdp` / `browser.drag.cdp` | `browser.click` | — | browser |
 | `browser.type.humanlike` / `browser.type.cdp` / `browser.press.cdp` | `browser.type` | — | browser |
 | `browser.screenshot` | `browser.screenshot` | — | browser |
 | `browser.evaluate` | `browser.evaluate` | — | browser |
--- a/docs/api/reference.md
+++ b/docs/api/reference.md
@@ -26,7 +26,7 @@ returns `EPERM`. Wire framing: newline-delimited JSON, one object per line.
 
 ## RPC methods
 
-Total: **160** methods (`ALL_RPC_METHODS` in
+Total: **162** methods (`ALL_RPC_METHODS` in
 `src/shared/rpc.ts`). Capability and risk class are read from
 `src/main/mcp/methodCapabilityMap.ts`:
 
@@ -145,6 +145,8 @@ Total: **160** methods (`ALL_RPC_METHODS` in
 | `browser.responseBody.get` | `browser.read` | `browser` |
 | `browser.type.cdp` | `browser.type` | `browser` |
 | `browser.click.cdp` | `browser.click` | `browser` |
+| `browser.hover.cdp` | `browser.click` | `browser` |
+| `browser.drag.cdp` | `browser.click` | `browser` |
 | `browser.press.cdp` | `browser.type` | `browser` |
 | `browser.cookies` | `browser.cookies` | `browser` |
 | `browser.resize` | `browser.evaluate` | `browser` |
--- a/src/main/browser-session/HumanBehavior.ts
+++ b/src/main/browser-session/HumanBehavior.ts
@@ -1,3 +1,5 @@
+import { generateTypingDelays, typingDelayFor } from '../../shared/humanRhythm';
+
 export interface HumanBehaviorConfig {
   typingDelay: { min: number; max: number };
   actionInterval: { min: number; max: number };
@@ -22,9 +24,15 @@ export class HumanBehavior {
     this.config = { ...DEFAULT_CONFIG, ...config };
   }
 
-  getTypingDelay(): number {
+  /**
+   * The pause after one keystroke. `char` is the character just typed, which
+   * lengthens the pause after punctuation and word breaks; omit it for a
+   * context-free draw. The distribution lives in `shared/humanRhythm` so this
+   * lane and the MCP lane stay identical.
+   */
+  getTypingDelay(char?: string): number {
     const { min, max } = this.config.typingDelay;
-    return Math.random() * (max - min) + min;
+    return typingDelayFor(char, { minDelay: min, maxDelay: max });
   }
 
   getActionInterval(): number {
@@ -59,6 +67,7 @@ export class HumanBehavior {
   }
 
   generateTypingSchedule(text: string): number[] {
-    return Array.from({ length: text.length }, () => this.getTypingDelay());
+    const { min, max } = this.config.typingDelay;
+    return generateTypingDelays(text, { minDelay: min, maxDelay: max });
   }
 }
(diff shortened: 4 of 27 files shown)


openwong2kim:

This PR makes agent browsing less mechanical: log-normal typing rhythm, pointer travel before off-centre clicks, CDP-routed hover/drag, Referer on navigation, and consistent UA client hints for `browser_emulate`. It touches both browser lanes plus the capability maps, so timing assumptions are the main risk; 36 new tests and a live smoke cover it. P3 — behavioral polish, no user-facing breakage.

PR #1174, opened 2026-09-02
feat(remote): a remote surface says so — on its tab, in its tooltip, and in the text you paste

openwong2kim:
From the #1141 dogfood: a remote-terminal tab was indistinguishable from a local one. The tab's text is an OSC title set by the shell on the *other* machine, so a Windows host renders `C:\Program Files\…\pwsh` exactly as a local pane does — the live dogfood for this PR had a local and a remote tab whose visible text was byte-identical. The tooltip was worse than no help: it showed a cwd that is a real directory on a machine that is not this one.

- **The tab carries a glyph** — `IconExternalLink`, the same one the ⋮ menu's "New remote pane" and "Split … — remote" entries use, so the action and the tab it produces read as one thing. `role="img"` with a name, because a bare `aria-label` on a span may simply be dropped and this is the only visual signal the tab is remote.
- **The tooltip leads with it**: one interpolated key (`Remote terminal — {path}`), so a locale owns its own order and punctuation, and a host name can be added later without reopening translations.
- **So does the text you paste.** `renderSurfaceLines` had no remote branch, so dragging a remote tab out — or copying pane info — still produced "Terminal", an empty `PTY ID:` line, and a bare `CWD:` naming somebody else's directory. Pasted into an agent's prompt, that is a path it will confidently try to use.

### Colour

Deliberately `currentColor`, not an accent. `--accent-blue` is *this file's* focus signal — the active pane's tab strip is underlined with it, and the workspace-tag comment a few lines up had already rejected a blue-ish mark on a tab because it "would read as focus". Inheriting the tab's own colour also means the glyph dims and brightens with the active/inactive text. Measured live: `rgb(239,238,236)` active, `rgb(133,130,123)` inactive, neither `#6E9BC4`.

### Review team (2-way: independent context-free Opus + GLM-5.3; Codex 402)

Both caught the `aria-label`-on-a-span problem. The independent review additionally caught the colour conflict above with direct evidence from this file, and showed that the "the tab body can't be rendered in a test" premise behind my source-scan assertions was simply false — `SurfaceTabs.actions.test.tsx` already mounts the real component with `surfaces` as a prop. The tests render now: exactly one glyph, inside the remote tab, absent from the local one, not painted with the focus accent.

### Dogfood (real remote session, two rounds)

A second wmux instance's web terminal server as the host, paired over `remote.hostsAdd`, a session minted through the ⋮ menu. Round 1 passed the glyph, colour, tooltip and narrow-pane layout, and **caught that two lines of my markdown branch were dead in the live app**: `addRemoteSurface` has exactly one caller, which passes no shell and no cwd, so `— unknown` was not a fallback but the only branch, and the CWD line could never fire. My unit test had hand-written a surface carrying both — passing against a shape the product cannot produce.

Fixed by printing what a remote surface actually knows: no shell on the heading, the OSC title on its own `Title:` line, the CWD line kept guarded for the day a caller supplies one. The suite now builds its fixture with `createRemoteSurface`, called the way the app calls it, so it cannot drift again. Round 2 verified the live clipboard output line by line, and confirmed the local surface block is byte-identical to before.

## Summary by CodeRabbit

* **New Features**
  * Remote terminal tabs now display an external-link indicator and accessible remote-specific labels.
  * Tooltips identify remote terminals and show their remote path.
  * Exported or copied pane information now includes remote terminal, host, session, and remote working-directory details.
  * Added localized remote terminal labels and tooltips in English, Korean, Polish, and Chinese.

* **Bug Fixes**
  * Prevented remote terminal details from being mislabeled as local terminals or showing misleading local PTY and working-directory information.


openwong2kim:
This PR makes remote-terminal tabs identifiable: a glyph on the tab, a "Remote terminal — {path}" tooltip, and correct labels in copied pane info. Changes are renderer-only (SurfaceTabs, sessionInfoMarkdown, four locales) with mounted tests and a live dogfood, so risk is low. P3 — UX clarity on top of working remote panes, no functional breakage addressed.

PR #1175, opened 2026-09-02
fix(fleet): the workspace dot sees pending questions and dead stashed panes (#1168)

### LINKED ISSUE #1168: The workspace status dot misses pending questions and dead stashed panes — it can read green while a pane is blocked
Found while reviewing #997's roster work; independent of it and worth fixing on its own.

`selectWorkspaceAgentStatus` (the sidebar workspace dot, and via `selectAllWorkspaceAgentStatus` also MiniSidebar, the titlebar vitals and the deck Fleet) rolls up `selectFleetPanes`. That pass reads neither of two signals the roster selector does:

1. **`surfacePendingQuestion`.** `workspaceAgentRoster.ts` promotes a surface with a transcript-derived pending question straight to `awaiting_input` — "the strongest evidence that this agent needs input". `fleet.ts` has no reference to it at all (grep: 0 hits). So a stop payload that carries `agentStatus: 'complete'` alongside a pending question renders a **green** workspace dot while the roster row underneath is red and prints the question on its own line.
2. **Stashed-pane liveness.** The roster treats a stashed pane whose `stashedLiveness === 'exited'` as `error` / `needsAttention`; `fleet.ts` has no `stashedLiveness` handling, so a workspace whose only entry is a dead stashed pane shows a neutral grey dot.

Both are cases where the dot says "nothing to see" about a workspace that wants the user — the exact failure the dot exists to prevent, and the reason #997 could not simply lean on it as a complete needs-you signal.

Fix belongs in `selectFleetPanes` / `selectWorkspaceAgentStatus` so every surface that consumes the roll-up is corrected at once. A regression test that pins "roster and dot report the same urgency for a pending-question pane" would fail today.

openwong2kim:
Closes #1168.

## The bug

A Claude Code turn that ends on a question fires one Stop hook, and
`buildTurnBoundaryMetadata` puts **both** `agentStatus: 'complete'` and the
transcript-derived `pendingQuestion` into a **single** broadcast. The roster
selector promotes that question to `awaiting_input` and prints it; the
roll-up that paints the dot above it never read the field at all, so it took
the `complete` and painted the row **green**.

The same roll-up had no stashed-pane liveness handling either, so a workspace
whose only entry was a stashed pane the daemon had confirmed dead showed a
neutral grey dot while the roster called it `error` and offered recovery.

Both are the dot saying "nothing to see" about a workspace that wants the
user — the one failure the dot exists to prevent.

## The fix

At the source, `selectFleetPanes`, so the sidebar dot, MiniSidebar, the
titlebar vitals chip and the deck Fleet all correct together:

- A non-empty `surfacePendingQuestion` folds into the existing **per-surface**
  attention scan, so a question asked by a background tab lights the dot the
  same way that tab's retained status already did.
- A stashed leaf whose `stashedPaneLiveness` is `exited` resolves to `error`
  ahead of the other derivations, matching the roster's own precedence.

Two consequences the fix drags along:

- Three consumers hand the selector a hand-built narrowed state instead of the
  store, so each had to start passing `surfacePendingQuestion` — otherwise the
  fix would have reached only the two that pass full state.
- `buildFleetSnapshots` (the deck/orchestrator mirror) is a **fifth** consumer,
  and it needed the promotion in its own per-surface row loop, not just the
  stripped base pass. Its rows were gated on `surfaceAgentStatus` alone, so the
  mirror described a blocked pane as `complete` — and that payload feeds the
  deck heartbeat and the completion gate, which would otherwise finish over an
  agent that is waiting on an answer.

## Evidence

**Unit.** `dotRosterParity.test.ts` feeds **one state object** to both the
roster and the dot and asserts the direction that is actually a contract:
when a roster row needs the user, the roll-up above it must say so too. It
deliberately does not assert the two are equal — the roster is per-surface and
agent-gated, the fleet pass is per-leaf and gates on nothing, so equality would
be pinning a coincidence. All four substantive cases fail on `main`; two of the
three new mirror cases fail without the `buildFleetSnapshots` change.

**Dogfood.** An isolated instance (`WMUX_DATA_SUFFIX=-dog1168`) driven over
CDP, with a realistic turn pushed through the **real** hook bridge
(`SessionStart → UserPromptSubmit → PostToolUse → Stop`, the Stop carrying a
transcript whose last assistant message ends in a question). A/B on the same
instance, same sequence:

| moment | before | after |
|---|---|---|
| turn ends on a question | dot `--accent-green`, no glow | dot `--accent-red` + waiting glow |
| the roster row underneath | "Needs input" + the question | unchanged |
| turn ends with a statement | `--accent-green` | `--accent-green` |

So the dot now tracks the signal rather than contradicting the row below it,
and it still returns to green when the question is answered.

Worth recording for the next dogfood: a bare `Stop` hook is **dropped** — the
completion alarm's turn gate needs evidence the agent worked, so the sequence
above is the minimum that reproduces anything.

## Known consequences, deliberately not fixed here

Surfaced by review; each is a separate decision, not part of "stop the dot
lying":

- **The dot loses its dismiss gesture for question-blocked panes.** Focusing a
  pane clears `surfaceAgentStatus` (that is the app's dismiss gesture); nothing
  clears `surfacePendingQuestion` on focus. A user who reads the question and
  declines to answer now has a red dot with no way to dismiss it until the
  agent emits bytes. Making the question dismissible on focus would change the
  roster too, so it wants its own issue.
- **`countNeedsAttention` still counts only `awaiting_input`/`waiting`**, while
  the roster's `needsAttention` also counts `error`. A dead stashed pane
  therefore renders a red error card in FleetView under a header that says
  "0 need you". The divergence predates this change (it applies to every error
  pane); this makes it easier to hit.
- **The new attention source is less gated than the roster's** — no
  `surfaceAgent` identity gate, no per-surface brain-pty filter. It inherits
  that from the retained-status scan it sits in rather than introducing it.

## Gates

Review team (this session + two independent reviewers) and a dogfood, per the
working agreement. The review caught, among other things, an over-correction I
had already committed: I had replaced a true comment with a false one on an
unverified claim and pinned it with a test of a state the data model forbids
(a browser surface holding a ptyId). Commit 3 reverts that and carries the real
argument instead.

## Summary by CodeRabbit

- **Bug Fixes**
  - Pending questions now mark affected sessions as awaiting input across fleet views, sidebars, status indicators, and navigation.
  - Exited stashed sessions now appear as errors with red status indicators instead of neutral grey.
  - Status indicators and roster entries now stay consistent across active, background, and stashed sessions.
  - Whitespace-only or stale pending questions no longer trigger attention states.

### DIFF
--- a/changelog.d/1175.md
+++ b/changelog.d/1175.md
@@ -1,0 +1,10 @@
+### Fixed
+
+- **The workspace dot no longer reads green while an agent is blocked.** When a
+  Claude turn ends on a question, the same signal carries "turn complete" and
+  the question itself — and the sidebar dot only ever read the first half, so it
+  went green while the row directly beneath it said "Needs input" and printed
+  the question. It now turns red, and so do the mini sidebar, the titlebar
+  vitals chip and the deck roster, which all read from the same place. A
+  workspace whose only pane is a stashed session that has since died now shows
+  red too, instead of the grey dot that meant "nothing here". (#1175)
--- a/src/renderer/components/Deck/DeckFleet.tsx
+++ b/src/renderer/components/Deck/DeckFleet.tsx
@@ -73,11 +73,12 @@ export default function DeckFleet({
   const surfaceActivity = useStore((s) => s.surfaceActivity);
   const paneLabel = useStore((s) => s.paneLabel);
   const surfaceAgent = useStore((s) => s.surfaceAgent);
+  const surfacePendingQuestion = useStore((s) => s.surfacePendingQuestion);
   const paneRole = useStore((s) => s.paneRole);
   const roleBindings = useStore((s) => s.orchestratorRoleBindings);
 
   const panes = useMemo(() => {
-    const all = selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, surfaceAgent });
+    const all = selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, surfaceAgent, surfacePendingQuestion });
     // Roster = live terminal panes of the ACTIVE workspace only (M1.5: the
     // deck is this workspace's orchestrator, so its roster is this
     // workspace's agents — the fleet-wide view lives in the titlebar vitals).
@@ -89,7 +90,7 @@ export default function DeckFleet({
       ),
       'attention',
     );
-  }, [workspaces, activeWorkspaceId, surfaceAgentStatus, surfaceActivity, paneLabel, surfaceAgent]);
+  }, [workspaces, activeWorkspaceId, surfaceAgentStatus, surfaceActivity, paneLabel, surfaceAgent, surfacePendingQuestion]);
 
   if (panes.length === 0) return null;
 
--- a/src/renderer/components/FleetView/FleetView.tsx
+++ b/src/renderer/components/FleetView/FleetView.tsx
@@ -54,6 +54,7 @@ export default function FleetView() {
   // #850: per-PTY agent identity — gates workspace metadata inheritance so a
   // non-agent active pane never borrows the real agent's name/status.
   const surfaceAgent = useStore((s) => s.surfaceAgent);
+  const surfacePendingQuestion = useStore((s) => s.surfacePendingQuestion);
   // X8 supervision mirror — subscribed here so the selector re-runs when a
   // supervised pane arms/stops or its restart count changes.
   const supervisionByPtyId = useStore((s) => s.supervisionByPtyId);
@@ -89,8 +90,8 @@ export default function FleetView() {
   // trees or the per-pty attention map change (the two inputs the selector
   // reads), not on every unrelated store mutation.
   const panes = useMemo(
-    () => sortFleetPanes(selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, supervisionByPtyId, surfaceAgent }), fleetSortMode),
-    [workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, supervisionByPtyId, surfaceAgent, fleetSortMode],
+    () => sortFleetPanes(selectFleetPanes({ workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, supervisionByPtyId, surfaceAgent, surfacePendingQuestion }), fleetSortMode),
+    [workspaces, surfaceAgentStatus, surfaceActivity, paneLabel, supervisionByPtyId, surfaceAgent, surfacePendingQuestion, fleetSortMode],
   );
   const needsCount = useMemo(() => countNeedsAttention(panes), [panes]);
   // Stable identity key of the terminal ptyIds to poll for RAM. `panes`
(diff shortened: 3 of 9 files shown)


openwong2kim:
Follow-up filed for the dismiss-gesture gap noted under "Known consequences": #1176.

openwong2kim (commented):
Fixes #1168 in `selectFleetPanes` (`fleet.ts`): the dot, vitals chip, and deck mirror now see pending questions and dead stashed panes, pinned by a roster/dot parity test. Scope is renderer selectors plus five consumers — contained, no daemon or protocol changes. P2: a green dot over a blocked agent breaks the attention system's core promise.

openwong2kim:
Re: the CodeRabbit suggestion to keep exited stashed panes past the `ptyId !== ''` filter in `StatusBar.tsx` (lines 116 and 133) — checked, and not applying it.

Both call sites feed exactly two things: `running` (counts `agentStatus === 'running'`) and `needsYou` (`countNeedsAttention`, which counts `awaiting_input` and `waiting` only). An exited stashed pane is `error`, so it contributes to neither. Letting it through the filter would therefore not change either number, and the chip only renders when a count is nonzero — so it would not appear. The single reachable effect would be on `jumpToUrgent`, where `sortFleetPanes` ranks `error` above `complete`/`running`/`idle` and the dead pane would become the jump target for a chip that is not on screen.

The real issue underneath is the `countNeedsAttention` / roster `needsAttention` divergence over `error`, which is listed under "Known consequences" in the PR body. It predates this change (it applies to every error pane, not just stashed ones) and deciding it means deciding whether the vitals chip counts errors at all — a separate call from "stop the dot lying".

Note a stashed pane that is ALIVE and blocked on a question is unaffected: its representative surface keeps a real ptyId, so it passes the filter, counts toward `needsYou`, and #977's unstash-before-jump already reaches it.

openwong2kim:
Correction to the first bullet under "Known consequences" in the PR body — I had it wrong, and #1176 has been rewritten accordingly.

The bullet describes the persisting red dot as a regression in feel, and says making the question dismissible on focus wants its own issue. On checking the clearing paths, that framing is backwards.

A pending question is not an event marker like the `surfaceAgentStatus` attention states. Those record "something happened while you were not looking", so focus is a legitimate acknowledgement and clearing them there is right. A pending question records "this agent cannot proceed until a human answers" — looking at the pane does not unblock it. Clearing it on focus would put the dot back to reporting "nothing to see" over a workspace that is still blocked, which is #1168 again with a different trigger.

And there is no stuck state to fix. `setSurfaceActivity` deletes the question on the agent's next tool hook, `markSurfaceRunning` does it on plain byte movement for hook-less agents, and every stop rewrites the field (`buildTurnBoundaryMetadata` sends `''` when the turn does not end on a question). So both ways a turn can resume after an answer clear it. The dot stays red exactly as long as the agent is blocked.

What remains is narrower and is an idea, not a defect: there is no seen/unseen distinction on a blocked pane. If wanted it belongs in the glow, not the status — `agentStatusIcon.ts` already separates `dotVar` from `glowClass`. #1176 now says that.

Nothing in this PR changes as a result.

PR #1177, opened 2026-09-02
feat(mcp): browser_repl — drive many browser steps in one tool call

openwong2kim:
## Summary

Adds `browser_repl`: one MCP tool call runs a JavaScript snippet that drives the browser through many steps — `await browser.snapshot()`, `refs.find(...)`, `await browser.click(...)`, `await browser.wait(...)`, `await browser.extract_text()` — where each of those used to be its own round trip.

Every `browser.X(args)` goes through this server's own `browser_X` handler, so automation leases, password redaction, frame-aware refs, snapshot diffing, and action-trace recording for `browser_replay` all apply unchanged. The bridge adds only what direct handler calls would skip: the SDK's Zod validation and the MCP connection scope.

### Design

- **Worker thread, not a vm context.** The snippet runs in a per-connection `worker_threads` Worker; handlers stay on the main thread and calls cross via `postMessage` (structured clone — no host Promises/Errors/functions leak into the snippet). A timeout calls `worker.terminate()`, which stops even a synchronous `while (true)`; the next run starts a fresh worker and says so. Not a sandbox claim: `repl_run` already gives the same caller an unrestricted Node process.
- **Serialized per connection.** Concurrent `browser_repl` calls on one connection run in order; a killed run's in-flight handler calls land before the next run touches the page.
- **Handler collector.** `collectingServer()` wraps `server.tool`/`registerTool` so the eight browser tool modules register unchanged while the bridge gets `{name, shape, handler}`.
- **Permission boundary.** One approval of `browser_repl` covers the observe/act whitelist (navigate, navigate_back, tabs, click, type, fill, press_key, hover, drag, select, scroll, scroll_into_view, snapshot, smart_snapshot, extract_text, extract_data, wait, console, highlight, resize, dialog). Cookies, storage, files, downloads, PDF, trace, screenshot, evaluate, session management, and replay stay separate calls.
- **Profile `full` only**; `browser_` prefix is excluded from core by derivation. tools/list stays within budget; `scripts/mcp-protocol-baseline.json` updated.

### Robustness (from code review)

Stray `browser.*` calls after a run finished are refused instead of hanging; bridge rejections are answered as tool errors; the run chain swallows rejections so a failing run cannot become an unhandled rejection in the server; a worker that dies between runs is detected immediately; the ledger is capped (200 lines) and masks typed text at every depth (`browser_fill` field values included); lease blocks are only classified before the handler's own body; under the broker a scope-less call is refused rather than landing on a shared session.

## Dogfood

Live against a fresh tab through the daemon: wait → snapshot → refs.find → click → wait → extract_text took **1 `browser_repl` call vs 6 direct tool calls**. `browser_replay save` afterwards found the steps in the action trace, confirming the handler path was used.

## Tests

- `src/mcp/browser-repl/__tests__/browserRepl.test.ts` — bridge (whitelist, re-validation, `full:true` default, scope re-entry, ref parsing for three snapshot formats, event/hint separation, ledger masking), session (multi-step run, throw-on-failure and catch, state across runs, timeout → terminate → fresh worker, serialization, stray-call refusal, straggler ordering, bridge rejection, ledger cap, redeclare hint, idle worker death), output formatting.
- `src/mcp/playwright/__tests__/toolCollector.test.ts`, catalog lockstep test, routing test updated.
- `npx tsc --noEmit` clean, full vitest suite green (14052 tests), `npm run build:mcp && npm run probe:mcp` exit 0.

## Notes

- No version bump; changelog fragment at `changelog.d/1177.md`.

## Summary by CodeRabbit

- **New Features**
  - Added the `browser_repl` tool for multi-step browser automation using JavaScript in a single call.
  - Supports persistent sessions, console output, browser-call traces, parsed snapshot references, error handling, and configurable timeouts.
  - Preserves existing browser permissions and automation behavior while protecting sensitive arguments.
  - Available for supported browser actions; resource-intensive operations remain separate calls.

- **Bug Fixes**
  - Improved cleanup and recovery when automation sessions time out, crash, or lose their connection.

### DIFF
--- a/changelog.d/1177.md
+++ b/changelog.d/1177.md
@@ -1,0 +1,23 @@
+### Added
+
+- **`browser_repl`: drive the browser through many steps in one tool call.**
+  A snapshot, a `refs.find`, a click, a wait and a text extraction used to cost an
+  agent five round trips, each one re-reading a page of tool output. The new
+  full-profile tool runs a JavaScript snippet instead, where every allowed
+  `browser_X` tool is `await browser.X(args)` with its usual arguments. Snapshot
+  results come back with the refs already parsed, a failed step throws a
+  catchable error, `console.log` is captured, and the run's call ledger (tool,
+  redacted arguments, timing, browser events) is printed with the result so the
+  agent still sees what happened. State survives between calls; a snippet that
+  overruns its timeout has its runtime terminated and the next call starts fresh.
+
+  The snippet runs in a worker thread and every browser step goes through the
+  existing tool handler, so automation leases, password redaction, ref maps and
+  the action trace behind `browser_replay` all behave exactly as they do for
+  direct calls. One thing does change: approving a single `browser_repl` call
+  covers the 21 tools it exposes (navigate, navigate_back, tabs, click, type,
+  fill, press_key, hover, drag, select, scroll, scroll_into_view, snapshot,
+  smart_snapshot, extract_text, extract_data, wait, console, highlight, resize,
+  dialog). Tools with a heavier footprint — `evaluate`, `screenshot`, cookies,
+  storage, downloads, uploads, PDF, tracing, replay and surface lifecycle — stay
+  separate calls. (#1177)
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 78000,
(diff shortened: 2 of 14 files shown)


openwong2kim:

Adds a `browser_repl` MCP tool that runs multi-step JavaScript automation against the existing browser handlers in a worker thread. Scope is well-contained in new `src/mcp/browser-repl/` files, gated to the `full` profile, with whitelist, masking, and timeout-termination tests. P3 — additive convenience feature, nothing existing depends on it.

PR #1179, opened 2026-09-02
fix(browser-replay): stop a replay when the same-name element population changed

openwong2kim:
## The defect

Live on 3.50.0 (chrome backend). A flow was recorded on a page holding one
`button "Submit order"` — `sameNameIndex 0`, `sameNameTotal 1`. A decoy button
with the same accessible name was then inserted **before** the real one, and the
flow was replayed:

```
ok 4. browser_click — clicked button "Submit order"
warning: step 4: button "Submit order": the page now has 2 such element(s), the
recording had 1 — the first one cannot have been displaced, so replaying against it
```

The replay clicked the decoy. The real submit never fired, and the step was
reported `ok`. "The first one cannot have been displaced" is false: an insertion
before index 0 displaces it, and index 0 of the live population is then a
different element entirely. This is the silently-wrong-element outcome the
feature was explicitly designed never to produce.

## The fix

`matchRefAxis` refuses at **every** position when the live same-name population
size differs from the recorded `sameNameTotal`, instead of exempting index 0.
Position N identifies the recorded element only while the population around it
is the one that was counted at record time. The step stops the run the way a
vanished element does, naming both counts and which way the count moved, and the
page is left where it stopped so the agent takes one snapshot and finishes live.

Three things the review panel found on top of that:

**The count is re-checked against the live page at the moment of acting.** The
population `matchRefAxis` compares comes from the internal snapshot — a
measurement taken *before* the step — so an element added between that snapshot
and the click is invisible to it. `resolveRef` already re-counts live but
skipped singleton populations, which is exactly where a look-alike inserted
above the recorded element hides. It now takes a `strictCount` option that drops
that exemption; the replay lane is the only caller that passes it, so live tools
keep the looser behaviour that avoids false rejections.

**Unnamed axes are exempt from the comparison.** Their stored total is not a
same-name count: `smartRefAxisEntry` records `roleIndex`/`roleTotal` in those
fields for an element with no accessible name, measured over dom-intelligence's
own walk, while the number compared against is counted from the snapshot ref map
(depth-capped, filtered on overflow). The two enumerations need not agree on an
unchanged page, so strict equality there could stop a flow with nothing wrong
with it. `resolveRef`'s own count check exempts unnamed refs for the same reason.

**A count-change stop no longer counts against the trace.** It fed the per-step
failure streak, so a page that merely grew a second "Save" could quarantine a
working recording on its second try and drop it out of the hint pipe for good.
Such a stop now travels as `inconclusive` from the runner through the stats RPC
to `applyRunOutcome`, which moves no counter but the clock. A vanished element
still counts as a failure — that one *is* evidence about the flow.

Also: the count comparison runs before the index lookup, so a population that
shrank past the recorded position reports the count rather than only "nothing at
that position".

## Known limitation, stated in the code, the docs and the changelog

Count equality is not identity. A change that leaves the count the same — one
look-alike added above, one element removed below — still measures N, still
resolves by index, and still lands on a different element. Telling those apart
needs a disambiguator the recording does not carry; widening the recorded format
is out of scope here.

A named smartRef axis is compared across those two enumerations anyway, since
the axis carries no record of which recorder minted it. That takes the trade the
whole check is built on: a stop the agent can finish live costs less than a click
on the wrong element.

Unchanged: population equal resolves by index at every position; a missing
element and a recorded position that no longer exists still stop with their own
messages.

Removed along the way: the per-step warning plumbing (`StepOutcome.warning`,
`Resolved.warning`, the `...warn` spreads). `matchRefAxis` was its only producer,
so it became unreachable. The page-shape warning is untouched.

## Tests

`src/mcp/browser-replay/__tests__/replayRunner.test.ts`

- decoy inserted before the recorded first element → stops, nothing clicked,
  marked inconclusive (replaces the test that asserted the old warn-and-click)
- population unchanged at index 0 → still resolves, no warnings
- element renamed out of the population → existing stop, and NOT inconclusive
- population shrank past the recorded position → reports the count
- population grew after the internal snapshot → stops on the strict live count
- unnamed axis with a mismatched total → still resolves

`src/mcp/playwright/__tests__/snapshot.refstability.test.ts`

- `strictCount` blocks a singleton whose live count changed
- `strictCount` still leaves an unnamed ref alone
- the existing default-behaviour test is untouched and still passes

`src/shared/browserReplay/__tests__/actionTrace.test.ts`

- two inconclusive stops in a row move no counter and do not quarantine

## Docs

`docs/how-to/replay-browser-flows.md` described the old position-dependent rule.

## Gates

- `npx tsc --noEmit` — clean
- `npx vitest run src/mcp/browser-replay src/shared/browserReplay` plus the
  refstability, browser-session and browser.rpc suites — all pass; the full
  `src/mcp` suite is green at 1022 tests
- `npm run build:mcp && npm run probe:mcp` — clean; tools/list bytes unchanged
  at full 75,457 / core 46,665 / commander 43,314, since no tool description
  text was touched

### DIFF
--- a/changelog.d/1179.md
+++ b/changelog.d/1179.md
@@ -1,0 +1,13 @@
+### Fixed
+
+- **A replay no longer clicks a look-alike element and calls it a success.** A
+  step recorded against the *first* `button "Submit order"` on a page used to
+  replay against whatever sat at that position later, on the reasoning that the
+  first element cannot have been displaced. It can: a decoy button with the same
+  accessible name inserted above the real one takes over position one, and the
+  run clicked the decoy, never fired the real submit, and still reported the step
+  as `ok`. A same-name population whose size differs from the recording now stops
+  the run at that step — at every position, the first included — naming both
+  counts. The page is left where it stopped, so the recovery is unchanged: take a
+  snapshot, finish the flow live, and `save` under the same name to re-record the
+  healed path. (#1179)
--- a/changelog.d/1179.md
+++ b/changelog.d/1179.md
@@ -1,13 +1,20 @@
 ### Fixed
 
-- **A replay no longer clicks a look-alike element and calls it a success.** A
-  step recorded against the *first* `button "Submit order"` on a page used to
-  replay against whatever sat at that position later, on the reasoning that the
-  first element cannot have been displaced. It can: a decoy button with the same
-  accessible name inserted above the real one takes over position one, and the
-  run clicked the decoy, never fired the real submit, and still reported the step
-  as `ok`. A same-name population whose size differs from the recording now stops
-  the run at that step — at every position, the first included — naming both
-  counts. The page is left where it stopped, so the recovery is unchanged: take a
-  snapshot, finish the flow live, and `save` under the same name to re-record the
-  healed path. (#1179)
+- **A replay stops when the number of same-name elements changed, instead of
+  clicking by position anyway.** A step recorded against the *first*
+  `button "Submit order"` on a page used to replay against whatever sat at that
+  position later, on the reasoning that the first element cannot have been
+  displaced. It can: a look-alike with the same accessible name inserted above
+  the real one takes over position one, and the run clicked the look-alike,
+  never fired the real submit, and still reported the step as `ok`. A live
+  population whose size differs from the recorded one now stops the run at that
+  step — at every position, the first included — saying which way the count
+  moved. The count is also re-checked against the live page at the moment of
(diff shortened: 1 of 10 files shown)


openwong2kim:

Closes a real correctness hole in browser-replay: a decoy same-name element inserted before the recorded one was clicked and the step reported `ok`; a changed population now stops the run at every position. Touches `replayRunner.ts` (warning plumbing removed), docs, and tests. P2 — silently acting on the wrong element is the one outcome a replay must never produce, and pages that merely grew now stop too.

PR #1180, opened 2026-09-02
fix(browser_repl): ref type contract, and the hints a run was swallowing

openwong2kim:
Both defects were found driving `browser_repl` against a real page on the chrome backend.

## 1. `refs[i].ref` had the wrong type

The bridge parsed snapshot refs with `Number(...)` and handed the script `refs[i].ref: number` with `param: 'ref'`. But every browser tool schema types `ref` as `z.string()` (only `smartRef` is a number). So the tool's own documented usage — *pass `refs[i].ref` as the arg named `refs[i].param`* — failed the bridge's schema re-validation:

```
browser.type({ref: refs[0].ref, text: 'x'})
→ browser.type: invalid arguments — ref: expected string, received number
```

- `parseSnapshotRefs` now emits `ref` as a string when the target argument is `ref`, and keeps it numeric for `smartRef`.
- The call path coerces a numeric `ref` / `sourceRef` / `targetRef` to a string before validation, at any depth (so `browser.fill({fields:[{ref, value}]})` is covered too). A snippet that computed a ref itself, or that kept the old numeric value, keeps working instead of being rejected before the handler runs.

The tool description already said only "pass `refs[i].ref` as the arg named `refs[i].param`" — no type claim to correct, so it is untouched and the tools/list byte budget is unchanged.

## 2. Hints never reached the run's author

The bridge dropped the lease's `[replay] ...` / `[skill] ...` blocks as "addressed to the model, not to code". But the model *is* the author of the snippet: on a promoted page a direct `browser_navigate` shows

```
[skill] fixture-submit — proven 4-step flow — browser_replay {action:"run", …}
```

while the same call inside `browser_repl` showed nothing — the one place a proven flow would have saved the most round trips.

A run now collects the hint blocks of every call (successful or failed), dedupes them line by line in first-seen order, caps them at 20 lines, and renders them as a `--- hints ---` block alongside `--- calls ---` and `--- console ---`. They stay out of the per-call `{text, events}` value, so nothing changes for a script parsing results.

## Tests

Added to `src/mcp/browser-repl/__tests__/browserRepl.test.ts`:

- a numeric `ref` is accepted where the schema wants a string, at top level and nested in `fields`, while `smartRef` stays a number;
- a failed call's hints are reported without leaking into the error text;
- a two-call run collects both hint lines once each and renders them in a `--- hints ---` block.

Existing tests were updated for the new `shapeResult` return shape (`{value, hints}`) and the string refs; the "fold several steps into one run" test now passes `login.ref` straight through rather than wrapping it in `String(...)`, which is the contract this PR restores.

## Gates

`npx tsc --noEmit` · `npx vitest run src/mcp/browser-repl` (117 passed) · `npm run build:mcp && npm run probe:mcp` — all green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Browser REPL snapshot references are now correctly handled as strings where required, while numeric smart references remain unchanged.
  * Fixed errors caused by numeric reference values, including nested reference fields.
  * Hints are now collected only from successful browser tool calls and cannot be spoofed by page content.

* **New Features**
  * Browser REPL runs display deduplicated, numbered replay and skill hints in a dedicated section.
  * Hint output is capped by line and total size, with omitted-line counts reported when limits are reached.

### DIFF
--- a/changelog.d/1180.md
+++ b/changelog.d/1180.md
@@ -5,8 +5,9 @@
   types `ref` as a string — so the tool's own documented usage, passing
   `refs[i].ref` as the arg named `refs[i].param`, was rejected as "expected
   string, received number" before the click ever happened. Refs now come back
-  as strings where the argument is `ref` (and stay numbers for `smartRef`), and
-  a numeric ref computed by the snippet is coerced rather than refused. (#1180)
+  as strings where the argument is `ref`, and stay numbers for `smartRef`. Note
+  for existing snippets: a snapshot-lane `refs[i].ref` is now a string, so a
+  comparison written as `r.ref === 3` has to compare `'3'`. (#1180)
 
 - **`browser_repl` shows the hints its calls carried.** A page with a promoted
   flow advertises it to a direct `browser_navigate` call — `[skill]
--- a/changelog.d/1180.md
+++ b/changelog.d/1180.md
@@ -1,0 +1,16 @@
+### Fixed
+
+- **`browser_repl` snapshot refs are the type their tool wants.** A snapshot
+  inside a script handed back `refs[i].ref` as a number, but every browser tool
+  types `ref` as a string — so the tool's own documented usage, passing
+  `refs[i].ref` as the arg named `refs[i].param`, was rejected as "expected
+  string, received number" before the click ever happened. Refs now come back
+  as strings where the argument is `ref` (and stay numbers for `smartRef`), and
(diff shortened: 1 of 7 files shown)


openwong2kim:

Fixes two browser_repl bridge defects: snapshot refs now arrive as the string type the tool schemas want (numeric refs coerced at any depth), and runs surface `[replay]`/`[skill]` hints as a deduped block. Contained to src/mcp/browser-repl with tests updated. P3 — the old failure was loud, never silent, so low risk.

PR #1181, opened 2026-09-02
fix(browser): make the emulated identity agree with itself

openwong2kim:
Three places where the browser announced one thing and contradicted it a moment later, each measured live on the chrome backend against httpbin.org/headers and a local fixture.

## Referer vs Sec-Fetch-Site
`browser_navigate` sent the referer through `page.goto(url, { referer })`, an address-bar navigation. Chromium labelled it `Sec-Fetch-Site: none` + `Sec-Fetch-User: ?1` while the Referer header claimed a link had been followed. No browser produces that pair. When there is a referer to send, the navigation now happens from inside the page (`location.assign` in the isolated world, awaited the same way goto is), so Chromium fills the whole set consistently. `page.goto` keeps the first navigation, the ones leaving about:blank, and the fallback when the in-page route does not commit, all without a referer.

## Half-applied device presets
`browser_emulate device: "iPhone 13"` set the UA string and viewport and stopped there: the page still read `navigator.platform === "MacIntel"`, `maxTouchPoints === 0`, `devicePixelRatio === 1`. The preset now carries `platform` on the UA override, device metrics (scale factor, mobile flag, screen size) and touch points on both the Playwright and the packaged CDP lane; `device: null` takes all of it back. Client Hints stay omitted for a Safari UA.

## One-millisecond key holds
Humanlike typing pressed and released each key in the same breath, so every keystroke had a ~1 ms dwell. Keys are now held for a log-normal draw from the shared rhythm module (median 70 ms, clamped 30-150). Characters that cannot be a key (CJK, emoji halves) still go through `press()`.

## Tests
New: `navigation.referer.test.ts`, `link-navigation.test.ts`, `ua-emulation.test.ts`, `human-typing.test.ts`; extended `humanRhythm`, `uaMetadata`, `state.fallback`.

Gates: `tsc --noEmit`, vitest on `src/mcp/playwright` + `src/shared`, `build:mcp && probe:mcp`.

## Summary by CodeRabbit

- **New Features**
  - Human-like typing now varies key-hold durations and reliably handles unsupported characters.
  - Device emulation now applies screen dimensions, pixel density, mobile settings, touch support, and matching platform information.
  - Device emulation reset now clears metrics and touch settings completely.

- **Bug Fixes**
  - Referer-aware navigation now follows browser-like in-page navigation behavior, with reliable fallback handling when needed.
  - Navigation errors are now distinguished from uncommitted navigations to avoid unnecessary retries.

- **Documentation**
  - Device emulation limitations and reset behavior are now documented.

### DIFF
--- a/changelog.d/1181.md
+++ b/changelog.d/1181.md
@@ -1,0 +1,18 @@
+### Fixed
+
+- **A navigation that carries a Referer no longer contradicts it.** When the
+  agent left one page for another, the Referer header said a link had been
+  followed while Chromium's own `Sec-Fetch-Site: none` said the URL was typed
+  into the address bar — a pair no real browser sends. With a referer to send,
+  the navigation now happens from inside the page, so the whole header set
+  agrees; without one, `page.goto` is used as before. (#1181)
+
+- **`browser_emulate` device presets apply the whole device.** Choosing
+  `iPhone 13` set the User-Agent and viewport and stopped there, leaving
+  `navigator.platform` at `MacIntel`, no touch points and a desktop pixel
+  ratio behind the mobile UA. The preset now carries platform, device metrics
+  and touch on both backends, and `device: null` clears all of it. (#1181)
+
+- **Humanlike typing holds each key for a human interval.** Keys were pressed
+  and released within a millisecond; the dwell is now drawn from the same
+  log-normal rhythm as the inter-key delay (median 70 ms). (#1181)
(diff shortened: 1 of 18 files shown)


openwong2kim:

Fixes three fingerprint inconsistencies in the browser lane: referer-carrying navigations, half-applied device presets, and ~1 ms key holds.
Touches `browser_navigate`'s route (in-page `location.assign`, with a goto fallback), `ua-emulation`/`state.ts` device metrics, and `humanRhythm`; each area gets new tests.
P3 — correctness polish on a working lane, no user-facing breakage.

PR #1184, opened 2026-09-03
docs(cli): list --profile on mcp register in the help text

openwong2kim:
`wmux mcp register --profile core` shipped in 3.50.0, but `wmux --help` still listed `mcp register` bare — the only way to learn the flag existed was the changelog. The row now names the flag and says what the two profiles are.

Help text only; no behaviour change.

## Summary by CodeRabbit

- **Documentation**
  - Updated `wmux --help` to document the optional `mcp register --profile` flag.
  - Clarified that the `full` profile registers all tools by default, while the `core` profile excludes browser tools.

### DIFF
--- a/changelog.d/1184.md
+++ b/changelog.d/1184.md
@@ -1,0 +1,6 @@
+### Fixed
+
+- **`wmux --help` now shows `mcp register --profile`.** The `core` profile
+  shipped in 3.50.0 but the help text still listed `mcp register` bare, so the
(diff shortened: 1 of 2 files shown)


openwong2kim:

Documents  in , which shipped in 3.50.0 but was only findable in the changelog, plus the changelog entry. Touches help text in  only — no behaviour change, no risk. P3, safe docs fix; good to merge.

PR #1185, opened 2026-09-03
fix(browser-replay): verify a replayed element by its neighbourhood, not only its position

### LINKED ISSUE #1182: browser_replay: a same-count swap of same-name elements still resolves by position
## What happens

#1179 stops a replay when the number of same-name elements changed. It cannot see a swap that keeps the count: one look-alike inserted above the recorded element and one element removed below leaves N unchanged, so `sameNameIndex` still resolves, the click lands on a different element, and the step reports ok.

Measured on a local fixture during the 3.50.0 dogfood; the review panel on #1179 flagged it independently (two models).

## Why it is not in #1179

The recorded axis is `role / name / sameNameIndex / sameNameTotal / frameKey` and carries nothing that identifies the element beyond its position. Closing this needs a disambiguator in the recording — a text or attribute fingerprint, an ancestor path, or a stable id — which widens the recording format and the promoted-skill store, so it belongs in its own change.

A related residual from the same review: a *named* smartRef axis is still compared across two enumerations (dom-intelligence's full AX walk versus the replay snapshot's depth-capped refMap), because the axis has no provenance field.

## Acceptance

- A flow recorded against one `button "Submit order"` stops, or resolves to the right element, when a decoy is inserted above and another same-name element removed below.
- Existing recordings keep replaying (format change is additive).

openwong2kim:
Closes #1182.

#1179 stops a replay when the number of same-name elements changed. It cannot see a swap that keeps the count: one look-alike inserted above the recorded element and one removed below leaves the count unchanged, so the step resolved by index and clicked a different element while reporting ok.

## What ships
Each recorded ref now carries a `context` field: the accessible name of the nearest ancestor with a structural role (a landmark, `row`, `listitem`, `dialog`, a table cell). It is minted inside the accessibility walk the snapshot already does — no extra CDP call, no DOM query, and it is not printed in the snapshot the agent reads. Capped at 96 characters.

At replay it is a verifier, never a locator. Nothing is ever located by it; it can only contradict what the index found and stop the run:
- exactly one live element carries the recorded context, at the recorded index → confirmed (this also clears a merely-changed population size, so some #1179 refusals become successful runs again);
- one carries it at a different index, or none carries it while others carry something → STOP, hand the page back live;
- nothing carries any context, or several carry the same → no verdict, the pre-existing count rules decide.

The element is deliberately not followed to its new index — that would be resolving by position, the brittleness this design refuses. The accessible description was considered as a second signal and rejected: `aria-describedby` routinely carries live text, so a verifier built on it would stop every replay.

## Residual, left open by design
Two genuinely identical siblings — same role, same name, same container, differing only in position — still cannot be told apart; the check abstains and the count rules decide. Separating those needs a positional/DOM-path axis. Documented in `actionTrace.ts`, in the replay runner, and in `docs/how-to/replay-browser-flows.md`, and pinned by a test.

## Accepted cost
A section or row renamed between recording and replay now stops a flow that would have worked. That is the module's designed failure mode, and the right side to be wrong on for a step that may be a Delete.

## Tests / gates
`tsc` clean; `vitest` on browser-replay + shared + playwright 818 passed (+15); `build:mcp && probe:mcp` clean, tools/list bytes unchanged.

Note: `stepsFingerprint` hashes the axis, so re-saving a pre-existing flow under the same name resets its success/fail history once — a cache reset, no correctness impact.

## Summary by CodeRabbit

* **New Features**
  * Browser replays now verify an element’s surrounding named container before performing actions, helping detect look-alike elements that have been structurally replaced.
  * Replays can continue safely when the number of matching elements changes, provided the recorded element is positively identified.

* **Bug Fixes**
  * Replays now stop and return control to the live page when the intended element is missing or has been replaced.

* **Documentation**
  * Updated replay guidance to explain container verification, its precedence, and its limitations.

### DIFF
--- a/changelog.d/1182.md
+++ b/changelog.d/1182.md
@@ -1,0 +1,15 @@
+### Fixed
+
+- `browser_replay` now records the accessible name of the nearest named
+  container each element sat in, and checks it before acting. A flow whose
+  page was restructured underneath it — a look-alike inserted above the
+  recorded element and another removed below, which leaves the same-name count
+  unchanged and used to resolve silently by position — now stops at that step
+  and hands the page back live instead of clicking a stranger. Recordings made
+  before this release keep replaying under the previous rules.
+
+### Changed
+
+- A replay whose same-name population merely changed SIZE no longer stops when
+  the recorded element can be positively identified by its container, which
+  turns some of the previous refusals back into successful runs.
--- a/docs/how-to/replay-browser-flows.md
+++ b/docs/how-to/replay-browser-flows.md
@@ -53,6 +53,13 @@ name, and position among the same-named elements in the same frame — not by a
 DOM path. A page that was restructured, or whose refs were renumbered by a
 restart, still replays as long as the button still reads as the same button.
 
+A recording also notes the *neighbourhood* each element sat in: the accessible
+name of the nearest container that has one — `region "Express checkout"`,
+`row "Alice Chen"`. It is never used to find the element, only to check the
+one the position found. That is what catches a swap the count cannot see (a
+look-alike inserted above the recorded element and another removed below
+leaves the count unchanged, so position alone still "resolves").
+
 ## When a replay stops
 
 A step whose element is gone stops the run at that step and reports which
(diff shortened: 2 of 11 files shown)


openwong2kim:

Adds a recorded container-name (`context`) to each browser_replay ref and verifies it before acting, stopping the same-count element swap that position alone cannot see. The recording format change is additive with old traces unaffected, and the runner plus snapshot walk get 15 new tests. P3: targeted correctness hardening, matching #1182's severity.

PR #1186, opened 2026-09-03
fix(browser): let a device preset keep navigator.platform

### LINKED ISSUE #1183: browser_emulate: device presets cannot own navigator.platform or touch input on a Playwright-driven page
## What happens

After #1181 a device preset applies the UA, Client Hints (Chromium UAs), device metrics, screen orientation and `maxTouchPoints`, and re-applies them after every main-frame commit. Two things it still cannot do, both stated in the tool's result today:

1. **`navigator.platform` keeps the host's value.** Every `Emulation.setUserAgentOverride` we send carries the preset's `platform` (verified by logging the payload), yet the page answers `MacIntel` under an iPhone UA. The same command sequence on a page with no other debugger session attached does change it, so the value is being decided by another session on the target (the library's own), not by our payload.
2. **Touch is only half there.** `maxTouchPoints` reads 5, but the Playwright context was created with `hasTouch: false`, so `page.tap()` is refused and every click is dispatched with `pointerType: "mouse"` — a touchscreen that never produces a touch event.

## Why it is not in #1181

Both are the same root cause: the preset is layered over a page whose emulation is owned by the automation library. Holding them means creating the page in a context built for the device (`browser.newContext({ ...devices[name] })`), which changes how chrome-backend surfaces are created and how the lease resolves pages. That is a larger change than the fix round.

## Acceptance

- Under `browser_emulate {device:"Pixel 7"}`, `navigator.platform`, `userAgentData.mobile`, `maxTouchPoints`, `devicePixelRatio` and `Sec-CH-UA-*` all agree with the UA after a navigation, and `browser_click` dispatches touch events.
- `device: null` returns the page to the host identity with no residue.

openwong2kim:
Closes #1183. **Stacked on #1181** (`fix/browser-fingerprint-consistency`), because it edits the same `ua-emulation.ts`; base this on #1181 or land it after.

## The diagnosis in the issue was wrong, corrected by measurement
The issue blamed Playwright's own `_updateUserAgent` outwriting our `platform`. It does not — that call is gated on context `userAgent`/`locale`, both empty over `connectOverCDP`, so it never fires.

Real cause: of everything the UA override carries, `platform` is the one value Chromium keeps as a **page-wide setting** rather than as state owned by the session that wrote it. **Any** DevTools session detaching from the page resets it — even one that never set a platform. `PlaywrightEngine` opens and detaches throwaway sessions on nearly every tool call (target-info probes in the tab-resolution path), so wmux was wiping its own preset.

```
after emulate                     platform=Linux armv8l
after one unrelated attach+detach  platform=MacIntel      ← wiped
after re-send on the live session  platform=Linux armv8l  ← restored
```

## Fix
`reassertUserAgentEmulation(page)` at the tail of `PlaywrightEngine.getPageForScope()` — the one door every browser tool comes through, after the throwaway sessions are gone. Re-sends only the UA override (metrics and touch survive a detach). No-op when nothing is emulated.

Live, `browser_emulate {device:"Pixel 7"}`: `navigator.platform` = `Linux armv8l` after emulate, after navigation, and after intervening tool calls; `maxTouchPoints 5`, `devicePixelRatio 2.625`, `userAgentData.mobile true`, `Sec-CH-UA-Platform "Android"`. `device:null` restores the host with no residue.

## newContext: rejected
Over `connectOverCDP` a new BrowserContext is incognito-like — no cookies, history or extensions — a louder bot signal than the one it removes, and it abandons the real-tab model and tab/lease identity.

## Touch input: documented residual, not shipped
The one in-place route (`Emulation.setEmitTouchEventsForMouse`) does deliver real touch events, but Playwright's mouse dispatch that produces them never returns (>30s per click) — it would deadlock `browser_click`. Rejected on evidence. The tool result now states it: `maxTouchPoints` and `(pointer:coarse)` match the device, but clicks still dispatch as mouse. The stale "navigator.platform still reports the host platform" note is removed (this PR makes it false).

## Gates
`tsc` clean; `vitest` 3092 passed (+3 new); `build:mcp && probe:mcp` clean, tools/list bytes unchanged.

### DIFF
--- a/changelog.d/1181.md
+++ b/changelog.d/1181.md
@@ -12,8 +12,7 @@
   touch points, a desktop pixel ratio and desktop viewport semantics behind the
   mobile UA. The preset now carries device metrics, screen size, orientation
   and touch on both backends, holds them across navigations, and `device: null`
-  clears all of it. Two limits the result now states outright: on a page under
-  automation `navigator.platform` keeps the host's value, and a non-Chromium
+  clears all of it. One limit the result now states outright: a non-Chromium
   preset such as an iPhone cannot fill `navigator.userAgentData` or the
   `Sec-CH-UA*` headers, which keep answering as Chromium — a Chrome-based
   preset gives a consistent mobile identity. (#1181)
--- a/changelog.d/1183.md
+++ b/changelog.d/1183.md
@@ -1,0 +1,24 @@
+### Fixed
+
+- **A device preset now holds `navigator.platform`.** Under `browser_emulate
+  {device: "Pixel 7"}` the page answered `MacIntel` from the next tool call
+  onward, while its User-Agent, Client Hints, pixel ratio and touch points all
+  said Android. Of everything a preset installs, the platform is the one value
+  Chromium keeps as a page-wide setting rather than as state owned by the
+  session that wrote it — and every debugger session that detaches from the
+  page clears that setting, including a session that never set a platform of
+  its own. wmux opens and closes short-lived sessions on the page while
+  resolving which tab a tool is talking to, so the preset was being undone by
+  wmux's own bookkeeping. The override is now re-sent once the page has been
+  resolved, so `navigator.platform` agrees with the rest of the identity across
+  navigations and tool calls, and `device: null` still returns the host's.
+  (#1183)
+
+### Changed
+
+- **`browser_emulate` says what a touch preset does not do.** A phone preset
+  reports a touchscreen — `maxTouchPoints`, `(pointer: coarse)` — but the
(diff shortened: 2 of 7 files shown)


openwong2kim:

Makes a `browser_emulate` device preset hold `navigator.platform` by re-sending the UA override in `PlaywrightEngine.getPageForScope` once throwaway CDP sessions have detached. Guarded, no-op when nothing is emulated, three new tests — low risk, though stacked on #1181 so it should land after that. P3: fidelity polish on an existing feature, no core flow affected.

PR #1187, opened 2026-09-03
fix(cli): wmux mcp check names the isolated instance it is reading for

### LINKED ISSUE #1151: Isolated instance (WMUX_DATA_SUFFIX) still writes the production ~/.claude.json MCP registration

openwong2kim:
Follow-up to #1157 (which closed #1151 in 3.50.0).

#1157 stopped an isolated instance (`WMUX_DATA_SUFFIX`) from writing the production agent configs, and `register`/`unregister` warn when run under a suffix. `wmux mcp check` was the one subcommand with no isolation signal: it listed the production `~/.claude.json`, `~/.codex/config.toml` and `~/.gemini/settings.json` exactly as on the daily instance, so the production entry read like this instance's own.

`check` now prints a heading naming the suffix and stating that the rows below belong to the production wmux and that an isolated instance never registers itself; `--json` gains an additive `isolated: { suffix, note }` field. The `register`/`unregister` warnings now read the suffix through the shared `dataSuffix()` helper instead of the env var ad hoc. No write behaviour changes: the warn-and-proceed for explicit CLI actions is the documented owner decision in `externalRegistrationSkipReason()`.

Tests: silent when unsuffixed, empty suffix treated as unsuffixed, `check` names the suffix, `register`/`unregister` name the production configs. Gates: `tsc` clean, vitest 645 passed, `build:mcp && probe:mcp` clean.

Possible follow-up, not in this PR: `wmux setup-hooks` still writes the Codex `notify` entry and the OpenCode plugin under a suffix via `installLifecycleIntegrations`, which #1157 deliberately left alone.

### DIFF
--- a/changelog.d/1151.md
+++ b/changelog.d/1151.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **`wmux mcp check` now says when it is speaking for an isolated instance.**
+  An instance launched with `WMUX_DATA_SUFFIX` never registers itself in the
+  agent configs, but `check` still listed `~/.claude.json`, `~/.codex/config.toml`
+  and `~/.gemini/settings.json` with no hint of that — so the production entry
+  read like this instance's own, which is exactly the confusion that made the
+  original hijack hard to spot. `check` now prints a heading naming the suffix
+  and saying the rows below belong to your production wmux (`--json` gains a
+  matching `isolated` field), and the `register` / `unregister` warnings read
+  the shared `dataSuffix()` helper instead of the env var directly, so every
+  subcommand agrees with the rest of the suffix-aware paths. (#1151)
--- a/changelog.d/1187.md
+++ b/changelog.d/1187.md
@@ -9,4 +9,4 @@
   and saying the rows below belong to your production wmux (`--json` gains a
   matching `isolated` field), and the `register` / `unregister` warnings read
   the shared `dataSuffix()` helper instead of the env var directly, so every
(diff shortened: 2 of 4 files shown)


openwong2kim:

This PR makes `wmux mcp check` name the isolated instance (`WMUX_DATA_SUFFIX`) it reads for, with an additive `isolated` JSON field, and routes the register/unregister warnings through one helper. It touches only CLI output in `src/cli/commands/mcp.ts` plus tests — no write behavior changes. P3: clarity polish on top of #1157 with no functional risk.

PR #1188, opened 2026-09-03
perf(browser): keep browser_repl snapshots diffed, with a complete refs list

openwong2kim:
Inside a `browser_repl` script, `browser.snapshot()` and `browser.smart_snapshot()` forced `full:true` whenever the caller left `full` unset. The reason was sound — the bridge parsed `refs[]` out of the returned text, and a diff only holds changed lines, so `refs.find(...)` would silently miss unchanged elements — but the cost was real: on the Node.js `fs` docs an act-then-verify loop paid ~78 KB and several seconds per snapshot, every iteration, while a direct call returned a diff of a few lines.

## What changes
The bridge no longer forces `full:true`; snapshots in a script get the same diff-or-full rendering a direct call gets, and `refs[]` comes from a separate, complete listing.

A small `AsyncLocalStorage` side channel (`snapshotListing.ts`): both snapshot handlers already hold the complete listing in a local at the moment they decide what to return, so each publishes it there — one line each, right after the baseline is set. The bridge wraps the handler call in the capture and parses refs from that listing. Nothing is published when no capture is active, so direct MCP calls are byte-identical to before. ALS rather than a module global because the broker hosts several callers in one process and two concurrent snapshots must not read each other's listing. Refs are parsed by the same `parseSnapshotRefs` from text equal to what `full:true` would have returned, so `ref` stays a string on the snapshot lane and a number for `smartRef` (#1180). The first line of `text` still says `[snapshot: full]` or `[snapshot: diff …]`.

The `browser_repl` description said "full listing by default", now false; edited to "diff text, all refs" (4 bytes smaller). `scripts/mcp-protocol-baseline.json` updated for the changed full-profile hash.

## Live (installed app, chrome backend, nodejs.org/api/fs.html)
| step | before text/refs | after text/refs |
|---|---|---|
| `snapshot()` #2 | 50,033 / 468 | **129** / 468 |
| `smart_snapshot()` #2 | 78,428 / 2,814 | **209** / 2,814 |
| `snapshot({full:true})` | 50,033 / 468 | 50,033 / 468 |

A ref taken from the diff call's `refs[]` for a line absent from the diff text was passed straight to `browser.highlight` and resolved — the property the old forcing existed to protect.

## Gates
`tsc` clean; vitest browser-repl + playwright 776 passed; `build:mcp && probe:mcp` clean. tools/list: full 75,453 / core 46,665 / commander 43,314.

## Summary by CodeRabbit

* **Bug Fixes**
  * `browser.snapshot()` and `browser.smart_snapshot()` now return concise diffs again during repeated script interactions, while still providing references for all page elements.
  * The first snapshot and requests using `full: true` continue to return the complete page tree.
  * Snapshot behavior now works consistently across direct calls and scripted browser interactions.

* **Tests**
  * Added coverage for diff responses, complete reference listings, forced full snapshots, and fallback behavior.

### DIFF
--- a/changelog.d/1188.md
+++ b/changelog.d/1188.md
@@ -12,4 +12,4 @@
   still says `[snapshot: full]` or `[snapshot: diff …]` — while `refs[]` keeps
   every element on the page. `full: true` still forces the whole tree, and the
   entries are unchanged: `ref` is a string for `snapshot`, a number for
-  `smart_snapshot`.
+  `smart_snapshot`. (#1188)
--- a/changelog.d/repl-diff.md
+++ b/changelog.d/repl-diff.md
@@ -1,0 +1,15 @@
+### Changed
+
+- **`browser_repl` snapshots return a diff again, with the refs still
+  complete.** A `browser.snapshot()` or `browser.smart_snapshot()` inside a
+  script used to be forced to render the whole tree, because the script's
+  `refs[]` was parsed out of the returned text and a diff names only what
+  moved — so an act-then-verify loop paid for a full listing on every
+  iteration (a Node.js API page cost about 78 KB and 7–8 seconds each time,
+  where the direct tool answered with a few lines). The handler now hands the
+  bridge the complete listing alongside whatever it returned, so the script's
+  `text` is the ordinary diff-or-full the direct tools give — its first line
+  still says `[snapshot: full]` or `[snapshot: diff …]` — while `refs[]` keeps
+  every element on the page. `full: true` still forces the whole tree, and the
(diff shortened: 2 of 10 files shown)


openwong2kim:

Stops forcing `full:true` on `browser_repl` in-script snapshots — `text` returns the ordinary diff while `refs[]` is parsed from the complete listing published via a new AsyncLocalStorage side channel (`snapshotListing.ts`). Touches the bridge and both snapshot handlers; direct MCP calls stay byte-identical, with a fallback to the returned text. P3: pure perf (78 KB / 7–8 s → a few lines), well-tested, no correctness stakes.

PR #1191, opened 2026-09-03
fix(browser-replay): verify by the element's own attribute, and count with the recorder's enumeration

openwong2kim:
Two follow-ups to #1185, as two commits.

## 1. The element's own attribute as a second verify-only signal
#1185 left identical siblings — same role, same name, same container — indistinguishable. Most real pages do distinguish them on the element itself: `data-testid`, `id`, `name`, `aria-label`, in that order of preference. Each recorded ref now carries one of those as `own` (`data-testid=submit-secondary`), capped at 96 characters, sanitized, additive (old traces replay unchanged).

Both lanes walk the CDP accessibility tree, not the DOM, so neither had attributes. `ownAttributes.ts` does one `DOM.getDocument` on the session that just fetched the AX tree and builds a `backendDOMNodeId → label` map — the same bridge `redact.ts` already uses for password fields — so both lanes mint the identical string by construction. Main frame only: `backendDOMNodeId` is per target, so an out-of-process iframe would collide; frame refs get no `own`, which the verifier reads as "no verdict", never as a mismatch.

`ownVerdict` runs only where the context verdict abstained (several share the context, or the page mints none), and decides the same three ways: unique match at the recorded index runs; elsewhere or missing while others carry one → STOP; nothing carries any → abstain. Never a locator.

## 2. Provenance, so a count compares like with like
A named smartRef axis recorded `sameNameTotal` from dom-intelligence's walk, while replay counted from the depth-capped snapshot ref map, so an unchanged page could STOP. The axis now carries `via: 'smart'` when the smart lane minted it (absent means a11y, so old traces are unchanged). On a count disagreement only, a smart axis is re-counted with the recorder's own walk on a scratch identity, so a mid-replay measurement cannot renumber refs an agent holds. Reconciling clears the count, never the position.

## Live (installed app, chrome backend, two `button "Submit order"` in one section differing only by data-testid)
Emitted axis at the transport boundary, both lanes: `context: region "Checkout"`, `own: data-testid=submit-secondary`; `via: smart` only on the smart lane. A smart-recorded flow replayed through the new path: `2/2 steps ran … clicked button "Submit order" (#2 of 2)`. (The installed 3.50.0 app's `sanitizeAxis` strips the new fields from disk, as with #1185; this PR updates the sanitizer.)

## Residual
Identical siblings carrying none of the four attributes remain irreducible; both verifiers abstain and a test pins it. Out-of-process iframes get no `own`.

## Gates
`tsc` clean; vitest browser-replay + shared + playwright 886 passed (+7 net, ~19 new cases); `build:mcp && probe:mcp` byte-identical (full 75,457 / core 46,665 / commander 43,314).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved browser replay reliability when multiple elements share the same role, name, and container.
  - Replay can now use identifying attributes such as `data-testid`, `id`, `name`, or `aria-label` to distinguish elements.
  - Smart snapshot replays now verify element counts against the same snapshot listing used during recording, reducing false change detections.
  - Elements without identifying attributes continue to use position and count-based verification.

- **Documentation**
  - Updated replay guidance with the new disambiguation and verification behavior.

### DIFF
--- a/changelog.d/1191.md
+++ b/changelog.d/1191.md
@@ -34,4 +34,4 @@
   a difference that is only between the two ways of counting no longer does.
   The extra reading happens only on a disagreement, so ordinary replays are
   unaffected, and traces recorded before this release keep the behaviour they
-  had.
+  had. (#1191)
--- a/changelog.d/replay2.md
+++ b/changelog.d/replay2.md
@@ -1,0 +1,21 @@
+### Fixed
+
+- **A replay can now tell two identical siblings apart by the attribute the
+  page gives them.** The neighbourhood check added in the previous release
+  abstains on the one shape it was built around: two elements with the same
+  role, the same accessible name, and the same container are under the same
+  neighbourhood, so it has nothing to say about them and the count rules decide
+  by position. Real pages almost always distinguish those siblings on the
+  element itself — a `data-testid`, an `id`, a `name`, or an `aria-label` — and
+  a recorded step now carries whichever of those the element had, as a second
+  verify-only signal. Where the neighbourhood gives no verdict, the sibling
+  carrying the recorded attribute wins if it is unique and still at the
+  recorded position; if it moved, or if the population carries attributes but
+  not that one, the run stops and hands the page back live instead of clicking
+  the twin. Nothing is ever *located* by the attribute, and a population that
+  carries none at all abstains exactly as before, so no existing flow changes
+  behaviour. Both recording lanes — `browser_snapshot` and
+  `browser_smart_snapshot` — mint the value from the same pass, so a flow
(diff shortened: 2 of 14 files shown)


openwong2kim:

Adds the element's own attribute (`data-testid`/`id`/`name`/`aria-label`) as a verify-only signal when identical siblings leave the context verifier silent, and re-counts smart-ref axes with the recorder's own enumeration. Touches the replay verifier, both snapshot lanes, and the trace format — additive, old traces replay unchanged. P3: accuracy refinement, not a regression fix; the never-locate design and branch tests contain the risk.

PR #1194, opened 2026-09-03
fix(browser-replay): let a step's navigation land before the next step, and record browser_wait

### LINKED ISSUE #1193: browser_replay: a click that navigates is not awaited, so the next step resolves against the old page — and browser_wait is not recorded
Found in live dogfood on the installed 3.50.1 (chrome backend), real GitHub pages.

## Repro

1. On `https://github.com/openwong2kim/wmux/issues`, `browser_click` the "Pull requests" tab (a soft navigation to `/pulls`).
2. `browser_click` the " 937 Closed" filter link on the resulting page.
3. `browser_replay {action:"save", name:"x"}` then `{action:"run", name:"x"}` from the issues page.

Result, 3/3 runs:

```
ok 1. browser_navigate — navigated to https://github.com/openwong2kim/wmux/issues
ok 2. browser_click — clicked link "Pull requests 3"
STOP 3. browser_click — no link " 937 Closed" on the page any more
```

A `browser_snapshot` taken immediately after the STOP shows the link present with exactly that name, and the same flow driven by hand through `browser_repl` succeeds every time. The recording is fine; the runner looks too early.

## Cause (code-verified on main)

`src/mcp/browser-replay/replayRunner.ts`: after each step the runner re-charges the ref map with `generateSnapshot(page, …)` right away. Nothing waits for a navigation the click started, so on any SPA/Turbo page the ref map is the pre-click document and the next `matchRefAxis` finds an empty population.

Second half: the natural workaround, recording a `browser_wait` between the two clicks, does not work because `src/mcp/playwright/tools/wait.ts` never calls `recordAction` — waits vanish from the trace (the saved flow reported 5 steps but replayed only the 2 clicks).

## Expected

Either (a) after a step whose page emitted a `navigated` event, wait for load/DOM-settled before re-snapshotting (the click tools already observe that event — the repl call log shows `1 event(s)` on the navigating click), or (b) record `browser_wait` as a replayable step, or both. (a) covers the common case without the author having to know.

## What did work in the same round

- Same-page 2-step flow (Sort menu open + Escape): 3/3 ok, promote, `[skill]` hint fired on exactly the recorded URL and not on other github.com pages, demote clean.
- `browser_repl` diffed snapshot with complete refs (#1188): second snapshot 129 chars vs 3727, refs 47/47.
- Pixel 7 emulation: UA, userAgentData.mobile/platform, navigator.platform, maxTouchPoints all agree; `browser_click` dispatches pointerdown(touch)/touchstart/touchend/click(touch) (#1190). iPhone 13 preset's Chromium-vs-Safari note is accurate.

openwong2kim:
Closes #1193.

## Problem

A replay re-snapshotted the instant a click resolved. On a Turbo/SPA page the click's fetch and render come after, so the next step was resolved against the document the click happened ON and stopped with "no link … on the page any more" while the element was there a moment later. Live on github.com this failed 3/3 runs; the same flow driven by hand succeeded every time.

The natural workaround, recording a `browser_wait` between the two clicks, did nothing: `wait.ts` never called `recordAction`, so waits vanished from the trace.

## Fix

- **Settle on the navigation signal, not on load state.** `waitForLoadState` resolves at once against the state the OLD document reached, so a same-document navigation was never waited for. The runner now watches the main frame's `framenavigated` (commit) and its navigation `request` (start, so a slow origin is not missed) around each step. When one fired, it waits for `domcontentloaded` and then until two consecutive ref-map shapes agree (250 ms apart, at most four reads, 3 s cap). The settle's last read is the re-charge, so there is no second dump. A step that did not navigate pays only the 300 ms grace.
- **`browser_wait` is recorded** (url as `urlGlob`, selector, text, network idle) on the Playwright lane, with no page shape stamped so it can never become the trace baseline. A `fn` wait is not recorded: the cache file is untrusted input and a stored script is never evaluated, and a hole would refuse the whole flow. The replayed timeout is clamped (0 means "for ever" to Playwright).
- Wire baseline refreshed for the `steps` description wording.

## Verification

- Unit: replayRunner (settle between steps, none after the last, request-signal path, wait replay by condition, timeout clamp), wait recording (`urlGlob`, no shape, fn skipped). 242 tests in the touched suites pass; `tsc -p tsconfig.mcp.json`, eslint, `build:mcp`, `probe:mcp` clean.
- Live: worktree MCP bundle attached to the installed 3.50.1 app, chrome backend, real github.com. The exact flow from #1193 (Issues → Pull requests → Closed) replayed 3/3 after failing 3/3 before; a 3-step flow with two soft navigations replayed 3/3.

## Known limit

The wait-recording half needs an app release: the installed main process's trace sanitizer drops steps whose tool it does not know, so a flow containing a `browser_wait` is refused by `save` until the app ships with this change.

## Review

Three-model panel (Claude, GLM; Codex unavailable, 402). Folded: cache-file timeout clamp, request-start navigation signal, no double dump after settle, read cap, fn-wait not recorded instead of a whole-flow hole, url stored as `urlGlob` outside the credential scrub, no shape stamped on a wait, RPC lane records nothing, recording tests.

## Summary by CodeRabbit

* **New Features**
  * Browser replays now wait for navigation to settle before continuing, improving reliability in Turbo and single-page applications.
  * Successful `browser_wait` actions are recorded and can be replayed using URL, selector, text, or network-idle conditions.
  * Recorded waits count toward flow and action limits, while JavaScript predicate waits remain unrecorded.
* **Bug Fixes**
  * Added bounded timeouts for replayed wait conditions to prevent excessively long waits.

### DIFF
--- a/changelog.d/1194.md
+++ b/changelog.d/1194.md
@@ -1,3 +1,3 @@
 ### Fixed
 - `browser_replay` now lets a step's navigation land before looking for the next element. A click that started a Turbo/SPA navigation used to have the next step measured against the old page and stopped with "no … on the page any more" while the element was there a moment later (#1193).
-- `browser_wait` is recorded as a replay step (url, selector, text, network idle). A wait on a JS predicate is listed as a `stored-script` hole rather than dropped, since a replay never evaluates a script from the cache file.
+- `browser_wait` is recorded as a replay step (url, selector, text, network idle) and counts toward the ring and per-flow step limits like any other action. A wait on a JS predicate is not recorded: a replay never evaluates a script from the cache file.
--- a/changelog.d/1194.md
+++ b/changelog.d/1194.md
@@ -1,0 +1,3 @@
+### Fixed
+- `browser_replay` now lets a step's navigation land before looking for the next element. A click that started a Turbo/SPA navigation used to have the next step measured against the old page and stopped with "no … on the page any more" while the element was there a moment later (#1193).
+- `browser_wait` is recorded as a replay step (url, selector, text, network idle). A wait on a JS predicate is listed as a `stored-script` hole rather than dropped, since a replay never evaluates a script from the cache file.
--- a/scripts/mcp-protocol-baseline.json
+++ b/scripts/mcp-protocol-baseline.json
@@ -3,7 +3,7 @@
   "profiles": {
     "full": {
       "maxListBytes": 78000,
-      "wireResultSha256": "0dd5f49dd55a3ada047100d74b93242d14ee1a823556e86263b4fa34c24135c0",
+      "wireResultSha256": "64e84f5161fc63162ef9115ffd0fe4931711d4b5d97658c876991ba1e981d3bf",
       "instructionSha256": "f18849bb1ea62bcf5a1e46a73d633c787b4e08e763cc04b01c4e557df8f833eb",
       "toolNames": [
         "browser_open",
--- a/src/mcp/browser-replay/__tests__/replayRunner.test.ts
+++ b/src/mcp/browser-replay/__tests__/replayRunner.test.ts
@@ -944,12 +944,12 @@ describe('settling after a step (#1193)', () => {
       { role: 'link', name: 'Closed', sameNameIndex: 0, sameNameTotal: 1, frameKey: '', ref: 2 },
     ];
     // Snapshot 1 (pre-flight), then the click navigates; the page is still
-    // the OLD document on the next two reads and becomes the destination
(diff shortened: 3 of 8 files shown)


openwong2kim:

This PR fixes #1193 by settling after navigation-triggering steps (framenavigated/request signal, shape-agreement poll) and recording `browser_wait` as a replayable step. It touches the replay runner, the wait tool, and the replayable-tool list, with the fn-wait skip and timeout clamp handled defensively. P2 — matches the issue it closes, and the unit tests plus live 3/3 verification back it.

PR #1197, opened 2026-09-03
feat(worktask): gate runner, task-level adopt/close/pr and read-only git tools for the orchestrator

openwong2kim:
## Why

The orchestrator brain could not judge or harvest a fan-out task without a human: closing, opening a PR and adopting were Electron-IPC-only, and the only way to run the project gate was to type commands into a pane and poll the screen. This is lane O2 of the orchestrator track (wave 1).

## What

- **Gate runner** (`TaskGateRunner`): runs a fixed, allow-listed command set in the task worktree — the trusted `wmux.json` verify script if declared, else `npm run lint` then `npm test` (each only if the script exists). Fixed argv, no shell. 15-minute timeout, process-group kill on cancel, one gate per task, `skipped: deps_missing` instead of a false fail when `node_modules` is missing or a symlink. The verdict is recorded on the task ledger through a `LedgerPort` seam with compare-and-swap on `rev`; the tail is bounded and treated as untrusted.
- **Pipe RPCs** `task.gate.run`, `task.gate.cancel`, `task.adopt`, `task.close`, `task.pr`, plus read-only `task.git.status`, `task.git.log`, `task.gh.prView`, on the same service instances the GUI drives, with the fan-out RPC's gates (local origin, server-resolved workspace, owner-scoped task lookup). New `TaskAdoptService` for task-level adopt (clean target, 3-way apply, left unstaged).
- **MCP tools** `task_gate_run`, `task_adopt`, `task_close`, `task_pr`, `git_status`, `git_log`, `gh_pr_view` as `register*` functions. They are NOT registered on any surface yet — registration, the `RpcMethod` union entries and the `TEARDOWN_DENY_METHODS` verdict for `task.close` land with lane F (see `docs/internal/orch-o2-tool-contract.md`).
- **UX A**: user-facing vocabulary unified to Task/Worker (en/ko/pl), Missions section keeps a one-line empty state and groups tasks under their parent, fan-out dialog confirms an all-empty launch, merges readiness toasts and widens the role select, Settings notes that `claude-pty` switches the panel to a terminal.

## Tests

14 gate runner, 7 adopt service, 80 worktask RPC, 29 MCP tool-surface, 5 mission grouping, 7 fan-out dialog. Full `npm test`: 14,345 passing, 0 failing. `npm run lint` errors are pre-existing on main (none introduced).

## Notes

- Changelog fragment: `changelog.d/orch-o2.md`.
[...]

PR #1199, opened 2026-09-03
feat(input): a real submit receipt, a verified brain sender, scoped approval presses and an honest channel nudge

openwong2kim:
## Why

Four links in the orchestrator loop lied or went nowhere: `input.send` reported `submitted:true` after writing an Enter it never verified; the brain (no PTY) was an `unverified_sender` on its own workspace so it could not address sibling panes; `approvalPress` had no scope, so any pane could be pressed; the channel nudge carried only a hint and never reported whether it landed, and the CLI and MCP unread views disagreed. This is lane O1 of the orchestrator track (wave 1).

## What

- **Submit receipt.** After the `\r`, `input.send` waits (400 ms, named constant) for a turn start (agentStatus → running/awaiting_input via the WorkspaceMirror) or the composer line clearing (positional, since Claude Code re-renders the prompt into the transcript). Raw byte activity is deliberately NOT a signal. No receipt ⇒ one Enter retry at 2× delay, then `accepted:false` with the last 10 screen lines. New result fields `accepted`, `agentStatusAfter`, `receiptSignal`, `enterRetried`, `screenTail`; `unobservable` when neither signal exists. `terminal_send` no longer implies ok = submitted; `terminal_send_key` points at `accepted`. Baseline SHAs refreshed (descriptions only).
- **Brain as verified sender.** Main stamps the router-validated commander workspace onto `a2a.task.send` (a wire-supplied claim is dropped first); `decideReplyDelivery` accepts it for both pane-keyed guards (`same_ws_no_anchor`, `unverified_sender`). `self_loop` unchanged; new `target_is_brain` refusal.
- **Approval press scope.** `decideApprovalPress`: press only when the target is a task workspace, autonomy is on, the request came from a hook, and verify-then-press confirms the prompt; any unestablished fact ⇒ refuse (`out-of-scope`, request not expired). Wired at `ApprovalRegistry.resolve` via a `pressScope` dependency — see Notes.
- **Channel nudge.** The wake injection carries the oldest owed message's first line (≤ 200 chars, control chars stripped) after the hint; each inject outcome is recorded and a failure marks the recipient `target_gone` (success is never promoted to `delivered`; only the ack path may). `wmux channel unread` and `channel_unread` now compute the same set from the same daemon call.

## Tests

~30 added: receipt (turn-start accepts, echo-only rejected + retry, late turn-start, unobservable, `submit:false` never accepted), brain addressing (sibling allowed, brain→brain refused, forged claim dropped), nudge body/outcome, CLI parity, approval scope (four refusals + press + registry enforcement). `npm test`: 14,048 + 193 runtime passing. `probe:mcp` green.

## Notes

- **Integration follow-up (required before release):** `ApprovalRegistry.pressScope` is not yet supplied by main, so on this branch alone every approval press is refused fail-closed. The integration PR wires `(workspaceId) => ({ isTaskWorkspace, autonomyMode })` from the worktask registry and the deck autonomy store, and exempts human-origin presses.
- Live 20/20 tally was not run: the installed daemon predates `accepted`, so a tally would have been fabricated; the fake-PTY tests carry the proof.
- `SUBMIT_ENTER_DELAY_MS` was not re-measured; the receipt path now catches a bad value instead.
- Behavior change: `wmux channel unread` no longer reads `$WMUX_MEMBER_ID` and now prints caught-up rows. Changelog fragment: `changelog.d/orch-o1.md`.

## Summary by CodeRabbit

- **New Features**
  - Terminal submissions now report whether they were accepted, with status and screen details when available.
  - Channel unread output now matches tool results, including up-to-date entries and message previews.
  - Wake notifications include a safe first-line preview for recognized agent interfaces.
  - Agents can receive messages from authorized commanders within the same workspace.

- **Bug Fixes**
  - Failed wake notifications are tracked as unavailable, while later acknowledgments can restore delivery.
  - Automated approval actions are restricted to eligible delegated tasks; refused approvals remain available for human review.

### DIFF
--- a/changelog.d/orch-o1.md
+++ b/changelog.d/orch-o1.md
@@ -7,3 +7,12 @@
   Enter is re-sent once before giving up. Previously `submitted: true` meant
   only "a carriage return was written", so an orchestrator reported progress on
   panes whose prompt was still sitting uncommitted in the composer.
+
+- An orchestrator brain can finally reach the agents in its own workspace. A
+  brain owns no pane, so every same-workspace A2A reply it sent was suppressed
+  as an "unverified sender" and merely stored — the brain was told the message
+  landed while the worker sat waiting. A caller carrying the daemon-validated
+  commander binding now satisfies that guard and the missing-anchor one, while
+  the self-loop protection for pane callers is unchanged. A reply addressed at
+  a brain is refused outright with `target_is_brain`: there is no pane behind
+  it to write into.
--- a/changelog.d/orch-o1.md
+++ b/changelog.d/orch-o1.md
@@ -9,28 +9,37 @@
   panes whose prompt was still sitting uncommitted in the composer.
 
 - An orchestrator brain can finally reach the agents in its own workspace. A
-  brain owns no pane, so every same-workspace A2A reply it sent was suppressed
-  as an "unverified sender" and merely stored — the brain was told the message
-  landed while the worker sat waiting. A caller carrying the daemon-validated
-  commander binding now satisfies that guard and the missing-anchor one, while
-  the self-loop protection for pane callers is unchanged. A reply addressed at
-  a brain is refused outright with `target_is_brain`: there is no pane behind
-  it to write into.
+  brain owns no pane, so every same-workspace A2A reply it sent to an ADDRESSED
+  pane was suppressed as an "unverified sender" and merely stored — the brain
+  was told the message landed while the worker sat waiting. A caller carrying
+  the daemon-validated commander binding, for the workspace that binding names,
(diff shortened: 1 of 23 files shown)


openwong2kim:

Lane O1: a real `input.send` submit receipt, a verified brain sender for A2A, scoped approval presses, and nudges that carry the owed message's first line. Touches the pipe handlers, ApprovalRegistry and ChannelService; the `pressScope` seam is unwired, so automated presses fail-closed until the integration PR. P2 — same wave as #1197/#1198; merge only with the integration.

PR #1198, opened 2026-09-03
feat(deck): task ledger, worker-event routing to the owning brain, and a ledger-backed Stop gate

openwong2kim:
## Why

A brain that fans out never learns its workers finished: a task workspace's `agent.stop` / `awaiting_input` is pushed only to that workspace, whose `wakePolicy` defaults to `none`, so the coalescer consumes it and the parent stays idle. "Delegate and forget" is not the model being lazy — the signal never arrives. This is lane F of the orchestrator track (wave 1).

## What

- **Worker event → owner brain.** The deck handler copies a task workspace's stop/awaiting_input event to the `WorkTask.owner` workspace tagged `{taskId, taskWorkspaceId}`; the coalescer lets tagged events through `wakePolicy: 'none'` and names the worker task in the wake prompt. An owner with no brain yet gets the event parked as `orphaned_event` and replayed on the human send that boots it.
- **Task ledger** (`src/daemon/ledger/TaskLedger.ts`): a status log keyed by WorkTask id — JSONL under the wmux data dir, single writer, torn-tail-tolerant replay, 5 MB rotation with snapshot, compare-and-swap on `rev`, `canActorSet` authorization, `completed` only by a brain with a passing gate (or `force` + reason), `closeTask` → `cancelled`, terminal-entry retention. Pipe RPC `ledger.list` / `ledger.update` (commander token = brain, senderPtyId = worker). Fan-out registers `working` and remembers the mission channel.
- **Commander-only registration lane** in the MCP entry (`COMMANDER_ONLY_TOOLS` SSOT next to `COMMANDER_TOOL_SURFACE`): `ledger_list` is commander-only; `ledger_update` (worker-scoped) is in full and core. Reserved names for lane O2's tools with a test that they are absent until wired. The probe's subset invariant is relaxed to `core ⊆ full`, `commander ⊆ core ∪ COMMANDER_ONLY_TOOLS`, with a leak assertion. Budgets after: full 76,714 / 78,000, core 47,895 / 49,000, commander 43,973 / 60,000.
- **Ledger-backed Stop gate**, behind `deck.ledgerGate` (default OFF, `deck-ledger-gate.json`): open tasks owned by the brain block Stop; stale fallback and the consecutive-block cap are kept; the pending-decision release is NOT flipped; after 3 consecutive ledger blocks the gate releases and logs `ledger_gate_released`; a ledger read failure falls back to today's inference. While on, `deck_ask_decision` prompts carry the open-task list.
[...]

PR #1200, opened 2026-09-04
test(mcp): scan only the index.ts wiring point for reserved commander tool names

openwong2kim:
## Why

#1197 (lane O2) landed the task/git tool implementations under `src/mcp` ahead of their registration, which is exactly the reserved state the invariant from #1198 (lane F) protects. The test scanned every `src/mcp` source for the reserved names, so the two PRs turned main red together (`reserved lane-O2 names are absent … task_gate_run is registered in src/mcp before being wired`).

## What

The invariant now scans only `src/mcp/index.ts`, the wiring point. A reserved name is "wired" when the entry file registers it, not when a module defines it. The integration PR that wires those tools flips the names from reserved to commander-only and updates this test again.

## Verification

`npx vitest run src/shared/__tests__/commanderSurface.test.ts` green; eslint clean. Test-only change, no changelog fragment.

## Summary by CodeRabbit

* **Tests**
  * Updated validation coverage to inspect the MCP entry file directly.
  * Simplified file-reading imports and removed broad recursive source scanning.
  * Clarified test descriptions and comments to reflect the narrower validation scope.

### DIFF
--- a/src/shared/__tests__/commanderSurface.test.ts
+++ b/src/shared/__tests__/commanderSurface.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { readFileSync, readdirSync, statSync } from 'node:fs';
+import { readFileSync } from 'node:fs';
 import path from 'node:path';
 import {
   COMMANDER_TOOL_SURFACE,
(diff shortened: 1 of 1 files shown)


openwong2kim:
This PR narrows the commanderSurface invariant so reserved lane-O2 tool names are checked against the src/mcp/index.ts wiring only, unbreaking main after #1197 landed unwired implementations. It touches a single test file with no runtime code, so the risk is minimal. P3 — test-only change at the correct seam (registration vs. definition).

PR #1209, opened 2026-09-04
Let the agent read a pasted image itself, and give WSL panes a path they can open

### LINKED ISSUE #1196: [Feature]: Native image paste for Claude
### Problem / motivation

I'm not sure if that's an intended behavior or workaround but right now when you make a screenshot and paste it inside the terminal (Claude in my scenario) it is pasted as path to the image and not the native claude `[Image #1]`. It's generally okay and not a big deal and doesn't cause any issues but when working from inside wsl, agent has to do a loop through the mtn to actually access that image instead of it just being available to it from the get fgo

### Proposed solution

- Add a toggle for how you want image or similar artifacts to be handled (pasted through wmux's path or natively)
- Add a bootleg for wsl specifically that appends mtn to the path (chaotic evil solution)
- Implement full native support for the screenshot/artifact paste operations (idk if it's too complex because you need to support each provider separately like claude, codex, gemini, agy, etc)

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

openwong2kim:
Closes #1196.

## The problem

Pasting a screenshot into a pane wrote a temp PNG and typed its path. That was the only route an agent ever saw, so Claude Code — which reads the OS clipboard on its own image-paste key and renders a real inline `[Image #N]` attachment — got a file path instead. Under WSL the path was not even reachable: the Windows temp directory needs a `/mnt` hop from the Linux side, which is what the reporter hit.

## What changed

Image-**only** clipboards now pick a route:

- **native** — send the agent its own image-paste key and let it read the clipboard itself. No temp file is written at all. Claude Code binds `ctrl+v` on macOS/Linux and `alt+v` on Windows, and binds **both** under WSL, so a win32 host can always send Alt+V and cover a native Windows pane and a WSL pane with one sequence.
- **path** — the historical behavior.
- **auto** (default) — native only for a pane whose detected agent is known to read the clipboard itself; path everywhere else. A key sent to a shell lands in readline as quoted-insert and the image is dropped, so the allowlist is deliberate rather than a guess.

A clipboard carrying **text is untouched**: it still pastes through the same chunked writer as before. Only the five image-fallback branches changed, and they now share one helper instead of four near-copies.

On the path route a WSL pane gets the `/mnt` view of the file. The rewrite is narrow on purpose — a Windows host, and a shell positively identified as a WSL entry point (`wsl.exe`, the Store's per-distro launchers, `System32\bash.exe`; Git Bash and `wslconfig` excluded) — because handing a PowerShell pane a `/mnt` path is worse than handing it nothing. Daemon sessions live on the far side of the socket, so both PTY modes record their shell in one registry for main to ask, including on reconnect.

POSIX targets also get POSIX quoting. A pasted path is a string the user never typed, and after the `/mnt` rewrite a Windows username containing `$` would otherwise expand at a bash prompt.

## Verification

- Live: with a PNG on the clipboard, injecting a raw `0x16` into a running Claude Code pane produced `❯ [Image #1]` — the premise the native route rests on, confirmed against the shipped binary's keymap (`chat:imagePaste`) and its clipboard readers (`osascript «class PNGf»` / `xclip` / `wl-paste` / `powershell.exe`).
- 26 new tests: route selection, per-platform key, WSL launcher detection and path mapping, POSIX vs Windows quoting, and the settings control.
- 873 tests across the touched areas pass; `tsc --noEmit` is clean apart from one error that predates this branch (`BrowserReplSession.ts:182`, present on `main`).

## Known limit

`surfaceAgent` keeps a pane's detected agent slug after the agent exits, so pasting an image into the plain shell left behind by a finished Claude session takes the native route and the image is dropped (the clipboard survives, so a retry works). The `path` setting is the escape hatch; agent-exit detection is filed separately.

## Summary by CodeRabbit

* **New Features**
  * Added configurable image-paste modes: Auto, Native, and File path.
  * Compatible agents can receive images directly; other terminals receive a temporary image path.
  * Added WSL-aware path conversion for pasted images.
  * Image-paste preferences are preserved across saved sessions.
* **Documentation**
  * Added changelog entries and translations for the new image-paste settings.
* **Tests**
  * Added coverage for routing, platform behavior, WSL paths, quoting, and settings controls.


openwong2kim:
Implements #1196: image-only pastes route native (agent reads its own clipboard key), path, or auto, plus a ptyId→shell registry so WSL panes get `/mnt` paths. Touches clipboard/PTY handlers and renderer paste paths; the auto-route risk (stale agent slug silently dropping an image) is already filed as #1210. P2 — worth timely review since it changes default paste behavior.

PR #1211, opened 2026-09-04
docs(internal): orchestrator wave 1 and wave 2 live dogfood record

openwong2kim:
## Summary

Records the two live dogfood passes of the orchestrator track on the isolated demo instance.

- Wave 1 (already on disk, never committed): fan-out routing, ledger and mission channel chain, the first-run stall, and the brain-boot backlog replay.
- Wave 2 (new): first-run handling and autonomy inheritance pass on the packaged build; the owner-brain loop is blocked because inheriting the owner's Deck mode spawns a brain per task workspace that consumes its worker's stop events (finding 7). Findings 8 to 10 cover the demo-instance traps (installed MCP bundle, shell model env, CDP port collision).

## Test plan

- [x] Documentation only; no code changes.

## Summary by CodeRabbit

- **Documentation**
  - Added internal documentation summarizing two dogfooding evaluation waves.
  - Recorded tested scenarios, validation outcomes, observed issues, follow-up findings, and merged changes.
  - Included an addendum documenting backlog replay behavior during system startup.

PR #1206, opened 2026-09-04
feat(orch): workers survive first-run screens, task workspaces inherit autonomy, and the brain presses approvals through the record

openwong2kim:
## Summary

Orchestrator wave 2, lane A: fan-out workers no longer freeze on Claude Code's first-run screens, task workspaces inherit their owner's autonomy, and the brain answers approvals through the daemon record instead of typing keystrokes.

- **First-run screens** (A-1): per-worker `CLAUDE_CONFIG_DIR` was tried first and rejected (the worker comes up "Not logged in" and detaches from the user's wmux hooks). Accepted: `CLAUDE_CODE_SANDBOXED=1`, which suppresses the trust dialog and onboarding without writing to the user's global config. Fallback detector answers an allow-list of known first-run headlines with ESC and reports `input_required` otherwise, so a real question is never blind-pressed. `WMUX_AGENT_FIRST_RUN=off` opts out. Verified live against two throwaway `claude` pty spawns (baseline shows the dialog, sandboxed lands authenticated on the composer).
- **Task workspace autonomy** (A-2 precondition): a fan-out task workspace inherits the owner workspace's autonomy mode. `assist` still means no unattended press; `modeToCaps` stays the single source of truth.
- **`approval.press` RPC + commander-only `approval_press` tool** (A-2): resolves the daemon approval record, presses the key the record names (never caller text), and declares itself `automated` so the press scope actually governs it. `terminal_send` and `terminal_send_key` are refused for a commander on a pane holding a record. Deadlock guard: a policy refusal (`press-capability-off`, `autonomy-off`, `not-a-task-workspace`, …) lifts the block for that pane for 10 minutes with a `typedFallback` note; transient refusals (`prompt-gone`, `not-found`) do not.
- **Coalescer verdicts** name `approval_press({ ptyId, decision })` with the task id, `status=awaiting_input` and the next-step contract ("END YOUR TURN — task X runs on").
- **`deadlineAt`** added additively to gate approvals (an AskUserQuestion has no timer and gets no fake clock). Lanes B and C render it.
- `docs/api/reference.md` regenerated for the new RPC method (drift guard).

Byte counts: commander 54,027 / 60,000 (52 tools), full 77,060 / 78,000 (unchanged; the tool is commander-only).

Spec: `plans/orch-wave2-plan-2026-09.md` (Lane A + Rev 2 amendments).

## Test plan

[...]

PR #1212, opened 2026-09-04
fix(orch): a fan-out task workspace never runs a brain of its own

openwong2kim:
## Summary

Wave 2 dogfood finding 7 (`docs/internal/dogfood-orchestrator-2026-09.md`, PR #1211): since #1206 a task workspace inherits its owner's Deck mode so the owner's brain may press its approvals. That mode also made the task workspace brain-eligible, so every fan-out worker got a Deck brain of its own within a second of spawn. Those brains consumed their worker's stop events (`[pane-events] … kind=stop`) before the owner's coalescer saw them, acted as operators on their own worker, and were held open by the Stop gate for as long as the worker ran (66 turns / 32 tool calls for one worker).

- `refuseWhenModeOff` also refuses an open fan-out task workspace with `task_workspace` — the single kill switch every turn path goes through (composer, wake, heartbeat, loop, scheduler, decision resume).
- The event-push path no longer pushes a task workspace's own lifecycle events to its own coalescer; they go only to the owner, tagged with the task (lane F), exactly as before the inheritance landed.
- The heartbeat's reviewed-workspace list skips task workspaces.
- A closed task hands the workspace back (the refusal reads the ledger's OPEN entry).

## Test plan

- [x] New tests in `deck.handler.loop.test.ts`: a worker's stop wakes the OWNER's brain tagged with the task and spawns nothing on the task workspace; wake and composer both answer `task_workspace`; a cancelled task lifts the refusal.
- [x] `tsc --noEmit` (root + daemon), eslint on touched files, `vitest run src/main/deck src/main/ipc/handlers/__tests__ src/main/worktask` 78 files / 1,232 tests.
- [x] Re-run of the wave 2 brain-side dogfood on the packaged build: no brain spawned for either task workspace, and the owner brain woke and drove its workers once a stale pending decision was answered (docs/internal/dogfood-orchestrator-2026-09.md, wave 3 section, PR #1211).

## Review round (Claude + GLM)

- An owner that is itself a task workspace (nested fan-out) parks its worker events instead of pushing them into a coalescer whose turn is always refused.
- A task reaching completed/failed/cancelled hands its workspace the default mode again (the inherited `danger` no longer outlives the task).
- The heartbeat filters task workspaces from its final set, not only the mirror branch.
- An unreadable ledger warns once instead of silently reviving the per-worker brain.
- Deferred: the first stop arriving before the ledger row exists (fan-out registers the row before the worker starts; only a failed registration is exposed), and a per-workspace index on the ledger lookup.

## Summary by CodeRabbit

* **New Features**
  * Fan-out task workspaces now operate without their own AI brain while tasks are active.
  * Worker stop events are routed to the owning workspace and associated with the relevant task.
  * Task workspaces resume normal operation after tasks finish, fail, or are cancelled.

* **Bug Fixes**
  * Prevented ambient and composer interactions from starting turns in active task workspaces.
  * Improved handling of nested task workspace events.

### DIFF
--- a/changelog.d/task-workspace-no-brain.md
+++ b/changelog.d/task-workspace-no-brain.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **A fan-out task workspace no longer spawns an orchestrator brain of its own.** Since the task workspace started inheriting its owner's Deck mode (so the owner's brain may press its approvals), that mode also made it brain-eligible: every worker got a brain that consumed the worker's own stop events before the owner ever saw them, one extra Claude session per worker, held open by the Stop gate. Task workspaces are now brain-less for as long as their task is open — the composer and every ambient driver answer `task_workspace` — and a worker's stop reaches only the owner, tagged with its task.
--- a/src/main/deck/CommanderSessionManager.ts
+++ b/src/main/deck/CommanderSessionManager.ts
@@ -35,12 +35,14 @@ export interface CommanderSendResult {
   /** Present on rejection: `busy` (a turn is running), `disposed`, `empty`,
    *  `invalid_workspace` (handler-level, M1.5 — the send carried no valid
    *  workspaceId to route to an orchestrator), or `mode_off` (handler-level —
-   *  the workspace's agent mode is `off`, so the brain must not run at all).
+   *  the workspace's agent mode is `off`, so the brain must not run at all),
+   *  or `task_workspace` (handler-level — a fan-out task workspace never runs
+   *  a brain of its own; the owner's brain drives it).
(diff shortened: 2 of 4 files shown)


openwong2kim:

This makes a fan-out task workspace refuse a brain of its own (wave 2 finding 7): turns answer `task_workspace` and worker stops route only to the owner. Changes are confined to `deck.handler.ts` (turn gating, event routing, heartbeat) plus tests and a changelog entry. P2 — it ends one held-open Claude session per worker and unblocks the orchestrator dogfood.

PR #1213, opened 2026-09-04
fix(deck): the stop gate and deck_complete_work only wait on agent panes

openwong2kim:
## Summary

Wave 3 finding 11 (`docs/internal/dogfood-orchestrator-2026-09.md`): `deck_complete_work` refused with `workers_outstanding` naming a plain zsh pane in the owner workspace that the operator had typed a command into; the brain then raised a decision and stopped. A shell is not a worker.

- `FleetSnapshotPane.isAgent?: boolean` (additive): the renderer's mirror snapshot now says whether a pane is an agent pane, from the same `surfaceAgent` detection (#850 guard) and pending-question signal (#1168) the store already keeps. `agentName` could not carry this: it is only set for the workspace's active pane, so every background worker reports `null` too.
- One exported rule, `isOutstandingWorkerPane` in `stopGate.ts`, used by both the Stop gate's snapshot inference and `deck.completeWork`, so the two sites cannot drift again. Reason strings say "agent pane(s)".
- Fail-open posture unchanged: a null/stale snapshot contributes nothing; a pane whose `isAgent` is unknown (older renderer) still blocks — only an explicit `false` releases.
- Trade-off: a just-spawned agent whose detector has not matched yet reads as a shell for those first seconds; the ledger gate and the A2A re-query still cover dispatched work.

## Test plan

- [x] 9 new tests: completeWork (running shell completes; agent pane beside it still refuses and the payload names only the agent; unknown `isAgent` blocks), stopGate (shell running/awaiting does not block; agent beside a shell blocks, reason names 1 agent pane, fingerprint omits the shell; unknown blocks), mirror snapshot (background agent true, pending-question true, byte-running shell false), handler parse pin.
- [x] `tsc --noEmit` root + daemon, eslint on touched files, vitest over pipe/ipc/renderer hooks+stores/deck: 170 files / 3,155 tests.
- [ ] Live: `deck_complete_work` from the demo brain with the harness shell pane busy (wave 3 dogfood re-run).

## Summary by CodeRabbit

* **Bug Fixes**
  * Plain shell panes no longer prevent turn completion or stopping when commands are running or awaiting input.
  * Completion and stop messages now clearly refer to outstanding agent panes.
  * Agent panes continue to block completion when they are still running or awaiting input.
  * Pane detection now reliably distinguishes agent activity from regular shell activity while preserving compatibility with older snapshots.


openwong2kim:
Stops the Stop gate and deck_complete_work from counting the operator's own shell as an outstanding worker, via one shared isOutstandingWorkerPane predicate and a per-pty isAgent flag on the mirror snapshot. Unknown isAgent still blocks (fail-open), so older renderers keep today's behavior. P2 — fixes a completion refusal no agent action could ever clear.

PR #1214, opened 2026-09-04
fix(orch): a pending decision no longer eats delegated wakes silently, refusals read REFUSED, and the ledger names its legal moves

openwong2kim:
## Summary

Wave 3 findings 12–14 (`docs/internal/dogfood-orchestrator-2026-09.md`).

- **F12 — pending decision as a silent kill switch.** The coalescer now logs one line per workspace per minute when a wake is dropped for a pending decision, naming how many events and which delegated task ids. Delegated (task-tagged) events are **held in the buffer** instead of consumed: `resolveDecision` kicks a resume turn without the human-send / brain-boot hooks, so a ledger park would never replay — the held events flush on the resume turn's idle instead. `fanout_start`'s accepted reply gains an additive `warnings` array when the owner workspace already has a pending decision.
- **F13 — refusals misread as success.** A refused `task_adopt` / `task_close` / `task_pr` now starts `REFUSED (<reason>): <error>` followed by `Nothing was adopted. Next step: …` (per reason), with the JSON after it; `isError` stays true. Gate tools are untouched (a failing gate is a successful call).
- **F14 — ILLEGAL_TRANSITION.** The ledger error appends `(allowed from <from>: <list>)`; the brain-side `ledger_update` description states the transition table in one line and names `force: { reason }` as the only brain-side exit; the task-tagged wake contract says which move is the brain's. The worker-side description is unchanged (the `full` profile has 398 B of headroom).

Byte counts: full 77,602 / 78,000 (unchanged), core 48,241, commander 55,698 / 60,000.

## Test plan

- [x] 7 new coalescer tests (log content and rate limit, snapshot path, hold-then-flush on resume, watermark not advanced, ambient consume unchanged), 3 fanout warning tests, 5 worktask refusal tests + 1 updated, 2 ledger message tests, 2 description tests, 2 contract assertions.
- [x] `tsc` root + daemon + mcp, eslint on 13 touched files, `build:mcp && probe:mcp` green, vitest over deck/mcp/ledger/pipe: 172 files / 2,941 tests.
- [ ] Live: refused adopt text and the pending-decision warning on the demo brain (wave 3 dogfood re-run).

Left: the fan-out warning is computed at accept time only; a decision raised afterwards is covered by the coalescer log.

## Summary by CodeRabbit

* **New Features**
  * Pending decisions now clearly indicate when worker events are being held, while delegated task events are preserved for later delivery.
  * Fan-out confirmations warn when the workspace has an unresolved decision.
  * Refused task adoption, closure, and pull-request actions now show a clear refusal reason and next step while preserving structured details.
  * Ledger errors now list valid next transitions, and ledger guidance documents the complete transition rules.

* **Bug Fixes**
  * Improved handling and diagnostics for blocked events and unresolved decisions.

### DIFF
--- a/changelog.d/brain-loop-honesty.md
+++ b/changelog.d/brain-loop-honesty.md
@@ -5,25 +5,34 @@
   deliberate — but until now the events it blocked were dropped without a
   trace, so a decision left over from a previous session ate every fan-out
   worker's "I'm done" and the orchestrator simply looked asleep. The block is
-  now logged once per workspace per minute, naming how many events it held and
-  which delegated tasks they belonged to, and a worker's own events are no
-  longer thrown away: they are held and delivered on the turn that follows the
-  answer. Ambient chatter is still dropped, as before.
+  now logged (once per workspace per minute, and again straight away for a new
+  decision), naming how many events it blocked and which delegated tasks they
+  belonged to. A worker's own events are no longer thrown away either: they are
+  parked in the same durable backlog that already holds events for a workspace
+  with no orchestrator, so they survive an app restart — the decision does —
+  and answering the decision replays them into the turn that follows. Answering
+  also restores the auto-wake budget the wait may have eaten. Ambient chatter is
+  still dropped, as before.
 
 - `fanout_start` now warns you at the moment it accepts when the launching
   workspace has a pending decision, instead of starting workers whose reports
-  will not reach anyone until it is answered.
+  will not reach anyone until it is answered. The warning leads the tool's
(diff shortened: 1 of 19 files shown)


openwong2kim:
Makes the orchestrator's pending-decision gate audible — blocked wakes logged, delegated events held for replay instead of consumed — and refusal and ledger messages now name their reasons and legal moves. Risk is concentrated in CommanderEventCoalescer buffer-retention semantics; well tested and the protocol baseline SHA is updated. P2 — closes a silent kill switch in the delegation loop.

PR #1216, opened 2026-09-04
fix(fanout): a worker's model is wmux's choice, not the operator's shell profile

openwong2kim:
## Summary

Wave 3 finding 15 (`docs/internal/dogfood-orchestrator-2026-09.md`, second run): every fan-out worker's first turn died with "issue with the selected model (glm-5.3)". The worker command is typed into the pane's interactive login shell after its rc files ran, so an `ANTHROPIC_MODEL` exported by `~/.zshrc` overrides anything the spawn env set. Reproduced three times, recovered by hand each time.

- `workerLaunchCommand` prefixes the launch with `env -u ANTHROPIC_MODEL` on POSIX for a `claude` worker that carries no `--model` and no role. A roled task is left alone on purpose: the prefix would hide the launcher stem from the renderer's role binding (`applyRoleAgent` / `applyRoleBinding`) and drop the binding's agent and model. win32 gets a logged note instead of a prefix.
- `ANTHROPIC_BASE_URL` / `ANTHROPIC_AUTH_TOKEN` are deliberately not scrubbed: routing claude through a proxy for every pane is the operator's call; only the model is wmux's.
- The recorded `initialCommand` is the line actually launched; `stripLaunchEnvPrefix` lets the first-run watch still read the launcher stem through the prefix.
- The first-run detector recognises the model error screen (`issue with the selected model (…)`) as a new kind that is never pressed; the ledger row moves to `input_required` with a summary naming the model and the fix.

## Test plan

- [x] 13 new tests: prefix rules (plain claude, `--model`/`-m` forms, non-claude stem, roled task, win32, environment-only launch), prompt-path quoting with a real `sh -c` round trip, prefix stripping, stuck summary wording, FanOutService records the launched command, model-error screen → `input_required` with zero keys sent, detector on the exact dogfood text.
- [x] `tsc --noEmit` root + daemon, eslint on touched files, vitest over worktask/pty/pipe handlers: 70 files / 1,587 tests.
- [ ] Live: fan-out on the demo instance with the shell profile still exporting `ANTHROPIC_MODEL` (wave 3 third run).

Known gap: a roled task whose binding has no model still launches unprotected (main cannot see renderer-side bindings); the detector is the backstop.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fan-out Claude workers now honor the model selected by the application instead of an overriding shell setting.
  - Workers encountering a selected-model error during their first turn are promptly marked as requiring input, with guidance to resolve the issue.
  - Model errors identify the rejected model and its source when available.
  - Existing explicit model selections, role-based configurations, gateway routing, and supported shell behavior remain unchanged.
  - Delayed model errors appearing after the worker screen loads are now detected reliably.

### DIFF
--- a/changelog.d/fanout-worker-model-env.md
+++ b/changelog.d/fanout-worker-model-env.md
@@ -1,0 +1,19 @@
+### Fixed
+
+- **A fan-out worker no longer launches on whatever model your shell profile
+  exports.** The worker's command is typed into the pane's interactive login
+  shell, so an `ANTHROPIC_MODEL` exported by `~/.zshrc` (or any other rc file)
+  overrode the environment wmux had set and every worker's first turn came back
+  "There's an issue with the selected model", indistinguishable from an idle
+  worker. A `claude` worker is now launched with that variable neutralised on the
+  command line itself, which is the only place late enough to win. A launch that
+  already names a model — an explicit `--model`, or a role bound to an agent and
+  model in Settings — is left exactly as it is. `ANTHROPIC_BASE_URL` and
+  `ANTHROPIC_AUTH_TOKEN` are deliberately untouched: routing claude through your
+  own proxy is a whole-machine choice, and a worker that quietly bypassed it
+  would be talking to a different endpoint than every other pane you open.
+
+- A fan-out worker whose first turn fails on its model is now reported as
+  `input_required` in the task ledger, with the model it was refused and the
+  `/model` fix, instead of sitting in `working` next to an idle-looking pane.
+  wmux reports that screen rather than typing at it.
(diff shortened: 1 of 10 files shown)


openwong2kim:
Neutralises a shell-exported `ANTHROPIC_MODEL` for plain claude fan-out workers via an `env -u` prefix on the launch line, and reports a first-turn model error as `input_required` instead of `working`. Changes stay inside fan-out launch construction and first-run detection, with roled, win32, and explicit-model cases deliberately skipped. P2 — it unblocks fan-out dogfooding, which failed this way in three runs.

PR #1224, opened 2026-09-05
feat(status): a Claude Code pane is running because its hook said so, not because it printed 2KB

openwong2kim:
## Summary

"Running" for a Claude Code pane is now a hook fact, not a byte-rate guess.

- **Turn start = `UserPromptSubmit`.** Registered on both install paths (plugin `hooks.json` 0.3.0 → 0.4.0, `wmux setup-hooks`); both ingest paths (daemon `HookIngest`, main `hooks.rpc`) broadcast `running` the moment the prompt is submitted. Fires once per turn, so it is compatible with the 2026-07-13 decision that removed the per-tool `PostToolUse` bridge.
- **Hook-governed panes ignore byte activity and silence.** Once a turn start has been seen for a pty (a per-agent latch in `HookSignalRouter`, 30-min TTL), `ActivityMonitor` neither re-stamps `running` nor clears to `idle`. The renderer keeps its own open-turn latch (`surfaceTurnOpenAt`), so a quiet turn longer than the old 120 s TTL stays lit.
- **Settle paths for a governed pane:** `Stop` → complete; `StopFailure` → error (red ✕, new kind `agent.stop_failure`); OSC 133 shell back at prompt → idle (the prompt-speed path; needs no process attribution); Ctrl+C / ESC ESC typed into a latched pane → idle (Claude Code fires no Stop on interrupt and stays the foreground command, so neither hooks nor OSC 133 can see it); process death edge (when the tracker attributed the pane) → idle; latch expiry → idle. None of them overwrite an unread `complete` / `awaiting_input` / `waiting` / `error`.
- **Unverifiable display state.** A latched pane with no signal for 30 min draws a hollow amber ring and "No update for 30m+" instead of a confident dot. Derived in the renderer; `AgentStatus` and the MCP pane_list schema are unchanged.
- **Orchestrator:** the Deck brain folds a re-submitted prompt into the open foreign turn only within 2 s or for identical text, so a new prompt after an ESC interrupt is a new turn. Stop-gate tests pin that a worker counts as running immediately after its prompt.

Panes with no turn-start evidence (plugin < 0.4.0, stale `setup-hooks`, hook-less agents) keep the byte heuristic exactly as before. `wmux setup-hooks --status` gained a "turn-start signal" row and reports a < 0.4.0 plugin as STALE.

## Review

Three-model panel (Claude + GLM; Codex unavailable), one round, 13 findings, all applied: renderer turn latch (running was previously re-derived from a decaying stamp, so the flag was unreachable), latch expiry timer, no latch on the daemon-down fallback, latch keyed by agent, process-exit guard (status + slug + no name clear), brain fold window, doctor plugin-version gate, fallback `agentName`, single-pass unverifiable roll-up, label cap.

## Live dogfood (Claude Code 2.1.236, `WMUX_DATA_SUFFIX` dev instance)

- Prompt submit → amber at t=0 (before any output); Stop → green at ~3 s; bridge log shows UserPromptSubmit/Stop reaching the daemon.
- 150 s quiet `sleep` turn: dot stayed amber through 165 s (old TTL would have dropped it at 120 s).
- Roster row and workspace dot agree.
- Single Ctrl+C mid-turn → "Interrupted" in Claude, sidebar idle within 2 s (the settle also clears the renderer's activity stamp and guards the redraw burst for 3 s, otherwise the byte TTL re-lit the dot for 2 minutes).
- Observed: `AgentProcessTracker` never attributed the pane in this environment (`agentAliveByPtyId` empty), and Claude Code did not emit `StopFailure` after 10/10 connection-refused retries — which is why the OSC 133 settle path exists. Both hooks stay registered.

## Follow-ups (not in this PR)

- Tee `agent.stop_failure` into `agent.lifecycle` so an orchestrator wakes when a worker's turn dies on an API error (touches the MCP events_poll description / token budget).
- A fresh Claude session at its prompt still reads as `waiting` (red "need you") from the footer regex — pre-existing, worth a separate pass now that needs-you has a rendition.

## Test plan

- [x] `tsc --noEmit` (only the pre-existing `BrowserReplSession.ts` error, also on main)
- [x] `env -C <wt> npx vitest run` over main/daemon/shared/cli/renderer/integrations: 13543 passed; the single failure is the known symlinked-node_modules `atlasCoherence` false negative
- [x] Live dogfood above

Existing installs: plugin users run `/plugin update` in Claude Code; plugin-less users re-run `wmux setup-hooks` (the app also refreshes on boot). Until then behaviour is unchanged.

## Summary by CodeRabbit

* **New Features**
  * Claude integration 0.4.0 adds prompt-start and API-failure status signals.
  * Panes turn amber on prompt submission and show a red error state for API failures.
  * Interrupts, agent exits, and returns to the shell prompt settle panes promptly.
  * Running panes without updates for 30 minutes show a hollow amber indicator and elapsed-time tooltip.

* **Bug Fixes**
  * Reduced status flicker and corrected inconsistent sidebar and roster states.
  * Prevented duplicate turn entries when prompts are resubmitted.

* **Documentation**
  * Updated integration documentation and setup guidance; existing installations may require rerunning hook setup.


openwong2kim:

Reworks Claude Code pane status from a byte-rate guess to hook facts: `UserPromptSubmit` starts a turn, `StopFailure` errors it, with new settle paths (interrupt keystroke, OSC 133, process exit). 87 files span the daemon→renderer status pipeline, so regressions would misreport every agent pane. P2: broad-surface core change, well tested, old heuristic kept as fallback.

PR #1225, opened 2026-09-05
fix(status): failed worker turns wake the brain, a fresh session is not needs-you, deck dots match the sidebar

openwong2kim:
## Summary

Three follow-ups to #1224 (stacked on `feat/hook-driven-running`; retarget to `main` after it merges).

- **An orchestrator wakes when a worker's turn dies on an API error.** `agent.stop_failure` now has its own lifecycle kind (`src/shared/events.ts`), is teed on both ingest paths (main `hooks.rpc`, daemon `DaemonNotificationRouter`) on emit *and* dedup like `agent.stop`, and the Deck `CommanderEventCoalescer` wakes the brain with a distinct verdict ("turn died on an API error — the work is incomplete") instead of folding it into a normal stop. `events.poll` needed no code change (string set); `wmux_events_poll` gained the kind name only (+19 B, full profile 77,767 / 78,000).
- **A Claude session sitting at its prompt is not "needs you".** The always-visible footer (`bypass permissions on` / `shift+tab to cycle`) mapped to `waiting` from the moment a session started (`SessionStart` grants hook authority but not lifecycle ownership), so a fresh `claude` — and every `/clear`, resume, compact — showed the red dot and "1 need you". For hook-governed panes the detector's `waiting` is now withheld outright; `awaiting_input` (permission prompts) is untouched, ungoverned panes unchanged. The post-Stop repaint was already covered by lifecycle ownership, so there is no green-to-red flip to fix.
- **Deck fleet roster derives running like the sidebar.** `DeckFleet` passed an inline fixture to `selectFleetPanes` without the activity / turn-latch / OSC 133 / process-alive slices, so its dots could never light from the latch. It now passes the real slices.

## Review

Three-model panel (Claude + GLM; Codex unavailable), one round, 10 findings, all applied: the failure tee is emitted on arrival (not from inside the alarm window), a vetoed footer read still traces its turn end for `events_poll` (60 s throttle), `status:'error'` without a hookKind falls back to `stop_failure`, `error` joined the attention statuses so the Deck snapshot sees a failed background worker, a pane that keeps dying is escalated after 3 consecutive failures instead of resumed forever, hook retries tee once per turn, the fleet roster no longer subscribes to the decay clock, and the docs / MCP `kinds` description name the new kind. Full MCP profile 77,840 / 78,000 B.

## Test plan

- [x] `tsc --noEmit` (only the pre-existing `BrowserReplSession.ts` error)
- [x] `env -C <wt> npx vitest run` main/daemon/shared/cli/renderer/mcp/integrations: 14800 passed; the single failure is the known symlinked-node_modules `atlasCoherence` false negative
- [x] `node scripts/build-mcp.js && npm run probe:mcp` green; baseline shas regenerated for the one-kind description change
- [x] Live: fresh `claude` in a `WMUX_DATA_SUFFIX` dev instance built from this branch → gray dot, no "need you" (was red before)

## Summary by CodeRabbit

- **Bug Fixes**
  - API-failed agent turns now wake orchestrators and are clearly distinguished from completed turns.
  - Repeated failures on the same pane escalate to a request for human guidance after three attempts.
  - New sessions at an idle prompt no longer incorrectly show “Waiting,” a red status dot, or “need you” counts.
  - Fleet and sidebar running indicators now remain consistent.
  - Approval prompts continue to trigger the expected notifications.

- **Documentation**
  - Added guidance for handling API-failed turn events.


openwong2kim:
## Review panel (Claude + GLM, one round; Codex unavailable)

**Two-model findings (all applied)**
- `waiting` withhold was also killing the dedup-ledger write and the `agent.lifecycle` tee for panes whose bridge died after `SessionStart` → the turn end is now still traced (`decision:'internal'`, 60 s per-pane throttle) while only the status broadcast is withheld.
- `lifecycleKindFor` fallback mapped a bare `status:'error'` to `agent.stop` → now `agent.stop_failure`.
- Repeated `stop_failure` on one pane looped wake → resume → fail → the coalescer counts consecutive failures and tells the brain to escalate via `deck_ask_decision` at 3.
- `DeckFleet` subscribed to the decay clock and re-ran the fleet selector every 2 s → it now reads a per-pty running map.
- Hook retries teed the same failed turn twice → one tee per (pty, kind) turn.

**Single-model findings (applied)**
- Claude CRITICAL: the failure tee lived inside the CompletionAlarm resume closure and was lost if the 1.5 s window was rebutted → emitted on arrival; the alarm gates only the toast.
- `error` was not an attention status, so the Deck level snapshot could never see a failed background worker; snapshot renders it with the failure verdict and the value-filtered policy keeps it.
- Stale rationale comment in `HookIngest.arbitrateDetector` rewritten.
- MCP baseline freshness verified by rebuilding from a reverted tree (probe passed against the committed shas) and again after the change; full profile 77,840 / 78,000 B.
- `docs/how-to/react-to-events.md` and the `wmux_events_poll` `kinds` description name `agent.stop_failure`.

**Deliberately not done**
- The vetoed-footer trace does not wake the Deck brain (`internal` only): a footer read mid-turn is not a turn boundary the brain should act on; `events_poll` consumers see it.

openwong2kim:

Publishes `agent.stop_failure` as its own lifecycle kind, withholds footer-driven `waiting` on governed panes, and fixes the Deck roster's running derivation. Touches the published event union, the #935 veto predicate, and coalescer verdicts, each with dedicated tests. P2 — additive and well-reviewed, but it re-cuts hook/detector arbitration and stacks on #1224, so merge order matters.

PR #1229, opened 2026-09-07
fix(audit): count only external-wire callers as legacy traffic (#1111)

openwong2kim:
## What

Pre-close evidence hygiene for the #1111 grandfather-lane close (announced for 2026-09-30, notice #1130).

The legacy contact recorder (`RpcRouter` → `plugin-trust.json`) and the per-method legacy traffic counter (`shadow-rejections.log`) fired for **every** envelope-less dispatch — including the trusted in-process surfaces. The renderer IPC bridge (`operator`) and the iframe plugin host (`firstParty`) send no `clientName` by design, so the audit trail that must answer "which external no-clientName callers still exist?" was dominated by wmux's own UI polling.

Measured on the primary dogfood box (`~/.wmux/shadow-rejections.log`, 2026-06-27 → 2026-09-07):

| method | counted calls | actual source |
|---|---|---|
| `events.poll` | 445,504 | renderer polling (`useRpcBridge`) |
| `task.mission.list` | 191,350 | renderer polling (`AppLayout` mission cache — identical daily-count pattern to `events.poll`, same caller) |
| `a2a.channel.getMembers` / `.list` | ~2.4k | renderer channel surfaces + ad-hoc scripts |
| `hooks.signal` | 1,248 | integrations hook bridge — the one true wire caller, already handled by draft PR #1139's curated hook-bridge lane |

Both side channels now gate on `isLocalExternalWireContext` — the same positive provenance predicate the enforcer lanes use. The in-process surfaces were never the grandfather's audience (#1139 exempts them at the enforcement points for the same reason). After this lands, every new `legacy-traffic` entry in the log is a real external envelope-less wire caller, which is exactly what the 2026-09-30 close decision needs to watch during the remaining notice window.

## Changes

- `src/main/pipe/RpcRouter.ts` — legacy recorder + counter block gated on `isLocalExternalWireContext(ctx)`; comments updated.
- `src/main/pipe/__tests__/RpcRouter.test.ts` — legacy suites dispatch with `{ externalWire: true }`; two regression tests: operator / firstParty lanes never tick either channel.
- `src/main/mcp/__tests__/phase-2-1-e2e.test.ts`, `phase-2-2-dynamic.test.ts` — legacy-path steps dispatch as wire callers.
- `src/main/audit/shadowRejectionLog.ts` — entry-kind doc precision.
- `docs/PROTOCOL.md` §4.1 — recording semantics: wire-only; in-process surfaces not recorded.

Note for draft PR #1139: it will want the same wire-only phrasing when it rebases (its §4.1 rewrite touches the same paragraph).

## Verification

- `vitest run src/main/pipe src/main/mcp src/main/audit` — 64 files, 1427 tests passed.
- `tsc --noEmit` — one pre-existing error on clean `main` (`src/mcp/browser-repl/BrowserReplSession.ts:182`), unrelated to this diff (verified via stash).
- `eslint` on touched files — clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved legacy-traffic auditing so records reflect external wire requests only.
  - Prevented trusted internal renderer and plugin-host activity from being incorrectly counted as legacy traffic.
  - Improved accuracy of audit thresholds and rejection logs used to evaluate legacy traffic.

- **Documentation**
  - Clarified protocol documentation regarding which requests are classified and recorded as legacy traffic.

### DIFF
--- a/changelog.d/1229.md
+++ b/changelog.d/1229.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Legacy-traffic audit now counts only external-wire callers.** The shadow log and trust-DB legacy records exist to show which no-`clientName` integrations are still calling before the grandfather lane closes (#1111, announced for 2026-09-30). They were also counting wmux's own renderer bridge and plugin host, which send no `clientName` by design — on a dogfood box that noise was ~636k of the ~639k recorded calls (`events.poll` 445k, `task.mission.list` 191k), drowning the signal the close decision relies on. Both audit channels now use the same wire-provenance check the permission lanes use, so anything new in `shadow-rejections.log` is a real external caller. (#1229)
--- a/docs/PROTOCOL.md
+++ b/docs/PROTOCOL.md
@@ -329,7 +329,7 @@ Either default can be overridden explicitly via the config key. The enforcement 
 
 Plugin identity rides the JSON-RPC envelope as `clientName` (and optional `clientVersion`). The MCP server (`src/mcp/index.ts`) populates them from the MCP `InitializeRequest.clientInfo` so the substrate can attribute every call to a declared plugin. Identity is **declared, not verified** — there is no root-of-trust today; threat-model details live in `api/mcp-plugin-spec.md`.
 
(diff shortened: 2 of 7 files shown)


openwong2kim:
This gates the legacy recorder and traffic counter on wire provenance, so the renderer bridge and plugin host no longer flood the #1111 evidence with in-process polling noise. It touches RpcRouter.ts, its tests, and PROTOCOL.md — narrow and reversible. P2 — the audit trail must be clean before the grandfather close on 2026-09-30.

Issue #1002, opened 2026-08-22
[Bug]: Splitting panes causes the initial pane to scroll from the start of the conversation


RudVlad473:
### What happened?

https://github.com/user-attachments/assets/9faba6eb-5682-46b1-84c4-c0166a170ae3

### Steps to reproduce

1. Have a long conversation with some kind of chat (in my case Claude)
2. Split pane horizontally
3. The initial pane 'restarts' and scrolls from the start of the convo to the botton

### Expected behavior

1. Have a long conversation with some kind of chat (in my case Claude)
2. Split pane horizontally
3. Initial pane is already at the bottom and doesn't have to scroll

### wmux version

3.44.0

### How did you install wmux?

Setup.exe (direct download)

### Shell in the affected pane

PowerShell 7+ (pwsh)

### Windows version

Windows 10 19045.6466

### AI agent involved (if any)

Claude Code

### Logs, screenshots, or recordings

```shell

```

### Confirmation

- [x] I searched existing issues and this is not a duplicate.
- [x] I am on Windows 10 or 11 (wmux is Windows-only).

openwong2kim:
## [wmux-hermes] Triage Summary
Splitting a pane horizontally makes the original pane replay its buffer from the top of the conversation and scroll down instead of staying at the bottom. This is pane-split scroll-position handling; low risk (no data loss), but disruptive during long agent sessions. P3 — a UX annoyance in a common workflow, worth batching with other scrollback fixes.

openwong2kim:
Confirmed the mechanism — this is not a repaint, the pane is genuinely being rebuilt.

**What happens on a split.** `splitPane` (`src/renderer/stores/slices/paneSlice.ts:502`) replaces the target leaf with a *new* branch node that carries a fresh id and holds the old leaf as a child. In the render tree that means one of two things, both of which React resolves by unmounting:

- the leaf was the workspace root → `PaneContainer` switches from returning `<Pane>` to returning `<Group>` at the same position (different component type → unmount);
- the leaf had a parent → the parent's `<Fragment key={child.id}>` key changes from the leaf id to the new branch id (different key → unmount).

Measured it rather than assuming: mounting the real `PaneContainer` against the real store with the leaf `Pane` mocked to log its mount/unmount, then calling `splitPane` on the root leaf, gives

```
MOUNTS   [pane-1168a916, pane-1168a916, pane-4de6c20a]
UNMOUNTS [pane-1168a916]
```

[...]

PR #1230, opened 2026-09-07
fix(term): repair idle-pane ghost glyphs with a post-settle verify repaint

openwong2kim:
## Problem

An idle pane can show a ghosted/stale frame indefinitely: the text buffer is correct (selecting the text instantly repairs it), but the pixels on screen are stale. Observed with opencode TUIs — CJK + box-drawing heavy, several agent panes streaming at once — the 2026-09-07 dogfood report of this exact symptom, and the same dirty-region desync class as #166/#318/#319 ("garbled glyphs until resize").

## Root cause

The trailing settle flush is a stream's final full-range repair, but its own WebGL raster can race and paint stale pixels on rows it just marked clean. Once the pane goes idle, no repaint trigger can ever fire again:

- `focus` never re-fires on a pane that is already focused
- `visible` needs a workspace switch
- `burst` needs output

So the raced final paint persists until the user manually selects text (selection forces `renderRows` from the correct buffer).

## Fix

`glyphRepaint` arms a one-shot **`settle-verify`** repaint `settleVerifyMs` (default 2s) after every settle flush — after the settle's own raster has committed — so a raced final paint cannot outlive ~`burstQuietMs + settleVerifyMs`. Cost: at most one extra refresh per stream that flushed (overlapping tails coalesce); noise-only trickles schedule nothing; the periodic-pulse standing cost (~0.4 refresh/s ceiling for `watch`-style panes) is documented and accepted. The verify is deliberately not cancelled by sub-floor writes (a keystroke echo would otherwise reopen the hole), and `dispose()` clears it.

## Review hardening (3 specialists + adversarial pass)

- **Adoption hole fixed (#1002 park/adopt):** the previous mount's scheduler death killed an armed verify; a pane split/drag-restructured within ~2s of stream settle carried the ghost with no self-repair left. The adopting mount now does one full-range refresh.
- Header cost claim corrected ("at most one", not "exactly one"), re-arm semantics comment fixed (latest settle wins).
- A fragile pre-existing test isolated from the new timer; new regression tests pin that only a settle (not a mid-stream flush) arms the verify, and that the useTerminal visibility gate covers both output-driven reasons.

## Verification

- `glyphRepaint.test.ts`: 28 tests (6 → 8 new for the verify path) — all pass
- `useTerminal.atlasClear.test.ts`: gate pin added — all pass
- Full `src/renderer/terminal/__tests__` + `src/renderer/hooks/__tests__` run: 1110 passed; the single failure (`atlasCoherence` I1/I3 patch marker) is pre-existing on `main` (node_modules shipped without the addon-webgl patch — separate infra issue, worth a look)
- eslint clean on changed files (remaining errors pre-exist on `main`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where terminal panes could display stale or ghosted content after output stopped.
  * Added a delayed verification repaint to automatically repair rendering issues in idle panes.
  * Improved terminal rendering when panes are adopted, split, or dragged after command completion.
  * Prevented unnecessary repainting for hidden panes.

### DIFF
--- a/changelog.d/1230.md
+++ b/changelog.d/1230.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **An idle pane no longer keeps a ghosted frame until you select its text.**
+  The final repaint after a stream of output could itself race on the GPU and
+  leave stale pixels on rows it had just marked clean — and once the pane went
+  quiet, nothing ever repaired it: focus doesn't re-fire on a pane you're
+  already looking at, and repairs were only scheduled while output flowed or on
+  workspace switches. Selecting the text fixed it instantly (the buffer was
+  always correct), which is what made it so annoying to live with. A pane now
+  schedules one more verification repaint a couple of seconds after output
+  settles, so a raced final paint self-heals instead of haunting the pane — and
+  a pane that gets split or dragged right after a command finishes heals too.
--- a/src/renderer/hooks/__tests__/useTerminal.atlasClear.test.ts
+++ b/src/renderer/hooks/__tests__/useTerminal.atlasClear.test.ts
@@ -39,4 +39,16 @@ describe('#191 — repaint must not clear the shared texture atlas (source-level
   it('repairs staleness with a full-range refresh instead', () => {
     expect(repaintBlock).toMatch(/terminal\.refresh\(0, terminal\.rows - 1\)/);
   });
+
+  // The repaint's output-driven reasons ('burst' and the idle-tail
+  // 'settle-verify') must stay behind the visibility gate: a hidden pane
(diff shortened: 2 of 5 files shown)


openwong2kim:
This arms a one-shot settle-verify repaint ~2s after each settle flush, so a raced final raster on an idle pane self-heals instead of ghosting until text selection. It touches glyphRepaint.ts and useTerminal.ts (adopt-path refresh, visibility gate) with tests pinning arm/re-arm/dispose semantics; risk is low. P2 — a user-visible rendering fix with bounded cost.

PR #1231, opened 2026-09-07
feat(codex): map the measured PermissionRequest hook to agent.awaiting_input (#1107)

openwong2kim:
## What

Closes the biggest remaining gap of #1107: `PermissionRequest` — the Codex lifecycle event that lets wmux stop screen-scraping Codex approval pauses — is now **measured and mapped**.

### The measurement

`codex exec` forces `approval: never`, so the 2026-08-31 spike could never fire it. This one drove the **interactive TUI in a PTY** (node-pty) against a stub Responses endpoint returning an `exec_command` call with `sandbox_permissions: "require_escalated"` — the model-side request for an unsandboxed run, which is what raises the TUI approval dialog. codex-cli 0.153.4, isolated `CODEX_HOME`, `--dangerously-bypass-hook-trust` for the measurement only.

Firing order for one gated call, all events carrying the same `turn_id`:

```
PreToolUse → PermissionRequest → (operator approves in the TUI) → PostToolUse → Stop
```

Captured payload (elided):

```json
{"session_id":"…","turn_id":"…","transcript_path":"…","cwd":"…",
 "hook_event_name":"PermissionRequest","model":"stub-1","permission_mode":"default",
 "tool_name":"Bash",
 "tool_input":{"command":"echo proof > /tmp/…","description":"write the spike proof file outside the sandbox"}}
```

Three facts that shaped the mapping: `tool_name` is Claude-normalized; `tool_input.description` is the call's user-facing `justification` (content — never forwarded, the envelope allowlist is the enforcement, asserted on the serialized envelope); and there is **no `tool_use_id`**, unlike the tool events around it.

### The mapping

`PermissionRequest` → `agent.awaiting_input` — the same pane state the three transcribed approval regexes in `AgentDetector.ts` produce — **not** `agent.awaiting_permission`, which is reserved for wmux's own blocking permission gate (#783: daemon holds the bridge RPC open until a phone resolves it), not the agent's local TUI dialog. The regexes stay as the fallback for panes whose operator never trusted the hook (the trust gate is silent, so that stays the common case).

### Changes

- Bridge `EVENT_TO_KIND` + `CODEX_HOOK_EVENTS`: `PermissionRequest` registered and mapped; `BRIDGE_VERSION` → 0.2.0.
- `codexHookEnvelope.test.ts`: measured payload fixture, mapping + metadata-only assertions; removed from the pinned unmapped list.
- `hookHarmlessness.mjs`: the real PermissionRequest payload (content fields included) added to the exercised events.
- README: "Not verified" section replaced with the measurement; firing-order, payload, and the two TUI dialogs the driver had to get past (directory trust — pre-answerable via `[projects."<path>"] trust_level = "trusted"` — and the update nag whose default runs a global npm install).
- `AgentDetector.ts`: the #1107 comment on the three regexes updated — hook is primary, regexes are the fallback.

## Verification

- `vitest run integrations/codex` — 35 passed.
- `vitest run src/main/pty/__tests__/AgentDetector.test.ts src/shared/hooks` — 182 passed.
- `hookHarmlessness.runtime.test.mjs` (runs the bridge against a fake daemon) — 18 passed, including the new PermissionRequest case.
- eslint on touched files clean (one pre-existing `no-control-regex` in `AgentDetector.ts:429` exists on clean `main`).

## What remains on #1107

- Install flow honesty (approve-then-verify in `lifecycleIntegrations`).
- Adding this bridge to `hookBridge.lockstep.test.ts`'s `BRIDGES` list when #1111 lands.

### DIFF
--- a/changelog.d/1231.md
+++ b/changelog.d/1231.md
@@ -1,0 +1,3 @@
+### Added
+
+- **Codex approval pauses are now reported by a hook instead of a screenshot guess.** The one Codex lifecycle event wmux could never measure — `PermissionRequest`, the moment a Codex pane stops and shows "would you like to run this command?" — finally fired under a lab rig (interactive TUI driven in a PTY against a stub model endpoint), and the hooks bridge now maps it to the same "needs a human" state the approval screen-regexes produce. A pane whose operator has trusted the hook gets the fact; the regexes stay as the fallback for everyone else. The captured payload confirmed Codex even names the tool the way Claude does, and the user-facing justification it carries never leaves the bridge. (#1231)
--- a/integrations/codex/README.md
+++ b/integrations/codex/README.md
@@ -8,7 +8,7 @@ different routes, and you want **one** of them, not both.
 | Registered as | `notify = [...]` in `config.toml` | `[[hooks.<Event>]]` in `config.toml` |
 | Payload arrives | last argv token | stdin |
 | Codex floor | any | **0.141.0** |
-| Reports | turn complete | turn complete, turn start, session start |
+| Reports | turn complete | turn complete, turn start, session start, approval pause |
 | Needs operator approval | no | **yes** (trust gate) |
 | Installed by wmux | yes (`lifecycleIntegrations`) | no — see *Installation* |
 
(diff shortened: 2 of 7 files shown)


openwong2kim:

Maps the measured Codex `PermissionRequest` hook to `agent.awaiting_input`, replacing the screenshot guess for panes with a trusted hook while the AgentDetector regexes stay as fallback. Touches only the codex bridge, its tests, and comments — the envelope stays metadata-only (asserted) and the bridge version bumps to 0.2.0. P2: clean, well-verified step on the #1107 track.

PR #1232, opened 2026-09-07
fix(hooks): TTL-exempt authority for turn-boundary-only bridges (#1009)

### LINKED ISSUE #1009: Signals-only leaves nothing to refresh hook authority, so the #935 status veto lapses 30 minutes into a turn
## What I found

`--signals-only` (#979) installs exactly the hook set the #935 report was filed
on — `Stop`, `SubagentStop`, `SessionStart`, plus the `AskUserQuestion`
`PreToolUse`/`PostToolUse` pair, and nothing per tool call. The fix that closed
the roster half of #935, `governsDetectorStatus`, rides
`HOOK_AUTHORITY_TTL_MS` (30 min). On the **full** profile that TTL is never
approached, because the wide gate hook refreshes it on every tool call even
when it is dormant:

`src/daemon/index.ts:3031` → `ingest.handle({ ...signal, kind: 'agent.tool_started' })`
→ `HookIngest.ts:712` → `router.touchAuthority(...)`.

Take the gate away and the only things left that touch authority are turn
boundaries. So on a signals-only install, a turn that runs longer than 30
minutes without a subagent stop or an `AskUserQuestion` **loses hook authority
mid-turn**, and the pane falls back to the detector's always-visible-footer
read — which is #935 exactly: `bypass permissions on` reads as `waiting`, lands
on the roster row and in the "N need you" roll-up, and pre-poisons the dedup
ledger so the real `Stop` arrives as `dedup`.

The margin is not hypothetical. #935's own snapshots caught
`Whirlpooling… (21m 27s)` and `(22m 18s)` on that same hook set — turns still
running at 9 minutes from the TTL, in the report that motivated the veto.

## Why I do not think this is just "the TTL is what it is"

The comment on `governsDetectorStatus` already reasons about the window, but in
the other direction:

> A shorter TTL is not the answer either: a single turn can run past twenty
> minutes with no bridge traffic at all on a turn-boundary-only hook install
> (the #935 report measured 21m), so a short window would simply restore the
> bug it fixes.

That is the same scenario — and the conclusion drawn from it was to keep 30
minutes, on the strength of `HOOK_AUTHORITY_TTL_MS`'s own premise: "long enough
to span a long tool call between hook signals." Before #979 that premise was
underwritten by the gate, which spoke on every tool call. `--signals-only` is
the first *supported* configuration in which nothing speaks between turn
boundaries, so the premise and the number stopped agreeing on 2026-08-21.

## Demonstration

Deterministic, real TTL, no clock mocking — `HookSignalRouter` takes `now`:

```ts
const t0 = 0;
const midTurn = t0 + 45 * 60_000; // a 45-minute turn, no Stop yet

// signals-only: SessionStart, then one Stop closing the previous turn.
const a = new HookSignalRouter({ latencyMeter: meter });
a.touchAuthority('p1', 'claude', t0, true, 'agent.session_start');
a.touchAuthority('p1', 'claude', t0 + 1_000, true, 'agent.stop');
a.governsDetectorStatus('p1', 'claude', 'waiting', midTurn); // => false  (lapsed)

// full: identical, plus the dormant gate's per-tool-call touch.
const b = new HookSignalRouter({ latencyMeter: meter });
b.touchAuthority('p1', 'claude', t0, true, 'agent.session_start');
b.touchAuthority('p1', 'claude', t0 + 1_000, true, 'agent.stop');
for (let t = t0; t <= midTurn; t += 60_000) {
  b.touchAuthority('p1', 'claude', t, true, 'agent.tool_started');
}
b.governsDetectorStatus('p1', 'claude', 'waiting', midTurn); // => true   (governed)
```

Both hold, and the boundary is exact — the same file pins `true` at
`HOOK_AUTHORITY_TTL_MS - 1` and `false` at `HOOK_AUTHORITY_TTL_MS`, measured
from the **previous** turn's `Stop`, which is where the clock actually starts:

```
✓ signals-only: the veto has lapsed mid-turn
✓ full: the dormant gate keeps the veto alive for the whole turn
Tests  2 passed (2)
```

Same router, same TTL, same turn: the profile alone decides whether the roster
believes the footer.

## What I am not proposing

Not a shorter TTL, and not a patch — the shape depends on an answer only you
can give. Three that fit the code as it stands:

1. **Make the TTL a function of the profile** (longer, or none, for
   signals-only). `expireAuthorityFor` already releases authority on *confirmed*
   process death from the liveness poll (`src/daemon/index.ts:5682`), so the
   TTL's remaining job is the narrow "bridge dead, process alive" case — a
   bridge killed with `-9`. Under signals-only that job could be left to the
   poll.
2. **Latch lifecycle ownership across the turn**: `SessionStart`/`Stop` bracket
   it, and `lifecycleOwned` already tracks something close to this.
3. **Leave it, and say so** — `--status` and the README already distinguish
   "absent because you chose the profile"; this would be one more consequence
   of choosing it, alongside remote approvals being off.

## The question

Is the 30-minute authority TTL a **freshness** guarantee (the bridge must keep
speaking to stay canonical) or a **death** backstop (only a bridge we believe is
gone should lose authority)? On the full profile the two are indistinguishable,
because the gate speaks constantly. `--signals-only` is what separates them, and
which one it is decides all three shapes above.

Filing rather than patching for the same reason as #970: three plausible shapes
follow from three different answers.

---

Verified against `main` at `0687bc4b` (v3.46.0), Windows 11. Happy to write whichever shape you pick.

openwong2kim:
## What

Implements the direction call from [#1009](https://github.com/openwong2kim/wmux/issues/1009#issuecomment-5517294397): **the authority TTL is a death backstop, not a freshness guarantee** — via the self-describing latch the owner specified, not install-profile plumbing.

### The bug

The 30-minute authority TTL was underwritten by the dormant permission gate, which touches authority on every tool call. `--signals-only` (#979) is the first supported profile where nothing speaks between turn boundaries, so a turn longer than 30 minutes lost the #935 detector-status veto mid-turn — the pane fell back to the always-visible-footer read: false `waiting`, roster red dot, "N need you" roll-up, real `Stop` pre-poisoned to `dedup`. #935's own snapshots caught 21-22 minute turns on exactly that hook set.

### The fix (as specified in the issue)

- An authority entry that has **ever** seen a per-tool-call kind (`agent.activity` / `agent.tool_started`) keeps the 30-minute freshness TTL. The full profile is unchanged, byte for byte — the dormant gate's synthetic touch is a tool kind.
- An entry that has **only ever spoken turn boundaries** gets no TTL. Released by confirmed process death (`expireAuthorityFor` + liveness poll), by relaunch (`agent.session_start` resets lifecycle ownership), or by pane dispose (`dropPty`) — nothing else.
- An entry with no kind named stays TTL-bound (pre-#1009 rule) — production callers always pass a kind, so this only preserves existing test semantics.

Because the latch reads what the bridge **does**, hand-edited hook configs and version-skewed bridges classify themselves correctly.

### The two asks, done

- `governsDetectorStatus`'s accepted-cost paragraph rewritten for the new rule: signals-only bridges killed with a live process leave `waiting`/`complete` withheld until death/relaunch — unbounded, but one-directional (`running` never vetoed → stale-busy, never false "needs you").
- The requested latch pin: a bridge that speaks `agent.activity` **once** and then goes boundary-only keeps the freshness TTL.

### Tests

- The issue's exact demonstration: signals-only 45-minute turn → veto holds (and stays held across further turns).
- Full profile: gate touches keep the veto alive mid-turn **and** age out at the TTL once quiet (both boundary edges).
- The owner's latch pin.
- The three release paths for TTL-exempt entries (death / relaunch / dispose).
- `authorityAgentFor` reports a TTL-exempt entry with true age — identity callers bound by their own `IDENTITY_TTL_MS` (2 min) on the uncorroborated path, so a signals-only pane no longer flips to screen-truth 30 minutes into a turn.

Out of scope, deliberately: the turn-start latch (`governsRunningState`) keeps its own timer — it releases the byte heuristic, a different and milder tradeoff with its own accepted-cost doc.

## Verification

`vitest run src/shared/hooks src/daemon/hooks` (231), `src/main/notification src/main/pty src/main/pipe/handlers/__tests__/hooks.rpc` (626 passed / 3 skipped), router suite 55/55. tsc clean (one pre-existing unrelated error on `main`); eslint clean.

Closes #1009.


openwong2kim:
Implements the #1009 direction call: turn-boundary-only bridges become TTL-exempt in HookSignalRouter, released only by death/relaunch/dispose. Touches the core authority path behind the detector veto and identity tier, with the latch pinned by tests on both edges. P3 to match #1009 — the failure it fixes is real but confined to signals-only installs.

PR #1235, opened 2026-09-07
fix(scripts): postinstall verifies the xterm patch landed instead of skipping silently

openwong2kim:
## Problem

The postinstall hook ran patch-package only when it was installed; when `patch-package` was absent it printed a `console.warn` and exited 0. That skip had one legitimate consumer (`npm ci --omit=dev`) but also silently masked a broken local state: a **hybrid node_modules** (dev deps present, patch-package missing — an `--omit=dev`-style install layered over a full one) skipped patch application with a warning nobody reads.

Observed fallout (2026-09-07 incident): the unpatched `@xterm/addon-webgl` made `atlasCoherence.test.ts` fail, and the partially-stripped type packages produced phantom `tsc` errors — poisoning every later diagnosis for days. Both vanished with a clean `npm ci`.

## Fix

`apply-patches.mjs` now verifies the patch actually landed after running (or skipping) patch-package, via `scripts/lib/patch-verify.mjs` — a probe of the installed addon-webgl bundle for the patch marker (the same string `atlasCoherence.test.ts` asserts; when patch-package never ran, EVERY patch is missing, so one marker detects the whole class):

| installed state | verdict |
|---|---|
| patched / probe target absent | pass |
| unpatched + dev deps present + no patcher (**the incident**) | **exit 1** with `npm ci` recovery hint |
| patcher ran but marker absent (patch drift) | **exit 1** with `npx patch-package @xterm/addon-webgl` hint |
| unpatched + genuinely `--omit=dev` | legacy skip preserved (packaging flows keep passing) |

## Verification (dogfooded)

- Unit tests: `scripts/__tests__/apply-patches-verify.test.mjs` — 5 tests over a fabricated fs tree (patched / absent / hybrid / prod / unreadable-probe)
- End-to-end dogfood in fabricated roots, all four states:
  - healthy full install → patches apply ✔, exit 0
  - hybrid (incident reproduction) → exit 1 + `Fix: run npm ci`
  - omit=dev → exit 0 + legacy warn
  - patch drift → exit 1 + regenerate hint
- Full `scripts/__tests__` suite: 345 passed, 1 skipped (pre-existing)

## Summary by CodeRabbit

* **Bug Fixes**
  * Install setup now verifies that required terminal-rendering patches were applied successfully.
  * Partially stripped development installs and patch failures now stop immediately with recovery guidance.
  * Production installs without development dependencies continue to complete successfully, with an appropriate warning.

* **Tests**
  * Added coverage for patched, missing-target, partially installed, production, and unreadable-install scenarios.

### DIFF
--- a/changelog.d/1235.md
+++ b/changelog.d/1235.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **A broken local install now fails loudly at install time instead of sabotaging you days later.** When patch-package was missing from node_modules — the signature of a partially stripped install — the postinstall hook skipped wmux's xterm.js patches with a console warning nobody reads. The unpatched terminal renderer and stale type packages then surfaced as a confusing scatter of test failures and phantom type errors, far from the cause. The hook now checks that the patch actually landed and fails the install on the spot with the recovery command, while genuine production installs (`npm ci --omit=dev`) keep passing exactly as before.
--- a/scripts/__tests__/apply-patches-verify.test.mjs
+++ b/scripts/__tests__/apply-patches-verify.test.mjs
@@ -1,0 +1,86 @@
+// Unit tests for the postinstall patch probe (scripts/lib/patch-verify.mjs).
+//
+// The 2026-09-07 incident: a hybrid node_modules (dev deps present,
+// patch-package missing) let the postinstall hook skip patch application
+// silently; the damage surfaced days later as an atlasCoherence failure and
+// phantom tsc errors. These tests pin the probe's four verdicts so the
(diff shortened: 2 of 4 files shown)


openwong2kim:

This makes the postinstall hook verify the @xterm/addon-webgl patch landed (apply-patches.mjs + patch-verify.mjs), failing loudly on hybrid installs and patch drift while keeping the omit=dev skip. Only install tooling is touched; the four verdicts are unit-tested and dogfooded. P3: dev-infra hardening, no user-facing urgency.

PR #1236, opened 2026-09-07
feat(codex): wire the hooks bridge into setup-hooks — approve-then-verify (#1107)

openwong2kim:
Closes the last open item on #1107: install-flow honesty for the Codex hooks bridge.

## What

`wmux setup-hooks` now installs the hooks bridge (script to the managed `~/.wmux/hooks/` location + a marker-bracketed `[[hooks.*]]` block in `~/.codex/config.toml`), but **writing is never reported as installing**. Codex requires the operator to trust a hook before it runs and stays silent when they haven't — so the flow is approve-then-verify:

1. **Write** — `setup-hooks` appends the block, bracketed by start/end marker comments so every later edit is bounded exactly. Refuses when:
   - `codex --version` is unprobeable or below the bisected **0.141.0** floor (fail closed — 0.140.0 parses the block, advertises `hooks stable true`, fires nothing)
   - any foreign `[[hooks.*]]` occupies the config (skip-if-foreign, the notify lane's rule one level wider)
   - the config is malformed, or a hand-pasted wmux block lacks its end marker (`manual` — never guess a boundary)
2. **Approve** — operator approves in Codex; the CLI print says `WRITTEN ≠ INSTALLED` and how to verify.
3. **Verify** — `--status` reports `WRITTEN but NOT trusted` vs `ACTIVE`. The verdict is evidence-based: the installer stamps the write time (`codex-hooks-install.json`), the bridge logs one JSON line per firing (`codex-hooks.log`), and only a log entry **newer than the stamp** reads as active. Idempotent re-runs never move the stamp; a path-changing refresh resets it (re-approval genuinely required, status says so). Manual installs (no stamp) degrade to any-firing-is-active.

## Also

- Refreshes preserve Codex trust annotations inside the region when the block is structurally current (no rewrite), and replace the region exactly when the bridge path or event set changed.
- `forceUnregister` removes the block marker-bounded beside the notify line.
- `copy-bridge.mjs` bundles the hooks bridge for packaged installs; its absence in older builds is non-fatal (`ok` stays true, own outcome field carries it).
- TS renderer + version gate are byte-identical mirrors of `integrations/codex/hooks/wmuxHooks.mjs` (src/ can't `require()` the ESM); a lockstep test holds the copies together.
- README Installation section rewritten around the new flow; manual flow kept (now with markers + the honest-floor semantics documented).

## What is deliberately NOT claimed

- The `PermissionRequest` regexes in `AgentDetector.ts` stay — they are the fallback for panes whose operator never approved the hook.
- No automatic approval: a programmatic installer cannot (and should not) pre-trust its own hook. The operator step is load-bearing.

## Tests

- `configIO`: block render validity, append/replace/idempotence, CRLF, region-boundary refusal (unterminated marker), foreign survival, removal, lockstep vs the `.mjs` (byte-identical render + version-gate agreement across the bisected table).
- `mcpRegistration`: all skip reasons, stamp lifecycle (not moved on idempotent re-run, reset on path change), annotation preservation, written-vs-active evidence window (pre-stamp firing does NOT count), manual floor, unregister.
- `lifecycleIntegrations`: version gate closed, null-when-unprobed, written verdict, ok unaffected by the new asset being source-missing.
- `codexHooksConfig`: end-marker bracket assertions.

tsc + eslint clean; shared/cli/integrations/main-mcp suites green (1481 + 251 passed).

## Summary by CodeRabbit

* **New Features**
  * `wmux setup-hooks` now installs the Codex hooks bridge alongside the notify bridge.
  * Installation status distinguishes hooks that are written but not yet trusted from active hooks.
  * Supports safe updates and removal of wmux-managed configuration while preserving user-defined hooks.
  * Requires Codex CLI 0.141.0 or newer and avoids modifying unsupported, malformed, or conflicting configurations.
* **Documentation**
  * Added guidance for approval, verification, status reporting, and manual hook setup.


openwong2kim:

Wires the Codex hooks bridge into `wmux setup-hooks` (approve-then-verify), completing #1107's last item. Touches config.toml handling (configIO, mcpRegistration, lifecycleIntegrations, CLI), with marker-bracketed regions, foreign-block refusal, and the 0.141.0 fail-closed floor covering the config-edit risk. P2 — closes a P2 track and the guards are well-tested.

openwong2kim:
Review pass complete: **no issues found.** The .d.mts signatures match the mjs exports; validate/bench/baseline all green.

openwong2kim:
Re-triggering CI: the synchronize event for 47cbe088 never started the workflows.

PR #1239, opened 2026-09-08
fix(term): repair stale frames on OS window refocus (#1234)

### LINKED ISSUE #1234: [Bug]: Rendering breaks/gets uglified when alt-tabbing
### What happened?

When alt-tabbing between windows, sometimes the rendering get uglified and is only restored to normal once you scroll (only through scrolling, even if you focus it and start typing, nothing happens).
I remember tagging a [bug](https://github.com/openwong2kim/wmux/issues/879) like this some time ago, I think this is a variation of it that was missed during initial fix.
I'm not sure if it happens only inside claude or reproducible in the regular terminal.
- <video src="https://github.com/user-attachments/assets/b2b66c63-d8f4-4203-aec1-c6a548f2f140" />

### Steps to reproduce

1. Launch wmux
2. Start a claude session
3. Alt tab to a different app (behavior isn't constant, sometimes you get the ugly render, sometimes you don't, but once you get it it can only be restored through scrolling that particular window).
4. See uglified output

### Expected behavior

1. Launch wmux
2. Start a claude session
3. Alt tab to a different app
4. See normal panes with proper text as expected

### wmux version

v3.50.1

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

Windows 10 19045.6466

Claude Code

```shell

```

### Confirmation

openwong2kim:
## What

Fixes #1234 (same corruption class as #879/#166/#318/#1230, on the alt-tab path).

**Why the existing triggers miss it.** Alt-tab away and back:
- the textarea never loses DOM focus in Electron → no `focus` repaint (clicking back into the pane also does not refire it — matches the report that focusing/typing does nothing);
- the pane never left the visible state → no `visible` repaint;
- an idle pane emits no output → no `burst`/`settle`.

The stale frame survived until the user scrolled — the one path that forces a full re-render.

**Fix.** New `window-focus` repaint reason:
- `glyphRepaint.onWindowFocus()` — immediate full-range repair **plus** the delayed verify the settle path already arms (#1230's lesson: a single refresh can itself race).
- Wired to DOM `window focus` **and** `visibilitychange → visible` (inert on Windows, #882, but the signal on macOS); a 250ms dedup window coalesces the pair into one repair.
- Deliberately **not** sharing the pane-focus throttle: a pane-focus repaint may have fired moments before the occlusion that caused the staleness.
- Gated on pane visibility — a hidden pane's repair is the `visible` repaint on reveal.

## Tests

5 new scheduler cases (33 total in the file): immediate+verify firing, signal-pair coalescing, non-interference with the focus throttle, verify-timer sharing with the settle path, dispose cancellation.

Reporter: it lands as draft since the corruption is driver/timing-dependent — if you can still repro on this build, the `[wmux:glyph-repaint]` console.debug line (Verbose) will say whether `window-focus` fired on refocus.

### DIFF
--- a/changelog.d/1239.md
+++ b/changelog.d/1239.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Terminal corruption after alt-tabbing away and back** (#1234): a stale/garbled frame could persist after refocusing the OS window — clicking into the pane or typing repaired nothing, only scrolling did. Refocusing the window now triggers an immediate full-pane repair (plus one delayed verification pass), without waiting for output or a workspace switch.
--- a/src/renderer/hooks/useTerminal.ts
+++ b/src/renderer/hooks/useTerminal.ts
@@ -1382,7 +1382,7 @@ export function useTerminal(containerRef: React.RefObject<HTMLDivElement | null>
     // Issue #166 — defensive full-range repaint for the "garbled glyphs until
     // resize" corruption (dirty-region desync). Strategy and trigger rationale
     // live in terminal/glyphRepaint.ts. Every reason (focus / visible / burst /
-    // settle-verify)
+    // settle-verify / window-focus)
     // does a plain full-range refresh; "focus" is throttled because it fires on
     // every keyboard pane-nav / MCP pane.focus via useActivePaneFocus's
     // term.focus(), not just mouse clicks, so the throttle is load-bearing. The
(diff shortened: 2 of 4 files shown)


openwong2kim:

Fixes #1234: adds a `window-focus` repaint reason to the glyph scheduler, so an alt-tab round-trip triggers an immediate full repair plus one delayed verify instead of waiting for a scroll. Change is confined to `glyphRepaint.ts`/`useTerminal.ts` with its own dedup window and correct teardown; well covered by 5 new scheduler tests. P3: cosmetic-but-annoying rendering glitch, intermittent and recoverable, matching the issue's priority.

openwong2kim:
Review pass complete: **no issues found.** Verified: the scheduler's dedup window, verify-timer sharing, dispose cleanup, and the hidden-pane visibility gate; the trigger analysis (textarea DOM focus never blurs on alt-tab, idle pane emits no burst) matches the reporter's 'only scrolling fixes it' symptom.

openwong2kim:
Closing this one. It adds another repaint trigger on OS window refocus, which repairs the stale frame after the fact rather than addressing why the frame goes stale — and the draft itself notes the corruption is driver/timing-dependent and unconfirmed.

Since this was opened, 3.52.0 shipped #1230 (post-settle verify repaint), #1258 (overlays bound to the live instance, scroll preserved across resync) and #1259 (real geometry re-asserted on recovery), which change exactly this recovery path. #1234 stays open; if the alt-tab corruption still reproduces on 3.52.0 we'll pick the fix back up from there with fresh evidence.

PR #1240, opened 2026-09-08
feat(pane): double-click a pane separator to even out the pair it separates (#1233)

### LINKED ISSUE #1233: [Feature]: When having 2 panes side by side, double clicking on the edge between them should split them 50/50 either vertically or horizontally
### Problem / motivation

Today when clicking on the edge between two side by side panes, instead of them aligning 50/50 against each other, one is getting pushed to the side.
I don't think that's intended behavior, since it just feels like it doesn't make sense and overall counter-intuitive.
I'm not sure if this is a me-only issue so just filing it as a feature request in case creator thinks this is a worthwhile endeavor
- <video src="https://github.com/user-attachments/assets/16a75e33-e23f-4970-a602-52de4d1d8f0b" />

### Proposed solution

Clicking on the edge between the panes (border) should make them split the width 50/50 either vertically or horizontally depending on what kind of edge (border) is being clicked on

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

openwong2kim:
Closes #1233.

## What

Double-clicking the divider between two panes now splits their combined space evenly between them — the two-pane case is exactly the asked-for 50/50, for either split direction. In a group of three or more panes, only the pair flanking the clicked divider is equalized, so the gesture stays local to the divider under the cursor.

## How

- `separatorEqualizePair` (pure, exported for tests) computes the new sizes; `PaneContainer`'s `Separator` handles `onDoubleClick`.
- The write goes through `updatePaneSizes` (the same path a drag uses), so it persists in the pane tree and the existing `paneSizes` sync effect drives the visual resize via `setLayout`.
- A pending drag-debounce write (200ms) is cancelled first — otherwise it would land after the equalize and snap the widths back to what the double-click just reset.

## Tests

5 unit cases for the arithmetic: 50/50 pair, local-pair-only in a 3-pane group, idempotence, out-of-range no-op, input immutability.


openwong2kim:

Implements #1233: double-clicking a pane separator evens out the two panes it flanks — exact 50/50 for a two-pane split, local to the clicked divider otherwise. The change is a pure `separatorEqualizePair` helper wired through the persistent `updatePaneSizes` path in `PaneContainer.tsx`, with a pending drag-write cancelled first; small, well-tested surface. P3: nice-to-have UX polish matching the issue's own priority.

openwong2kim:
Review pass complete: **no issues found.** The pending-drag debounce cancellation is correct, and a stale sizes read within the 200ms window is harmless — adjacent-pair drag conserves the pair sum, so equalization lands identically. Pure-arithmetic tests cover the branches.

Issue #1176, opened 2026-09-02
A blocked pane offers no seen/unseen distinction (the original "cannot dismiss the red dot" premise was wrong)


openwong2kim:
**Rewritten. The original framing of this issue was wrong, and acting on it would have reintroduced #1168.**

## Why the original premise was wrong

It said: focus clears every other attention status, so a pending question should be clearable too. That treats a pending question as an *event marker*, which it is not.

The attention statuses in `surfaceAgentStatus` (`complete`, `waiting`, `awaiting_input`) mark that **something happened while you were not looking**. Focusing is the acknowledgement, so `Pane.tsx` clearing them on focus is right — the event has been seen, and seeing it is the whole resolution.

`surfacePendingQuestion` is not that. It is derived from the transcript and means **this agent cannot proceed until a human answers**. It is a description of the agent's current state, not a record of a past event. Looking at the pane does not unblock the agent, so clearing the signal on focus would make the dot report "nothing to see" about a workspace that is still, in fact, blocked — the exact failure #1168 was filed for, just triggered by "the user looked and walked away" instead of "the stop said complete".

## The signal already clears on the only thing that resolves it

Verified in code; there is no stuck state to fix:

- `setSurfaceActivity` deletes the question the moment the agent emits a tool hook (`paneSlice.ts`) — "the agent is demonstrably working again, so any question it was blocked on has been answered".
- `markSurfaceRunning` does the same for agents with no tool hooks, on plain byte movement.
- Every stop rewrites the field: `buildTurnBoundaryMetadata` sends `pendingQuestion: ''` when the turn does not end on a question, and `setSurfacePendingQuestion('')` deletes the key.

So: answer the question and the agent moves → cleared. Answer it and the agent replies without tools and stops → cleared by the empty write. The red dot persists exactly as long as the blocked condition does. That is accurate, not sticky.

## What is actually left, if anything

One narrow thing, and it is an idea rather than a defect: there is no *seen / unseen* distinction on a blocked pane. Across many workspaces you cannot tell which blocked agents you have already triaged from ones you have not opened at all.

If that is worth solving, it belongs in the **visual treatment**, not the status: keep the dot red (the agent is still blocked) and drop only the animated glow once the pane has been focused. `agentStatusIcon.ts` already separates `dotVar` from `glowClass`, so the seam exists.

Not doing that speculatively — no one has reported wanting it. Leaving this open at P3 as the record of the analysis; close it if the idea is not wanted.

[...]

PR #1241, opened 2026-09-08
feat(roster): seen/unseen distinction for blocked panes (#1176)

openwong2kim:
Implements the narrow idea #1176 records (its own analysis ruled the rest a non-defect).

## What

Across many workspaces you cannot tell which blocked agents you have already triaged. Now:

- **Unseen blocked agent** — red dot + animated glow (as today).
- **Seen blocked agent** (you focused the pane since it asked) — red dot, **no glow**. The agent is still blocked; looking does not answer the question (#1168's lesson — the dot must not lie), so only the animation drops.

## Mechanics

- `surfaceQuestionSeen: Record<ptyId, questionText>` in paneSlice. `markSurfaceQuestionSeen` fires from `Pane.tsx`'s focus effect (keyboard nav included, same placement as the attention clear).
- Keyed by **text**: a re-asked question is unseen again; the marker is deleted on every question-clear path (tool activity, byte movement, empty stop write, pane close, surface close) — a reused ptyId inherits nothing.
- Roster: `row.questionSeen` in the projection (and in `rowsEqual`); the dot drops `glowClass` only when seen AND no attention event is live (attention statuses are unseen by definition — focus clears them).
- The workspace-level aggregate dot is untouched: still red while blocked, by design.

## Tests

- paneSlice lifecycle (3 new): marker records current text only, new text stays unseen, marker dies on all three clear paths.
- Selector (1 new, 4 assertions): seen/unseen/text-mismatch/no-question cases, and that `status`/`needsAttention` are unchanged by seen-ness.


openwong2kim:

Implements #1176: a blocked pane you have focused keeps its red dot but drops the roster glow, keyed by question text so a re-asked question glows again. Touches paneSlice, the roster selector, and Pane.tsx, with lifecycle tests on every clear path; status semantics are unchanged, so risk is low. P3 — visual refinement from our own analysis.

openwong2kim:
Review pass complete: **no issues found** (the dotRosterParity crash was caught and fixed pre-push during the original pass). Semantics check: focus marks the CURRENT question seen; a re-asked question (new text) is unseen again; stashed panes can never be seen — all correct per the issue's analysis.

PR #1242, opened 2026-09-08
fix(firstrun): sequence the first-boot overlays instead of stacking them (#1164)

### LINKED ISSUE #1164: First boot: auto-update dialog and onboarding wizard stack, and the overlay swallows all clicks
Observed twice during isolated-instance dogfood runs (fresh data directory, dev build): on first boot the "Automatic Updates" consent dialog and the "Welcome to wmux" onboarding wizard mount at the same time, overlapping; the full-screen overlay then swallows every click (keyboard shortcuts still pass). A user on a genuinely fresh install would see the same stack.

Expected: the two first-boot surfaces sequence (updates consent after — or inside — the wizard), and whichever is on top actually receives pointer events.

Repro: boot with a fresh `WMUX_DATA_SUFFIX` data dir and watch the first window.

openwong2kim:
Fixes #1164.

## Root cause

On a fresh data dir both startup IIFEs fire at once: `session.load() → null` arms the auto-update consent prompt (z-modal, 60) and `firstRun.check()` arms the wizard (z-dialog, 70). Both render as `fixed inset-0` siblings — the wizard's backdrop covers the prompt's buttons, so every click lands on a handler-less backdrop. Keyboard still works because the wizard owns focus and wires capture-phase Escape.

## Fix — sequencing, not stacking

One first-boot surface at a time (same principle as the D8 ruling that keeps the onboarding spotlight behind the wizard):

1. **Prompt waits for the wizard**: renders only when `showFirstRunWizard === null`. The consent stays pending and appears the instant the wizard completes or is dismissed (ESC included).
2. **Spotlight waits out the prompt**: the onboarding-start effect now skips while `showAutoUpdatePrompt` — its z-9999 full-screen backdrop would have covered the prompt's buttons, the identical bug one layer deeper.

First-boot order: **wizard → consent → spotlight**. Upgrade installs (marker exists, no wizard) are unchanged.

## Verification note

There is no AppLayout mount fixture today (the known TODO at the old line 1374); the change is two render/effect gates, tsc + eslint clean (AppLayout's 4 pre-existing lint errors are untouched), FirstRunWizard suites still pass (35).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved first-time setup sequencing so the onboarding wizard appears before the automatic updates consent prompt.
  * Ensured each first-boot prompt remains interactive and appears in the correct order.
  * Improved onboarding behavior for upgrade installations and users with multiple workspaces.

* **Tests**
  * Added coverage for onboarding, update consent, and first-boot prompt sequencing.


openwong2kim:

This PR fixes #1164 by sequencing the first-boot surfaces (wizard → update consent → spotlight) instead of stacking them. Both changes are narrow render/effect gates in AppLayout.tsx, so the risk is low; upgrade installs without the wizard are unchanged. P3 — small, well-verified fix for a fresh-install-only annoyance.

openwong2kim:
Review pass complete: **no issues found.** Both gates (prompt waits for the wizard; spotlight waits out the prompt) produce the intended first-boot order with no pointer-dead state; upgrade installs and reopen paths unaffected.

PR #1243, opened 2026-09-08
feat(remote): per-session agent metadata over the host API, remote rows in the roster (#1163)

### LINKED ISSUE #1163: Remote host API exposes no per-session agent metadata — roster cannot count remote Claude sessions (#1140 item 2 groundwork)
Split out of #1140 item 2 as the structural half, per the triage review of the remote-parity backlog.

`workspaceAgentRoster.ts` is keyed on local `ptyId`s (`surfaceAgent[ptyId]`, `surfaceAgentStatus[ptyId]`), and a remote-terminal surface has `ptyId: ''` by contract, so remote sessions are structurally invisible to the roster. The remote host's `/api/workspaces` payload carries only `sessionId/shell/cwd` — no agent name or lifecycle status — so there is no signal to build even a presence row from (a title-pattern heuristic was considered and rejected as false-positive-prone throwaway work).

Full parity needs both sides:
1. The host daemon exposes per-session agent metadata (name, status) in the `/api/workspaces` payload — it already has its own local roster to draw from.
2. `RemoteHostClient` parses it (with protocol-version tolerance for older hosts) and the roster selector admits remote sessions under a synthetic key (`remote:{hostId}:{sessionId}`), rendered with a host badge so remote rows are never mistaken for local agents.

#1140 stays open scoped to the UX; this issue tracks the protocol/selector work.

openwong2kim:
Closes #1163 (the structural half of #1140 item 2; the UX half stays there).

## What

**Host** — `GET /api/workspaces` panes gain `agentName` + `agentStatus` (additive-optional per the protocolVersion additive-field rule; older hosts simply omit both):
- name: creation-time role `displayName`, else the same daemon agent-state resolution the desktop's `daemon.getAgentName` uses (`readDaemonAgentState`) — so an agent that has exited, or a stale persisted `lastDetectedAgent` slug after a daemon restart, no longer produces a row;
- status: the same `bridge.getAgentStatus()` snapshot the desktop reconnect path trusts.

**Desktop** — `RemoteHostClient.normalizeWorkspaces` keeps a valid `agentName`, whitelist-checks `agentStatus` (`REMOTE_AGENT_STATUSES`) so a NEWER host's unknown status degrades to name-only, and drops a status without a name. `samePanes` compares agent fields so a remote status flip re-renders its row despite the 10s poll's identity stability.

**Roster** — `selectWorkspaceAgentRoster` admits a remote-terminal surface whose attached mirror reports agent metadata, under the synthetic `remote:{hostId}:{sessionId}` key (remote surfaces have `ptyId: ''` by contract, and the synthetic key can never collide with a local ptyId or A2A address). Remote agents count into `agentCount`/`needsAttentionCount`; the row shows the origin glyph (steel, matching the tab strip's remote-glyph rule) and `@hostLabel` instead of the local pane coordinate; clicking resolves the mirror surface. A host snapshot has no event channel → never `hasAttention`; status is the whole signal.

## Compatibility

Older desktop ↔ newer host: fields dropped. Newer desktop ↔ older host: fields absent, roster unchanged. No protocol version bump needed (additive rule).

## Tests

- Host: agent precedence matrix + absence (5 assertions over 2 fetch phases).
- Client: keep/drop matrix incl. unknown status and status-without-name.
- Selector: remote row shape, synthetic key, counting, detached host, no-metadata pane. 226 + 290 tests green in the touched suites.


openwong2kim:

This PR closes #1163 by surfacing per-session agent metadata over the host API and admitting remote agents into the workspace roster under a synthetic key. It touches a trust boundary (host data entering local unions), but the whitelist check and additive-optional fields keep the risk low. P3 — additive groundwork matching the tracked issue priority.

openwong2kim:
Adversarial review pass complete (fix-first): **4 findings, all fixed** — the roster's remote lookup only searched the FIRST attached workspace per host (multi-workspace-per-host silently missing rows), remote rows keyed by the colliding synthetic ptyId (duplicate React keys on multi-attach), unbounded agentName (now capped 256), and the slug fallback surfacing 'claude-code' instead of 'Claude Code' (spurious vendor column). Trust-boundary verification of the status whitelist came back clean — no foreign value can enter the local AgentStatus union.

PR #1244, opened 2026-09-08
feat(window): transparency + Windows 11 backdrop material (#1133)

### LINKED ISSUE #1133: [Feature]: Window transparency / acrylic (frosted-glass) background
### Problem / motivation

wmux is a native Windows terminal multiplexer, and on Windows 11 its window is fully opaque. An agent-heavy workflow means the wmux window is on screen most of the day, and a large opaque dark rectangle is visually heavy — especially next to Windows Terminal / Tabby / WezTerm, which all support translucency. There is currently no way to get a translucent or frosted-glass look in wmux.

### Proposed solution

Appearance options (e.g. Settings → Appearance):

1. **Background opacity slider (0–100%)** — apply the theme background color as `rgba()` at the renderer level, so the desktop shows through.
2. **Window material (Windows 11)** — expose Electron's `BrowserWindow#setBackgroundMaterial()` (`'mica'` / `'acrylic'`) behind a setting, combined with a transparent window background, for system-composited frosted glass at near-zero GPU cost. Acrylic is the classic "frosted terminal" look.
3. Optionally: custom background image + image opacity, as Windows Terminal and Tabby have.

### Alternatives considered

- Windows Terminal: per-profile `opacity` + acrylic material. Tabby and WezTerm (`window_opacity`) offer the same. wmux currently has neither, which is the one thing keeping my desktop terminals from matching.
- This can't be done externally — Chromium transparency/material must be set on the window itself.

### Additional context

Interesting detail: `%APPDATA%/wmux/settings.json` on my machine (3.49.1) contains a leftover `wmux-appearance-prefs` block with `terminalBgOpacity`, `windowTransparency`, and `windowMaterial: "mica"` — but no code in the repo reads these keys (searched current `main`). It looks like this feature was planned or existed in an experimental build at some point; if so, this request is just asking to revive it.

I searched existing issues and Discussions and this is not a duplicate.

openwong2kim:
Closes #1133 (items 1 and 2; item 3, background image, deliberately not attempted here).

## What

- **Window transparency slider** (30–100%, Settings → Appearance, Windows/macOS): the window background paints `--bg-base` at the chosen opacity over the desktop. Inner surfaces (panels, terminals) stay fully opaque — readability over blur is the tint model, matching Windows Terminal.
- **Backdrop material** (Windows 11 only): `mica` / `acrylic` via Electron's `setBackgroundMaterial`, system-composited behind the tint at near-zero GPU cost. `acrylic` is the classic frosted-terminal look.
- Defaults unchanged — opaque, exactly today's behaviour — so nothing shifts for users who never open the slider (DESIGN.md defaults untouched).

## How

- **Main owns the prefs** (`window-appearance.json`, the BrowserBackendStore pattern) because Chromium's `transparent` flag is **creation-only**: read synchronously before the first window, and all three creation sites (boot, updater rebuild, macOS activate) go through one `createMainWindow` wrapper that tracks whether the live window is translucent.
- **Honest restart gating**: opacity/material values apply live (IPC + change event; renderer tints via `data-window-translucent` + `--window-bg-opacity` with attribute-specificity CSS — no per-component edits), but crossing the translucent/opaque boundary needs a rebuild, and the Settings UI says "pending restart" when it detects that (main returns `active` alongside the prefs).
- **Platform gating**: Linux excluded (transparent without a compositor guarantee = black window); material row additionally requires Windows 11 (build ≥ 22000 via the preload's `windowsBuildNumber`).
- The leftover `wmux-appearance-prefs` keys the issue found in a settings.json never existed in this repo's history (`git log -S` empty) — likely from an experimental build; this PR supersedes them.

## Tests

- `WindowAppearanceStore.test.ts` (9): per-field corruption degradation, clamp/round, transparent-creation predicate, round-trip, normalize-on-write.
- Polish coverage lock satisfied (en + pl strings added).
- Settings suites (155) and tsc/eslint clean.

### DIFF
--- a/changelog.d/1244.md
+++ b/changelog.d/1244.md
@@ -1,0 +1,3 @@
+### Added
+
+- **Window transparency & Windows 11 backdrop material** (#1133): a new Appearance setting paints the window background at reduced opacity so the desktop (or the system-composited mica/acrylic backdrop, Windows 11 only) shows through — the frosted-terminal look, with panels and terminals staying fully readable. Opacity and material changes apply live; enabling or disabling transparency itself takes effect after a restart (the window is built with the flag at startup). Off by default.
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -34,9 +34,13 @@ process.on('uncaughtException', (err) => {
 import { markBoot, emitBootSummary } from './util/bootTrace';
 import * as crypto from 'crypto';
 import * as path from 'path';
+import * as os from 'os';
+import { parseWindowsBuildNumber } from '../shared/platform';
+import { WINDOWS_11_FIRST_BUILD } from '../shared/conptyWindows';
+import { normalizeWindowAppearance, type WindowAppearancePrefs } from '../shared/windowAppearance';
 import { app, BrowserWindow, dialog, ipcMain, powerMonitor } from 'electron';
 import { checkUserDataIsolation } from './dataIsolation';
-import { createWindow, loadMainRenderer } from './window/createWindow';
+import { createWindow, loadMainRenderer, windowWillBeTranslucent } from './window/createWindow';
 import { attachDesktopPresenceReporter, reportDesktopPresence } from './window/desktopPresence';
 import { PTYManager } from './pty/PTYManager';
 import { PTYBridge } from './pty/PTYBridge';
@@ -109,6 +113,7 @@ import { McpRegistrar } from './mcp/McpRegistrar';
 import { BrokerSupervisor, isMcpBrokerEnabled } from './mcp/BrokerSupervisor';
 import { WebviewCdpManager } from './browser-session/WebviewCdpManager';
 import { BrowserBackendStore } from './browser-session/BrowserBackendStore';
+import { WindowAppearanceStore } from './window/WindowAppearanceStore';
 import { ChromeLauncherRegistry } from './browser-session/ChromeLauncher';
 import { ChromeProfileStore } from './browser-session/ChromeProfileStore';
 import { ChromeSurfaceStore } from './browser-session/ChromeSurfaceStore';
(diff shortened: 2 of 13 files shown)


openwong2kim:
Adds window transparency and Windows 11 mica/acrylic backdrop material (Settings → Appearance), closing #1133 items 1–2. Main owns the prefs because the transparent flag is creation-only; all three window-creation sites route through one wrapper, defaults stay opaque, and Linux is excluded. P3 — appearance polish, off by default, with corruption and platform gating covered by tests.

openwong2kim:
Adversarial review pass complete (fix-first): **4 findings, all fixed** — the P1 first: WINDOW_APPEARANCE_CHANGED now carries `active` and the renderer gates the tint on it (enabling pre-restart used to paint translucent over the opaque native backgroundColor — wrong blend on every non-amber theme). Plus: persist debounced (was ~70 sync fs writes per slider drag on main), material row gated on 22H2/22621 (Electron's documented floor), and picking a material at opacity 100 nudges the tint to 90 so the backdrop is actually visible. All three createWindow sites verified routing through the wrapper; IPC normalization verified.

openwong2kim:
Closing this implementation, keeping #1133 open.

Review found the effect is close to invisible in practice: the tint applies to html/body/#root, but xterm (no `allowTransparency`, theme background), the titlebar (`bg-base`) and the sidebar (`mantle`) all paint opaque on top of it, so only the gaps between them would show the desktop — not the Windows Terminal-style see-through terminal #1133 asks for. Enabling it still carries Electron's documented transparent-window costs (no native shadow on macOS; on Windows no maximize via titlebar double-click and no edge resize).

Doing it properly means making the terminal background itself translucent and re-validating text contrast, which is a design-system change (DESIGN.md: no effects, `backgroundColor` = bgBase) that needs an explicit owner decision first. Smaller issues found along the way, for whoever picks this up: the persistence debounce timer is `unref()`'d while its comment says it guards against loss on exit (a quit within 150 ms rolls back the last value), the Windows 11 build threshold differs between main (22000) and the UI (22621), and the opacity predicate is duplicated in three places.

PR #1245, opened 2026-09-08
feat(wsl): pick which WSL distro the default terminal boots (#1103)

### LINKED ISSUE #1103: [Feature]: Select what exact WSL you have by default
### Problem / motivation

Right now if you select your default terminal as 'WSL', chances are (happens to me, not sure if there's some randomness involved in how wsl type is chosen) if you have Docker Desktop installed, this is the wsl that you get booted into by default.
This is slightly annoying when your main workspace lives inside an ubuntu wsl and wmux boots you into docker desktop's wsl, basically making 'Default terminal = wsl' feature useless.
Not a big deal, since you can just hop into wsl on every terminal load manually, but since wmux gives you the default terminal selection option, it'd be nice to have it be more agile.

### Proposed solution

When selecting default terminal as wsl, have a separate dropdown that specifies what exact wsl you want.
My personal preference would be for wmux to recognize that when working from wsl, you usually don't want docker desktop's one, but instead the one that you configure separately.

### Alternatives considered

This is a personal preference and I might be wrong/oblivious on how other people use that, so safe option here is to just give a separate dropdown or just have initial dropdown be like 'WSL - Docker Desktop' , 'WSL - Ubuntu', 'WSL - distro bla bla bla'

### Additional context

_No response_

### Confirmation

openwong2kim:
Closes #1103.

## What

When the default terminal is **WSL**, a new **WSL distribution** picker (Settings → Terminal) lists every distro (`wsl --list --quiet`) with **Docker-owned distros sorted last**, plus a **System default** option. Chosen once, every WSL pane — local and daemon mode, splits, respawns — boots into it via `wsl.exe -d <name>`.

## Trust-path design

The distro travels as **exactly `['-d', '<name>']`**, validated at every boundary:

- **Name charset** (`WSL_DISTRO_NAME_RE`): anchored, no flags/spaces/quotes — the value becomes a spawn argument, so it must not be able to smuggle argv.
- **Main IPC mirror**: the renderer owns the persisted choice (session.json) and pushes it to main on change + boot (`settings:default-wsl-distro`) — injection happens at the single place `pty.create` resolves the effective shell; zero per-create-site threading.
- **Daemon RPC**: `DaemonCreateSessionParams.args` is validated with `isWslDistroSpawnArgs` (exact arity, shape, and wsl cmd; anything else refused) — an external RPC caller cannot widen it into arbitrary argv.
- **Spawn order**: the flag goes **first**, before hook/exec args, because `wsl.exe` parses its own options before anything that follows.

Enumeration is total-to-`[]` (off Windows, missing wsl.exe, timeout, garbage) so the picker simply hides; the parser tolerates the UTF-16LE output older WSL installs emit.

## Tests

14 unit cases (charset, args builder, RPC-boundary shape, UTF-8/UTF-16 parsing, docker-last ordering); pl coverage lock satisfied; Settings + loadSession suites green (227); tsc (root + daemon) clean — the 3 eslint `no-var-requires` in pty.handler are pre-existing on main (unchanged lines).


openwong2kim:
Adds a WSL distribution picker (Settings → Terminal) so WSL panes boot the chosen distro instead of the system default, closing #1103. It touches both spawn paths (local PTY and daemon RPC), with the distro name charset-validated and the exact -d argument shape enforced at the boundary. P3 — a self-contained settings enhancement with the spawn-argument risk properly guarded.

openwong2kim:
Adversarial review pass complete (fix-first): **6 findings, all fixed** — the P1: daemon recovery/restart/promote replays DROPPED the distro args (a recovered WSL pane silently booted the system default — the exact bug this PR fixes); the selection now persists on DaemonSession.args and rides all four replay paths. Also: local-mode hook injection skipped when a distro selection is present (`--rcfile` stacked behind `-d` is not a wsl.exe flag), local-branch fallback-shell parity, Unicode distro names (`wsl --import '우분투'`) via a widened-but-injection-safe charset + BOM-sniffed buffer decoding, stale-choice reconciliation, and an unrelated TODOS.md sweep removed. Spawn-argv injection surface verified solid — no bypass found at any boundary.

PR #1246, opened 2026-09-08
feat(sidebar): list and reclaim orphaned daemon sessions (#1101)

### LINKED ISSUE #1101: [Feature]: Track active agent session even after pane is closed
### Problem / motivation

Today was debugging why my WSL was eating so much memory and happened to bump into the following thing: since when wmux crashes or you close the pane it doesn't kill the agent session, just your terminal view of it I would say, you might run into a situation when you have multiple active sessions open even tho you already forgot about them (closed the pane).
It's nice of wmux to not kill them once wmux pane/whole app are closed, since it doesn't interrupt the workflow, and I appreciate the design, but it would be even better if you could get some kind of view onto the sessions that were orphaned.

### Proposed solution

It would be great if there was some kind of way to know what hanging sessions you have from inside the wmux, and maybe reconnect to them by clicking a button or smth, or maybe kill them by clicking on trash can button next to them, don't really know how to implement it so it's congruent with app's design & style.

### Alternatives considered

_No response_

### Additional context

_No response_

### Confirmation

openwong2kim:
Closes #1101.

## What

A new sidebar section (under the workspace list, above the remote section) lists **daemon sessions that are still running but owned by no pane** — the ones that survive closing a pane/app by design, which until now were visible only as the tray's background-session count.

- **Click a row** → the session comes back as a new pane bound to its id (same shape as a restored surface, so the terminal's reconnect path attaches rather than spawning a duplicate), activated in its origin workspace when that still exists, the active one otherwise.
- **✕ on a row** → kills the session (the existing `pty.dispose` → `daemon.destroySession`).
- Renders nothing while empty; rows are monochrome (alive-but-unnecessary is not the amber alive grammar), with cwd and an age trailer.

## How

- Orphans = live daemon sessions minus `getWorkspacePtyIds` across all workspaces — visible tree **and stash**, so a stashed pane never reads as orphaned; dead tombstones and this app's attached sessions are excluded.
- `pty.list` now also carries each session's spawn-time workspace stamp and the daemon-derived agent name (same precedence as `/api/sessions`).
- Slow 30s poll + refresh after adopt/kill: orphans appear via quit/crash/dispose races — none emit renderer-visible events.

## Tests

7 slice cases: the set difference (incl. stash totality), agent-name precedence, adopt binding + activation + list removal + origin-workspace preference, dispose round-trip, unknown-id safety. pl coverage lock satisfied; Sidebar suites (140) green.


openwong2kim:
Adds a sidebar section listing daemon sessions still running after their pane closed, with click-to-adopt and ✕-to-kill. Touches `pty.list` (additive `workspaceId`/`agentName`) plus a new renderer slice and component; 7 slice tests cover the set-difference. P3 — additive visibility feature for #1101, low risk: renders nothing when empty.

openwong2kim:
Adversarial review pass complete (fix-first): **7 findings, all fixed** — the two P1s: the floating terminal's session is now counted as owned (its row was listed as an orphan whose ✕ killed it under the pane), and adopt re-checks ownership at CLICK time (boot race / 30s-poll staleness must not double-bind one pty to two surfaces). Hardening: the 20-pane cap, the #977 ordinal rule, born-hidden un-zoom, pane.created/focused publishes + saveSessionNow, and a two-step inline kill confirm. The adopt-into-reconnect path itself verified correct (born-with-ptyId skips self-create; reconcile's 2-strike cannot clear a live id).

PR #1247, opened 2026-09-08
feat(remote): rename + color-tag aliases for attached remote workspaces (#1086)

### LINKED ISSUE #1086: Remote workspaces: one click doesn't return to a local workspace, and remote rows lack rename/color/startup parity with local ones

openwong2kim:
Closes #1086's remaining scope (Bug 2, the local-alias half — startup/duplicate parity stays out per the issue's own #1091 spectate-only direction; Bug 1 was already fixed on main).

## What

Remote workspace rows get the two parity verbs that are **local by design**:

- **Rename** — context-menu item + double-click, same interaction as local rows. The alias lives on this desktop's attachment descriptor; the remote host still owns the real name, so a host-side rename updates the snapshot **behind** the alias without clobbering it, and clearing the alias reveals the snapshot again.
- **Color tag** — the same palette and dot grammar as `WorkspaceItem`, stored the same way. A tagged row paints its tag color in the status-dot slot; untagged rows are unchanged.

## Persistence

Descriptor round-trip: slice action → `attachmentsAdd` upsert (same-key overwrite, the store's own rule) → disk (main store now passes the optional fields through, validated) → boot restore. Descriptors built inside the immer producer (drafts must not escape `set()` — the trap stashPane's comments document).

## Tests

3 new slice cases: alias set/clear + host-rename non-interference, color set/clear, unknown-key no-ops. Restore-path typing covered by tsc; lifecycle + Sidebar suites green.


openwong2kim:

Adds rename and color-tag aliases for attached remote workspaces (#1086), kept locally on the attachment descriptor while the host owns the real name. Touches the sidebar row, the remoteWorkspaces slice, and descriptor persistence — additive optional fields, no protocol change, slice tests included. P3: a cosmetic parity enhancement with low risk.

openwong2kim:
Adversarial review pass complete (fix-first): **1 finding, fixed + regression-tested** — attachRemoteWorkspace replaces the entry wholesale on a same-key re-attach with a fresh snapshot carrying no label/color, silently wiping the user's rename/tag. Aliases now survive re-attach (and an incoming label still wins). The immer-draft-escape hazard in the alias persistence was verified handled (descriptors built inside the producer).

PR #1248, opened 2026-09-08
feat(i18n): blame-based locale drift advisory (#1037)

### LINKED ISSUE #1037: Five pre-existing Polish strings translate English that has since changed, and the drift class needs history-aware tooling
## What

Follow-up from #1029. While applying the review fix there (`update.readyToInstall` translating the pre-#908 English), I audited the rest of the pre-existing Polish keys for the same defect class — **the English source changed after the translation was written** — by comparing per-key `git blame` timestamps between `en.ts` and `pl.ts`.

13 candidates, 7 false positives (line moved/reformatted, translation still accurate), 1 fixed in #1029. **5 remain stale:**

| key | en.ts today | pl.ts today (stale meaning) |
|---|---|---|
| `settings.fontFamilyDesc` | Any installed font — type a name or pick one | "Monospace font for the terminal" |
| `fanout.title` | Multi Task — N parallel isolated tasks | "Fan-out — 1 prompt → N isolated tasks" |
| `fanout.promptLabel` | Shared prompt (optional — common context for every task) | "Prompt" |
| `fanout.promptPlaceholder` | Common context every task receives… (optional when each task has its own prompt) | missing the optional-clause |
| `fanout.titlesLabel` | Per-task title & prompt (editable — branch/slug identity) | missing "& prompt" |

The fan-out four are one event — the feature became Multi Task and its shared prompt became optional (#1008 era), and the Polish still describes the old dialog. All five predate #1029's 1075-key pass; they are in the original 374.

@p-poppe offered on #1029 to take these; the wording is theirs to make as the speaker. This issue exists so the list doesn't vanish into a PR comment.

## The structural part

This drift class is invisible to the coverage test #1029 added, by construction: the key exists, the placeholders match, only the *meaning* is stale. Structure can't see it; only history can.

Proposal: a CI **advisory** (not a gate) that runs the blame comparison — flag keys whose `en.ts` blame is newer than the locale's — and posts the list as a PR comment when a locale file is touched. Advisory because the method has a real false-positive rate (7 of 13 here): reformatting or moving an English line trips it without any meaning change. A human still judges; the tool just stops the class from being invisible.

Scope note: the same audit presumably applies to `ko` (71%) and `zh` (98%), the other two maintained locales. The 20 stalled locales are a known, accepted gap (#997) and not worth auditing until someone owns them.

openwong2kim:
Closes #1037.

## What

The five stale Polish strings the issue catalogued are **already retranslated on main** (verified — `fanout.title`, `settings.fontFamilyDesc` & co all carry current wording). What was still missing is the structural half, so the class stops being invisible:

- `scripts/locale-drift-report.mjs` — compares per-key `git blame` timestamps between `en.ts` and each maintained locale (pl/ko/zh; the 20 stalled locales are #997's accepted gap) and lists keys whose English changed after the translation was written.
- **Advisory by design, never a gate**: blame dates false-positive on reformats/moves (7 of 13 candidates in the original audit), so the CI step (`validate` job) prints the report and exits 0 regardless. A human judges; the tool restores visibility.

Current report on main: 17 candidates across pl/ko/zh — spot-checks show a mix of real candidates (the `settings.prefix.*` trio) and known false-positive shapes, exactly the advisory profile the issue asked for.

## Usage

```
node scripts/locale-drift-report.mjs            # pl ko zh
node scripts/locale-drift-report.mjs pl --check # CI mode (exit 0)
```


openwong2kim:

Adds a blame-based locale drift advisory (#1037): a script comparing en.ts vs locale blame dates, wired into CI as a print-only step that always exits 0. Advisory by design, no gating and no product code touched, so CI risk is nil. P3: useful tooling, nothing urgent.

openwong2kim:
Review pass complete: **no issues found.** Advisory-only by design (exit 0 in --check mode unless a locale file is missing); blame parsing walks full line numbering consistently with lineOfKey.

PR #1249, opened 2026-09-08
fix(ime): scan Claude Code v2.1.246+'s rule-and-chevron input chrome (#1035)

### LINKED ISSUE #1035: The IME anchor's Claude input-line scan looks for chrome Claude Code no longer draws, so src=marker never fires

openwong2kim:
Fixes #1035.

## What

`scanClaudeInputLine` now recognizes the v2.1.246+ input chrome in addition to the old rounded box:

```
────────────────────────────
› 정확히어떻게해야하는지알려주
```

- A `› ` row counts only with a **full-width rule (4+ `─`) directly above** — the same structural guard the box top played: transcript output may quote a bare chevron, but it does not place a rule directly above the bottom-most one.
- Column math: `› x` → indent + 2 (vs the box's indent + 4).
- Wrapped input has no structural continuation marker in the new chrome (no side borders), so `rowSpan` stays the conservative 1 — the preedit gate treats the typed row as caret territory, same as the box's first interior row.
- Both shapes scan bottom-up in one pass; whichever is bottom-most wins when both appear (quoted chrome in the transcript above a live input of either kind).

## Tests

6 new cases: the exact pane clip from the report (Korean input), indent math, no-rule rejection, bottom-most-wins, mixed old/new chrome in both directions, short-dash-run rejection. 95 total in the file.


openwong2kim:

Teaches the IME anchor's input-line scan (#1035) Claude Code v2.1.246+'s rule-and-chevron chrome, guarded by the full-width-rule-above test so quoted transcript chevrons don't match. One scan branch plus regexes in imeAnchor.ts, old box shape intact, 6 new tests. P3: contained fix restoring an inert path, no regression risk.

openwong2kim:
Review pass complete: **no issues found.** The rule-above guard mirrors the box-top guard; old-box scanning unchanged; mixed-chrome precedence covered by tests. The regexes are anchored and tolerate indentation/trailing whitespace.

PR #1251, opened 2026-09-08
feat(workspace): archive — Active → Archived → Permanently Deleted (#1011)

### LINKED ISSUE #1011: [Feature]: Archive / Recently Closed Workspaces to keep the sidebar manageable
## Problem / motivation

I use wmux with multiple workspaces for different projects and tasks.

Over time, the workspace sidebar naturally becomes crowded. Some workspaces are used intensively for a few days or weeks and then are no longer needed on a daily basis.

To keep the sidebar usable, I periodically close/remove old workspaces.

The problem is that this currently feels like a destructive action:

- the workspace disappears from the sidebar
- there is no visible "Recently Closed" list
- there is no archive
- there is no simple way to restore it later
- if I realize weeks later that I still need the workspace configuration, I have to recreate it manually

This creates an unfortunate choice:

**Keep everything → overloaded sidebar**

or

**Clean up the sidebar → lose easy access to previously configured workspaces**

For users who use wmux as a long-running project/agent environment, I think a third state would be useful:

**Active → Archived → Permanently Deleted**

## Proposed solution: Workspace Archive

Add an archive for workspaces.

Instead of only closing/removing a workspace, there could be separate actions such as:

```text
Archive
Delete permanently
```

An archived workspace would disappear from the normal sidebar but remain available for later restoration.

The archived section could remain collapsed by default, for example:

```text
WORKSPACES

▸ Current Project
▸ Research
▸ Development
▸ Testing

──────────────

Archived Workspaces (12)
```

Alternatively, archived workspaces could be accessible only through the command palette:

```text
Ctrl+K / Command Palette
→ Open Archived Workspace
```

This keeps the normal sidebar focused on active work while still preserving older workspace configurations.

## Restore workflow

An archive view could show entries such as:

```text
Archived Workspaces

Oxaliplatin Review
Archived 12 days ago
D:\Projects\oxaliplatin

HIPEC Dashboard
Archived 31 days ago
D:\Projects\hipec
```

with actions:

```text
Restore
Delete permanently
```

Restoring would put the workspace back into the normal sidebar.

## Important distinction: archived workspaces should not keep PTYs alive

I would not expect an archived workspace to keep all terminal processes running indefinitely. That would defeat the purpose and could consume considerable resources.

Instead, archiving could save a lightweight workspace snapshot and then terminate the live PTYs.

Possible snapshot contents:

- workspace name
- workspace color
- pane layout / splits
- surface configuration
- working directories
- workspace profile
- startup commands
- relevant resume/session metadata if available

On restore, wmux could create fresh PTYs from that configuration.

Conceptually:

```text
ACTIVE WORKSPACE
    │
    │ Archive
    ▼
ARCHIVED SNAPSHOT
    │
    │ Restore
    ▼
ACTIVE WORKSPACE
with freshly created PTYs
```

## Suggested implementation

A possible persisted structure could be something like:

```ts
interface ArchivedWorkspace {
  id: string;
  archivedAt: string;
  workspace: WorkspaceSnapshot;
}
```

and session state:

```ts
interface SessionData {
  workspaces: Workspace[];
  archivedWorkspaces?: ArchivedWorkspace[];
}
```

When archiving:

1. Snapshot the workspace configuration
2. Add the snapshot to `archivedWorkspaces`
3. Perform the normal PTY/resource cleanup
4. Remove the workspace from active workspaces
5. Save the session

When restoring:

1. Load the archived snapshot
2. Generate fresh pane/surface/PTY identifiers where necessary
3. Recreate the workspace
4. Remove the snapshot from `archivedWorkspaces`
5. Activate the restored workspace

Some of this may be reusable from the existing workspace duplication / fresh pane-tree logic.

## MVP / simpler version

A smaller first implementation could simply be **Recently Closed Workspaces**.

Keep the last 20–50 removed workspaces and expose them through:

```text
Command Palette

Workspace: Reopen Closed Workspace
```

For example:

```text
1. Oxaliplatin Review
2. HIPEC Dashboard
3. Test Environment
4. Old Project
```

This alone would already solve most of the problem. It could later evolve into a full archive.

## Retention options

Possible approaches:

- keep archived workspaces indefinitely until explicit permanent deletion
- configurable retention such as 30 / 90 days / forever
- fixed maximum such as the most recent 50 or 100 archived workspaces

Personally, explicit archive + permanent delete would be the least surprising behavior.

## Optional improvement: temporary Undo

Immediately after removing/archiving a workspace:

```text
Workspace archived                         Undo
```

The Undo action could remain available for a few seconds. This would cover accidental clicks, while the archive solves the long-term problem.

## Why this would improve wmux

wmux encourages users to create multiple workspaces for different projects, agents, experiments and tasks. That naturally means the workspace list grows over time.

An archive would let users treat the sidebar as a list of **currently relevant workspaces**, rather than as permanent storage for every workspace ever created.

It would provide both:

- a cleaner sidebar
- confidence that cleaning up the sidebar does not permanently discard useful workspace configuration

I think this becomes increasingly valuable as users adopt wmux for long-running multi-project workflows.

### Confirmation

- [x] I searched existing issues and did not find an existing workspace archive / recently closed workspaces feature request.

openwong2kim:
Closes #1011.

## What

The sidebar's missing third state:

- **Archive** (workspace context menu, beside Duplicate) — the same teardown as Close (sessions die; promotion, multiview pruning, a2a force-fail and recovery cleanup all run through the existing `removeWorkspace` — one teardown, no drift), but the **configuration survives**: name, color tag, profile, pane arrangement, archivedAt.
- **Archived section** (below the workspace list, collapsed by default, invisible when empty) — one row per snapshot with its color dot and an age trailer. Click = **Restore**; hover ✕ = **Delete permanently**.
- **Restore** rebuilds a live workspace with the same name/color/profile/layout and **fresh identities everywhere**: new workspace id, new pane ids (the snapshot is a LayoutNode — no ids), and an ordinal taken from the true free high-water. A restored workspace can never collide with a live `w<N>` auto-name or A2A address.

## Persistence

`session.json` gains `archivedWorkspaces` (shape-guarded on hydration — malformed entries drop, never crash the load).

## Tests

5 slice cases: snapshot content + teardown/promotion, last-workspace refusal, restore identity-freshness (ws id, every pane id, ordinal vs a live one), permanent delete, unknown-id no-ops. Sidebar/i18n suites green; pl coverage lock satisfied.

## Summary by CodeRabbit

- **New Features**
  - Workspaces can now be archived instead of closed, preserving their name, color, profile, and pane arrangement.
  - Archived workspaces appear in a collapsible sidebar section with timestamps.
  - Restore an archived workspace with one click, or permanently delete its saved snapshot with confirmation.
  - Archived workspaces persist across sessions and the last remaining workspace cannot be archived.

- **Bug Fixes**
  - Invalid archived workspace data is safely ignored during session loading.


openwong2kim (commented):
Implements workspace archive (closes #1011): snapshot on archive, collapsed Archived section, one-click restore with fresh ids. Touches the sidebar, workspaceSlice and session.json persistence; low risk — teardown reuses removeWorkspace and the restore path is identity-fresh and tested. P3 — a self-contained feature PR with no behavior change for non-archive flows.

openwong2kim:
Adversarial review pass complete (fix-first): **6 findings, all fixed** — the two P1s: the four archive i18n keys were never added (every locale rendered RAW keys — the batch edit that carried them aborted on an anchor mismatch and CI stayed green because nothing asserts these keys), and restore now calls clearRemoteSelection (the every-assignment-site contract). Hardening: tree-shape + archivedAt validation on hydration, profile re-normalization on restore (the only profile-entry path that skipped the policy), snapshot push conditioned on confirmed removal, dead wsOrdinal field dropped.

PR #1252, opened 2026-09-08
chore(ci): vitest 4.1.11 + Node 24 validate — the #984 experiment

### LINKED ISSUE #984: Node 24 on windows-latest kills one vitest fork per run — runtime move blocked, npm-11 lineage decoupled
Found while moving CI to Node 24 for the npm-11 lockfile lineage (#983). The lineage move landed decoupled from the runtime (Node 22 + `npm install -g npm@11`), because the Node 24 runtime itself does this on Windows:

## Symptom

On `windows-latest` under Node 24, one vitest worker fork dies silently and the run fails with:

```
Error: [vitest-pool]: Worker forks emitted error.
Caused by: Error: Worker exited unexpectedly
```

Every test that runs passes. The dead fork's queued file is left unreported — `779 passed | 3 skipped (783)`, one file unaccounted. Extracting the ✓-reported set from the log and diffing against the suite list identifies it: both baseline attempts on the same commit were missing exactly `src/daemon/transcript/__tests__/TranscriptProjector.test.ts` (pure-JS imports, passes everywhere else), which says the fork died mid-run and vitest only discovered the corpse when it tried to hand it the next file.

## Evidence matrix (all on the same npm-11 lock, vitest 4.1.0, forks pool)

| runtime | job | result |
| --- | --- | --- |
| Node 22 (many runs, `main`) | baseline win + validate | green |
| 24.19.0 | Baseline (windows-latest) | **fail ×2** (same missing file both attempts) |
| 24.19.0 | validate (windows) | pass ×1 |
| 24.18.1 | Baseline (windows-latest) | **fail ×1** |
| 24.18.1 | validate (windows) | **fail ×1** (`780 passed | 2 skipped`) |
| 24.x | Baseline macOS / ubuntu | green |
| 24.x | release rehearsal (3-OS build), bench, tsc | green |

So: high-frequency but not deterministic (validate passed once on 24.19.0), Windows-only, test-suite-only — build paths are unaffected. Pinning 24.18.1 was run as an experiment against the open 24.19.0 memory-regression reports (nodejs/help#5165, nodejs/node#60482) and **did not help**, which weakens the simple-OOM attribution.

## Nearest ecosystem matches

- vitest-dev/vitest#8861 — "v4: inconsistent errors with pool: forks/vmForks", Windows 11 + Node 24.10. Closed by #8967 (Windows Defender fix), which predates 4.1.0 — already in our tree, and our symptom (silent fork death, not syntax errors) differs.
- vitest-dev/vitest#9762 — same error string, Linux/bun, needs-reproduction.

Neither is a confirmed match. vitest 4.1.0 → 4.1.11 has eleven untried patches; a bump is the cheapest next experiment when someone picks this up.

## What unblocking looks like

Re-attempt the runtime move (this repo wants CI on the same major the runner forces actions onto, #964) when any of these is true:

1. vitest patch (try 4.1.11 first) makes Windows + Node 24 green for 3 consecutive baseline runs, or
2. a Node 24.x release lands with a child-process/Windows fix that names this class, or
3. the fork death is root-caused here (instrument with `--maxWorkers=1` / threads pool on a scratch branch to bisect worker count vs pool type — #8861's reporters saw both help).

Until then `node-version: 22` + the npm-11 install step is the supported CI configuration; the lockfile lineage does not depend on the runtime major.

openwong2kim:
Experiment for #984, per its own triage order ("the vitest bump to 4.1.11 is the cheapest first experiment").

## What

- `vitest` 4.1.0 → **4.1.11** (the windows-latest one-dead-fork-per-run failure was reproduced under 4.1.0 on Node 24.18.1/24.19.0).
- The **validate job's runtime moves to Node 24** — the exact runtime move #984 tracks — so CI itself becomes the experiment.

## Reading the result

- **Green across several runs** (windows-latest is the only platform that matters; the fork death was Windows-only) → the blocker is lifted; follow-up flips the remaining workflows to Node 24 and closes #984.
- **A file still left unreported / a fork still dies** → the `--maxWorkers=1` / threads-pool bisection is the next step, and this branch's runs are the evidence.

Local verification on darwin/arm (Node 24.15.0): full runtime suite green under 4.1.11 (one timing-band flake in `human-typing` under load, passes twice in isolation — same behavior on 4.1.0).

Kept: the npm-11 lineage guard step (the lockfile lineage decoupling from #983 stays regardless of the runtime major).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,7 +65,13 @@ jobs:
         run: npm run build:daemon
 
       - name: Test
-        run: npm test
+        # #984 bisection step 2: vitest 4.1.11 did NOT fix the fork death
+        # (run 34180314730: "Worker forks emitted error — Worker exited
+        # unexpectedly", all 986 executed files green, one fork dead) — so the
+        # main suite runs on the THREADS pool here instead of forks. If this is
+        # green across runs on windows-latest, the failure is fork-pool
+        # specific on Node 24 and the runtime move can proceed pool=threads.
+        run: npx vitest run --exclude "**/*.runtime.test.ts" --exclude "**/*.runtime.test.tsx" --exclude "**/*.runtime.test.mjs" --pool=threads && npm run test:runtime
         env:
           # Tells those suites the bundle was promised, so removing the step
           # above fails a test instead of silently retiring both files.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,13 +15,12 @@ jobs:
 
       - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
         with:
-          # Node 22 runtime, npm 11 package manager — deliberately split. The
-          # lockfile lineage needs npm 11 (npm 10's `npm ci` rejects an
-          # npm-11-written lock, which is how all five Dependabot PRs arrived
-          # dead); the Node 24 runtime is NOT needed for that and kills one
-          # vitest fork per run on windows-latest (#984). Recouple when #984
-          # is resolved.
-          node-version: 22
+          # #984 experiment: Node 24 runtime + vitest 4.1.11 — the windows
+          # one-dead-fork-per-run failure was reproduced under vitest 4.1.0 on
+          # 24.18.1/24.19.0; 4.1.11 is the upstream line that could carry the
+          # fix. Green runs here unblock the runtime move; a still-dead fork
+          # means the --maxWorkers=1 / threads-pool bisection is next.
+          node-version: 24
           cache: npm
 
       # npm 11 owns package-lock.json (see CONTRIBUTING). Node 22 bundles npm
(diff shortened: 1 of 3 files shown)


openwong2kim (commented):
Runs the #984 experiment: vitest 4.1.0 → 4.1.11 and the validate job moves to Node 24, so CI itself is the test bed. Touches only ci.yml and dev dependencies, so product risk is nil; the open question is whether the windows-latest fork death disappears. P3 — outcome is observational; green runs unblock the runtime move, red means the bisection is next.

openwong2kim:
Closing: this was the #984 bisection experiment (CI-as-evidence vehicle — vitest 4.1.11 probe, then the temporary `--pool=threads` switch), not a fix. The conclusion it gathered (4.1.11 does not fix the fork death; threads pool is the viable path on Node 24) is recorded here for the follow-up that will do the real runtime move. Not merging an experiment branch.

PR #1257, opened 2026-09-09
fix(updater): honour autoUpdateEnabled from process start (#1250)

### LINKED ISSUE #1250: [Bug]: autoUpdateEnabled: false is ignored — the updater quits the app to install, fails silently, and the app relaunches into the same loop (4x in 3 minutes)
### What happened?

With auto-update turned off (`"autoUpdateEnabled": false` in %APPDATA%\wmux\session.json), wmux 3.46.0 still downloaded 3.51.0 and ran the install handoff four times in about three minutes. Each handoff quit the app; none of them installed anything; the app came back up 5-17 seconds later and did it again.

Three distinct problems:

1. **The setting is not honoured.** I turned auto-update off deliberately because I keep long-lived agent panes open and want to choose when the app restarts. The updater downloaded the installer and called the install handoff anyway — and that handoff quits the app.

2. **The failures are completely silent.** No `install-aborted` marker in %APPDATA%\wmux. No `previous install was refused: ...` line on any subsequent boot (this is where it differs from #1084). Nothing in %LOCALAPPDATA%\SquirrelTemp\Squirrel-Install.log — the newest entry there was from 2026-08-13, three weeks earlier, so Setup.exe was never launched at all. The only evidence was indirect: four `wmux-install-waiter-*` directories and a 146 MB installer in %TEMP%, and a version number that never changed.

3. **It is a loop, not a one-off.** Because nothing records the failure, the relaunched app sees the same pending update and starts over. Four cycles here. Two of them logged `install already in progress — ignoring re-entrant call`, so some in-progress state also survives across the failed handoff (possibly the same class as #1043).

The daemon is not the blocker in this case. It acked shutdown in 35 / 34 / 44 / 50 ms on the four attempts — this is not #980's "daemon never dies", and it is far faster than #1084's 634 ms.

And the install itself is fine: ~21 minutes after the last attempt I ran the very same installer file by hand, on the same machine, having killed nothing and changed nothing. It completed in 9 seconds with a clean `Finished Squirrel Updater`, kept all 4 workspaces and 24 ptys, and produced zero rate-limit errors. By then the app and daemon had already exited on their own and the install root was fully unlocked (I probed every .exe/.dll/.node/.asar under it with FileShare.None — 0 locked). That is what makes me think the handoff gives up while the tree is still coming down, rather than the tree failing to come down.

### Steps to reproduce

1. Turn auto-update off. Confirm %APPDATA%\wmux\session.json contains "autoUpdateEnabled": false.
2. Be on a version behind the latest GitHub release (I was on 3.46.0; 3.51.0 was out).
3. Launch wmux normally (Start Menu stub) with a few workspaces to restore — I had 4 workspaces / 24 ptys.
4. Watch %APPDATA%\wmux\logs\main-<date>.log and %TEMP%.

Observed on this machine 4/4 times, unattended, within 3 minutes of a cold boot. I was not clicking anything — the first attempt fired 32 seconds after the app started on its own.

### Expected behavior

1. With autoUpdateEnabled: false, the updater should not download an installer and should never call the install handoff. Checking and surfacing "an update is available" would be fine; quitting the app is not.

2. If a handoff runs and Setup.exe does not start, that must be recorded — the install-aborted marker written, a "previous install was refused" line on the next boot, and ideally a toast. As it stands a failed handoff is indistinguishable from no attempt at all.

3. A recorded failure should also stop the relaunched app from immediately retrying, and should clear the in-progress state so the next attempt is not swallowed by "install already in progress".

### wmux version

3.46.0 (the version that attempted the updates). Now 3.51.0 after installing manually — autoUpdateEnabled is still false there, so I expect this to recur on the next release.

### How did you install wmux?

winget

PowerShell 7+ (pwsh)

### Windows version

Windows 10 Education 22H2 (build 19045.6456)

Claude Code

```shell
Timeline — %APPDATA%\wmux\logs\main-2026-09-08.log. Times are UTC (local = UTC+9). Machine booted at 00:05.

  00:09:04  app start   v3.46.0 pid 2980   (started on its own after boot)
  00:09:36  handoff #1  waiter 1dFraX   daemon.shutdown ack 35ms   -> quit
  00:09:53  app start   v3.46.0 pid 3240   (17s after the quit)
  00:10:04  handoff #2  waiter XvoRm2   daemon.shutdown ack 34ms   -> quit
  00:10:09  app start   v3.46.0 pid 6568   (5s after the quit)
  00:10:33  handoff #3  waiter MkerBT   daemon.shutdown ack 44ms   -> quit
  00:12:17  app start   v3.46.0 pid 17988  (104s after the quit)
  00:12:32  handoff #4  waiter hDeisp   daemon.shutdown ack 50ms   -> quit
  ...       nothing for 21 minutes; no app, no daemon
  00:33:14  I ran the same Setup.exe by hand
  00:33:23  Squirrel: "Finished Squirrel Updater"
  00:33:21  app start   v3.51.0

The 5s and 17s gaps are not me clicking anything.

Shape of every one of the four handoffs (paths shortened):

[AutoUpdater] Session save triggered before update install
[AutoUpdater] install handoff: waiter=%TEMP%\wmux-install-waiter-1dFraX\wait-and-install.ps1 watching 4 process(es) under %LOCALAPPDATA%\wmux; force-killed 0/0 foreign, 4 of our own left to app.quit(), daemon pid <n> left to the graceful shutdown
[AutoUpdater] quitting for install — the daemon goes down with us so the installer runs against a dead tree
[Main] Full shutdown — racing daemon.shutdown (8000ms budget)
[main] full-shutdown: racing daemon.shutdown
[Main] daemon.shutdown ack received (elapsed=35ms)

Nothing follows the ack line in any of the four.

Attempts 2 and 4 were preceded by:
[AutoUpdater] install already in progress — ignoring re-entrant call.

Attempts 2-4 logged:
[AutoUpdater] reusing verified installer already in temp (wmux-update-3.51.0-<pid>-wmux-3.51.0.Setup.exe) — skipping download

%LOCALAPPDATA%\SquirrelTemp\Squirrel-Install.log — no entry near any of the four attempts. Timestamps jump from 2026-08-13 straight to my manual install:

[08-09-26 09:33:14] info: Program: About to install to: %LOCALAPPDATA%\wmux
[08-09-26 09:33:15] info: CheckForUpdateImpl: First run, starting from scratch
[08-09-26 09:33:19] info: ApplyReleasesImpl: Writing files to app directory: ...\app-3.51.0
[08-09-26 09:33:23] info: Program: Finished Squirrel Updater

State I measured ~20 minutes after the last attempt, before touching anything:
  wmux.exe processes .................. 0
  daemon processes .................... 0
  locked files under the install root . 0   (FileShare.None probe on every .exe/.dll/.node/.asar)
  %TEMP% .............................. 4x wmux-install-waiter-*, 1x 146 MB wmux-update-3.51.0-*.Setup.exe
  %APPDATA%\wmux\update-install-aborted.txt ... absent

Setting, in both the live file and the pre-install snapshot (session.json.tmp.*):
  "autoUpdateEnabled": false

Minor side observation: the last attempt left session.json.tmp.17988.12 behind (pid 17988 = the 4th app), i.e. the atomic session save did not finish its rename on that quit. It was harmless here — it is actually the only surviving record of my pre-update workspace list — but it suggests the quit path can cut the save short.

Related: #980 and #1084 are the same "install never happens" family, but both left a refusal trace; this one leaves none, and the daemon shuts down promptly. #1043 covers the swallowed re-entrant handoff.
```

### Confirmation

openwong2kim:
## Root cause (#1250, claim 1: "the setting is not honoured")

The `AUTO_UPDATE_ENABLED` listener was registered inside `start()`, which runs at the **end** of the ready sequence — the daemon bootstrap can hold it for tens of seconds on a cold boot (measured 39s in dogfood, per the #866 comment in the same file). The renderer sends the persisted toggle during `loadSession` ~1s in, via fire-and-forget `ipcRenderer.send` — so on essentially every cold boot the send arrived before any listener existed and was **silently dropped**. `enabled` stayed at its default `true` for the whole session: background checks, downloads, and the update-ready toast all ran against an explicit `autoUpdateEnabled: false`. That matches the reporter's 4/4 reproducibility.

Same bug class as #866, one channel over — and the same fix pattern (constructor registration) was already established in this file for the two #866 handlers.

## Changes

- **Constructor registration** of the `AUTO_UPDATE_ENABLED` listener (`AutoUpdater.ts`) so the renderer's boot-time send lands.
- **Boot read from session.json** (`SessionManager.readAutoUpdateEnabled`, injected as an `AutoUpdaterHooks` reader so the updater stays free of the session module — same injection rationale as `getDaemonPid`). Applied in `start()` only when the renderer has not delivered the toggle: the renderer's value is fresher than disk and always wins. Covers boots where the IPC never arrives at all (window still loading, or the renderer's session load failed on a locked file and fell back without dispatching `loadSession`).
- **Re-check `enabled` at `downloadUpdate()` entry** — `check()`'s gate runs when the poll *starts*; the toggle can land while `fetchUpdate` is awaited, and the 150 MB download must still be stopped. `oneShotInstall` stays exempt, matching `check()`'s gate: the toggle must never brick the manual update path.
- `performInstall` is deliberately **not** gated: every path into it is user-initiated (Restart button / one-shot press); background polls only ever download.

## Scope note on the issue's claims 2 & 3

The silent-failure and relaunch-loop halves of #1250 were observed on v3.46.0 and are the failure family already hardened by #1057/#1058/#1059/#1085/#1136 (all shipped by v3.48.1–v3.51.0): the waiter is now provably started before quitting, refusals are recorded and surfaced, and the marker feeds the boot notice. The one piece this PR does not explain is the *unattended* install handoff in the report — both v3.46.0 and current code require a user press (toast/Settings) to cross from download to install, so that part is worth re-verifying on 3.51.0 with the reporter. A follow-up comment on the issue asks for that.

## Tests

- `AutoUpdater.platform.test.ts`: listener registered at construction before `start()` (the exact live shape of the bug); `start()` seeds from the persisted hook when the renderer never sends; renderer-delivered value wins over a stale disk value.
- `AutoUpdater.download.test.ts`: a toggle landing after the check started still stops the background download; a manual press remains exempt.

Full suite: 988 files / 15,135 tests green. `tsc --noEmit` clean. (The 4 pre-existing `no-var-requires` eslint errors in these files are untouched.)

Fixes #1250 (claim 1; claims 2–3 already mitigated on production — see scope note).

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-update preferences are now restored when the application starts, even if the renderer is unavailable.
  * Disabling automatic updates during a background check now prevents pending downloads.
  * Manual, one-time update installations continue to work when automatic updates are disabled.
  * Early toggle changes are now reliably applied before update checks begin.

* **Documentation**
  * Added a changelog entry describing the auto-update behavior improvements.

### DIFF
--- a/changelog.d/1257.md
+++ b/changelog.d/1257.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **The auto-update toggle is honoured from process start.** Turning auto-update off used to be silently ignored on cold boots: the setting reached the updater before its listener existed, so background checks, downloads, and the "update ready" prompt kept appearing. The toggle now lands reliably, is also seeded from the saved session when the window is still loading, and a download that has not started yet is stopped if you flip the toggle mid-check. (#1257)
--- a/src/main/index.ts
+++ b/src/main/index.ts
@@ -445,6 +445,12 @@ const autoUpdater = new AutoUpdater(() => mainWindow, {
   // before-quit takes the daemon.shutdown branch instead of detaching.
   onInstallRequiresFullShutdown: () => { fullShutdownRequested = true; },
   getDaemonPid: () => readDaemonPid(getWmuxDir()),
+  // #1250: seed the updater's enabled flag from session.json. The renderer
+  // sends the toggle during loadSession, but that send races this module's
+  // own ready sequence; a locked session.json can stop it from ever arriving.
+  // The disk read fills the silence; the renderer's IPC still wins when it
+  // lands (AutoUpdater.start applies the read only if nothing arrived).
+  readAutoUpdateEnabled: () => sessionManager.readAutoUpdateEnabled(),
 });
 
 // #1046: a Squirrel install can half-complete (an AV lock race inside the
--- a/src/main/session/SessionManager.ts
+++ b/src/main/session/SessionManager.ts
@@ -265,6 +265,26 @@ export class SessionManager {
     }
   }
 
+  /**
+   * #1250: the persisted auto-update toggle, as a targeted read.
+   *
+   * The updater needs the value before the renderer's session load runs, and
+   * load() is the wrong tool: it pulls the whole workspace payload through
+   * the migration machinery and logs a second load line, all for one boolean.
+   * Returns null when nothing is persisted (first launch, or a locked/corrupt
+   * file — the caller keeps its default rather than failing the boot).
+   */
+  readAutoUpdateEnabled(): boolean | null {
+    try {
+      const loaded = atomicReadJSONSync<SessionData>(this.filePath, {
+        validate: SessionManager.isSessionData,
+      });
+      return loaded?.autoUpdateEnabled ?? null;
+    } catch {
+      return null;
+    }
+  }
+
   load(): SessionData | null {
     // v2 RCA fix (adversarial review): distinguish "no session file" (true
     // first launch → null) from "file exists but unreadable" (transient AV/
--- a/src/main/updater/AutoUpdater.ts
+++ b/src/main/updater/AutoUpdater.ts
@@ -197,6 +197,20 @@ export interface AutoUpdaterHooks {
    * install.
    */
   getDaemonPid: () => number | null;
+  /**
+   * The persisted auto-update toggle, read straight from session.json (#1250).
+   *
+   * Injected rather than imported for the same reason as getDaemonPid above:
+   * pulling the session module into the updater drags its migration machinery
+   * into the electron-mocked unit tests. main owns the SessionManager, so it
+   * hands over a reader instead.
+   *
+   * Used once, in start(), to initialize `enabled` from what the user last
+   * chose — covering the boots where the renderer never delivers the toggle
+   * (window not yet loaded, or its session load failed on a locked file).
+   * A null/undefined means "nothing persisted" and leaves the default.
+   */
+  readAutoUpdateEnabled?: () => boolean | null;
 }
 
 export class AutoUpdater {
@@ -205,6 +219,11 @@ export class AutoUpdater {
   private hooks: AutoUpdaterHooks;
   private isChecking = false;
   private enabled = true;
+  // #1250: whether the renderer's AUTO_UPDATE_ENABLED send has been received.
+  // start() initializes `enabled` from session.json ONLY while this is false —
+  // once the renderer has spoken, its value is fresher than anything on disk
+  // and must not be clobbered by the boot read.
+  private enabledFromRenderer = false;
   private pendingUpdate: UpdateInfo | null = null;
   private downloadedPath: string | null = null;
   // The digest `downloadedPath` was accepted under. Kept so the bytes can be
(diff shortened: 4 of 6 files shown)


openwong2kim:

This fixes the #1250 core defect: the `AUTO_UPDATE_ENABLED` listener moves to the constructor and `start()` seeds the toggle from session.json, so the renderer's boot-time send is no longer dropped. Changes are confined to AutoUpdater/SessionManager, low risk, with the manual one-shot update path deliberately exempt. P2 — the bug silently overrode an explicit user setting on essentially every cold boot.

PR #1258, opened 2026-09-09
fix(terminal): bind overlays to the live instance, preserve scroll across resync (#1256)

### LINKED ISSUE #1256: Scroll-to-bottom button is an intermittent no-op; scroll position is yanked to bottom on pane resync
**Environment:** wmux 3.51.0, Windows 10, Electron renderer.

**Symptoms:**
1. The scroll-to-bottom button often does nothing.
2. After a hidden pane is shown/reconnected, manual scroll position is thrown back to the bottom.

**Root cause (from static analysis of the unpacked app.asar):**
- The button component (`yB`, i18n key `terminal.scrollToBottom`) is rendered by the terminal wrapper (`dx`). `dx` passes `H.current` — a **snapshot** of a ref value — instead of the ref itself. The ref is populated by mutation without a re-render, so on early renders the button receives `null` (no button, no `onScroll`/`onWrite`/`onResize` subscriptions) and only comes alive if `dx` re-renders for some other reason. After a resync/reconnect of a hidden pane, a **stale** terminal instance is retained, so `onClick → scrollToBottom()` and the visibility subscription target a detached terminal → silent no-op.
- "Blocks manual scroll" is not a separate blocker: on show, the resync does `reset()` + a full buffer rewrite, which forces the viewport to the bottom. Resyncing while the user is scrolled up rips the position down.
- (The button hiding while at the bottom is by design, `viewportY < baseY`; combined with the stale instance it feels permanently stuck.)

**Suggested fix:** Pass the **ref** to `yB`/`hB` rather than `H.current`, and after a resync restore the previous scroll position instead of `reset()`-ing to the bottom.

openwong2kim:
## Root cause

Two defects, one root pattern — render-time snapshots of a ref that is populated by mutation:

**1. The scroll-to-bottom button intermittently no-ops.** `ScrollToBottomButton` (and `BookmarkIndicator` — same pattern, found during verification) received `terminalRef.current` at render time. `useTerminal` populates that ref inside its mount effect — both fresh creation and parked-terminal adoption — with no re-render to follow. So the snapshot the overlays hold is `null` after mount until some unrelated state change re-renders, and stale after an adoption swaps the instance. Null/stale means: the button's `onScroll`/`onWriteParsed`/`onResize` subscriptions target a detached terminal (visibility never updates), and clicks call `scrollToBottom()` on a dead buffer — the silent no-op from the report.

**2. Resync yanks the viewport to the bottom.** Both resync repaint paths (`paintDeadSnapshot` for dead sessions, `completeResyncFromFlush` for the live snapshot/reflow ladder) do `reset()` + full buffer rewrite, and `reset()` snaps the viewport to the bottom. Showing a hidden pane while scrolled up ripped the position down.

## Changes

- `useTerminal` publishes the live instance as **state** (`terminalInstance`), set at both assignment sites (creation/adoption, dispose-on-cleanup). Identity changes now re-render; `Terminal.tsx` binds `terminalInstance` for `ScrollToBottomButton`, `BookmarkIndicator`, and the vi-copy gate. Prop contracts of the two components are unchanged.
- Both repaint paths capture **distance-from-bottom** before the reset and restore it after the recovered screen parses — via a trailing empty `term.write('', cb)` as a parse barrier (xterm invokes write callbacks in order; same shape as the existing `hydrateForRead` barrier). Rows-based, the same convention `terminalPark.restoreParkedViewport` uses.

## Tests

- `useTerminal.hiddenRetention.test.ts`: new source-contract tests pinning (a) capture-before-reset → parse-barrier → scrollToLine ordering in both repaint paths, (b) the instance state publication at both assignment sites and the hook return.
- New `Terminal.liveBinding.test.ts`: Terminal.tsx binds `terminalInstance` (never a `terminalRef.current` snapshot) for both overlays and the vi-copy gate.
- Existing repaint-ordering contracts kept (window widened — ordering unchanged).

Full suite green: 990 files / 15,135 tests. `tsc --noEmit` clean; no new lint findings (the 3 `no-empty-function` errors in useTerminal.ts are pre-existing on main).

Fixes #1256.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed the scroll-to-bottom button so it consistently responds when the terminal is ready.
  - Fixed bookmark indicators and copy-mode availability when terminals are refreshed or replaced.
  - Preserved the current scroll position when a hidden or disconnected pane is recovered, preventing unexpected jumps to the bottom.

### DIFF
--- a/changelog.d/1258.md
+++ b/changelog.d/1258.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **The scroll-to-bottom button and a resynced pane's scroll position.** The floating scroll button could silently do nothing, and revealing a hidden pane could throw you to the bottom of its output. Both came from one pattern: overlay components captured the terminal at render time, before the terminal actually existed — and the pane-recovery repaint reset the viewport. The button (and the bookmark markers) now bind the live terminal instance, and a recovered pane puts you back where you were scrolled to. (#1258)
--- a/src/renderer/components/Terminal/Terminal.tsx
+++ b/src/renderer/components/Terminal/Terminal.tsx
@@ -239,9 +239,12 @@ export default function TerminalComponent({ ptyId: externalPtyId, shell, cwd, on
   // `isActive` for the stacked/tab case (one tab visible at a time).
   const shown = visible ?? isActive;
   const isVisible = isWorkspaceVisible && shown;
-  const { terminal: terminalRef, findNext, findPrevious, clearSearch } = useTerminal(containerRef, { ptyId, isVisible, scrollbackFile, onFirstData: scrollbackFile ? handleFirstData : undefined, onContextMenu: handleContextMenu });
+  const { terminal: terminalRef, terminalInstance, findNext, findPrevious, clearSearch } = useTerminal(containerRef, { ptyId, isVisible, scrollbackFile, onFirstData: scrollbackFile ? handleFirstData : undefined, onContextMenu: handleContextMenu });
 
-  const showViCopyMode = viCopyModeActive && isActive && terminalRef.current !== null;
+  // terminalInstance (state, #1256) — not terminalRef.current (a render-time
+  // snapshot): the ref is populated after this render ran, so a snapshot read
+  // here sees null until an unrelated re-render happens.
+  const showViCopyMode = viCopyModeActive && isActive && terminalInstance !== null;
   const showSearchBar = searchBarVisible && isActive;
 
   const handleCloseSearch = () => {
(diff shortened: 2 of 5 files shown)


openwong2kim:

This fixes #1256: the terminal instance is now published as state, so the scroll-to-bottom button and bookmark indicator bind the real instance instead of a null/stale ref snapshot, and both resync repaints restore the pre-reset scroll position. Changes stay confined to useTerminal/Terminal with contract tests. P3, matching the issue — a daily UX annoyance, not a correctness risk.

PR #1259, opened 2026-09-09
fix(terminal): floor tiny fits, re-assert real geometry on recovery (#1255)

### LINKED ISSUE #1255: Terminal panes render ~1 column wide (cols clamped to 10) after resize/split/restore and never recover
**Environment:** wmux 3.51.0, Windows 10, Electron renderer.

**Symptom:** After resizing the window, splitting/closing panes, or restoring a saved session, a terminal pane renders extremely narrow — roughly one word per line. Fullscreening the pane or resizing the window does **not** fix an already-broken pane; only a fresh pane/session renders correctly.

**Root cause (from static analysis of the unpacked app.asar):**
- The daemon enforces a hard floor `MIN_SAFE_COLS = 10` (`clampCols = Math.max(10, cols)` in `resizeSession`, `daemon-bundle/index.js`). Any tiny cols value gets clamped to 10 → ~1 word/line.
- Source of the tiny value: the FitAddon wrappers in the renderer (`ResizeObserver` handler and the `refit` path, `index-B7_D3s94.js`) gate only on `offsetWidth === 0 || offsetHeight === 0`. There is no sane minimum width and no wait for layout to settle. During split/resize, or before `react-resizable-panels` settles on session restore, a "small-but-nonzero" measurement yields cols 2–9 → clamped to 10.
- Why it never self-heals: two dedup caches are never reconciled. The `refit` path sends a `pty.resize` without updating the renderer's cached cols and without a readback of the daemon's actual size. After that, correct fits (cols equal to the cached value) are silently dropped, and the daemon stays at `cols = 10`. Reconnect/resync does not recompute size from the DOM.

**Suggested fix:** In the fit path, (1) enforce a reasonable minimum width, (2) wait for layout to settle before fitting, and (3) read back the daemon's actual cols and re-assert if they diverge from the DOM-derived value.

**Repro:** Split a pane (or resize the window) while content is present, or restore a multi-pane session; observe a pane collapse to ~1 column and stay there.

openwong2kim:
## Root cause

The report's static analysis checked out at every step. A pane collapsing to ~1 column after resize/split/restore and never recovering needs two failures, both present:

**1. Sub-floor fits were applied.** The renderer's fit sites gated only on *zero* dimensions (`offsetWidth === 0`). A transient small-but-nonzero container (mid-split, session restore before `react-resizable-panels` settles) was fitted for real: the tiny columns were applied to the xterm buffer, and the reflow re-wrapped the entire scrollback at that width — damage a later correct fit does not undo. This is why a broken pane stayed broken even after fullscreening.

**2. Nothing reconciled the two sides of the pipe.** The daemon clamps its PTY to `MIN_SAFE_COLS(10)` (`DaemonSessionManager.ts`) while the renderer's `lastSentCols` dedup believed the tiny value it sent was the live geometry. Correct fits equal to the cached value were dropped, and reconnect/resync never recomputed size from the DOM — so the daemon could sit at the clamp indefinitely.

## Changes

- **`shared/terminalGeometry.ts`** (new): the floor constants move to a shared module — the daemon imports them (single source; value and semantics unchanged, the zle.so SIGBUS guard intact), and the renderer never proposes what the daemon would have to clamp.
- **Every `fit()` apply site is floor-gated** (exported `fit`, initial mount, `fonts.ready`, `runFit`, font/theme effect) via `proposedDimensions()` + `isSafeGeometry()`: sub-floor proposals are *skipped, not applied*. Skipping is self-healing — the ResizeObserver re-fires when the layout settles. Floor skips record no selection-debt (`claimFit` mechanism stays reserved for deferrals); an adoption on a sub-floor container holds its parked viewport exactly like a hidden one.
- **Recovery re-asserts geometry**: both the resync settle (`completeResyncFromFlush`) and the daemon reattach now push DOM-derived dimensions through `sendResize` — which carries no dedup — so a session recovered at the clamp, or left there by a transient tiny fit, is handed the real size.

## Deliberately out of scope

A daemon-side readback/echo of actual geometry (belt-and-braces reconciliation in the other direction) — noted in the issue as a possible follow-up; the two fixes above cover both known divergence paths.

## Tests

- `shared/__tests__/terminalGeometry.test.ts`: the floor semantics as a shared contract.
- `useTerminal.deferredFit.test.ts` (#1255 block): source-contract pins — every fit apply site gated (ordering vs the zero guard / `claimFit`), both re-assert points present.
- Daemon geometry tests unchanged and green (behavioral, not source-matched).

Full suite green: 15,140 tests. `tsc --noEmit` clean for both the app and daemon projects; no new lint findings (the 3 `no-empty-function` errors in useTerminal.ts predate this branch).

Fixes #1255.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed terminal panes collapsing into a single column after resizing, splitting, or restoring sessions.
  - Prevented unsafe intermediate dimensions from reflowing terminal content.
  - Improved pane recovery so terminal and session sizes remain synchronized after layout changes.

- **Tests**
  - Added regression coverage for terminal resizing, font and theme updates, session recovery, and minimum safe dimensions.

### DIFF
--- a/changelog.d/1259.md
+++ b/changelog.d/1259.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- **Panes no longer collapse to a single column after resizing, splitting, or restoring a session.** A transient mid-layout measurement used to be applied to the terminal for real, re-wrapping all its scrollback at a broken width — damage a later resize could not undo — while the background session was independently clamped to a minimum width, leaving the two permanently out of step. Sub-floor measurements are now skipped until the layout settles, and every pane recovery re-asserts the real size. (#1259)
--- a/src/daemon/DaemonSessionManager.ts
+++ b/src/daemon/DaemonSessionManager.ts
@@ -6,6 +6,7 @@ import os from 'node:os';
 import fs from 'node:fs';
 import path from 'node:path';
 import type { DaemonSession, DaemonSessionState, DaemonSessionSupervision, DaemonConfig } from './types';
+import { MIN_SAFE_COLS, MIN_SAFE_ROWS } from '../shared/terminalGeometry';
 import { RingBuffer } from './RingBuffer';
 import { DaemonPTYBridge } from './DaemonPTYBridge';
 import { PromptEventLog } from './PromptEventLog';
@@ -119,10 +120,9 @@ const DEFERRED_UNMUTE_DELAY_MS = 100;
  * observed 6/7 boundary, which may shift with prompt width or locale. The
  * renderer's xterm view can briefly be narrower than the PTY during a layout
  * transition — harmless compared to a dead shell, and the next settled resize
- * reconciles them.
+ * reconciles them. (#1255: the renderer now skips sub-floor fits entirely —
+ * shared constant, see shared/terminalGeometry.ts.)
  */
-const MIN_SAFE_COLS = 10;
-const MIN_SAFE_ROWS = 2;
 const clampCols = (cols: number): number => Math.max(MIN_SAFE_COLS, cols);
 const clampRows = (rows: number): number => Math.max(MIN_SAFE_ROWS, rows);
 
--- a/src/renderer/hooks/__tests__/useTerminal.deferredFit.test.ts
+++ b/src/renderer/hooks/__tests__/useTerminal.deferredFit.test.ts
@@ -91,3 +91,71 @@ describe('#747 — a deferred fit must be recorded and settled', () => {
     expect(block).not.toMatch(/fitAddon\.fit\(\)/);
   });
 });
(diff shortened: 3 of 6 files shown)


openwong2kim:

This fixes #1255: every fit() apply site is now floor-gated so a transient mid-layout measurement is skipped instead of applied, the floor constants are single-sourced in a new shared/terminalGeometry module, and both recovery paths re-assert real geometry past the dedup cache. Risk is moderate — it touches every resize path in useTerminal — but the gate is skip-and-retry and well covered by contract tests. P3, matching the issue.

PR #1263, opened 2026-09-09
fix(wsl): restore directories and exact Claude sessions

thomas-pedersen:
## Summary

Preserve WSL project directories and exact Claude conversation IDs across app/daemon restarts. Resolve the directory in the selected distro, keep the Linux target separate from the Windows PTY directory, and use per-pane Claude hooks through the existing authenticated Windows bridge.

This revision addresses the requested changes in [the maintainer review](https://github.com/openwong2kim/wmux/pull/1263#pullrequestreview-5187019581). WSL probes are asynchronous, and unsuccessful recovery retains the original pane, binding and scrollback with an error and a Retry control.

## Changes

- Honor #1245's distro picker for new panes; save its resolved distro/user and normalize saved distro arguments to that target. Existing panes keep their target when the default changes.
- Replace the synchronous WSL probe with asynchronous `execFile` (60-second budget, matching caller timeout). Coalesce concurrent probes for the same target/directory; do not cache completed directory validation, since directories and default users can change.
- Publish pending WSL panes without waiting for their distributions. Recover concurrently in the background under the startup cap, including exec units without a GUI. Attach and explicit Retry share an in-flight promotion. Cap-skipped WSL panes also retain their identity for on-demand recovery.
- Keep failed recovery metadata and buffer files across subsequent restarts. Explicit close removes pending recovery; failed recovery does not expire via suspended TTL. Cancel pending creates during close/shutdown so a late probe cannot spawn an orphan.
- Restore exact resume bindings during promotion and persist them before removing the old buffer. Validate the Linux cwd before any exec resume command runs; missing directories keep the pane retryable instead of resuming another project.
- Preserve native local `shellArgs`; reject double-quoted WSL paths before spawning; add bootstrap cwd/WSLENV diagnostics; avoid interactive startup files for exec units.
- Ignore renderer-provided distro/user arguments for fresh panes. Dead-pane replacement carries a lookup ID; main reads the saved target from daemon-owned metadata.
- Add the Retry UI through the existing reconnect path, including visibility and geometry restoration. Retain the maintainer's merged performance baseline and Ctrl+G changes.

Usage and limitations: [WSL session recovery](https://github.com/thomas-pedersen/wmux/blob/fix/wsl-session-recovery/docs/how-to/wsl-session-recovery.md).

### Fixed

- **WSL panes keep their project and Claude conversation after restart.** Linux working directories are validated inside the selected WSL distribution and restored after Bash startup files. Per-pane Claude hooks capture the exact session ID for Resume, including when several panes share a project. The distro picker is honored for new panes; recovery retains the pane's resolved distro and Linux user. (#1263)
- Cold WSL starts no longer block the daemon. Failed recovery keeps the pane and scrollback with an error and a Retry control, including across further restarts.

Recorded in `changelog.d/1263.md`; no release version bump.

## Stability tier impact

- [x] Change to an `internal` surface (no public contract impact).

Optional WSL/recovery metadata is confined to internal daemon RPC/session persistence and renderer IPC. Existing distro `args` retain #1245's validated shape. Existing attach/promote RPCs perform recovery; no new public RPC method.

n/a — no PaneMetadata, EventBus, permission-enforcement or named-pipe security contract changes. The existing authenticated bridge is reused.

## Test plan

- Type checking, daemon build and focused async-probe, cancellation, persistence, reconnect and keyboard tests pass.
- Windows focused regression suite: **8 suites / 78 passed**, 4 platform skips.
- Real Windows + Ubuntu-24.04 runtime regression passes. Two simulated Claude sessions in the same quoted/Unicode directory execute the real hooks/bridge and retain separate IDs through detach/reattach and two daemon restarts. The test then removes the project directory, verifies failed retries retain IDs and nonempty buffer files, restarts again while the directory is missing, restores it and retries the original conversations successfully. Double-quoted paths are rejected explicitly.
- Additional tests cover event-loop progress while probes are pending, concurrent probe coalescing, retry after failure, cancellation on close, pending metadata retention beyond TTL, native shell argument preservation, exec output without `.bashrc` banners and missing-cwd diagnostics.
- Final merged Linux `npm test`: **1,012 unit suites / 15,420 tests passed**, plus **197 runtime tests passed**, with platform/opt-in skips retained. Runtime execution required the built daemon bundle (`WMUX_REQUIRE_DAEMON_BUNDLE=1`).
- **All CI gates pass on `64cf042`**: [Windows validation](https://github.com/openwong2kim/wmux/actions/runs/34823632694), [Windows/macOS/Linux baselines](https://github.com/openwong2kim/wmux/actions/runs/34823632740), and [performance plus artifact contract](https://github.com/openwong2kim/wmux/actions/runs/34823632671). Fresh-runner performance confirmation was not needed. CodeRabbit still skips automatic review because this PR is a draft.

The WSL runtime fixture uses an isolated daemon/profile and a simulated CLI; it makes no model requests. Run it with `WMUX_TEST_WSL=1` and optionally `WMUX_TEST_WSL_DISTRO` as documented.

Scope: Bash and Windows interop are required. Exec commands use the non-interactive Linux PATH (or an explicit PATH in the command). This does not add Codex hooks, zsh/fish integration or Linux transcript access in Windows chat view. Existing panes without captured IDs need a one-time manual `claude --resume` selection in a fresh pane. Draft pending a manual check with real Claude conversations.

## Related issues / PRs

- Integrates #1245 (distro picker), #1290 (test timing fixes), and the maintainer's merge of #1291 (performance baseline) and #1286 (Ctrl+G).
- #1265 (external terminal hyperlinks) remains a separate, already-merged fix.

## Reviewer checklist

- [x] CHANGELOG entry and fragment included.
- [x] Stability impact classified.
- [x] Substrate contract docs: n/a; no listed public surface changed.
- [x] Regression tests cover the requested behavior and failure paths.
- [x] No secrets, tokens, `.env`, or binaries committed.

## Summary by CodeRabbit

- **New Features**
  - WSL panes now restore their distribution, user, working directory, project, and Claude conversation after restarts.
  - Claude sessions can resume automatically, including after detach and reattach.
  - Panes can be restored on demand when startup recovery limits are exceeded.

- **Bug Fixes**
  - WSL recovery no longer blocks daemon startup.
  - Failed recovery preserves the pane and scrollback, displays an error, and provides a Retry option.
  - Recovery and supervised restarts now complete reliably.

- **Documentation**
  - Added guidance for WSL directory handling, Claude session recovery, requirements, and troubleshooting.


openwong2kim:

Validates WSL startup directories inside the distro, pins distribution/user across recovery, and adds a pane-local Claude shim so Resume uses the exact conversation ID. It reworks both spawn paths and crosses the Windows/WSL boundary via WSLENV, so the argv/env handling deserves close review. P2: fixes real WSL bugs but is a draft pending a live Claude check.

openwong2kim:
Heads-up before your next rebase: #1245 (pick which WSL distro the default terminal boots, for #1103) just merged to main and touches the same spawn path as this PR: the spawn args in `DaemonSessionManager.createSession`, the `createSession` replay calls in `src/daemon/index.ts`, `src/daemon/types.ts`, `pty.handler.ts` and `PTYManager.ts`.

Two things will need reconciling:

1. **The `-d <distro>` flag.** #1245 injects the chosen distro into the WSL spawn args (validated where the session spawns, and falling back to the system default when the chosen distro is no longer installed). Your WSL arg rebuild regenerates those args, so as written it would drop that flag. It should carry it through.
2. **One source of truth for the distro.** This PR persists the distro as `wslTarget`, while #1245 persists the user's choice in settings and in the saved spawn args. Ideally `wslTarget` is derived from the distro #1245 actually resolved, so recovery can never disagree with the setting.

The CI runs for this PR are approved, so you'll get results on your next push. Thanks for the careful work here, and for #1265, which is merged.

openwong2kim (changes requested):
Thank you for this PR — the WSL recovery design (distro-side cwd validation, per-pane Claude hooks through the existing bridge, the opt-in runtime test) is well thought out. I ran a detailed review; findings below, ordered by importance.

## Blocking

### 1. The WSL probe runs synchronously on the daemon's event loop
`resolveWslCwd`'s default probe is `execFileSync(wsl.exe, …, { timeout: 15_000 })`, and it is called inside `DaemonSessionManager.createSession` (also `PTYManager.create` in local mode, and every recovery/restart/promote replay path). WSL cold boots routinely take 10–60s. Consequences:

- A pane create while WSL is cold freezes the whole daemon for up to 15s — every other pane's output and every RPC stalls with it.
- After a daemon restart with N WSL panes, `recoverSessions` probes them serially: up to N×15s of frozen daemon before any pane is back.

Suggestion: make the probe async (`execFile` + timeout), cache the resolved target per (distro, user), and in the recovery path probe panes in parallel (or skip-and-retry) instead of blocking per pane.

### 2. Probe failure during recovery discards the pane silently
In `recoverSessions`, a probe throw (distro uninstalled, directory deleted, wslpath missing, 15s timeout) is not classified as transient (the ERROR_INVALID_PARAMETER/87 retry only covers ConPTY), so the session is marked dead: the scrollback snapshot is not restored and the pane is gone — with one log line. The changelog says "an unavailable recovery target fails visibly", but nothing reaches the user, and the old homedir fallback is gone for WSL. Consider keeping the pane in a retryable/pending state with a visible error instead of dead.

## Non-blocking

- `src/main/pty/PTYManager.ts` — with the non-WSL path, `spawnArgs = hookInjection.args` now drops `options.shellArgs` entirely (previously `[...shellArgs, ...injection.args]`). No current caller hits it, but it's a silent breaking change of a public option; keeping the old concatenation for non-WSL costs nothing.
- `resumeLaunchCommand` (`src/daemon/index.ts`) skips the cwd-exists check for WSL exec panes, so a deleted Linux cwd still gets `--resume` — the bootstrap's `cd` then fails, the pane exits 1, and the fresh-start fallback no longer exists.
- Quoting: the runtime test covers single quotes/spaces/Unicode in the cwd, but not embedded `"`. If node-pty's ConPTY argv join doesn't escape embedded double quotes, a `"` in a Linux path could split wsl.exe argv. Either reject `"` in WSL cwds or add a round-trip quoting test.
- `WMUX_WSL_CWD` transport failure → the bootstrap's `builtin cd` fails and the pane exits 1 with no diagnostic. A stderr line (value + reason) would make this debuggable.
- The WSL exec unit's bootstrap sources the interactive `.bashrc` (`. $WMUX_WSL_BASHRC; eval "$1"`), so `.bashrc` stdout (echo/MOTD) can pollute exec output and OSC parsing. An exec-specific non-interactive variant (cd only) may be safer.
- `recoveryCwds.args`/`wslTarget` come from the renderer and, after format validation, become wsl.exe spawn argv (including `--user`). For dead-pane replacement that's fine; consider ignoring them on fresh-pane creates so the settings picker stays the only source for new panes.

## CI

The `bench-confirm` failure on this PR is not from this diff — main itself is currently red on Perf (a same-tree rerun of the last green main commit reproduces it; attribution shows the renderer ready-tail grew ~1s while daemon internals are unchanged). I'm resolving that separately and will rerun the gate here once it clears.

thomas-pedersen:
Thanks for the detailed review. The fixes are pushed through `64cf042`; I also incorporated your merge containing #1291 and #1286.

Both blocking findings are addressed:

- **Non-blocking WSL startup:** probing is asynchronous, concurrent probes for the same target share an in-flight request, and pending sessions recover in the background. Completed cwd results are not cached, so directory removal or a changed default user is detected. Background exec recovery also works without a GUI attached.
- **Retryable recovery failures:** a failed probe preserves the pane ID, conversation binding, and saved scrollback across retries and daemon restarts. The terminal shows the error with **Retry connection**. Closing the pane or shutting down cancels pending creation; successful retry restores terminal visibility and geometry.

The other review findings are covered too: native local `shellArgs` are retained; missing Linux directories cannot silently resume in home; double-quoted WSL paths are rejected before spawn; missing cwd transport produces a diagnostic; exec launches avoid interactive Bash startup output; and recovery distro/user/arguments come from daemon-owned metadata rather than renderer-supplied overrides. Non-interactive PATH requirements are documented.

Validation on this head:

- [Windows validation](https://github.com/openwong2kim/wmux/actions/runs/34823632694), [Windows/macOS/Ubuntu baselines](https://github.com/openwong2kim/wmux/actions/runs/34823632740), and [performance/artifact checks](https://github.com/openwong2kim/wmux/actions/runs/34823632671) all passed.
- Local Linux: 15,420 unit tests and 197 runtime tests passed, with the daemon bundle required. Focused Windows checks: 78 passed.
- Real Windows + Ubuntu-24.04/ConPTY regression passed, including two distinct conversation IDs in one directory, repeated restarts, a missing-directory failure that retains scrollback through another restart, and successful exact-ID recovery after restoring the directory.

The Windows/WSL regression uses a simulated Claude CLI with the actual hooks and Windows bridge. The PR remains draft pending a manual check with real Claude conversations. The review fixes are ready for another look.

openwong2kim:
Thanks for the thorough follow-up, @thomas-pedersen. We re-reviewed `64cf042` and ran a live macOS pass of the non-WSL paths this change touches. Both blocking items are fixed. The async probe coalesces concurrent requests and recovery runs in parallel, and a failed probe now keeps the pane pending with its id, binding and scrollback. Most of the non-blocking items check out too.

We did find one regression that is **not WSL-specific**: it affects supervised panes on macOS, Linux and native Windows. `restartSupervisedSession` is now `async`, but `PaneSupervisor.fireRestart` still calls it inside a synchronous `try/catch` (and `deps.restartSession` is still typed `(id) => void`). A failed restart escapes as an unhandled rejection, skips the backoff/`onSessionDied` path, and `session.restarted` is broadcast as if it succeeded.

Because this touches platforms outside WSL and we can verify it on our side, **we'll take it from here and push the remaining fixes directly to this branch now**. Please hold off on pushing to `fix/wsl-session-recovery` until we post again, so we don't step on each other. Planned:

- Handle async restart failures in `PaneSupervisor` (route rejections into the existing failure path, broadcast only on success) with a rejecting-mock test.
- Limit the `listSessions` recovery lookup in `pty.handler.ts` to WSL shells, so non-WSL dead-pane replacement stays identical to main.
- Add per-id cancellation to local-mode `PTYManager.createAsync`, so closing a surface mid-probe can't spawn an orphan PTY.
- Check the non-WSL behavior change in `promoteSession`: keep it (and note it in the changelog) if the fix needs it, otherwise scope it to WSL.

We'll track the smaller follow-ups separately: a "start fresh" escape for deleted exec directories, a TTL for pending recovery entries, and PATH for nvm-installed `claude` under `--norc`. You don't need to do anything more for this PR. We'll post again once it's ready to merge. Thanks again for the careful work on this.

openwong2kim (approved):
Both blocking findings are fixed. The maintainer follow-up commits (supervisor async restart, WSL-only recovery lookup, local-mode create cancellation, promote parity) passed tests and a live macOS check of the supervised-restart success and failure paths. Approving pending CI.

openwong2kim:
Merged — thank you, @thomas-pedersen. WSL panes now keep their Linux directory and their exact Claude conversation across restarts, and the fix ships in the next release.

What we added on top of your branch before merging:

- `58b37136` — `PaneSupervisor` now handles async restarts: sync throws and async rejections both go through the existing backoff/guard path, and `session.restarted` is broadcast only after a restart actually succeeds. Verified live on macOS: a restart that fails is counted and eventually trips the runaway guard, with no `session.restarted` and no unhandled rejection. A successful restart broadcasts only after the pane is re-created.
- `0765b80c` — the `listSessions` recovery lookup in `pty.handler.ts` now runs only for WSL shells, so non-WSL dead-pane replacement matches main.
- `46dcc5d9` — local-mode `PTYManager.createAsync` reserves the id before probing, and `dispose(id)` during the probe cancels the spawn.
- `f51c620f` — we kept your `promoteSession` change on all platforms. Boot recovery already does `--resume`, supervisor re-arm, binding exposure and an immediate save everywhere; cap-skipped promotes were the one path missing them. There is now a changelog line and a test for it.

The remaining follow-ups are tracked in #1305: a "start fresh" escape for deleted exec directories, a TTL for pending recovery entries, PATH for nvm-installed `claude` under `--norc`, and renderer-side cancellation for in-flight local-mode creates. Contributions there are welcome if any of them interest you, but there's no expectation.

Thanks again for the careful, well-tested work on a hard cross-platform problem.

openwong2kim:
## Live dogfood, 2026-09-16 — not provable on this machine

This PR was on the Windows live-dogfood list for the 2026-09-16 pass (isolated
`WMUX_DATA_SUFFIX=-dog916` instance built from `origin/main` @ `2cbeea8e`,
Windows 11 26200). It could not be exercised: **the box has no WSL distribution
at all.**

```
> wsl.exe --list --verbose
Wsl/CallMsi/Install/REGDB_E_CLASSNOTREG

> Get-AppxPackage MicrosoftCorporationII.WindowsSubsystemForLinux
Version : 2.6.3.0   Status : Ok

> Get-ChildItem HKCU:\SOFTWARE\Microsoft\Windows\CurrentVersion\Lxss
(no distributions registered)
```

The store package is present but the component is not registered and no distro
is installed, so none of the central claims — Linux cwd validated inside the
selected distro and restored after Bash startup files, exact Claude conversation
ID restored per pane, the distro picker honoured for new panes, failed recovery
keeping pane + scrollback with a Retry control across restarts — can be observed
live. This is the largest Windows-only surface in 3.55.0 (587 Windows/WSL marker
hits in the diff) and it is currently shipping unverified outside CI.

Getting it covered needs `wsl --install` plus a reboot on the dogfood machine;
not done here because it is a machine-wide change. Say the word and I will set
the distro up and run the restart/recovery matrix properly.

Same situation for #1308.

PR #1268, opened 2026-09-11
fix: hand the real $? to the wrapped pwsh prompt

### LINKED ISSUE #1267: [Bug]: pwsh shell integration clobbers $? for the wrapped prompt, breaking oh-my-posh exit-code segments
### What happened?

The pwsh shell-integration `prompt` wrapper destroys `$?` before delegating to the prompt function it wrapped. Any prompt engine that reads `$?` to detect the last command's success therefore always sees "success", regardless of the real exit code.

Concretely: with oh-my-posh, a `status`-type segment renders the success color forever in a wmux pane. The same config in Windows Terminal colors correctly. wmux's own OSC 133 `D;<exit>` report is unaffected — it uses the snapshot taken at the top of the wrapper — so only the *wrapped* prompt is broken.

**Root cause.** The wrapper correctly snapshots `$?` and `$LASTEXITCODE` as its first two statements, and its comments call out this exact trap:

> `# Capture $? and $LASTEXITCODE as the VERY FIRST statements. Any`
> `# comparison, assignment, or cmdlet call inside this function resets`
> `# $? to true [...] This same trap bites VS Code / Windows Terminal`
> `# integrations; the fix is to snapshot both variables before doing anything else.`

But the snapshot is never handed to the wrapped prompt. Roughly ten statements run between the snapshot and the delegation — the `$ec` assignment, `$esc`/`$bel`/`$pre` assignments, the `if ($loc.Provider.Name -eq 'FileSystem')` test, the `-split | ForEach-Object | -join` pipeline, the `$pre +=` — and each one resets `$?` to `$true`. By the time control reaches:

```powershell
$body = if ($global:__wmux_prev_prompt) {
    try { & $global:__wmux_prev_prompt } catch { ... }
}
```

`$?` is `$true` unconditionally. `$LASTEXITCODE` does survive intact (it's only written by native executables, and the wrapper restores it at the end anyway) — but that doesn't save consumers that check `$?` first.

**Why that breaks oh-my-posh specifically.** `Update-PoshErrorCode` in oh-my-posh's pwsh init short-circuits on `$?` and only falls back to `$LASTEXITCODE` when `$?` is false:

```powershell
if ($script:OriginalLastExecutionStatus) { $script:ErrorCode = 0; return }   # <-- always taken under wmux
# ... error-record check ...
if ($script:OriginalLastExitCode -is [int] -and $script:OriginalLastExitCode -ne 0) {
    $script:ErrorCode = $script:OriginalLastExitCode                          # <-- unreachable under wmux
}
```

So `.Code` is pinned at `0` and the intact `$LASTEXITCODE = 7` is never consulted.

### Steps to reproduce

1. Install oh-my-posh (reproduced on 29.30.0) and save this minimal config as `%TEMP%\omp-repro.toml`. The `{{ .Code }}` in the template makes the value visible as text rather than only as a color:

   ```toml
   "$schema" = "https://raw.githubusercontent.com/JanDeDobbeleer/oh-my-posh/main/themes/schema.json"
   version = 4

   [[blocks]]
   alignment = "left"
   type = "prompt"

   [[blocks.segments]]
   type = "status"
   style = "plain"
   foreground_templates = ["{{ if eq .Code 0 }}#a3be8c{{ else }}#bf616a{{ end }}"]
   template = " CODE[{{ .Code }}] "

   [blocks.segments.options]
   always_enabled = true
   ```

2. In **Windows Terminal**, start pwsh and run:

   ```powershell
   oh-my-posh init pwsh --config $env:TEMP\omp-repro.toml | Invoke-Expression
   cmd /c exit 7
   ```

   Prompt shows `CODE[7]` in red. Correct.

3. In a **wmux** pwsh pane, run the same two commands. Prompt shows `CODE[0]` in green. Wrong — the exit code was 7.

4. Confirm the shell itself still has the exit code, and that `prompt` is not oh-my-posh's function:

   ```powershell
   cmd /c exit 7; "Q=$? LEC=$LASTEXITCODE ISOMP=$((Get-Command prompt).ScriptBlock.ToString() -match 'PoshPrompt')"
   ```

   Observed in a wmux pane:

   ```
   Q=False LEC=7 ISOMP=False
   ```

   `Q=False`/`LEC=7` prove the data is present at the shell level; `ISOMP=False` confirms `prompt` is the wmux wrapper, with oh-my-posh's version displaced into `$global:__wmux_prev_prompt`.

Note for step 3: because the wrapper is injected *after* the user profile runs, re-running `oh-my-posh init` inside an existing pane replaces the wrapper and masks the bug. To reproduce inside a pane that has already been re-inited, restore the chain by stashing the fresh oh-my-posh prompt into `$global:__wmux_prev_prompt` and reinstalling the wrapper scriptblock over `function:global:prompt`.

### Expected behavior

The wrapped prompt should observe the same `$?` the user's command actually produced, so exit-code-driven prompt segments work identically in wmux and Windows Terminal. With the fix below applied, the instrumented template reads:

| Last command | Expected `.Code` | Before fix | After fix |
|---|---|---|---|
| `cmd /c exit 7` | 7 | **0** | 7 |
| `cmd /c exit 0` | 0 | 0 | 0 |
| `Get-Item Q:\nope-xyz` (failing cmdlet) | 1 | **0** | 1 |

### Suggested fix

oh-my-posh's pwsh init already provides an escape hatch for prompt wrappers — it prefers `$global:NVS_ORIGINAL_LASTEXECUTIONSTATUS` over `$?` whenever that global is a `[bool]`:

```powershell
if ($global:NVS_ORIGINAL_LASTEXECUTIONSTATUS -is [bool]) {
    # make it compatible with NVS auto-switching, if enabled
    $script:OriginalLastExecutionStatus = $global:NVS_ORIGINAL_LASTEXECUTIONSTATUS
}
else {
    $script:OriginalLastExecutionStatus = $?
}
```

It's named for NVS but is a general hook. Setting it from the snapshot the wrapper already holds fixes this in one line, immediately before the delegation:

```powershell
$global:NVS_ORIGINAL_LASTEXECUTIONSTATUS = $__wmux_ok
$body = if ($global:__wmux_prev_prompt) {
    try { & $global:__wmux_prev_prompt } catch { "PS $($executionContext.SessionState.Path.CurrentLocation)> " }
}
```

`$__wmux_ok` is already captured at the top of the wrapper, so this adds no work and no new failure mode. Verified against all three rows of the table above.

Two notes if you'd rather solve it generically than lean on an oh-my-posh-specific global:

- Restoring `$?` for real is awkward in PowerShell — it isn't assignable. Forcing it false via something like `Write-Error -ErrorAction Ignore` would work mechanically, but it pollutes prompt-time state and, with `SilentlyContinue`, pushes a synthetic record onto `$Error` that oh-my-posh's error-record branch then reads as `Code = 1`, masking the true exit code. The `NVS_*` global avoids all of that.
- Deferring the snapshot won't help either: any statement before the delegation is enough to reset `$?`, so the wrapper can't both do its OSC work and leave `$?` pristine. Passing the snapshot forward is the only clean option.

### wmux version

3.52.0

### How did you install wmux?

winget

PowerShell 7+ (pwsh)

### Windows version

Windows 11 Enterprise 23H2, build 10.0.22631

None / not relevant

```shell
No daemon logs needed — the failure is entirely in the injected pwsh prompt wrapper and is fully captured by the `Q=False LEC=7 ISOMP=False` probe in step 4. To dump the exact wrapper source a pane is running:


(Get-Command prompt).ScriptBlock.ToString() | Set-Content $env:TEMP\wmux-prompt.txt
```

### Confirmation

byte2pixel:
## Summary

The pwsh shell integration wraps the user's `prompt` function and snapshots `$?` as its very first statement, but never handed that snapshot to the prompt it wraps — and every statement in between resets `$?` to true. Any prompt engine that reads `$?` to detect the last command's result therefore saw "success" forever inside a wmux pane. This passes the real status through to the wrapped prompt, in both the daemon-mode wrapper and the local-mode hook.

## Changes

- `src/daemon/shell-integration.ts` — re-assert the `$?` snapshot immediately before delegating to `$global:__wmux_prev_prompt`. Bump `INTEGRATION_VERSION` 9 → 10 so `installShellIntegration()` actually rewrites the on-disk script.
- `src/main/pty/shell-hooks/pwsh.ps1` — same defect on the local-mode fallback path: the `Test-Path` guard sits between the function entry and the delegation. Snapshot on entry, re-assert inside the `if`, immediately before the call.
- Unit tests pinning the placement and the `-ErrorAction Ignore` choice, plus a runtime test that drives a real ConPTY pane.

## Why `Write-Error -ErrorAction Ignore`

`$?` is not assignable, and the snapshot cannot be deferred — any statement before the delegation is enough to reset it — so the value has to be re-created. `-ErrorAction Ignore` is the variant with no side effects. Measured, not assumed:

- sets `$?` to `$false`, and reaches `$global:?` too (so Starship is covered, not just oh-my-posh);
- records **nothing** in `$Error`. `SilentlyContinue` does push a synthetic record, and oh-my-posh reads the newest error record as exit code 1 — so that variant would mask the real exit code rather than fix anything;
- does not throw under `$ErrorActionPreference = 'Stop'`;
- behaves identically on Windows PowerShell 5.1 and pwsh 7+.

The issue suggested `$global:NVS_ORIGINAL_LASTEXECUTIONSTATUS` instead. I went the generic route because that global only helps oh-my-posh, and once written it is a `[bool]` forever — so oh-my-posh never consults `$?` again, and if the wrapper is later displaced (re-running `oh-my-posh init` inside a pane, which the issue notes people do) the stale value freezes the status permanently. Happy to switch if you'd rather keep the change narrower.

**Placement is the whole fix.** A single cmdlet call between the restore and the delegation resets `$?` straight back to true and the restore becomes a silent no-op — it looks identical in review and fails only at runtime. That is exactly how the local-mode hook's `Test-Path` broke it, and my first attempt there reproduced the mistake. The tests assert adjacency for that reason.

### Fixed

- **Exit-code segments in a custom pwsh prompt work again.** wmux wraps your `prompt` function to emit its command-boundary markers, and it snapshotted `$?` correctly — but never handed that snapshot to the prompt it wrapped. Roughly ten statements ran in between, and in PowerShell every statement resets `$?` to true, so oh-my-posh's `status` segment (and Starship, and anything else reading `$?`) reported success after every failed command inside a wmux pane, while the same config coloured correctly in Windows Terminal. The wrapper now re-asserts the real status immediately before delegating, and the local-mode prompt hook — the one used when the daemon is gone — is fixed the same way. (#1267)

## Stability tier impact

The OSC 133 wire format is unchanged; this only affects what the wrapped prompt observes.

n/a — no PaneMetadata, EventBus, permission, pipe, or RPC surface touched.

## Test plan

**Unit** — `src/daemon/__tests__/shellIntegration.test.ts`, `src/main/pty/__tests__/pwshHook.test.ts`: the restore is adjacent to the delegation with nothing in between, sits after the snapshot, and uses `-ErrorAction Ignore` rather than `SilentlyContinue`.

**Runtime** — `src/daemon/__tests__/shellIntegration.runtime.test.ts`: spawns a real `powershell.exe` pane through `DaemonSessionManager.createSession()`, installs a probe as `$global:__wmux_prev_prompt`, and asserts the probe observes `False` after a failing command and `True` after a successful one. Whatever that scriptblock sees is exactly what oh-my-posh or Starship would see, so this tests the real contract rather than the text of the script.

All four new assertions fail without the fix (verified by reverting it).

**Manual** — against the emitted script, on pwsh 7.6.6 and Windows PowerShell 5.1:

| last command | before | after |
|---|---|---|
| `cmd /c exit 7` | `$? = True` | `$? = False` |
| `cmd /c exit 0` | `$? = True` | `$? = True` |
| `Get-Item Q:\nope-xyz` | `$? = True` | `$? = False` |

`$Error` stays empty and `$LASTEXITCODE` still reads 7 afterwards. Confirmed `.version` flips 9 → 10 on disk and the installed script carries the fix.

**Suite** — `npx tsc --noEmit` clean; `npm test` green (15,151 parallel + 225 runtime).

## Related issues / PRs

Closes #1267

## Two adjacent bugs found while measuring — want follow-up issues?

Neither is touched by this PR. Both are in wmux's *own* exit-status reporting rather than the wrapped prompt, and both change what daemon consumers observe, so they seemed to deserve their own review rather than riding along here. Both reproduced on the real script. Say the word and I'll open issues (or fold them in, if you'd prefer).

**1. `shell-integration.ts` emits `133;D;0` after a failing cmdlet.**

```powershell
$ec = if ($null -ne $__wmux_le) { $__wmux_le } elseif ($__wmux_ok) { 0 } else { 1 }
```

`$LASTEXITCODE` is preferred over the `$?` snapshot, and it is non-null from the first native command onward — so the `$__wmux_ok` branch is effectively dead for the rest of the session. `Get-Item Q:\nope-xyz` emits `133;D;0`. This feeds `agent.lifecycle.exitCode`. VS Code's integration checks `$?` first. Note this contradicts the original issue's assumption that OSC 133 reporting was unaffected — it is, for native exit codes, but not for cmdlet failures.

**2. `pwsh.ps1` never restores `$LASTEXITCODE`.**

Its own `git rev-parse` (line ~52) zeroes `$LASTEXITCODE` on every prompt render and nothing puts it back, so after `cmd /c exit 7` the user sees `$LASTEXITCODE = 0`. The daemon wrapper handles this deliberately; the local-mode hook does not.

## Reviewer checklist

- [x] Substrate contract docs (PROTOCOL.md, inventory.md, stability.md) updated if the surface changed — n/a, no surface change.

https://claude.ai/code/session_01GBZkAo17s8gJxM8AwK4WUy

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed PowerShell prompt status reporting so oh-my-posh, Starship, and similar tools correctly reflect whether the previous command succeeded or failed.
  * Preserved accurate command status in both connected and local terminal sessions.
  * Updated shell integration to ensure the corrected behavior is applied consistently.

### DIFF
--- a/changelog.d/1268.md
+++ b/changelog.d/1268.md
@@ -1,0 +1,12 @@
+### Fixed
+
+- **Exit-code segments in a custom pwsh prompt work again.** wmux wraps your
+  `prompt` function to emit its command-boundary markers, and it snapshotted
+  `$?` correctly — but never handed that snapshot to the prompt it wrapped.
+  Roughly ten statements ran in between, and in PowerShell every statement
+  resets `$?` to true, so oh-my-posh's `status` segment (and Starship, and
+  anything else that reads `$?`) reported success after every failed command
+  inside a wmux pane, while the same config coloured correctly in Windows
+  Terminal. The wrapper now re-asserts the real status immediately before
+  delegating, and the local-mode prompt hook — the one used when the daemon is
+  gone — is fixed the same way. (#1268, #1267)
--- a/src/daemon/__tests__/shellIntegration.runtime.test.ts
+++ b/src/daemon/__tests__/shellIntegration.runtime.test.ts
@@ -79,6 +79,44 @@ function waitForEventAfter(
   });
 }
 
+/**
+ * Poll the session's raw output until `pattern` matches something written
+ * after `baselineBytes`.
+ *
+ * The prompt body is rendered by the shell itself — it is not reported as a
+ * PromptEvent — so verifying what the WRAPPED prompt observed means reading
+ * the PTY stream rather than promptLog.
+ */
+function waitForOutputAfter(
+  managed: ManagedSession,
+  baselineBytes: number,
+  pattern: RegExp,
+  label: string,
+  timeoutMs = EVENT_TIMEOUT_MS,
+): Promise<RegExpMatchArray> {
+  return new Promise((resolve, reject) => {
+    const deadline = Date.now() + timeoutMs;
+    const tick = () => {
+      const fresh = managed.ringBuffer.readAll().subarray(baselineBytes).toString('utf8');
+      const match = fresh.match(pattern);
+      if (match) {
+        resolve(match);
+        return;
+      }
+      if (Date.now() > deadline) {
+        reject(
+          new Error(
+            `timed out waiting for ${label} — tail after baseline: ${JSON.stringify(fresh.slice(-400))}`,
+          ),
+        );
+        return;
+      }
+      setTimeout(tick, 50);
+    };
+    tick();
+  });
+}
+
 describe.runIf(hasPowerShell)('OSC 133 runtime — powershell.exe', () => {
   let manager: DaemonSessionManager;
 
@@ -162,6 +200,53 @@ describe.runIf(hasPowerShell)('OSC 133 runtime — powershell.exe', () => {
       'command_end with non-zero exitCode',
     );
     expect(cmdEnd.exitCode).toBe(7);
+  }, EVENT_TIMEOUT_MS + 2000);
+
+  // Issue #1267. The wrapper delegates to $global:__wmux_prev_prompt, so
+  // whatever $? that scriptblock observes is exactly what oh-my-posh or
+  // Starship would observe. Installing a probe there exercises the real
+  // contract through a real ConPTY — a substring assertion on PWSH_INIT can
+  // prove the restore is present but not that it actually survives to the
+  // delegation, which is the half that was broken.
+  it('hands the real $? to the wrapped prompt', async () => {
+    manager = new DaemonSessionManager();
+    const id = `rt-pwsh-status-${Date.now()}`;
+    manager.createSession({
+      id,
+      cmd: POWERSHELL,
+      cwd: path.resolve(process.cwd()),
+    });
+
+    const managed = manager.getSession(id)!;
+
+    // Stand in for the user's prompt engine. $q must be assigned as the
+    // scriptblock's first statement — reading $? later would measure this
+    // probe's own bookkeeping instead of the command that just ran.
+    managed.ptyProcess.write('$global:__wmux_prev_prompt = { $q = $?; "WMUXPROBE[$q]" }\r');
+    await waitForOutputAfter(managed, 0, /WMUXPROBE\[(?:True|False)\]/, 'the probe prompt to render');
+
+    // A failing native command: the wrapped prompt must see $? = False.
+    const beforeFailure = managed.ringBuffer.readAll().length;
+    managed.ptyProcess.write(`& "${CMD_EXE}" /c exit 7\r`);
+    const afterFailure = await waitForOutputAfter(
+      managed,
+      beforeFailure,
+      /WMUXPROBE\[(True|False)\]/,
+      'a prompt render after a failing command',
+    );
+    expect(afterFailure[1]).toBe('False');
+
+    // ...and True again after one that succeeds, so a fix that simply pins
+    // the status to False cannot pass.
+    const beforeSuccess = managed.ringBuffer.readAll().length;
+    managed.ptyProcess.write(`& "${CMD_EXE}" /c exit 0\r`);
+    const afterSuccess = await waitForOutputAfter(
+      managed,
+      beforeSuccess,
+      /WMUXPROBE\[(True|False)\]/,
+      'a prompt render after a successful command',
+    );
+    expect(afterSuccess[1]).toBe('True');
   }, EVENT_TIMEOUT_MS + 2000);
 });
 
--- a/src/daemon/__tests__/shellIntegration.test.ts
+++ b/src/daemon/__tests__/shellIntegration.test.ts
@@ -181,3 +181,39 @@ describe('BASH_INIT — OSC 7 cwd report (#540)', () => {
     expect(BASH_INIT).not.toContain('"${HOSTNAME-localhost}" "$p"');
   });
 });
+
+// Issue #1267: the wrapper snapshots $? as its very first statement but used
+// to delegate to the wrapped prompt ~10 statements later — and in PowerShell
+// EVERY statement resets $? to true. So any prompt engine that reads $? to
+// detect the last command's status (oh-my-posh `status`, Starship) saw
+// "success" forever inside a wmux pane, while the same config in Windows
+// Terminal coloured correctly. $? is not assignable and the snapshot cannot be
+// deferred, so the wrapper re-creates the value immediately before delegating.
+describe('PWSH_INIT — the wrapped prompt sees the real $? (#1267)', () => {
+  const promptBody = PWSH_INIT.slice(PWSH_INIT.indexOf('function global:prompt'));
+
+  it('re-asserts the status with nothing between it and the delegation', () => {
+    // Placement IS the fix. A single cmdlet call in between — a Test-Path,
+    // say — resets $? straight back to true and the restore silently becomes
+    // a no-op, which looks identical in review and fails only at runtime.
+    expect(promptBody).toMatch(
+      /if \(-not \$__wmux_ok\) \{ Write-Error [^\n]*-ErrorAction Ignore \}\s*\r?\n\s*\$body = if \(\$global:__wmux_prev_prompt\)/,
+    );
+  });
+
+  it('re-asserts AFTER the snapshot is taken, never before', () => {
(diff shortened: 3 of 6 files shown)


openwong2kim:

Fixes #1267: re-asserts the snapshotted `0` immediately before the wrapped prompt is invoked, in both the daemon wrapper (`shell-integration.ts`) and the local-mode `pwsh.ps1` hook. The `Write-Error -ErrorAction Ignore` restore is placement-critical, and unit tests plus a real-ConPTY runtime test pin it. P2 — clean, well-tested fix for a P2 prompt-integration bug with no OSC 133 wire change.

openwong2kim:
Shipped in v3.53.0 — thank you, and welcome as a first-time contributor.

What stood out in this one: you measured the PowerShell semantics instead of assuming them. The `-ErrorAction Ignore` vs `SilentlyContinue` comparison (that `SilentlyContinue` pushes a synthetic record oh-my-posh reads as exit code 1) is exactly the detail that would have turned a "fix" into a different bug, and the ConPTY runtime test pins the real contract rather than the text of the script. Fixing the local-mode hook alongside the daemon wrapper, and bumping `INTEGRATION_VERSION` so existing installs actually pick it up, closed the whole path rather than half of it.

The two adjacent bugs you spotted while measuring are now filed as #1269 (the local hook never restoring `$LASTEXITCODE`) and #1270 (`133;D;0` after a failing cmdlet). No obligation to take them — but if you want them, they're yours.

PR #1238, opened 2026-09-08
feat(pane): snap running panes into a layout template (#1237)

openwong2kim:
## What

Closes #1237.

- **`snapToLayoutTemplate`** (`uiSlice`): re-fits the RUNNING panes into a saved layout template — the non-destructive twin of `applyLayoutTemplate`, which replaces the tree with fresh empty leaves and kills the sessions.
  - Existing leaves map onto template slots in tree order (top-left → bottom-right), keeping pane ids, surfaces and ordinals.
  - Surplus panes: **stashed** when stashable (daemon ring required), **discarded** when empty (nothing alive), **refused** with a named toast when a tab cannot survive unmount (editor/diff/git) or the daemon is down.
  - Deficit: fresh empty leaves, ordinals continuing past the workspace high-water (the #977 collision rule).
  - Pane cap enforced counting stashed panes, same as apply.
- **Entry points**: one palette command per template (`Snap Running Panes to Layout: X`) and a snap section in the pane ⋮ / right-click menu (`Snap panes to X`).

## Design notes

- No sixth cluster button: the five-button width contract (`PANE_ACTIONS_CLUSTER_WIDTH`) is the subject of two DESIGN.md rulings — the verb is workspace-level anyway, so the menu + palette are its chrome-free homes.
- Snap stashes carry no `origin` anchor: the old topology is gone, so unstash honestly falls back to "next to the active pane".
- User-defined presets already existed (`LayoutTemplate`, palette "Save Current Layout as Template"); this PR supplies the missing snap-back action the issue asked for.

## Tests

- `uiSlice.snapLayout.test.ts` (7): identity/surface/ordinal preservation, surplus→stash, empty surplus discard, daemon refusal, unstashable-surface refusal, cap block, focus/zoom hygiene.
- `SurfaceTabs.actions.test.tsx`: menu key set + arrow-key walk updated for the snap entries (196 passing in the touched suites).

PR #1277, opened 2026-09-11
fix(pane): make divider double-click even out the pair; keep pane sizes in sync across splits, closes and snaps

### LINKED ISSUE #1233: [Feature]: When having 2 panes side by side, double clicking on the edge between them should split them 50/50 either vertically or horizontally

openwong2kim:
## Summary

Double-clicking a pane divider now actually evens out the two panes it separates. Before this, a real double-click collapsed one pane instead (80/20 became 3/97; a pair dragged to 81/19 became 16.7/83.3 about 200 ms later), in both split directions and in nested splits. Along the way this makes the store ↔ panel-library size sync hold across child-set changes (split, close, snap, template count changes), and makes unstash always produce sizes that sum to 100.

Fixes #1233. Follow-up to #1240 (the first attempt) and #1238.

## Root cause

Verified against the shipped `react-resizable-panels` 4.7.4 source (`dist/react-resizable-panels.js`) and measured in jsdom with React 19 and the real library.

**Double-click**
- **Units.** The size parser (`zt`) maps a `number` to pixels; only strings carry a unit (`"50%"`). `PaneContainer` passed the stored percentage as `defaultSize={pane.sizes[i]}`, so 81 became 81 px, about 16.7% of a 485 px group — the value measured live.
- **The library's own double-click.** A capture-phase `dblclick` listener on `document` (`De`) hit-tests the divider's grab band and resizes the first panel of the pair to its `defaultSize`, the px-misread value. This collapse **predates #1240**: it is what #1233 reported.
- **Why an element handler never sees it.** The library grabs a divider anywhere in a 16 px band (37 px on a coarse pointer) around the 1 px line. A double-click beside the line targets the neighbouring panel. One exactly on the line has its pointerdown on the divider and its pointerup on a panel, so the browser dispatches the dblclick to their common ancestor, `BODY`.

**Size sync on child-set changes**
- The library registers a changed panel set one commit **after** `PaneContainer`'s sync effect runs (`registerPanel` only schedules a re-render). The sync effect's `setLayout` therefore landed on the previous registration. With a changed child count it threw `Invalid N panel layout`. Otherwise it left the library holding a layout keyed by ids that no longer exist: after split-then-close, the returning pane rendered at `flexGrow: 1` next to its neighbour's 70, and every later drag reported the stale ids.
- On re-registration the library revives its cached layout for that panel-id set. Its report was swallowed by the `programmaticRef` flag, which was still armed by the misdirected `setLayout`. So a snap back onto the same pair kept an old drag's widths on screen while the store said 50/50.

**Unstash sizes**
- `attachBeside` copied the stashed pane's recorded pair verbatim. That pair is the two panes' shares of the ORIGINAL parent, so a pane stashed out of a three-way split recorded e.g. `[33.3, 33.3]`, and the new two-child branch persisted that sum to the session file and archive snapshots.

## Fix

One rule: **the store is the only source of truth for sizes; the library only ever receives `setLayout` for the panel set it has actually registered.**

- `defaultSize` is a percent string.
- **Store → library.** The sync effect pushes the store's sizes only when `getLayout()` is keyed by exactly the current children. Otherwise it marks a resync pending and does nothing, so there is no misdirected `setLayout`, no poisoned keys, and no throw. When the push happens, a pending drag write is cancelled because it is superseded.
- **Library → store** (`onLayoutChanged`): every report is sorted into one of four cases:
  - keyed by other ids → never persisted (the library is catching up);
  - equal to the store → an echo or a no-op drag; nothing is persisted, and a pending write is dropped;
  - the first report after the panel set changed (resync pending) → the library's cache or defaults, never the user. The store's sizes are pushed back (deferred a microtask, re-checked against the current children) instead of being persisted over them;
  - otherwise → a user drag, persisted through the existing 200 ms debounce.

  This replaces the `programmaticRef` flag, which could stay armed.
- **Double-click.** Detected on `window` in the capture phase, ahead of the library's handler, and decided **by coordinates**:
  - the point is inside this group's rect;
  - it lies within the band of one of this group's own dividers, using the library's band (37 px when `(pointer: coarse)` matches, 16 px otherwise); the nearest divider wins where bands overlap;
  - no nested group has a band there, so at a T-junction only the innermost pair is evened out.

  Targets in an unrelated subtree (a portalled popover) are ignored; `BODY` and other ancestors are accepted. 0x0 (display:none) dividers never match. A hit evens out that pair, reading the live layout when it covers the current children (so a drag on another divider still inside its debounce is kept). It writes through `updatePaneSizes` to the tile's own workspace, which now takes an optional `workspaceId`, so a non-active multiview tile works. Then it calls `preventDefault()`, which makes the library's `De` return early.
- `attachBeside` rescales the recorded pair to sum to 100, keeping its proportion.

**Not chosen: keying the `Group` by its child-id set.** That would give fresh library state on every structural change, but it remounts every `Panel`'s subtree. Every sibling leaf's terminal would then go through the park/adopt cycle (#1002) on any split, close or snap, which is the churn the adoption work exists to avoid.

Not changed: `minSize={10}` is also read as pixels, and has been since the initial commit. Moving it to `"10%"` would make a row of 11+ panes (the per-workspace cap is 20) unsatisfiable, so it stays a 10 px floor.

## Tests

- `PaneContainer.layoutSync.test.ts` (new): real library, with panel `offsetWidth`/`offsetHeight` stubbed so it measures a real group, reports real layouts and keeps its per-panel-set cache. Asserts the **rendered** `flexGrow` for:
  - split then close on the same group;
  - drag, split, then snap back onto the same pair (converges to 50/50, and the cached 17/83 is not written back after the debounce);
  - child count 3 → 2 → 3 (no throw, no stranded pane).

  All three fail against the previous commit of this branch.
- `PaneContainer.separatorEqualize.test.ts`:
  - a double-click on the neighbouring panel inside the band (and the event reaches the library's document-capture listener already `defaultPrevented`);
  - exactly on the line, dispatched to `BODY`;
  - the nested line, including its centre;
  - a T-junction (innermost pair only);
  - away from every divider (left alone);
  - a portalled popover (left alone);
  - the coarse vs fine band;
  - a non-active multiview tile;
  - `defaultSize` reaching `Panel` as `"80%"`;
  - `separatorIndexAt` picks the nearest divider and skips 0x0 ones.
- `paneSlice.stash.test.ts`: a pane stashed out of a 20/30/50 row comes back as `[40, 60]`; an entry with no origin comes back as `[50, 50]`.
- `paneSlice.test.ts`: `updatePaneSizes` with a workspace id writes to that workspace, not the active one.
- `uiSlice.snapLayout.test.ts`: unstashing a snap-stashed pane lands in a pair that sums to 100.

## Residual

- A dogfood report mentioned a persisted `[100,100]` after unstash. No store writer produces it, and the renderer paths that could persist a bad sum are closed here, but I did not reproduce that exact value.
- The resync relies on the library reporting after it re-registers a panel set, which it does whenever the new layout differs from its last report. Confirmed in jsdom with the real library; the live dogfood is the final check.

## Summary by CodeRabbit

- **Bug Fixes**
  - Double-clicking pane dividers now evenly splits only the adjacent panes, including nested layouts, touchscreens, and inactive multiview tiles.
  - Improved divider hit detection, including clicks directly on divider lines and overlapping divider areas.
  - Pane sizes now remain consistent after splitting, closing panes, snapping layouts, and changing layout structures.
  - Closing panes no longer leaves neighboring panes as thin slivers.
  - Restored panes now receive normalized split sizes, with balanced sizing when previous size information is unavailable.


openwong2kim:

Root-causes what the earlier #1233 fixes missed: `defaultSize` read as pixels, and the library's own dblclick reset beating the dead separator handler. Renderer-only (`PaneContainer.tsx`, `paneSlice.ts` unstash), but jsdom can't cover the drag-debounce cancellation, so that needs a live check. P2 — third attempt at a visible layout bug that also persisted broken split sizes.

PR #1281, opened 2026-09-12
fix(search): clear highlight decorations when the search bar goes away

### LINKED ISSUE #1266: [Bug]: Search isn't functional
### What happened?

<video src="https://github.com/user-attachments/assets/963e98f9-222f-43fd-a4e7-68134658e06c" />

### Steps to reproduce

1. Launch wmux
2. Press 'Ctrl + F'
3. Search a letter
4. It highlight unrelated characters + the highlighting doesn't update with the scrolling, so you get the search artifacts frozen into the pane

### Expected behavior

1. Launch wmux
2. Press 'Ctrl + F'
3. Search a letter
4. It highlight words & characters you search for

### wmux version

3.52.0

### How did you install wmux?

Setup.exe (direct download)

PowerShell 7+ (pwsh)

### Windows version

Windows 10 19045.6466

Claude Code

```shell

```

### Confirmation

openwong2kim:
## Summary

Ctrl+F/Cmd+F search highlights get stranded in a pane: they survive the
search bar being closed with no way to dismiss them, and an abandoned pane
keeps re-applying them as new output arrives, at coordinates that no longer
match anything. Fixes #1266.

## Root cause

Two layers, both ownership problems rather than search problems.

**1. The bar belongs to the wrong thing.** `Terminal.tsx` gated the search
bar on `searchBarVisible && isActive`, but `Pane.tsx` passes
`isActive={surface.id === activeSurfaceId}` — the selected **tab inside a
pane**, not "this pane has focus". `searchBarVisible` is a single global
flag, so *every* pane rendered its own search bar at once (live dogfood
measured two bars up simultaneously), and moving focus to another pane never
unmounted the abandoned pane's bar.

**2. A search stays live until it is explicitly cleared.**
`@xterm/addon-search` caches the term on `findNext(term, { decorations })`
and installs an `onWriteParsed` hook that re-runs the search and re-creates
every highlight decoration 200ms after each later chunk of output.
`clearSearch()` was only wired to the bar's close button. So closing the bar
stranded the highlights, and an abandoned pane went on re-highlighting for
its whole life — dogfood measured those decorations sitting at stale
coordinates (-245,-76), drawn as orange marks bleeding over the neighbouring
pane. That is the visual half of "highlights unrelated characters".

### Two hypotheses checked and rejected

- **Scan bound.** An earlier draft patched `@xterm/addon-search` to scan to
  `buffer.active.length` instead of `buffer.active.baseY + rows`. Rejected:
  `baseY` is `ybase` (`lines.length - rows`) and scrolling moves
  `ydisp`/`viewportY`, not `baseY`, so the bounds are identically equal and
  the patch is a semantic no-op. Measured with the real addon against a real
  terminal (patched vs unpatched bundles, scrolled up): identical results.
  Confirmed independently by two reviewers and by dogfood — search from a
  scrolled-up position finds and cycles matches across the whole scrollback.
- **Renderer.** `BufferDecorationRenderer` failing to reposition on
  redraw-only frames. Dogfood shows scrolling with an active search tracks
  correctly. Dead end, no follow-up needed.

## Fix

- Gate the bar on real pane focus: the workspace's `activePaneId`, resolved
  from the surface via a new `findLeafBySurfaceId` helper in
  `shared/paneUtils.ts`. Exactly one bar is up.
- Clear the addon's decorations whenever the bar goes away for any reason —
  close button, global toggle, tab switch within a pane, or focus leaving
  the pane.

### Intended UX consequence

Only the focused pane shows the search bar (previously all of them did), and
leaving a pane **ends** its search: highlights and the addon's cached term
are both dropped, and reopening the bar starts from an empty query. The
query box is per-mount `useState` and already did not survive a pane switch;
the highlights now match it instead of outliving it.

## Tests

`src/renderer/components/Terminal/__tests__/searchBarUnmountClears.test.tsx`
— mounts the real Terminal component (jsdom, `createRoot` + `act`, in the
style of `ContextMenu.portal.test.tsx`) and asserts the addon's
`clearDecorations()` actually ran:

- bar up → not cleared; global toggle off → cleared; tab swap → cleared;
- **the reporter's scenario**: two panes side by side, focus on A, then
  focus moves to B — exactly one bar is up throughout, and A's highlights
  are cleared when focus leaves. Both of those cases fail without the focus
  gate (verified by reverting it).

`src/renderer/components/Terminal/__tests__/searchDecorationLeak.test.ts` —
real `@xterm/addon-search` against a real `@xterm/headless` terminal, no
addon mock; `registerDecoration` returns a decoration with the lifecycle
surface the addon uses, so creation and teardown are genuinely exercised:

- an abandoned search keeps re-running on every later write (the "output
  arrives in A" half of the reporter's scenario);
- `clearDecorations()` disposes every decoration it created and quiesces the
  write hook;
- matches are found across the whole scrollback while scrolled up (guards
  the rejected scan-bound direction).

Gates: `npx tsc --noEmit` clean; eslint clean on changed files (the two
warnings on `Terminal.tsx` — unused `useMemo` import, non-null assertion —
are pre-existing on `main`); `vitest run src/renderer/terminal
src/renderer/components/Terminal` → 28 files / 421 tests, and `vitest run
src/shared src/renderer/stores` → 130 files / 2367 tests, all passing.

## Reviewer checklist

- [ ] One search bar in the focused pane is the wanted model (see "Intended
      UX consequence").
- [ ] `findLeafBySurfaceId` falling back to "focused" for surfaces it cannot
      place (stashed panes, transitional trees) is the right default — it
      preserves today's behaviour rather than hiding a bar.

## Follow-ups, deliberately out of scope

- Scoping search state per pane properly (a `searchBarVisible` + query keyed
  by pane, so a pane can keep its query across a focus round-trip). This PR
  makes the existing global flag mean what it claims; owning per-pane state
  is a store-slice change with its own keybinding and results-panel
  implications.
- #1266 reply: on macOS Ctrl+F is consumed by xterm and Cmd+F opens the bar.
  If the reporter pressed Ctrl+F on Windows and got terminal behaviour, that
  is a separate discoverability question to answer in the issue.

Fixes #1266

### DIFF
--- a/changelog.d/1281.md
+++ b/changelog.d/1281.md
@@ -8,3 +8,11 @@
   re-highlighting an abandoned term — landing on whatever text occupied those
   cells by then, with no search box left to dismiss it. The highlights are now
   cleared whenever the search bar goes away, not only via its close button.
+
+### Changed
+
+- **Leaving a pane now ends its search.** Focusing another pane drops the
+  highlights and the addon's cached term, and reopening the search bar starts
+  from an empty query. The query box was already per-mount local state, so it
+  never survived a pane switch; the highlights now match it instead of
+  outliving it.
--- a/changelog.d/1281.md
+++ b/changelog.d/1281.md
@@ -1,18 +1,21 @@
 ### Fixed
 
-- **Search highlights no longer linger in a pane after you leave it (#1266).**
-  The search bar is only rendered for the active pane, so focusing another
-  pane made it disappear without ever ending the search. `@xterm/addon-search`
-  treats a search as live until it is explicitly cleared and re-applies its
-  highlights after every subsequent chunk of output, so the pane kept
-  re-highlighting an abandoned term — landing on whatever text occupied those
-  cells by then, with no search box left to dismiss it. The highlights are now
-  cleared whenever the search bar goes away, not only via its close button.
+- **Search highlights can no longer be stranded in a pane (#1266).** Closing
+  the search bar left every highlight behind with no way to dismiss them —
+  the "artifacts frozen into the pane" in the report. Worse, the bar was
+  shown per *tab* rather than per focused *pane*, so every pane rendered its
+  own bar and moving to another pane never ended the abandoned pane's
+  search: `@xterm/addon-search` treats a search as live until it is
+  explicitly cleared and re-applies its highlights after every subsequent
+  chunk of output, so that pane kept re-highlighting a term nobody was
+  searching for, at coordinates that no longer matched anything — orange
+  marks drifting over neighbouring panes. The search bar now belongs to the
+  focused pane, and its highlights are cleared whenever it goes away.
 
 ### Changed
 
-- **Leaving a pane now ends its search.** Focusing another pane drops the
-  highlights and the addon's cached term, and reopening the search bar starts
-  from an empty query. The query box was already per-mount local state, so it
-  never survived a pane switch; the highlights now match it instead of
-  outliving it.
+- **Only the focused pane shows the search bar, and leaving a pane ends its
+  search.** Previously every pane showed one simultaneously. Moving focus
+  elsewhere now drops that pane's highlights and the addon's cached term;
+  reopening the bar starts from an empty query (the query box was already
+  per-mount state and never survived a pane switch).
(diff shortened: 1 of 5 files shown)


openwong2kim:
Clears xterm search highlights whenever the search bar unmounts (one effect in `Terminal.tsx`), stopping the addon from re-highlighting an abandoned term on later output — fixes #1266. A small, well-tested change, though the reporter's scroll artifact may have a second renderer-level cause worth a Windows dogfood. P2 — restores broken search with minimal risk.

PR #1282, opened 2026-09-12
fix(remote): return to a local workspace on the first click

openwong2kim:
## Summary

Removes a class of "the first click did nothing" bugs around attached remote
workspaces, found while chasing the remaining Bug 1 of #1086.

**Scope honesty, up front:** the reporter's original repro — click the
already-active local workspace row (or a nested pane row under it) after
visiting a mirror — could NOT be reproduced on current `main`. Three independent
reads agree: `setActiveWorkspace` has no same-id early return and always calls
`clearRemoteSelection`, the sidebar row's `onSelect` dispatches unconditionally,
and `useNotificationListener.ts:142` / `:235` already carry `|| state.activeRemoteKey`
from #1089 / #1138. What this PR actually fixes is adjacent drift off the same
invariant (orphan-adopt and Ctrl+click multiview), plus the reason that drift
was possible at all. **#1086 Bug 1 stays open** pending a live trace, alongside
its other open item (startup parity for remote workspace creation) — hence
`Refs`, not a closing keyword. The rename/color half shipped in #1247 and is
untouched.

## Root cause

`activeWorkspaceId` (workspaceSlice) and `activeRemoteKey` (remoteWorkspacesSlice)
are independent fields, and `WorkspaceCenter` checks the remote key FIRST — the
local pane tree stays `display: none` until something drops the remote selection.
The store's answer was a convention documented on `clearRemoteSelection`: "call
this at every site that assigns `activeWorkspaceId`". A convention is not a
mechanism, and two paths had drifted off it — both of which present as a
swallowed click:

- `orphanSessionsSlice.adoptOrphanSession` assigns `state.activeWorkspaceId`
  directly and never called `clearRemoteSelection`: adopting an orphan session
  row while a mirror is showing changed the local tree behind a mirror that kept
  covering it.
- `uiSlice.toggleMultiviewWorkspace` (Ctrl+click) never assigns
  `activeWorkspaceId` at all, so no amount of assignment-site discipline could
  cover it — yet the multiview grid IS the local viewport.

Both reproduce with the real store on `main` (see Tests).

## Fix

Make the invariant hold in one place instead of at every call site:

- `activateLocalWorkspace(state, id)` in `workspaceSlice` is now the only way to
  make a local workspace active: assign + `clearColdParkEntry` +
  `clearRemoteSelection`. Every site routes through it — the six workspaceSlice
  sites, `removeWorkspace`'s promotion, both companySlice promotions, and
  orphan-adopt.
- `toggleMultiviewWorkspace` clears the selection for the WHOLE action rather
  than only the join branch. Chosen deliberately: joining the grid, leaving it,
  and collapsing it by un-picking the last partner all land the user on the
  local view, and a rule that fires on some Ctrl+clicks but not others is the
  drift this PR exists to remove.
- `isRemoteMirrorVisible(state)` in `remoteWorkspacesSlice` is the single
  definition of "a mirror is what's on screen". `WorkspaceCenter`, `ToolbarHost`,
  `CommandPalette`'s fan-out guard and `Pane`'s question-seen guard all read it,
  so chrome and centre cannot disagree. `useNotificationListener` deliberately
  keeps the raw flag: its `FocusTargetState` is a minimal surface built by hand
  in fixtures, the predicate needs `remoteWorkspaces` too, and there it errs the
  safe way (one extra activation). Documented at the field.
  The predicate also ignores a key with no live entry behind it. That is a
  defensive invariant, **not** a bug fix: no reachable path produces a dangling
  key today (`detachRemoteWorkspace` nulls it, `attachRemoteWorkspace` sets both
  in one mutation, `restoreRemoteWorkspace` never selects, and the key is not
  persisted).
- A guard test fails the build if a raw `activeWorkspaceId` assignment reappears
  anywhere in `src/renderer` — `workspaceSlice.ts` included, with the one legal
  assignment exempted by an inline `// guard-allow` marker that a second test
  pins to exactly one line. It matches any receiver name and the
  `set({ activeWorkspaceId: … })` patch shape, verified against a throwaway
  probe file containing both.

Note: this adds a `uiSlice` → `workspaceSlice` import. Safe — `clearRemoteSelection`
is called at action time, not module-init time.

## Tests

New `remoteSelectionInvariant.test.ts` (real store: workspace + pane + ui +
orphan + remote slices) and `activeWorkspaceIdAssignment.guard.test.ts`.

Against `origin/main`:

- adopt an orphan session while a mirror shows → `activeRemoteKey` stays
  `'host-1:ws-1'` (expected `null`) — **fails**
- Ctrl+click into (and out of) the multiview grid → same — **fails**
- guard scan → offenders in `companySlice.ts`, `orphanSessionsSlice.ts` and
  (with the hardened version) `workspaceSlice.ts` — **fails**
- the already-active-click case, the `activatePaneTarget` /
  `focusNotificationTarget` cases and the render-gate cases pass on `main` once
  the new export exists; they are regression locks for #1089 / #1138 and for the
  reporter's path, not demonstrations of a bug.

Gates: `npx tsc --noEmit` clean, eslint clean on all changed files, full
`npx vitest run` = 15507 passed / 1 failed — the known `THIRD_PARTY_NOTICES`
drift guard, which fails only because this worktree symlinks `node_modules`.

## Dogfood

With a mirror attached and selected, one click each: the already-active local
workspace's sidebar row, a nested agent row under it, an orphan-session row, and
Ctrl+click a second local workspace (grid must appear at once), then Ctrl+click
it back out. The first two are the reporter's path and are expected to already
work on `main` — if they take two clicks there, the real mechanism is still
unfound and #1086 Bug 1 needs a live trace.

### DIFF
--- a/changelog.d/1282.md
+++ b/changelog.d/1282.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- **Returning to a local workspace after visiting a remote one takes one click
+  again.** The remote mirror and the local workspace were tracked by two
+  independent selections, and the mirror won the tie — so any action that made
+  the local viewport the thing you wanted to see, but forgot to drop the mirror
+  selection, left the mirror covering it and looked like a swallowed click.
+  Adopting an orphan session and Ctrl+clicking a workspace into the multiview
+  grid both did that. Local activation now runs through one place that always
+  drops the mirror, and a detached mirror can no longer leave the centre blank
+  (#1086).
--- a/changelog.d/1282.md
+++ b/changelog.d/1282.md
@@ -1,11 +1,9 @@
 ### Fixed
 
-- **Returning to a local workspace after visiting a remote one takes one click
-  again.** The remote mirror and the local workspace were tracked by two
-  independent selections, and the mirror won the tie — so any action that made
-  the local viewport the thing you wanted to see, but forgot to drop the mirror
-  selection, left the mirror covering it and looked like a swallowed click.
-  Adopting an orphan session and Ctrl+clicking a workspace into the multiview
-  grid both did that. Local activation now runs through one place that always
-  drops the mirror, and a detached mirror can no longer leave the centre blank
-  (#1086).
+- **Adopting an orphan session, or Ctrl+clicking the multiview grid, no longer
+  leaves a remote mirror covering the local view.** A remote mirror and the
+  local workspace are two independent selections and the mirror wins the tie, so
+  these two actions changed the local view underneath a mirror that kept
+  covering it — the click looked swallowed and only a second, different
+  selection got you back. Local activation now runs through one place that
+  always drops the mirror selection (#1086).
--- a/src/renderer/components/AgentToolbar/ToolbarHost.tsx
+++ b/src/renderer/components/AgentToolbar/ToolbarHost.tsx
@@ -3,6 +3,7 @@ import { useStore } from '../../stores';
 import AgentToolbar, { AGENT_TOOLBAR_HEIGHT } from './AgentToolbar';
 import { useComposeShortcut } from './useComposeShortcut';
 import { useHoverReveal, HOVER_TRIGGER_ZONE_PX } from './useHoverReveal';
+import { isRemoteMirrorVisible } from '../../stores/slices/remoteWorkspacesSlice';
 
 /**
  * Owns the reveal state for AgentToolbar and the element the trigger band is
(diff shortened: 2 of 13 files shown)


openwong2kim:
Fixes the two-click return to a local workspace (#1086 Bug 1) by routing every `activeWorkspaceId` assignment through `activateLocalWorkspace`, which always drops the remote mirror selection, with `isRemoteMirrorVisible` as the single render gate. Renderer store slices plus regression and guard tests make it low risk. P3 — usability fix with an existing workaround, cleanly tested.

PR #1283, opened 2026-09-12
fix(updater): start the Windows install waiter outside wmux's job object

### LINKED ISSUE #1264: Windows auto-update always aborts: install waiter is killed before it can run Setup.exe
## Summary

On Windows the in-app auto-update never completes while wmux is running. Every attempt ends with the **generic** abort marker:

```
[AutoUpdater] previous install was refused: install-aborted: wmux quit to install the
update, but the installer step was interrupted before it could report an outcome.
Try again, or run the installer from the releases page.
```

The wording is the key detail: `wait-and-install.ps1` writes that generic text **pessimistically at the top**, then overwrites it with a specific reason on every one of its own exit paths (`a process under the install root would not exit` → exit 3, `install root still locked` → exit 2, `the installer could not be started` → exit 4, `incomplete installation` → exit 6). The generic text surviving means the waiter reached **none** of its own branches — it was killed mid-flight. `Setup.exe` is never started at all.

## Environment

- Windows 10 Pro 22H2 (19045), x64
- wmux 3.51.0 → 3.52.0, Squirrel install under `%LOCALAPPDATA%\wmux`
- Installed via winget (`openwong2kim.wmux`)

## Evidence

Four failed attempts left artifacts in `%TEMP%`, and they split into two distinct failure modes:

| Attempt | Artifacts present | Interpretation |
|---|---|---|
| 2026-09-05 10:27:09 | `wait-and-install.ps1` only | waiter script written, **never launched** (no stamp file) |
| 2026-09-05 10:28:38 | `wait-and-install.ps1` only | same |
| 2026-09-05 10:29:50 | `wait-and-install.ps1` only | same |
| 2026-09-09 08:41:43 | `launch-waiter.vbs`, `launched-w.txt`, `wait-and-install{,-b,-w}.ps1` | waiter **started**, then died before any branch |

The 09-09 run also left `%APPDATA%\wmux\update-install-ready.tmp` containing `alive` — written on the waiter's first lines — and `launched-w.txt` containing `launched`. So the process definitely ran, and definitely never got to clean up.

Relevant `main-2026-09-09.log` lines:

```
[AutoUpdater] reusing verified installer already in temp (wmux-update-3.51.0-712-wmux-3.51.0.Setup.exe) — skipping download
[AutoUpdater] Session save triggered before update install
[installTeardown] waiter verified via hidden wscript launcher (pid 118372): C:\Users\Daniel\AppData\Local\Temp\wmux-install-waiter-eOsG6n\wait-and-install-w.ps1
[AutoUpdater] install handoff: waiter=...\wait-and-install-w.ps1 watching 5 process(es) under C:\Users\Daniel\AppData\Local\wmux; force-killed 0/0 foreign, 5 of our own left to app.quit(), daemon pid 80908 left to the graceful shutdown
[AutoUpdater] quitting for install — the daemon goes down with us so the installer runs against a dead tree
[AutoUpdater] previous install was refused: install-aborted: ... interrupted before it could report an outcome.
```

## Analysis

The waiter is spawned *from* wmux (via the hidden `wscript` launcher) and is then expected to outlive it. Two candidate reasons why it does not, both consistent with "died before reaching any branch":

1. **Windows Job Object.** If the Electron processes run in a job object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, every descendant is terminated when the job closes — the `wscript` hop does not escape a job unless the child is created with `CREATE_BREAKAWAY_FROM_JOB`. This would kill the waiter at exactly the moment wmux quits, which matches the log ordering.

2. **The waiter's own force-kill.** It runs

   ```powershell
   & taskkill.exe /PID $procId /T /F
   ```

   over `$forceKillEligible` (4 of the 5 pids). `/T` terminates the **entire process tree** of that pid. If the waiter is still a descendant of one of those pids, it kills itself here.

I could not distinguish the two after the fact — the processes are gone. A `Get-CimInstance Win32_Process` parent-chain dump at handoff time, logged before quitting, would settle it.

Two secondary observations:

- The 09-05 runs suggest an earlier variant where the waiter was never launched at all. The `launch-waiter.vbs` hop looks like a fix for that, and it did change the symptom — but not the outcome.
- The daemon pid is deliberately excluded from `$forceKillEligible` and gets the full 60 s `$budget`. With live panes running agents, a graceful shutdown may well exceed 60 s, which would produce exit 3. That is a *separate* failure waiting to happen; it is not what fired here (the specific message never appeared).

## Suggested fix

Launch the installer from something that is structurally outside the app's process tree and job object. The most robust option on Windows is a one-shot scheduled task:

```
schtasks /create /tn wmux-update-<rand> /sc once /st <now+1min> /tr "<setup.exe> ..." /f
```

Task Scheduler runs it under `svchost`, so neither `app.quit()` nor `taskkill /T` can reach it. Alternatives: create the waiter with `DETACHED_PROCESS | CREATE_BREAKAWAY_FROM_JOB`, or ship a tiny separate updater `.exe` rather than a PowerShell child.

Independently, it would help a lot if the pessimistic marker distinguished *"waiter never started"* from *"waiter started and vanished"* — right now both surface as the same generic message, which is what made this take a while to pin down.

## Workaround (works reliably)

The nupkg path via `Update.exe` bypasses the whole handoff and does not need the app fully dead:

```
%LOCALAPPDATA%\wmux\Update.exe --update=https://github.com/openwong2kim/wmux/releases/download/v3.52.0/
```

This installed 3.52.0 cleanly on the first try (`app-3.52.0` created, `RELEASES` updated, Run key and shortcuts repointed). Note it still terminates running instances (`[Squirrel] terminated running wmux instance(s) before install work: 17680`), and unlike the in-app path there is no preceding session save — so panes should be closed first.

openwong2kim:
## Summary

Windows in-app auto-update never completed: the install waiter started, wrote its
`alive` heartbeat and its launch stamp, and then died before reaching any of its
own exit branches, so `Setup.exe` was never launched. This adds a waiter
transport that the app's exit cannot reach.

## Root cause, with evidence

The reporter's artifact set is decisive: `update-install-ready.tmp` = `alive` and
`launched-w.txt` = `launched` (the waiter's first two statements ran), and then
the abort marker still holds the *pessimistic* text that every one of the
script's exit branches overwrites. A process killed by the kernel cannot reach a
branch.

In `src/main/updater/installTeardown.ts`, all three transports (W = `wscript`
trampoline, A = `cmd.exe` trampoline, B = direct) reach the waiter through
`child_process.spawn` from the Electron main process. On Windows that makes the
waiter a **member of whatever job object the app is in**: libuv never passes
`CREATE_BREAKAWAY_FROM_JOB` (its own `win/process.c` comment says so), and
`detached: true` only buys `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` —
neither leaves a job. A job created with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`
terminates every member when its last handle closes, i.e. exactly when wmux's
last process exits. The `wscript` hop does not help; it is spawned the same way,
so the whole trampoline is inside the job along with the PowerShell it hosts.

The second hypothesis in the issue — the waiter's own `taskkill /T` reaching back
up its parent chain — is **not** what fired here. That call only runs after
`forceKillGraceMs` expires for a force-kill-eligible pid, and the reported run
logged `5 of our own left to app.quit()` with no stuck pid and no exit-3 marker.
It remains a real hazard on the refusal path, and the new transport happens to
remove it too (the scheduler's child is in no tree of ours).

## Fix

Add **transport S**, tried before W/A/B: an **on-demand Task Scheduler entry**,
registered with `schtasks /Create /XML` and started with `schtasks /Run`. The
Task Scheduler service starts the process, so it is not our child, not in our
job, and not in any tree our own `taskkill /T` can walk. No admin rights and no
stored password: `LogonType InteractiveToken`, run as the logged-on user.

Registration is from **XML, not `/TR`** — and that is load-bearing, not tidiness.
A `/TR`-registered task inherits `DisallowStartIfOnBatteries=true` and
`StopIfGoingOnBatteries=true`, which on a laptop means `/Run` starts nothing (we
fall through to the transport that cannot survive the job) and unplugging
mid-wait lets the scheduler terminate the waiter — #1264 again, wearing the same
pessimistic marker. The definition sets `DisallowStartIfOnBatteries=false`,
`StopIfGoingOnBatteries=false`, `AllowHardTerminate=false`,
`ExecutionTimeLimit=PT0S` (the waiter's own budgets are the only deadline) and
`MultipleInstancesPolicy=IgnoreNew`.

It carries **no `<Triggers>` element at all**. A leaked registration therefore
cannot fire by itself — which matters because a late waiter would open the
recorded pids *by number*, `taskkill /T /F` whatever inherited them, possibly
relaunch `Setup.exe` out of `%TEMP%`, and write a false "install root still
locked" marker against an already-updated install. The second half is
`sweepStaleWaiterTasks()` at startup (`schtasks /Query` + `/Delete` for names
matching `^wmux-update-`, re-validated against the same pattern the builder
enforces), plus a best-effort delete in the `/Create` failure path (a timeout can
mean the server committed anyway) and a logged, no-longer-swallowed delete
result.

`RunLevel` mirrors the current process's elevation (`fltMC` probe →
`HighestAvailable` when elevated, `LeastPrivilege` otherwise). Without it an
elevated wmux would hand the install to a medium-IL waiter whose `taskkill`
against the app's own pids fails, `$stuck` fires, and the install is refused with
exit 3 — a failure mode the in-tree transports never had, because they inherited
the app's token.

Other details kept from review:

- The action is still the hidden `wscript` launcher, so the console window #1136
  removed does not come back.
- Every S bail-out (definition unbuildable, `/Create` or `/Run` failure, stamp
  timeout) logs its cause and errno before falling through. The reporter is told
  to look for `waiter verified via scheduled task`; its absence now says *which*
  of the four fired.
- **Quit-path blocking is capped.** schtasks is a local RPC, so the calls get 3s
  (create/run) and 2s (delete); only the stamp poll keeps a generous window, and
  it was cut from 10s to 6s to match W. Worst case S adds ~14s rather than ~40s.
- If the stamp gate expires, the task is deleted and the stamp re-polled for a 1s
  grace, so a scheduler instance that starts just past the deadline is adopted
  rather than raced.
- W/A/B are untouched and still carry the install where Task Scheduler is
  unavailable. Every S failure is fail-closed to them.

The honest-reporting design is kept: every branch still writes its specific
reason. The pessimistic sentinel now states that the waiter **did start** and was
stopped before running the installer, so "killed waiter" reads differently from
"refused install".

## Test plan

**Are the battery settings provably applied?** Yes, at two levels. A unit test
asserts the generated XML carries `DisallowStartIfOnBatteries=false`,
`StopIfGoingOnBatteries=false`, `ExecutionTimeLimit=PT0S`,
`MultipleInstancesPolicy=IgnoreNew`, `AllowHardTerminate=false` and
`InteractiveToken`, and that there is no trigger; and the `windows-latest`
runtime test re-reads the task back with `schtasks /Query /XML ONE` after the
real import and asserts the scheduler **stored** those values. What is *not*
proven by either is behaviour on an actual battery transition — no CI runner has
a battery.

New/changed tests:

- `installTeardown.test.ts` — `buildScheduledTaskXml`: the power settings, the
  absence of a trigger, elevation mirrored onto `RunLevel`, XML escaping of `&`
  and pass-through of non-ASCII with a matching UTF-16 declaration, fail-closed
  on `\r`/`\n`, and the Settings/top-level element **order** (the importer
  validates a sequence; a definition that does not round-trip would fail closed
  and silently cost the fix). `buildScheduledTaskCreateArgs`: `/XML` not `/TR`,
  no `/RU`/`/RP`, and a name it could not address again later is refused.
- `installTeardown.spawn.test.ts` (new file) — launch **path**, win32 forced and
  `child_process` mocked, runs on any host: `/Create` → `/Run` → `/Delete`
  happens before any `spawn` of ours, all three address the same validated name,
  the registered definition on disk is the hidden wscript launcher with the
  battery settings, every exec is `windowsHide` and **≤3s**; fall-through to W on
  a scheduler that proves nothing or throws; a `/Create` failure still attempts a
  delete; and the startup sweep deletes only names matching our pattern (a
  `\wmux-update-nested\evil` and a `GoogleUpdateTask...` are left alone).
- `AutoUpdater.platform.test.ts` — `start()` sweeps on win32 and not elsewhere.
- `installTeardown.runtime.test.ts` — real kernel, `windows-latest` only. A stub
  parent PowerShell joins a real `KILL_ON_JOB_CLOSE` job, launches the waiter and
  exits; a lock holder **outside** the job keeps the install root busy past the
  parent's death, so the waiter can only run the stub `Setup.exe` if it outlived
  the parent.
  - **control**: launched the old way (`Start-Process wscript` from inside the
    job) — the launch stamp appears and the stub installer never runs. This is
    the mechanism proof, and it is red on the pre-fix launch path by
    construction.
  - **fix**: registered and started through the real builders — the waiter
    survives the job closing, the stored XML carries the battery settings, and
    the stub installer's file is written.
  - Per review, the stub's **exit code is returned and diagnosed** (1/2/3 =
    job-object premise, 4/5 = schtasks premise, −1 = exited 0 without arming).
    On `windows-latest` an unmet premise now **FAILS** with a message naming
    which one, unless `WMUX_SKIP_JOB_RUNTIME_TESTS=1` is set. It can no longer
    go green by not running.
- The existing #1136 transport-identity test accepts either hidden transport (S
  or W) and still refuses the visible `cmd.exe` fallback.

Gates: `tsc --noEmit` clean; eslint 0 errors on the changed files (warnings are
the suite's pre-existing non-null-assertion style); `vitest run src/main
src/daemon/__tests__` → 5443 passed / 53 skipped (the Windows-only suites skip on
macOS).

## Residual risk — what still needs the reporter

- **CI proves the mechanism, not the field.** The runtime tests now fail loudly
  if the job-object or schtasks premise is unmet on the runner, so "green" means
  they ran — but a GitHub Windows runner is still not a desktop: no battery, no
  UAC-elevated session, and its Task Scheduler behaviour for an interactive-user
  task may differ from a real user's box.
- **Not proven at all here**: that a winget/Squirrel install under
  `%LOCALAPPDATA%\wmux` on Windows 10 22H2 actually reaches transport S, and that
  a real `Setup.exe` then completes. @DanielIshi — the line to look for in
  `main-*.log` is `waiter verified via scheduled task (wmux-update-…, runLevel=…)`.
  If it instead says `scheduled-task transport unavailable (…)`, that message now
  carries the exact cause and errno, and that is the datum to paste back.
  Battery-specific confirmation (start an update while unplugged, and pull the
  plug mid-install) would be the most valuable single data point.
- Narrow window, logged not fixed: deleting the task **definition** cannot stop
  an instance the scheduler has already started, so a waiter can still appear
  just after we give up on S. The waiter's root mutex makes that safe (the
  newcomer yields with exit 5 and writes nothing), but it is the one case where a
  user told the install did not start may still get it.
- The daemon's 60 s graceful-shutdown budget producing a spurious exit 3 with
  busy panes — the reporter's separate observation — is untouched here.

Fixes #1264

### DIFF
--- a/changelog.d/1283.md
+++ b/changelog.d/1283.md
@@ -4,8 +4,12 @@
   before it could report an outcome" message. The install waiter was spawned as
   a child of the app, which on Windows put it inside the app's job object — so
   it was killed the instant wmux quit, before it could start `Setup.exe`. It is
-  now started by the Task Scheduler instead, outside the app's job object and
-  outside any process tree the teardown can reach. ([#1264](https://github.com/openwong2kim/wmux/issues/1264))
+  now registered as an on-demand Task Scheduler entry and started by the
+  scheduler, outside the app's job object and outside any process tree the
+  teardown can reach. ([#1264](https://github.com/openwong2kim/wmux/issues/1264))
+- Windows: the update install waiter now runs on battery power and is not
+  stopped when a laptop is unplugged mid-install, and it keeps the elevation the
+  app itself was running with.
 - Windows: when the install waiter is killed mid-flight, the message on the next
   start now says the waiter started and was stopped, rather than reading the
   same as a refused install.
--- a/changelog.d/1283.md
+++ b/changelog.d/1283.md
@@ -1,0 +1,11 @@
+### Fixed
+
+- Windows: in-app auto-update no longer aborts with a generic "interrupted
+  before it could report an outcome" message. The install waiter was spawned as
+  a child of the app, which on Windows put it inside the app's job object — so
+  it was killed the instant wmux quit, before it could start `Setup.exe`. It is
+  now started by the Task Scheduler instead, outside the app's job object and
+  outside any process tree the teardown can reach. ([#1264](https://github.com/openwong2kim/wmux/issues/1264))
+- Windows: when the install waiter is killed mid-flight, the message on the next
+  start now says the waiter started and was stopped, rather than reading the
+  same as a refused install.
--- a/src/main/updater/AutoUpdater.ts
+++ b/src/main/updater/AutoUpdater.ts
@@ -33,6 +33,7 @@ import { LocalUpdateFeed } from './LocalUpdateFeed';
 import {
   collectInstallRootProcesses,
   clearAbortMarker,
+  sweepStaleWaiterTasks,
   freeSpaceShortfall,
   INSTALL_ABORT_MARKER,
   INSTALL_READY_MARKER,
@@ -298,6 +299,15 @@ export class AutoUpdater {
       console.log(`[AutoUpdater] In-app updates are not supported on ${process.platform}; skipping auto-check (update via your package manager).`);
       return;
     }
+
+    // #1283 review — sweep waiter task registrations a previous update failed
+    // to clean up (a `/Create` that timed out after the server committed, a
+    // `/Delete` that failed). They carry no trigger so they cannot fire on
+    // their own, but they point at a `%TEMP%` path the installer has since
+    // invalidated and they are litter in the user's task list. Startup is
+    // where this belongs: the quit path is the one place blocking costs the
+    // user something.
+    if (process.platform === 'win32') sweepStaleWaiterTasks();
 
     void this.sweepStaleArtifacts();
 
--- a/src/main/updater/__tests__/AutoUpdater.download.test.ts
+++ b/src/main/updater/__tests__/AutoUpdater.download.test.ts
@@ -176,6 +176,7 @@ async function loadWin32(
     // never leaks an UPDATE_ERROR into tests about other flows.
     readAbortMarker: vi.fn((): string | null => null),
     clearAbortMarker: vi.fn(),
+    sweepStaleWaiterTasks: vi.fn(() => 0),
     waitForWaiterHeartbeat: vi.fn(async (): Promise<boolean> => true),
     INSTALL_ABORT_MARKER: 'update-install-aborted.txt',
     INSTALL_READY_MARKER: 'update-install-ready.tmp',
--- a/src/main/updater/__tests__/AutoUpdater.platform.test.ts
+++ b/src/main/updater/__tests__/AutoUpdater.platform.test.ts
@@ -162,6 +162,7 @@ async function loadForPlatform(
     readDaemonPid: vi.fn((): number | null => null),
     readAbortMarker: vi.fn((): string | null => null),
     clearAbortMarker: vi.fn(),
+    sweepStaleWaiterTasks: vi.fn(() => 0),
     // #1056 — resolves true by default so existing tests exercise the same
     // happy path as before this check existed; the dedicated test below
     // overrides it to prove the refusal branch.
@@ -208,6 +209,24 @@ describe('AutoUpdater platform gating', () => {
 
     expect(requestUrls).toContain(EXPECTED_WIN32_FEED);
     updater.stop();
+  });
+
+  it('win32: start() sweeps leftover waiter task registrations; other platforms do not', async () => {
+    // #1283 review — the two leak paths (a /Create that timed out after the
+    // server committed, a /Delete that failed) can only be cleaned up later,
+    // and startup is the one place blocking costs the user nothing.
+    vi.useFakeTimers();
+    const win = await loadForPlatform('win32');
+    const winUpdater = new win.AutoUpdater(() => null, quitHooks());
+    winUpdater.start();
+    expect(win.teardown.sweepStaleWaiterTasks).toHaveBeenCalledTimes(1);
+    winUpdater.stop();
+
+    const mac = await loadForPlatform('darwin');
+    const macUpdater = new mac.AutoUpdater(() => null, quitHooks());
+    macUpdater.start();
+    expect(mac.teardown.sweepStaleWaiterTasks).not.toHaveBeenCalled();
+    macUpdater.stop();
   });
 
   it('win32: periodic timer keeps polling the win32 feed', async () => {
--- a/src/main/updater/__tests__/installTeardown.runtime.test.ts
+++ b/src/main/updater/__tests__/installTeardown.runtime.test.ts
@@ -750,6 +750,31 @@ describe.skipIf(!onWindows)('#1264 — the waiter must outlive the app that spaw
     fs.writeFileSync(vbs, '\uFEFF' + (launcher as string), 'utf16le');
   }
 
+  /**
+   * The property CRITICAL 2 actually needs: the stored task has no trigger that
+   * could ever fire it, so a registration that leaks cannot start a waiter on
+   * its own later (a late waiter would open the recorded pids BY NUMBER and
+   * taskkill whatever inherited them).
+   *
+   * Measured on both CI runners (validate and the cross-platform baseline,
+   * identical output): Task Scheduler normalises a trigger-less definition by
+   * storing an EMPTY, self-closing `<Triggers />` — it materialises no trigger
+   * of its own. So a substring match on `<Triggers` is the wrong test; what
+   * must be asserted is the absence of CHILD trigger elements.
+   */
+  function expectNoOwnTrigger(xml: string): void {
+    if (!xml.includes('<Triggers')) return;
+    if (/<Triggers\s*\/>/.test(xml)) return;
+    const block = /<Triggers[^>]*>([\s\S]*?)<\/Triggers>/.exec(xml);
+    expect(block, `could not parse the Triggers element out of:\n${xml}`).not.toBeNull();
+    const inner = (block as RegExpExecArray)[1];
+    const found = [...inner.matchAll(/<([A-Za-z][\w.]*)/g)].map((m) => m[1]);
+    // Listed, not just counted: if the scheduler ever DOES materialise a
+    // trigger, the failure has to name it — that would mean this transport
+    // needs a different guarantee (register disabled, enable only for /Run).
+    expect(found, `the scheduler stored trigger element(s) we never asked for: ${found.join(', ')}\nreadback was:\n${xml}`).toEqual([]);
+  }
+
   /** Poll for `p` for up to `ms`. */
   function waitFor(p: string, ms: number): boolean {
     const deadline = Date.now() + ms;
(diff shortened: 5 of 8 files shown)


openwong2kim:
Adds a Task Scheduler transport (S) in `installTeardown.ts` so the Windows install waiter escapes the app's kill-on-close job object, fixing #1264; W/A/B remain as fallbacks. It touches only the update quit path, well covered by unit/spawn/runtime tests, though the CI survival test may skip and field verification is pending. P2 — restores the broken Windows auto-update, with a manual workaround available.

openwong2kim:
## Live dogfood, 2026-09-16 — Windows 11 26200

The packaged update flow itself cannot be exercised on a dev build (no update
can be staged), so I tested the two things that actually decide whether this fix
holds on a given machine: the failure mechanism, and transport S.

### 1. The root-cause mechanism reproduces here

A process assigned to a job object with
`JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE` (`SetInformationJobObject` +
`AssignProcessToJobObject`, both returning `True`) spawns a child through
`CreateProcess` — `ProcessStartInfo` with `UseShellExecute = false`, i.e. the
shape libuv uses:

```
set=True assign=True
child-pid=19300
child 19300 after the job's last handle closed: DEAD
```

So on this box a waiter reached by `child_process.spawn` is killed exactly when
wmux's last process exits, which is the death the PR's artifact set points at.

(The first attempt at this control used `Start-Process`, which goes through
`ShellExecuteEx` and is therefore *not* our child — it survived. Worth knowing
if anyone re-runs this: the control only means something with
`UseShellExecute = false`.)

### 2. Transport S starts a waiter that the job cannot reach

Using this PR's own builders (`buildScheduledTaskXml`,
`buildScheduledTaskCreateArgs`, `buildScheduledTaskRunArgs`, bundled out of
`installTeardown.ts` with esbuild) against a stand-in VBS waiter:

```
task XML built: 37 lines
create args: ["/Create","/TN","wmux-update-dog916probe","/XML","…\dog916-1283-task.xml","/F"]
SUCCESS: The scheduled task "wmux-update-dog916probe" has successfully been created.
SUCCESS: Attempted to run the scheduled task "wmux-update-dog916probe".
waiter stamp written: true
waiter process: {"ProcessId":13824,"ParentProcessId":16816,"parent":"cmd.exe"}
our pid       : 1596
```

The task registers under a plain (non-elevated) user, `/Run` starts it, the
waiter reaches its first statement (stamp written), and it is parented by the
Task Scheduler's chain — **not** by the creating process. Nothing in our job,
which is the whole point of the change. `/Delete` removed it cleanly.

### Still unverified

The end-to-end handoff — waiter survives wmux's real exit, then launches
`Setup.exe` — needs a staged update and a packaged build. Same class of
limitation as #1156 in the 2026-09-06 dogfood.

PR #1285, opened 2026-09-12
fix(terminal): arm the atlas wake recovery with resume, not every alt-tab

openwong2kim:
Investigation of #1234 (rendering corruption after alt-tab, Windows 10 / GTX 1660 Ti). Fixes the cause the reporter's 15-minute log points at, rather than adding another repaint trigger.

## What the log proves

633 lines, 08:44:00 – 08:59:20 (15m20s). Every event in it:

| event | count |
| --- | --- |
| `glyph-repaint burst` | 440 |
| `glyph-repaint settle-verify` | 145 |
| `glyph-repaint focus` | 30 |
| `atlas-guard recover (visibility)` | 18 |
| `glyph-repaint visible` | **0** |
| `atlas-guard prevent` / `cure` | **0** |

- 12 of the 18 recoveries follow a `glyph-repaint focus` by 13–26 ms (median 22.5 ms). That is an alt-tab: the textarea refocuses, then Chromium's occlusion tracker flips `visibilityState` a frame later.
- 11 of the 18 landed within ±1 s of a `burst`, i.e. **mid-stream**.
- Inter-recovery gaps range 3.9 s – 264 s; one recovery per ~51 s of ordinary use.

## The mechanism

`atlasWakeRecovery.ts` fires `atlasGuard.recoverNow()` — an unconditional wipe of the *shared* WebGL glyph atlas plus a render-model drop on every owner pane — on two triggers: `system-resumed` and `visibilitychange`. Its own header says window focus is deliberately **not** a trigger, because "wiping the shared atlas while output is flowing re-arms xterm's page-merge race", and it asserts that on Windows the visibility trigger never fires (measured on Electron 41: `visibilityState` stays `visible` behind another window).

That assertion is falsified by this log. On Windows 10.0.19045 with native occlusion active, `visibilitychange` *is* the window-focus event — so the reporter's machine takes the exact path the module refuses to take, 18 times in 15 minutes, 61% of them mid-stream. Each one empties a shared atlas that is then repacked from scratch under a CJK/box-drawing TUI workload, driving the page pool back toward the merge trigger that #741/#798/#902 exist to contain. `[wmux:glyph-repaint] focus` is not the problem; it is the 20 ms warning that the wipe is about to happen.

## The fix

Visibility no longer recovers on its own. Its documented job is narrow — cover the unlock-screen gap where main's `resume` lands while the window is still hidden, i.e. where the rebuild it triggers can be undone by Chromium before first present — and that is a **one-shot latch**, not a time window:

- A resume delivered while the window is **hidden** arms exactly one visibility recovery. It fires on the next transition to visible and clears the latch.
- The latched recovery is **throttle-exempt**, so a fast unlock is not swallowed by the rebuild the resume itself just performed, and it **never expires**, so a slow password entry still recovers.
- A resume delivered while already **visible** does not arm, which keeps `WAKE_RECOVER_THROTTLE_MS`'s original job of collapsing a real wake into one rebuild.
- An unarmed visibility change (every alt-tab) does nothing but log at Verbose.

### Where the latch is required

| platform | behaviour | why |
| --- | --- | --- |
| **win32, darwin** | latch required from the first visibility change | `powerMonitor` resume is dependable on both, and on Windows we now have field evidence that visibility fires on an ordinary alt-tab, so an unarmed visibility change is known to be noise |
| **linux, unknown/absent platform** | **keeps the pre-#1234 unconditional behaviour** until a resume has been delivered once; the latch is required from then on | `powerMonitor` 'resume' exists as an API everywhere but is not reliably emitted on some Linux session setups. Gating on the API's *presence* would silently leave such a machine with no wake recovery at all. The cost is one unjustified wipe per visibility change there until its first sleep — accepted deliberately |

A first draft gated purely on first delivery, on **every** platform. Live dogfood killed it: it left #1234 unfixed on the reporter's own machine until that machine happened to sleep once, and their report is 15 minutes of alt-tabbing with no reason to think it ever did. A fix that waits for a suspend to take effect is not a fix.

The rejected alternative was to stop trusting the event and detect the wake directly — compare wall-clock elapsed against expected elapsed and treat a large unexplained jump as a sleep. More honest in principle, and it would cover Linux too, but its discriminator is unsound in exactly the state it has to work in: a hidden renderer has its timers throttled, and can have them frozen outright, so an ordinary alt-tab away produces the same unexplained jump as a real suspend and would re-arm the wipe on every app switch — the bug being fixed. The monotonic-clock variant only trades that for a different unknown, since whether the monotonic clock advances across suspend differs by platform; if it pauses, the jump never appears and wake recovery dies silently everywhere. The platform gate rests on something actually measured.

Net effect on the reporter's machine: 18 mid-stream atlas wipes per 15 minutes become 0, while real sleep→wake recovery is strictly improved (a slow unlock now recovers where the pre-existing code's rebuild-while-hidden could be undone before first present).

## Instrumentation for the next report

`recover` lines carried only `atlases=1`, which cannot distinguish a wipe that did nothing from a wipe that repacked a full pool. An entry line is now emitted before the rebuild loop (so a throwing rebuild still leaves an attributable record), followed per atlas by `panes=`, pre-wipe `pages(pre)=U/N used`, `clear=` outcome and `gen=before->after` (sampled on both sides, so one line never mixes a pre-wipe reading with a post-wipe one). The suppressed visibility trigger logs at Verbose. If the reporter still sees corruption on a build with this change, the next log will show zero atlas wipes — which exonerates the guard and moves the investigation to the WebGL context/driver path.

## Live dogfood

Module-level A/B inside the real renderer, on `193bda37`:

- **Proved:** `main` recovers on every visibility transition even after a resume; this build recovers exactly once for an armed hidden→visible transition, logs `visibility ignored — unarmed` otherwise, and adds no extra wipes on repeat transitions. The new `recover (...) rebuilt — ...` line matches the intended shape. No rendering regression after app switches: canvases intact, `focus` / `burst` / `settle-verify` all still firing.
- **Also found:** the first-delivery-only gate made this build behave identically to `main` when no resume had ever been delivered (8 recover lines vs main's 4, the extra being the new `rebuilt` lines) — which is what motivated the platform gate above.
- **Could NOT prove:** macOS does not emit `visibilitychange` on app switch, so the transitions had to be simulated by dispatching the event. That is the same listener path the reporter's platform drives natively, but it is a simulation of the trigger, not of their GPU. No real OS sleep or screen lock was performed, so the armed-recovery path was exercised only through a synthetic resume.

## What is proven vs inferred

Tests: `atlasWakeRecovery.test.ts` rewritten around the latch — 10 cases. Resume-then-unlock recovers twice; the latch survives a 120 s unlock and is consumed exactly once; an unarmed alt-tab storm recovers nothing; a resume while visible does not arm; the latch is required immediately on darwin; linux and unknown/absent platforms stay unconditional until a first delivered resume; teardown detaches both triggers. Plus **the reporter's exact condition**: win32, no resume ever delivered, 20 hide/show cycles, zero recoveries. `tsc`, `eslint`, and all 414 tests under `src/renderer/terminal` pass.

### DIFF
--- a/changelog.d/1285.md
+++ b/changelog.d/1285.md
@@ -1,0 +1,3 @@
+### Fixed
+
+- Terminal rendering no longer risks corruption on every alt-tab on Windows. The shared WebGL glyph atlas rebuild meant for sleep→wake was firing on each window refocus (Chromium's occlusion tracker flips page visibility there), wiping and repacking the atlas mid-stream. It now runs only when the system actually resumed.
--- a/changelog.d/1285.md
+++ b/changelog.d/1285.md
@@ -1,3 +1,3 @@
 ### Fixed
 
-- Terminal rendering no longer risks corruption on every alt-tab on Windows. The shared WebGL glyph atlas rebuild meant for sleep→wake was firing on each window refocus (Chromium's occlusion tracker flips page visibility there), wiping and repacking the atlas mid-stream. It now runs only when the system actually resumed.
+- Terminal rendering no longer risks corruption on every alt-tab on Windows. The shared WebGL glyph atlas rebuild meant for sleep→wake was firing on each window refocus (Chromium's occlusion tracker flips page visibility there), wiping and repacking the atlas mid-stream. It now runs only when the system actually resumed, latched one-shot so a slow unlock still recovers and an alt-tab never does.
--- a/src/renderer/components/Layout/AppLayout.tsx
+++ b/src/renderer/components/Layout/AppLayout.tsx
@@ -816,6 +816,9 @@ export default function AppLayout() {
     return initAtlasWakeRecovery({
       onSystemResumed:
         typeof onResumed === 'function' ? onResumed : () => () => {},
+      // #1234: without a resume push, visibility is the only wake signal this
+      // build has, so it must keep recovering unconditionally.
+      hasSystemResumeSignal: typeof onResumed === 'function',
     });
   }, []);
 
--- a/src/renderer/components/Layout/AppLayout.tsx
+++ b/src/renderer/components/Layout/AppLayout.tsx
@@ -811,22 +811,24 @@ export default function AppLayout() {
   // Rationale and trigger choice live in terminal/atlasWakeRecovery.ts.
   useEffect(() => {
     // optional-chain electronAPI — jsdom (tests) has no preload bridge; an
-    // older main without the push degrades to visibility-only recovery.
+    // older main without the push degrades to visibility-only recovery (the
+    // module tracks whether a resume is ever actually DELIVERED, so a platform
+    // where powerMonitor never fires keeps that fallback — see #1234).
     const onResumed = (window as any).electronAPI?.system?.onResumed;
     return initAtlasWakeRecovery({
       onSystemResumed:
         typeof onResumed === 'function' ? onResumed : () => () => {},
-      // #1234: without a resume push, visibility is the only wake signal this
-      // build has, so it must keep recovering unconditionally.
-      hasSystemResumeSignal: typeof onResumed === 'function',
     });
   }, []);
 
   // #882 — one renderer-wide subscription to "is anyone looking at this
   // window" (minimized / hidden to tray / screen locked), which panes fold
   // into their #766 viewer-visibility report. All platforms: the bit is right
-  // everywhere, it is only Windows where `document.visibilityState` could not
-  // supply it. See hooks/useWindowDisplayed.ts.
+  // everywhere, whereas `document.visibilityState` cannot supply it — on
+  // Windows it is occlusion-driven, so it flips on an ordinary alt-tab and says
+  // nothing about whether the window is minimized or the screen is locked
+  // (#1234 corrected the earlier claim that it never flips there at all).
+  // See hooks/useWindowDisplayed.ts.
   useEffect(() => windowDisplayedStore.init(), []);
 
 
--- a/src/renderer/terminal/__tests__/atlasWakeRecovery.test.ts
+++ b/src/renderer/terminal/__tests__/atlasWakeRecovery.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from 'vitest';
-import { initAtlasWakeRecovery, RESUME_ARM_MS, WAKE_RECOVER_THROTTLE_MS } from '../atlasWakeRecovery';
+import { initAtlasWakeRecovery, WAKE_RECOVER_THROTTLE_MS } from '../atlasWakeRecovery';
 
 type Listener = () => void;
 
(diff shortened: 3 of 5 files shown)


openwong2kim:

For #1234, atlasWakeRecovery now wipes the shared glyph atlas on a visibility flip only within 10s of a real system resume, so Windows alt-tabs stop triggering mid-stream rebuilds. Touches the sleep/wake recovery path; old behaviour is kept when no resume push exists, and recover logs gain page/gen diagnostics. P3 — matches the issue's severity, well tested and low risk.

openwong2kim:
### bench-confirm on this branch is a runner draw, not a regression from this PR

`main` fails the same metric on its own first bench leg. Both runs emit the identical gate message — `the failure reproduced on the same runner — coldStart.firstPtyDataMs (best) (FAIL)` — and escalate to the fresh-runner confirmation job. Only the second machine's draw differs:

| leg | main (`52e1206a`) | this branch (`cd5a6496`) |
| --- | --- | --- |
| bench (first runner) | 2441 / 2449 / 2465 / 2466 ms → **FAIL, escalated** | 2180 / 2371 / 2447 / 2558 ms → **FAIL, escalated** |
| bench-confirm (fresh runner) | 1818 / 1883 ms → passed, recorded as "runner-health noise, not a regression (#940)" | 2534 / 2769 ms → reproduced → **FAIL** |

`bench/baseline-ci.json` records `coldStart.best.firstPtyDataMs = 1134` and was last updated 2026-07-27 (`e5a4f182`). Every cold-start sample on any branch right now is 1800–2850 ms, so whether the gate goes green is decided by how fast the confirmation runner happens to be. The same `coldStart` failure also appeared on the unrelated `fix/1280-ctrl-g-rich-input` branch earlier today.

The `frameBudget.N8.frameDeltaMs.p95` FAIL from the previous run did **not** recur here, consistent with it being noise: in that run N8 measured p95 46.9 ms while N16 measured 31.3 ms, and sixteen flooded panes cannot be cheaper than eight. Every value in that scenario is an exact multiple of the 15.6 ms vsync tick, so p95 moves only in whole dropped frames.

This PR cannot affect either metric: the changed code runs only on `visibilitychange` / `system-resumed`, and nothing in it executes before first PTY data. `bench/baseline-ci.json` is deliberately untouched here — re-baselining the perf gate is a separate, owner-level decision, since it would accept a repo-wide ~2.2x cold-start move.

openwong2kim:

`2cbeea8e`, renderer console watched over CDP while the window really lost and
regained focus (`SetForegroundWindow` to a *maximised* `cmd.exe` and back, 8
round trips, ~24 s).

```
glyph-repaint focus                     : 57
glyph-repaint visible                   : 73   <- visibilitychange DOES fire here
atlas-wake "visibility ignored — unarmed": 1   <- the guard ran and declined
atlas-guard recover (visibility)        : 0
atlas-guard recover (any reason)        : 0
```

The two things worth separating:

1. **The #1234 precondition is present on this machine too.** Chromium's native
   occlusion flips `visibilityState` on an ordinary alt-tab on Windows 11
   26200, not just on the reporter's Windows 10 19045 — `glyph-repaint visible`
   fires 73 times across 8 round trips. The assertion the module used to carry
   ("on Windows the visibility trigger never fires") is falsified here as well.

2. **And no atlas is wiped.** Zero `recover (visibility)`, zero recoveries of
   any reason. The `[wmux:atlas-wake] visibility ignored — unarmed (no pending
   system-resumed)` line proves this is the latch declining, not the module
   failing to load — which is exactly the distinction that debug line was added
   for.

Not checked: a real sleep/wake (`powerMonitor` `resume`) and the unlock-screen
latch, so the arming half of the change is unverified — only the declining half.

PR #1286, opened 2026-09-12
fix(toolbar): Ctrl+G owns Rich Input exactly, and stops at the pane

### LINKED ISSUE #1280: [Bug]: ctrl + g 동작이 항상 리치입력을 활성화 함
### What happened?

- ctrl + g 동작이 항상 리치입력을 활성화 합니다. 3.52.0 버전부터 발생함. 3.51.0은 정상 동작
콘솔입력창에서 ctrl + g입력 -> 콘솔에 ^G 출력 + 리치 입력창 활성화
에이전트 cli에서 ctrl + g입력 -> 설정된 editor 출력 + 리치 입력창 활성화
- ctrl + shift + g 입력 시 : 리치 입력창만 출력

### Steps to reproduce

1. wmux 실행
2. ctrl + g 입력
3. 에이전트 cli 실행
4. ctrl + g 입력

### Expected behavior

ctrl + shift + g 입력할 경우에만 리치 입력 창 활성화.

### wmux version

3.52.0

### How did you install wmux?

Setup.exe (direct download)

Windows PowerShell 5.1

### Windows version

windows 11 25H2

None / not relevant

```shell

```

### Confirmation

openwong2kim:
## Summary

Reported by @River-Jang (#1280, 3.52.0, Windows 11 / PowerShell 5.1; 3.51.0 was fine):

- Ctrl+G in a shell pane → `^G` echoed **and** Rich Input opened.
- Ctrl+G in an agent CLI → the configured external editor opened **and** Rich Input opened.
- Ctrl+Shift+G → Rich Input only.

Both observations are the same defect seen from two sides: the Rich Input binding fired when it should not, and when it fired it did not take the key away from the pane. The binding stays on **Ctrl+G**, and this PR also adds a Settings toggle so anyone who wants Ctrl+G to reach the pane can have it.

## Root cause

**1. The chord gate matched a superset of the modifier set.** `useComposeShortcut` tested

```ts
if (!(e.ctrlKey || e.metaKey) || (e.key !== 'g' && e.key !== 'G')) return;
```

`KeyboardEvent.key` is `'G'` precisely when Shift is held, and nothing rejected Shift or Alt. So **Ctrl+Shift+G** — which `WMUX_KEYMAP` and `useKeyboard` own as `clearMultiview` — also toggled Rich Input, and so did Ctrl+Alt+G. That is why the reporter found Ctrl+Shift+G "working": it was riding a different binding.

**2. The leak to the pane is a regression from #1228 (b6225d3c, "fix: Dvorak Ctrl+C, Codex Shift+Enter, and stale agent slug after exit").** xterm's own encode path ends in `cancel()`, which calls `preventDefault()` **and `stopPropagation()`** — so before #1228 a Ctrl+G was consumed by xterm and never reached the document-level listener, and Rich Input's advertised key was quietly dead inside a terminal. #1228 added a catch-all `resolveCtrlLetterByte` encoder that writes the control byte itself with only `preventDefault()`. The event then keeps bubbling: BEL (0x07) to the pane **plus** the popover. Nothing about Ctrl+G changed in #1228; it changed who stops the event.

`resolveCtrlLetterByte` itself is exact (`if (!e.ctrlKey || e.shiftKey || e.altKey || e.metaKey …) return null`), so the same mistake is not in that helper.

### Other bindings audited

- `useKeyboard.ts` — every branch spells `shift` or `!shift` explicitly. The three that deliberately omit it (`Ctrl+=`, `Ctrl+-`, `Ctrl+Tab`) are documented as intentional in `keymap.ts`, and their shifted aliases are listed in `WMUX_KEYMAP` so the conflict checker sees them. No hazard.
- `matchesDisabledShortcut` in `shared/keymap.ts` builds the combo from the actual `shiftKey` value. Exact.
- `useTerminalCopyShortcut` — `if (!e.ctrlKey || e.shiftKey || e.altKey || e.metaKey) return;`. Exact.
- `useTerminal`'s custom key handler — every bubble branch is gated on `!e.shiftKey`, with a separate Ctrl+Shift catch-all. Exact.
- Every other `keydown` listener in the renderer is an Escape / outside-dismiss handler with no chord matching.

`useComposeShortcut` was the only loose matcher in the renderer.

## Fix

- `useComposeShortcut` matches **exactly**: no Shift, no Alt, and the platform picks the base modifier — ⌘ on macOS (where Ctrl+G is a readline byte), literal Ctrl elsewhere. Adds a physical `KeyG` fallback so the binding survives a Hangul / non-Latin IME, and yields when the user has switched Ctrl+G off in Settings → Shortcuts.
- `useTerminal` bubbles the chord to that listener instead of encoding it, so **neither xterm nor the catch-all ctrl encoder writes anything** when the binding fires — no more `^G`, no more external editor. It gets its own branch rather than a row in the `bubbleKeys` / `bubbleCodes` allowlists, whose condition is only `ctrlKey && !shiftKey`: a row there would also swallow Ctrl+Alt+G and Ctrl+Meta+G, which no handler claims. macOS is untouched: ⌘G is the binding there and Ctrl+G stays a readline byte.
- **Both gates call one exported pure predicate**, `isComposeChord` (`src/renderer/terminal/composeChord.ts`). The general defect behind #1280 is that each gate had its own opinion of the same key: xterm's encode path calls `stopPropagation`, so the pane gate runs first and the popover gate only sees what it let bubble — every condition one gate applies that the other cannot see is a silently dead key. The predicate can be wrong; it cannot be inconsistent. It also rejects `e.isComposing` (with a Hangul preedit open, `key='Process'` + the physical `KeyG` fallback used to pop Rich Input mid-composition, while every other ctrl-letter path defers) and accepts an unshifted capital `G`, which Caps Lock produces.
- **Only the active-leaf terminal owns the chord, on BOTH gates.** The bubble branch runs in every `useTerminal` consumer, but `useComposeShortcut` acts on the workspace's active leaf, so `FloatingPane` (Ctrl+`) and Deck's `BrainTerminalEmbed` had the key swallowed by the pane gate and declined by the popover gate — a dead key, or Rich Input aimed at a different pty. A new `ownsComposeShortcut` option (true from `Terminal.tsx`, defaulting to false) keeps those surfaces encoding 0x07 exactly as 3.51.0 did. Resolving the popover against the pty that produced the event was the alternative, and was rejected: the popover is rendered by the active leaf's toolbar, so it would open over the wrong pane.

  The first cut of this wired ownership into the pane gate only, and the live dogfood caught what that leaves: from the floating pane, one Ctrl+G still opened Rich Input over the background active leaf while the floating pty received nothing — on Windows, the original complaint reproduced inside the floating pane. Ownership now lives on the DOM and both gates read the same marker: `useTerminal` stamps every container with `data-terminal-pty` and adds `data-compose-owner` only where the option is set, and `composeOwnerHost(target)` resolves a keydown to `{ ptyId, owns }`. The document gate acts only when the keydown came from an owning terminal whose pty **is** the active leaf. A keydown from no terminal at all (focus on `<body>`) keeps the old behaviour — the binding never required terminal focus — and the toolbar-owned-input exemption is unchanged.
- **Auto-repeat is a decision, not an accident.** The predicate accepts repeats, so both gates agree the key belongs to the binding; the popover gate declines only to *toggle* on one. A held Ctrl+G toggles once and is then inert — a deliberate non-repeating chord — rather than producing neither a byte nor an action.
- `useComposeShortcut` picks up the `inspectModeActive` early-out `useKeyboard` applies to every global shortcut, and the pane gate checks the same flag so the key is not merely swallowed.
- `WMUX_KEYMAP` gains the `Ctrl+G` row it never had, **advertised** with a `settings.sc.richInput` label. Advertised rows are exactly the rows `SettingsPanel` renders with a disable toggle, and the disabled gate is wired on both sides (`useComposeShortcut` yields; `useTerminal`'s disabled branch writes the control byte), so switching it off hands Ctrl+G straight back to the pane. The row also reserves its accelerator, so the app-menu drift test now covers it.

## The binding stays Ctrl+G — the reporter's expectation is declined deliberately

The reporter expected "only Ctrl+Shift+G opens Rich Input". That is the *accidental* spelling — the one the superset bug created — not the designed one:

- The toolbar renders a literal `Ctrl G` kbd chip next to the Rich Input button.
- `settings.agentToolbarShowDesc` says "Ctrl+G always works" in all 23 locales.
- The design spec and the implementation plan for the toolbar both specify Ctrl+G, and the hook, `ToolbarHost`, `ComposeHost` and `AgentToolbar` comments all say ⌘G / Ctrl+G.
- **Ctrl+Shift+G is already `clearMultiview`** (`useKeyboard`, `WMUX_KEYMAP`), so Rich Input cannot own it without both actions firing on one keypress.

One case the built-in still loses, worth knowing: a user who has bound a **custom keybinding** to Ctrl+G keeps winning via `useKeyboard`'s `stopImmediatePropagation`, so Rich Input silently never opens for them. Their explicit rebind beating the built-in is the right precedence — it is just silent.

The real cost of keeping Ctrl+G is that it no longer reaches the pane: Claude Code's external-editor key and readline's abort. That is what the new Settings → Shortcuts toggle is for — turn the shortcut off and Ctrl+G behaves exactly as it did in 3.51.0, with no other binding touched.

## Tests

- `src/renderer/components/AgentToolbar/__tests__/useComposeShortcut.test.tsx` — new. Ctrl+G toggles and consumes the key; **Ctrl+Shift+G does not**; Ctrl+Alt+G does not; bare G does not; the IME `key === 'Process'` + `code === 'KeyG'` path does, but **not while composing**; a disabled Ctrl+G yields; inspect mode yields; no focused terminal surface yields; key repeat does not flap. Plus the keymap sweep on both platforms. macOS block: ⌘G toggles, Ctrl+G does not, ⌘+Shift+G does not.
- `src/renderer/terminal/__tests__/composeChord.test.ts` — new. The predicate directly: Shift rejected (Ctrl+Shift+G), Alt and Meta rejected, unshifted capital G accepted (Caps Lock), the physical `KeyG` fallback under a jamo / `Process` key, composing deferred, macOS ⌘G accepted with literal Ctrl+G and Ctrl+⌘+G rejected, repeat accepted — plus a sweep of **every `WMUX_KEYMAP` row on both platforms**, with ⌘ substituted for the cmdOrCtrl family the way `useKeyboard` dispatches it and `code` filled in from the letter.
- `src/renderer/hooks/__tests__/useTerminal.composeBubble.test.ts` — new. jsdom cannot run xterm's custom key handler, so the wiring is pinned at the source like the neighbouring `macCtrlPassthrough` / `ctrlLetterEncoding` locks, but kept to the few lines that matter now that the predicate is tested on its own: the branch is `ownsComposeShortcut && isComposeChord(…)`, it precedes the catch-all encoder, `g` / `KeyG` stay out of the shared allowlists, and the disabled-shortcut branch precedes it **and writes the control byte inside that same branch**. Plus the ownership wiring in all three consumers (`Terminal.tsx` opts in; `FloatingPane.tsx` and `BrainTerminalEmbed.tsx` do not; the option defaults to false) and a behavioural check that a non-owning terminal still resolves Ctrl+G to `0x07` — the floating-pane case.
- `src/shared/__tests__/keymap.test.ts` — Ctrl+G is in `ADVERTISED_SHORTCUTS` (the list SettingsPanel maps over) with the right label key and `literalCtrl: false`; it reserves its accelerator; disabling it matches a real Ctrl+G keydown by `key` and by physical `code`, does **not** match Ctrl+Shift+G, and on macOS matches ⌘G but not literal Ctrl+G.
- `src/renderer/i18n/__tests__/advertisedShortcutsLocale.test.ts` — new. Every advertised row has an English string (a missing one would render the raw key in the Settings list, in every language), and the Rich Input label is present in ko/zh too.

### Locale convention followed

`settings.sc.richInput` was added to **en (authoritative), pl, ko, zh**. Per #997 and the note in `scripts/locale-drift-report.mjs`, those are the maintained locales — pl is gated by `plLocaleCoverage.test.ts`, ko/zh are maintained by hand — while the other 20 are an explicitly accepted stalled gap (ko is at 1159/1607 keys, most locales at 377) that falls back to English at runtime. No translation was invented: each value reuses that locale's existing "Toggle …" phrasing from `settings.sc.toggleNotifications` and its own `toolbar.richInput` term.

Gates: `npx tsc --noEmit` clean; eslint on the changed files clean (the 3 remaining `no-empty-function` errors in `useTerminal.ts` are pre-existing on `main`); full `npx vitest run` → 15580 passed, 1 failed — `generateNotices` only, the known spurious failure under a symlinked `node_modules`.

Fixes #1280

## Summary by CodeRabbit

- **New Features**
  - Added the Rich Input shortcut to **Settings → Shortcuts**, with localized labels and an option to disable it.
  - Rich Input now responds to **Ctrl+G on Windows/Linux** and **⌘G on macOS**.
  - The shortcut is handled only by the active pane terminal; floating panes and embedded terminals retain their existing behavior.

- **Bug Fixes**
  - Prevented duplicate toggles from held keys.
  - Suppressed the shortcut during IME composition and inspect mode.
  - Improved handling across keyboard layouts and modifier combinations.

### DIFF
--- a/changelog.d/1286.md
+++ b/changelog.d/1286.md
@@ -9,7 +9,7 @@
   Holding the key toggles once rather than repeating, and it stays out of the
   way while an IME composition is open. In the floating pane (Ctrl+`) and the
   Deck brain terminal, where Rich Input does not apply, Ctrl+G remains a plain
-  `^G`. On macOS the binding is ⌘G, and Ctrl+G stays a plain readline byte.
+  `^G` and no longer opens a popover over whichever pane happens to be active. On macOS the binding is ⌘G, and Ctrl+G stays a plain readline byte.
 
 ### Added
 
--- a/changelog.d/1286.md
+++ b/changelog.d/1286.md
@@ -7,3 +7,10 @@
   combination that included Ctrl and the letter G. It now matches exactly, and
   when it fires the key is consumed instead of being handed to the pane too.
   On macOS the binding is ⌘G, and Ctrl+G stays a plain readline byte.
+
+### Added
+
+- **Rich Input's shortcut can be switched off.** Settings → Shortcuts now
+  lists Ctrl+G with the same toggle the other built-ins have. Turn it off and
+  Ctrl+G goes to the pane instead — the agent's external editor, readline's
+  abort — with no other binding changed.
--- a/changelog.d/1286.md
+++ b/changelog.d/1286.md
@@ -6,7 +6,10 @@
   multiview — opened Rich Input as well, because the binding accepted any
   combination that included Ctrl and the letter G. It now matches exactly, and
   when it fires the key is consumed instead of being handed to the pane too.
-  On macOS the binding is ⌘G, and Ctrl+G stays a plain readline byte.
+  Holding the key toggles once rather than repeating, and it stays out of the
+  way while an IME composition is open. In the floating pane (Ctrl+`) and the
+  Deck brain terminal, where Rich Input does not apply, Ctrl+G remains a plain
+  `^G`. On macOS the binding is ⌘G, and Ctrl+G stays a plain readline byte.
 
 ### Added
 
--- a/changelog.d/1286.md
+++ b/changelog.d/1286.md
@@ -1,0 +1,9 @@
+### Fixed
+
+- **Ctrl+G opens Rich Input and nothing else.** Since 3.52.0 it did two things
+  at once: the pane got a `^G` (or the agent CLI opened your external editor)
+  *and* the Rich Input popover appeared. Ctrl+Shift+G — which clears a
+  multiview — opened Rich Input as well, because the binding accepted any
+  combination that included Ctrl and the letter G. It now matches exactly, and
+  when it fires the key is consumed instead of being handed to the pane too.
+  On macOS the binding is ⌘G, and Ctrl+G stays a plain readline byte.
--- a/src/renderer/components/AgentToolbar/__tests__/useComposeShortcut.test.tsx
+++ b/src/renderer/components/AgentToolbar/__tests__/useComposeShortcut.test.tsx
@@ -16,6 +16,7 @@ import { createElement, act } from 'react';
 import { createRoot, type Root } from 'react-dom/client';
 import { WMUX_KEYMAP } from '../../../../shared/keymap';
 import { useComposeShortcut } from '../useComposeShortcut';
+import { TERMINAL_PTY_ATTR, COMPOSE_OWNER_ATTR } from '../../../terminal/composeChord';
 
 const setToolbarPopover = vi.fn();
 let state: Record<string, unknown>;
@@ -37,10 +38,25 @@ function Probe() {
   return null;
 }
 
-function press(init: KeyboardEventInit): void {
+function press(init: KeyboardEventInit, from?: HTMLElement): void {
   act(() => {
-    document.dispatchEvent(new KeyboardEvent('keydown', { bubbles: true, cancelable: true, ...init }));
+    const e = new KeyboardEvent('keydown', { bubbles: true, cancelable: true, ...init });
+    (from ?? document).dispatchEvent(e);
   });
+}
+
+/**
+ * A terminal container as useTerminal stamps it: every terminal carries its
+ * ptyId, only the pane-surface one carries the owner marker.
+ */
+function mountTerminal(ptyId: string, owns: boolean): HTMLElement {
+  const host = document.createElement('div');
+  host.setAttribute(TERMINAL_PTY_ATTR, ptyId);
+  if (owns) host.setAttribute(COMPOSE_OWNER_ATTR, '');
+  const textarea = document.createElement('textarea');
+  host.appendChild(textarea);
+  document.body.appendChild(host);
+  return textarea;
 }
 
 /**
@@ -85,6 +101,7 @@ beforeEach(() => {
 afterEach(() => {
   act(() => { root.unmount(); });
   container.remove();
+  document.querySelectorAll(`[${TERMINAL_PTY_ATTR}]`).forEach((el) => el.remove());
 });
 
 describe('useComposeShortcut modifier matching (#1280)', () => {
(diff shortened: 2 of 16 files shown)


openwong2kim:

Fixes #1280 two ways: `useComposeShortcut` matches Ctrl+G exactly (Shift/Alt disqualified), and `useTerminal` bubbles g/KeyG so the pane stops receiving `^G`. Risk: Ctrl+G no longer reaches the pane (Claude's external-editor key) and the new keymap row isn't togglable in Settings yet. P2 — direct fix of a P2 regression, well tested.

openwong2kim:

`2cbeea8e`. The pane runs a raw `setRawMode` byte dumper, so anything that leaks
to the PTY — a BEL `07`, anything — is printed. Keys are **real OS keystrokes**
(`SendKeys` → `WM_KEYDOWN`), not synthesised DOM events.

| chord | Rich Input | bytes reaching the pane |
|---|---|---|
| Ctrl+G | **opens** | **none** |
| Ctrl+Shift+G | does not open | none |

Both halves of the report are fixed: the binding fires on exactly Ctrl+G, and
when it fires the key stops at the pane — no `^G` echo, no `07`. Ctrl+Shift+G no
longer rides the Rich Input binding.

Screenshot at the moment Ctrl+G was pressed: the Rich Input popover is open and
the terminal still shows only `DUMPER-READY-plain`, with no `^G` on the line and
no `BYTES[07]` from the dumper.

The Settings entry for the toggle is present in the built UI.

Environment note: this is Windows 11 / PowerShell 5.1, the same combination as
the original report (#1280).

Issue #1288, opened 2026-09-12
Perf gate is red repo-wide: coldStart baseline is 7 weeks stale against a newer runner image


openwong2kim:
The perf gate now fails on unrelated PRs for two independent reasons, so it no longer distinguishes a regression from the environment. Needs an owner call.

## 1. `coldStart.firstPtyDataMs` — stale baseline vs a newer runner image

`bench/baseline-ci.json` records `coldStart.best.firstPtyDataMs = 1134` and was last touched on **2026-07-27** (`e5a4f182`). The current windows runner (image `win25-vs2026/20260907.229`) measures **2629 / 2639 / 2673 ms across three boots** — about 2.3× the baseline, consistently.

Reproduced on at least two unrelated branches on 2026-09-12: `fix/1234-atlas-guard-recovery` (run at 09:12, where it was the only failure) and `fix/1280-ctrl-g-rich-input` (run 34685338042, 09:14). Neither touches anything that runs before first PTY data.

Two readings, and they need different responses:
- The runner image genuinely got slower (or its disk/CPU profile changed) → refresh the baseline against the current image, and record which image it was measured on so the next shift is attributable.
- Something in the repo really did slow cold start by 1.5 s since late July → that deserves its own investigation, and re-baselining would bury it.

Worth deciding deliberately: a blind refresh silently accepts a 2.3× move.

## 2. `frameBudget.N8.frameDeltaMs.p95` — vsync-quantized noise

Same run, and the numbers refute a real regression internally:

| scenario | baseline p95 | measured p95 |
|---|---|---|
| N4 | 15.7 | 15.7 |
| **N8** | 15.7 | **46.9** |
| N16 | — | 31.3 |

Sixteen flooded panes cannot be cheaper than eight, so N8 > N16 is incoherent as a rendering cost. Every value is an exact multiple of the ~15.6 ms vsync tick (15.6 / 31.3 / 46.9), so p95 can only move in whole dropped frames and a single noisy sample shifts it by a full tick. Previously tracked in #1274; captured here too because the two causes now stack.

## Consequence

Every PR inherits a red `bench-confirm` from main, which trains everyone to ignore the perf gate — including for a real regression. Suggested shape: re-measure the baseline on the current image with the image tag recorded, and make N8 advisory (or median-based, or sampled enough that one dropped frame can't move it) until the quantisation is handled.

openwong2kim:
## [wmux-hermes] Triage Summary

[...]

Issue #1274, opened 2026-09-11
CI: four tests time out on slow runners and fail unrelated PRs


openwong2kim:
On 2026-09-11 the same four tests failed repeatedly across unrelated PRs (and once on main) purely by exceeding their timeout; every one passed on a re-run of the same commit. With GitHub's runners slow that day, each red run cost a ~15 min re-run.

| Test | Runner | Timeout | Observed | Seen on |
|---|---|---|---|---|
| `src/daemon/__tests__/DaemonSessionManager.test.ts` › honours a custom session.maxSessions from setConfig | macos-14 | 5000 ms | 5.9–7.3 s | #1251, #1249, #1273 |
| `scripts/__tests__/opencode-sync-render.runtime.test.mjs` › paints completed synchronized frames while OpenCode-style output remains active | ubuntu-22.04 | 30000 ms | 30.0 s | main 3c352278, #1238, #1273 |
| `src/main/git/__tests__/mergeSession.test.ts` › a conflicting merge returns the unmerged file list, a clean merge returns an empty list | windows-latest | 5000 ms | 6.0 s | #1241 |
| `src/main/ipc/handlers/__tests__/diff.handler.test.ts` › 미커밋 3파일(수정2+untracked1)을 파일 트리·numstat로 반환 | windows-latest (validate) | — | 10.2 s | #1251 |

None of these PRs touched the code under test. The maxSessions case spawns real sessions up to the cap, and the git cases shell out to real `git`, so their wall time scales with runner load rather than with the code.

Suggested direction: give the process-spawning tests an explicit per-test timeout sized for the slowest runner (or shrink the work, e.g. a lower cap in the maxSessions case), and have the render runtime test assert on frame completion events rather than racing a fixed 30 s wall clock.

Separately, the perf gate's `frameBudget.N8` on windows-latest went bimodal the same day (p50 15.6 ms vs 31.3 ms on identical code, #1265 vs main bfbd553e); worth watching before tightening that threshold.

openwong2kim:
## [wmux-hermes] Triage Summary

Four timing-sensitive tests (maxSessions, opencode-sync-render, mergeSession, diff.handler) exceed fixed timeouts on loaded runners and fail unrelated PRs. The risk is CI noise and wasted re-runs, not user-facing breakage. P3 — real but intermittent; explicit per-test timeouts and event-based assertions in the render test will fix it.

openwong2kim:
[...]

PR #1290, opened 2026-09-12
test: fix the seven timing-sensitive CI flakes

openwong2kim:
Fixes the seven timing-sensitive test failures tracked in #1274. All of them failed on loaded GitHub runners, on PRs that never touch the code under test, and all of them passed on rerun — today they cost a ~15-minute rerun on almost every PR.

Test-only change. No production file is modified, no assertion is weakened, and the global vitest timeout in `vitest.config.ts` is untouched.

## Per-test cause and fix

**1. `src/daemon/__tests__/DaemonSessionManager.test.ts` › honours a custom session.maxSessions from setConfig** (5 s limit, measured 5.4–7.3 s on macos-14)

The three-session test itself takes **4 ms in isolation**. The cost belonged to its neighbour: `throws an actionable error when the session cap is reached` spawned 200 mock sessions to reach the *default* ceiling, and that garbage was collected during the next test — measured as 283 ms locally, which is exactly the 5.4–7.3 s seen on a loaded macos-14 runner. The ceiling is read from config (the no-config path falls back to `createDefaultConfig`), so the cap test now configures a cap of 4 and asserts the full user-facing message; the default-200 contract is pinned by an explicit assertion on `createDefaultConfig()` rather than by spawning 200 PTYs. Work shrunk 50×, no coverage lost. The two near-duplicate cap tests (the rewritten one and the adjacent configurable-cap one) are folded into a single test, since they are the same branch.

**2. `scripts/__tests__/opencode-sync-render.runtime.test.mjs` › paints completed synchronized frames…** (30 s wall clock, ubuntu + windows)

Three problems. The probe slept a fixed 80 ms after the warm-up frame and another 80 ms before reading the viewport, so a runner whose `requestAnimationFrame` callback landed later than that read a stale screen and failed `finalFrameVisible` — every wait is now driven by xterm's `onRender` event, resolving on the first render that shows the frame (the deadline only bounds the failure case). Separately, the wall clock here is a cold `chromium.launch()` of the system Chrome/Edge channel plus the first WebGL context: measured at **31.5 s** in a full parallel suite run locally. File budget raised to 120 s with that measurement in the comment, and `chromium.launch` now carries an explicit 60 s timeout so the one unbounded step fails with a clear Playwright error instead of eating the whole budget. Third, `waitForVisible` leaked an `onRender` listener every round the 50 ms timer won the race (~200 over the 10 s deadline); the disposable is held and the losing side disposed.

**3. `src/main/ipc/handlers/__tests__/deck.handler.loop.test.ts:1169` nested fan-out brain parking** (`expected +0 to be 1`)

There is no debounce on this path: `deck.handler.ts:1753` skips `coalescer.push` for a task workspace and calls `routeWorkerEventToOwner` inline. The only async step is `TaskLedger.recordOrphanedEvent` (serialize + jsonl append, `taskLedger.ts:534`), whose rejection is swallowed by a `.catch(console.warn)` in `taskLedgerHost.ts:80` — so a slow or failed append surfaces here only as the poll never seeing the event. `vi.waitFor`'s *default* timeout is 1000 ms, which a loaded Windows runner's fs append can outlast. The wait is now 10 s at a 25 ms interval and the test past it; locally the park lands in ~36 ms, so a healthy run is unaffected.

**4. Real-process files — explicit, named, justified budgets**

These exec real `git` and real shells, so their runtime is process-spawn cost and scales with runner load rather than with code speed. Per-suite timeouts are applied to the spawning suites only; pure-parser suites in the same files keep the 5 s default.

- `src/main/git/__tests__/mergeSession.test.ts` — real `git init/commit/worktree add/merge` against a per-test temp repo (each test mutates its repo, so sharing a fixture would break isolation). `GIT_PROCESS_TIMEOUT_MS = 30_000` on the 5 git-driven suites.
- `src/main/ipc/handlers/__tests__/diff.handler.test.ts` — every test builds two temp repos and shells out to `git diff/apply/apply --check`. 30 s on all 13 suites.
- `src/main/worktask/__tests__/FanOutService.test.ts` — the assertion *is* a real `bash -c` / `sh -c` round-trip proving the launch string survives an alias-expanding shell; irreducible. 30 s on the two shell-spawning suites (`workerLaunchCommand (F15)` and its sibling `buildInitialCommand (§4 D4)`, the same latent flake).

**5. `src/shared/daemon/__tests__/daemonScriptIdentity.test.ts` › refuses an unrelated process whose argv merely ends in daemon/index.js**

On win32 `getProcessArgv` shells out to `powershell … Get-CimInstance Win32_Process` with `timeout: 5000` and returns `null` on any failure. Under `definitiveOnly: false` a null cmdline next to a non-mismatching image is documented and intended to proceed to SIGKILL — so on a loaded runner the test killed its own sleeper, and the old `toBe(false)` was racing WMI latency rather than stating anything about the matcher. `killVerifiedDaemonPid` is unchanged.

The first revision of this PR asserted the refusals under `definitiveOnly: true`; review correctly flagged that as a tautology on exactly the platform that flaked — production returns false on *any* indeterminate probe (`daemonLauncherCore.ts:1477/1488`) before `argvIdentifiesDaemonScript` is ever called, so the #1025 entry-position guard would no longer be exercised there, and the repro would no longer be asserted in the mode `killDaemonByPidFile` actually uses. So the three refusal assertions keep `definitiveOnly: false`, and only the win32 CIM call is stubbed — to the sleeper's real command line — via the same `vi.doMock` technique as the new indeterminate test. The matcher therefore decides the outcome on every platform and probe latency decides nothing; every other `execFileSync` call, the tasklist / `ps` image lookup included, passes through to the real implementation, and on macOS/Linux the argv probe is untouched (it is fast and never flaked).

The indeterminate branch keeps its own test with the probe fully stubbed (`definitiveOnly: false` → proceeds, `definitiveOnly: true` → refuses), and now SIGKILLs both sleepers in its `finally` so a failed first expect cannot orphan 30 s sleepers on the runner.

## Triage sweep

Applied the issue's rule once across the whole suite (`vitest run --reporter=verbose`, 15,607 tests): any test whose runtime under a full parallel run is within ~2× of its per-test timeout.

It found an eighth flake that **actually failed locally**: `src/renderer/stores/slices/__tests__/channelsSlice.test.ts` › `is reachable through the composed store via channels selector` — 348 ms in isolation, **5022 ms** in the parallel run, i.e. over the 5 s default. The assertions are instant; the cost is the one-off Vite transform of the entire composed renderer store graph pulled in by `await import('../../index')`. Fixed properly rather than budgeted: the import is hoisted into a `beforeAll`, so that one-off cost is no longer charged to the test body. The test now measures 0 ms and keeps the 5 s default.

Remaining candidates, reported not fixed (runtime under a full parallel suite run vs the 5 s default, no explicit timeout on the test):

| Test file | Slowest test | Runtime | Margin |
|---|---|---|---|
| `src/renderer/__tests__/typeScaleLint.test.ts` | fires on a string className | 4566 ms | 1.1× |
| `src/main/worktask/__tests__/fanoutEnvironment.test.ts` | kills the whole process tree on timeout | 4221 ms | 1.2× |
| `src/renderer/stores/slices/__tests__/notificationSlice.test.ts` | decrements the evicted surface's index | 4188 ms | 1.2× |
| `src/main/pipe/__tests__/RpcDispatchProvenance.sourceInvariant.test.ts` | keeps externalWire writes inside the envelope | 4023 ms | 1.2× |
| `src/shared/__tests__/humanRhythm.test.ts` | stays inside the clamp on every draw | 3211 ms | 1.6× |
| `src/daemon/web/__tests__/WebTerminalServer.test.ts` | collapses a resize storm | 3188 ms | 1.6× |
| `src/main/ipc/handlers/__tests__/worktree.handler.merge.test.ts` | status — conflict path | 2635 ms | 1.9× |
| `src/mcp/playwright/__tests__/PlaywrightEngine.test.ts` | getPage auto-opens with the session | 2516 ms | 2.0× |
| `src/mcp/playwright/__tests__/lazyPlaywright.test.ts` | loads playwright-core on first call | 2103 ms | 2.4× |

Also noted inside a file this PR touches: `FanOutService.test.ts` › `fan-out worker first run (A-1)` › "reports a worker still stuck on a first-run screen instead of calling it working" — 1507 ms against the 5 s default. Not fixed, because shrinking it would mean changing the watcher's real poll interval (production) or faking timers (which changes what the test proves).

## Runtime table (macOS, local)

| File | Slowest test | Before | After | Per-test timeout in effect |
|---|---|---|---|---|
| `DaemonSessionManager.test.ts` | maxSessions + cap (now one test) | 283 ms + 43 ms / 200 spawns | 6 ms / 3 spawns | 5 s (default) |
| `opencode-sync-render.runtime.test.mjs` | the probe | 31.5 s (parallel suite) | 1.6–1.9 s standalone | 120 s, launch 60 s |
| `deck.handler.loop.test.ts` | nested fan-out parking | 53 ms (waitFor capped at 1 s) | 36 ms | 15 s (wait 10 s) |
| `deck.handler.loop.test.ts` | fast reject-and-requeue | 3244 ms | 3244 ms | 10 s (pre-existing) |
| `mergeSession.test.ts` | clean merge → Land round-trip | 1125 ms | 340 ms warm | 30 s |
| `diff.handler.test.ts` | alreadyApplied hunk rejection | 2406 ms | 660 ms warm | 30 s |
| `FanOutService.test.ts` | same-shell form, aliased claude | 8 ms local / 7.4 s windows | 8 ms | 30 s |
| `daemonScriptIdentity.test.ts` | #1025 repro | 6.3 s windows (failing) | 335 ms | 15 s |
| `daemonScriptIdentity.test.ts` | requirement 1, executed | 350 ms | 317 ms | 15 s |
| `daemonScriptIdentity.test.ts` | requirement 3, executed | 647 ms | 628 ms | 30 s |
| `daemonScriptIdentity.test.ts` | indeterminate branch (new) | — | 612 ms | 15 s |
| `channelsSlice.test.ts` | composed store selector | 5022 ms (failing) | 0 ms (import in `beforeAll`) | 5 s (default) |

File totals after the change, measured over 3 consecutive local runs each: `DaemonSessionManager` 0.35–0.41 s; `deck.handler.loop` 5.86–5.92 s; `daemonScriptIdentity` 4.36–4.40 s; `channelsSlice` 0.48–0.54 s; `mergeSession` 1.69–1.74 s; `diff.handler` 8.88–9.18 s; `FanOutService` 1.87–1.90 s; `opencode-sync-render` 1.75–2.19 s. All green on every run.

Note on comparing numbers: local figures are a single file run serially on macOS; CI figures are a file sharing a runner with parallel vitest workers, on Windows where process spawn costs an order of magnitude more. The budgets are sized for the latter.

## Gates

- `npx tsc --noEmit` — clean.
- `npx eslint` on all changed files — 0 new problems. The one error (`FanOutService.test.ts:540 no-empty-function`) and all warnings are byte-identical at HEAD.
- Every touched test file run 3× and green every time; the daemon and git ones included. Full suite (before the fixes): 15,531 passed / 2 failed — `generateNotices` drift, which needs a real `npm ci` rather than a symlinked `node_modules`, and the `channelsSlice` timeout this PR fixes.

No changelog fragment: CI-only, no user-facing change.

Refs #1274 — deliberately does not close it, since the `frameBudget.N8` perf-gate half (#1288) is separate and unaddressed here.

### DIFF
--- a/scripts/__tests__/opencode-sync-render.runtime.test.mjs
+++ b/scripts/__tests__/opencode-sync-render.runtime.test.mjs
@@ -45,9 +45,15 @@ const pageHtml = `<!doctype html>
   // another 80 ms before reading the screen, so a loaded CI runner whose
   // requestAnimationFrame callback landed later than that read a stale
   // viewport and failed on finalFrameVisible.
-  const nextRender = () => new Promise((resolve) => {
-    const sub = term.onRender(() => { sub.dispose(); resolve(); });
-  });
+  // Returns the promise AND its disposer, so a caller that loses a
+  // Promise.race can drop the listener instead of leaking one per round.
+  const nextRender = () => {
+    let sub;
+    const done = new Promise((resolve) => {
+      sub = term.onRender(() => { sub.dispose(); resolve(); });
+    });
+    return { done, dispose: () => sub.dispose() };
+  };
 
   const visibleText = () => {
     const buffer = term.buffer.active;
@@ -65,7 +71,19 @@ const pageHtml = `<!doctype html>
     const until = Date.now() + deadlineMs;
     while (!visibleText().includes(needle)) {
       if (Date.now() >= until) return false;
-      await Promise.race([nextRender(), new Promise((r) => setTimeout(r, 50))]);
+      const render = nextRender();
+      let timer;
+      try {
+        await Promise.race([
+          render.done,
+          new Promise((r) => { timer = setTimeout(r, 50); }),
+        ]);
+      } finally {
+        // Whichever side won, drop the other: ~200 rounds fit inside the
+        // deadline, and each one used to leave its onRender listener behind.
+        clearTimeout(timer);
+        render.dispose();
+      }
     }
     return true;
   };
@@ -74,7 +92,7 @@ const pageHtml = `<!doctype html>
     const warmedUp = nextRender();
     await write('\\x1b[?2026h' + frameBody(-1));
     await write('\\x1b[?2026l');
-    await warmedUp;
+    await warmedUp.done;
 
     let renders = 0;
     const sub = term.onRender(() => { renders++; });
@@ -114,6 +132,10 @@ it('paints completed synchronized frames while OpenCode-style output remains act
     browser = await chromium.launch({
       channel,
       args: ['--enable-unsafe-swiftshader'],
+      // Bound the one unbounded step. Everything after this — the probe's
+      // waits — is capped at 10 s, so a hung launch should fail here with a
+      // clear Playwright error rather than burn the whole file budget.
+      timeout: 60_000,
     });
     const page = await browser.newPage();
     await page.goto(origin);
--- a/scripts/__tests__/opencode-sync-render.runtime.test.mjs
+++ b/scripts/__tests__/opencode-sync-render.runtime.test.mjs
@@ -40,10 +40,41 @@ const pageHtml = `<!doctype html>
     return '\\x1b[H\\x1b[2J' + rows.join('\\r\\n');
   };
 
+  // #1274: every wait in this probe is driven by an xterm event, not by a
+  // fixed sleep. The old version slept 80 ms after the warm-up frame and
+  // another 80 ms before reading the screen, so a loaded CI runner whose
+  // requestAnimationFrame callback landed later than that read a stale
+  // viewport and failed on finalFrameVisible.
+  const nextRender = () => new Promise((resolve) => {
+    const sub = term.onRender(() => { sub.dispose(); resolve(); });
+  });
+
+  const visibleText = () => {
+    const buffer = term.buffer.active;
+    let text = '';
+    for (let y = 0; y < term.rows; y++) {
+      text += buffer.getLine(buffer.viewportY + y)?.translateToString(true) ?? '';
+    }
+    return text;
+  };
+
+  // Resolve as soon as the wanted frame is on screen. The deadline only
+  // bounds the failure case (the assertion then reports a stale viewport);
+  // a healthy run returns on the very next render event.
+  const waitForVisible = async (needle, deadlineMs) => {
+    const until = Date.now() + deadlineMs;
+    while (!visibleText().includes(needle)) {
+      if (Date.now() >= until) return false;
+      await Promise.race([nextRender(), new Promise((r) => setTimeout(r, 50))]);
+    }
+    return true;
+  };
+
   window.runProbe = async () => {
+    const warmedUp = nextRender();
     await write('\\x1b[?2026h' + frameBody(-1));
     await write('\\x1b[?2026l');
-    await new Promise((resolve) => setTimeout(resolve, 80));
+    await warmedUp;
 
     let renders = 0;
     const sub = term.onRender(() => { renders++; });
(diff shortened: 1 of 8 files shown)


openwong2kim:

Test-only PR fixing the seven timing-sensitive CI flakes from #1274 across eight test files. Fixed sleeps become event-driven waits, a 200-session spawn shrinks to a configured cap of 4, and real-git/shell suites get explicit budgets — no production code touched and no assertion weakened. P3: CI-reliability change only, but it removes a ~15-minute rerun from almost every PR.

PR #1291, opened 2026-09-12
ci(perf): re-bless CI baseline after windows-latest fleet slowdown

openwong2kim:
## Why

Since 2026-09-12 07:08Z the Perf workflow fails on main and on every open PR, flapping around the gate threshold. This is not a product regression — evidence below — but a uniform slowdown of the `windows-latest` runner fleet, so the blessed baseline (cut 2026-07-27) no longer describes what the runners measure.

## Investigation

- **Same-tree rerun fails.** A rerun of `e1cd04d7` (the last main commit whose Perf run was green, 2026-09-12 06:55Z, best firstPtyData 1282 ms) measures best 2410/2542 ms later the same day (runs 34702798103, 34703562118). The failure reproduced on a second machine both times (`bench-confirm`), so all three designed controls (best-of-N, same-runner re-run, fresh-runner confirmation) saw the shift.
- **Trend timeline.** `bench-history` shows the jump between two consecutive main runs nine minutes apart: 06:59Z best 1282 ms → 07:08Z 2012 ms → 11:53Z 3252 → 12:18Z 2620 → 15:35Z 2425. Runs that stayed green after the shift measured 1542/1815 ms — the whole distribution moved up and straddles the 2134 ms gate threshold, which is why verdicts flip per run.
- **Not the runner image.** Green and red runs report the same image provisioner version (20260828.587).
- **Not the daemon.** Boot-phase attribution is unchanged vs. the baseline (module-eval/lock/config/recovery/pipe all within tens of ms; daemon ready ~1060 ms in both eras). The entire delta is the renderer `ready tail` (renderer-load-triggered → ready-end): ~700 ms → ~1780 ms, plus `frameBudget.N8` stepping from 1 to 3 frame intervals — the signature of slower disk/GPU on the host, matching the known fleet behavior class of actions/runner-images#12386 / #8755.

## What this PR does

Blesses `bench/baseline-ci.json` from the perf artifact of run 34703562118 (same-tree rerun of `e1cd04d7`), per the baseline policy's deliberate-PR path: the numbers moved for an external reason, the cause is documented above, and the gate is currently blocking all merges on a coin flip.

`coldStart.best.firstPtyDataMs` gate threshold moves 2134 → ~3813 ms (best 2542) and `frameBudget.N8` allows 4 frame intervals at p95 (blessed 46.9 + 1 frame margin). That is a real loss of sensitivity and it is deliberate: it is temporary.

## When the fleet recovers

`perf-compare` flags an improvement when `current < baseline * 0.8` — once runners measure ~2000 ms or below again, every run will print "consider refreshing baseline". Re-bless then from a fresh main artifact (the same one-command operation as this PR) and reference this PR in reverse.

## Summary by CodeRabbit

* **Performance**
  * Updated benchmark baselines with refreshed measurements for startup time, memory usage, input latency, rendering performance, WebGL recovery, and background processing.
  * Added coverage for additional resource and process metrics, including tree handling and application-level measurements.
  * No changes to public features or user-facing functionality.

### DIFF
--- a/bench/baseline-ci.json
+++ b/bench/baseline-ci.json
@@ -3,17 +3,17 @@
   "meta": {
     "tool": "perf-bench.mjs",
     "mode": "ci",
-    "appVersion": "3.20.0",
-    "commit": "2395e5e",
-    "startedAt": "2026-07-10T08:42:16.313Z",
-    "finishedAt": "2026-07-10T08:44:49.002Z",
+    "appVersion": "3.53.0",
+    "commit": "e1cd04d",
+    "startedAt": "2026-09-12T15:54:55.245Z",
+    "finishedAt": "2026-09-12T15:57:20.715Z",
     "platform": "win32",
     "arch": "x64",
     "cpuModel": "AMD EPYC 7763 64-Core Processor                ",
     "cpuCount": 4,
     "totalMemBytes": 17174360064,
     "osRelease": "10.0.26100",
-    "nodeVersion": "v22.23.1",
+    "nodeVersion": "v22.23.2",
     "config": {
       "coldRuns": 3,
       "inputSamples": 80,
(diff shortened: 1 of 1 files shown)


openwong2kim:
Re-blesses `bench/baseline-ci.json` from a same-tree rerun artifact to unblock the perf gate after the windows-latest fleet slowdown tracked in #1288. Single JSON file, no product code; the looser thresholds are documented as temporary. P2 — the gate is currently red on every open PR, so landing this promptly restores merge signal.

Issue #1305, opened 2026-09-14
WSL session recovery follow-ups from #1263


openwong2kim:
Follow-ups identified in the review of #1263 (WSL directory and Claude session recovery). None block that PR; each is small and independent.

## 1. No way out when a WSL exec pane's Linux directory is gone
A recovered WSL exec pane whose working directory was deleted now stays pending with **Retry connection**, instead of silently resuming in home. That is the right default, but as long as the directory is missing every Retry fails, and there is no way to start fresh.

**Proposal:** when the probe reports the directory as missing, offer an explicit "Start fresh in home" action that drops `--resume` and launches in the user's home directory, alongside Retry.

## 2. Pending recovery entries never expire
Suspended entries carrying `recoveryError` are exempt from the suspended-session TTL in `src/daemon/StateWriter.ts`. If a surface or workspace disappears without an explicit pane close, nothing ever promotes or closes the entry. `sessions.json` grows without bound, and every boot schedules a background retry for it.

**Proposal:** a long TTL for pending entries based on last activity (e.g. 30 days), or pruning after N consecutive boots with a failed promote and no client access.

## 3. `claude` installed via nvm is not on PATH in exec panes
The WSL exec bootstrap now runs Bash with `--noprofile --norc`, which keeps interactive startup output out of exec streams. As a side effect, a `claude` whose PATH comes from `.bashrc` (typical for nvm) can exit 127 in exec panes. `docs/how-to/wsl-session-recovery.md` documents the PATH requirement.

**Proposal:** resolve the agent binary's absolute path once, quietly (e.g. a login shell with output discarded), and pass that path to the exec launch instead of relying on PATH.

openwong2kim:
## 4. Local-mode create cancellation is only half closed

#1263 (commit `46dcc5d9`) added a per-id pending token to `PTYManager.createAsync`, so `dispose(id)` during the WSL probe window now cancels the spawn with `'PTY creation cancelled'`, the same error `disposeAll` uses.

No renderer-side caller learns the PTY id before the create resolves, though. Closing a surface while a local-mode WSL create is still in flight can therefore still leave an orphan PTY that nothing owns. This only affects local (non-daemon) mode with a WSL shell; daemon mode already guards this with `pendingCreates`.

**Proposal:** have the renderer pass a client-generated create id (or get a cancel handle back immediately) so closing the surface can call `dispose` on a create that is still pending.

openwong2kim:
[...]

PR #1308, opened 2026-09-14
fix(wsl): let unattempted recovery entries expire

thomas-pedersen:
## Summary

WSL panes skipped by the startup recovery limit were given a synthetic `recoveryError`, making them exempt from suspended-session expiry even though recovery had never failed. Preserve the normal expiry for these placeholders, including entries already saved with the old waiting marker, while retaining genuine failures for Retry or explicit close.

## Changes

- Allow pending recovery without an error; startup forwards only the persisted error instead of inventing one.
- Clear the exact legacy `WSL session is waiting to reconnect.` marker when rebuilding pending state. Preserve other errors and original activity timestamps.
- Add a manager/StateWriter regression covering unattempted, legacy-placeholder, and genuinely failed recovery across save/load cycles and expiry.
- Address the other two CodeRabbit findings from #1263: document mock no-op methods and statically import `homedir` for both cwd fallbacks.

### Fixed

- **WSL recovery placeholders expire normally.** Panes waiting for recovery, including those skipped by the startup recovery limit, no longer receive a synthetic failure that prevents suspended-session cleanup. The old saved waiting marker is also cleared. Genuine recovery failures keep their saved state for Retry or explicit close.

## Stability tier impact

n/a — no RPC, metadata schema, permission, or event changes. `recoveryError` is already optional.

## Test plan

- `npx tsc --noEmit` and `npm run build:daemon`: passed.
- `WMUX_REQUIRE_DAEMON_BUNDLE=1 npm test`: 15,604 unit tests and 197 runtime tests passed on Linux (30 unit and 43 platform/opt-in runtime tests skipped).
- Focused recovery/persistence/IPC checks: 70 tests passed. The legacy-marker regression was also observed failing before its fix.
- ESLint compared with `origin/main` across all four changed TypeScript files: all seven reported errors removed, no new errors. Existing unrelated lint errors remain (20 in daemon index, two in PTY IPC); this is not a claim that repository-wide lint passes.
- No manual Windows UI test for this follow-up; the new persistence test mocks PTY/WSL and runs on all CI platforms.

## Related issues / PRs

Follow-up to #1263 and its [post-merge CodeRabbit review](https://github.com/openwong2kim/wmux/pull/1263#pullrequestreview-5196469036). Related to #1305; this only addresses mistakenly non-expiring placeholders, not the broader policy for genuinely failed/orphaned recovery, PATH, or start-fresh behavior.

The session-limit impact in the review is overstated: pending recovery entries are separate from live PTYs and do not directly consume the live-session cap. The confirmed defect is indefinite metadata/scrollback retention.

## Reviewer checklist

- [x] CHANGELOG entry above is filled in.
- [x] Substrate contract docs: n/a, no surface change.
- [x] Tests cover the new behavior and regression case.
- [x] No secrets, tokens, `.env`, or unrelated binaries included.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WSL session recovery and cleanup for panes waiting to reconnect, including those skipped by recovery limits.
  * Removed outdated waiting markers so suspended sessions can expire normally.
  * Preserved genuine recovery failures, allowing them to remain available for retry or explicit closure.


openwong2kim:
Thanks @thomas-pedersen — clean follow-up to #1263. Dropping the synthetic `recoveryError` at the source, while clearing the legacy marker already on disk, is exactly the right cut: unattempted placeholders age out again, and genuine failures still keep their state for Retry. The save/load/expiry regression covering all three cases (including the `lastActivity` non-renewal check) is much appreciated, and so is the correction on the session-cap claim. Merging.

openwong2kim:
## Live dogfood, 2026-09-16 — not provable on this machine

This PR was on the Windows live-dogfood list for the 2026-09-16 pass (isolated
`WMUX_DATA_SUFFIX=-dog916` instance built from `origin/main` @ `2cbeea8e`,
Windows 11 26200). It could not be exercised: **the box has no WSL distribution
at all.**

```
> wsl.exe --list --verbose
Wsl/CallMsi/Install/REGDB_E_CLASSNOTREG

> Get-AppxPackage MicrosoftCorporationII.WindowsSubsystemForLinux
Version : 2.6.3.0   Status : Ok

> Get-ChildItem HKCU:\SOFTWARE\Microsoft\Windows\CurrentVersion\Lxss
(no distributions registered)
```

The store package is installed but the component is not registered and no distro
exists, so a WSL pane cannot be created, which means none of the claims here —
placeholder entries expiring, the legacy `WSL session is waiting to reconnect.`
marker being cleared on rebuild, genuine failures keeping their saved state for
Retry — can be observed live. Fixing that needs `wsl --install` plus a reboot on
the dogfood machine; not done here because it is a machine-wide change.

Everything else on the Windows list for this pass was exercised; this one and
#1263 are the two that were not. Flagging so it is not mistaken for "dogfooded
and fine."

PR #1313, opened 2026-09-14
fix(browser): calls without surfaceId stay on the calling agent's own tab

openwong2kim:
## The defect

Two agent panes in one workspace hold two MCP connections (the broker gives each accepted connection its own server and scope), but browser targeting had one answer for a call that named no `surfaceId`: **the newest browser surface in the workspace**.

With agents A and B in workspace `ws-1`:

1. A calls `browser_open` → surface `S1`, and starts working in it.
2. B calls `browser_navigate {url}` with no `surfaceId` → resolves to `S1`.
3. B has now navigated A's tab. A's next snapshot shows B's page, A re-navigates, and the two keep overwriting each other with nothing in either result saying why.

Two more paths made it worse. On the builtin backend `browser_open` reuses the workspace's *first* browser surface, so B's open navigated A's pane and handed B `S1` as its own handle. And the lanes disagreed among themselves: the engine took the workspace's *newest* surface while main's automation lease and the RPC fallback took its *oldest* live session, so one call could lease one tab and drive another.

## The fix

**A per-connection opener key.** Each MCP connection mints a random id once (`ConnectionScope`, with a module fallback for the single-child stdio server) and sends it with every open. Main records it against the surface, in memory, and answers callers with a *verdict* — mine, another's, or unclaimed — never with anyone's key.

**A per-connection pin.** The surface returned by `browser_open`, `browser_tabs new`, an auto-open or an adoption becomes that connection's default target. Passing an explicit `surfaceId` does not move it — only opening does, which is the contract the tool descriptions already stated.

**Fallback order** when `surfaceId` is omitted:

1. the connection's pin, while that surface still exists;
2. otherwise the newest surface **this** connection opened;
3. otherwise the newest surface **nobody** claims — restored after a restart, opened by a person, or opened before this change — which the connection then **adopts** (`browser.surface.adopt`), so the next connection does not land on it too;
4. otherwise nothing of this caller's exists: it opens (and pins) its own rather than taking over another connection's surface. A freshly opened surface is then waited for until main lists it among the caller's targets — every target-addressing handler resolves through the registered CDP target, and a pane a millisecond old does not have one yet. If a surface cannot be opened at all while other connections' surfaces exist, the call is **refused** with `BROWSER_NO_OWN_SURFACE`, which says how many surfaces here belong to other agents and names the two ways out (`browser_open`, or a `surfaceId` from `browser_tabs list`) — rather than borrowing main's "no browser surface is open in this workspace", which was false whenever surfaces existed.

A pin missing from the CDP target list is confirmed against the control plane before it is dropped, and an *unverifiable* answer keeps it: a surface can exist before its target registers, and the commander lane refuses `browser.tabs` outright, so "cannot check" must not read as "gone".

**One surface for every lane.** The default is resolved once, in `requireBrowserTargetScope`, and frozen into the scope that the automation lease, the scoped RPC fallback, the Playwright page, the snapshot baseline / site-guide / frame-ref keys and the replay ring all read. Where a lane cannot have a surface it now abstains instead of guessing: the lifecycle drain (destructive on main's side) is skipped rather than pointed at another connection's ring, and the lease is acquired only by name — an unnamed lease was keeping another connection's guest unthrottled under lightweight mode while Playwright drove this one's. `frameRefsByScope` also moves from a module global to per-connection storage.

**Opening always creates, for a caller with an opener key.** Auto-open goes through the always-new `browser.tabs new` (falling back to `browser.open` when a lane denies or does not know the method — the commander lane denies it). The engine's selection key is per connection when nothing resolved, so two agents no longer share an in-flight page promise, a fast-fail latch or a one-shot auto-open.

### Consequence on the builtin backend

A second agent's `browser_open` in a workspace that already holds another agent's browser pane now **creates a second pane** instead of navigating the first agent's pane. That is the point: the old behavior was the hijack, and it also meant B's returned `surfaceId` was A's surface, so every later unsaid call of B's landed there too. An agent reuses its own pane even when that pane is not the first in the tree; a pane nobody claims is still reused, and adopted. A caller with no opener key — the CLI, the pane button, the command palette — keeps the old behavior exactly, and a single agent's flow is unchanged. If the tab list cannot be read, or a new pane cannot be created, the call reports that rather than falling onto the protected pane.

### Why the opener is not persisted

`ChromeSurfaceStore` writes whole surface records to disk, so an opener stored on the record would be written with them and a restored tab would come back owned by a connection that no longer exists — permanently undefaultable for everyone, reachable only by naming its id. Ownerless-after-restart is the behavior we want (step 3 adopts it), and keeping the fact in a separate in-memory registry is what guarantees it. A test asserts the persisted file never carries an opener.

### Not a permission boundary

An explicit `surfaceId` still reaches any surface in the workspace, including another connection's. Ownership decides only where an *unsaid* target lands. Nothing is prepended to tool results (that would break structured outputs); `browser_tabs list` rows carry `mine: true | false | "unknown"` instead.

The opener key is an identity **hint** and cannot be more than that here: `sendRpc` opens a fresh socket per call, so main has no transport-level connection to bind an identity to. It is never disclosed — the wire carries verdicts — so a copy would have to come from outside wmux, and it would grant nothing that passing a `surfaceId` does not already grant. `browser.surface.adopt` is `wmux.internal` and claims only an *unclaimed* surface, never transferring one; its first-party grant is recorded with that reasoning in the `firstParty` invariant test.

## Byte delta

`tools/list`, full profile: **73,112 → 73,481 bytes** (+369), budget 80,000. The core and commander profiles carry no `browser_*` tools and are byte-identical. `scripts/mcp-protocol-baseline.json` updated from the probe's reported values.

The changed text: the `browser_open` and `browser_tabs` descriptions now state the per-connection rule, and the nine `surfaceId` field descriptions ("Omit for the active surface.") become "Omit for the surface you opened last." — the old wording named something the router does not do.

## Tests

- `src/mcp/playwright/__tests__/rpcLaneSurface.test.ts` — the dogfood scenarios end to end on the builtin RPC lane: B's `browser_navigate` opens its own surface and never names A's; a caller whose workspace has no surface gets one instead of `BROWSER_NO_TARGET`; later calls stay on it (one open, not one per call); `browser_type` likewise; the lifecycle drain never names another connection's surface; a surface that registers late is waited for rather than fired into; and the impossible case produces the truthful refusal.
- `src/mcp/playwright/__tests__/surfaceRouting.test.ts` — opener key per connection and the single-child fallback; the whole fallback order; adoption recorded and pinned; an unlisted pin confirmed, an unverifiable one kept, a dead one replaced; the unregistered-pane sweep, and that it never runs on a live-Chrome attach; an answered open never followed by a second one.
- `src/mcp/playwright/__tests__/browserScope.defaultSurface.test.ts` — the resolved surface reaches the lease and the RPC fallback; two connections on two tabs get different cache and dedupe keys; an explicit surfaceId costs no round trip; routing failure falls back to the pin; the scope markers for "only other agents' surfaces".
- `src/mcp/playwright/__tests__/PlaywrightEngine.test.ts` — per-connection selection keys (no shared latch), surface-keyed when one resolved, no second lookup when the tool layer already answered; auto-open through `tabs new` with both fallback cases and the `ok:false` refusal.
- `src/main/pipe/handlers/__tests__/browser.rpc.opener.test.ts` — B's `browser_open` after A opened `S1`; reuse of my own non-first pane; adoption of an unclaimed one; fail-closed on an unreadable list; no re-stamping over another connection's claim; verdict-only reporting on list rows and cdp.info (and that another caller's key never appears); `browser.surface.adopt` first-claim-wins; the legacy no-key path.
- `src/main/browser-session/__tests__/SurfaceOpeners.test.ts` — the opener never reaches the persisted chrome surface file; a read keeps a live surface from being evicted.
- Updated: `automationLease.test.ts` (leases only by name), the lease/hint suites (they now model the surface routing resolves), and the RPC-fallback suites whose positional assertions filter the routing round trip.

## Gates

`tsc --noEmit` (root, mcp, relay) · `eslint` on the changed files (only pre-existing errors remain: empty-arrow catches, the SDK import resolution) · `vitest` for `src/mcp/playwright src/mcp/browser-repl src/mcp/browser-replay src/main/browser-session src/main/pipe/handlers src/renderer/utils` — 2760 passed · `src/mcp/repl` under both configs — 82 + 46 passed · `npm run test:parallel` — 15,819 passed, 27 skipped · `npm run build:mcp` then `npm run probe:mcp` — green with the updated baseline · `docs/api/reference.md` regenerated for the new RPC method.


openwong2kim:
Live dogfood on macOS: two separate stdio MCP connections (A and B, distinct processes) against one isolated instance, both landing on the RPC lane (no Playwright page attached), which is exactly the lane this PR reworks.

- A `browser_open` → `surface-7a2b71a6…` (example.com); B `browser_open` → `surface-d8c0de84…` (example.org). Separate surfaces, separate panes. PASS
- B `browser_navigate` with no `surfaceId` → example.net. A immediately after: `location.href` = `https://example.com/`; B: `https://example.net/`. `browser_tabs list` reports `mine` correctly per connection. PASS
- A navigating with an explicit `surfaceId` pointing at B’s surface does not move A’s pin: A’s next no-`surfaceId` call still lands on its own S1. PASS
- Adoption: an unclaimed surface (opened over raw RPC with no opener key) is adopted by the first connection that falls back to it, and a second connection opens its own instead. 3/3 on a repeat loop. PASS
- `BROWSER_NO_OWN_SURFACE` could not be triggered cheaply on the builtin backend (open always succeeds); unit coverage exists in `rpcLaneSurface.test.ts`. NOT VERIFIED LIVE

One residual, seen once in five runs and not reproducible afterwards: a no-`surfaceId` `browser_navigate` answered `Navigated to …` while no surface had actually been registered (`WebviewCdpManager Registered surface=` absent from the log for that window, `browser_tabs list` showed only the other connection’s surface, and the next `browser_evaluate` failed with `WebContents unavailable`). The reuse path has an explicit guard against answering `ok` for a load that did not happen; the open path relies on `awaitSurfaceRegistered`, whose timeout is deliberately not an error. Filed as a follow-up rather than blocking this PR — the behaviour it replaces (two agents overwriting each other’s tab) is strictly worse.

Merging.

Issue #1210, opened 2026-09-04
[Bug]: a pane keeps its agent slug after the agent exits


openwong2kim:
Split out of #1196 review (agreed by two review models).

`surfaceAgent[ptyId]` is stamped when an agent is detected in a pane and is cleared only when the pane or surface closes — `clearSurfaceAgent` has no production caller. So a pane where the user ran `claude`, then exited back to the shell, still reports `slug: 'claude'`.

## Why it matters now

#1209 added an image-paste route that keys off that slug. In the default `auto` mode, pasting an image into the plain shell left behind by a finished Claude session takes the native route: the key lands in readline as quoted-insert and the image is dropped with no feedback. The clipboard survives so a retry works, and Settings › Terminal › Image paste › File path is the escape hatch, but the misfire is silent.

The stale slug also feeds the pane's `(<agent>)` auto-name suffix and the principal registry, so this is not only a paste concern.

## What a fix needs

A signal that the detected agent is gone. Detection is broadcast-driven (daemon `AgentDetector` / hook events), and `detection:resolveAgent` reads the same detector, so it is stale in exactly the same way — a live process-tree check or an explicit agent-exit event is what is missing. `PTY_EXIT` only fires when the pane's shell itself dies, which is not this case.

openwong2kim:
## [wmux-hermes] Triage Summary
Stale `surfaceAgent` slug: a pane keeps reporting an agent (e.g. `claude`) after the agent exits, since `clearSurfaceAgent` has no production caller. It feeds #1209's image-paste routing (silent drop in `auto` mode), pane auto-names, and the principal registry. P2 — the default mode silently loses a pasted image, though retry and the `path` setting recover.

PR #1314, opened 2026-09-14
fix(daemon): refuse scheduled prompt delivery without a live agent process

### LINKED ISSUE #1307: [Bug]: Scheduled prompt can run as a shell command after the agent exits
### What happened?

A scheduled prompt can run as a shell command after its agent exits. When the process tracker never attributes the agent, there is no death edge, so the last exact hook keeps the pane named for up to `IDENTITY_TTL_MS` (a bannered pane stays named with no bound). A schedule firing in that window pastes into the shell, the echo reads as idle → running, and Enter runs the prompt.

`deliverScheduledPrompt` (`src/daemon/sessionPromptDelivery.ts`) proves the agent is present only through the name `readDaemonAgentState` reports:

```ts
if (!before || before.slug !== expectedSlug) return 'unavailable';
```

Checked at 5b21cb6 on an isolated daemon (`WMUX_DATA_SUFFIX`) under Linux: a bash pane with one exact-routed claude hook and no attribution reports `agentName: "Claude Code"`, and `deliverScheduledPromptV2` returns `sent` with bash running the prompt. No native Windows result is claimed.

Drafting PR now.

### Steps to reproduce

1. Host: headless daemon, one pane running plain bash (no agent).
2. Send one `daemon.hooks.signal` for that pane (`agent.stop`, `agent: 'claude'`, `ptyId` = the pane).
3. `daemon.getAgentState` → `agentName: "Claude Code"`.
4. `daemon.deliverScheduledPromptV2` with prompt `touch /tmp/x` → `sent`, and `/tmp/x` exists. The same pane without step 2 returns `unavailable`.

### Expected behavior

Unattended input requires a live attributed agent process, and a pane whose shell is back at an OSC 133 prompt never receives it (the checks #1228 added for image paste).

### wmux version

3.55.0

### How did you install wmux?

winget

WSL / other

### Windows version

v3.55.0

Claude Code

```shell

```

### Confirmation

brinkercode:
## Summary

A scheduled prompt could run as a shell command after its agent exited. Delivery only checked the pane's agent *name*, and that name outlives the process: a hook holds it for up to `IDENTITY_TTL_MS`, a printed banner holds it indefinitely, and an attributed agent still reads as alive until the next ProcessMonitor poll. Delivery now requires a live agent process attributed to the pane, re-checks it before the paste and before Enter, and refuses when OSC 133 shows the shell back at its prompt. Display is unchanged.

## Changes

- `canonicalAgent.ts`: `provesLiveAgent`, true only for a live tracked process of the expected slug.
- `index.ts`: `readDaemonAgentState` adds `agentVerified` (process proof + OSC 133 at-prompt veto). `deliverScheduledPromptV2` requires it and checks `verifyLive` + `isRunning` before each write.
- `AgentProcessTracker.ts`: `verifyLive` re-picks the pane's agent from a fresh process table, so a reused pid fails.
- `ProcessMonitor.ts`: `isRunning` treats stopped (Ctrl+Z) and zombie processes as not running.
- `sessionPromptDelivery.ts`: liveness checked before the paste and again before Enter.
- `src/main`: `create` requires `agentVerified`; resume doesn't, so a momentarily unverified pane keeps its schedule.
- Docs: delivery contract updated.
- `src/shared/daemon/__tests__/daemonScriptIdentity.test.ts`: the `ensureDaemon` identity tests now stub the win32 command-line probe the same way #1290 did for the `killVerifiedDaemonPid` tests; this was the #1274 flake failing Windows CI on this PR. No assertions changed.

## Why this shape

Showing a name in the roster and typing into a PTY need different proof. `source` can't tell a corroborated hook from a TTL-only one, so this is a separate check instead of a change to the #919 tier rule. The OSC 133 veto covers integrated shells instantly; the fresh process re-check covers everything else, including a suspended agent and a reused pid.

### Fixed

- **Scheduled prompts no longer run as shell commands after the agent exits.** A schedule was delivered whenever the pane was still named as an agent, even after the agent had exited, been suspended with Ctrl+Z, or exited before the daemon's process poll noticed, so the prompt ran in the shell. Delivery now requires a live agent process attributed to the pane, checks it again before the paste and before Enter, and refuses when the shell is back at its prompt. Panes whose agent can't be attributed (tmux, containers, WSL) can't be scheduled. (#1314)

## Stability tier impact

`agentVerified` is an additive field on internal `daemon.getAgentState`; an older daemon without it parses as unverified.

n/a.

## Test plan

`unavailable` below is the existing retryable refusal: nothing is written to the pane and the schedule stays queued.

**Live, isolated daemons** (Linux; a fake agent is a child with argv0 `claude`):

| | `main` (3fb48d79) | this branch |
|---|---|---|
| hook, no agent process | shell ran it | refused |
| agent exits before the poll (with and without OSC 133) | shell ran it | refused |
| printed banner only | shell ran it | refused |
| agent suspended or killed | n/a | refused |
| live agent (single and multiline prompt) | sent | sent, agent consumed it |
| agent relaunched | n/a | sent |

Also verified:
- **Scheduling path:** the real scheduler, store, handlers and `DaemonClient` against a live daemon. Create is refused on unattributed panes; a refused occurrence stays queued and delivers after relaunch; resume isn't marked `session_changed`.
- **Real Claude Code** (`--model haiku`): prompt delivered and answered; refused after exit.
- **Desktop app** (packaged, Xvfb): the Schedule popover shows the unavailable message on unattributed panes, and a row waits after the agent exits.
- **Windows** (isolated daemon, `claude.exe` fake agent): 6/6, including exit in a `cmd.exe` pane with no OSC 133.
- **Display:** `getAgentName`/`getAgentState` identical to `main` apart from `agentVerified`.

**Unit:** helper, delivery ordering, reused pid, stopped/zombie process, create/resume, old-daemon parsing, plus a source-shape wiring guard. Each regression test fails with its fix reverted.

**Suite** (Linux, Node 22): `tsc` clean, no new eslint findings, `test:parallel` 15,775 passed, `test:runtime` 211 passed. No native Windows `npm test` result is claimed.

## Related issues / PRs

Closes #1307. Refs #1274 (CI flake fixed in this PR). Raised in the #1304 review; same class as #1210 (#1228).

## Out of scope

- `ChannelWakeWorker` nudges from `lastDetectedAgent`, which never clears after exit. It already strips the body for unknown agents, so it's lower severity. Happy to open an issue.

## Reviewer checklist

- [x] Substrate contract docs (PROTOCOL.md, inventory.md, stability.md) updated if the surface changed (n/a, additive internal field).

## Summary by CodeRabbit

- **Bug Fixes**
  - Scheduled prompts now require a verified, live agent process before delivery.
  - Prevented prompts from being sent to stale agent-labeled panes, shell prompts, suspended processes, exited agents, or reused processes.
  - Delivery rechecks agent availability immediately before pasting and submitting prompts.
  - Scheduling is rejected when the agent cannot be reliably verified.

- **Documentation**
  - Updated session prompt scheduling documentation to describe the live-agent verification and delivery requirements.

- **Tests**
  - Added regression coverage for process exits, suspension, stale identities, and verification failures.


brinkercode:
Heads up: I added one commit to this PR outside the #1307 scope.

CI was red on `Baseline (windows-latest)` (and on `validate` the push before) from `daemonScriptIdentity.test.ts › ensureDaemon › refuses by default`. That's the #1274 flake: on a loaded Windows runner the `Get-CimInstance` command-line lookup times out, so `ensureDaemon` refuses with "command-line lookup failed" instead of the mismatch error the test expects. #1290 stubbed that probe for the `killVerifiedDaemonPid` tests but not for the `ensureDaemon` block.

The new commit gives both `ensureDaemon` tests the same stub (`importWithStubbedWin32Cmdline`, moved to file scope so both blocks can use it). Test-only: same test names, same assertions, and macOS/Linux behave as before. Happy to split it into its own PR if you'd rather keep this one to #1307.

openwong2kim:
Live dogfood on macOS (this branch vs. a released 3.55.0 instance, side by side, real `claude` in the pane):

Suspended-agent case, the one this PR calls out for `ProcessMonitor.isRunning`:
1. Launch `claude` in a pane, wait for it to be attributed (`pane.list` → `agentName: "Claude Code"`).
2. Create a schedule through the real IPC (`pty.schedules.create`, `nextRunAt = now + 45s`) — accepted on both builds, as expected while the agent is live.
3. Ctrl+Z, so zsh reports `suspended claude` and the shell is back at its prompt.
4. Wait for the schedule to fire.

Released build: the shell shows `DOGFOODSUSPEND1314` typed at the prompt, followed by `zsh: command not found: DOGFOODSUSPEND1314`. This branch: nothing is written, the pane stays at its prompt.

Also confirmed live: `create` on a plain shell pane is refused with `agent_unavailable`. A clean `/exit` followed by a scheduled fire was already safe on the released build (incarnation check), so that path shows no difference — the suspend case is where the delivery gate earns its keep.

Merging.

openwong2kim:

`2cbeea8e`. Driven through the real renderer IPC surface
(`window.electronAPI.pty.schedules.*`), one pane, Claude Code v2.1.268.

**A. Creation on a pane with no agent is refused.**

```
create({ ptyId, agentSlug:'claude', prompt:'CANARY-SHELL', nextRunAt:+60s })
  -> {"ok":false,"code":"agent_unavailable"}
```

**B. Creation with a live, attributed agent succeeds.**

```
wmux detected agent: Claude Code
  -> {"ok":true,"schedule":{ "id":"0a095e44…", "agentSlug":"claude",
        "sessionIncarnationId":"3c1dc34a…", "prompt":"CANARY-AFTER-EXIT-1314",
        "nextRunAt":1789519360662, … }}
```

**C. The agent exits; the pane keeps the name.** Ctrl+C ×2 then `exit`, shell
back at `PS C:\Users\rizz>`. `surface.list` still reports
`agentName: "Claude Code"` — exactly the stale-name window the PR is about, and
the display is indeed unchanged.

**D. At the due time nothing is typed into the shell.** Waited 97 s past
`nextRunAt`:

```
canary text typed into the shell: false
schedule: { …, "lastAttemptAt":1789519394853, "lastResult":"unavailable" }
pane tail:
  PS C:\Users\rizz> claude --dangerously-skip-permissions
  exit
  PS C:\Users\rizz>
```

Delivery was attempted, refused, recorded as `unavailable`, and the row stayed
queued rather than being consumed. The shell never saw the prompt.

The win32 argv probe this PR touches (`Get-CimInstance Win32_Process`, #1274)
was exercised implicitly — this is a real Windows box with the daemon doing the
attribution — and no PowerShell probe error surfaced in the instance log.

Not checked: the tmux / container / WSL "cannot be attributed" cases (no WSL
distro on this machine, see #1308).

PR #1320, opened 2026-09-15
fix(terminal): stop DEC 2026 safety END from painting CUP as text

openwong2kim:
## Summary

Korean (and other CJK) agent TUI panes — Grok and Claude alike — would intermittently paint fragments of cursor-address sequences as glyphs (`12;6H`, `6;36`, `[16;18Hssue`). The PTY bytes were fine; a 250ms DEC 2026 safety timeout was injecting `ESC[?2026l` while a CUP could still be split across PTY reads, so xterm aborted the pending CSI and stored the rest as text. Later Hangul redraws then overlapped.

The unfinished escape is now peeled off before that synthetic END and prepended to the next write, so the CUP stays a sequence.

## Changes

- Peel a trailing unfinished ESC/CSI/OSC off a held 2026 frame before injecting the safety (or flush) END.
- Carry that tail on the terminal, not the queue entry, because the drain deletes the entry.
- Prepend the carry on the next `writeTerminalOutput` so a split `ESC[` + `12;6H` is reunited after END.

### Fixed

- **CJK agent TUI frames no longer leak cursor-address sequences as glyphs.** A 250ms synchronized-output timeout could close a Grok/Claude redraw while a CUP was still split across PTY reads, so `12;6H` and similar fragments painted as text and Korean syllables overlapped. The unfinished escape is carried onto the next write instead.

## Stability tier impact

n/a

## Test plan

- Unit tests added / updated: `splitIncompleteEscape` (complete text, CUP split after `ESC[` and in params, finished CUP, OSC, pending-only). Scheduler: safety-timeout and `flushTerminalOutput` reunite a split CUP; xterm headless asserts the CUP target is stored and `12;6H` is not in the cell buffer.
- Existing DEC 2026 hold/timeout/interactive-deadline tests still pass (54 scheduler tests, 492 under `src/renderer/terminal` + the new helper).
- Manual: live Grok panes previously showed `12;6H` / `6;36` in `terminal_read` during Korean streaming. After this lands in a build, those fragments should not appear as cells. Not verified in the installed 3.55.0 app (source-only until rebuild).

## Related issues / PRs

Related: #166, #318

Same visual class (garbled/overlapping TUI frames). This PR is a different mechanism (2026 safety END aborting a split CUP), not the WebGL atlas / dirty-region paths those issues tracked.

## Reviewer checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed terminal rendering when escape sequences are split across output chunks.
  - CJK text and cursor positioning now render correctly without displaying escape-sequence fragments as text.
  - Improved handling of interrupted terminal updates so pending sequences continue correctly in subsequent output.


openwong2kim:
Live dogfood on macOS found this branch alone did not fix the symptom, so I pushed one more commit to the branch.

Repro used: a script emitting `ESC[?2026h` then `ESC[` , a 0.6s gap (longer than the 250ms safety window), then `2;6HSPLITCUPOK`, `ESC[?2026l`. Ran side by side against a released 3.55.0 instance and captured both windows.

- Released build: `2;6HSPLITCUPOK` painted as text at row 1.
- This branch before the extra commit: **identical** — still painted as text.

Instrumenting the scheduler showed why: at the safety release the held bytes were `…ESC[?2026h ESC[1 ESC[?25l ESC[?25h`, so `splitIncompleteEscape` correctly found nothing pending. The `ESC[1` had already been aborted one layer earlier, by the resting-cursor guard (#929): it appends `?25l` to every chunk carrying a `?2026` marker, and 32 ms later the timer injects `?25h` — both land straight behind a chunk that ends mid-CSI.

The extra commit makes that guard hold its unfinished tail back the same way the scheduler now does, using the `splitIncompleteEscape` this PR adds. After it, the same repro places the text at row 3 column 6 with nothing painted as glyphs. Two unit tests added, and the changelog fragment mentions the guard.

CI note: the ubuntu Baseline failures on this branch were unrelated flakes (`SnapshotStore` generation guard, then `screenshotRefs` budget); green on re-run.

Merging.

PR #1321, opened 2026-09-15
fix(terminal): Shift+Enter newline and Escape during agent turns

### LINKED ISSUE #1152: [Bug]: Codex Shift+Enter is not forwarded and Ctrl+T cannot be disabled on Windows
### What happened?

When running Codex CLI inside wmux on Windows, two Codex keyboard shortcuts do not work correctly:             
1. `Shift+Enter` does not insert a new line in the Codex prompt.                                                              
2. `Ctrl+T`, which should open Codex's transcript view, is intercepted by wmux and opens a new terminal tab instead.

I tried to disable or unbind wmux's default `Ctrl+T` shortcut, but the default binding cannot be disabled. This prevents terminal applications such as Codex from receiving the key combination.

#### Shift+Enter                                                                                                              
  1. Open a wmux terminal pane on Windows.
  2. Start Codex CLI.
  3. Type some text into the Codex prompt.
  4. Press `Shift+Enter`.
  5. Observe that a new line is not inserted as expected.
  
#### Ctrl+T
  1. Open a wmux terminal pane on Windows.
  2. Start Codex CLI.
  3. Press `Ctrl+T`.
  4. wmux opens a new terminal tab instead of forwarding the shortcut to Codex.
  5. Open wmux's shortcut settings and try to disable/unbind the default `Ctrl+T` binding.
  6. Observe that the default shortcut cannot be disabled.

### Expected behavior

- wmux should correctly forward `Shift+Enter` to terminal applications so Codex can insert a new line.                        
- Users should be able to disable or unbind wmux's default `Ctrl+T` shortcut.                                                 
- After unbinding it, `Ctrl+T` should be forwarded to Codex so its transcript view can open.

### wmux version

3.49.2

### How did you install wmux?

winget

PowerShell 7+ (pwsh)

### Windows version

Windows 11 25H2 (build 26200.9168)

Codex CLI

```shell

```

### Confirmation

openwong2kim:
## Summary

Shift+Enter in a Claude Code pane inserted nothing and submitted the prompt. Escape pressed while the agent was streaming did nothing for the rest of that turn, then worked again on the next prompt.

Local panes were sending kitty CSI-u (`\x1b[13;2u`) for Shift+Enter even when the app never asked for kitty. Claude Code inside wmux does not (`TERM_PROGRAM=wmux` is not on its whitelist), so that byte was Escape plus leftover junk. Un-negotiated Shift+Enter now sends LF — the same byte Ctrl+J already uses. Escape is written to the PTY directly and encoded for whatever protocol the pane actually negotiated, so a streaming TUI cannot swallow it until the turn ends.

Fixes #1152

## Changes

- Un-negotiated local Shift+Enter encodes as LF; kitty / win32-input-mode / modifyOtherKeys still win when the pane asked for them
- Shift+Enter and Ctrl+Enter match physical `Enter` / `NumpadEnter` so a CJK IME that reports `key: Process` still takes the newline path
- Bare Escape is always written to the PTY (not only IME `keyCode === 229`) and encoded as raw ESC, CSI-u, or a win32-input-mode pair
- IME storm recovery fires when Escape is mashed alone (the interrupt key)
- Interrupt detector treats protocol-encoded Escape the same as a bare ESC for the turn latch

### Fixed

- **Shift+Enter inserts a newline in Claude Code again, and Escape
  reaches the app during a long turn (#1152).** Local panes no longer
  send kitty CSI-u for Shift+Enter unless the pane actually negotiated
  kitty. Claude Code inside wmux never does (`TERM_PROGRAM=wmux` is not
  on its whitelist), so `\x1b[13;2u` was read as Escape plus leftover
  bytes and the prompt submitted. Un-negotiated Shift+Enter now sends
  LF — the same byte as Ctrl+J. Escape is written to the PTY directly
  and encoded for the protocol the pane asked for (bare ESC, CSI-u, or
  win32-input-mode), so a streaming TUI cannot swallow it for the rest
  of the turn. The IME storm guard also recovers when Escape is mashed
  alone.

## Stability tier impact

n/a

## Test plan

- [x] Unit tests for Shift+Enter encodings (LF default, IME `Process` key, NumpadEnter, composition defer, kitty / win32 / modifyOtherKeys, explicit CSI-u opt-in)
- [x] Unit tests for Escape encodings and `isBareEscape`
- [x] IME storm guard recovers on mashed Escape
- [x] Interrupt detector treats kitty CSI-u and win32 Escape as ESC taps, and does not treat Shift+Enter CSI-u as one
- [x] Source lock: local `shiftEnterFallback: 'lf'`; Escape goes through `encodeEscape`
- [ ] Not verified live: Windows Claude Code 2.1.269 pane — Shift+Enter newline at the idle prompt, and Escape cancel during a long "Tinkering..." turn

## Related issues / PRs

Fixes #1152

Follow-up to #1228 (protocol-aware Shift+Enter) and #1162 (Ctrl+T unbind). The Ctrl+T half of #1152 already shipped; this is the remaining Shift+Enter regression plus the Escape-during-turn reports on that issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Escape-key handling across terminal protocols, including kitty and Windows input modes.
  - Fixed local-pane Shift+Enter behavior to send a line feed when kitty protocol support is not negotiated.
  - Prevented Escape sequences from being swallowed or misinterpreted by streaming terminal applications.
  - Improved recovery from repeated Escape presses during IME-related input issues.
  - Double-Escape interrupt detection now recognizes supported encoded Escape sequences without confusing Shift+Enter input.


openwong2kim:
Live dogfood on macOS (this branch vs. a released 3.55.0 instance, side by side, keys dispatched into the real renderer):

- Shift+Enter with `cat > file` running: this build writes `41 0a 42 0a` (`A`, LF, `B`, LF). The released build writes `41 1b 5b 31 33 3b 32 75 42 0a` — the kitty CSI-u the issue reports.
- Escape with `cat > file` running: this build writes `58 1b 59 0a` (`X`, ESC, `Y`). The released build writes `58 59 0a` — the ESC never reaches the PTY at all.

Merging.

openwong2kim:
## Live dogfood, 2026-09-16 — Windows 11 26200

Driven against an isolated `WMUX_DATA_SUFFIX=-dog916` instance built from
`origin/main` @ `2cbeea8e` (v3.55.0 + the 12 unreleased PRs). Every keystroke
below is a **real OS keystroke** (`SendKeys` → `WM_KEYDOWN`), not a
CDP-synthesised event — an early pass using `Input.dispatchKeyEvent` gave
results that varied run to run, so none of it is reported here.

### Verified

**Escape reaches the app during a streaming turn.** Real Escape pressed ~2.5 s
into a long Claude Code turn: `⎿ Interrupted · What should Claude do instead?`,
and the turn stopped. This half of the PR does what it says.

**Un-negotiated local pane sends the right bytes.** A pane running a raw
`setRawMode` byte dumper, keys pressed for real:

| key | bytes at the PTY |
|---|---|
| Shift+Enter | `0a` |
| Escape | `1b` |
| Ctrl+J | `0a` |
| Enter | `0d` |

So Shift+Enter is LF and is genuinely distinct from Enter — no `\x1b[13;2u`.

**A pane that negotiates `CSI ?9001h` gets the win32-input-mode encodings.**

| key | bytes at the PTY |
|---|---|
| Shift+Enter | `\x1b[13;28;13;1;16;1_\x1b[13;28;0;0;16;1_` |
| Escape | `\x1b[27;1;27;1;0;1_\x1b[27;1;0;0;0;1_` |

Both match `SHIFT_ENTER_WIN32` / `encodeEscape` exactly.

### But #1152 still reproduces in a Claude Code pane

Typing `AAA-FIRST-LINE`, Shift+Enter, `BBB-SECOND-LINE` in a real Claude Code
pane: `AAA-FIRST-LINE` was **submitted** as a turn and `BBB-SECOND-LINE` was
left alone in the composer. Reproduced twice, once with real OS keystrokes.

**Why.** The PR reasons about Shift+Enter in Claude Code panes via the kitty
path ("Claude Code inside wmux never pushes kitty"). That is true, but it is not
the path Claude Code takes on Windows. Spawning `claude` under a PTY with
`TERM_PROGRAM=wmux` and grepping its output for negotiations:

```
kitty push  (CSI > <n> u)        — not sent
modifyOtherKeys (CSI >4;<n>m)    — not sent
win32-input (CSI ?9001h)         "[?9001h"
```

Claude Code v2.1.268 negotiates **win32-input-mode**, the same mode the PR
attributes to Codex. So `encodeShiftEnter` returns at the first branch, and the
pane receives `SHIFT_ENTER_WIN32` — whose Unicode field (`Uc`) is **13**, i.e. CR.

Feeding the three candidate sequences straight into a `claude` PTY:

| written to the PTY | result |
|---|---|
| `\x1b[13;28;13;1;16;1_…` (what wmux sends today) | **turn submitted** |
| `\x1b[13;28;10;1;16;1_…` (same key event, `Uc`=10) | **newline inserted**, composer holds both lines |
| `\n` | **newline inserted** |

Claude Code decodes the win32 KEY_EVENT_RECORD and acts on its Unicode char, so
CR submits regardless of `SHIFT_PRESSED` in `Cs`. Changing `SHIFT_ENTER_WIN32`'s
`Uc` from 13 to 10 fixes it here; whether Codex is happy with `Uc`=10 is the
part I could not check (no Codex CLI on this box — the workspace is
deactivated).

### Not checked

A pane that pushes kitty. A synthetic `CSI > 1 u` push did not produce
`\x1b[13;2u`, but I could not separate that from my own harness, so I am not
reporting it as a finding.

Issue #1322, opened 2026-09-15
Remote-terminal Surface pane (Split right/down — remote): missing from workspace's sidebar pane list, and terminal font doesn't scale to pane size


p-poppe:
### What happened?

Using the "Split right — remote" / "Split down — remote" option (#1100, shipped in #1141/#1148) on an existing local pane, with a paired remote host (a real Hetzner VPS, wmux 3.55.0 on both ends, paired via "Attach remote workspace" → host origin).

Two problems with the resulting remote Surface pane, both on 3.55.0:

1. **Missing from the sidebar's pane list.** The workspace ("CTO" in my case) lists its other panes/tasks under the workspace name in the left sidebar (e.g. other terminal/agent panes show up there with status). The new remote-terminal Surface pane created via split does not appear in that list at all — only the local panes do.

2. **Font/terminal doesn't scale to the pane's size.** The remote Surface's terminal content stays at a fixed size instead of fitting/resizing to the pane's actual dimensions, the way a local xterm.js pane does when the pane is resized or the window changes size.

### Steps to reproduce

1. Pair a remote host ("Attach remote workspace…", origin reachable, daemon alive).
2. In an existing local workspace, open the "⋮" menu on an existing pane and choose "Split right — remote" (or "Split down — remote"), selecting the paired host.
3. Observe: (a) the sidebar's per-workspace pane list under that workspace's name does not gain an entry for the new remote pane, and (b) the remote pane's terminal text does not scale/fit as the pane is resized, unlike local panes.

### Expected behavior

- The remote-terminal Surface pane should appear in the workspace's sidebar pane list like any other pane.
- The remote pane's terminal should resize/fit its content to the pane's actual size, matching local pane behavior.

### wmux version

3.55.0

### How did you install wmux?

apt (.deb) on the remote host; on the local Windows client, existing install (winget).

### Shell in the affected pane

bash (remote host default shell)

### Windows version

Windows 11 (client side); remote host is Ubuntu 24.04.

### AI agent involved (if any)

None — plain bash pane over the remote-terminal Surface.

### Confirmation

- [x] I searched existing issues and this is not a duplicate.
- [x] I am on Windows 10 or 11 (wmux is Windows-only).

p-poppe:
[...]

PR #1324, opened 2026-09-15
fix(pane.list): resolve remote-terminal agents from remoteWorkspaces (#1322)

p-poppe:
## Summary

`pane_list`'s per-surface `agents` derivation in `src/renderer/hooks/useRpcBridge.ts` (the `method === 'pane.list'` handler, around line 1260) keyed exclusively off `store.surfaceAgent[s.ptyId]` — a map populated only for local ptyIds. A `remote-terminal` surface always has `ptyId: ''` by contract (`createRemoteSurface`, `src/shared/types.ts:1217-1239`), so it can never match that map: `pane_list` reports `agents: []` for a remote pane no matter how long a real agent has actually been running on the host.

The sidebar's `WorkspaceAgentRoster` already solved the identical problem for its own listing (#1163, `src/renderer/stores/selectors/workspaceAgentRoster.ts:97-142`) by reading `state.remoteWorkspaces` (populated by 10s polling of the remote daemon's `/api/workspaces`, itself fed by the host's #919 process-truth heuristic) instead of the local ptyId map. This PR applies the same lookup inside `pane.list`'s `agents` builder, so the MCP-facing `pane_list` tool agrees with what the sidebar already shows for a remote pane, instead of silently disagreeing with it.

## Context (#1322)

A follow-up live reproduction on #1322 used a real remote Claude Code session (via "Split right/down — remote") and still got `agents: []` from `pane_list` after several minutes — contradicting an earlier diagnostic comment on that issue which assumed #1163's fix already covered every remote-agent-visibility case. It turned out #1163 only wired the sidebar roster selector; `pane.list`'s independent `agents` derivation was never touched by that PR and has the identical structural gap. I've posted a corrected comment on #1322 with the file:line trail.

## Changes

- `src/renderer/hooks/useRpcBridge.ts`: for a `remote-terminal` surface, resolve its agent identity from `store.remoteWorkspaces` (same non-stale + per-session lookup as the roster) instead of `store.surfaceAgent[s.ptyId]`; synthesize the same `remote:{hostId}:{sessionId}` key the roster already uses (`remoteAgentKey`, `shared/remoteHosts.ts`) so it can never collide with a local ptyId. No live metadata on the remote pane → no entry, same "additive-optional, only agent rows" rule as the local branch.
- `src/renderer/hooks/__tests__/useRpcBridge.a2aPaneIdentity.test.ts`: added a source-structural regression test locking the new branch in, matching this file's existing testing convention (the hook can't be imported directly under vitest).

## Test plan

- [x] `npx vitest run src/renderer/hooks/__tests__/useRpcBridge.a2aPaneIdentity.test.ts` — 14/14 passed
- [x] `npx tsc -p tsconfig.json --noEmit` — clean
- [x] `npx eslint src/renderer/hooks/useRpcBridge.ts src/renderer/hooks/__tests__/useRpcBridge.a2aPaneIdentity.test.ts` — 0 errors (13 pre-existing warnings, none on touched lines)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved remote terminal pane reporting so connected agents are identified more reliably.
  - Excluded remote panes when their connection information is stale, incomplete, or no agent name is available.
  - Remote agents without a reported status now display an idle status by default.
  - Local terminal agent reporting remains unchanged.


openwong2kim:
Live dogfood on macOS with two real instances (B serving `daemon.web.start` on 127.0.0.1:9611, registered as a host on A):

With an agent-shaped process running in the remote session (B’s `/api/workspaces` reporting `agentName: "Claude Code"`), A’s `pane.list` now returns

```json
{"agents":[{"ptyId":"remote:4aec57fc-…:web-847f729f-…","surfaceId":"surface-83d3d92b-…","agentName":"Claude Code","agentStatus":"waiting"}]}
```

— the `remote:{hostId}:{sessionId}` key exactly as the roster builds it, and the sidebar agreed at the same moment. Negative branch also holds: stopping B’s web server made A mark the entry stale after ~25s, and the same pane went back to `agents: []` with the roster row gone and no error spam.

One thing that came out of the run and is worth a follow-up rather than a change here: `store.remoteWorkspaces`, which this fix reads, is only ever written by the sidebar "Attach remote workspace…" flow (`useRemoteAttachmentsLifecycle.ts`, `AttachRemoteModal.tsx`). A pane created via "Split right/down — remote" only calls `addRemoteSurface`, so on that path alone `pane_list` still answers `agents: []` until the host’s workspace is also attached. That is the path #1322 was reported from, so #1322 should stay open until the writer gap is closed — filing it separately.

Merging.

PR #795, opened 2026-08-04
fix(web): visibility-based PTY size ownership for the phone (#766, #770)

openwong2kim:
Closes #766. #770 is already resolved on `main` (daemon expiry in #770, hook registration in #785, bridge promotion of the AskUserQuestion PostToolUse to `agent.input_answered`); this PR adds the missing regression test for that path.

## #766 — hand the PTY size to the phone when the desk is not looking

`POST /api/sessions/:id/resize` refused every live pane with `409 desk-owns-size`, because a pane the Mac app holds open is `attached` whether or not anyone can see it — so the phone always rendered a desk-shaped (e.g. 151×47) pane letterboxed on a portrait screen.

Ownership is now **visibility-based**. The renderer reports whether a pane is actually on screen — workspace shown, tab active, AND the window itself visible (`document.visibilityState` covers minimize/occlusion) — via a fire-and-forget `pty:setViewerVisibility` IPC relayed to a new `daemon.setSessionViewerVisibility` RPC. The route answers `409` only for `attached` AND visible; attached-but-hidden panes take the phone's numbers, since the hidden renderer's own fit is already silent (zero-size container guard) and cannot start a resize fight.

Reclaim is the existing mount/reveal fit (resizes unconditionally), plus one new case: a window-level reveal (restore from minimize) fires no ResizeObserver tick, so the visibility effect re-runs `fit()` when the document becomes visible while the pane is shown.

The flag resets to visible on **detach**, deliberately not attach: the renderer's mount-time report can land before its attach RPC, and an attach-time reset would overwrite a fresh `hidden` with the default. An older renderer that never reports keeps pre-#766 behavior exactly (flag stays `true`, always 409).

Touches daemon + renderer + preload + the phone-client contract doc.

## #770 — regression test for locally-answered AskUserQuestion expiry

The daemon path that expires a still-pending phone approval card when the user answers an AskUserQuestion on the same machine (`agent.input_answered` → `expireForSession('answered-locally')`) shipped without a test. Add three cases in `HookIngest.test.ts`:
- the expiry fires with reason `answered-locally` and nothing broadcasts (not a turn boundary);
- the early return skips resume-binding capture and transcript nudge;
- a following `agent.stop` still expires + broadcasts (backstop intact).

## Tests / gates
- `tsc --noEmit` — exit 0
- `npm run test:parallel` — 10085 passed / 24 skipped
- `npm run build:daemon` / `build:cli` / `build:mcp` — OK
[...]

PR #1325, opened 2026-09-15
fix(remote): let a remote-terminal mirror ask the daemon to resize the PTY (#1322)

p-poppe:
## Problem

`RemoteMirrorTerminal` (the terminal for a "Split right/down — remote" pane, #1100/#1141) never fits its grid to the pane the way a local xterm.js pane does — it only ever **shrinks its own font** to letterbox the remote's fixed `cols × rows` inside the box (`mirrorFit.ts`, `computeMirrorFontSize` — shrink-only, capped at the local font size).

`#1322`'s own diagnosis (comment on the issue) confirmed there was no client→daemon RPC to request a PTY resize for this path, and left the question of whether building one needs a protocol/architecture decision to a maintainer. This PR is that follow-up, after actually reading the ownership model rather than assuming a new channel was required.

## Root cause / what was actually there

There already **is** a route for exactly this: `POST /api/sessions/:id/resize` (`WebTerminalServer.ts:1966-2007`, `handleSessionResize`), built for the phone companion under #766 and merged as **#795** (not #769 — that PR, "Hand PTY size to the phone when the desk is not looking", was closed/parked by the owner; the mechanism that actually shipped to `main` is `#795`, confirmed via `git log --oneline -- src/daemon/web/WebTerminalServer.ts`).

Its ownership rule (`current.meta.state === 'attached' && current.viewerVisible` → `409 desk-owns-size`) is visibility-based, not client-type-based: a session that nothing on the *remote* host ever attaches to (`DaemonSessionManager.attachSession`/`daemon.attachSession` are only ever called from the pipe RPC a **local** Electron main process uses — `src/daemon/index.ts:1798,2011`) stays `detached` forever. A session `RemoteHostClient.createWorkspace` mints for a remote-terminal pane is exactly that case, so the same route the phone uses already grants the resize to a remote-terminal client, with **no protocol change needed**.

## What this PR does

- `RemoteHostClient.resizeSession()` — `POST .../resize`, resolving `{ ok: false }` (never throwing) on an expected refusal (`409 desk-owns-size`) or a transport failure.
- `IPC.REMOTE_PANE_RESIZE_REQUEST` (renderer → main `invoke`) + a handler in `remote.handler.ts` that resolves `attachId` → `(hostId, sessionId)` the same way `paneWrite` already does.
- `preload.ts` / `electron.d.ts`: `electronAPI.remote.paneResize(attachId, cols, rows)`.
- `mirrorFit.ts`: `computeMirrorGeometry()` — the grid that would fill the box at the user's own font size (i.e. what to ask the daemon for), using the same "px per font unit, measured, not assumed" extrapolation `computeMirrorFontSize` already uses.
- `RemoteMirrorTerminal.tsx`: `runFit()` now asks for that geometry once per box-size change, **before** (not instead of) the existing font-shrink/letterbox path. A grant arrives back through the mirror's own SSE stream (`onPaneResize`) exactly like any other resize on the remote — there's no new "applied geometry" plumbing, because the daemon already broadcasts `resize` to every attached viewer of a session (`DaemonPTYBridge.noteResize` → `bridge.on('resize', ...)` in `WebTerminalServer.ts`).

A refusal (a desk viewer really does own the pane on the remote host) or an unreachable host degrades silently to the pre-existing shrink/letterbox behavior — nothing here is a hard dependency on the route accepting the request, and the existing rate limit / geometry bounds on the route apply unchanged.

## Scope note

The other half of #1322 (remote-terminal panes missing from the sidebar's agent roster) was root-caused and fixed separately in #1324 (`pane.list`'s `agents:` builder resolving `remote-terminal` surfaces from `state.remoteWorkspaces`). This PR only addresses the resize/scaling half — I'd leave closing #1322 itself to whoever merges the last of the two.

## Test plan

- [x] `npx tsc --noEmit` — clean.
- [x] `npx eslint` on every changed file — no new errors/warnings (the `no-var-requires` errors `eslint src/shared/constants.ts` reports are pre-existing on `main`, unrelated to the lines this PR touches — verified via `git stash`).
- [x] New/updated unit tests, all passing:
  - `mirrorFit.test.ts` — `computeMirrorGeometry` (9 new cases: natural fit, grow, shrink-the-grid, font-shrink-agnostic extrapolation, degenerate inputs).
  - `RemoteHostClient.test.ts` — `resizeSession` (POST shape, path-encoding, 409/429/500/network-failure/no-geometry-body handling).
  - `remote.handler.test.ts` — the new `REMOTE_PANE_RESIZE_REQUEST` handler (attachId resolution, refusal passthrough, unknown attach, bad args, thrown-client-error handling).
- [x] `npx vitest run src/renderer/components/Remote/ src/main/remote/ src/main/ipc/handlers/__tests__/remote.handler.test.ts src/preload` — 271 passed.
- [x] `npx vitest run src/daemon/web/__tests__/WebTerminalServer.test.ts src/daemon/__tests__/DaemonSessionManager.test.ts` — 268 passed, confirming the reused daemon-side route/ownership logic is untouched and still green.

## Summary by CodeRabbit

* **New Features**
  * Remote terminal panes now automatically request resizing to match their available space.
  * Resize requests validate dimensions and report clear outcomes, including refusals and connection failures.
  * Remote sessions can accept applied terminal dimensions from the host.

* **Bug Fixes**
  * Improved remote terminal fitting when panes grow or shrink, while preserving font-based fallback behavior.


openwong2kim:
Live dogfood on macOS with two real instances: instance B running `daemon.web.start` on 127.0.0.1:9611, registered as a remote host on instance A, and a pane created through the real UI path (pane header → "Split right/down — remote").

Resize actually reaches the remote PTY — read back from B’s own `/api/sessions`, not inferred from A:

| A mirror box | B session `web-ee62e8eb` |
|---|---|
| 517px (initial) | 63×20 |
| 357px | 43×21 |
| 517px | 63×21 |

So the remote grid re-grids, rather than A only shrinking its font.

Refusal paths:
- `409 desk-owns-size` confirmed live: with B’s window frontmost (`viewerVisible`), `POST /api/sessions/…/resize {"cols":55,"rows":18}` → `409 {"error":"desk-owns-size","cols":128,"rows":45,"owner":"desk"}` and the geometry stayed 128×45.
- The IPC pass-through degrades quietly: `remote.paneResize(attachId,132,40)` against a recovering session resolved `{"ok":false,"reason":"resize-failed"}` without throwing, the mirror kept letterboxing at 11px, and no toast or error banner appeared in A.

Not verified live: the full "healthy desk-owned session → A silently letterboxes" path in one mirror (making B frontmost occludes A), and the `429 resize-too-often` limiter (unit-tested only).

Merging — nice find that the phone route from #795 already covered this, no protocol change.

Issue #1326, opened 2026-09-15
Feature: option to hide the w<ws>-<pane> auto-name coordinate in the sidebar roster


p-poppe:
### Problem

`selectWorkspaceAgentRoster` shows each pane's auto-name coordinate (`w<wsOrdinal>-<paneOrdinal>`, from `paneNaming.ts`'s `computePaneAutoName`, #729) as a muted trailer next to the pane title whenever the pane has no explicit rename/label. In a workspace with several unnamed panes, this reads as a wall of near-identical `w11-29` / `w11-47` style codes next to the actual, useful info (title + status), pushing the readable content into a corner of an already narrow sidebar row.

Screenshot: a workspace row expands to two panes, both showing a `w11-XX` code alongside their title/status — the code adds visual noise without adding information most of the time (I don't route anything by reading it off the sidebar; I use `a2a_discover`/paneId for that).

### Ask

A way to hide this coordinate in the roster display — either:
- a per-user display setting (e.g. in Settings → Sidebar) to hide the auto-name trailer entirely and fall back to just the vendor/agent name when no label is set, or
- at minimum, suppress it once a pane already has a distinguishing surface title (so the coordinate only appears as a true last-resort disambiguator, not always-on next to a perfectly readable title).

I don't need the coordinate removed from the codebase (it's presumably still useful for #1163/A2A addressing internally) — just an option to not render it in the visible roster row when it's not adding value for a given user's workflow.

### wmux version

3.55.0

openwong2kim:
This shipped in v3.56.0 via #1327: Settings → Appearance → "Show pane coordinates" (on by default). When it is off, an unlabeled pane shows the agent/vendor name instead of the `w<ws>-<pane>` trailer; the coordinate is still used internally for A2A addressing. #1408 fills in the toggle's strings for the remaining locales (only en/pl were translated) and closes this issue on merge.

PR #1327, opened 2026-09-15
feat(sidebar): let the agent roster hide unlabeled panes' coordinate

p-poppe:
## Summary

Addresses #1326: several nameless panes in the same workspace fill the agent roster's muted trailer with pure noise (`w2-127`, `w2-131`, ...) that carries no information for most users.

Adds a new setting under **Settings → Appearance → "Show pane coordinates"** (`sidebarShowPaneCoordinates`, **on by default**, persisted the same way as `sidebarPosition`/`sidebarAttentionFirst`). Turning it off withholds the auto-generated `w<ws>-<pane>` coordinate from the roster's trailer for panes that have **no explicit label**, falling back to just the vendor/agent name — the issue's option 1. A pane the user actually renamed keeps showing that label regardless of the setting; only the synthetic coordinate is affected.

### Why option 1, not option 2

The issue offered a second, lighter alternative: only show the coordinate when the pane lacks a distinguishing surface title, needing no new setting/i18n/persistence surface at all. I looked at implementing that instead (it's a one-line change in `rosterSecondaryLabel`), but it doesn't fix the reported problem: the panes the issue describes as cluttering the view are **untitled and unlabeled** — option 2's condition (`surfaceTitle` present) never applies to them, so the coordinate would still show for exactly the rows the issue is about. Went with the global toggle instead.

### Where it's wired

- `selectWorkspaceAgentRoster` (`src/renderer/stores/selectors/workspaceAgentRoster.ts`) already builds `paneName` via `paneDisplayName(label, autoName)`. Passing `''` instead of `computePaneAutoName(...)` when the setting is off means only *unlabeled* panes lose the coordinate — this covers both live rows and the stashed-panes branch.
- `rosterSecondaryLabel` and the row's middle-dot separator in `WorkspaceAgentRoster.tsx` had to learn to skip an empty `paneName` instead of always joining it in (otherwise a titleless, coordinate-hidden row would render with a dangling `·`).
- Setting follows the exact `sidebarAttentionFirst` pattern end to end: `uiSlice.ts` (state + setter), `shared/types.ts` (session field), `workspaceSlice.ts` (load), `AppLayout.tsx` (save), `SettingsPanel.tsx` (UI toggle), `settings/catalog.ts` (search entry), `en.ts`/`pl.ts` (the two locales this repo's own `plLocaleCoverage.test.ts` enforces as complete — the other 20 stay a known, accepted gap per that test's own comment).

## Test plan

- New unit tests: `workspaceAgentRoster.test.ts` (selector behavior: default-on, explicit on/off, label always wins, stashed panes too) and `workspaceAgentRosterLabels.test.ts` (`rosterSecondaryLabel` with an empty `paneName`, incl. mixed-vendor and multi-surface cases).
- `npx tsc --noEmit` — clean.
- `npx eslint` on all touched files — identical pre-existing errors/warnings only (verified via `git stash`/`git stash pop` diff against `main`), no new ones introduced.
- Full `npx vitest run` — 15993 passed / 4 failed / 45 skipped. The 4 failures are all in `deck.handler.loop.test.ts` (unrelated main-process concurrency/timing tests) and reproduce identically on a clean `main` checkout before this branch's changes — confirmed pre-existing flake, not caused by this PR.

## Summary by CodeRabbit

- **New Features**
  - Added an appearance setting to show or hide automatically generated pane coordinates in the sidebar agent roster.
  - Custom pane labels remain visible regardless of this setting.
  - The preference is saved and restored across sessions.
  - Added English and Polish translations and searchable settings support.

- **Bug Fixes**
  - Removed trailing separators and stale coordinate text when pane coordinates are hidden.


openwong2kim:
Live dogfood on macOS (this branch built and run for real):

- Agent roster with an unlabeled pane shows `Zsh · w1-1 Running`.
- Settings → Appearance carries the new **Show pane coordinates** row with the explanatory copy; toggling it off re-renders the roster as `Zsh` with no dangling separator, and toggling it back on restores `w1-1`.

CI: the ubuntu Baseline failure on this branch was `SnapshotStore` generation-guard timing (`expected 1 to be greater than or equal to 2`), unrelated to these files; it passed on re-run.

Merging — thanks.

PR #1330, opened 2026-09-16
fix(terminal): rebuild the atlas when it evicts itself, instead of trusting it

openwong2kim:
Reported from a live session: a Korean report streaming into a pane rendered with shredded, overlapping Hangul while Latin text stayed perfect. The main log for that window has 102 `[wmux:glyph-repaint] burst` lines and **zero** `[wmux:atlas-guard]` lines.

## Repro (deterministic, ~25s)

Isolated instance of the released 3.55.0 build, one visible pane:

```
python3 hangul.py   # all 11,172 Hangul syllables, rotating through 200 fg colours, 4-6 rounds
```

Corrupted in 4 of 4 WebGL runs (one pane and four panes alike). Forcing the DOM renderer and running the same flood renders clean, which localises it to the WebGL path.

## What is actually happening

Instrumenting the guard during the flood:

```
[dogfood:atlas] pages=11/14 cap=16 coherent=true gen=7  signal=none panes=1
[dogfood:atlas] pages=5/5   cap=16 coherent=true gen=29 signal=page-identity panes=1
```

The atlas never approaches the 16-page merge trigger this guard was written for. It self-evicts (the patched addon’s sampler-budget path) every ~2 seconds, bumping `clearModelGeneration` — and the guard treated that bump as proof that the addon had already rebuilt its owners, so it skipped. That skip is the only branch that ever ran during the corruption, which is why the log is silent while the screen is visibly wrong.

Two experiments on top of that:

- Dropping only the owners’ render models on each bump: **still corrupt**.
- Full coherent rebuild (texture wipe, then models, then refresh): **clean**.

## The fix

A generation advance becomes a CURE trigger instead of a reason to stand down, and the generation is re-baselined from the *post-rebuild* atlas — our own wipe bumps it too, and reading the pre-wipe value back would rebuild on every poll forever (the first draft did exactly that; the new unit test pins it).

## Cost

Edge-triggered, not level-triggered. Two full floods on the fixed build fired **one** rebuild in total and rendered clean throughout; the unfixed build corrupted every time.

## Tests

- `atlasGuard.test.ts`: the two tests that asserted the skip now assert the rebuild, plus a new one that pins "one rebuild per bump, and none while the generation holds steady".
- 27 terminal-related suites, 485 tests, green. `tsc --noEmit` clean, eslint clean on the touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed scrambled CJK characters during heavy terminal output.
  - Improved WebGL glyph rendering recovery when the glyph cache reaches capacity or reorganizes.
  - Terminal content now refreshes automatically after cache evictions, preventing stale character positions from remaining visible until a window resize.
  - Recovery retries failed repairs and avoids unnecessary repeated refreshes during stable rendering.


openwong2kim:
Panel review (Claude + GLM; Codex is 402 `deactivated_workspace` on this machine and a Grok pass stalled without emitting findings) landed three fixes on top of the first commit:

- **`recoverNow` never re-baselined the generation** (both models, CRITICAL). Its own wipe bumps it, so with this PR every sleep/wake recovery would be followed a poll later by a second full re-raster, logged as a self-eviction that never happened.
- **A `failed` wipe consumed the bump anyway** (both models, CRITICAL). Nothing was repaired, and the pane would stay scrambled until the atlas happened to evict again — never, on a quiet pane. The baseline is kept on failure and the next poll retries.
- **Nothing bounded the new CURE path** (Claude, CRITICAL). A rebuild re-rasters the viewport and re-mints every visible glyph, so under a flood the repair can feed the eviction that triggers it. Three consecutive rebuilding polls now fall back to the existing cooldown.
- Cure line says `merge` when the removal latch fired (GLM, INFO), so field triage can still separate a real page merge from a sampler eviction.

Four tests added for exactly those paths: two evictions inside one poll gap collapse to one rebuild, a failed wipe retries, the streak backs off and recovers after the cooldown, and `recoverNow` leaves no duplicate rebuild.

Not taken, with reasons: partial group coverage when a pane sharing the atlas is not registered yet (pre-existing on the PREVENT/CURE paths, not introduced here) and an unconditional rebuild on the first poll of a newly registered pane (would re-raster every pane at startup for a case the next eviction repairs).

Live after the fixes: same Hangul flood, one pane and four panes, clean in both, 1 and 2 rebuilds. 489 terminal-suite tests green, tsc and eslint clean.

PR #1332, opened 2026-09-16
fix(daemon): phone activity resets the idle clock (#1316)

### LINKED ISSUE #1316: idle.timeout kills the daemon during phone-only use — HTTP/push activity does not reset the 5-minute idle clock
## 현상
폰(wmux-ios)만으로 데몬을 사용하는 시나리오에서 `idle.timeout` 5분이 지나면 서버가 꺼진다. 폰의 HTTP 요청·APNs 푸시 활동이 idle 타이머를 리셋하지 않는다.

## 재현
1. 데스크톱 wmux 클라이언트를 종료(터미널 창 없음)
2. 폰 앱으로 페인 열기·채널 읽기·승인 응답 등 활발히 사용
3. 5분 후 데몬 종료 → 폰은 연결 끊김

## 기대
폰 발신 활동(인증된 API 호출 최소한)은 idle로 세지 않아야 한다.

## 배경
wmux-ios 채널 도그푸드(2026-09-15)에서 발견. 심각도 [높음].

openwong2kim:
Fixes #1316.

## The bug

The daemon's idle-shutdown window was anchored to one signal: the moment the
last client let go of the **control pipe** (`DaemonPipeServer.getLastDisconnectAt`).
That was a fair proxy for "is anyone using me?" while the desktop app was the
only client. A phone is not on the pipe at all.

So in phone-only use — desktop closed, no live PTY left behind — someone could
be reading channels and answering approvals from wmux-ios continuously and the
daemon still counted itself abandoned. Five minutes in, `idle.timeout` fired and
the phone lost its connection.

## The fix

`WebTerminalServer` stamps `lastApiActivityAt` once, right after the auth gate
in `handleApi` — so every authenticated route is covered and no route can forget
it. The daemon's `onIdleCheck` folds that into the **same** anchor as the pipe's
`lastDisconnectAt` via `newerTimestamp`, rather than running a second clock that
could disagree with the first.

Two things this deliberately does **not** do:

- **Unauthenticated traffic does not count.** `/api/pair` and the static shell
  are reachable by anything that can open the port; "a stranger can keep my
  daemon alive" is not a property worth having.
- **A silent SSE viewer still does not hold the daemon up.** That decision is
  already recorded beside `pendingApprovals` in `onIdleCheck` and this PR leaves
  it alone. It costs nothing to leave alone: a phone watching a live pane is
  watching a live PTY, which `sessions` already covers.

## Tests

- `idleAnchor.test.ts` — the null semantics, which are the part that could go
  wrong quietly: `null` means "this source never saw anybody", not a timestamp
  of zero, so folding it in must not drag the anchor back to the epoch.
- `WebTerminalServer.test.ts` — no stamp before any request, none from a 401,
  none from `/api/pair`, a stamp from an authenticated call.

## Summary by CodeRabbit

* **Bug Fixes**
  * Phone activity now keeps the daemon running when users read channels or respond to approvals with the desktop app closed.
  * Unauthenticated requests do not prevent idle shutdown, preserving existing security behavior.


openwong2kim:
## Live dogfood

Against the **bundled daemon** (`dist/daemon-bundle/index.js`) in an isolated
state dir, `WMUX_IDLE_SHUTDOWN_MS=6000`, no client on the pipe and no live PTY —
the reported shape. `/api/config` is poked once a second with the operator token
for 18s (3× the idle window), then the poking stops.

**Before (`main`):**

```
FAIL A. authenticated phone traffic holds the daemon open past the idle window
      idle window 6000ms, poked for 18000ms (7 requests, last HTTP 200)
      alive after poking: false (expected true)
      shutdown reason: [shutdown.phase] idle.timeout idleMs=6431 cfgMs=6000
```

Seven answered, authenticated requests and the daemon still exited under them —
exactly the bug.

**After (this branch):**

```
PASS A. authenticated phone traffic holds the daemon open past the idle window
      idle window 6000ms, poked for 18000ms (18 requests, last HTTP 200)
      alive after poking: true (expected true)
      exited after going quiet: true
      shutdown reason: [shutdown.phase] idle.timeout idleMs=6055 cfgMs=6000

PASS C. a stranger with a bad token does NOT hold it open
      idle window 6000ms, poked for 18000ms (6 requests, last HTTP 401)
      alive after poking: false (expected false)
```

All 18 requests survive, and the daemon still dies ~6s after the traffic
stops — the fix does not trade a five-minute timer for an immortal daemon. A
caller presenting a bad token is refused and does not hold it open, so the
unauthenticated carve-out is load-bearing rather than decorative.

Driver: `scripts/_1316-idle-live.mjs` (untracked, like the repo's other `_*`
dogfood scripts).

PR #1333, opened 2026-09-16
fix(browser_replay): an unnamed smart step stops when its element is gone (#1300)

### LINKED ISSUE #1300: browser_replay: a smartRef step reports success when its target element is gone
Found while dogfooding the per-site memory stack (#1298/#1299) on an isolated instance with the chrome backend.

## What happened
A flow step recorded on the `smartRef` axis, replayed on a page where the target element no longer exists, still reports `ok … typed into textbox "…"`. The run counts as a success.

## Why it matters
- The replay result is wrong: nothing was typed into the intended field.
- With #1299, a successful run bumps the domain's `provenFlowCount` (observed rising to 2 on a page whose form was removed), so the per-site memory learns a false 'proven' flow.

## Expected
A smartRef step whose element cannot be resolved on the live page should STOP the run like the css axis does (`matched 0 elements, expected exactly 1`), not resolve to a different element or a stale ref.

## Repro shape
1. Save a flow that types into a form field, recorded with smartRef.
2. Remove the field from the page (or navigate to a variant without it).
3. `browser_replay {action:"run"}` → step reports ok.

Not caused by #1298/#1299; filed separately so the memory feature's proven-count is not trusted until this is fixed.

openwong2kim:
Fixes #1300.

## The bug

A `smartRef`-recorded step whose element had been removed replayed as `ok …
typed into textbox "…"`. Nothing was typed into the intended field, and with
#1299 a successful run bumps the domain's `provenFlowCount` — so the per-site
memory learned a "proven" flow that had not actually run.

## Root cause

`smartRefAxisEntry` recorded the axis like this:

```ts
sameNameIndex: named ? element.sameNameIndex : element.roleIndex,
sameNameTotal: named ? element.sameNameTotal : element.roleTotal,
```

The fallback is correct for `resolveSmartRefLocator` a few lines above it,
which locates through `getByRole(role)` with **no** name filter and therefore
counts every same-role element. It is wrong for the only consumer this axis
has. The replay runner builds its live population as

```ts
entry.role === axis.role && entry.name === axis.name && frameKey matches
```

— role **and** name. So an unnamed axis carried a role-space index into a
name-space lookup.

`matchRefAxis` then exempted unnamed axes from the population-size check, with
a comment naming exactly this substitution as the reason ("their stored total
is not a same-name count at all"). That left `indexLookup` — pure position — as
the only thing between a removed element and whichever unnamed element had
taken its slot.

Concretely: a page with `button "Save"` plus three unnamed buttons. The second
unnamed button is role-index 2 of 4, name-index 1 of 3. Record a step on it,
remove it, replay: the recorded index 2 no longer exists in name space… but on
a page with one more unnamed sibling it does, and it belongs to a different
element. Resolved, acted on, reported ok.

## The fix

Both halves, because either one alone leaves the other's hole open:

1. `smartRefAxisEntry` records the same-name pair unconditionally.
   `finalizeSmartPopulations` already computes it for the unnamed population —
   it keys on `role\0name`, so "no name" is a real population, not a degenerate
   one. There was never anything to fall back *from*.
2. `matchRefAxis` count-checks every axis. The exemption existed only because
   of (1).

An unnamed smart axis keeps the same escape hatch a named one has: when the
snapshot ref map and the smart walk genuinely enumerate different sets, the
smart lane's own recount (`countSmartNamedPopulation`, which takes the empty
name as a real key) settles it, exactly as for a named axis.

## Consequence, accepted

An unnamed smart step recorded **before** this change carries a role-space
total, so its next replay stops instead of resolving. There is no trace version
to gate on and no way to tell such an axis from a correct one.

That stop is `inconclusive: true`, so it stays out of the failure streak that
quarantines a trace, and the agent finishing the flow live is also the next
recording — which lands correct. Stopping is what the issue asks for, and it is
the opposite failure from acting on the wrong element.

## Tests

- `dom-intelligence.refstability.test.ts` — the axis for an unnamed element
  carries the name-space pair while the element itself still reports the
  role-space one, so the two enumerations are visibly distinct in the fixture.
- `replayRunner.test.ts` — an unnamed step whose population no longer matches
  stops (and nothing is typed); an unnamed step the smart lane's recount
  vindicates still runs. The existing test that asserted the exemption is
  rewritten to the new contract.

## Summary by CodeRabbit

- **Bug Fixes**
  - Replay steps targeting unnamed elements now verify that the page structure has not changed before acting.
  - Replays stop and report that the page changed instead of typing into a neighboring or incorrect field.
  - Smart replay continues when the recorded element population still matches, even if other elements have been reordered.
  - Flows recorded before this fix may stop once during replay; completing the flow live will record the corrected behavior.


openwong2kim:
## Live dogfood

Real browser (Edge over CDP), real accessibility tree, the real
`getSmartSnapshot` → `smartRefAxisEntry` → `replayTrace` path. The unit tests
mock the page and the ref map, which is exactly where this bug lived: the two
lanes disagreed about what the recorded numbers *count*, and a fake page cannot
disagree with itself.

Fixture — one named textbox (name from a **wrapping** `<label>`) plus four
unnamed ones, and **no** `id` / `name` / `aria-label` / `data-testid` anywhere.
That last part matters: my first fixture gave the inputs ids, and the replay
stopped on the own-attribute verifier for a reason unrelated to this bug — it
passed on the old code too. With no attributes to verify, the population count
is the only evidence left, which is the thing this PR changes.

The recorded target is the 2nd unnamed input: role-space index 2 of 5,
name-space index 1 of 4. It is then removed and the flow replayed.

**Before (`main`):**

```
element   : roleIndex=2/5 sameNameIndex=1/4
axis says : sameNameIndex=2/5  (via=smart)
axis carries ROLE-space numbers (the bug): true
replay ok=true inconclusive=false
step    : true — typed into textbox "" (#3 of 5)
fields after replay: #0="" #1="" #2="hunter2" #3=""
```

The recorded field is gone, and `hunter2` went into **#2 — a different field** —
while the step reported success. That is the issue verbatim, and the run that
would have bumped `provenFlowCount`.

**After (this branch):**

```
element   : roleIndex=2/5 sameNameIndex=1/4
axis says : sameNameIndex=1/4  (via=smart)
axis carries ROLE-space numbers (the bug): false
replay ok=false inconclusive=true
step    : false — textbox "" (#2 of 4) can no longer be identified by position:
          the page has 3 element(s) with that role and name, the recording had 4
          — element(s) were removed, so position 2 no longer counts the same
          population
fields after replay: #0="" #1="" #2="" #3=""
```

Stops, says why, and nothing is typed anywhere.

Driver: `scripts/_1300-smartref-live.mjs` (untracked, like the repo's other `_*`
dogfood scripts).

PR #1334, opened 2026-09-16
fix(cdp): claim the debugging port, and say whether it came up (#1331)

### LINKED ISSUE #1331: A second wmux instance can silently start with no CDP port
Found while running two instances side by side on macOS (3.55.0 and a local build, separate `WMUX_DATA_SUFFIX`).

`src/main/index.ts` picks the remote-debugging port as `18800 + crypto.randomInt(100)` and logs it:

```
[WinMux] CDP enabled on port 18885
```

The second instance drew 18885 as well, which the first one already held. Chromium could not bind it, and the process came up with **no listening CDP port at all** — `lsof -nP -iTCP -sTCP:LISTEN -p <pid>` was empty — while the log still said "CDP enabled on port 18885". Everything that needs CDP (browser automation, screenshots, DOM snapshots) is then unavailable with no error anywhere.

Two things worth fixing:
1. Verify the port is actually listening after startup, and retry with another port on failure rather than trusting the draw.
2. Do not log "CDP enabled on port N" until that check passes; log the failure otherwise.

Low priority for single-instance users, but anyone running an isolated dogfood instance next to their real one hits it at roughly 1 in 100 launches, and the symptom (browser tools mysteriously dead) gives no hint of the cause.

openwong2kim:
Fixes #1331.

## The bug

```ts
cdpPort = 18800 + crypto.randomInt(100);
app.commandLine.appendSwitch('remote-debugging-port', cdpPort.toString());
console.log(`[WinMux] CDP enabled on port ${cdpPort}`);
```

Three lines, two problems. The draw can collide with another wmux instance —
about once in a hundred launches when you run a dogfood build next to the real
one — and when it does, Chromium cannot bind and comes up with **no listening
CDP port at all**. The third line says "enabled" regardless. Browser
automation, screenshots and DOM snapshots are then dead, and the only route to
the cause is `lsof` on the process.

## The fix

**1. The port is claimed, not just drawn.** `claimCdpPort` walks the range from
a random offset — the draw stays unpredictable, which is the reason it was
randomized — and skips any port a *live* wmux already holds. The claim is a pid
written into a tmpdir file with `openSync(…, 'wx')`.

- **tmpdir, not `~/.wmux{suffix}`**: the collision in the report was between two
  instances under *different* data suffixes. A per-instance directory would
  leave each invisible to the other, which is the bug.
- **pid, not a lock to release**: a claim whose pid is gone is reclaimed, so
  nothing has to run at exit. That matters because a crash is precisely when no
  exit handler runs.
- **sync by necessity**: the port must be chosen before `appendSwitch`, inside
  the synchronous `appInit()`. There is no await there to hang a `net.listen`
  probe on, and making boot async for one would reorder far more than this is
  worth. `openSync(…, 'wx')` is atomic and is the shape available.

**2. Nothing claims the port is enabled until it answers.** The pre-ready line
now says `CDP requested on port N` — which is all that is known before Chromium
starts. After ready, a probe of `/json/version` (the same surface
`WebviewCdpManager` already talks to) logs either the browser that answered or
an error naming the likely cause and the consequence:

```
[WinMux] CDP is NOT listening on port 18885 (connect ECONNREFUSED …). Chromium
was asked for this port and did not get it — most often another process,
commonly a second wmux instance, already holds it. Browser automation … will be
unavailable until wmux is restarted.
```

A claim file is a weaker check than a bind — it only knows about wmux — and
that split is deliberate: the claim prevents the collision that can be
prevented, and the probe is the thing that decides what gets logged.

## What this does not do

The issue asks to "retry with another port on failure". There is no in-process
retry available: `--remote-debugging-port` is consumed at Chromium startup, so
changing the port means relaunching the app. Restarting the whole app under the
operator to recover a *debugging* socket is worse than a diagnosis they can act
on, so the failure is named loudly instead. Happy to add the relaunch if you
want it.

## Tests

`cdpPort.test.ts` — the claim steps past a live holder (the reported
collision), reclaims a dead one's port (the crash case), wraps at the top of the
range, still returns a port when everything is held (a boot must not fail over a
debugging socket) and refuses to steal a claim it cannot parse — including the
empty file another instance leaves for the instant between creating it and
writing its pid, which is the race that would put two processes on one port
again. The probe reports the browser, a refused connection, and a non-200.

## Summary by CodeRabbit

- **New Features**
  - Improved browser debugging port coordination to prevent conflicts between running instances.
  - Verifies that the debugging endpoint is responding before reporting it as available.
  - Reports clear status and failure reasons when the debugging port cannot be used.

- **Bug Fixes**
  - Updated startup and diagnostics tooling to recognize current and legacy debugging-port status messages.
  - Prevented false claims that browser debugging is enabled when the endpoint is unavailable.


openwong2kim:
## Live dogfood

Against a **packaged build of this branch** (`npm run package`), not the dev
tree.

### 1. Two instances side by side, different `WMUX_DATA_SUFFIX`

The exact configuration from the report — separate temp HOMEs, started ~300ms
apart:

```
[a] requested line: CDP requested on port 18815
[b] requested line: CDP requested on port 18875
[a] verdict line : CDP listening on port 18815 (Chrome/146.0.7680.216)
[b] verdict line : CDP listening on port 18875 (Chrome/146.0.7680.216)
[a] says "CDP enabled": false
[b] says "CDP enabled": false
ports: [18815,18875] distinct=2
claim files: ["wmux-cdp-18815.lock=6060","wmux-cdp-18875.lock=17180"]
instance pids: [6060,17180]
[a] GET /json/version on 18815 -> 200 Chrome/146.0.7680.216
[b] GET /json/version on 18875 -> 200 Chrome/146.0.7680.216
```

Distinct ports, claim files holding the right pids **across the suffix
boundary**, and both ports really answer.

### 2. The failure path — the one the report is about

A plain TCP listener is invisible to the claim file (that only knows about
wmux), so occupying the range forces Chromium to fail its bind the same way a
colliding instance did:

```
occupied 98/100 ports in the CDP range
LOG | [WinMux] CDP requested on port 18885
LOG | [WinMux] CDP is NOT listening on port 18885 (This operation was aborted).
      Chromium was asked for this port and did not get it — most often another
      process, commonly a second wmux instance, already holds it. Browser
      automation (MCP browser tools, screenshots, DOM snapshots) will be
      unavailable until wmux is restarted.
claims "enabled": false
admits not listening: true
```

It even drew 18885 — the port from the issue. Before this PR that boot would
have printed `CDP enabled on port 18885` and left every browser tool dead with
no hint.

### 3. The claim under real concurrency

Twelve **separate OS processes** claiming at once, each holding its claim open
for 10s so they genuinely coexist:

```
12 distinct of 12
```

For contrast, the old bare draw over 2000 simulated 12-instance runs:
`46.7% of runs collide`.

### One thing this found

The probe answers *"is anything listening"*, not *"is it ours"*. A second wmux
holding the port would answer 200 and read as success — the claim file is what
actually covers that case, so the pair is only defeated by a claim file deleted
out from under a running instance. Recorded in the code (b8e9083) rather than
left for a reader to rediscover.

Drivers: `scripts/_1331-two-instances.mjs`, `scripts/_1331-bind-failure.mjs`,
`scripts/_1331-claim-race.mjs` (untracked, like the repo's other `_*` dogfood
scripts).

PR #1335, opened 2026-09-16
fix(pwsh): report the exit code of the command that actually ran (#1270, #1269)

### LINKED ISSUE #1269: pwsh local-mode hook never restores $LASTEXITCODE
Follow-up from #1268 (reported by the PR author, confirmed in review).

`src/main/pty/shell-hooks/pwsh.ps1` runs `git rev-parse` on every prompt render and never puts `$global:LASTEXITCODE` back. After `cmd /c exit 7` the user sees `$LASTEXITCODE = 0` (or 128 outside a git repo).

Since #1268 the wrapped prompt now correctly sees `$? = $false`, so prompt engines that fall back to `$LASTEXITCODE` when no `$Error` record matches can now render wmux's own git exit code instead of the user's.

The daemon-mode wrapper in `src/daemon/shell-integration.ts` already handles this: snapshot `$__wmux_le = $LASTEXITCODE` next to `$__wmux_ok = $?` and restore `$global:LASTEXITCODE = $__wmux_le` before returning. The local-mode hook should do the same.

### LINKED ISSUE #1270: pwsh shell integration emits 133;D;0 after a failing cmdlet
Follow-up from #1268 (reported by the PR author).

In `src/daemon/shell-integration.ts`:

```powershell
$ec = if ($null -ne $__wmux_le) { $__wmux_le } elseif ($__wmux_ok) { 0 } else { 1 }
```

`$LASTEXITCODE` is preferred over the `$?` snapshot, and it is non-null from the first native command onward, so the `$__wmux_ok` branch is effectively dead for the rest of the session. `Get-Item Q:\nope-xyz` after any earlier native command emits `133;D;0`. This feeds `agent.lifecycle.exitCode`.

VS Code's integration checks `$?` first. Suggested order: if the `$?` snapshot is false and `$LASTEXITCODE` is 0/null, report 1; otherwise report `$LASTEXITCODE`. Needs a runtime test in `shellIntegration.runtime.test.ts` (failing cmdlet after a successful native command).

openwong2kim:
Fixes #1270. Fixes #1269.

Two follow-ups from #1268, both about `$LASTEXITCODE` in a prompt wrapper. They
are one PR because they are the same mistake in the two copies of that wrapper —
fixing one and not the other is how they drifted apart in the first place.

## #1270 — the daemon's OSC 133 integration reports the wrong command

```powershell
$ec = if ($null -ne $__wmux_le) { $__wmux_le } elseif ($__wmux_ok) { 0 } else { 1 }
```

`$LASTEXITCODE` is set **only** by native commands and is never cleared, so from
a session's first native command onward it is non-null for good — and the
`elseif ($__wmux_ok)` branch is dead for the rest of the session. Every failing
cmdlet after that point emitted `133;D;0`. `Get-Item Q:\nope-xyz` reported
success, and that number is what `agent.lifecycle.exitCode` is built from.

The obvious repair — check `$?` first — trades the bug for its mirror image.
After `cmd /c exit 7`, a cmdlet that **succeeds** leaves `$LASTEXITCODE` at 7,
so reporting it marks a successful command failed. Neither variable can answer
alone.

What separates the two cases is whether `$LASTEXITCODE` **moved** since the last
prompt:

| | moved | did not move |
|---|---|---|
| meaning | a native command ran | the last command was a cmdlet |
| exit code | `$LASTEXITCODE` | `$?` → 0 or 1 |

The residual case, stated in the code: two native commands in a row exiting with
the **same** code. The second one's number has not moved, so its failure reports
as 1 rather than its own code. Both are non-zero, the consumer is a
failed/succeeded signal, and exit 1 — far and away the most common failure code
— reports as exactly itself.

`INTEGRATION_VERSION` goes to 11 so installed copies are rewritten. That is not
bookkeeping: without the bump the v10 script on disk keeps running, which is
exactly how the new runtime tests first failed — reproducing both directions of
the bug against the old script before passing against the new one.

## #1269 — the local-mode hook eats the user's `$LASTEXITCODE`

`src/main/pty/shell-hooks/pwsh.ps1` runs `git rev-parse` on every prompt render
and never puts `$LASTEXITCODE` back, so after `cmd /c exit 7` the user's own
shell reads 0 — or 128 outside a git repo.

Since #1267 this is worse than harmless. The wrapped prompt now correctly sees
`$? = $false`, so a prompt engine that falls back to `$LASTEXITCODE` when no
`$Error` record matches renders **wmux's** git exit code as the user's. The fix
is the one the daemon-mode wrapper already had: snapshot beside `$?`, restore as
the last statement before the return.

## Tests

- `shellIntegration.runtime.test.ts` — real ConPTY, both directions: a failing
  cmdlet after a native command reports non-zero, and a succeeding cmdlet after
  a failing native command reports zero (so the naive `$?`-first fix cannot
  pass). Both anchor on their own `command_start`; the startup prompt render
  emits a `D;0` of its own and matching that reads the wrong command's code.
- `pwshHook.test.ts` — the snapshot sits beside `$?` and the restore is the last
  statement before the return.


openwong2kim:
## Live dogfood

### #1270 — real ConPTY

The runtime suite spawns a real `powershell.exe` through ConPTY and reads the
actual OSC 133 bytes, so this half was live from the start. Worth stating how it
was confirmed non-vacuous: both new tests **failed first**, against the v10
script still on disk, and each failure was the bug itself —

```
AssertionError: expected +0 to be 1     ← failing cmdlet after a native command
AssertionError: expected 7 to be +0     ← succeeding cmdlet after `cmd /c exit 7`
```

Both pass after the `INTEGRATION_VERSION` bump rewrites the installed script.
The second is the reason `$?`-first is not the fix: it is the same bug pointed
the other way.

### #1269 — real PowerShell 5.1

Dot-source the hook, run `cmd /c exit 7`, render the prompt once (what the REPL
does before the next command), then read `$LASTEXITCODE`:

| | before | after |
|---|---|---|
| inside a git repo | 7 → **0** | 7 → **7** |
| outside a git repo | 7 → **128** | 7 → **7** |

Both numbers from the issue reproduced exactly — 0 from `git rev-parse`
succeeding, 128 from it failing outside a repo — and both fixed. (pwsh 7 is not
installed on this box; 5.1 is the engine the issue was filed against.)

Driver: `scripts/_1269-hook-exitcode.ps1` (untracked, like the repo's other `_*`
dogfood scripts).

Issue #1356, opened 2026-09-17
browser_snapshot with a text query takes 5-38 s while a CSS selector scope takes 0.6 s


openwong2kim:
## Summary

`browser_snapshot` with a text query (`q`) is very slow. Narrowing with a CSS selector is ~10-60x faster on the same page.

## Measured (dogfood 2026-09-17)

| Mode | Time |
|---|---|
| `q` with a dropdown open | 38 s |
| `q` normally | 5-16 s |
| CSS selector scope | 0.6 s |

## Expected

Query-filtered snapshots should be in the same order of magnitude as selector-scoped ones. Likely the query path walks / serializes the full tree before filtering.

## Pointers

- `src/mcp/playwright/tools/inspection.ts` (`browser_snapshot`)
- `src/mcp/playwright/dom-intelligence.ts`

## Source

Dogfood report against a real-world web app, 2026-09-17. Severity #3.

openwong2kim:
Fixed by #1370, landed in main through #1376 (289d1cd5). A literal q now narrows the CDP fetch with DOM.performSearch instead of serialising the whole accessibility tree. Verified live against Chrome 141 on a 10k-node page: 105 ms against 376 ms for the full-tree path, with a byte-identical body. Follow-up for the selector lane is #1371.

PR #1370, opened 2026-09-17
perf(browser_snapshot): a text query narrows the CDP fetch, not the result (#1356)

openwong2kim:
Stacks on #1366 (`fix/1358-screenshot-scale`) — both touch
`src/mcp/playwright/tools/inspection.ts`, so this branches from it and targets
it. Review the top commit only; merge #1366 first.

## Profile first

A 5 000-node fixture is committed (`__tests__/fixtures/large-page.html`, 500
rows x 7 nodes); `SNAPSHOT_BENCH_SCALE=10` grows it to ~35 000 accessibility
nodes, the size the dogfood page had with a dropdown open. Real Chrome 141,
this machine, 2026-09-17:

| Stage | 3 506 nodes | 35 006 nodes |
|---|---|---|
| `Accessibility.getFullAXTree` | 273 ms | **9 872 ms** |
| `DOM.performSearch` + `getSearchResults` | 21 ms | 119 ms |
| `Accessibility.getPartialAXTree` (one hit) | 6 ms | 12 ms |
| `getDomFacts` | 66 ms | 767 ms |
| `collectOcclusion` | 31 ms | 51 ms |
| `collectPageFacts` | 74 ms | 131 ms |
| buildTree + `q` prune + serialise (derived) | < 30 ms | < 30 ms |
| **`generateSnapshot({q})` end to end** | 424 ms | **11 079 ms** |
| `generateSnapshot({})` end to end | 454 ms | 11 071 ms |

So the hypothesis in the plan splits cleanly: serialisation does **not**
dominate — everything downstream of the fetch is under 30 ms even at 35 000
nodes, so filtering during the walk would have bought nothing. The CDP fetch
dominates, and it is superlinear: 10x the nodes cost 36x the time. `q` was
paying for the whole document before it had anything to filter.

(The dogfood's 0.6 s selector scope is the same lane's DOM-listing fallback —
`generateScopedSnapshot` also fetches the full tree, and measured 5 266 ms on
the 35 000-node page. That lane is untouched here.)

## After

| | 35 006 nodes |
|---|---|
| `q` "Unobtainium" before | 11 079 ms |
| `q` "Unobtainium" after | **271 ms** |

41x on this page, and the gap grows with the page, because what is left is no
longer proportional to the document.

## Cause

`inspection.ts` passes `q` to `generateSnapshot`, which fetches the whole
accessibility tree through `Accessibility.getFullAXTree` and only then prunes
it to the matches and their ancestors (`snapshot.ts`, `pruneChildrenToQuery`).
The prune is cheap; the fetch is the bill.

## Fix

A literal `q` narrows the fetch instead: one `DOM.performSearch` for the text
(the call `getPasswordFieldBackendIds` already uses), then one
`Accessibility.getPartialAXTree` per hit, which comes back with the hit's
ancestor chain in it. Matched nodes' subtrees are completed outwards with
`Accessibility.getChildAXNodes`. Everything after that is the code that ran
before, unchanged — the same `buildTree`, the same prune, the same ref
numbering off `backendDOMNodeId` — so the output is the same output: same
ancestors, same ref numbers, same lines. A unit test asserts that byte for
byte against the full-fetch route, and a real-Chrome check confirmed it on the
fixture.

The route abstains — and the full fetch runs exactly as it does today —
whenever it cannot promise that:

- a `/regex/` query (DOM search takes literal text only);
- a document containing an `<iframe>`/`<frame>` (frame nodes reach the tree
  only through the graft, which needs the full fetch);
- zero search hits — a `q` naming a **role** ("button") is invisible to a text
  search, and that is exactly the query whose answer must not silently shrink;
- more than 200 hits or more than 2 000 fetched nodes, past which the round
  trips cost more than the one big fetch they replace;
- any CDP failure, including the RPC lane, where `performSearch` is not there.

Tool description updated: `selector` is named as the cheapest way to narrow a
big page, and `q` as a text filter that costs about as little when it is
literal text and still reads the whole tree for a regex or a framed page.

## Tests

- `snapshot.querysearch.test.ts` (new, 7 tests, runs in CI): the search route's
  output equals the full route's output for the same query; ref numbers are
  identical; ancestors and the matched subtree are kept; each of the four
  abstain conditions falls back and still answers; "no nodes match" is
  unchanged.
- `snapshot.querybudget.test.ts` (new, skipped without `CHROME_PATH`): the
  real-Chrome correctness check that `q` and the equivalent `#needle-row`
  selector scope list the same rows with the same refs, plus the budget
  assertion that `q` costs under half an unfiltered snapshot. Run it with
  `CHROME_PATH=... SNAPSHOT_BENCH_SCALE=10 npx vitest run
  src/mcp/playwright/__tests__/snapshot.querybudget.test.ts` — the header
  comment says so.
- `npx vitest run src/mcp` → **1455 passed, 2 skipped**.
- `npx vitest run src/mcp/playwright/__tests__/inspection*
  src/mcp/playwright/__tests__/dom-intelligence* src/mcp/playwright/__tests__/snapshot*`
  → 260 passed, 1 skipped.
- Budget test with real Chrome at `SNAPSHOT_BENCH_SCALE=10` → passed.
- `npx tsc -p tsconfig.mcp.json --noEmit` and `npx tsc -p tsconfig.json
  --noEmit` → clean.

Closes #1356


openwong2kim:
Landed in main through #1376, which merged this branch along with the other nine. GitHub cannot mark this one merged because its base was fix/1358-screenshot-scale rather than main; retargeting is refused with There are no new commits between base branch 'main' and head branch 'fix/1356-snapshot-query-speed', which is the confirmation that every commit here is already in main (289d1cd5).

PR #1396, opened 2026-09-17
fix(daemon): gate the orchestrator brain pane per credential class

### LINKED ISSUE #1388: GET /api/stream resolves panes with a bare getSession — a paired device can stream a brain pane's raw bytes
## What

\`handleStream\` in the daemon web server resolves the requested pane with a bare \`getSession\`, without the brain-pane exclusion that \`/api/sessions\` and the transcript routes apply (\`readableSession\`). A paired phone device that guesses or learns an orchestrator brain pane id can open \`GET /api/stream?session=<brain-id>\` and receive that pane's raw terminal bytes.

Found while implementing #1315 in #1386. That PR deliberately did **not** widen the hole (liveness delivery on the stream takes the \`readableSession\` gate) and did not fix it either, because the fix is not a one-liner.

## Why it is not a one-line fix

The operator token legitimately uses the same route to stream any pane, including brain panes. Adding \`readableSession\` to \`handleStream\` unconditionally would cut the operator off. The gate has to depend on the caller's credential class (device vs operator), which the route does not currently distinguish at that point.

## Proposal

Apply the brain-pane exclusion in \`handleStream\` only for device-class credentials, keep operator-token behaviour unchanged, and add a test that a device stream request for a brain id answers 404/403 while an operator request still streams.

Related: #1315, #1386.

openwong2kim:
Fixes #1388

## The hole

`handleStream` resolved the requested pane with a bare `getSession`, which answers existence and nothing else. A paired phone that learned an orchestrator brain pane id (the `brain-` prefix is guessable, and an approval or notify event can carry one) could open `GET /api/stream?session=<brain-id>` and receive the orchestrator's raw terminal bytes.

Auditing the rest of the surface, the same hole ran through **every** route that takes a caller-supplied session id, not just the stream:

| Route | Before | After (device) | After (operator) |
|---|---|---|---|
| `GET /api/stream` | bare `getSession` | 404 | unchanged |
| `POST /api/input` | bare `getSession` | 404 | unchanged |
| `GET /api/sessions/:id/diff` | bare `getSession` | 404 | unchanged |
| `GET /api/sessions/:id/commands` | bare `getSession` | 404 | unchanged |
| `POST /api/sessions/:id/resize` | bare `getSession` | 404 | unchanged |
| `DELETE /api/sessions/:id` | bare `getSession` | 404 | unchanged |
| `GET /api/sessions/:id/turns[/block]` | flat `readableSession` | unchanged | unchanged |

So a phone could not only read the brain pane but type into it, read its diff, SIGWINCH it, and close it.

`POST /api/upload` and `POST /api/stream-ticket` take no session id, so neither carries this hole.

## Why it was not a one-liner

`readableSession` (the existing brain-pane exclusion) is **flat**: it refuses the operator too. That is right for a transcript, but the operator legitimately streams, types into, diffs and closes a brain pane from their own desk — and an existing `★` test asserts the operator gets a 200 streaming a brain id. Switching `readableSession` on at these routes would have cut the operator off from their own orchestrator.

The gate therefore keys on the **credential class**:

```ts
private addressableSession(sessionId, principal) {
  if (principal.kind === 'operator') return this.deps.sessionManager.getSession(sessionId);
  return this.readableSession(sessionId);
}
```

The operator arm is byte-identical to the call that was already there, so operator behaviour is unchanged on every route. A stream ticket resolves to a device principal, so that path is covered with no code of its own. Refusals answer the identical `404 {"error":"session not found"}` a missing pane does — a distinct error would confirm the id.

The method's doc comment carries the route-to-gate table, so the next route that takes a caller-supplied pane id has somewhere to look. Picking no gate is what this issue was.

## The other half: `broadcastEvent`

Found while reviewing the above, and fixed here because leaving it would have made this PR a user-visible regression.

`DaemonSessionManager` re-emits `session:critical` / `session:notification` for **every** pty, the brain included, and that fan-out is not keyed by pane: it wrote to every open stream and recorded the entry in the replay window `GET /api/events` serves (whose backlog route takes no principal at all). Each entry carries the pane id **plus pane-authored text** — the critical detector's `matchedLine`, an OSC 9/777 `title` and `body`.

Two consequences:

1. Refusing the id at the pane routes while the daemon kept handing out fresh ones closes the door and leaves the sign up. `addressableSession` names this channel as the way an id leaks.
2. With the pane routes gated but this one open, a phone that tapped a brain toast would call `/api/stream?session=brain-…`, get the new 404, and `EventSource` would go to `CLOSED` **without retrying** — a permanently dead tile with no row in `/api/sessions` to explain it.

`broadcastEvent` now refuses brain panes at the producer, mirroring the gate `emitAgentLiveness` already took in #1386. It is a **brain** filter rather than `readableSession`'s brain-or-unknown one: these events come from a live pty, so an id the manager no longer knows is a pane that closed while its event was in flight — a race, not an invention — and dropping those would lose real signals.

## Deliberately not changed

- `/api/sessions/:id/turns[/block]` keep the flat `readableSession`. Moving them to the new gate would **widen** operator access to brain transcripts, which is not this fix's job and is locked by an existing `★` test.
- `emitAgentLiveness`'s own gate (#1386, merged today) is untouched, so `agent.liveness` delivery on the pane stream still refuses the brain pane on both sinks.
- The lookups that follow a side effect stay bare `getSession`: the applied-geometry read-back at the end of `handleSessionResize` and the resize debounce inside an already-open `handleStream`. Gating those would answer 404 for work that had already happened — a lie about state rather than a refusal. The resize re-read that happens *before* the pane is acted on does gate, so authorization is re-checked immediately before the side effect rather than trusted from whenever the headers landed.

## Tests

Six tests in `src/daemon/web/__tests__/WebTerminalServer.test.ts`, all covering **both** `isBrainPty` marks (the `WMUX_BRAIN_PTY=1` env marker and the `brain-` id prefix) on separate panes, so a build that omits `env` from a session listing is still caught:

- `★ a paired device is refused on every pane route, for both brain marks` — table-driven over all six routes × both marks, asserted as one table so a partial regression reports every row. Status alone is not the assertion: it also proves `ptyProcess.write`, `resizeSession`, `lifecycle.destroy` and the git runner were never reached, and that `/api/stream` registered no SSE client.
- `★ the operator still streams a brain pane` — the constraint that made this non-trivial, asserting a `snapshot` frame actually arrives rather than just a 200.
- `the operator still drives the other pane routes on a brain id` — input, diff, commands, resize and DELETE, each with its side effect proven to have run.
- `a device still streams and types into an ordinary worker pane` — positive control on every gated route, so an over-broad gate cannot ship green.
- `★ a brain pane raises no attention event on any wire` — neither the live stream nor the `/api/events` replay backlog may contain the brain's pane id or its pane-authored text.
- `a stream ticket is refused on a brain pane too` — and the same ticket still opens an ordinary pane.

Three of the six fail on `main` (the bug proofs); the other three pass on `main` and guard the operator path against an over-broad fix. Verified both directions by reverting the source and re-running.

## Follow-ups found, not fixed here

- **`POST /api/approvals/:id` / `GET /api/approvals`** resolve an approval id to a pane and write keystrokes into it, with no brain filter anywhere in the approvals pipeline. `approvalWire` ships `screenTail`, `question` and `toolInputSummary`. Not reachable for a brain pane today — `ClaudePtyBrainAdapter` sets `WMUX_HOOKS_TO_MAIN=1` so the hooks that create approval records are not armed for it — but that is a spawn-profile property of a different module, not a gate. Worth its own issue.
- `RemoteHostClient.ts` comments that `RemoteHost.token` "is always the operator credential … never a paired device's", but the pair-exchange path mints a **device** credential. No functional regression (remote attach picks panes from `/api/workspaces`, which is brain-filtered), but the comment is wrong.

## Verification

```
npx vitest run src/daemon/web            # 217/217 pass (18 files, 518+ tests)
npx vitest run src/daemon                # 2764 pass, 5 fail
                                         #   (all 5 pre-existing PowerShell OSC 133
                                         #    runtime failures on this box, unrelated)
npx tsc --noEmit -p tsconfig.daemon.json # clean
npx tsc --noEmit -p tsconfig.json        # clean
npx eslint <both changed files>          # 0 errors (10 pre-existing warnings)
```

https://claude.ai/code/session_01JFTfQwRqFaJqU1UaERfewm

## Summary by CodeRabbit

- **Bug Fixes**
  - Paired devices can no longer access the orchestrator’s internal pane through terminal streaming, typing, resizing, diff, command, or close actions.
  - Requests targeting the internal pane now receive the same response as requests for an unavailable pane, preventing pane discovery.
  - Alerts and notifications generated by the orchestrator’s internal pane are no longer sent to paired devices or retained in event history.
  - Operator access to the orchestrator’s internal pane remains unchanged.


openwong2kim:
Superseded: the per-credential-class gate on the six pane routes landed as #1401 (same design — `attachableSession`, operator keeps `getSession`, device gets `readableSession`), and the `broadcastEvent` brain filter from this PR is ported in #1403 with a regression test. Closing so the conflicting rebase is not needed; both halves are on `main` or in review.

PR #1409, opened 2026-09-18
feat(remote): desktop notifications for remote agents

### LINKED ISSUE #1344: [Feature]: Desktop notifications for remote agents (finished / needs input)
### Problem / motivation

A local agent that finishes or waits for input raises a desktop notification. A remote agent never does, so on a desktop attached to a remote host you only find out by looking at the pane.

Local notifications come from `DaemonNotificationRouter`, which listens to the local daemon's session events (`session:agent`, `session:idle`, `session:notification`, …). Nothing in `src/main/notification/` handles remote agents. Remote agent status reaches the desktop only through the renderer polling `/api/workspaces` every 10 s (`useRemoteAttachmentsLifecycle`, `POLL_INTERVAL_MS`), which has no event channel; the roster even notes it has "no unseen-attention state".

### Proposed solution

Two ways to get the signal, and I'd like a direction call:

- **A. Subscribe to the host's `/api/events` (preferred).** The host already streams attention entries and `agent.liveness` over SSE for the phone client. The main process would hold one subscription per attached host and route agent transitions into the existing toast path, with the host label in the notification. It's event-driven, so there's no poll delay, and it reuses a channel that already exists.
- **B. Derive transitions from the existing 10 s poll.** No host-side change, but notifications arrive up to 10 s late and a short-lived state between polls is missed.

Either way, remote notifications go through the same settings and dedup as local ones.

Direction questions:

- Is the desktop subscribing to `/api/events` intended use, given it was built for the phone client? And is the attached host's token the right credential for it?
- Should remote notifications follow the global notification settings, or get a per-host toggle?

### Alternatives considered

- **Status quo:** watch the remote panes, or rely on the phone client's notifications.

### Additional context

Pairs with the remote Fleet View request: both read the same remote agent state. Happy to send the PR once the direction is agreed.

### Confirmation

openwong2kim:
Closes #1344

Option A from the issue: the main process subscribes to each attached host's `/api/events` SSE stream and routes agent transitions into the existing desktop notification path.

## Direction calls

**Main, not the renderer.** The attach roster that decides which hosts need a subscription is already main-owned and persisted (`RemoteAttachmentsStore`), the bearer token lives in main and structurally never crosses into the renderer (`RemoteHostPublic`), and `dispatchNotification` is a main-process function. A renderer-side subscription would need the token, would die on every Cmd+R, and would have to bridge every event back over IPC to reach the same dispatcher. So this reuses the existing auth path with no new plumbing: the same `Authorization: Bearer <host.token>` header `RemoteHostClient` already uses for `/api/stream`.

**Attention channel only, not `agent.liveness`.** The daemon delivers `agent.liveness` only to principals registered as transcript watchers — which happens by reading a pane's turn view. That is a per-pane phone affordance, not a fleet signal, so subscribing to it would mean a turn-view fetch per remote session. The recorded attention kinds (`critical`, `notify`, `approval`) already carry both halves the issue asks for: "needs input" (critical / approval) and "finished" (notify, the same OSC/agent notification path whose local twin is `session:notification`).

**Global settings, no per-host toggle.** Everything goes through `dispatchNotification`, so remote events honour the renderer's notification policy, the mirrored per-category mute, idle suppression and the toast dedup exactly as local ones do — one setting, one behaviour. `notify` maps to `terminal` and `critical`/`approval` to `approval`, matching the local categories for the same events.

## How it works

- `RemoteAttentionSubscriber` — one long-lived `/api/events` stream per **attached** host (a registered-but-not-attached host is one the user is not watching). Reconnects with jittered backoff (1s → 60s cap) and, unlike the per-pane `RemoteHostClient`, never gives up: a host asleep overnight must start notifying again without a re-attach.
- `remoteAttention.ts` — the pure derivation, split from the transport so the rule is testable without a socket. `formatRemoteAttention` mirrors `frontend/attentionFormat.js` so the desktop and the phone say the same thing about the same event; approval *echoes* (`tier: 'info'` — resolved/expired/superseded) are dropped, since notifying about an answered question is noise.
- **No storm on reconnect.** Every connect deliberately sends no cursor (no `Last-Event-ID`, no `?since=`), so the daemon always answers with a `reset` frame naming its head id followed by the whole replay window. Everything at or below that head id existed before the connection and is suppressed — a reconnect behaves exactly like a first connect. Cross-connection `epoch:id` dedup covers the pane-stream tee of the same event. Accepted cost, stated in the code: an event raised while the desktop was disconnected is not banner-ed when the stream returns. A desktop sitting next to its roster is not a phone in a tunnel; a banner burst on every tailnet blip is the worse failure.
- Lifecycle: `syncAttentionSubs()` reconciles subscriptions against the attach roster on attach add/remove, on host removal, and once at registration (so they survive an app restart), and the disposer stops them all.
- The notification carries `ptyId: null` and no `workspaceId`: the event names a *remote* session id, and passing it as either would make the toast click jump to whatever local surface happened to share it.

## Tests

`src/main/remote/__tests__/remoteAttention.test.ts` (5 tests, no network — a fake `Response` whose body yields canned SSE chunks):
- kind → notification vocabulary, including body-only `notify` and the dropped approval echo
- initial snapshot suppressed, first post-head event fires, same `epoch:id` never fires twice
- reconnect replays the window and re-fires nothing, then delivers the next live event
- malformed frames and unknown event names ignored
- end-to-end through the subscriber: the stream drops, the backoff timer fires, the reconnect notifies only for the genuinely new event; asserts the URL and the `Bearer` header

Gates: `tsc --noEmit`, `eslint` on the changed files, the new suite and the existing `remote.handler` suite (83 tests) all pass.


openwong2kim:
Review round (2-model panel: Claude + GLM 5.2; Codex unavailable — workspace deactivated, HTTP 402). Fixes pushed in 3258fb7e.

**Agreed by both models**
- Replay gate could be bypassed: an event with no numeric `id` (or no `epoch`) skipped both the boundary check and the dedup, and a reset with no usable `headId` left the boundary at 0 so the whole window fired as live. The gate now starts **closed** and only the first reset naming a usable head id opens it; a later reset can no longer raise the boundary and silently swallow live events; an unplaceable event is dropped.
- No cap on the SSE frame buffer — a peer that never sends a separator grew main's heap without bound. Capped at 256KB, then abort + backoff.
- Unread response bodies were never cancelled on the non-ok / early-return paths, leaking sockets behind an unbounded retry loop.
- `stop()` then `start()` on one instance could leave the aborted run's continuation alive and split into two loops. Generation counter on every async resumption point.
- No rate limit on remote notifications. Per-host token bucket (5 burst, 1 per 4s).

**Solo, accepted**
- (Claude) **Misattribution, the most consequential one.** `dispatchNotification` with `ptyId: null` and no `workspaceId` falls through `resolveNotificationTarget` to the *active local workspace* — so a remote notification would flash an unrelated local pane, jump there on click, and be silenced by that workspace's `notificationsMuted`. The remote path now uses the two gates main owns outright (mirrored per-category mute + ToastManager's `toastEnabled`/focus suppression) instead of the local-surface machinery it cannot honestly feed.
- (Claude) A zombie connection — TCP alive, no bytes, after a laptop sleep or NAT rebind — parked `read()` forever and the host silently stopped notifying until an app restart. Idle watchdog at three missed 25s heartbeats.
- (Claude) 401/403/404 retried like a network blip forever. Demoted to the slowest backoff step.
- (Claude) Backoff reset on a heartbeat comment let a host that only comments sit in a 1s reconnect loop. Reset moved onto real event frames.
- (Claude/GLM) CRLF frame and line separators, per the SSE spec, in case a proxy normalises newlines.
- (Claude) Remote-supplied `title`/`body` went into an OS toast unbounded and with newlines intact, which also let a title forge the ` · ` host-label prefix. Stripped and capped (120/240), matching what `RemoteHostClient` already does to remote strings.

**Rejected, with reasons**
- (GLM) *Subscriber caches a stale host record.* Not reachable: there is no in-place host mutation — `store.add` mints a new id and removal cascades through `syncAttentionSubs`. Same property the existing `clients` cache relies on.
- (Claude) *Disposer leaks subscriptions and the will-quit listener.* It already stops and clears `attentionSubs`, and `app.removeListener('will-quit', onWillQuit)` predates this PR.

Tests grew to 10 (from 5): fail-closed boundary, dropped unplaceable events, text sanitising, 401 backoff, and no delivery after `stop()`. `tsc`, `eslint` and the notification + remote.handler suites (212 tests) pass.

PR #1413, opened 2026-09-18
feat(browser): continue a truncated snapshot with a cursor instead of re-reading the page

openwong2kim:
## Problem

`browser_snapshot` and `browser_smart_snapshot` truncate at their length budget
(50 000 characters by default) and, past that point, the agent had exactly two
ways forward — a bigger `maxLength`, or a narrower `selector`/`q`. Both
**re-capture the page**. The page moves under a re-capture (a modal closes, a
list re-renders), and every re-capture opens a new ref generation, so reading
the second half of a big page cost the refs of the first half. Reading a large
settings page or a long result list meant snapshot → act on what fits → snapshot
again → discover the refs moved.

## Design

A **continuation cursor** over the SAME capture.

- When a result does not fit one window, the whole text is stored once, per
  connection scope, under an opaque `captureId`, and the window ends with:

  ```
  … truncated at line 812 of 2804. Pass cursor:"NGYzYT…OjgxMg" to continue this same capture (no re-read).
  ```

- Both tools gain `cursor?: string`. With a cursor the handler returns **before
  anything touches the page**: the next window of the same stored text, closing
  with the same trailer if more remains, or `(end of capture)` when exhausted.
  No page read, no CDP, no `beginRefGeneration` — the refs in every window are
  the ones that one capture minted, so `browser_click`/`browser_fill` resolve
  them exactly as they resolve the first window's.
- **Line granularity.** A snapshot line is one node with its ref, so a window is
  never cut mid-line; the budget bounds the window, and a line that is alone
  wider than the budget is emitted whole rather than split (a cursor that cannot
  make progress is worse than one oversized window). Such a lone line stores no
  capture — there is no second window to page to.
- **The budget has two halves and a line must fit both**: characters, because
  that is what the snapshot's own `maxLength` counts, and UTF-8 bytes, because
  every result then passes the dispatch-layer cap (`src/mcp/resultCap.ts`), which
  elides the *middle* of an over-cap text block. A characters-only window of a
  CJK page would come back with a hole in it while its trailer still promised
  continuity from the line it ended on — lines no cursor could reach, and an
  `(end of capture)` that said otherwise. Sizing the window to the byte cap up
  front means the elision never fires. `browser_smart_snapshot` declares
  `maxBytes`, so its windows follow the ceiling the caller asked for.
- **Never diffed, never combined.** A continuation neither reads nor writes the
  diff baseline — a window of a tree the caller is part-way through is not a new
  observation to compare. `format`/`selector`/`filter`/`q`/`full`
  (`maxContentLength`/`full` on the smart tool) cannot change text that already
  exists, so they are reported as ignored on one line instead of being silently
  honoured or silently dropped.
- **Invalidation.** Captures live in a sibling store in `snapshotCache.ts` with
  the same discipline as the diff baselines — per connection scope with a
  module-global fallback, an entry cap, a 5-minute TTL. A capture is retired by:
  any new snapshot of that surface from either tool (that snapshot re-mints
  refs, which is exactly what would make the older capture's numbers lie); the
  navigation/close invalidation the lifecycle drain in `automationLease.ts`
  already performs; or the TTL, which is an **idle** timeout so a walk in
  progress does not expire under the agent. A cursor into a retired capture
  returns an error result whose text starts `cursor_expired:`.
  The post-body drain keeps a capture that carries no URL of its own — where the
  baseline side deliberately fails closed — because fail-closed there deleted
  the capture the very call that minted it had just stored.
- **One cut, not two.** `generateSnapshot`/`generateScopedSnapshot` gain
  `deferTruncation`, set only by the two tools: the cut has to happen once,
  after assembly, where the window offsets are counted. Every other caller
  (replay, the snapshot runner) keeps the hard cut and its `... (truncated)`
  marker, and the per-frame `FRAME_TRUNCATED_NOTE` budget is untouched. Both
  tools then cut the assembled text to the capture ceiling (1 MB, at a line
  boundary) through `capCaptureText`, which leaves a line naming what it
  dropped — so the diff baseline and the repl listing stay bounded as they were
  before, and the last window is honest about the tail.
- The overflow retry that replaces the tree with its interactive-only strip used
  to be silent. For a hard-cut caller that is fine — its result ends in
  `... (truncated)` and claims nothing — but a capture the agent pages to
  `(end of capture)` does claim completeness, so on that lane the strip says so.

`scripts/mcp-protocol-baseline.json`: the `full` profile's `wireResultSha256` is
re-pinned for the two new `cursor` parameters (`tools/list` is 76 726 bytes
against the 80 000 budget).

## Verification

- `npx tsc --noEmit` — clean.
- `npx vitest run src/mcp/playwright/__tests__/snapshotCursor.test.ts src/mcp/playwright/__tests__/inspection.snapshotCursor.test.ts src/mcp/playwright/__tests__/extraction.smartSnapshotCursor.test.ts` — 32 passed.
  Covers: truncation → cursor → continuation → exhaustion; the windows
  reassembling byte-exactly (line granularity at the boundary, no ref delivered
  twice or half); a ref absent from the first window arriving verbatim later;
  zero page reads during a walk; a CJK walk through the real result-cap wrapper
  with the guard's elision marker never appearing; the byte half of the budget
  and `maxBytes` raising it; `capCaptureText` cutting at a line boundary with a
  note; a lone over-budget line storing no capture; cursor + the other
  parameters excluded with a note; the diff baseline left intact; the smart
  tool's caveats landing in the first window; invalidation on a new snapshot, on
  a drained `navigated` event, and on the TTL; the TTL behaving as an idle
  timeout across a slow walk; the store's entry cap and eviction; and
  per-connection isolation (two scopes never see each other's captures, same
  token shape).
- `npx vitest run src/mcp` — 1655 passed, 3 skipped.
- `npm run build:mcp && npm run probe:mcp` — passes with the re-pinned hash.
- `npm run test:parallel` — 16 329 passed, 12 failed, 32 skipped. The failures
  are the same nine files before and after this branch, all environmental on
  this machine: symlink `EPERM` (settingsFile, SiteGuideStore ×3,
  WebTerminalServer turns/image ×2, worktask.rpc, setupStatusline), a missing
  `node_modules/@xterm/*` in the worktree (atlasCoherence, imeAnchor), a stale
  `THIRD_PARTY_NOTICES` relative to `package-lock.json` (generateNotices — no
  dependency changed here), and one flaky `killVerifiedDaemonPid.crossHost`.
- `npx eslint` on every changed file — no errors, and no warnings in the lines
  this branch touches.
- `/code-review high` was run on the diff; all eight findings are addressed in
  the second commit.

## Not covered

- **No live dogfood.** The app was not launched, so the walk was exercised
  through the tool handlers on the RPC lane (DOM listing) and the pure store,
  not against a real Chrome page.
- **A continuation trusts the lifecycle drain for navigation.** It never reads
  `page.url()` itself, because that is the page access the feature exists to
  avoid. A navigation that never reached the ring (a drain won by another
  consumer) is caught by the next snapshot or the TTL rather than immediately;
  the refs in such a window then fail loudly at `resolveRef`'s own navigation
  guard rather than clicking the wrong element.
- **The capture is whatever the snapshot produced**, including the
  interactive-only strip and per-frame truncation on a very large page. Both are
  now stated in the capture rather than silently applied, but a cursor cannot
  recover what the serializer never emitted — a `selector` or `q` re-read still
  can.
- **`maxLength` is still not a `browser_snapshot` parameter**, so its window is
  the 50 000-character default; the cursor is the way to read more.


openwong2kim:

PR #1414, opened 2026-09-18
feat(browser): machine-readable effect_state on every mutating browser result

openwong2kim:
## Problem

A mutating browser tool reports its outcome as prose, and prose is the one thing
an agent cannot branch on after a timeout. `locator.click: Timeout 30000ms
exceeded` does not say whether the click reached the page, so the agent either
retries — a second submit on a form that already posted — or abandons a flow that
actually worked. Two tools already had to warn about exactly this in words
(`browser_file_upload`'s measured transfer timeout, `browser_download`'s
post-click timeout that also strands the tab); nothing was machine-readable, and
nothing covered the other twelve mutating tools.

## Design

New module `src/mcp/playwright/resultTrailer.ts`. The tools that put input on a
page now end every result — success and `isError` alike — with:

```
effect_state: none | committed | unknown
error_code: <code>            (error results only)
```

- `none` — nothing was dispatched (ref/selector not found, scope refused, invalid
  params, backend unsupported). A retry is free.
- `committed` — the input reached the page: the Playwright call or the RPC lane
  returned.
- `unknown` — a dispatch went out and the outcome cannot be read from here
  (timeout after dispatch, transport loss mid-action). That error text also
  carries one sentence: "Inspect the page before retrying — the action may
  already have taken effect."

The state is **structural, not guessed**. Each tool body runs its dispatch sites
through an effect probe, so a failure raised before any input went out is `none`
whatever its wording, and one raised after it is `unknown`. A branch that can
prove otherwise declares the state itself — the RPC `browser_select` lane, whose
page evaluation reads the ref, finds it gone and changes nothing, reports `none`
even though it did evaluate.

`leasedMutation` (automationLease.ts) owns the probe, because
`withAutomationLease` settles the workspace scope *before* it calls the body: a
scope refusal rejects outside the body's own `catch`, and that was the one
failure shape these tools would have advertised a trailer for and answered
without one.

`error_code` is a closed set of twelve, exported once with its type. It is tagged
at the failure site wherever wmux raises the error itself (`taggedFailure`, kept
non-enumerable so it never reaches a log line), and pattern-matched only for the
errors wmux does not author: Playwright's, the RPC transport's, main's refusals.
An unclassifiable failure is `unknown_error` — never a code outside the set.

Trailered (14): `browser_click`, `browser_type`, `browser_fill`,
`browser_press_key`, `browser_hover`, `browser_drag`, `browser_select`,
`browser_navigate`, `browser_navigate_back`, `browser_file_upload`,
`browser_download`, `browser_scroll`, `browser_scroll_into_view`,
`browser_dialog`. Each description gains one short sentence pointing at the
trailer. Read-only tools (snapshot, screenshot, extract, console, network,
cookie/storage reads, waits) carry none — they have nothing to report, and a
trailer there would teach the agent to read one where it means nothing.

Existing human-readable text is unchanged, except where it contradicted the
trailer: the RPC fill lane's "Nothing was typed" now admits that the focusing
click did reach the page, and the upload timeout drops its own "check the page
before retrying" half (the trailer carries that sentence; the duplicate-upload
warning stays). The trailer is appended to the **last** text block, so it stays
last whatever the automation lease prepends (`[browser events]`, replay/skill
hints).

One behavior fix fell out of the contract: `browser_scroll`'s RPC ref lane asked
the page whether the ref still existed and discarded the answer, reporting
`Scrolled down by 500px (element ref=7)` for a ref that was gone. Its sibling
`browser_scroll_into_view` already checked that sentinel; now both do, so the
trailer cannot certify a scroll that never happened.

`browser_replay` is unaffected — it drives Playwright itself and never reads a
tool result — which two new tests pin.

## Verification

- `npx tsc --noEmit` — clean.
- `npx vitest run src/mcp/playwright/__tests__/resultTrailer.test.ts` — 35 passed
  (helper placement/idempotence, the closed code set, the classifier, the probe).
- `npx vitest run src/mcp/playwright/__tests__/interaction.effectTrailer.test.ts`
  — 12 passed: `none` / `committed` / `unknown` through the real `browser_click`,
  `browser_fill`, `browser_press_key` and `browser_scroll` handlers with fakes,
  including a scope refusal raised by the lease (before the body runs), one
  raised by the page lane, and an RPC transport loss mid-click.
- `npx vitest run src/mcp/playwright/__tests__/file.effectTrailer.test.ts` — 7
  passed (`browser_download` committed/unknown/none, `browser_dialog`, and
  `browser_wait_for_download` carrying no trailer).
- `npx vitest run src/mcp/browser-replay/__tests__/replayRunner.test.ts` — 57
  passed, two of them new.
- `npx vitest run src/mcp` — 1679 passed, 2 skipped, 0 failed.
- `npm run build:mcp && npm run probe:mcp` — passes with the refreshed `full`
  baseline hash; `wireResultBytes` 76918 of the 80000 budget, `core` and
  `commander` unchanged.
- `npm run test:parallel` — 16351 passed, 12 failed, 32 skipped. All 12 are
  pre-existing on this machine and unrelated: the 8 symlink-privilege ones
  (settingsFile, setupStatusline, SiteGuideStore ×3, WebTerminalServer turns/image
  ×2, worktask.rpc) plus the 4 known flakes (generateNotices drift guard,
  atlasCoherence, imeAnchor, killVerifiedDaemonPid crossHost).

## Not covered

- Tools that change something other than the page's own state are outside the
  fourteen: `browser_evaluate`, `browser_tabs`, `browser_open`/`close`,
  `browser_emulate`, `browser_resize`, `browser_highlight`, and the cookie and
  storage writes. `browser_evaluate` is the most timeout-ambiguous of them and
  would be the obvious next one; `browser_tabs` already answers with its own
  structured codes. The module docstring names this gap rather than implying
  coverage.
- A partly filled form reports `effect_state: committed` and no `error_code` —
  the three states have no "partly", and the `Filled 2/3 field(s).` count plus
  the `Errors:` block remain the per-field truth.
- No live dogfood: every check is unit-level against fakes. No real agent has yet
  read the trailer against a real page.
- The MCP baseline hash is correct only for this branch's final state; if another
  description-changing PR lands first, this one needs the hash refreshed again
  (`scripts/mcp-protocol-baseline.json`, no `--update` flag — see
  `npm run probe:mcp`).

## Summary by CodeRabbit

* **New Features**
  * Browser actions now report machine-readable effect states: `none`, `committed`, or `unknown`.
  * Failed actions include standardized error codes, making retry decisions clearer.
  * Uncertain outcomes provide guidance to inspect the page before retrying.
  * Read-only tools and browser replays remain unchanged and do not include effect trailers.

* **Bug Fixes**
  * Scrolling with a stale element reference now correctly reports that the element was not found instead of falsely indicating a scroll occurred.


openwong2kim:

PR #1416, opened 2026-09-18
feat(browser): browser_request_help hands one blocked step to the operator

openwong2kim:
## Problem

Login walls, CAPTCHAs, OTP fields, payment confirmations and consent screens
block every browser flow, and there was **no browser-side way to ask the
operator** — `grep` over `src/mcp/playwright/` confirms none of the ~35 browser
tools could hand a step back to a human. The orchestrator's `deck_ask_decision`
is a different surface: it asks the *brain* a question about the *work*. So an
agent facing a sign-in page either looped on it or gave up, and the human never
learned there was one keystroke between the agent and the rest of the flow.

## Design

`browser_request_help` opens one request for the caller's surface, brings that
surface's pane into view, and **waits**. It settles on one of four outcomes, and
the result's last two lines are machine-readable so the agent branches without
parsing prose:

```
help_state: completed | continued | cancelled | timed_out
url: <page url after>
```

`completed` = the caller's `completion` criteria held; `continued` = the human
pressed Done; `cancelled` = Cancel; `timed_out` = the deadline.

**Main owns the request.** `src/main/browser-session/HelpRequests.ts` holds the
deadline — a renderer that reloads or is never looked at must not be able to
leave a request open forever, and the caller's own timeout is a client-side guess
about a process it does not own. It also enforces one open request per surface
(a second call is refused with text starting `help_already_pending:`; a queue
would let an agent stack prompts the human answers in an order nobody chose),
answers another workspace's id *exactly* as it answers an unknown one (so this
cannot become an id-probe for other workspaces' requests), and runs the 500 ms
completion poll under a continuous-1s-hold rule so one frame of a redirect chain
cannot complete a request. The poller has an in-flight guard: a `Runtime.evaluate`
against a guest mid-navigation routinely outruns the 500 ms cadence — exactly
when a login flow is redirecting — and two overlapping probes resolve out of
order. Every page read is bounded (2 s), because two things hang off it: the
request's RPC reply (the client gives up at 10 s) and the row's removal at settle.
A wedged guest must cost the caller a missing `url`, never a dropped reply or a
Done button still on screen after the request is over.

**The tool polls; it never holds one long RPC.** `browser.help.status` is asked on
a ~1 s cadence inside `withAutomationLease` (which renews the lease through the
wait). The client's default request timeout is 10 s, so a five-minute call would
be dropped by the transport long before anyone answered.

**Plumbing** mirrors the permission-prompt trio exactly (`BROWSER_HELP_OPEN` /
`_RESOLVE` / `_CLOSED`), deliberately not the `RPC_COMMAND` path: the payload is
an agent-authored string with a two-button answer, so the channel carrying it
stays structurally incapable of reaching anything else. `external` workspaces
delegate opens to the OS browser and get `not_supported`; on `chrome`/`live`
there is no in-window webview, so the Chrome tab is raised via the backend's
`selectSurface` and an OS `Notification` carries the ask.

### DESIGN.md rules applied

- **Attention grammar, max two renditions.** The in-pane bar and the Fleet inbox
  row are the two. No titlebar chip, no modal, no third place.
- **Status dot vocabulary.** Red = needs input, on both renditions.
- **Not a 36px chrome row.** The bar sits in the inspector toast's slot as a
  transient ~30px band, so the page loses no lasting height (`minHeight: 30` is
  asserted in the RTL test).
- **Primary action = solid warm fill.** Done is the *one* `--accent` fill on the
  surface; Cancel is raised neutral, not the destructive treatment the approval
  rows' Deny wears — cancelling a help request abandons one step, not the flow.
- **`--surface-highlight`, not `--text-main`, for the inset on the accent fill**
  (`tokenDiscipline.test.ts`): a text-main mix inverts on the light themes and
  presses a groove into the button.
- **Typography.** The prompt is 13px **sans** — "someone talking", not machine
  evidence. No mono, no emoji; the glyph is a monochrome dot.
- **Radii 5px, hit areas at least 24x24, transitions 120 ms** — all asserted in
  tests.
- **Every claim one click from its evidence.** The inbox row carries a jump
  affordance to the browser pane.
- **i18n.** All six strings go through `useT`; `en`, `ko` and `pl` have entries
  (`pl` is gated by `plLocaleCoverage.test.ts`; the other locales fall back to
  `en` per `i18n/index.ts`).

The prompt is **agent-authored and untrusted**: control characters are stripped
in one shared place (`sanitizeHelpPrompt`) and it is rendered as a text child, so
it can never be markup. The bridge hook deliberately does **not** jump on open —
`focusNotificationTarget` switches the operator's active workspace and
auto-unstashes the target pane, which `useNotificationListener` documents as
right "here and only here: the user clicked… an agent rearranging the layout as a
side effect IS the surprise this feature exists to prevent." Main's non-yank
`surface.focus` brings the pane into view instead; the operator-initiated jump
lives on the inbox row.

## Verification

`npx tsc --noEmit` — clean.
`npx eslint` on every touched file — no new findings (the 8 in `BrowserPanel.tsx`
are pre-existing, in the inspector/webview code).

```
npx vitest run src/main/browser-session/__tests__/HelpRequests.test.ts \
  src/mcp/playwright/__tests__/help.test.ts \
  src/renderer/stores/slices/__tests__/browserHelpSlice.test.ts \
  src/renderer/stores/selectors/__tests__/approvalInbox.test.ts \
  src/renderer/components/Browser/__tests__/ \
  src/main/pipe/handlers/__tests__/browser.rpc.scopeCoverage.test.ts \
  src/main/mcp/__tests__/firstParty.test.ts
→ 9 files, 87 tests passed
```

New coverage: `HelpRequests` (create / status / resolve / cancel / timeout /
one-per-surface / workspace scoping / the completion hold-and-restart rule / no
polling without criteria), the slice plus `selectHelpRequestForSurface`, the inbox
selector's new source and its ordering, the tool with fakes across all four
`help_state`s plus `help_already_pending:` and `not_supported:`, and an RTL test
for the bar (Done/Cancel wiring, other-surface isolation, and that an
`` prompt renders as text with no nodes created).

`npm run test:parallel` — **1058 files passed / 9 failed; 16352 tests passed /
13 failed.** All 9 failing files are this machine's pre-existing environment
limits, none of them touched here: `settingsFile`, `setupStatusline`,
`SiteGuideStore`, `worktask.rpc` (symlink `EPERM`), `WebTerminalServer`,
`generateNotices` + `atlasCoherence` + `imeAnchor` (packages absent from this
worktree's `node_modules`), and `perfConfirmFresh` (its fixture pins the repo's
HEAD sha, so it fails on every feature branch).

`npm run build:mcp && npm run probe:mcp` — passes.
`scripts/mcp-protocol-baseline.json` gains the new tool name and a refreshed
`full.wireResultSha256`; the full profile measures **77,646 / 80,000** bytes, so
the tools/list budget is **not** raised (the description and param text were
deliberately trimmed to keep it under). `docs/api/reference.md` regenerated
(179 → 182 methods, three capability rows) — no version change, `package.json`
untouched.

## Not covered

- **UI verified by tests only.** No live dogfood was run — the app and daemon
  were never launched from this worktree. The lead owns that pass.
- `completion.selector` and `urlIncludes` are evaluated in the **top frame** of a
  builtin webview. A CAPTCHA widget or a 3-D Secure step usually lives in a
  cross-origin iframe, so a selector there will never match — documented in the
  tool schema and `inventory.md`, with `urlIncludes` named as the condition that
  works across those flows. No iframe traversal was attempted.
- `completion` does nothing on the `chrome`/`live` backends (no guest webview to
  read); Done, Cancel and the deadline all still work there.
- No **resync after a renderer reload**: main keeps the request pending but the
  bar and the row are gone until the deadline. `permissionPrompt` has the same
  gap, so this is consistent rather than new — closing both wants one channel,
  not two.
- A **timed-out** request leaves no trace in the Fleet inbox's auto-rejected log.
  That log is labelled "auto-rejected approvals" and reusing it would mislabel a
  timeout; giving help requests their own line is a follow-up.
- The OS notification's title/body on `chrome`/`live` are **English only** —
  main has no i18n layer, like every other main-side notification string.


openwong2kim:

PR #1417, opened 2026-09-18
feat(browser): confine Live Chrome writes to the agent's own tabs

openwong2kim:
## Problem

`docs/browser-backends.md` said of the `live` backend: *"There is no way to scope
it to one tab or one profile: the agent gets the browser."* That was accurate.
`LiveChromeClient.hasSurface()` returned true for every tab, and the ownership
map it kept (`tabOwners`) was read only to seed `browser.cdp.info` — never to
decide whether a write was allowed. So an agent bound to Live Chrome could
navigate, click, type into, evaluate JavaScript in, set cookies on or close any
tab in the user's real browser, including every logged-in session in it.

Binding a workspace to Live Chrome is a deliberate consent, and it plainly covers
reading: the point of the backend is an agent working inside a session the user
signed into themselves. It does not follow that it covers typing into the tab the
user happens to have open.

## Design

**Reads unchanged, writes confined.** A live target is agent-owned by workspace W
iff `tabOwners.get(targetId) === W`, or W borrowed it. Ownership is *exact*: a
tab opened for workspace A is a foreign tab from workspace B, refused the same
way one of the user's own is. A refused call fails before it touches the page,
with an error starting `agent_window_scope:` that names the tab and carries the
one remedy (`borrow it first: browser_tabs action:"borrow" surfaceId:"<id>"`).

**Both lanes, because neither covers the other.** `browser.rpc.ts` gates the
write RPCs (`navigate`, `goBack`, `click/type/press/hover/drag.cdp`, `evaluate`,
`cookies` set/clear, `emulate`, `resize`, `close`). The MCP Playwright lane
drives fill/select/upload/type over CDP without main ever seeing the call, so
`getPageForScope` takes an `intent` — default `'read'`, which leaves every
existing call site byte-identical — and the mutating tool call sites pass
`'write'`. Ownership is always *main's* answer: `browser.cdp.info` target rows
carry `owner: 'agent' | 'borrowed' | 'user'`, and a resolved live target absent
from that list is one the agent may read and must not drive. `browser.cookies`
is gated as a method but lets `action:"get"` through; `browser.type.humanlike` is
deliberately out of the set (it computes a keystroke schedule and touches no
page — the keystrokes it describes go through gated methods).

**Mixed mode.** A manually opened builtin pane still exists under this backend;
its surface is a webview, not a live tab, so the gate skips ids the
`WebviewCdpManager` owns (exact-match, plus a discarded guest the handler is
about to wake). Both are pure lookups: no wake, no dispatch before the refusal.

**Lending.** `browser_tabs action:"borrow" surfaceId:"<id>"` asks the human
through the *existing* MCP permission pipeline — the same `ApprovalQueue`, the
same dialog, the same Fleet approvals inbox row, no new UI — reading
`Agent in workspace <name> wants to control tab "<title>" (<origin>)`. 60 s
deadline; timeout is a deny. The three refusals stay distinguishable because the
agent's next move differs: `user_denied:`, `borrow_timeout:`, `borrow_pending:`
(one request per tab, so prompts cannot be stacked on the user). `action:"return"`
hands a tab back; a Chrome disconnect, `dispose()` or a profile-binding change
clears every grant. A borrow prompt writes **no** plugin-trust record: it is a
question about a tab, not a standing decision about a client.

**`browser_tabs list`** labels every row `agent` / `borrowed` / `user` and takes
`scope?: 'agent' | 'user' | 'all'` (default `'all'` — today's behaviour).

**Window grouping is best-effort; ownership is not.** The first tab a workspace
opens asks for `newWindow: true` and records the window via
`Browser.getWindowForTarget`; later opens activate one of that workspace's own
tabs first so Chrome places the new one beside it, then record where it actually
landed. CDP has no command that moves an existing tab between windows, so a
stray placement is recorded rather than corrected — and the gate never consults
the window.

**Setting.** `liveWriteScope: 'agent' | 'all'` in `browser-backend.json`, default
`'agent'`. `'all'` restores the old behaviour and is the *larger* grant (every
session in the browser, not one window's worth), which is why it has no Settings
toggle — it should cost a file edit. Under `'all'` the gate is off, so `borrow`
asks nobody and `scope:"agent"` filters nothing.

Clean-room: implemented from a behavioural description. The credit comment sits
where the policy is defined (`src/shared/liveWriteScope.ts`).

### Second commit

Six findings from `/code-review high`, each a real hole: `browser_evaluate` and
`browser_replay action:"run"` resolved their pages with read intent (and replay
swallowed the refusal into "no live page"); `borrow` rode `sendRpc`'s 10 s
default against a 60 s prompt, capping the human's answer at ten seconds;
per-workspace grant clearing had no caller, so grants outlived the binding;
`'all'` still prompted and still filtered; and `onDisconnect` never fired on the
endpoint-moved path, so a Chrome restart kept every grant.

## Verification

`npx tsc --noEmit` clean. `npm run test:parallel`: **16435 passed, 11 failed** —
all 11 pre-existing on this box and unrelated to the diff (8 symlink-`EPERM`:
`settingsFile`, `setupStatusline`, `SiteGuideStore`×3, `WebTerminalServer`
turns/image×2, `worktask.rpc`; 3 needing a worktree-local `node_modules`:
`generateNotices`, `atlasCoherence`, `imeAnchor`). `npm run probe:mcp` green
after updating the `full` profile hash in `scripts/mcp-protocol-baseline.json`
(76804 / 80000 bytes).

New tests:

- `LiveChromeClient.scope.test.ts` (16) — exact ownership, borrow lifecycle
  (pending once / granted / returned / denied), clear on dispose, on Chrome
  disconnect and on a Chrome *restart* (endpoint moved), per-workspace clearing,
  window grouping incl. a fake CDP that reports a **different** window, a Chrome
  that refuses `newWindow`, one that will not answer `getWindowForTarget`, and
  the `owner`-labelled `cdpInfoTargets` rows.
- `browser.rpc.liveScope.test.ts` (78) — every write RPC × (own tab passes /
  borrowed passes / user tab refused before dispatch / another workspace's tab
  refused), reads on a user tab still pass, cookie `get` vs `set`, no-surfaceId
  writes not gated, builtin and discarded surfaces not gated, dedicated Chrome
  not gated, `'all'` bypass (write, borrow, list, close), `list` scope filtering
  and labels, borrow approve/deny/timeout/pending-twice, return, and the
  `cdp.info` policy + owner rows.
- `PlaywrightEngine.liveWriteScope.test.ts` (12) — write refused / allowed /
  borrowed allowed, read untouched and free of the extra round trip, `'all'`
  bypass incl. a policy switched since the cache, refusal when the target id
  cannot be read or main cannot be reached, and that `allowScopedRpcFallback`
  re-raises it instead of falling back.
- `liveBorrowApproval.test.ts` (9), `ApprovalQueue.test.ts` +5 (consent writes no
  trust record), `BrowserBackendStore.test.ts` +6 (setting defaults fail-safe),
  `navigation.tabs.test.ts` +12 (tool-side contract, owner rows, borrow deadline).

## Not covered

- **Real Chrome 144 live attach was not exercised.** Everything here is unit
  tests and fakes; the lead does the live dogfood. Specifically unverified
  against a real browser: whether `Target.createTarget({newWindow:true})` plus
  `activateTarget` actually groups tabs the way Chrome 144 documents, and whether
  the permission dialog and the borrow prompt interact well on screen.
- Each gated write costs one extra `browser.cdp.info` round trip and a throwaway
  CDP session. A cached owned-id set would remove it, and would have to be
  invalidated on borrow/return/open/close from a process that sees most of them
  second-hand — a stale cache there hands an agent a tab the user took back.
  Measured need first.
- A grant follows the **tab**, not the origin the prompt showed. An agent that
  navigates a lent tab keeps writing to it; that is how a flow proceeds through a
  site, so it is documented rather than fixed.
- No Settings UI for `liveWriteScope`, deliberately (see Design).
- `browser_tabs select` is not in the write set: it moves UI focus, and was
  already permitted on any live tab.


openwong2kim:

PR #1433, opened 2026-09-20
feat(web): serve transcript media over a streamed /turns/file route

openwong2kim:
## What

`GET /api/sessions/:id/turns/file?path=<absolute>` — the bytes behind a path the transcript named, for the files a phone can *play* as well as the ones it can render.

`/turns/image` is PNG/JPEG/GIF/WebP only, so an mp4 or mov an agent produced (a screen recording, an ffmpeg render) answers 415 and the phone can only show a filename chip. This adds the route beside it. **`/turns/image` is not touched** — shipped iOS builds depend on it, and on its promise that what comes back renders as an image.

Written to a contract from the wmux-ios workspace; the three points it left open were agreed with that side before implementation (tag strings, gate order, shrink guard).

## Gate

Copied from `handleSessionTurnImage`, not extracted into a shared helper — rewiring the old handler to reach a new abstraction is a change to it, whatever the diff says about behaviour. Same `--allow-transcript` / `transcript-disabled:` prefix, same roots (the pane's **spawn** cwd ∪ uploads dir, never `meta.cwd`, which OSC 7 lets any pane process move), both sides realpath'd, `path.relative` containment (never a string prefix), `O_RDONLY | O_NOFOLLOW | O_NONBLOCK`, `isFile()`, and one handle from the gate to the last byte.

## What differs from the image route

**Streamed, not buffered.** A 128 MiB cap behind `Buffer.allocUnsafe(stat.size)` is one request that can hold 128 MiB. Two details the shape forces:
- `autoClose: false` on the read stream — otherwise it closes the FileHandle at `end` and the growth probe reads a closed descriptor on every *successful* transfer.
- `pipe(res, { end: false })` + `await finished(stream)` — the probe, not pipe, decides whether the response ends cleanly or is cut.

**Sniff before cap**, because which cap applies (8 MiB image / 128 MiB video) is a fact about the type. Consequence, agreed with the client side: **a 200 MB text file is `415 unsupported-type`, not `413`.** Saying "too large" about a file that would be refused at any size is a lie the client would render as a retry.

**New tags** — `bad-file-ref`, `file not found`, `unsupported-type` — rather than reusing the image route's. The client keeps a separate error map per route, and shared strings would make a 404 in the logs ambiguous about which route produced it.

## Two guards the buffered route did not need

- **Client walks away.** `pipe` unpipes on a closed response but never destroys its source, so an aborted download would strand `finished()` for ever with the handle held. `res.once('close', …)` destroys the stream.
- **File shrank.** The contract covered growth only. Fewer bytes than the `Content-Length` already promised leaves URLSession waiting for a remainder that is never coming, so a short transfer cuts the socket the same way a grown one does.

Note on the growth probe: the stream is bounded by `end: stat.size - 1`, so growth can never ride along behind the approved bytes. The probe's job is to notice and cut rather than finish a response whose file is no longer what the gate approved.

## Also

- `/api/config` gains `turnFiles`, beside `turnImages`, under the same grant and **omitted rather than false** — the shape a daemon predating the route serves, so a client decides once per connection instead of learning it from a 404 per file.
- `docs/phone-client-contract.md` gains the matching section.
- `qt  ` stays split from `video/mp4` → `video/quicktime`. The client names its cache file from the Content-Type, and a QuickTime movie saved as `.mp4` will not open.
- Ranges are not supported and `Accept-Ranges` is not advertised (the client downloads, then plays locally).

## Verification

- `tsc --noEmit` clean; `eslint` adds no new warnings.
- `vitest run src/daemon/web/__tests__/WebTerminalServer.test.ts` — **255 passed**, including every pre-existing `/turns/image` case unchanged (that describe block has a zero-line diff).
- 16 new cases: the brand table (mp4/mov/m4v/png, `qt  ` split out), 415 for text, 415-not-413 for an oversized unsupported file, both caps, boundary escape via symlink + FIFO + string-prefix sibling + directory + wandered `meta.cwd`, 400 refs, unknown pane and brain pane, uploads dir, `turnFiles` in `/api/config`, the shrink cut, the growth bound, and a mid-download abort that leaves the server answering.
- Streaming is proved at the handle, not by RSS: server and client share one vitest process, so RSS counts the client's own chunks. The test wraps the FileHandle and asserts the largest single `read` is one stream chunk (64 KiB), never the 100 MB a buffered route would ask for, with `createReadStream` called exactly once.

Fixture note: the 100 MB / 128 MiB files are `truncate`d sparse, so they cost no disk and no wall clock.

## Summary by CodeRabbit

* **New Features**
  * Added support for viewing MP4 and QuickTime transcript videos in the phone turn view.
  * Eligible media files can be streamed from supported transcript-generated locations alongside existing transcript images.
  * Added support for fragmented MP4 videos produced by streaming tools.
  * Video files support sizes up to 128 MiB; image limits remain 8 MiB.
  * Availability is reported in configuration when transcript access is enabled.

* **Documentation**
  * Documented supported formats, access requirements, size limits, and streaming behavior.


openwong2kim:
Two findings from the review panel are **not** addressed here, on purpose — both are inherited from `/turns/image` rather than introduced by this PR, and fixing either means touching a route with shipped clients:

- #1434 — `O_NOFOLLOW`/`O_NONBLOCK` are POSIX-only, so both media routes lose the link-swap and FIFO guards on Windows without saying so.
- #1435 — no response idle timeout anywhere on this server, so a client that stops reading holds this route's file handle for as long as it holds the socket.

Everything else the panel raised is in the last commit.

